PR #642, opened 2016-07-11
Create CONTRIBUTING.md

Manouchehri:
There's a lot of bug reports where people don't even mention what build they're running.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,29 @@
+When reporting a bug, please be sure post the following information:
+
+1. The build of Windows you're running. Example:
+```cmd
+C:\Users\Dave>ver
+
+Microsoft Windows [Version 10.0.14385]
+```
+
+2. The exact command you tried to run. Example:
+```cmd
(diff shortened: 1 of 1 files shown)


benhillis:
This is a good suggestion, I will work with @russalex on getting this added.

Chnkr:
I would strongly suggest using an issue template as well, see: https://github.com/blog/2111-issue-and-pull-request-templates

Manouchehri:
Was about to suggest that as well; wasn't sure how you guys would want it
formatted though.

sundhaug92:
A few suggestions:
1. Indent the code-blocks in the example so that the numbering continues in the list
2. Add a point for "expected behaviour", especially if there exists links to documentation/specification for "the Linux way"
3. Change the text "the exact command" to "Steps required to reproduce", some bugs are not triggered by commands but rather by code written by the user themselves.
4. Link people to a [manual to write good bug-reports](https://musescore.org/en/developers-handbook/how-write-good-bug-report-step-step-instructions), to help people consider the how, why and whats of bug-reporting. 
5. Request that people tag their bugs so that it's easier to sort and filter through.

PR #859, opened 2016-08-11
Added link to stackoverflow with tag

SRGOM:
Already existing and in use tag `WSL` is somewhat unique so makes a much better search query as @aseering has in mind. Added to number 4 because I feel like its more important than the rest of the things for an end user.


benhillis:
Thanks for the suggestion, I've made the change with some minor edits.

PR #907, opened 2016-08-16
Added a small FAQ section

SRGOM:
Please feel free to reject if this is not the right place


benhillis:
Thanks for the suggestion. We already have a FAQ page linked on our ReadMe.

PR #1837, opened 2017-03-31
Make markdown headers render right CONTRIBUTING.md

VoLuong:


benhillis:
Not sure how that happened, thanks for noticing!

PR #1923, opened 2017-04-15
Tweak the issue template and contributing guidelines

aseering:
* Drop some questions that seldom generate useful information
* Re-order remaining questions to try to guide users into providing the most-useful details
* Elaborate on how to gather requested information
* Link directly from the Issues template to the UserVoice and to security@microsoft.com, to direct people to the appropriate resource
* Update the Contributing instruction text for Creators

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,7 @@
 Please provide as much information as possible when reporting a bug or filing an issue on the Windows Subsystem for Linux.
 
+This Github issues page is a bugtracker.  It's intended for technical discussion and for debugging specific issues.  If you have a feature request, you should instead post it to the [UserVoice](https://wpdev.uservoice.com/forums/266908).  The UserVoice is better for tracking feature requests; it has a voting mechanism so the WSL team can put a higher priority behind issues that affect more people.
+
 Do not open Github issues for Windows crashes (BSODs) or security issues.  Please direct all Windows crashes and security issues to secure@microsoft.com.  Issues with security vulnerabilities may be edited to hide the vulnerability details.
 
 A well written bug will follow the template:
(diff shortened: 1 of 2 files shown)


aseering:
I don't necessarily expect this to get accepted as-is.  But the existing template hasn't been revised in a while.  I feel like some of the questions on it seldom end up being useful; others are almost always useful, but they're confusing and near the bottom of the list.  Plus people get confused about whether they even should be posting here, etc.  So I took a shot at refactoring to account for those observations.  Thoughts welcome.

jackchammons:
Looks great. I like the changes/updates. Merging. Thanks!

therealkenc:
The suggested `strace` command in the boilerplate needs to be tweaked. The -o option is just confusing people, and separate files per thread is generally not what is wanted or particularly easy to post.
```
strace -ff <cmd> 2> strace.txt
```
Right now the instructions read:
```
`strace -o strace.txt -ff <cmd>`
```
...which does not actually give you an strace.txt file to post.

aseering:
@therealkenc -- it's a fair point that this is confusing, but your invocation has the disadvantage (IMO) that it captures, and mixes up, the process's stderr with the trace.  My reasoning was that, while people might get confused by the additional process IDs in the filenames, they would probably figure it out, whereas I think understanding and tracking down stderr is more confusing for a novice.

therealkenc:
I suppose the the instructions could be updated to explain they have to zip up the files (as opposed to quote "post the strace.txt", which doesn't exist) and include the stderr output in the zip that's awesome too I guess. A couple of paragraphs ought to do it. Or not. Just trying to be constructive.

aseering:
I mean, you're welcome to submit a PR too :-)  You're commenting on my already-merged PR; no one's going to make any changes on this page, so presumably this page's purpose is for discussing the existing change?

i could be convinced by your syntax.  I would definitely not be a fan of adding a couple paragraphs of explanation; people will ETIMEOUT / get scared off and ignore it more often than they'll benefit from it, I think.

therealkenc:
I was just being inappropriately sardonic about the expository paragraph suggestion. And yes, on me to submit a PR instead of opine from the cheap seats. Really it's the sorry state of the open issue list I have a problem with, not your changes -- which on balance are appreciated. I should have used my inside voice.

aseering:
Ah, no problem :-)  Yeah, tickets here could use some clean-up.

PR #2692, opened 2017-11-25
modify strace guidance

therealkenc:


jstarks:
I think we can stick with the -o flag (which nicely keeps the trace output separate from the process's stderr) if we request -f instead of -ff.

PR #5312, opened 2020-06-01
Template updates for WSL2 release

vkecla:
Ask submitted to specify their distribution and whether they are running WSL1 or WSL2, plus lipstick restructuring.


craigloewen-msft:
👍 Looks good to me!

PR #5320, opened 2020-06-02
Added new bug template

craigloewen-msft:
I edited parts of our bug template to make them more user friendly. Specifically, I added the notes into comments so users didn't have to delete anything and could just fill in the template, and changed the wording to have clearer sections on what info is required to submit a bug.


therealkenc:
LGTM.

PR #6009, opened 2020-09-30
minor

shashwatharsh:


benhillis:
case matches the case of the labels

PR #6436, opened 2021-01-15
Add a wpr log profile for WSL and instructions on how to use it

OneBlue:
This change adds a log profile that we can use to collect logs from users that can't use Feedback Hub.


craigloewen-msft (approved):
Looks good to me!

PR #6452, opened 2021-01-20
Improve network diagnostic scripts

OneBlue:
This change improves the networking diagnostics script by stopping WSL, removing the HNS network, and running networking.sh.
The objective of this logic is to capture the HNS network creation logs.

This also automates the wprp log collection.

### DIFF
--- a/diagnostics/networking.bat
+++ b/diagnostics/networking.bat
@@ -5,7 +5,7 @@
 net session >nul 2>&1 || goto :admin
 
 :: Validate that required files are here
-if not exist wsl.wprp (echo wsl.wprp not found && exit /b 1)
+if not exist wsl_networking.wprp (echo wsl_networking.wprp not found && exit /b 1)
 if not exist networking.sh (echo networking.sh not found && exit /b 1)
 
 :: List all HNS objects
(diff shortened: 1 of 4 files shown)


diagnostics/networking.bat
```diff
@@ -0,0 +1,36 @@
...
+:: Stop WSL
+sc stop LxssManager
+
+:: List all HNS objects
+echo HNS objects: 
+hnsdiag list all -df
+
+:: The WSL HNS network is created once per boot. Resetting it to collect network creation logs
+echo Deleting HNS network
+hnsdiag reset networks
```
dcantah, line 21: This would reset all the networks according to hnsdiag. Why not just delete the wsl ics network only? It has a fixed name iirc

dcantah (approved):

diagnostics/networking.bat
OneBlue: Good point. Changed the script to capture the correct network GUID.

craigloewen-msft (approved):

benhillis (approved):

benhillis (approved):

PR #6465, opened 2021-01-25
Add more providers and packet capture to networking.bat

OneBlue:
Taking inspiration from this [PR](https://github.com/microsoft/SDN/pull/470/files), this change adds more logging provider and a network capture to `networking.bat`.

With these new providers, the .etl file size is around `40MB`.


erfrimod (approved):
Looks good

craigloewen-msft (approved):

Keith-Mange (commented):
LGTM.  Over time we'll figure out if there is more we should add.

PR #6479, opened 2021-01-27
Fallback to wget if curl is not available

OneBlue:
This change adds a call to print the Windows routing table, and a fallback logic to use wget if curl isn't available

### DIFF
--- a/diagnostics/networking.sh
+++ b/diagnostics/networking.sh
@@ -27,5 +27,8 @@ traceroute 1.1.1.1
 cat /etc/resolv.conf
 
 # Validate that everything is functionning correctly
-curl -m 5 -v https://microsoft.com
-
+if which curl > /dev/null; then
+    curl -m 5 -v https://microsoft.com
+else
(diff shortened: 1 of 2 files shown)


diagnostics/networking.sh
```diff
@@ -27,5 +27,8 @@ traceroute 1.1.1.1
 cat /etc/resolv.conf
 
 # Validate that everything is functionning correctly
-curl -m 5 -v https://microsoft.com
-
+if which curl > /dev/null; then
+    curl -m 5 -v https://microsoft.com
+else
```
benhillis, line 32: should we have an else if for wget, and an else that prints an error?

benhillis (approved):

diagnostics/networking.sh
OneBlue: We don't do `set -e` here so I'm ok letting bash printing `wget not found` here

craigloewen-msft (approved):

PR #7034, opened 2021-06-06
minor shell script changes

slycordinator:
Changed to check "ip route" command success with "if (cmd); then ... ; else ... ; fi" construct instead of checking exit codes

changed which to type
type is a shell builtin, which is preferable to using the external which command


OneBlue (approved):
LGTM, thanks !

benhillis (approved):

PR #7316, opened 2021-08-16
Add rebuildregkeys.ps1 to rebuild the regkey if user missed them

licanhua:
I saw user reported distros are missing after Windows update. Because we don't have telemetry on this, and I have no idea how many users are experiencing this, and don't know if I need to create a new feature on this.
As a start, I created simple script to help user to rebuild the regkeys. It has a lot of restrictions:
1. only support known distros downloaded from Windows Store
2. only support WSL2
3. regkeys are hard coded. not support any customization like changing user id, changing distro to folders other than c:

### DIFF
--- a/diagnostics/rebuildregkeys.ps1
+++ b/diagnostics/rebuildregkeys.ps1
@@ -9,15 +9,40 @@ If WSL 2 distros are missing from registry but provided in Get-AppxPackage, this
 then user can import it to recover the regkeys
 
 Example:
-D:\repo\>powershell.exe .\rebuildregkeys.ps1
+D:\repo\WSL\diagnostics>powershell .\rebuildregkeys.ps1
 Below distros are detected in registry
 Below apps are installed
   Ubuntu
 Below apps are installed but not detected in registry which need to rebuild
   Ubuntu
 
-File D:\repo\wsl.reg is generated, please import it to Windows registry
+File D:\repo\WSL\diagnostics\wsl.reg is generated, please import it to Windows registry
+
+DefaultUid is set 0(root) in the script. After you imported the registry, you may reconfig the --default-user with the command the application provided.
+This an example for ubuntu:
+    ubuntu --help
+    ubuntu config --default-user demo
+
 #>
+
+$defaultUidMesasge = @"
+DefaultUid is set 0(root) in the script. After you imported the registry, you may reconfig the --default-user with the command the application provided.
+This an example for ubuntu:
+    ubuntu --help
+    ubuntu config --default-user demo
+"@
+
+$template = @"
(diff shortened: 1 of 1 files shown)


Biswa96:
So, the registry is now documented, right? 😎 

The registry changes assume the GNU/Linux distribution is in WSL2 with default user (UID = 1000) already configured. Will it be correct?

licanhua:
> So, the registry is now documented, right? 😎
> 
No. Please don't think it as public API. It's subject to change

> The registry changes assume the GNU/Linux distribution is in WSL2 with default user (UID = 1000) already configured. Will it be correct?
>
Yes.

diagnostics/rebuildregkeys.ps1
```diff
@@ -0,0 +1,138 @@
...
+
+$template = @"
+[HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Lxss]
+[HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Lxss\{UUID}]
+"State"=dword:00000001
+"DistributionName"="{DISTRONAME}"
+"Version"=dword:00000002
+"BasePath"="{BASEPATH}"
+"Flags"=dword:0000000f
+"DefaultUid"=dword:000003e8
```
benhillis, line 93: Safer to leave DefaultUid as 0 (root).

benhillis:
> > So, the registry is now documented, right? 😎
 @licanhua - please add to the comment that the registry state is subject to change.

diagnostics/rebuildregkeys.ps1
```diff
@@ -0,0 +1,138 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root for license information.
+
+<#
+Script to rebuild the WSL2 distros regkeys when they are missing from user's registry but provided in Get-AppxPackage
```
benhillis, line 5: Why is this WSL2 only? Can support be added for WSL1?

benhillis (changes requested):
Let's fix to work with WSL1 and make it clear that the registry key state is subject to change and should not be used programatically.

diagnostics/rebuildregkeys.ps1
licanhua: no. WSL1 is complex than this, and this script only rebuild for wsl2

licanhua:
> Let's fix to work with WSL1 and make it clear that the registry key state is subject to change and should not be used programatically.

Added a warning message

scottmai (approved):
LGTM

PR #7783, opened 2021-12-01
Add download URL and package family names to DistributioInfo.json

OneBlue:
This change adds the package URLs and family names to DistributionInfo.json

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -5,56 +5,80 @@
             "FriendlyName": "Ubuntu",
             "StoreAppId": "9NBLGGH4MSV6",
             "Amd64": true,
-            "Arm64": true
+            "Arm64": true,
+            "Amd64PackageUrl": "https://wsldownload.azureedge.net/Ubuntu.2020.424.0_x64.appx",
+            "Arm64PackageUrl": "https://wsldownload.azureedge.net/Ubuntu.2020.424.0_ARM64.appx",
+            "PackageFamilyName": "CanonicalGroupLimited.UbuntuonWindows_79rhkp1fndgsc"
         },
         {
             "Name": "Debian",
             "FriendlyName": "Debian GNU/Linux",
             "StoreAppId": "9MSVKQC78PK6",
             "Amd64": true,
-            "Arm64": false
+            "Arm64": false,
+            "Amd64PackageUrl": "https://wsldownload.azureedge.net/DebianGNULinux_1-1-3-0_x64__76v4gfsz19hv4.Appx",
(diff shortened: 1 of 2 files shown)


distributions/DistributionInfo.json
```diff
@@ -5,56 +5,80 @@
...
+            "Arm64": false,
+            "Amd64PackageUrl": "https://wsldownload.azureedge.net/openSUSE-42_v1.appx",
+            "Arm64PackageUrl": null,
+            "PackageFamilyName": "46932SUSE.openSUSELeap42.2_022rs5jcyhyac"
         },
         {
             "Name": "SLES-12",
             "FriendlyName": "SUSE Linux Enterprise Server v12",
-            "StoreAppId": "9NVGTG6752JM",
+            "StoreAppId": "9MZ3D1TRP8T1",
```
OneBlue, line 46: @benhillis: I noticed that this AppId was wrong testing

distributions/validate.py
```diff
@@ -0,0 +1,60 @@
+import requests
```
OneBlue, line 1: This script is useful to quickly validate the urls. It doesn't have to be checked in, but it would be nice to have

distributions/DistributionInfo.json
```diff
@@ -5,56 +5,80 @@
...
+            "PackageFamilyName": "46932SUSE.SUSELinuxEnterpriseServer12SP2_022rs5jcyhyac"
         },
         {
             "Name": "SLES-15",
             "FriendlyName": "SUSE Linux Enterprise Server v15",
             "StoreAppId": "9NJFZK00FGKV",
             "Amd64": true,
-            "Arm64": false
+            "Arm64": false,
+            "Amd64PackageUrl": null,
```
benhillis, line 59: We should add a copy of this one to the offline distro install location.

benhillis (approved):

Biswa96:
Just wondering. Will this JSON file be shipped with system WSL like Windows Terminal has default.json file?

distributions/DistributionInfo.json
OneBlue: Agreed. We can merge this now though and update the URL once it's published there

OneBlue:
> Just wondering. Will this JSON file be shipped with system WSL like Windows Terminal has default.json file?
@Biswa96: No, this file is to be used by wsl --install to figure out which distributions are installable and where to get them from.  It will be read from Github at runtime

Biswa96:
Q. There are some distribution missing arm64 version here. If they add arm64 version will this list be updated?

PR #8966, opened 2022-10-06
Update DistributionInfo.json

craigloewen-msft:
Changes Ubuntu target to the latest


craigloewen-msft:
I've verified this works.
As well as an FYI this Ubuntu package has Canonical's new install experience with the GUI.

benhillis (approved):

PR #9073, opened 2022-10-26
Update DistributionInfo.json to reflect 22.04 version

MSDNAndi:
Corrected Ubuntu 22.04 entry, moved 20.04.4 to separate entry. May need renaming and added PackageURLs


MSDNAndi:

PR #9100, opened 2022-11-01
Updated Ubuntu package

craigloewen-msft:
Updated Ubuntu package from "Ubuntu on Windows" to Ubuntu to rev it to latest (Ubuntu 22.04)


benhillis (changes requested):
Can we get separate appx's from Canonical? This will mean users are downloading twich as much as they need and this package is not small.

craigloewen-msft:
> 

Split it out to two packages!

OneBlue (approved):

benhillis (approved):

PR #9286, opened 2022-12-08
Update DistributionInfo.json

sbradnick:
Making adjustments for SLES-12 & SLES-15 as well as adding Tumbleweed.

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -31,19 +31,29 @@
             "PackageFamilyName": "KaliLinux.54290C8133FEE_ey8k8hqnwqnmg"
         },
         {
-            "Name": "SLES-12",
-            "FriendlyName": "SUSE Linux Enterprise Server v12",
+            "Name": "SLES-12SP5",
+            "FriendlyName": "SUSE Linux Enterprise Server 12SP5",
             "StoreAppId": "9MZ3D1TRP8T1",
             "Amd64": true,
             "Arm64": false,
-            "Amd64PackageUrl": "https://wsldownload.azureedge.net/SLES-12_v1.appx",
+            "Amd64PackageUrl": null,
             "Arm64PackageUrl": null,
-            "PackageFamilyName": "46932SUSE.SUSELinuxEnterpriseServer12SP2_022rs5jcyhyac"
+            "PackageFamilyName": null
         },
         {
-            "Name": "SLES-15",
-            "FriendlyName": "SUSE Linux Enterprise Server v15",
-            "StoreAppId": "9NJFZK00FGKV",
+            "Name": "SLES-15SP4",
+            "FriendlyName": "SUSE Linux Enterprise Server 15SP4",
+            "StoreAppId": "9NMQXWVJJSZJ",
+            "Amd64": true,
+            "Arm64": false,
+            "Amd64PackageUrl": null,
+            "Arm64PackageUrl": null,
+            "PackageFamilyName": null
+        },
+        {
+            "Name": "Tumbleweed",
+            "FriendlyName": "openSUSE Tumbleweed Rolling",
+            "StoreAppId": "9MSSK2ZXXN11",
             "Amd64": true,
             "Arm64": false,
             "Amd64PackageUrl": null,
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -51,7 +51,7 @@
             "PackageFamilyName": null
         },
         {
-            "Name": "Tumbleweed",
+            "Name": "openSUSE-Tumbleweed",
             "FriendlyName": "openSUSE Tumbleweed Rolling",
             "StoreAppId": "9MSSK2ZXXN11",
             "Amd64": true,


distributions/DistributionInfo.json
```diff
@@ -31,19 +31,29 @@
...
+            "Arm64": false,
+            "Amd64PackageUrl": null,
+            "Arm64PackageUrl": null,
+            "PackageFamilyName": null
         },
         {
-            "Name": "SLES-15",
-            "FriendlyName": "SUSE Linux Enterprise Server v15",
-            "StoreAppId": "9NJFZK00FGKV",
+            "Name": "Tumbleweed",
```
OneBlue, line 54: `Tumbleweed` is a bit too generic. Maybe `OpenSuse-rolling`, or `OpenSuse-Tumbleweed` ?

distributions/DistributionInfo.json
```diff
@@ -31,19 +31,29 @@
...
         {
-            "Name": "SLES-12",
-            "FriendlyName": "SUSE Linux Enterprise Server v12",
+            "Name": "SLES-12SP5",
+            "FriendlyName": "SUSE Linux Enterprise Server 12SP5",
             "StoreAppId": "9MZ3D1TRP8T1",
             "Amd64": true,
             "Arm64": false,
-            "Amd64PackageUrl": "https://wsldownload.azureedge.net/SLES-12_v1.appx",
+            "Amd64PackageUrl": null,
```
OneBlue, line 39: @craigloewen-msft : FYI we should upload this package to the CDN

OneBlue (changes requested):

distributions/DistributionInfo.json
sbradnick: That's the name, no more generic than "Debian" or "Ubuntu". I don't object to 'openSUSE-Tumbleweed' (the capitalization is specific) if need be.

distributions/DistributionInfo.json
sbradnick: (I know your comment wasn't for me, but ...) I opted to remove this because I don't know what "_v1" is in reference to and I can't discern if the `--install` will key in on the `StoreAppId` (which I _can_ predict which version will be installed based off the partner portal) or this `Amd64PackageUrl` which SUSE has no control over.

distributions/DistributionInfo.json
OneBlue: `openSUSE-Tumbleweed` sounds good to me

distributions/DistributionInfo.json
OneBlue: @sbradnick: This package URL is used when `wsl.exe --install`  is called with `--web-download` and on server SKU's, so if this is removed, all of these scenarios will not work.

sbradnick: I don't disagree, but SUSE doesn't [have the same level of] control over that `PackageUrl` and IIRC when I checked ahead of entering this PR, the `SLES-12_v1.appx` wasn't something we really want to offer up anyway (off the top of my head, I can't recall that 9MZ3D1TRP8T1 == SLES12SP5 =? `SLES-12_v1.appx`). Last references I see in chats w/ my teammates is it being SLES12SP2 which isn't something we want to offer up; and the files contained in that .appx are from 2017. The `/etc/os-release` contained in `install.tar.gz` also point to SLES12SP2.

Basically, it feels like side-stepping the Store in the sense that we'd like to offer up a curated list of _supported_ distros (as listed by: https://apps.microsoft.com/store/search/suse?filteredCategories=Developer%20tools) and this doesn't fit into that (as evidenced by openSUSE-42 being listed for an extended period of time by `--online` but having been EOL'd by SUSE since July 2019) and doesn't offer much flexibility outside of people installing from github.com/microsoft/wsl/releases AFTER someone such as myself enters a PR to get it fixed.

I'd argue against the idea of a Store method to install distros AND this whole `DistributionInfo.json` method by `--online` unless Microsoft is going to give development partners like SUSE more control over it. _MAYBE_ if it was a mirror of Store offerings, straight up, it'd be ok. The Store may be heavy handed and take lots of clicks to implement, but it's a repeatable process w/ understood results - this seems like a wild-west approach and those who **can** change it aren't necessarily versed in what SUSE is doing and what _should_ be available. There are more than a few Distros out in the wild - so I can understand that - but my team is more than happy to take this on :smiley:

OneBlue: That's understandable. Maybe an acceptable middle ground would be for SUSE to host its own .appx and have a URL pointing to SUSE's CDN's here.

@benhillis  / @craigloewen-msft: what would be your thoughts on this ?

sbradnick: I'm not sure SUSE has much interest in a parallel install method like this. Speaking for myself, as a Linux user since the late 90's - yes, CLI is my preference - but we're getting wildly different experiences between using the Store as a source and this idea of `--web-download` maintained by _someone_ who clearly isn't a mainline SUSE user; that's not a dig at anyone, but as evidenced below - what exists now makes no sense whatsoever.

As the current DistributionInfo.json stands before my PR - running `wsl --install SLES-15` (which doesn't have a `PackageUrl`) uses [9NJFZK00FGKV](https://apps.microsoft.com/store/detail/opensuseleap151/9NJFZK00FGKV) which is actually Leap 15.1 (an EOL version) - this is TERRIBLE and confusing on many levels (i.e. requesting "SUSE Linux Enterprise Server 15SP?" (SLES) and getting "openSUSE Leap 15.1".  No matter what someone would do w/ this installation method/source, they'd get an EOL distro installed).

2023-01-06-091343_1280x867_scrot

It does appear `--web-download` doesn't work when a `PackageUrl` isn't present (I agree with you there - and it seems like a nice-to-have when it's setup properly), but there's little precedent, for SUSE at least, that this is working "as intended" as the results are either NOT what was requested based off the naming or produce a "Catastrophic failure":

```
PS C:\Users\scott>  wsl --install SLES-15 --web-download
Catastrophic failure
Error code: Wsl/InstallDistro/E_UNEXPECTED
```

And by this logic, [SLES-12](https://apps.microsoft.com/store/detail/9MZ3D1TRP8T1) would appear to be CORRECT if done via `wsl --install SLES-12` but INCORRECT if done via `wsl --install SLES-12 --web-download` (You'd actually get SLES12SP2 because of `"Amd64PackageUrl": "https://wsldownload.azureedge.net/SLES-12_v1.appx"`).

OneBlue (approved):

OneBlue (changes requested):
Mislicked. 

This change should be superseded by: https://github.com/microsoft/WSL/pull/9562/files

sbradnick:
Closing in favor of https://github.com/microsoft/WSL/pull/9562 - thanks!

PR #9469, opened 2023-01-10
Adding the Ubuntu-22.04 LTS app

CarlosNihelton:
By updating DistributionInfo.json to bring the latest Ubuntu-22.04 app from the store for those who prefer installing with `wsl --install`.

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -76,8 +76,8 @@
             "StoreAppId": "9PN20MSR04DW",
             "Amd64": true,
             "Arm64": true,
-            "Amd64PackageUrl": "https://wslstorestorage.blob.core.windows.net/wslblob/Ubuntu2204LTS-230109.AppxBundle",
-            "Arm64PackageUrl": "https://wslstorestorage.blob.core.windows.net/wslblob/Ubuntu2204LTS-230109.AppxBundle",
+            "Amd64PackageUrl": "https://wslstorestorage.blob.core.windows.net/wslblob/Ubuntu2204LTS-230117.appx",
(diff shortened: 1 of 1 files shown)


craigloewen-msft:
@benhillis @OneBlue   want me to break out the ARM64 and X64 packages?

OneBlue:
> @benhillis @OneBlue want me to break out the ARM64 and X64 packages?

Yeah that would divide the package size by two, so I  think we should do that

frivard-coveo:
Quick question:  would this PR resolve this issue: https://github.com/microsoft/vscode-remote-release/issues/7858

where Ubuntu 22.04 (my only distro) is not recognized by the WSL Remote extension?

distributions/DistributionInfo.json
```diff
@@ -70,6 +70,16 @@
...
             "PackageFamilyName": "CanonicalGroupLimited.Ubuntu20.04onWindows_79rhkp1fndgsc"
         },
+        {
+            "Name": "Ubuntu-22.04",
+            "FriendlyName": "Ubuntu 22.04 LTS",
+            "StoreAppId": "9PN20MSR04DW",
+            "Amd64": true,
+            "Arm64": true,
+            "Amd64PackageUrl": "https://wslstorestorage.blob.core.windows.net/wslblob/Ubuntu2204LTS-230109.AppxBundle",
+            "Arm64PackageUrl": "https://wslstorestorage.blob.core.windows.net/wslblob/Ubuntu2204LTS-230109.AppxBundle",
```
craigloewen-msft, line 80:
```suggestion
            "Arm64PackageUrl": "https://wslstorestorage.blob.core.windows.net/wslblob/Ubuntu2204LTS-230117_ARM64.appx",
```

OneBlue (approved):

PR #9522, opened 2023-01-23
Added Open Euler

craigloewen-msft:
Added Open Euler


craigloewen-msft:
image

OneBlue:
Thanks for doing this @craigloewen-msft. Looking install the .tar.gz, I see some files that we shouldn't ship: 

- /boot (100MB of kernels & initramfs)
- /dev
- /.dockerenv
- /afs
- /run

Maybe we should ask the maintainers to remove those to reduce the size of this distro a bit?

craigloewen-msft:
image

benhillis:
Agreed with @OneBlue, we should engage with the OpenEuler folks to make sure they aren't shipping hundreds of megs of stuff that aren't needed.

pkking:
Hi, we have pushed a new release to ms store, could you please check the new wsl app.

pkking:
图片
@craigloewen-msft Hi, the newest WSL already pruned these content, can you please rebase and merge?

PR #9717, opened 2023-02-28
Make WSL Version separate.

pmartincic:


Biswa96:
Doesn't the collect-wsl-logs.ps1 already provide the lifted WSL version?

pmartincic:
The problem is that most people don't read both sentences or provide logs. And sometimes I don't need logs, I just need to see that they're on the old version to confirm it's an known issue.

craigloewen-msft (approved):

Issue #9802, opened 2023-03-17
Wsl/0x80070020


NickLiu635:
### Windows Version

Microsoft Windows [版本 10.0.22621.1413]

### WSL Version

0.0.0.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [X] WSL 1

### Distro Version

Ubuntu

### Repro Steps

![image](https://user-images.githubusercontent.com/36353209/225791735-42613347-b4f1-4dd0-8dac-bae4162d2981.png)

### Expected Behavior

---

### Actual Behavior

---

NickLiu635:
I see that there is a WSL window that is still functioning normally. Here are some details.
```
1:/$ cat /proc/version
Linux version 4.4.0-22621-Microsoft (Microsoft@Microsoft.com) (gcc version 5.4.0 (GCC) ) #1105-Microsoft Fri Jan 01 08:00:00 PST 2016
1:/$ lsb_release -r
Release:        22.04
```

NickLiu635:
Same issue when installing a distro.
![image](https://user-images.githubusercontent.com/36353209/225793923-adcddde9-2435-44b7-aa21-ac2a37a6cf4f.png)

OneBlue:
Thanks @NickLiu635. 

If there is no crash dump we'll need to follow what's happening with a debugger. Can you please follow the ['12) Collect wslservice time travel debugging traces'  section](https://github.com/microsoft/WSL/blob/f9bc074b180b07c7825f36619a5930b49a3bde2b/CONTRIBUTING.md#12-collect-wslservice-time-travel-debugging-traces) and share the traces so I can have a deeper look ?

NickLiu635:
> Thanks @NickLiu635.
> 
> If there is no crash dump we'll need to follow what's happening with a debugger. Can you please follow the ['12) Collect wslservice time travel debugging traces' section](https://github.com/microsoft/WSL/blob/f9bc074b180b07c7825f36619a5930b49a3bde2b/CONTRIBUTING.md#12-collect-wslservice-time-travel-debugging-traces) and share the traces so I can have a deeper look ?

there is no `wslservice.exe` ; `Error code: Wsl/0x80070020` while `wsl -l`

OneBlue:
Thank you @NickLiu635.

Unfortunately, the trace captured C:\windows\system32\wsl.exe, which only calls inside the store application.

[...]

PR #9807, opened 2023-03-17
Add instructions on how to capture TDD traces

OneBlue:
This change adds instructions on how to capture time travel debugging traces.

This will be useful for issues like #9802

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -222,3 +222,31 @@ Once you're done, crash dump collection can be disabled by running the following
 ```
 reg.exe delete "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps" /f
 ```
+
+### 12) Collect wslservice time travel debugging traces
+
+To collect time travel debugging traces:
+
+1) [Install Windbg preview](https://apps.microsoft.com/store/detail/windbg-preview/9PGJGD53TN86?hl=en-us&gl=us&rtc=1)
+
+2) Open windbg preview as administrator by running `windbgx` in an elevated command prompt
+
+3) Navigate to `file` -> `Attach to process`
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -221,4 +221,32 @@ Once you're done, crash dump collection can be disabled by running the following
...
 reg.exe delete "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps" /f
-```
\ No newline at end of file
+```
+
+### 12) Collect wslservice time travel debugging traces
+
+To collect time travel debugging traces:
+
+1) [Install Windbg preview](https://apps.microsoft.com/store/detail/windbg-preview/9PGJGD53TN86?hl=en-us&gl=us&rtc=1)
```
pmartincic, line 230: Will this work for server users?

pmartincic (approved):

CONTRIBUTING.md
OneBlue: Unfortunately not, but I couldn't find an official way to get the .appx so for now these instructions will not work with desktop SKU's

pmartincic: How do you feel about this?
image

OneBlue: I'd like to but winget is not available on Windows 10 so I'd rather not depend on it

OneBlue: (Not on all the Windows 10 builds we support with Store WSL at least)

pmartincic: @craigloewen-msft, do you have numbers on whether more users will be on win10 builds without winget, or if there are more on server?

benhillis: The vast majority of users are on client SKUs so I think the store instructions + winget instructions for server is a good compromise.

PR #9992, opened 2023-04-18
Updated validate.py to optionally accept a specific distro

craigloewen-msft:
Makes testing faster and easier

### DIFF
--- a/distributions/validate.py
+++ b/distributions/validate.py
@@ -61,7 +61,7 @@ def is_unique(collection: list):
 
 if __name__ == "__main__":
     if len(sys.argv) < 2:
-        print(f'Usage: {sys.argv[0]} /path/to/file', file=sys.stderr)
+        print(f'Usage: {sys.argv[0]} /path/to/file [distroName]', file=sys.stderr)
         exit(1)
 
     with open(sys.argv[1]) as fd:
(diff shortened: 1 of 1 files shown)


distributions/validate.py
```diff
@@ -70,8 +70,15 @@ def is_unique(collection: list):
     distros = content['Distributions']
     assert is_unique([e.get('StoreAppId') for e in distros if e])
     assert is_unique([e.get('Name') for e in distros if e])
+    
+    specificDistro = ""
+
+    if len(sys.argv) > 2:
+        specificDistro = sys.argv[2]
```
OneBlue, line 77: nit: We could take a different branch and do: 
```
    if len(sys.argv) > 2:
        content =[e for e in content if e['name'] == sys.argv[2]]
        if not content:
            raise RuntimeError(f'No distribution found for name: {sys.argv[2]}')
```

So that way the script fails if we pass an invalid distribution name

OneBlue (approved):

distributions/validate.py
craigloewen-msft: Fixed!

OneBlue (approved):

momohnasirumuri:
Run

PR #10119, opened 2023-05-18
Updated Ubuntu 2204 package

craigloewen-msft:
Update to fix an error where Ubuntu 2204 LTS package couldn't be installed on Windows Server due to a bundled dependency


craigloewen-msft:
image

OneBlue (approved):

PR #10180, opened 2023-06-09
Add networking diagnostics to collect-wsl-logs script

CatalinFetoiu:
Collect networking diagnostics as part of the existing collect-wsl-logs script

The networking diagnostics will be added in the zip archive in a file named "wsl_network_snapshot.log"


diagnostics/collect-wsl-logs.ps1
```diff
@@ -130,6 +130,100 @@ if ($Dump)
...
 }
 
+# Collect networking state relevant for WSL
+$wslNetworkLog = "$folder/wsl_network_snapshot.log"
+$stdOutLog = "$folder/wsl_network_stdout.log"
+$stdErrLog = "$folder/wsl_network_stderr.log"
+
+try
+{
+    Write-Output "WSL Net Snapshot" | Out-File -FilePath $wslNetworkLog
```
OneBlue, line 140: Instead of merging everything in the same file, could we use the same model as the rest of the script and have each command write its output in a single file ? This would allow us to get rid of the: '-----------------------------------------------------------------------' prints and would remove the need to try / catch this section as long as we pass `-ErrorAction Ignore` to all commands

OneBlue (changes requested):

diagnostics/collect-wsl-logs.ps1
```diff
@@ -131,86 +131,56 @@ if ($Dump)
...
 # Collect networking state relevant for WSL
-$wslNetworkLog = "$folder/wsl_network_snapshot.log"
 $stdOutLog = "$folder/wsl_network_stdout.log"
 $stdErrLog = "$folder/wsl_network_stderr.log"
 
+# Using a try/catch as some of the commands below do not exist on some OS versions
 try
 {
-    Write-Output "WSL Net Snapshot" | Out-File -FilePath $wslNetworkLog
+    Write-Output "Get-NetAdapter" | Out-File -FilePath "$folder/Get-NetAdapter.log" -Append
```
OneBlue, line 140: Given that the file name describes its content, I think we can remove the strings as well to simplify things

diagnostics/collect-wsl-logs.ps1
```diff
@@ -130,6 +130,70 @@ if ($Dump)
     }
 }
 
+# Collect networking state relevant for WSL
+$stdOutLog = "$folder/wsl_network_stdout.log"
+$stdErrLog = "$folder/wsl_network_stderr.log"
+
+# Using a try/catch as some of the commands below do not exist on some OS versions
```
benhillis, line 137: Would it be better to wrap individual operations in try / catch so collection doesn't stop at first error?

diagnostics/collect-wsl-logs.ps1
CatalinFetoiu: updated to output to multiple files.
will keep the try/catch based on offline discussion

diagnostics/collect-wsl-logs.ps1
```diff
@@ -131,86 +131,56 @@ if ($Dump)
...
-    Write-Output "hnsdiag.exe list all" | Out-File -FilePath $wslNetworkLog -Append
+    Write-Output "hnsdiag.exe list all" | Out-File -FilePath "$folder/hnsdiag_list_all.log" -Append
     Start-Process -FilePath "hnsdiag.exe" -ArgumentList "list all" -RedirectStandardOutput $stdOutLog -RedirectStandardError $stdErrLog -NoNewWindow -wait
-    Get-Content $stdOutLog, $stdErrLog | Out-File -FilePath $wslNetworkLog -Append
-
-    Write-Output -----------------------------------------------------------------------   | Out-File -FilePath $wslNetworkLog -Append
+    Get-Content $stdOutLog, $stdErrLog | Out-File -FilePath "$folder/hnsdiag_list_all.log" -Append
 
-    Write-Output "hnsdiag.exe list endpoints -df" | Out-File -FilePath $wslNetworkLog -Append
+    Write-Output "hnsdiag.exe list endpoints -df" | Out-File -FilePath "$folder/hnsdiag_list_endpoints.log" -Append
```
OneBlue, line 159: Could we simplify by : `& hnsdiag.exe list endpoints -df  2>&1 >> "$folder/hnsdiag_list_endpoints.log"`  ?

OneBlue: (Same for the below calls)

diagnostics/collect-wsl-logs.ps1
CatalinFetoiu: agreed will update for this and the other commands

diagnostics/collect-wsl-logs.ps1
CatalinFetoiu: thanks :) this should help remove the need for the additional stderr and stdout files

diagnostics/collect-wsl-logs.ps1
CatalinFetoiu: good point, will update

diagnostics/collect-wsl-logs.ps1
CatalinFetoiu: ended up using the "$cmd | cmd | Out-Null" construct to run cmd commands

diagnostics/collect-wsl-logs.ps1
```diff
@@ -131,68 +131,73 @@ if ($Dump)
...
+    $cmd | cmd | Out-Null
+}
+catch {}
 
-    Write-Output "hnsdiag.exe list endpoints -df" | Out-File -FilePath "$folder/hnsdiag_list_endpoints.log" -Append
-    Start-Process -FilePath "hnsdiag.exe" -ArgumentList "list endpoints -df" -RedirectStandardOutput $stdOutLog -RedirectStandardError $stdErrLog -NoNewWindow -wait
-    Get-Content $stdOutLog, $stdErrLog | Out-File -FilePath "$folder/hnsdiag_list_endpoints.log" -Append
+try
+{
+    $cmd = "hnsdiag.exe list endpoints -df 2>&1 > " + "$folder/hnsdiag_list_endpoints.log"
```
OneBlue, line 175: What's the reason for taking this approach instead of  let's say:

```
& hnsdiag.exe list endpoints -df 2>&1 > $folder/hnsdiag_list_endpoints.log
```

?

CatalinFetoiu: I was not able to get the right syntax for the vfpctrl commands below (which need to be dynamically created) and use this approach instead

I updated for the hnsdiag commands

let me know if you think we can simplify the vfpctrl commands as well

diagnostics/collect-wsl-logs.ps1
```diff
@@ -130,6 +130,72 @@ if ($Dump)
...
+    & hnsdiag.exe list endpoints -df 2>&1 > $folder/hnsdiag_list_endpoints.log
+}
+catch {}
+
+try
+{
+    foreach ($port in Get-NetFirewallHyperVPort)
+    {
+        $vfpLogFile = "$folder/vfp-port-" + $port.PortName + "-get-port-state.log"
+        $cmd = "vfpctrl.exe /port " + $port.PortName + " /get-port-state > " + $vfpLogFile
```
OneBlue, line 183: Could this also use the &<command> 2>&1 > $folder/<log> structure ?

CatalinFetoiu: I wasn't able to get the right syntax here using operator "&" - since the command string needs to be dynamically created, I tried doing just & $dynamic_cmd_variable, but this did not seem to work

mitschmi: Should be resolved with latest fix

mitschmi: Oh, after taking a look at what this was previously, did you want this to be called using "& vfpctrl.exe ..." rather than use the $cmd | cmd | Out-Null and just add "2>&1 >" to the commands? I can give this a try

mitschmi: okay, I think i found right escape characters, so should be resolved with latest push

keith-horton (approved):

diagnostics/collect-wsl-logs.ps1
```diff
@@ -130,6 +130,104 @@ if ($Dump)
...
+try
+{
+    & vfpctrl.exe /list-vmswitch-port 2>&1 > $networkingFolder/vfpctrl_list_vmswitch_port.log
+}
+catch {}
+
+# Container networking info
+try
+{
+	wsl -e ip a s | Out-File -FilePath "$networkingFolder/wsl_ip_a_s.log" -Append
```
OneBlue, line 209: We interact with the user distribution as part of the log collection because this might block forever is WSL is in a bad state. 

If we need container information, that should be done in a separate script that can users can run if needed

OneBlue (changes requested):

diagnostics/collect-wsl-logs.ps1
mitschmi: Could an alternative solution be to run the WSL commands asynchronously with some kind of timeout? 
maybe something using https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/wait-job?view=powershell-7.3 ? (still need to look into how to use this exactly)

OneBlue: I would prefer to avoid doing this since we don't know which version of powershell the user will have installed and we also don't know which distribution the issue is with (calling wsl.exe will invoke the default distribution, but that might not the one having the issue).

We have [dedicated script for that purpose](https://github.com/microsoft/WSL/blob/master/diagnostics/networking.bat) that we ask users to run specifically for networking issues. I'd recommend adding to that one instead

mitschmi: Makes sense, thanks Pierre, will add to that script then.

OneBlue (approved):

PR #10266, opened 2023-07-06
Add WSL Community Telegram channel to Community Links

sirredbeard:
The WSL Community Telegram channel is the longest-running chat dedicated to WSL and branched from the original series of WSLConf conference series.


OneBlue:
Thank you for this but a telegram channel doesn't fit well within our README

PR #10409, opened 2023-08-22
Add new script for WSL networking diagnostics

CatalinFetoiu:
The script collects the regular WSL logs (using collect-wsl-logs.ps1).

In addition, it collects:
traces from networking providers (wsl_networking.wprp)
tcpdump from the container
pktmon logs
WFP logs
host routes
It runs the networking.sh scripts before and after repro and collects it's output

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,7 +119,32 @@ Availability  Capabilities  CapabilityDescriptions                              
 
 #### Networking issues
 
-If the issue is about networking, run [networking.bat](https://github.com/Microsoft/WSL/blob/master/diagnostics/networking.bat) in an administrative command prompt:
+Install tcpdump in your WSL distribution using the following commands.
+Note: This might not work if WSL has Internet connectivity issues.
+
+```
+# sudo apt-get update
+# sudo apt-get -y install tcpdump
+```
+
+Run [networking_diagnostics.bat](https://github.com/Microsoft/WSL/blob/master/diagnostics/networking_diagnostics.bat) in an administrative command prompt:
+
+```
+$ git clone https://github.com/microsoft/WSL --depth=1 %tmp%\WSL
+$ cd %tmp%\WSL\diagnostics
+$ networking_diagnostics.bat
+```
+
+Note:
+If tcpdump is installed, the script will open 2 more shells, one of which will ask for your sudo password.
+If tcpdump is not installed, only 1 additional shell will be opened
+Wait for those shells to be opened before reproducing the issue
+
+After reproducing the issue, in the shell that asked for your sudo password, hit Ctrl + C. In the other shells, press any key.
+
+Collect `WslNetworkingLogs.zip`
+
+Run [networking.bat](https://github.com/Microsoft/WSL/blob/master/diagnostics/networking.bat) in an administrative command prompt:
 
 ```
 $ git clone https://github.com/microsoft/WSL --depth=1 %tmp%\WSL
--- a/diagnostics/networking_diagnostics.bat
+++ b/diagnostics/networking_diagnostics.bat
@@ -1,0 +1,18 @@
+powershell invoke-expression 'cmd /c start powershell -Command { .\collect-wsl-logs.ps1 }'
+wsl.exe tr -d "\r" ^| bash < ./networking.sh > wsl_network_configuration_before.log
+powershell Get-NetRoute > get_netroute.log
+powershell invoke-expression 'cmd /c start powershell -Command { "wsl.exe sudo tcpdump -n -i any > tcpdump.log" }'
+
+wpr -start .\wsl_networking.wprp -filemode -instanceName wpr_networking
+pktmon start -c --flags 0x1A
+netsh wfp capture start
+
+pause
+
+netsh wfp capture stop
+pktmon stop
+wpr -stop wsl_networking.etl -instanceName wpr_networking
+
+wsl.exe tr -d "\r" ^| bash < ./networking.sh > wsl_network_configuration_after.log
+
+echo "Finished log collection"
--- a/diagnostics/networking_diagnostics.bat
+++ b/diagnostics/networking_diagnostics.bat
@@ -1,6 +1,12 @@
+set networking_folder=".\networking_logs"
+set neworking_logs_zip=".\WslNetworkingLogs.zip"
+
+mkdir %networking_folder%
+
+wsl.exe tr -d "\r" ^| bash < ./networking.sh > %networking_folder%\wsl_network_configuration_before.log
+powershell Get-NetRoute > %networking_folder%\get_netroute.log
+
 powershell invoke-expression 'cmd /c start powershell -Command { .\collect-wsl-logs.ps1 }'
-wsl.exe tr -d "\r" ^| bash < ./networking.sh > wsl_network_configuration_before.log
-powershell Get-NetRoute > get_netroute.log
 powershell invoke-expression 'cmd /c start powershell -Command { "wsl.exe sudo tcpdump -n -i any > tcpdump.log" }'
 
 wpr -start .\wsl_networking.wprp -filemode -instanceName wpr_networking
@@ -9,10 +15,23 @@ netsh wfp capture start
 
 pause
 
+:: allow some time for the user to stop logs in all the spawned shells
+:: (particularly the shell running collect-wsl-logs.ps1 will take a bit to stop)
+timeout 20
+
 netsh wfp capture stop
 pktmon stop
-wpr -stop wsl_networking.etl -instanceName wpr_networking
+wpr -stop %networking_folder%\wsl_networking.etl -instanceName wpr_networking
 
-wsl.exe tr -d "\r" ^| bash < ./networking.sh > wsl_network_configuration_after.log
+wsl.exe tr -d "\r" ^| bash < ./networking.sh > %networking_folder%\wsl_network_configuration_after.log
+
+move tcpdump.log %networking_folder%
+move PktMon.etl %networking_folder%
+move wfpdiag.cab %networking_folder%
+move "*zip" %networking_folder%
+
+powershell Compress-Archive -Path %networking_folder% -DestinationPath %neworking_logs_zip%
+
+rmdir /s /q %networking_folder%
 
 echo "Finished log collection"


CONTRIBUTING.md
```diff
@@ -119,7 +119,32 @@ Availability  Capabilities  CapabilityDescriptions
...
+
+```
+# sudo apt-get update
+# sudo apt-get -y install tcpdump
+```
+
+Run [networking_diagnostics.bat](https://github.com/Microsoft/WSL/blob/master/diagnostics/networking_diagnostics.bat) in an administrative command prompt:
+
+```
+$ git clone https://github.com/microsoft/WSL --depth=1 %tmp%\WSL
```
OneBlue, line 133: The git command might not be available on the user's machine so I'd recommend downloading the files with powershell instead

CONTRIBUTING.md
```diff
@@ -119,7 +119,32 @@ Availability  Capabilities  CapabilityDescriptions
...
+$ cd %tmp%\WSL\diagnostics
+$ networking_diagnostics.bat
+```
+
+Note:
+If tcpdump is installed, the script will open 2 more shells, one of which will ask for your sudo password.
```
OneBlue, line 139: Instead of doing that, let's directly invoke the script as root (wsl.exe -u root <command>)

diagnostics/networking_diagnostics.bat
```diff
@@ -0,0 +1,37 @@
+set networking_folder=".\networking_logs"
+set neworking_logs_zip=".\WslNetworkingLogs.zip"
+
+mkdir %networking_folder%
+
+wsl.exe tr -d "\r" ^| bash < ./networking.sh > %networking_folder%\wsl_network_configuration_before.log
+powershell Get-NetRoute > %networking_folder%\get_netroute.log
+
+powershell invoke-expression 'cmd /c start powershell -Command { .\collect-wsl-logs.ps1 }'
+powershell invoke-expression 'cmd /c start powershell -Command { "wsl.exe sudo tcpdump -n -i any > tcpdump.log" }'
```
OneBlue, line 10: Let's not call sudo and use wsl -u root instead

OneBlue:
Overall this script looks very similar to `networking.bat`. Having both `networking.bat` and `networking_diagnostics.bat` feels a bit confusing. Could we merge those into one ?

OneBlue (changes requested):

diagnostics/networking_diagnostics.bat
```diff
@@ -0,0 +1,37 @@
...
+
+mkdir %networking_folder%
+
+wsl.exe tr -d "\r" ^| bash < ./networking.sh > %networking_folder%\wsl_network_configuration_before.log
+powershell Get-NetRoute > %networking_folder%\get_netroute.log
+
+powershell invoke-expression 'cmd /c start powershell -Command { .\collect-wsl-logs.ps1 }'
+powershell invoke-expression 'cmd /c start powershell -Command { "wsl.exe sudo tcpdump -n -i any > tcpdump.log" }'
+
+wpr -start .\wsl_networking.wprp -filemode -instanceName wpr_networking
```
OneBlue, line 12: Instead of having two log collection running at the same time, let's pass this custom log profile to collect-wsl-logs instead

diagnostics/networking_diagnostics.bat
```diff
@@ -0,0 +1,37 @@
...
+netsh wfp capture stop
+pktmon stop
+wpr -stop %networking_folder%\wsl_networking.etl -instanceName wpr_networking
+
+wsl.exe tr -d "\r" ^| bash < ./networking.sh > %networking_folder%\wsl_network_configuration_after.log
+
+move tcpdump.log %networking_folder%
+move PktMon.etl %networking_folder%
+move wfpdiag.cab %networking_folder%
+move "*zip" %networking_folder%
```
OneBlue, line 31: This expansion is dangerous since it might move unrelated user files.If we need to also collect regular logs I'd recommand a way to pass a path down to collect-wsl-logs instead

diagnostics/networking_diagnostics.bat
```diff
@@ -0,0 +1,37 @@
...
+:: (particularly the shell running collect-wsl-logs.ps1 will take a bit to stop)
+timeout 20
+
+netsh wfp capture stop
+pktmon stop
+wpr -stop %networking_folder%\wsl_networking.etl -instanceName wpr_networking
+
+wsl.exe tr -d "\r" ^| bash < ./networking.sh > %networking_folder%\wsl_network_configuration_after.log
+
+move tcpdump.log %networking_folder%
```
OneBlue, line 28: Could we run these commands in the right directory directly instead of having to move the files afterwards ?

diagnostics/networking_diagnostics.bat
```diff
@@ -0,0 +1,37 @@
...
+move PktMon.etl %networking_folder%
+move wfpdiag.cab %networking_folder%
+move "*zip" %networking_folder%
+
+powershell Compress-Archive -Path %networking_folder% -DestinationPath %neworking_logs_zip%
+
+rmdir /s /q %networking_folder%
+
+echo "Finished log collection"
```
OneBlue, line 37: The script should display the path to the folder it created so the user knows where it is

CatalinFetoiu:
> Overall this script looks very similar to `networking.bat`. Having both `networking.bat` and `networking_diagnostics.bat` feels a bit confusing. Could we merge those into one ?

good point, I'll look more into what networking.bat does and move things over to this new script

diagnostics/networking_diagnostics.bat
CatalinFetoiu: Currently wsl_networking.wprp does not include the trace providers that wsl.wprp has - we need the traces providers from both wsl.wprp and wsl_networking.wprp

thoughts? should we merge wsl.wprp and wsl_networking.wprp in a single file?

OneBlue: I think we should keep wsl.wprp as it currently to avoid creating too much noise in log files, but I think adding all the profiles from wsl.wprp into wsl_networking.wprp would make sense here

keith-horton:
> 

the existing networking.bat solves a different problem than this script. That script attempts to repro a bad network config by tearing down all HNS objects, shutting down containers, and starting over.

we deliberately do not want to destroy the current repro environment. we want to snap everything about the current state, and attempt to watch packet flows.

OneBlue:
> > 
> 
> the existing networking.bat solves a different problem than this script. That script attempts to repro a bad network config by tearing down all HNS objects, shutting down containers, and starting over.
> 
> we deliberately do not want to destroy the current repro environment. we want to snap everything about the current state, and attempt to watch packet flows.

That makes sense. Could we potentially add parameters to that script then ? Destruction of the networking state could be opt-in.

CatalinFetoiu:
> > > 
> > 
> > 
> > the existing networking.bat solves a different problem than this script. That script attempts to repro a bad network config by tearing down all HNS objects, shutting down containers, and starting over.
> > we deliberately do not want to destroy the current repro environment. we want to snap everything about the current state, and attempt to watch packet flows.
> 
> That makes sense. Could we potentially add parameters to that script then ? Destruction of the networking state could be opt-in.

@keith-horton @OneBlue easiest option I can think of is to keep the old networking.bat. And add a flag to the new networking script, that will invoke the old networking.bat

And we can update the instructions to mention just the new script.

Thoughts?

OneBlue:
> > > > 
> > > 
> > > 
> > > the existing networking.bat solves a different problem than this script. That script attempts to repro a bad network config by tearing down all HNS objects, shutting down containers, and starting over.
> > > we deliberately do not want to destroy the current repro environment. we want to snap everything about the current state, and attempt to watch packet flows.
> > 
> > 
> > That makes sense. Could we potentially add parameters to that script then ? Destruction of the networking state could be opt-in.
> 
> @keith-horton @OneBlue easiest option I can think of is to keep the old networking.bat. And add a flag to the new networking script, that will invoke the old networking.bat
> 
> And we can update the instructions to mention just the new script.
> 
> Thoughts?

Could we merge both into one, with a parameter to decide if we want to destroy the HNS state or not ? 

I think we should really try to only have one entry point for networking diagnostics to reduce potential confusion

CONTRIBUTING.md
CatalinFetoiu: fixed in iteration 3

CONTRIBUTING.md
CatalinFetoiu: fixed in iteration 3

diagnostics/networking_diagnostics.bat
CatalinFetoiu: fixed in iteration 3

diagnostics/networking_diagnostics.bat
CatalinFetoiu: fixed in iteration 3

diagnostics/networking_diagnostics.bat
CatalinFetoiu: fixed in iteration 3

diagnostics/networking_diagnostics.bat
CatalinFetoiu: fixed in iteration 3

CatalinFetoiu:
> > > > > 
> > > > 
> > > > 
> > > > the existing networking.bat solves a different problem than this script. That script attempts to repro a bad network config by tearing down all HNS objects, shutting down containers, and starting over.
> > > > we deliberately do not want to destroy the current repro environment. we want to snap everything about the current state, and attempt to watch packet flows.
> > > 
> > > 
> > > That makes sense. Could we potentially add parameters to that script then ? Destruction of the networking state could be opt-in.
> > 
> > 
> > @keith-horton @OneBlue easiest option I can think of is to keep the old networking.bat. And add a flag to the new networking script, that will invoke the old networking.bat
> > And we can update the instructions to mention just the new script.
> > Thoughts?
> 
> Could we merge both into one, with a parameter to decide if we want to destroy the HNS state or not ?
> 
> I think we should really try to only have one entry point for networking diagnostics to reduce potential confusion

@keith-horton @OneBlue I remove the old script and added a "--stop-wsl" to the new script

Note that I did not port other things from the old script to the new one. for example the old script also had a packet capture, but the new one already does that using pktmon and tcpdump

diagnostics/networking_diagnostics.bat
CatalinFetoiu: resolved in iteration 3

diagnostics/networking_diagnostics.bat
```diff
@@ -0,0 +1,60 @@
...
+   net.exe stop WslService || net.exe stop LxssManager
+)
+
+wsl.exe tr -d "\r" ^| bash < ../networking.sh > wsl_network_configuration_before.log
+powershell Get-NetRoute > get_netroute.log
+
+powershell invoke-expression 'cmd /c start powershell -Command { ..\collect-wsl-logs.ps1 }'
+powershell invoke-expression 'cmd /c start powershell -Command { "wsl.exe -u root sudo tcpdump -n -i any > tcpdump.log" }'
+
+wpr -start ..\wsl_networking.wprp -filemode -instanceName wpr_networking
```
OneBlue, line 30: Instead of starting another log collection, can you pass this file to collect-wsl-logs.ps1 ?
 `collect-wsl-logs.ps1 -LogProfile wsl_networking.wprp` 

If needed we can copy all the providers in the default log file to `wsl_networking.wprp`. 

I think this would make a things a bit simpler, and collect-wsl-logs.ps1 has the benefit of handling the case where the log collection is already running

diagnostics/networking_diagnostics.bat
```diff
@@ -0,0 +1,60 @@
...
+pktmon start -c --flags 0x1A
+netsh wfp capture start
+
+pause
```
OneBlue, line 34: We should display a message telling the user to reproduce the issue here

diagnostics/networking_diagnostics.bat
```diff
@@ -0,0 +1,60 @@
...
+
+:: allow some time for the user to stop logs in all the spawned shells
```
OneBlue, line 36: This isn't a great UX for the user. Could we kill those processes ourselves instead ? We might need to switch to a powershell script for that, but I think it would be worth it because users will most likely get confused here

OneBlue: Actually looking at this script it's probably a good idea to move it to native powershell given how much we call powershell.exe in it

OneBlue (changes requested):

diagnostics/networking_diagnostics.bat
CatalinFetoiu: "Instead of starting another log collection, can you pass this file to collect-wsl-logs.ps1" makes sense, will do.

"If needed we can copy all the providers in the default log file to wsl_networking.wprp" the latest iteration of the PR already does this

diagnostics/networking_diagnostics.bat
CatalinFetoiu: moved the script to powershell and changed it to automatically stop the log collection - now the user just needs to press a key to stop logs, similar to collect-wsl-logs.ps1

diagnostics/networking_diagnostics.bat
CatalinFetoiu: updated wsl_networking.wprp and used wpr with wsl_networking.wprp (instead of calling collect-wsl-logs.ps1 from the new script)

diagnostics/collect-networking-logs.ps1
```diff
@@ -0,0 +1,210 @@
...
+pktmon start -c --flags 0x1A --file-name "$folder/pktmon.etl" | out-null
+
+# Start WFP capture
+netsh wfp capture start file="$folder/wfpdiag.cab"
+
+# Start tcpdump. Using a try/catch as tcpdump might not be installed
+$tcpdumpProcess = $null
+try
+{
+    $tcpdumpProcess = Start-Process wsl.exe -ArgumentList "-u root /bin/bash -c 'tcpdump -n -i any > $folder/tcpdump.log'" -PassThru
```
OneBlue, line 63: Since this is no longer a task, you can remove the '/bin/bash -c' part

diagnostics/collect-networking-logs.ps1
```diff
@@ -0,0 +1,210 @@
...
+        $Key = $Host.UI.RawUI.ReadKey('NoEcho,IncludeKeyDown')
+    }
+
+    Write-Host "`nSaving logs..."
+}
+finally
+{
+    try
+    {
+        if ($tcpdumpProcess -ne $null)
```
OneBlue, line 112: FYI This will stop wsl.exe but it will no synchronously stop the underlying tcpdump process so this might miss some logs.

diagnostics/networking.sh
```diff
@@ -1,40 +1,19 @@
...
-if gateway=$(ip route show | awk '/default/ { print $3 }'); then
-    ping -c 4 "$gateway"
-else
-    echo 'No gateway found.'
-fi
-
-ping -c 4 1.1.1.1
+# This will include the HTTP proxy env variables, if present
+echo "Printing all environment variables"
+printenv
```
OneBlue, line 16: We should avoid printing the environment here because it might contain secrets. If you want the proxy variables I recommend printing whether the env variables are set or not but not their content

OneBlue (changes requested):

diagnostics/networking.sh
CatalinFetoiu: thanks for pointing it out. will update to print if HTTP_PROXY, http_proxy and the other http proxy variables are set or not

diagnostics/collect-networking-logs.ps1
CatalinFetoiu: thanks for pointing it out

I think that's ok, because by the time we reached this point in the script, the user already finished the repro so we collected tcpdump output during the repro

OneBlue: What I meant is that tcpdump's writing to a file is not synchronous and might still be in buffer at the point where it gets killed so can potentially be a lot of packets missing which might make the investigation harder

CatalinFetoiu: got it, thanks for the clarification. do you have suggestions for avoiding this?

OneBlue: I think the easiest would be to send ctrl-c or either kill the process on the linux side and wait for wsl.exe to return

CatalinFetoiu: thanks. what do you think of something like the below?
wsl.exe killall tcpdump
Stop-Process -InputObject $tcpdumpProcess

OneBlue: `wsl.exe killall tcpdump` -> I think this looks good as long as you add `-u root`

`Stop-Process -InputObject $tcpdumpProcess` -> instead of doing that you should wait for wsl.exe to exit normally. At that point the buffers would be flushed for sure

CatalinFetoiu: updated to use "killall tcpdump" and Wait-Process

diagnostics/networking.sh
CatalinFetoiu: changed to print if the proxy variables are set or not

keith-horton (approved):

diagnostics/collect-networking-logs.ps1
```diff
@@ -0,0 +1,217 @@
...
+    Write-Host "`nSaving logs..."
+}
+finally
+{
+    try
+    {
+        wsl.exe -u root killall tcpdump
+        if ($tcpdumpProcess -ne $null)
+        {
+            Wait-Process -InputObject $tcpdumpProcess
```
OneBlue, line 115: nit: We might want a timeout of let's say 10 seconds here, to make sure we don't wait forever in case something went wrong

OneBlue (approved):

diagnostics/wsl_networking.wprp
```diff
@@ -9,17 +9,62 @@
...
     <EventProvider Id="mup" Name="20c46239-d059-4214-a11e-7d6769cbe020" />
     <EventProvider Id="rfsmon" Name="51734B23-5B7E-4892-BA8E-45BC110B735C" />
     <EventProvider Id="hyperv_storage" Name="c7ad62c6-5c99-5a1b-bbc4-0821ae5b765e" />
     <EventProvider Id="hns" Name="0c885e0d-6eb6-476c-a048-2457eed3a5c1" />
+    <EventProvider Id="netmgmt" Name="93f693dc-9163-4dee-af64-d855218af242" />
+    <EventProvider Id="EventProvider_Microsoft.Windows.Mobile.Provisioning.AppDownload" Name="0BBE6221-EF09-4A3F-82EE-BE00DBB6A98A" />
+    <EventProvider Id="EventProvider_Microsoft.Windows.Mobile.Provisioning.Datastore" Name="42C60CEA-0FE7-4541-A86B-9E11F95BD9BF" />
+    <EventProvider Id="EventProvider_Microsoft.Windows.Mobile.Provisioning.PhoneProvisioner" Name="B876B1FC-C7F1-443E-9012-86677F7DE580" />
+    <EventProvider Id="EventProvider_Microsoft.Windows.Mobile.Provisioning.PPOEM" Name="7EDBED09-1FF7-4FEE-B8C3-5DB694420830" />
+    <EventProvider Id="EventProvider_Microsoft.Windows.Provisioning.API" Name="82ADD491-01D7-4B85-9EAD-192C3CAACA23" />
```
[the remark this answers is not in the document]
CatalinFetoiu: discussed offline, I copied those from wsl.wprp, so that wsl_networking.wprp includes all providers from wsl.wprp

diagnostics/collect-networking-logs.ps1
```diff
@@ -0,0 +1,210 @@
...
+    {
+        Write-Host -ForegroundColor Red "Couldn't start log collection (exitCode: $LastExitCode)"
+    }
+}
+
+# Start packet capture using pktmon
+pktmon start -c --flags 0x1A --file-name "$folder/pktmon.etl" | out-null
+
+# Start WFP capture
+netsh wfp capture start file="$folder/wfpdiag.cab"
```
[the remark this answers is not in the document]
CatalinFetoiu: discussed offline, we can ask separately for InternetClient traces if neeeded. wsl_networking.wprp already covers some of the important providers that InternetClient collect

diagnostics/collect-networking-logs.ps1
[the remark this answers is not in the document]
CatalinFetoiu: tested now and confirmed

PR #11014, opened 2024-01-12
collect-networking-logs.ps1: Collect HNS events and update RestartWslReproMode option

CatalinFetoiu:
Update the script to collect HNS events from the past 24 hours, after the problem was reproduced

Also update the HNS network deletion commands from RestartWslReproMode option, as the NAT network can also have the name "WSL (Hyper-V firewall)"

### DIFF
--- a/diagnostics/collect-networking-logs.ps1
+++ b/diagnostics/collect-networking-logs.ps1
@@ -30,7 +30,9 @@ if (Test-Path $wslconfig)
 if ($RestartWslReproMode)
 {
     # The WSL HNS network is created once per boot. Resetting it to collect network creation logs.
+    # Note: The below commands apply only for WSL in NAT mode
     Get-HnsNetwork | Where-Object {$_.Name -eq 'WSL'} | Remove-HnsNetwork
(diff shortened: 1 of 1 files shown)


diagnostics/collect-networking-logs.ps1
```diff
@@ -30,7 +30,9 @@ if (Test-Path $wslconfig)
 if ($RestartWslReproMode)
 {
     # The WSL HNS network is created once per boot. Resetting it to collect network creation logs.
+    # Note: The below HNS commands apply only to WSL in NAT mode
     Get-HnsNetwork | Where-Object {$_.Name -eq 'WSL'} | Remove-HnsNetwork
+    Get-HnsNetwork | Where-Object {$_.Name -eq 'WSL (Hyper-V firewall)'} | Remove-HnsNetwork
```
OneBlue, line 35: nit: could be merged with the above call by: 
`    Get-HnsNetwork | Where-Object {$_.Name -eq 'WSL' -Or $_.Name -eq 'WSL (Hyper-V firewall)' } | Remove-HnsNetwork

OneBlue (approved):

diagnostics/collect-networking-logs.ps1
CatalinFetoiu: thanks, fixed

PR #11249, opened 2024-03-04
Minor updates to collect-networking-logs.ps1

SlavyMihov:
Need the following changes in order to be able to automate log collection (use _collect-networking-logs.ps1_ in test automation):
* enable log collection to work offline (using local copies of _networking.sh_ and _wsl_networking.wprp_, if existing)
* enable alternative mechanism to end log collection, in addition to keypress - termination of process _tcpdump_.

### DIFF
--- a/diagnostics/collect-networking-logs.ps1
+++ b/diagnostics/collect-networking-logs.ps1
@@ -11,9 +11,26 @@ mkdir -p $folder
 $logProfile = "$folder/wsl_networking.wprp"
 $networkingBashScript = "$folder/networking.sh"
 
-# Download supporting files
-Invoke-WebRequest -UseBasicParsing "https://raw.githubusercontent.com/microsoft/WSL/master/diagnostics/wsl_networking.wprp" -OutFile $logProfile
-Invoke-WebRequest -UseBasicParsing "https://raw.githubusercontent.com/microsoft/WSL/master/diagnostics/networking.sh" -OutFile $networkingBashScript
+# Copy/Download supporting files
+if (Test-Path "$PSScriptRoot/wsl_networking.wprp")
+{
+    Copy-Item "$PSScriptRoot/wsl_networking.wprp" $logProfile
+}
+else
+{
+    Write-Host -ForegroundColor Yellow "wsl_networking.wprp not found in the current directory. Downloading it from GitHub."
+    Invoke-WebRequest -UseBasicParsing "https://raw.githubusercontent.com/microsoft/WSL/master/diagnostics/wsl_networking.wprp" -OutFile $logProfile
+}
+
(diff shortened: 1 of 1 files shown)


keith-horton:
@SlavyMihov , did you verify the script can be run on a clean / customer configuration who installed WSL? then run a 2nd and 3rd time successfully? (the normal way customers run our scripts)

SlavyMihov:
> @SlavyMihov , did you verify the script can be run on a clean / customer configuration who installed WSL? then run a 2nd and 3rd time successfully? (the normal way customers run our scripts)

Yes, I did a few successful test runs - the behavior remains unchanged.

diagnostics/collect-networking-logs.ps1
```diff
@@ -105,9 +122,16 @@ try
           183  # Application 2
 
     $Key = $null
-    while ($Key -Eq $null -Or $Key.VirtualKeyCode -Eq $null -Or $KeysToIgnore -Contains $Key.VirtualKeyCode)
+    while (($Key -Eq $null -Or $Key.VirtualKeyCode -Eq $null -Or $KeysToIgnore -Contains $Key.VirtualKeyCode) -and ($null -eq $tcpdumpProcess -or $tcpdumpProcess.HasExited -eq $false))
```
CatalinFetoiu, line 125: right now the tcpdump process will not exit on its own, it will be stopped only when the user presses a key, so we can't use this as an indication to stop the script

SlavyMihov: Here I am trying to introduce an alternative mechanism to end log collecting. Instead of having the user press a key, we can have a test automation script kill process _tcpdump_ which will be detected here and will effectively end the log collection.
This does not change the behavior of _collect-networking-logs.ps1_ since the very next thing it does is to terminate _tcpdump_ anyway.

CatalinFetoiu: I see, thanks for clarifying

the WSL distro might not have tcpdump installed. In the instructions we have for running the script, we tell users to manually install it (https://github.com/microsoft/WSL/blob/master/CONTRIBUTING.md#collect-wsl-logs-for-networking-issues)

if they don't have installed, we'll exit the script too early because $tcpdumpProcess will be null

can we find a different way to automatically stop the script?

CatalinFetoiu: sorry, I missed that the change already takes into account that $tcpdumpProcess can be null

CatalinFetoiu: this looks good in this case, I am resolving the comment

CatalinFetoiu (approved):

Issue #11497, opened 2024-04-22
Error code: Wsl/CallMsi/REGDB_E_CLASSNOTREG


malezjaa:
### Windows Version

Microsoft Windows [Version 10.0.22631.3447]

### WSL Version

Doesn't work

### Are you using WSL 1 or WSL 2?

- [X] WSL 2
- [ ] WSL 1

### Repro Steps

When I try to install wsl command with `wsl --install` I get this error: `Error code: Wsl/CallMsi/REGDB_E_CLASSNOTREG`

Same with any other command 

### Expected Behavior

To install wsl

### Actual Behavior

Error

malezjaa:
Logs
[WslLogs-2024-04-22_21-42-43.zip](https://github.com/microsoft/WSL/files/15067911/WslLogs-2024-04-22_21-42-43.zip)

malezjaa:
```sh
Get-WindowsOptionalFeature: C:\dev\collect-wsl-logs.ps1:56
Line |
  56 |  Get-WindowsOptionalFeature -Online > $folder/optional-components.txt
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Klasa niezarejestrowana.
Log collection is running. Please reproduce the problem and once done press any key to save the logs.
Saving logs...
100%  [>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>]
Logs saved in: C:\dev\WslLogs-2024-04-22_21-47-38.zip. Please attach that file to the GitHub issue.
```

OneBlue:
@malezjaa: Did you reproduce the issue while log collection was running ? There's no WSL traces in the logs. 

/logs

malezjaa:
I reproduced this error, I ran collection log again, but it looks the same as logs I uploaded first
[WslLogs-2024-04-23_08-39-50.zip](https://github.com/microsoft/WSL/files/15072522/WslLogs-2024-04-23_08-39-50.zip)

OneBlue:
Your WSL installation appears to be in a bad state. Can you try running the following command to reinstall WSL (elevated powershell) and see if that solves the issue?

```
[...]

PR #11500, opened 2024-04-23
Add script and instructions to install the latest version of WSL when the machine is in a bad MSIX state

OneBlue:
This will allow automated diagnostic and recovery on issues like #11497

### DIFF
--- a/triage/install-latest-wsl.ps1
+++ b/triage/install-latest-wsl.ps1
@@ -1,0 +1,50 @@
+#Requires -RunAsAdministrator
+
+# This script downloads and installs the latest version of the WSL MSI package
+
+$ErrorActionPreference = "Stop"
+Set-StrictMode -Version Latest
+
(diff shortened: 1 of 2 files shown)


triage/install-latest-wsl.ps1
```diff
@@ -0,0 +1,50 @@
+#Requires -RunAsAdministrator
+
+# This script downloads and installs the latest version of the WSL MSI package
+
+$ErrorActionPreference = "Stop"
+Set-StrictMode -Version Latest
+
+$release = Invoke-WebRequest 'https://api.github.com/repos/microsoft/WSL/releases/latest' | ConvertFrom-Json
+$systeminfo = & systeminfo | findstr /C:"System Type"
```
pmartincic, line 9: I think this is also under: `$env:PROCESSOR_ARCHITECTURE`

PR #11528, opened 2024-04-27
Update DistributionInfo.json Kali Linux Arm

crramirez:
Kali Linux does include an ARM build in the bundle, so it should be shown to users of ARM devices


benhillis (approved):
nice catch, thanks!

PR #11815, opened 2024-07-19
Update missing logs automatic message to mention collecting networking logs

CatalinFetoiu:
When a github issue is created without logs, an automatic message is posted asking the user to collect logs.

The PR updates this message to mention that for networking issue, the networking logs script should be used


benhillis (approved):

keith-horton (approved):
Thanks!

Issue #12282, opened 2024-11-16
错误代码: Wsl/CallMsi/Install/REGDB_E_CLASSNOTREG


Sherry520:
### Windows Version

Microsoft Windows [版本 10.0.19045.4780]

### WSL Version

2.3.26

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Repro Steps

wsl --update

### Expected Behavior

I expected it would update to 2.3.26 successfully.

### Actual Behavior

When I updated for the first time, an error occurred:
```Could not write value to key \SOFTWARE\Classes\Drive\shell\WSL```,and with a log file.
Then I followed the solution as shown in https://github.com/microsoft/WSL/issues/11697#issuecomment-2168104220,
Then I ran `wsl --update` again, another error occurred:
```
没有注册类
错误代码: Wsl/CallMsi/Install/REGDB_E_CLASSNOTREG
```

### Diagnostic Logs

[wsl-install-logs.txt](https://github.com/user-attachments/files/17783752/wsl-install-logs.txt)

**Comment by Sherry520** [comment]
```
.\collect-wsl-logs.ps1

wsl --update
没有注册类
错误代码: Wsl/CallMsi/Install/REGDB_E_CLASSNOTREG
```
[WslLogs-2024-11-16_14-51-44.zip](https://github.com/user-attachments/files/17783810/WslLogs-2024-11-16_14-51-44.zip)

**Comment by Sherry520** [comment]
By reference this method: https://github.com/microsoft/WSL/issues/12278#issuecomment-2479699777, sovled error "Wsl/CallMsi/Install/REGDB_E_CLASSNOTREG". But for Chinese language users, additional operations are required in powershell:

中文语言用户参考以下我修改后的命令:
(For other non-English users, check the corresponding language number)
```
# 修改界面语言为英文
chcp 437

Invoke-WebRequest -UseBasicParsing "https://raw.githubusercontent.com/microsoft/WSL/master/triage/install-latest-wsl.ps1" -OutFile install-latest-wsl.ps1
Set-ExecutionPolicy Bypass -Scope Process -Force
.\install-latest-wsl.ps1

# 修改界面语言为中文
chcp 936
```

### Things to note:
[...]

PR #12284, opened 2024-11-17
Fix language issues in install WSL ps1 script

Sherry520:
https://github.com/microsoft/WSL/issues/12282#issuecomment-2480894130 The install-latest-wsl.ps1 script cannot parse the operating "System Type" '$systeminfo = & systeminfo | findstr /C:"System Type"' due to language problems. In the Chinese interface environment, "System Type" is the Chinese "系统类型".

### DIFF
--- a/triage/install-latest-wsl.ps1
+++ b/triage/install-latest-wsl.ps1
@@ -3,7 +3,7 @@
 # This script downloads and installs the latest version of the WSL MSI package
 
 # Get current language code
-$chcp_num = (chcp) -replace '\D+\(\d+)','$1'
+$chcp_num = (chcp) -replace '[^\d]+(\d+).*','$1'
 
 # Set language to english
 chcp 437
--- a/triage/install-latest-wsl.ps1
+++ b/triage/install-latest-wsl.ps1
@@ -2,6 +2,12 @@
 
 # This script downloads and installs the latest version of the WSL MSI package
 
+# Get current language code
+$chcp_num = (chcp) -replace '\D+\(\d+)','$1'
+
+# Set language to english
+chcp 437
+
 $ErrorActionPreference = "Stop"
 Set-StrictMode -Version Latest
 
@@ -48,3 +54,6 @@ if ($exitCode -Ne 0)
 Write-Host 'Installation complete'
 
 Remove-Item $target -Force
+
+# Restore original language
+chcp $chcp_num


linghengqian (commented):
- So, what is preventing the current PR from being merged? I can say that the bug corresponding to this PR prevents all Windows using non-English environments from installing WSL.
- image
- I will say I'm still confused as to why we use `systeminfo`, it's just too weird.

- Considering this is windows and not linux, can't we use dotNet's `Environment.Is64BitOperating` and `HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment` registry? See https://learn.microsoft.com/en-us/troubleshoot/windows-server/setup-upgrade-and-drivers/determine-the-type-of-processor and https://learn.microsoft.com/en-us/dotnet/api/system.environment.is64bitoperatingsystem?view=net-9.0 .

linghengqian (commented):
- Some new changes happened at https://github.com/microsoft/WSL/issues/12282#issuecomment-2670913566 .

triage/install-latest-wsl.ps1
```diff
@@ -2,6 +2,12 @@
 
 # This script downloads and installs the latest version of the WSL MSI package
 
+# Get current language code
+$chcp_num = (chcp) -replace '\D+\(\d+)','$1'
```
OneBlue: Unfortunately this expression doesn't work on my end: 


```
The regular expression pattern \D+\(\d+) is not valid.
At install-latest-wsl.ps1:6 char:1
+ $chcp_num = (chcp) -replace '\D+\(\d+)','$1'
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (\D+\(\d+):String) [], RuntimeException
    + FullyQualifiedErrorId : InvalidRegularExpression

```

OneBlue (changes requested):

triage/install-latest-wsl.ps1
Sherry520: The problem has been solved. https://github.com/microsoft/WSL/pull/12284/commits/abe03f0a9f95ce12bda3733100b9d9eaa2f8402e

OneBlue (approved):
LGTM, thank you !

PR #12314, opened 2024-11-25
Introduce new CI to validate new distributions entries

OneBlue:
Sample CI run: https://github.com/microsoft/WSL/pull/12315 

This change will help us review & make suggestions to incoming tar based WSL distribution updates

### DIFF
--- a/.github/workflows/modern-distributions.yml
+++ b/.github/workflows/modern-distributions.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     paths: ['distributions/**']
 
+permissions:
+  pull-requests: write
+
 jobs:
   check:
     name: Validate tar based distributions changes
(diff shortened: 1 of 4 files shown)


.github/workflows/modern-distributions.yml
```diff
@@ -0,0 +1,34 @@
...
+          run: pip install -r distributions/requirements.txt
+          shell: bash
+
+        - name: Run validation
+          run: | 
+            python distributions/validate-modern.py \
+            --repo-path . \
+            --compare-with-branch 'origin/${{ github.base_ref }}' \
+            --manifest distributions/DistributionInfo.json \
+            --github-token '${{ secrets.GITHUB_TOKEN }}' \
```
benhillis, line 31: How is this token managed?

OneBlue: It's a token that's generating by github actions. We don't actively have to manage it (https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication)

craigloewen-msft (approved):

benhillis (approved):

PR #12407, opened 2024-12-25
Update DistributionInfo.json

angualberto:


craigloewen-msft:
Hi it looks like the links you provided: `https://publicwsldistros.blob.core.windows.net/wsldistrostorage/Fedora-35-230801_ARM64.Appx` don't exist. For making a custom distro I recommend reading these docs :

https://learn.microsoft.com/en-us/windows/wsl/build-custom-distro

PR #12451, opened 2025-01-10
Add SUSE distributions to DistributionInfo.json

sbradnick:
Adding "openSUSE" section to "ModenDistributions".


OneBlue:
Thank you @sbradnick, this is looking good. 

This is just missing a default entry for `openSUSE` (`Default=True`). That's for wsl to know which version to install if the user runs: 

```
wsl --install opensuse
```

sbradnick:
> Thank you @sbradnick, this is looking good. 
> 
> This is just missing a default entry for `openSUSE` (`Default=True`). That's for wsl to know which version to install if the user runs: 
> 
> ```
> wsl --install opensuse
> ```

I've set TW as the default.

OneBlue:
Thanks @sbradnick. 

Sorry I just noticed this: Could you use gz compression instead of xz ? 
The reason for that is that wsl1 unfortunately doesn't support xz 

Also could you change the distribution url's to have the .wsl extension  ? 
That way if users manually downloaded the distribution, they could just install them via double clicking.

sbradnick:
> Sorry I just noticed this: Could you use gz compression instead of xz ?

kiwi-ng defaults to `xz` and that doesn't appear to be a configurable option --> https://osinside.github.io/kiwi/image_description/elements.html

So at this point, I'd say no I can't.

It also isn't an option to build a `.tar.xz` image but as a last step, rename it to `.wsl`. But that's such a trivial action for the end user ...
 
> The reason for that is that wsl1 unfortunately doesn't support xz

I don't believe we're overly concerned with support for wsl1 on our side.

> Also could you change the distribution url's to have the .wsl extension  ? 
> That way if users manually downloaded the distribution, they could just install them via double clicking.

I can going forward (don't really want to cut a new GH release just for that); but if that's the expectation, I wasn't aware it applied here since `--web-download` isn't a mouse action. I'd agree it's convenient and 'neat', but so trivial it should be optional, not required.

OneBlue (approved):

benhillis (changes requested):
We're discussing internally how to handle xz support. Holding off on merging this until we have decided on a stance here.

sbradnick:
Any updates on this? I dug into the kiwi-ng code and was able to monkey patch it to produce a gzip tarball vs. xz tarball but it's seemingly not something upstream kiwi seems interested in.

Additionally, we were able to get a repository in the SUSE org for SLE tarballs so I can start putting together the SUSE "flavor" as well.

OneBlue:
Thank you @sbradnick. We've decided to add support for tar.xz tarballs for WSL1 in the next WSL pre-release, so this change is good to go !

OneBlue:
Latest iteration LGTM. Feel free to merge whenever ready.

OneBlue:
Thank you for rebasing @sbradnick. 

Is this change ready to be merged ? If so I can merge it now

sbradnick:
> Is this change ready to be merged ? If so I can merge it now

I'd say yes ✅; if you guys are adding xz support for WSLv1 in the next release, then all should be good. Next update I'll change the file extension on the openSUSE side to `.wsl`.

OneBlue:
Ok sounds good. 

The next pre-prerelease version will support .tar.xz indeed, so merging this now. 

Thank you !

PR #12472, opened 2025-01-17
Add Kali to modern distributions

Gamb1ts:


Gamb1ts:
> @Gamb1ts please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
>  > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

@microsoft-github-policy-service agree company="OffSec"

OneBlue:
Thank you @Gamb1ts ! 

Feel free to ignore the bot's warnings on the masked systemd units, I need to go fix that ! 

The tar & manifest look great ! One small suggestion that I have would to use `systemctl is-system-running` in `/usr/lib/wsl/wsl-oobe` so we could get remove the `sleep 25` and make the distribution start a bit faster.

Something like this should do the trick: 

```
timeout 25 bash -c 'while ! systemctl is-system-running | grep -E "running|degraded" ; do sleep 1; done'  || echo "warning: timed out waiting for systemd to start"
```

benhillis:
What's the goal of sleeping to wait for systemd in the oobe script? If you want to make sure that no user processes are launched before systemd is up, that should already be guaranteed as part of WSL. Is that not the behavior you are seeing?

Gamb1ts:
Early in my testing I found that the system setup was too quick which led to issues using certain commands. However, this was prior to masking the recommended systemd units. I will re-visit this and see if in its current state the wait is necessary. 

Unfortunately I will have to mark this as WIP until next week, we will need to change where we host the image. Next week I will supply the new URL and checksums.

Gamb1ts:
I have disabled the wait on systemd, I was able to confirm that it is not needed. I have also updated the URL to utilize our mirror kali.download. Unless there is any other feedback, the Kali team is good with the PR at this stage.

OneBlue:
FYI when I tried it locally, I got the following output: 

```
Please create a default Kali WSL user. The username does not need to match your Windows username.
For more information visit: https://aka.ms/wslusers
Enter new UNIX username: blue
logger: socket /dev/log: No such file or directory
logging to syslog failed: command line logger --id=139 --tag=adduser --priority=user.info  -- Adding user `blue' ... returned error: 256
logger: socket /dev/log: No such file or directory
logging to syslog failed: command line logger --id=139 --tag=adduser --priority=user.info  -- Adding new group `blue' (1000) ... returned error: 256
logger: socket /dev/log: No such file or directory
logging to syslog failed: command line logger --id=139 --tag=adduser --priority=user.info  -- Adding new user `blue' (1000) with group `blue (1000)' ... returned error: 256
logger: socket /dev/log: No such file or directory
logging to syslog failed: command line logger --id=139 --tag=adduser --priority=user.info  -- Creating home directory `/home/blue' ... returned error: 256
logger: socket /dev/log: No such file or directory
logging to syslog failed: command line logger --id=139 --tag=adduser --priority=user.info  -- Copying files from `/etc/skel' ... returned error: 256
```

Looking through dmesg, I can see the following logline: 

`Legacy cgroup v1 support selected. This is no longer supported. Will proceed anyway after 30s. `. This is most likely what's causing the systemd delay that you originally saw. 

Let me see if this can be solved from the WSL side.

OneBlue:
Looks like we can solve this on the WSL via: `SYSTEMD_CGROUP_ENABLE_LEGACY_FORCE=1`.

In the meantime, I'd recommend adding the logic to wait to systemd to the OOBE, so it doesn't print the above the error. I refined my earlier suggestion to make the output a bit cleaner: 


```
echo 'Waiting for systemd to start...'
timeout 25 bash -c 'while ! systemctl is-system-running 2> /dev/null | grep -E "running|degraded > /dev/null"; do sleep 1; done' || echo "Warning: timed out waiting for systemd to start"
```

(that way the user won't get `Failed to connect to bus: No such file or directory`  on stderr when waiting for systemd)

Gamb1ts:
Yeah that was the same issue I was seeing previously. With the latest pre-release WSL installed I was no longer having the issue FWIW. 

I have re-added the systemd wait and re-built the images. The checksums have been updated but the link is the same. The site is heavily cached so may need to clear it.

OneBlue:
Thank you @Gamb1ts. I restarted the pipeline, and the hashes match now. 

Although I still see the systemd wait logic being commented `/usr/lib/wsl/wsl-oobe` (and I validated the distribution hash, so I don't think I got a cached file).

Gamb1ts:
I will rebuild the image and verify locally it is back in

OneBlue:
Thank you @Gamb1ts. Just tried it out, the OOBE works perfectly now. 

Feel free to merge whenever ready.

OneBlue (approved):

Gamb1ts:
Thank you!

> Feel free to merge whenever ready.

Unfortunately I am unable to merge this as I don't have the permissions. Are you able to?

OneBlue:
Merged !

PR #12482, opened 2025-01-20
feat: add an unofficial Arch Linux distro to the modern distributions list

ateoi:
Tar image is based on the official Arch Linux Docker image. Scripts used to build the image are available at <https://github.com/DevelopersCommunity/archlinux-wsl>

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -1,5 +1,17 @@
 {
-    "ModernDistributions": {},
+    "ModernDistributions": {
+        "Arch-Linux-Unofficial": [
+            {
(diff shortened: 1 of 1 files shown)


craigloewen-msft:
Hi @ateoi ! Since this is an unofficial version of Arch, we aren't going to take this pull request. The reasoning here is that we don't want to confuse general WSL users with an 'unofficial' version of the Arch distro. If you'd like to include Arch in this list, please reach out and work with official Arch maintainers to put an official Arch version here! Otherwise you can always direct people to the download link you have listed there and have users manually download your distro. Then they can just double click to install it. 

Thank you for the submission!

Issue #11693, opened 2024-06-14
WSL no internet connection / DNS issues


cyberjj999:
### Windows Version

Microsoft Windows [Version 10.0.22621.3737]

### WSL Version

2.2.4.0

### Are you using WSL 1 or WSL 2?

- [X] WSL 2
- [ ] WSL 1

### Kernel Version

5.15.153.1-microsoft-standard-WSL2

### Repro Steps

- `ping google.com` returns `ping: google.com: Temporary failure in name resolution`
- `pip install streamlit` or other packages return network error message; same for sudo-apt install

This indicate a clear network problem.

### Expected Behavior

No issues with network problem - `ping` works and `pip install <package>` should work.

## Clear Network/Internet Connection Problem:
- `ping google.com` returns `ping: google.com: Temporary failure in name resolution`
- `pip install streamlit` or other packages return network error message; same for sudo-apt install

My host machine (Windows 11) has no internet issues at all.

## What I've Tried
1. Disable windows firewall entirely shutdown WSL and `ping google.com` again: doesn't work

2. Ran the following command to flush my dns on windows:
```
netsh winsock reset 
netsh int ip reset all
netsh winhttp reset proxy
ipconfig /flushdns
``` 
then restarted my computer: doesn't work

3. Updated my `/etc/resolv.conf` and `/etc/wsl.conf` to put nameserver of `8.8.8.8` and `8.8.8.4` and even make the `/etc/resolve.conf` immutable... and it doesn't work.
```
sudo rm /etc/resolv.conf
sudo bash -c 'echo "nameserver 8.8.8.8" > /etc/resolv.conf'
sudo bash -c 'echo "[network]" > /etc/wsl.conf'
sudo bash -c 'echo "generateResolvConf = false" >> /etc/wsl.conf'
sudo chattr +i /etc/resolv.conf
```

4. Disabled "fast start-up" option in Power Options then restarted my comp... still doesn't work

5. Chnaged from my company WiFi to my personal mobile hotspot - doesn't work

## Suspected Reasons

1. Change of network (VPN?) but disabling VPN doesn't yield a meaningful difference

2. Windows Update (including Quality Updates)

Somehow I have WSL update automatically with an old kernel version though my WSL Ubuntu is installed from Microsoft store?
[enter image description here][1]

  [1]: https://i.sstatic.net/Q6NEXInZ.png

But my WSL version seems to be okay
```
WSL version: 2.2.4.0
Kernel version: 5.15.153.1-2
```

[...]

PR #12520, opened 2025-01-30
Add superuser detection in networking logs script

1MLightyears:
Solve a side case not considered in #11693 by adding superuser detection in the diagnostic script diagnostics/collect-networking-logs.ps1.
Didn't find any other calls to wsl.exe -u root, which should all be changed at the same time. Did I miss anything?

### DIFF
--- a/diagnostics/collect-networking-logs.ps1
+++ b/diagnostics/collect-networking-logs.ps1
@@ -126,6 +126,19 @@ mkdir -p $folder
 
 $logProfile = "$folder/wsl_networking.wprp"
 $networkingBashScript = "$folder/networking.sh"
+$superUser = "root" # user name of the super user.
+
+# Detect the super user first.
+# Actually it's not definite that the super user is named "root". Instead, a user with uid=0 is what we are looking for. See #11693.
+$_passwd = wsl -- sh -c "getent passwd"
+$_lines = $_passwd -split "`n"
+foreach ($_l in $_lines) { 
+    $_col = $_l -split ":"
+    if ($_col[2] -eq "0") {  
+        $superUser = $_col[0] 
+        break
+    }
+}
 
 # Copy/Download supporting files
 if (Test-Path "$PSScriptRoot/wsl_networking.wprp")
@@ -158,7 +171,7 @@ if (Test-Path $wslconfig)
 }
 
 # Collect Linux & Windows network state before the repro
-& wsl.exe -u root -e $networkingBashScript 2>&1 > $folder/linux_network_configuration_before.log
+& wsl.exe -u $superUser -e $networkingBashScript 2>&1 > $folder/linux_network_configuration_before.log
 
 Collect-WindowsNetworkState "before_repro"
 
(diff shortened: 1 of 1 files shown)


1MLightyears:
> @1MLightyears please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

OneBlue:
Hi @1MLightyears. Thank you for doing this ! 

The overall change looks good, but let's remove the /etc/passwd parsing and assume that `root` is uid=0 account so we can simplify to `wsl.exe -u root` when invoking WSL

1MLightyears:
> Hi @1MLightyears. Thank you for doing this !
> 
> The overall change looks good, but let's remove the /etc/passwd parsing and assume that `root` is uid=0 account so we can simplify to `wsl.exe -u root` when invoking WSL

Thank you @OneBlue , but that will be exactly what it was, and it triggered an error on my WSL. In case you have no time for reading #11693 (It's absurdly long, and I have to admit that), a useful little trick to avoid ssh weak password exploit is to **change the name of superuser from `root` to a custom name.** Thus, implying the superuser is `root` would cause an error. The `wsl.exe` call would directly fail and exit.

I understand that this would introduce more complexity to the script, but I think that side cases should also be taken into consideration. If you don't want to merge it now, maybe we can just leave this open for some time and see if any more modifications can be made to the script, as #11693 is still growing even longer and more people are likely to diagnostic with the script, which implies that more bugs may be found in the future.

OneBlue:
Ok that's fair. To find a simpler approach to that, how we about instead of looking for the username ourselves we used: 

```
 wsl -- id -nu 0
``` 
?

CatalinFetoiu:
thanks @1MLightyears for submitting the PR!

1MLightyears:
> Ok that's fair. To find a simpler approach to that, how we about instead of looking for the username ourselves we used:
> 
> ```
>  wsl -- id -nu 0
> ```
> 
> ?

Oh, that's concise. I don't know this before and had trouble passing a command with pipe to `wsl.exe`. That was why I chose to parse the output in powershell as in the commit message :(

Changes applied √

CatalinFetoiu (approved):

diagnostics/collect-networking-logs.ps1
```diff
@@ -127,6 +127,10 @@ mkdir -p $folder
 $logProfile = "$folder/wsl_networking.wprp"
 $networkingBashScript = "$folder/networking.sh"
 
+# Detect the super user first.
+# Actually it's not definite that the super user is named "root". Instead, a user with uid=0 is what we are looking for. See #11693.
+$superUser = wsl -- id -nu 0  # user name of the super user.
```
OneBlue: Let's use this to make sure this doesn't fail if there's an stderr warning, or a "wsl" alias: 

```
$superUser = & wsl.exe -- id -nu 0
```

OneBlue (changes requested):

diagnostics/collect-networking-logs.ps1
1MLightyears: Makes sense.

OneBlue (approved):

FELLIX000 (approved):

PR #12534, opened 2025-01-31
Change the modern distribution workflow to use pull_request: 

OneBlue:


.github/workflows/modern-distributions.yml
```diff
@@ -11,19 +11,10 @@ jobs:
     permissions:
         pull-requests: write
```
benhillis, line 12: I don't think we want write access here.

OneBlue: The pull_request trigger requires review before running the pipeline so I think it's acceptable in this workflow

benhillis (approved):

craigloewen-msft (approved):

PR #12617, opened 2025-02-21
Update to 24.04.2 for Ubuntu and Ubuntu 24.04 WSL instances

didrocks:
We now have 24.04.2 released and hosted on our official ubuntu infrastructure. Point to those WSL images.

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -6,12 +6,12 @@
                 "FriendlyName": "Ubuntu",
                 "Default": true,
                 "Amd64Url": {
-                    "Url": "https://publicwsldistros.blob.core.windows.net/wsldistrostorage/Ubuntu2404-250130_x64.wsl",
-                    "Sha256": "64d20e0b641d2d3ad00c8015f9e905a63932f499bd743ab77224f5242539da8d"
+                    "Url": "https://releases.ubuntu.com/noble/ubuntu-24.04.2-wsl-amd64.wsl",
+                    "Sha256": "5d1eea52103166f1c460dc012ed325c6eb31d2ce16ef6a00ffdfda8e99e12f43"
                 },
                 "Arm64Url": {
-                    "Url": "https://publicwsldistros.blob.core.windows.net/wsldistrostorage/Ubuntu2404-250130_ARM64.wsl",
-                    "Sha256": "a97f5513a3c9946b4b4ac3035eb441e9a6cb797fe46f0a7462c81b842d25a608"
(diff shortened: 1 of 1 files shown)


OneBlue:
Thank you @didrocks. The update looks good.

Looking through the OOBE process, I have two minor suggestions: 

1) In `/usr/lib/wsl/wsl-setup:102`:

This could use ${WSL_DISTRO_NAME} like this (which should be more reliable than calling wslpath):

```
echo "Provisioning the new WSL instance ${WSL_DISTRO_NAME}"
``` 

2) In `/usr/lib/wsl/wsl-setup:135`:

Instead of writing to `/etc/wsl.conf`, the oobe script could instead add `--uid 100` in `/usr/lib/wsl/wsl-setup:39` like this:

```
/usr/sbin/adduser --uid 1000 --quiet --gecos ''  "$username"
```

Which will automatically work because `oobe.defaultUid` is already set to 1000 in `/etc/wsl-distribution.conf` (and WSL keeps track of the default uid outside of the distro itself). 


Anyway, those are not blockers at all so let me know if you prefer to merge as-is and potentially fix those in a future release or if you prefer to do another iteration on this PR.

OneBlue (approved):

didrocks:
Hey @OneBlue Thanks for the review! This could be as is as the current version referenced have already the same issue.

For 1. Oh nice, we will definitively use it.
For 2. we may work with pre-installed images where a non interactive user has been already preprovisionned with `1000`. This is why we then update the `wsl.conf`, as updating `/etc/wsl-distribution.conf` is more a distribution file. The issue is the one we raised as feedback with the new format: updating in the oobe `wsl.conf` the user id isn’t then logging you in with the user for the current session (only the next one), as the file is probably not released.

So, unsure about 2. we will tackle 1., but this will be in the next batch of updates as this is a no-change, apart from hosting our official 24.04.2 images. I think you can merge it. Thanks!

crramirez:
Writing to `/etc/wsl.conf` also guarantees that if the distro is exported and then imported on WSL version prior 2.4, it will still use the right user and not root.

OneBlue:
> Hey @OneBlue Thanks for the review! This could be as is as the current version referenced have already the same issue.

Sounds good @didrocks. Merged

> The issue is the one we raised as feedback with the new format: updating in the oobe `wsl.conf` the user id isn’t then logging you in with the user for the current session (only the next one), as the file is probably not released.

There was a bug causing this behavior an early preview on this feature, but it's resolved now (the first shell the user gets should always use the correct user account via `defaultUid`, even immediately after OOBE).

Issue #12704, opened 2025-03-16
"TypeError: read_tar() takes 3 positional arguments but 4 were given" when running validate-modern.py


Antiz96:
Hi,

Since [these recent changes](https://github.com/microsoft/WSL/commit/1e8588fdeaa4493cd43f62bc2e456bb46f6cf2ea), the validate-modern.py script fails with `TypeError: read_tar() takes 3 positional arguments but 4 were given`.

It indeed seems like the `read_tar` function [now expects 3 arguments](https://github.com/microsoft/WSL/blob/master/distributions/validate-modern.py#L323) while [4 arguments are passed from the main function](https://github.com/microsoft/WSL/blob/master/distributions/validate-modern.py#L68).

Antiz96:
```
Traceback (most recent call last):
  File "/builds/mark/archlinux-wsl/workdir/clone/distributions/validate-modern.py", line 499, in <module>
    main()
    ~~~~^^
  File "/builds/mark/archlinux-wsl/workdir/venv/lib/python3.13/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/builds/mark/archlinux-wsl/workdir/venv/lib/python3.13/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/builds/mark/archlinux-wsl/workdir/venv/lib/python3.13/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/mark/archlinux-wsl/workdir/venv/lib/python3.13/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/builds/mark/archlinux-wsl/workdir/clone/distributions/validate-modern.py", line 68, in main
    read_tar(tar, '<none>', fd, ARM64_ELF_MAGIC if arm64 else  X64_ELF_MAGIC)
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: read_tar() takes 3 positional arguments but 4 were given
```

mhegreberg:
Im also experiencing this

elsaco:
[...]

PR #12707, opened 2025-03-17
Fix --tar argument handling in validate-modern.py

OneBlue:
Solves https://github.com/microsoft/WSL/issues/12704


mhegreberg:
I tested this fix locally and it seems to fix the issue for the Arch linux WSL Image

benhillis (approved):

PR #12710, opened 2025-03-18
Add Fedora Linux to distribution list

jeremycline:
Fedora has started building WSL images in Fedora 42. The Fedora 42 beta shipped today so images are now available.

I'm not sure how folks feel about adding a beta image to this list. Fedora 42 stable is due to arrive April 15th or April 22nd, and it's totally fine if we want to wait until then to add Fedora images.

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -157,6 +157,21 @@
                     "Sha256": "13f675407f0fa792d8561f7e7866e748899a62f39b7e7dd1f5561b27ba7d7a1a"
                 }
             }
+        ],
+        "Fedora": [
+            {
+                "Name": "FedoraLinux-42",
+                "FriendlyName": "Fedora Linux 42",
+                "Default": true,
+                "Amd64Url": {
+                    "Url": "https://download.fedoraproject.org/pub/fedora/linux/releases/42/Container/x86_64/images/Fedora-WSL-Base-42-1.1.x86_64.tar.xz",
+                    "Sha256": "99fb3d05d78ca17c6815bb03cf528da8ef82ebc6260407f2b09461e0da8a1b8d"
+                },
+                "Arm64Url": {
+                    "Url": "https://download.fedoraproject.org/pub/fedora/linux/releases/42/Container/aarch64/images/Fedora-WSL-Base-42-1.1.aarch64.tar.xz",
+                    "Sha256": "a5a2ceb8ca56b7245b909d021b0fd620427db349f02b8ef3b82b741bcb5611cd"
+                }
+            }
         ]
     },
     "Default": "Ubuntu",


OneBlue:
Thank you for publishing this @jeremycline ! 

This is looking pretty good. We don't have concerns with publishing a beta image in this manifest, as long as it's updated once stable. 

Some minor feedback on the manifest and the distribution tars: 

1) If possible, could you rename the tars to ".wsl" ? That would allow users directly downloading them to install them by double clicking on it

2) Regarding the distribution name, we would prefer to use `Fedora-42` directly, so that way running wsl --install won't break after this version is no longer in beta 

3) Could you add a shortcut icon the distribution? (`shortcut.icon` in `/etc/wsl-distribution.conf`) ? That way Fedora would have its own icon in the start menu instead of the default one 

Let me know if you have any questions, or anything else I can help with

jeremycline:
> Thank you for publishing this @jeremycline !
> 
> This is looking pretty good. We don't have concerns with publishing a beta image in this manifest, as long as it's updated once stable.
> 
> Some minor feedback on the manifest and the distribution tars:
> 
> 1. If possible, could you rename the tars to ".wsl" ? That would allow users directly downloading them to install them by double clicking on it

This is on our radar and is tracked at https://github.com/OSInside/kiwi/issues/2678 - I can likely devote some time to getting it taken care of soon, but it definitely won't happen with the Fedora 42 images. I could _possibly_ convince a release engineer to manually rename the F42 final images before pushing them out, but I'd rather not add manual steps to their release process.

Is it okay to initially include these with the `tar.xz` suffix with the expectation that we improve the tooling soon?

> 2. Regarding the distribution name, we would prefer to use `Fedora-42` directly, so that way running wsl --install won't break after this version is no longer in beta

Totally fair. Given this, though, I think I'd like to wait a few weeks for the Fedora 42 final release to avoid confusion. I'll mark this as a draft for now and update the PR with the final images.

> 3. Could you add a shortcut icon the distribution? (`shortcut.icon` in `/etc/wsl-distribution.conf`) ? That way Fedora would have its own icon in the start menu instead of the default one
> 

I've got a [pull request](https://pagure.io/fedora-logos/pull-request/27) open to get a logo into the image and will do my best to get this and the necessary wsl.conf change in before the change freeze. Fingers crossed it'll happen.

> Let me know if you have any questions, or anything else I can help with

Thanks!

benhillis:
Sounds great, thanks Jeremy!

PatrickLang:
:) can't wait for this one. If there are any plans around testing updates to WSL or this distro list in CI - let me know how I can help. I wrote some basic install & functional tests at https://github.com/PatrickLang/fedora-wsl-tests that we'll probably put into use on the Fedora side. Getting Windows up in the Fedora test environments is a new area, so I have been testing in a fleet of VMs in my home lab instead.

jeremycline:
Today is Fedora 42 release day, so I've updated the images to the GA image.

The shortcut icon did make it into this build. The CI error about `shortcut.enabled` being an unexpected key is a bit confusing since it's [documented as being valid](https://learn.microsoft.com/en-us/windows/wsl/build-custom-distro#add-the-wsl-distribution-configuration-file).

OneBlue:
Thank you for doing this @jeremycline!  I've updated the validation script in #12826. The error should go away if you rebase your branch. 

Looking at the tar, I can see that many of the files have selinux extended attributes, which unfortunately break installation with WSL1 (you can reproduce this with `wsl --install --from-file  Fedora-WSL-Base-42-1.1.x86_64.tar.xz   --version 1`).

Would it be possible to create a tar with `'--xattrs-exclude=security.selinux*'` ? I manually created a tar like that from yours, and that custom tar works with WSL1.

Besides that selinux issue, everything looks good to me !

jeremycline:
> Looking at the tar, I can see that many of the files have selinux extended attributes, which unfortunately break installation with WSL1 (you can reproduce this with `wsl --install --from-file Fedora-WSL-Base-42-1.1.x86_64.tar.xz --version 1`).
> 
> Would it be possible to create a tar with `'--xattrs-exclude=security.selinux*'` ? I manually created a tar like that from yours, and that custom tar works with WSL1.

Hmm. We didn't consider or test against WSL1. While I can see about excluding the SELinux extended attributes in future releases, there's not much I can do for Fedora 42 as the release was this morning.

OneBlue:
> Hmm. We didn't consider or test against WSL1. While I can see about excluding the SELinux extended attributes in future releases, there's not much I can do for Fedora 42 as the release was this morning.

I think skipping WSL1 support for this first release is fine (your call). 

I'll approve the change as it is, and if modifying the already released tar is not possible, fixing this in the next release should be OK.

Let me know if you want to merge this as it is, or if you'd like to try to update the existing tar.

OneBlue (approved):

craigloewen-msft (approved):

PatrickLang:
If anyone's looking for the docs, they're at https://docs.fedoraproject.org/en-US/cloud/wsl/


I'm fixing the URLs for today's release at https://pagure.io/fedora-cloud/docs-site/pull-request/3#

jeremycline:
> > Hmm. We didn't consider or test against WSL1. While I can see about excluding the SELinux extended attributes in future releases, there's not much I can do for Fedora 42 as the release was this morning.
> 
> I think skipping WSL1 support for this first release is fine (your call).
> 

Yeah, I'm okay with it.

> I'll approve the change as it is, and if modifying the already released tar is not possible, fixing this in the next release should be OK.
> 
> Let me know if you want to merge this as it is, or if you'd like to try to update the existing tar.

Great, I appreciate it.

To provide more details that I left out of my previous comment in my rush to start cooking dinner, the issue is that Fedora relies on a community of mirrors to distribute content, and while images are built nightly after the release, the nightly images aren't retained forever and are never sent to the mirror network. What goes out on release day (or really, a few days before) is what it is.

For some images, like Azure, AWS, and GCP cloud images, we upload these nightly images to the respective cloud platforms, and things like container images are pushed to a registry, so nightly builds are retained in that way. The WSL image doesn't have a great place (at the moment) to be retained, although in the coming months I think it would be interesting to come up with a plan there and perhaps set up CI and automated updates to the image we provide here. In the shorter term we can document where adventurous users who want to test out the images on WSL1 can grab the nightly images when we've fixed the extended attributes.

All that to say, I'd prefer not to fix the release image and merge this as is, but perhaps we won't need to wait until F43 to fix things for WSL1.

jeremycline:
I rebased with one small tweak (sorry): I changed the name to align with the value in `/etc/os-release` which is `Fedora Linux`, so the new name is `FedoraLinux-42` (which also matches the `FriendlyName` value).

OneBlue:
Ok that sounds good. 
Merged! 

Thank you for doing this !

massimozaffarano1979-boop:
Ciao

PR #12715, opened 2025-03-19
Add AOSC OS to modern distributions.

KexyBiscuit:
Hi, I'm a maintainer of AOSC OS, a general purpose Linux distribution that strives to simplify user experience and improve free and open source software for day-to-day productivity.

Our distribution has been available on Microsoft Store for years <https://www.microsoft.com/store/apps/9NMDF21NV65Z>, and we'd like to adopt the modern distribution solution.

The .wsl files are identical to the version I just submitted to the Store for certification, and are generated by our in-house tool <https://github.com/AOSC-Dev/aoscbootstrap/blob/9182de9024ad24c6294233eb4eb58623be66c238/contrib/generate-releases.sh>.

By the way, we also release AppX sideload packages for machines with no Microsoft Store available, however I see that all other distributions have their AppX on Microsoft's server, does Microsoft accept new submissions to that list too?

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -138,6 +138,21 @@
                     "Sha256": "9c5804fc58108a138f53ff111961e17217b1cf429f8e5c8365b52062b621e8e5"
                 }
             }
+        ],
+        "AOSCOS": [
+            {
+                "Name": "AOSCOS",
+                "FriendlyName": "AOSC OS",
+                "Default": true,
+                "Amd64Url": {
+                    "Url": "https://releases.aosc.io/os-amd64/wsl/aosc-os_wsl_20250319_amd64.wsl",
+                    "Sha256": "8f507177cf42649e158fdf0da0d1dfb61d57a4b994c1026ec6aa1441eabfb69f"
+                },
+                "Arm64Url": {
+                    "Url": "https://releases.aosc.io/os-arm64/wsl/aosc-os_wsl_20250319_arm64.wsl",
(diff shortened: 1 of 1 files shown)


KexyBiscuit:
Since `oobe.command` and `shortcut.icon` are architecture-independent data, we put them under `/usr/share/wsl` instead of `/usr/lib/wsl`.

Not sure the reason that `validate-modern.py` is complaining about the regular file (`S_IFREG`, `0100000`) mask value...

KexyBiscuit:
Gentle ping, is there anything I missed?

KexyBiscuit:
Loop @OneBlue how can we get this proceeded?

craigloewen-msft:
Hi @KexyBiscuit ! Thank you for the submission. For distros that are available in `wsl --install` we generally focus only on larger scale distros who meet criteria like [the membership criteria for operating system distribution security contact lists](https://oss-security.openwall.org/wiki/mailing-lists/distros#membership-criteria) for security purposes.

However, modern distros are ones that don't require the store, and so can be downloaded and installed from any location! So for users to get access to AOSC OS, using the Store is one great way, or for modern distros you can have them download the .wsl file and double click it, or you could show a script like: `curl https://releases.aosc.io/os-amd64/wsl/aosc-os_wsl_20250319_amd64.wsl -o aosc-os_wsl.wsl; wsl --install --from-file ./aosc-os_wsl.wsl` so users can install from the command line. 

I'm going to close this pull request out for our own bookkeeping purposes, but will still be on the thread and so would be happy to hear any feedback if you'd wish to keep commenting. Thank you again!

Issue #12731, opened 2025-03-23
the kali linux distro require create user many times


usersever:
### Windows Version

10.0.26100.3476

### WSL Version

2.4.12.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Distro Version

kali-linux (i got from microsoft store)

### Other Software

Kernel version: 5.15.167.4-1
WSLg version: 1.0.65
MSRDC version: 1.2.5716
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release

### Repro Steps

```
Installing, this may take a few minutes...
Please create a default UNIX user account. The username does not need to match your Windows username.
For more information visit: https://aka.ms/wslusers
Enter new UNIX username: aqua
New password:
Retype new password:
passwd: password updated successfully
Installation successful!
Waiting for systemd to start...
running
Please create a default Kali WSL user. The username does not need to match your Windows username.
For more information visit: https://aka.ms/wslusers
Enter new UNIX username:
```
than i try to reenter that user in that linux it say 
```
fatal: The user `aqua' already exists.
```
or trying to create other user
```
fatal: The UID 1000 is already in use.
```
even when i try to reinstall kali linux it give me the same issue.

### Expected Behavior

idk

### Actual Behavior

still require create new user

usersever:
[WslLogs-2025-03-23_14-38-46.zip](https://github.com/user-attachments/files/19407644/WslLogs-2025-03-23_14-38-46.zip)

cyberofficial:
Yea this happens to me as well, so I thought the windows store stuff was broken, so manually downloaded the application, extracted the app package then ran the install manually, same issue.

[WslLogs-2025-03-23_13-25-28.zip](https://github.com/user-attachments/files/19410466/WslLogs-2025-03-23_13-25-28.zip)

https://github.com/user-attachments/assets/62456512-c252-4fc1-a148-bbd55e916e35

cyberofficial:
I found a work around

`wsl --exec /bin/bash`

If it's not default then add -d kali-linux before the exec `wsl -d kali-linux --exec /bin/bash`

![Image](https://github.com/user-attachments/assets/e92a050e-0d89-41d5-a4ca-28f6eae79a1c)

![Image](https://github.com/user-attachments/assets/22610c88-4f87-4200-8a2c-dd4e2cc6411b)

Little annoying, but can save it as shortcut on desktop

[...]

PR #12738, opened 2025-03-25
Change checksums to match current Kali image

Gamb1ts:
Includes fix for store downloads OOBE user creation loop


Gamb1ts:
Forgot to link to the issue in the commit message, so adding a link here: https://github.com/microsoft/WSL/issues/12731

OneBlue:
Thank you for the quick update @Gamb1ts ! 

As a small nit, next time if possible it would be better to create the new tars under a different URL, so that way kali installs don't break between the time the tars are updated and the new hashes are merged to the manifest

OneBlue (approved):

Gamb1ts:
Good call, thank you!

PR #12838, opened 2025-04-18
Fix indentation in DistributionInfo.json

Antiz96:
:wave: 

Replace tabs by spaces.  
It seems like I accidentally introduced some tabs with the Arch Linux image, sorry 🫣 🤭

### DIFF
--- a/distributions/validate.py
+++ b/distributions/validate.py
@@ -3,6 +3,7 @@ import json
 import sys
 import hashlib
 import base64
+import difflib
 from urllib.request import urlretrieve
 from xml.etree import ElementTree
 import tempfile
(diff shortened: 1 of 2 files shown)


Antiz96:
I added a quick check to the `validate.py` script which should prevent this type of mistake and ensure a constant / coherent formatting for the future by verifying that a 4 space indentation is used across the whole DistributionInfo.json file (which is the formatting currently applied).  

If it isn't desirable, I can remove it.

Antiz96:

distributions/validate.py
```diff
@@ -65,7 +66,16 @@ def is_unique(collection: list):
...
+        data = fd.read()
+        content = json.loads(data)
+        diff = difflib.unified_diff(
+            data.splitlines(keepends=True),
+            (json.dumps(content, indent=4) + "\n").splitlines(keepends=True),
+            fromfile="a" + sys.argv[1],
+            tofile="b" + sys.argv[1],
+        )
+        diff = "".join(diff)
+        assert diff == "", diff
```
OneBlue, line 78: Nice, thank you for doing this !

OneBlue (approved):
No worries at all @Antiz96. I was planning to reorganize this file either way. 

Thank you for making the fix though !

PR #12918, opened 2025-05-19
Update the pipelines to reference the default branch

OneBlue:
This changes updates various .yml files to correctly reference the current default branch, master


.pipelines/wsl-build-nightly-localization.yml
```diff
@@ -1,7 +1,7 @@
 trigger:
   branches:
     include:
-    - main
+    - master
```
Croydon, line 4: Given that both git and GitHub changed the default name for the default branch from master to main, why making this active step backwards?

Loafer19: they are Old Believers

OneBlue: We eventually want to move to `main` as our default branch, but right now existing WSL installations point to the [distribution manifest](https://github.com/microsoft/WSL/blob/master/distributions/DistributionInfo.json) on the main branch (this is used for wsl --instal & wsl --list --online).

craigloewen-msft: We are investigating changing this, but it will require some effort (As we have a public URL with the current branch name in it and so will need to adjust that, etc.)

benhillis (approved):

Coolis1362 (commented):
This is Ok I will Approve This In My Fork!

Coolis1362 (commented):
well never i don't have acess

.pipelines/wsl-build-nightly-localization.yml
erkinalp: @OneBlue why not `default`?

PR #12920, opened 2025-05-19
Added Support For Ubuntu 25.04 WSL amd64 and Arm64

Coolis1362:
Whlie Checking For Any Updates In The Ubuntu 25.04 Releae Page There Was A WSL Image So Yeah I Chaged It!

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -6,11 +6,11 @@
                 "FriendlyName": "Ubuntu",
                 "Default": true,
                 "Amd64Url": {
-                    "Url": "https://releases.ubuntu.com/noble/ubuntu-24.04.2-wsl-amd64.wsl",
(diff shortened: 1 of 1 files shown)


benhillis:
@Coolis1362 - thanks for the submission, but I think we want to leave updating the distros up to the maintainers of those distros (in this case Canonical). We have monthly syncs with Canonical, and I do not think they are quite ready yet to switch their default version of Ubuntu to 25.04. @craigloewen-msft might have more detail on timeline.

Coolis1362:
Ok, btw I'm A JSON and Python Programmer Yeah, And I fork The Repo to Update [https://github.com/microsoft/WSL/blob/master/distributions/DistributionInfo.json](https://github.com/microsoft/WSL/blob/master/distributions/DistributionInfo.json)!

PR #12943, opened 2025-05-20
Test PR triggers from repo forks

benhillis:
DO NOT MERGE


benhillis:
Ok this seems to be working now.

benhillis:
For anybody interested, running into some fun internal policy issues that we're working around. We have some strict policies in place about building from fork pull requests and we're working on a solution.

PR #12982, opened 2025-05-28
Fix incorrect memset call on an std::string

OneBlue:

This pull request corrects an incorrect call to memset, writing zeroes in an std::string instead of simply calling clear()

## PR Checklist

- [X] **Closes:** Link to issue #xxx

This was originally caught by theoparis, see https://github.com/microsoft/WSL/discussions/12927 

## Validation Steps Performed
Ran the tests


src/linux/init/util.h
```diff
@@ -93,7 +93,7 @@ class InteropServer
     InteropServer(InteropServer&& other) noexcept :
         m_InteropSocketPath(std::move(other.m_InteropSocketPath)), m_InteropSocket(std::move(other.m_InteropSocket))
     {
-        memset(&other.m_InteropSocketPath, 0, sizeof(other.m_InteropSocketPath));
+        m_InteropSocketPath.clear();
```
benhillis: Shouldn't this be other.m_InteropSocketPath?

OneBlue: Thank you, it is. 

I had a deeper look and actually the explicit clear() isn't even needed, since that string is moved during initialization

benhillis: agreed

benhillis (approved):

benhillis:
@OneBlue - looks like pipelines aren't getting triggered.

OneBlue:
> @OneBlue - looks like pipelines aren't getting triggered.

Yeah I'm still working on it

OneBlue:
> > There is another similar compiler warning with memset probably.
> > ```
> > src/linux/init/init.cpp:1075:12: warning: first argument in call to 'memset' is a pointer to non-trivially copyable type 'struct _CREATE_PROCESS_PARSED' [-Wnontrivial-memcall]
> > ```
> 
> Yes this is similar, I'll take a look in a separate PR.

Ended up adding it to this PR

benhillis (approved):

PR #13002, opened 2025-05-29
cleanup: use wil::unique_fd in CreateProcessParse

benhillis:
This non-functional change switched to using the RAII file descriptor types to make ownership of file descriptors more explicit and to reduce the risk of resource leaks. This also removes the need for explicit initialize / cleanup methods on the struct (can you tell this file used to be written in C?).

### DIFF
--- a/src/linux/init/init.cpp
+++ b/src/linux/init/init.cpp
@@ -77,7 +77,6 @@ typedef struct _CREATE_PROCESS_PARSED
     wil::unique_fd EventFd;
     wil::unique_fd StdFd[LX_INIT_STD_FD_COUNT];
     wil::unique_fd ServiceFd;
-    wil::unique_fd TtyFd;
 } CREATE_PROCESS_PARSED, *PCREATE_PROCESS_PARSED;
 
 struct sigaction g_SavedSignalActions[_NSIG];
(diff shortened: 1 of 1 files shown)


src/linux/init/init.cpp
```diff
@@ -2823,7 +2671,7 @@ try
 }
 CATCH_LOG();
 
-int SessionLeaderCreateProcess(gsl::span<gsl::byte> Buffer, int MessageFd, int TtyFd, const wsl::linux::WslDistributionConfig& Config)
+int SessionLeaderCreateProcess(gsl::span<gsl::byte> Buffer, int MessageFd, wil::unique_fd&& TtyFd, const wsl::linux::WslDistributionConfig& Config)
```
benhillis: todo: wrap this method in try / catch_return. #Resolved

src/linux/init/init.cpp
```diff
@@ -3169,22 +2986,16 @@ Return Value:
             FATAL_ERROR("Invalid message size {}", Message.size());
         }
 
-        switch (Header->MessageType)
+        if (Header->MessageType == LxInitMessageCreateProcess)
+        {
+            SessionLeaderCreateProcess(Message, MessageFd, std::move(TtyFd), Config);
```
OneBlue: We shouldn't move the tty fd here, otherwise the next `LxInitMessageCreateProcess` call wil receive an empty fd.

Given that the fd is owned `SessionLeaderEntry()`, I think passing down the FD by value should be OK here

src/linux/init/init.cpp
```diff
@@ -2967,17 +2788,13 @@ Return Value:
...
     // N.B. CreateProcess does not return.
     //
 
-    CreateProcess(&CreateProcessParsed, Config);
-
-SessionLeaderCreateProcessEnd:
-    CreateProcessParseFree(&CreateProcessParsed);
-    return Result;
+    CreateProcess(&Parsed, Config);
+    assert(false);
```
OneBlue: We probably want FATAL_ERROR() here in case we ever somehow hit this

src/linux/init/init.cpp
benhillis: True, need to plumb this another way.

OneBlue (approved):

Issue #13006, opened 2025-05-29
WSL version updated to 2.5.7.0, caused docker Desktop to stop 


NihaNallappagari:
### Windows Version

windows 11 enterprise

### WSL Version

2.5.7.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

5.15.153.1-2

### Other Software

wsl -v -l
WSL version: 2.5.7.0
Kernel version: 6.6.87.1-1
WSLg version: 1.0.66
MSRDC version: 1.2.6074
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release
Windows version: 10.0.26100.4202

### Repro Steps

Upgrade wsl to 2.5.7.0 and start docker desktop, docker desktop fails to start

### Expected Behavior

It shouldn't interrupt docker behavior 

### Actual Behavior

Upgrade wsl to 2.5.7.0 and start docker desktop, docker desktop fails to start. Downgraded to WSL version: 2.2.4.0 and everything working as expected.

### Diagnostic Logs
Different errors did popup
Error occured during wsl startup
failed to fetch extensions
running engine: waiting for the Docker API: context deadline exceeded

SirLouen:
Some logs:
[WslLogs-2025-05-31_18-57-54.zip](https://github.com/user-attachments/files/20534774/WslLogs-2025-05-31_18-57-54.zip)

leongj:
Same issue - I downgraded to 2.4.13 and Docker started working

SirLouen:
> Same issue - I downgraded to 2.4.13 and Docker started working I was also having this issue: [docker/for-win#14704](https://github.com/docker/for-win/issues/14704)

Yes, this is what I also did to make it work

``` 
> wsl --version
WSL version: 2.4.13.0
Kernel version: 5.15.167.4-1
WSLg version: 1.0.65
MSRDC version: 1.2.5716
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release
Windows version: 10.0.26100.4061
```

mikeleppane:
Hey! +1 also happens to me. It also affects VS Code when using the WSL extension. This version of WSL is now causing serious issues, making Docker Desktop and VS Code usage almost impossible. 


```powershell
WSL-versio: 2.5.7.0
Ytimen versio: 6.6.87.1-1
[...]

PR #13036, opened 2025-06-03
Make the post import distribution checks less strict

OneBlue:

This change allows distributions to be imported, even if they don't have a /bin/sh, as long as they have at least `/etc`.

## PR Checklist

- [X] **Closes:** Link to issue #13006

This change solves docker failing to register a distribution if it doesn't provide `/bin/sh`. See #13006 for details. 

## Validation Steps Performed

Added test coverage for this scenario

### DIFF
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -2620,7 +2620,7 @@ void PostProcessImportedDistribution(wsl::shared::MessageWriter<LX_MINI_INIT_IMP
 
     Message->ValidDistribution = false;
 
-    for (auto* path : {"/etc", "/bin/sh", "/etc/wsl-distribution.conf"})
+    for (auto* path : {"/etc", "/bin/sh"})
     {
         if (access(path, F_OK) >= 0)
         {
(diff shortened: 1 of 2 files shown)


benhillis (approved):

src/linux/init/main.cpp
```diff
@@ -2618,17 +2618,20 @@ void PostProcessImportedDistribution(wsl::shared::MessageWriter<LX_MINI_INIT_IMP
...
-            LOG_ERROR("Failed to access {} {}", path, errno);
-            Message->ValidDistribution = false;
-            return;
+            Message->ValidDistribution = true;
         }
     }
 
-    Message->ValidDistribution = true;
+    if (!Message->ValidDistribution)
+    {
```
giacomo, line 2632: missing previous implemented `LOG_ERROR("Failed to access {} {}", path, errno);`. ❓

OneBlue: Yes, that's the objective of this change

giacomo: So you dont want to log any error when ValidDistribution is false? Sorry if the question is stupid.

benhillis (approved):

src/linux/init/main.cpp
benhillis: The error will message will come from the side that is processing this message. Since there is only a single way this can fail (if /etc is not present) it's not interesting to log the failure here.

PR #13054, opened 2025-06-04
Fix missing null terminator when calling execv() on modprobe

OneBlue:

This change fixes a bug causing WSL to boot without kernel modules. The root cause of the bug is a missing null terminator in an argv when calling execv() on modprobe. 

## PR Checklist

## Validation Steps Performed
Validated the fix on an ARM64 device.


benhillis (approved):

benhillis:
Nice, good catch @chessturo and @OneBlue

PR #13061, opened 2025-06-05
Fix edge cases around .vhd support

benhillis:
While experimenting with the new --fixed-vhd support I realized we do not correctly handle all cases where a distro might be stored as a .vhd. This resolves those issues and clarifies usage that we support both .vhd and .vhdx.

### DIFF
--- a/src/windows/service/exe/LxssUserSession.cpp
+++ b/src/windows/service/exe/LxssUserSession.cpp
@@ -927,7 +927,7 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
     std::filesystem::path newVhdPath = Location;
     RETURN_HR_IF(E_INVALIDARG, newVhdPath.empty());
 
-    newVhdPath /= LXSS_VM_MODE_VHD_NAME;
+    newVhdPath /= distro.VhdFilePath.filename();
 
     auto impersonate = wil::CoImpersonateClient();
 
@@ -952,7 +952,7 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
 
     // Update the registry location
     registration.Write(Property::BasePath, Location);
-    registration.Write(Property::VhdFileName, LXSS_VM_MODE_VHD_NAME);
+    registration.Write(Property::VhdFileName, newVhdPath.filename().c_str());
 
     revert.release();
 
@@ -1077,8 +1077,26 @@ HRESULT LxssUserSessionImpl::ExportDistribution(_In_opt_ LPCGUID DistroGuid, _In
         {
             if (WI_IsFlagSet(Flags, LXSS_EXPORT_DISTRO_FLAGS_VHD))
             {
+                if (GetFileType(FileHandle) != FILE_TYPE_DISK)
+                {
+                    THROW_HR_WITH_USER_ERROR(E_INVALIDARG, wsl::shared::Localization::MessageExportVhdInvalidArg());
+                }
+
                 const wil::unique_handle userToken = wsl::windows::common::security::GetUserToken(TokenImpersonation);
                 auto runAsUser = wil::impersonate_token(userToken.get());
+
+                // Ensure the target file has the correct file extension.
+                std::wstring exportPath;
+                THROW_IF_FAILED(wil::GetFinalPathNameByHandleW(FileHandle, exportPath));
+
+                const auto sourceFileExtension = configuration.VhdFilePath.extension().native();
+                const auto targetFileExtension = std::filesystem::path(std::move(exportPath)).extension().native();
+                if (!wsl::windows::common::string::IsPathComponentEqual(sourceFileExtension, targetFileExtension))
+                {
+                    THROW_HR_WITH_USER_ERROR(
+                        WSL_E_EXPORT_FAILED, wsl::shared::Localization::MessageRequiresFileExtension(sourceFileExtension.c_str()));
+                }
+
                 const wil::unique_hfile vhdFile(CreateFileW(
                     configuration.VhdFilePath.c_str(), GENERIC_READ, (FILE_SHARE_READ | FILE_SHARE_DELETE), nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr));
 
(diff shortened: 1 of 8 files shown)


benhillis:
Considering adding some tests, need to convert some vhdx to vhd.

OneBlue (approved):
Change looks good. Regarding test coverage, looks like https://learn.microsoft.com/en-us/powershell/module/hyper-v/convert-vhd?view=windowsserver2025-ps can convert from vhdx to vhd

benhillis:
> Change looks good. Regarding test coverage, looks like https://learn.microsoft.com/en-us/powershell/module/hyper-v/convert-vhd?view=windowsserver2025-ps can convert from vhdx to vhd

Done.

src/windows/service/exe/LxssUserSession.cpp
```diff
@@ -1079,6 +1079,19 @@ HRESULT LxssUserSessionImpl::ExportDistribution(_In_opt_ LPCGUID DistroGuid, _In
             {
                 const wil::unique_handle userToken = wsl::windows::common::security::GetUserToken(TokenImpersonation);
                 auto runAsUser = wil::impersonate_token(userToken.get());
+
+                // Ensure the target file has the correct file extension.
+                std::wstring exportPath;
+                THROW_IF_FAILED(wil::GetFinalPathNameByHandleW(FileHandle, exportPath));
```
OneBlue: We should only do that this if `FileHandle` is a file, since the user might be exporting to a pipe here

src/windows/service/exe/LxssUserSession.cpp
```diff
@@ -1077,8 +1077,26 @@ HRESULT LxssUserSessionImpl::ExportDistribution(_In_opt_ LPCGUID DistroGuid, _In
         {
             if (WI_IsFlagSet(Flags, LXSS_EXPORT_DISTRO_FLAGS_VHD))
             {
+                if (GetFileType(FileHandle) != FILE_TYPE_DISK)
+                {
+                    THROW_HR_WITH_USER_ERROR(E_INVALIDARG, wsl::shared::Localization::MessageExportVhdInvalidArg());
```
OneBlue: What's the reason for this ? We could let a user export to stdout. With this could do something like: 

```
vhd --export --vhd <distro> | ssh <machine> wsl --import --vhd <distro>

Comment (OneBlue) [reply src/windows/service/exe/LxssUserSession.cpp]: We could skip the extension check logic if the HANDLE is not a file, that way we can support pipes

File: src/windows/service/exe/LxssUserSession.cpp [hunk 3]
```diff
@@ -1448,6 +1462,30 @@ HRESULT LxssUserSessionImpl::RegisterDistribution(
                 wil::CreateDirectoryDeep(distributionPath.c_str());
             }
 
+            // If importing a vhd, determine if it is a .vhd or .vhdx.
+            std::wstring vhdName{LXSS_VM_MODE_VHD_NAME};
+            if (WI_IsFlagSet(Flags, LXSS_IMPORT_DISTRO_FLAGS_VHD))
+            {
+                if (GetFileType(FileHandle) != FILE_TYPE_DISK)
```
OneBlue: I think we can skip this as well. If the user wants to export to a file, and then reimport from it either, that file's extension doesn't strictly need to be .vhdx

test/windows/UnitTests.cpp
```diff
@@ -6276,5 +6275,74 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
+        VERIFY_ARE_EQUAL(err, L"");
+
+        // Attempt to export to a .vhdx (should fail).
+        std::tie(out, err) = LxsstuLaunchWslAndCaptureOutput(std::format(L"--export {} {} --format vhd", newDistroName, vhdxPath), -1);
+        VERIFY_ARE_EQUAL(
+            out, L"The specified file must have the .vhd file extension.\r\nError code: Wsl/Service/WSL_E_EXPORT_FAILED\r\n");
+        VERIFY_ARE_EQUAL(err, L"");
+
+        // Attempt to import to a non VHD file.
+        auto tempFile = wsl::windows::common::filesystem::TempFile(
```
OneBlue, line 6329: nit: we could use TempFileFlags::DeleteOnClose here and get rid of the scope_exit

test/windows/UnitTests.cpp
```diff
@@ -6276,5 +6275,74 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
+            GENERIC_ALL, 0, CREATE_ALWAYS, wsl::windows::common::filesystem::TempFileFlags::None, L"txt");
+
+        auto deleteFile =
+            wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { LOG_IF_WIN32_BOOL_FALSE(DeleteFile(tempFile.Path.c_str())); });
+
+        tempFile.Handle.reset();
+
+        constexpr auto negativeVariationDistro = L"negative-variation-distro";
+        std::tie(out, err) = LxsstuLaunchWslAndCaptureOutput(
+            std::format(L"--import {} {} {} --vhd", negativeVariationDistro, negativeVariationDistro, tempFile.Path), -1);
```
OneBlue, line 6336: Is this the behavior we want ? If we're not importing in place, as long as the content is a valid VHD, I don't think we should fail if the file extension isn't what we expect

test/windows/UnitTests.cpp
benhillis: I thought so too, but I kept getting an error message - either invalid handle or some kind of sharing violation, I'll try again...

test/windows/UnitTests.cpp
benhillis: Yes, it is. HCS does checks on file extension and will not allow non .vhd / .vhdx files.

test/windows/UnitTests.cpp
benhillis: Ah now I remember how this works. This flag means "set the delete on close flag during handle creation". We don't want that, because we need to close the handle. The TempFile class handles this already - I can just get rid of the scope exit.

test/windows/UnitTests.cpp
OneBlue: Hmm I was going to except for the case of an import in place, we'll copy the VHD under the correct name, but it looks like we actually need to know if it's a VHD, or VHDX (and I'm guessing that HCS cares about the extension ?).

In that case that's probably OK, we can always revisit if that turns out to be an issue

OneBlue (approved):

PR #13078, opened 2025-06-09
Correctly report corrupted disks when mount() fails with EUCLEAN

OneBlue:

This change updates the distribution creating logic to consider `EUCLEAN` as a symptom of a corrupted disk, allowing the correct error message to be displayed

## PR Checklist

- [X] Closes #13074 

### DIFF
--- a/src/windows/service/exe/WslCoreInstance.cpp
+++ b/src/windows/service/exe/WslCoreInstance.cpp
@@ -59,7 +59,8 @@ WslCoreInstance::WslCoreInstance(
 
     if (result.Result != 0)
     {
-        if (result.Result == EINVAL && result.FailureStep == LxInitCreateInstanceStepMountDisk)
+        // N.B. EUCLEAN (117) can be returned if the disk's journal is corrupted.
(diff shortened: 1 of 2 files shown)


src/windows/service/exe/WslCoreInstance.cpp
```diff
@@ -59,7 +59,8 @@ WslCoreInstance::WslCoreInstance(
 
     if (result.Result != 0)
     {
-        if (result.Result == EINVAL && result.FailureStep == LxInitCreateInstanceStepMountDisk)
+        // N.B. EUCLEAN (117) can be returned if the disk's journal is corrupted.
+        if ((result.Result == EINVAL || result.Result == 117) && result.FailureStep == LxInitCreateInstanceStepMountDisk)
```
benhillis, line 63: Why not use EUCLEAN?

benhillis: I guess that's not defined anywhere in Windows headers.

benhillis (approved):

src/windows/service/exe/WslCoreInstance.cpp
OneBlue: Yeah exactly

PR #13094, opened 2025-06-11
Distributions: Add Rocky Linux 9 and 10

lumarel:

This PR will add Rocky Linux as a new distribution to the DistributionInfo.json list. More specifically the releases of 9.6 and 10.0.
The WSL images are hosted on the main CDN of the project and I made sure with releng that the links should not break in the future, even if the content is moved to the vault.

Further these images are built with KIWI-ng with these recipes: [9](https://git.resf.org/sig_core/rocky-kiwi-descriptions/src/branch/r9/rocky-wsl.xml) and [10](https://git.resf.org/sig_core/rocky-kiwi-descriptions/src/branch/r10/rocky-wsl.xml)

And finally I looked at the PR for Fedora and hopefully checked everything that might be necessary, if there is anything I/we might be able to improve I'm sure going to do so! 🙂
(beside updating the image when we build a new version of course in the future)

Cheers, Lukas

## PR Checklist

- [ ] **Closes:** Link to issue
- [ ] **Documentation updated:** If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/wsl/) and link it here

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -160,6 +160,34 @@
                 }
             }
         ],
+        "RockyLinux": [
+            {
+                "Name": "RockyLinux-9",
+                "FriendlyName": "Rocky Linux 9",
+                "Default": false,
+                "Amd64Url": {
+                    "Url": "https://dl.rockylinux.org/pub/rocky/9.6/images/x86_64/Rocky-9-WSL-Base-9.6-20250531.0.x86_64.wsl",
+                    "Sha256": "9ce7566838dbd4166942d8976c328e03ec05370d9f04006ec4327b61bf04131a"
+                },
+                "Arm64Url": {
+                    "Url": "https://dl.rockylinux.org/pub/rocky/9.6/images/aarch64/Rocky-9-WSL-Base-9.6-20250602.0.aarch64.wsl",
+                    "Sha256": "7ff27a7cddd781641b5990d15282f2a9631d5bbfd80afac6c598f10cd7739bfd"
+                }
+            },
+            {
+                "Name": "RockyLinux-10",
+                "FriendlyName": "Rocky Linux 10",
+                "Default": true,
+                "Amd64Url": {
+                    "Url": "https://dl.rockylinux.org/pub/rocky/10.0/images/x86_64/Rocky-10-WSL-Base-10.0-20250611.0.x86_64.wsl",
+                    "Sha256": "3e84270955361269d2abf53c89da98f17d91b55ff837412ef683a39602b590cb"
+                },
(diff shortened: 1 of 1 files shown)


lumarel:

craigloewen-msft (approved):
Tried the distro out and looks great to me. Thanks for doing the work to add this!! :)

lumarel:
Cool!
Thanks for making it possible to add it 😇 (open source is great!)

OneBlue:
Thank you for doing @lumarel. 

This looks good overall, but I want to point a couple things: 

1) `/usr/libexec/wsl/oobe.sh` manually writes the default user to `/etc/wsl.conf`. We don't recommend doing that since this can break `wsl --manage <distro> --set-default-user`, and shouldn't be needed since you already set `defaultUid` to 1000 in `/etc/wsl-distribution.conf`

2) The tars all have the `security.selinux` extended attribute set, which breaks when importing the distribution with WSL1. Given that WSL2 doesn't support selinux, I'd recommend removing those to remain compatible with WSL1 

You can use [validate-modern.py](https://github.com/microsoft/WSL/blob/master/distributions/validate-modern.py) to validate that the extended attributes are supported by WSL1. Sample usage: 

```
pip install -r distributions\requirements.txt
python distributions\validate-modern.py --tar Rocky-9-WSL-Base-9.6-20250531.0.x86_64.wsl
```

Let me know what you'd like to do above those two, you get the final say as the distribution maintainer, but we definitely recommend fixing those before merging, since that will give your users a better experience.

OneBlue (changes requested):
Waiting on maintainers to reply on above feedback

jeremycline:
> Thank you for doing @lumarel.
> 
> This looks good overall, but I want to point a couple things:
> 
> 1. `/usr/libexec/wsl/oobe.sh` manually writes the default user to `/etc/wsl.conf`. We don't recommend doing that since this can break `wsl --manage <distro> --set-default-user`, and shouldn't be needed since you already set `defaultUid` to 1000 in `/etc/wsl-distribution.conf`

For what it's worth this is probably because I did this in Fedora (sorry) - it's now fixed there so if Rocky pulls the latest sources from Fedora Rawhide this particular issue will be addressed.

lumarel:
Took me a long time to get back on this (sorry, work ate my time)
That's a great hint, thanks for that, @jeremycline. (beside I'm really happy to having had your implementation as template)

I will get that sorted out hopefully very soon.

tinlittle:
It's been 70 years meme inserted here.

benhillis:
Closing old pull requests with no activity.

PR #13101, opened 2025-06-12
Fix various issues with systemd user sessions 

OneBlue:

This change offers major systemd compatbility improvements, which should fix various systemd-related issues. 

## PR Checklist

- [ ] **Closes:** Link to issue #13053

Design changes:

- WSL doesn't create /run/user/<uid> anymore, but lets systemd create it on its own (this will solve the double mount issues which can sometimes break user session, as shown in #13053), 

- WSL now actually waits for the user session to be ready before running the user command (and prints a warning if the session fails to start

- The wslg sockets are created with a systemd-unit (this is required because /run/user/<uid> won't exist before the user session is created with this change )

## Validation Steps Performed

- Added test coverage and re-enabled systemd tests
- Updated the test distro to include a proper systemd user session configuration
- Manually validated that wayland and pulse work as expected


test/windows/UnitTests.cpp
```diff
@@ -2234,7 +2267,8 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
         // Keys that are created by the optional component and the service.
         const std::vector<LPCWSTR> inboxKeys{
             L"SOFTWARE\\Classes\\CLSID\\{B2B4A4D1-2754-4140-A2EB-9A76D9D7CDC6}",
-            L"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Explorer\\Desktop\\NameSpace\\{B2B4A4D1-2754-4140-A2EB-9A76D9D7CDC6}",
+            L"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Explorer\\Desktop\\NameSpace\\{B2B4A4D1-2754-4140-A2EB-"
+            L"9A76D9D7CDC6}",
```
benhillis, line 2271: Hmm why is this getting wrapped? Do you have a different version of clang format?

src/linux/init/init.cpp
```diff
@@ -236,6 +242,61 @@ int WslEntryPoint(int Argc, char* Argv[])
...
+        wil::unique_file File{fopen("/etc/wsl.conf", "r")};
+        if (File)
+        {
+            std::vector<ConfigKey> ConfigKeys = {
+                ConfigKey(wsl::linux::c_ConfigAutoMountRoot, automountRoot),
+
+            };
+            ParseConfigFile(ConfigKeys, File.get(), CFG_SKIP_UNKNOWN_VALUES, STRING_TO_WSTRING(CONFIG_FILE));
+            File.reset();
+        }
```
benhillis, line 267: Could this be passed as another argument to avoid having to parse the config file again?

benhillis: Same question for below helper.

benhillis: Or maybe we should combine the two now that you have your fancy command line argument parsing lib :)

src/linux/init/config.cpp
```diff
@@ -576,6 +515,11 @@ Return Value:
 
     wsl::linux::WslDistributionConfig Config{CONFIG_FILE};
 
+    if (getenv(LX_WSL2_SYSTEM_DISTRO_SHARE_ENV) != nullptr)
+    {
+        Config.GuiAppsEnabled = true;
+    }
```
benhillis, line 521: Why is this needed?

test/windows/UnitTests.cpp
OneBlue: Potentially. We use the version that comes with VS now, so it gets updated automatically

src/linux/init/init.cpp
OneBlue: Unfortunately not. This code is hit when the systemd generator runs, so we can't pass arguments to it

src/linux/init/config.cpp
OneBlue: That's because we need `GuiAppsEnabled` to be set before we decide whether to create a systemd user generator or not. The Windows side setting is set in mini_init, which will only set `LX_WSL2_SYSTEM_DISTRO_SHARE_ENV` if gui apps are enabled

benhillis (approved):

benhillis (approved):

benhillis:
Looks good to me, nice job!

justinkb:
this assumes systemctl is available in /usr/bin, which isn't always true, for example with NixOS, seems problematic

PR #13103, opened 2025-06-12
cleanup: add ChildName to UtilCreateChildProcess error message

benhillis:
While doing some code auditing, I noticed that the UtilCreateChildProcess logs an error that could use more context about the operation that failed.

### DIFF
--- a/src/linux/init/config.cpp
+++ b/src/linux/init/config.cpp
@@ -1034,23 +1034,17 @@ try
 
     if (Config.BootCommand.has_value())
     {
-        const int ChildPid =
-            UtilCreateChildProcess("BootCommand", [Command = Config.BootCommand.value(), SavedSignals = g_SavedSignalActions]() {
-                //
-                // Restore default signal dispositions for the child process.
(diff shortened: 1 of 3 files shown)


src/linux/init/config.cpp
```diff
@@ -1034,23 +1034,17 @@ try
...
-            });
+            THROW_LAST_ERROR_IF(UtilSetSignalHandlers(SavedSignals, false) < 0);
+            THROW_LAST_ERROR_IF(UtilRestoreBlockedSignals() < 0);
 
-        if (ChildPid < 0)
-        {
-            LOG_ERROR("fork failed {}", errno);
-        }
+            execl("/bin/sh", "sh", "-c", Command.c_str(), nullptr);
+            LOG_ERROR("execl() failed, {}", errno);
```
OneBlue, line 1046: Probably for a future change, but now that we have warnings support in init, we could just an stderr warning when the boot command fails

OneBlue (approved):

PR #13117, opened 2025-06-16
Localization change from build: 124622348

OneBlue:


benhillis (approved):

OneBlue (commented):
Let's merge the one that will get created once #13118 is merged since it won't have my additional commit. 

That way we can also test that the trigger works correctly

PR #13157, opened 2025-06-20
Look up systemctl from $PATH for user session checks

K900:
This is already done in other instances where init calls systemctl, and we rely on init to find systemctl in PATH on NixOS, as we don't (really) have a /usr.

## PR Checklist

### DIFF
--- a/src/linux/init/config.cpp
+++ b/src/linux/init/config.cpp
@@ -2704,7 +2704,7 @@ try
 
     if (Config.BootInitTimeout > 0)
     {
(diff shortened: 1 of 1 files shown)


benhillis:
This seems reasonable to me.

OneBlue (approved):
This makes sense. We call `systemctl` by $PATH in other places so we should do the same here. 

@K900: Do you know distributions that put `systemctl` outside of `/usr/bin` ?

K900:
NixOS does, and I assume GuixSD too.

PR #13212, opened 2025-07-02
Expose the VM ID inside user distros

chelnak:
Prior to this change it was not possible to find the ID of the VM responsible for providing WSL distro instances.

With this change, the VM ID is now exposed inside user distros via `wslinfo` and as an environment variable.

A benefit of this change is that a listener can be configured host side using the ID allowing connections from inside user distros.

For example, using Go, we can:

```
hvsock.Listen(hvsock.Addr{
		VMID:      guid,
		ServiceID: svcid,
	})
```

Where guid is the ID of the VM.

And from inside a distro, we can establish a vsock connection by using the appropriate port.

Looking forward to discussing this PR a bit more!

## PR Checklist

## Validation Steps Performed
The changes here have been validated by locally testing the change inside a user distro & also by adding and running unit tests.

The vm id can be accessed by:

* Using `wslinfo --vm-id`
* Fetching at the environment variables inside the distro

image

### DIFF
--- a/test/windows/UnitTests.cpp
+++ b/test/windows/UnitTests.cpp
@@ -842,6 +842,18 @@ class UnitTests
                 L"Invalid command line argument: --invalid\nPlease use 'wslinfo --help' to get a list of supported "
                 L"arguments.\n");
         }
+
+        if (LxsstuVmMode())
+        {
+            // Get the VM ID from the distro and validate that it not null.
+            auto [vmId, vmIdErr] = LxsstuLaunchWslAndCaptureOutput(L"env | grep 'WSL2_VM_ID' | awk -F= '{print $2}'");
+            VERIFY_ARE_NOT_EQUAL(vmId, L"");
+
+            // Ensure that the response from wslinfo matches the VM id from the distros environment
+            auto [out, err] = LxsstuLaunchWslAndCaptureOutput(L"wslinfo --vm-id");
+            VERIFY_ARE_EQUAL(out, std::format(L"{}", vmId));
+            VERIFY_ARE_EQUAL(err, L"");
+        }
     }
 
     TEST_METHOD(WslPath)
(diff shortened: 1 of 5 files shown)


chelnak:
Doh... i'll fix that localisation error shortly!

chelnak:
Should be fixed now (double checked with the python script locally!)

benhillis:
I am good with this approach, but I think @oneblue has some thoughts about not using an environment variable that people might end up taking a dependency on (and might not be present in all situations).

OneBlue (approved):
Overall change looks good ! 

Architecture wise I we should not expose `WSL2_VM_ID` to user processes and instead tell users to use `wslinfo` to fetch the VM ID. 

Doing that cleanly would require a bit of a deeper change (since wslg currently depends on it being set inside the system distro), so let's merge this PR, and then we'll do a follow up change so that wslg moves to calling `wslinfo`, and then we can remove the environment variable from both the user and system distro. 

Thank you for your contribution !

benhillis (approved):

test/windows/UnitTests.cpp
```diff
@@ -842,6 +842,17 @@ class UnitTests
                 L"Invalid command line argument: --invalid\nPlease use 'wslinfo --help' to get a list of supported "
                 L"arguments.\n");
         }
+
+        {
+            // Get the VM ID from the distro and validate that it not null.
+            auto [vmId, vmIdErr] = LxsstuLaunchWslAndCaptureOutput(L"env | grep 'WSL2_VM_ID' | awk -F= '{print $2}'");
```
OneBlue, line 849: This test case is failing when the tests run in WSL1 mode. Let's wrap in in ` if (LxsstuVmMode())`  like we do for `--networking-mode` above

OneBlue (changes requested):

test/windows/UnitTests.cpp
chelnak: I'll get that sorted now! Thanks.

chelnak: Done!

chelnak:
@benhillis @OneBlue Thanks both for the reviews.

On the architecture point - I see what you mean and the plan sounds good!

benhillis (approved):

OneBlue (approved):

PR #13278, opened 2025-07-23
Merge User/oneblue/prototype lsw to a feature/wsl-for-apps

OneBlue:

This change merges the current WslForApps prototype to a feature branch 

## PR Checklist


src/linux/inc/lxwil.h
```diff
@@ -373,6 +373,11 @@ class unique_fd
         return fd;
     }
 
+    int* addressof() noexcept
```
benhillis, line 376: Should we override the address-of operator instead?

src/linux/init/LSWInit.cpp
```diff
@@ -0,0 +1,689 @@
...
+    sockaddr_vm SocketAddress{};
+    wil::unique_fd ListenSocket{UtilListenVsockAnyPort(&SocketAddress, 1, false)};
+    THROW_LAST_ERROR_IF(!ListenSocket);
+
+    Channel.SendResultMessage<uint32_t>(SocketAddress.svm_port);
+
+    wil::unique_fd Socket{UtilAcceptVsock(ListenSocket.get(), SocketAddress, SESSION_LEADER_ACCEPT_TIMEOUT_MS)};
+    THROW_LAST_ERROR_IF(!Socket);
+
+    THROW_LAST_ERROR_IF(dup2(Socket.get(), Message.Fd) < 0);
```
benhillis, line 77: This scares me a bit, how do we control the lifetime of this new file descriptor?

src/linux/init/LSWInit.cpp
```diff
@@ -0,0 +1,689 @@
...
+void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const LSW_TTY_RELAY& Message, const gsl::span<gsl::byte>&)
+{
+    THROW_LAST_ERROR_IF(fcntl(Message.TtyMaster, F_SETFL, O_NONBLOCK) < 0);
+
+    pollfd pollDescriptors[2];
+
+    pollDescriptors[0].fd = Message.TtyInput;
+    pollDescriptors[0].events = POLLIN;
+    pollDescriptors[1].fd = Message.TtyMaster;
+    pollDescriptors[1].events = POLLIN;
```
benhillis, line 89: Would there be any benefit in creating two threads to do the relays?

src/linux/init/LSWInit.cpp
```diff
@@ -0,0 +1,689 @@
...
+        {
+            THROW_LAST_ERROR_IF(prctl(PR_SET_CHILD_SUBREAPER, 1) < 0);
+
+            winsize ttySize{};
+            ttySize.ws_col = Message.TtyColumns;
+            ttySize.ws_row = Message.TtyRows;
+
+            wil::unique_fd ttyMaster;
+
+            auto result = forkpty(ttyMaster.addressof(), nullptr, nullptr, nullptr);
```
benhillis: Pass ttySize

src/linux/init/LSWInit.cpp
```diff
@@ -0,0 +1,689 @@
...
+                sigemptyset(&SignalMask);
+                try
+                {
+                    childLogic();
+                }
+                CATCH_LOG();
+                exit(0);
+            }
+
+            Response.PtyMasterFd = ttyMaster.release();
```
benhillis, line 268: >PtyMasterFd [](http://example.com/codeflow?start=21&length=11)

How is the lifetime of this controlled?

src/linux/inc/lxwil.h
OneBlue: Potentially. This felt a bit easier to read (similar to the put()) method that wil has

src/linux/init/LSWInit.cpp
```diff
@@ -0,0 +1,689 @@
...
+}
+
+int LswEntryPoint(int Argc, char* Argv[])
+{
+
+    //
+    // Mount devtmpfs.
+    //
+
+    int Result = UtilMount(nullptr, "/dev", "devtmpfs", 0, nullptr);
```
benhillis, line 484: We might want to think about if we want to mount some virtual filesystems after chrooting.

src/linux/init/LSWInit.cpp
OneBlue: That's done from the service side. We could have an RAII object to represent the fd from the service side, but in most situations the process will either die at the end of the method, or exec()

src/linux/init/LSWInit.cpp
OneBlue: I don't think so. This is tty I/O so I expect very small amounts of data here. Having one thread also makes synchronization much easier

src/linux/init/main.cpp
```diff
@@ -3875,6 +3876,16 @@ int main(int Argc, char* Argv[])
     // Determine which entrypoint should be used.
     //
 
+    if (getenv(LSW_ROOT_INIT_ENV))
```
benhillis, line 3879: This makes me wonder if the LSW init should be in a separate initrd. Somebody could technically set this variable via the kernel command line in WSL.

src/linux/init/LSWInit.cpp
OneBlue: That one is also owned by the service

src/linux/init/LSWInit.cpp
OneBlue: Yeah so far I've going with the assumption that the caller would do it, but we could also have a flag to mount all the basic ones

src/linux/init/util.cpp
```diff
@@ -1634,6 +1634,19 @@ Return Value:
     return 0;
 }
 
+int UtilMountFile(const char* Source, const char* Destination)
+try
+{
+    wil::unique_fd Fd{open(Destination, (O_CREAT | O_WRONLY | O_TRUNC), 0755)};
```
benhillis: Do we want O_TRUNC? Won't that clobber the file that's there if it currently exists?

src/linux/init/main.cpp
OneBlue: We could, although that would also be true for our WSL1/WSL2 init. If the user actively tries to break init, there's plenty of ways to do that for sure. 

My thinking for now is that we should try to avoid adding extra build complexity unless we have to

src/linux/init/util.cpp
OneBlue: Hmm probably not. Although we're mounting over the file so the expectation would be for that file to be empty. We can probably remove O_TRUNC though

benhillis:
Looking like a good start!

src/linux/init/LSWInit.cpp
benhillis: I don't understand what you mean by owned by the service, what about the Linux FD?

src/linux/init/util.cpp
benhillis: IMO bind mounting over an existing file should be non-destructive.

src/linux/init/LSWInit.cpp
OneBlue: The TtyMasterFd is returned to the service, which then passed it again to the linux when creating the relay process. 

On the linux side, we want the fd to remain open between messages because of that. If let's say something went wrong, the service would close the channel which would close the process to exit, closing the fd.

OneBlue: We could potentially model this a bit better, which RAII object on the service side, but probably in a future change

src/windows/lswclient/DllMain.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    DllMain.cpp
+
+Abstract:
+
+    This file the entrypoint for the LSW client library.
```
benhillis, line 11: I'm still not convinced we need the DLL. We already have a COM interface, why not just use that?

benhillis (approved):

PR #13304, opened 2025-07-29
Localization change from build: 127287347

OneBlue:


localization/strings/en-GB/Resources.resw
```diff
@@ -1100,7 +1100,7 @@ Error code: {}</value>
     <value>Failed to query networking mode</value>
   </data>
   <data name="MessageMismatchedKernelModulesError" xml:space="preserve">
-    <value>Customised kernel modules were supplied without specifying a customised kernel. Please see https://aka.ms/wslcustomkernel for more information.</value>
+    <value>Customised kernel modules were supplied without a specifying a customised kernel. Please see https://aka.ms/wslcustomkernel for more information.</value>
```
OneBlue, line 1103: This does not match the US version

benhillis: booo

crramirez: why adding "a" is the right way? I see the right way even on British is without "a", no?

OneBlue:
Closing since translation is incorrect

PR #13318, opened 2025-07-31
Switch `wslinfo --vm-id` to not rely on presence of VM ID environment variable

benhillis:
This change is a follow up from https://github.com/microsoft/WSL/pull/13212. This is preferred because it discourages users of WSL from making assumptions about what environment variables may be present, which is fragile because processes can and do modify their environment.


packages.config
```diff
@@ -24,7 +24,7 @@
   <package id="Microsoft.WSL.LxUtil.amd64fre" version="10.0.26100.1-240331-1435.ge-release" />
   <package id="Microsoft.WSL.LxUtil.arm64fre" version="10.0.26100.1-240331-1435.ge-release" />
   <package id="Microsoft.WSL.TestDistro" version="2.5.7-47" />
-  <package id="Microsoft.WSLg" version="1.0.66" />
```
benhillis: This version of WSLg cannot be used, it has a gig of coredump files in the ARM64 vhd. We are currently debugging the WSLg ARM64 pipeline.

src/linux/init/config.cpp
```diff
@@ -436,6 +436,18 @@ try
         ResponseChannel.SendResultMessage<uint8_t>(static_cast<uint8_t>(Config.NetworkingMode.value()));
         break;
 
+    case LxInitMessageQueryVmId:
+    {
+        if (Config.VmId.has_value())
+        {
+            wsl::shared::MessageWriter<LX_INIT_QUERY_VM_ID> Response(LxInitMessageQueryVmId);
```
OneBlue: We should send a reply even if `Config.VmId` is unset, otherwise the caller could get stuck waiting on a reply. Sending an empty string would probably be OK (then caller can just throw on empty string)

benhillis: That's not entirely true, the socket connection will be closed and the sender will get an error.

OneBlue: That's true. Probably better to reply to the message than closing though. An unexpectedly closed channel might make us thing that the interop server is crashing if we ever have to debug this path in the future

benhillis:
Still need an updated WSLg to make this work.

src/linux/init/main.cpp
```diff
@@ -1827,6 +1827,11 @@ try
     AddEnvironmentVariable(LX_WSL2_USER_PROFILE, UserProfile);
     AddEnvironmentVariable(LX_WSL2_NETWORKING_MODE_ENV, std::to_string(static_cast<int>(Config.NetworkingMode)).c_str());
 
+    if (UserProfile)
```
OneBlue, line 1830: Is UserProfile only set with the system distro ? Looking through the code it's looking like it's passed to both the system distro and the regular init. Let me know if I'm missing something

OneBlue: We might want to bring in this test case from my change to validate that the env variable isn't passed to user commands: https://github.com/microsoft/WSL/blob/715adf391cb3e83fd88ce1add45f18b67f981e57/test/windows/UnitTests.cpp#L846C8-L856C32

benhillis: Yes user profile is only set for the system distro, and since this logic is temporary, I think this is fine for now. I'll add your test case.

OneBlue (approved):

Issue #13339, opened 2025-08-05
ModernDistro: failed to create Windows Terminal Profile fragment


CarlosNihelton:
### Windows Version

Microsoft Windows [version 10.0.26200.5670]

### WSL Version

2.6.0.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.87.2-1

### Distro Version

Ubuntu 24.04

### Repro Steps

Make sure to have your Windows user named with some special characters, such as `Martín` or `joão`. Register a modern distro.

### Expected Behavior

Terminal profile fragment exist after registration.

### Actual Behavior

Closing and reopening the Windows Terminal doesn't show the expected profile. No fragment was created at  `%LOCALPPADATA%\Microsoft\Windows Terminal\Fragments\Microsoft.WSL\`.

cnihelton:
I investigated the event trace a little bit and found a nice suggestion. I think I know where the exception comes from. I'll push a pull request soon. There is an encoding handling issue coming from the user name having special characters.

<img width="1892" height="877" alt="Image" src="https://github.com/user-attachments/assets/85473b5b-a710-4ec9-b442-169f464fc285" />

OneBlue:
Interesting. Looking at the code, this is most likely an issue in the icon path, which would contain your username. 

Free to submit a pull request ! Here's a [code pointer](https://github.com/microsoft/WSL/blob/8106556cec10c8130ae558c4cee5488cd024d00a/src/windows/service/exe/LxssUserSession.cpp#L2686) to where the issue is most likely coming from

OneBlue:
Fixed in [2.6.1](https://github.com/microsoft/WSL/releases/tag/2.6.1)

PR #13340, opened 2025-08-05
Fix json invalid utf-8 from IconPath

cnihelton:

When storing into the json container the `IconPath` value directly as a `std::string` we are subject to unspecified behaviour as Windows stores paths as 16-bit `wchar_t` , `std::string` doesn't convey a specific char encoding and the `std::filesystem` library doesn't mandate a specific conversion.

We need to access the icon path in its native format (which we assume to be `std::wstring`) and convert it to multibyte.

Details in the linked issue: #13339 

## PR Checklist

- [x] **Closes:** #13339 

## Validation Steps Performed

Repeat the process described in the issue #13339 with a build containing the proposed fix.

### DIFF
--- a/src/windows/service/exe/LxssUserSession.cpp
+++ b/src/windows/service/exe/LxssUserSession.cpp
@@ -2683,7 +2683,7 @@ try
 
         e["name"] = WideToMultiByte(Configuration.Name);
         e["guid"] = WideToMultiByte(distributionProfileId);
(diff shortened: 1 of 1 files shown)


OneBlue:
Thank you for doing this @CarlosNihelton! I reproduced the issue, and validated the fix.

OneBlue (approved):

benhillis (approved):

PR #13345, opened 2025-08-05
Disable Firewall in WSLA VM config

OneBlue:

This change disables firewall support in WSLA for now, since this is only supported on Windows 11 

## PR Checklist

### DIFF
--- a/src/windows/service/exe/LSWVirtualMachine.cpp
+++ b/src/windows/service/exe/LSWVirtualMachine.cpp
@@ -15,6 +15,7 @@ Abstract:
 #include "hcs_schema.h"
 #include "LSWApi.h"
 #include "NatNetworking.h"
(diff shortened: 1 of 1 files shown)


src/windows/service/exe/LSWVirtualMachine.cpp
```diff
@@ -277,6 +277,7 @@ void LSWVirtualMachine::ConfigureNetworking()
 
         // TODO: refactor this to avoid using wsl config
         static wsl::core::Config config(nullptr);
+        config.FirewallConfig.reset(); // Don't enable firewall for now since it's only supported on >= Windows 11
```
benhillis: Can we replace this with a version check?

benhillis (approved):

Issue #13330, opened 2025-08-03
`wsl --debug-shell` hangs when exits


chansey97:
### Windows Version

Microsoft Windows [Version 10.0.26100.4652]

### WSL Version

2.5.9.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.87.2

### Distro Version

Mariner based root file system

### Repro Steps

This is not an urgent issue. I was just reading this document from https://learn.microsoft.com/en-us/windows/wsl/wsl-plugins#plugins-linux-user-space and wanted to inspect the content of the "very minimal Mariner based root file system".

But it still appears to be a BUG, as it cannot exit properly.

1. Open an Command Prompt in Windows Terminal
2. Input `wsl`
   NOTE: This step is necessary, otherwise 
   ```
   The system cannot find the file specified.
   Error code: Wsl/DebugShell/ERROR_FILE_NOT_FOUND
   ```
3. Open an elevated Command Prompt in Windows Terminal
4. Input `wsl --debug-shell`   
   ```   C:\Users\Chansey>wsl --debug-shell   
   Welcome to CBL-Mariner 2.0.20240829 (x86_64) - Kernel 6.6.87.2-microsoft-standard-WSL2 (hvc2)
   DESKTOP-7942QN8 login: root (automatic login)
   ```
5. Input `exit`
   ```
   logout
   <<hangs....>>
   ```

### Expected Behavior

Return to the Command Prompt normally.

### Actual Behavior

Hangs forever... 

Other things you can do:

- Open the 2nd elevated Command Prompt in Windows Terminal
   ```
   C:\Users\Chansey>wsl --debug-shell
   All pipe instances are busy.
   Error code: Wsl/DebugShell/ERROR_PIPE_BUSY
   ```
- The only way to reset everything, is `sc.exe stop wslservice`.

### Diagnostic Logs

[WslLogs-2025-08-03_22-21-44.zip](https://github.com/user-attachments/files/21566329/WslLogs-2025-08-03_22-21-44.zip)

benhillis:
Thanks for reporting, I have an idea on how to fix this.

PR #13379, opened 2025-08-11
Relaunch the debug shell if user exits it.

benhillis:
While looking at #13330 I realized that it's not ideal that the debug shell will become unusable if the user runs "exit" in it. Resolve that issue by adding a relaunch loop.

This does not actually fix the reported issue, but I think is a useful change to take in isolation as it improves debugging capability.

### DIFF
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -1194,9 +1194,27 @@ Return Value:
 --*/
 
 {
-    UtilCreateChildProcess("agetty", []() {
-        execl("/usr/bin/setsid", "/usr/bin/setsid", "/sbin/agetty", "-w", "-L", LX_INIT_HVC_DEBUG_SHELL, "-a", "root", NULL);
-        LOG_ERROR("execl failed, {}", errno);
+    // Spawn a child process to handle relaunching the debug shell if it exits.
+    UtilCreateChildProcess("DebugShell", []() {
+        for (;;)
+        {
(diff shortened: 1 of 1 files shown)


src/linux/init/main.cpp
```diff
@@ -1194,9 +1194,27 @@ Return Value:
 --*/
 
 {
-    UtilCreateChildProcess("agetty", []() {
-        execl("/usr/bin/setsid", "/usr/bin/setsid", "/sbin/agetty", "-w", "-L", LX_INIT_HVC_DEBUG_SHELL, "-a", "root", NULL);
-        LOG_ERROR("execl failed, {}", errno);
+    // Spawn a child process to handle relaunching the debug shell if it exits.
+    UtilCreateChildProcess("DebugShell", []() {
```
OneBlue, line 1198: nit: this could be a thread now that it forks before execing agetty

OneBlue (approved):

src/linux/init/main.cpp
benhillis: I recall there being an issue when I tried that originally, let me double check.

benhillis: Now I recall the issue. This won't work because there is another thread in the process calling waitpid for all children.

OneBlue: Ah I see. Thank you !

PR #13393, opened 2025-08-14
Implement WSLA API to mount and unmount Windows folders via plan9

OneBlue:

This change introduces WLSA APIs to mount and unmount Windows folders in the vm via plan9

## PR Checklist

### DIFF
--- a/src/windows/lswclient/LSWApi.h
+++ b/src/windows/lswclient/LSWApi.h
@@ -205,6 +205,10 @@ HRESULT WslInstallComponents(enum WslInstallComponent Components, WslInstallCall
 // Used for testing until the package is published.
 HRESULT WslSetPackageUrl(LPCWSTR Url);
 
+HRESULT WslMountWindowsFolder(LSWVirtualMachineHandle VirtualMachine, LPCWSTR WindowsPath, const char* LinuxPath, BOOL ReadOnly);
+
+HRESULT WslUnmountWindowsFolder(LSWVirtualMachineHandle VirtualMachine, const char* LinuxPath);
+
 #ifdef __cplusplus
 }
 #endif
--- a/test/windows/LSWTests.cpp
+++ b/test/windows/LSWTests.cpp
@@ -887,6 +887,7 @@ class LSWTests
             VERIFY_ARE_EQUAL(WslMountWindowsFolder(vm.get(), L"relative-path", "/win-path", true), E_INVALIDARG);
             VERIFY_ARE_EQUAL(WslMountWindowsFolder(vm.get(), L"C:\\does-not-exist", "/win-path", true), HRESULT_FROM_WIN32(ERROR_PATH_NOT_FOUND));
             VERIFY_ARE_EQUAL(WslUnmountWindowsFolder(vm.get(), "/not-mounted"), HRESULT_FROM_WIN32(ERROR_NOT_FOUND));
+            VERIFY_ARE_EQUAL(WslUnmountWindowsFolder(vm.get(), "/proc"), HRESULT_FROM_WIN32(ERROR_NOT_FOUND));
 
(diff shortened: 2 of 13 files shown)


src/windows/lswclient/LSWApi.h
```diff
@@ -205,6 +205,10 @@ HRESULT WslInstallComponents(enum WslInstallComponent Components, WslInstallCall
 // Used for testing until the package is published.
 HRESULT WslSetPackageUrl(LPCWSTR Url);
 
+HRESULT WslMountWindowsFolder(LSWVirtualMachineHandle VirtualMachine, LPCWSTR WindowsPath, const char* LinuxPath, BOOL ReadOnly);
+
+HRESULT WslUnmountWindowsFolder(LSWVirtualMachineHandle VirtualMachine, const char* LinuxPath);
```
benhillis, line 210: It's a little odd to take the Linux path for an api to mount a Windows folder.

test/windows/LSWTests.cpp
```diff
@@ -814,4 +814,79 @@ class LSWTests
...
+            VERIFY_SUCCEEDED(WslMountWindowsFolder(vm.get(), testFolder.c_str(), "/win-path", true));
+            expectMount("/win-path", "/win-path*9p*rw,relatime,aname=*,cache=5,access=client,msize=65536,trans=fd,rfd=*,wfd=*");
+
+            // Validate that folder is not writeable from linux
+            VERIFY_ARE_EQUAL(RunCommand(vm.get(), {"/bin/bash", "-c", "echo -n content > /win-path/file.txt"}), 1);
+
+            VERIFY_SUCCEEDED(WslUnmountWindowsFolder(vm.get(), "/win-path"));
+            expectMount("/win-path", {});
+        }
+
```
benhillis, line 884: A test where we call WslUnmountWindowsFolder after something else in the guest has already unmounted would be interesting.

src/windows/lswclient/LSWApi.h
OneBlue: I'm not sure I'm following, this the API that unmounts the folder. Given that the same Windows path can be mounted in multiple places in the guest, the linux path is the only way to uniquely identify the mount point

test/windows/LSWTests.cpp
OneBlue: Good call. Added a bit of logic to correctly handle that case

src/windows/lswclient/LSWApi.h
benhillis: I mean it's odd for a method called "UnmountWindowsFolder" to take "LinuxPath" as an argument, but I see your point.

test/windows/LSWTests.cpp
```diff
@@ -887,6 +887,13 @@ class LSWTests
             VERIFY_ARE_EQUAL(WslMountWindowsFolder(vm.get(), L"relative-path", "/win-path", true), E_INVALIDARG);
             VERIFY_ARE_EQUAL(WslMountWindowsFolder(vm.get(), L"C:\\does-not-exist", "/win-path", true), HRESULT_FROM_WIN32(ERROR_PATH_NOT_FOUND));
             VERIFY_ARE_EQUAL(WslUnmountWindowsFolder(vm.get(), "/not-mounted"), HRESULT_FROM_WIN32(ERROR_NOT_FOUND));
+
```
benhillis, line 891: Hard to tell, but do you have a testcase that tries to umount something that wasn't mounted by the mount api, for example /proc?

OneBlue: That would take the same code path as "/not-mounted" here since we look at `m_plan9Mounts` before trying to unmount, but added a test case to be safe

benhillis (approved):

benhillis (approved):

Issue #11206, opened 2024-02-24
Add NixOS to WSL Linux distros


onny:
Hey,
would like to see NixOS provided as a distribution for WSL. 
It is already working fine by using https://github.com/nix-community/NixOS-WSL

If it's possible to publish it on Microsoft Store, I'm willing to help out.

Best regards
Jonas

sirredbeard:
Commented on the upstream issue, but I'm happy to help.

Ifaen:
so seems like there is no progress right?

PR #13401, opened 2025-08-16
Add NixOS 25.05 to the distribution list

nzbr:

Add NixOS as a modern distribution to `DistributionInfo.json`
The tarball/.wsl-file is built from the sources available here: https://github.com/nix-community/NixOS-WSL/tree/release-25.05

## PR Checklist

- [x] **Closes:** Link to issue #11206

If there's anything that needs to be changed before NixOS can be added to the catalog, let me know. I will then create a new release and update this PR accordingly

## Validation Steps Performed

Several people use NixOS as their main distro for WSL without significant issues regularly. We have automated tests to prevent bugs we had in the past from being re-introduced.

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -185,6 +185,17 @@
                     "Sha256": "a5a2ceb8ca56b7245b909d021b0fd620427db349f02b8ef3b82b741bcb5611cd"
                 }
             }
+        ],
+        "NixOS": [
+            {
+                "Name": "NixOS-25.05",
+                "FriendlyName": "NixOS 25.05",
+                "Default": true,
+                "Amd64Url": {
+                    "Url": "https://github.com/nix-community/NixOS-WSL/releases/download/2505.7.0/nixos.wsl",
+                    "Sha256": "b9db52dc217e7852f715a4693d604da91591158b2da884d67a637cb70d63b1e1"
+                }
+            }
         ]
     },
     "Default": "Ubuntu",


craigloewen-msft:
Is this an official build / release from NixOS itself? 

I noticed that the build release came from the [`nix-community` org](https://github.com/nix-community) not the [`nixos` org](https://github.com/NixOS). 

Why this matters is if this is a community project that isn't a directly sanctioned release from NixOS, then we should look at either making it official and directly approved, or marking it appropriately as a community project.

OneBlue:
Thank you for doing this @nzbr. The distribution content looks good, just two minor things: 

- I'm not very familiar with Nixos, but it's looking like `/etc/wsl.conf` is not writable, is that by design? Is there a specific way for user to make changes to that configuration file? 

- I see `user.default=nixos` set in `/etc/wsl.conf`. Instead of doing this, we recommend setting `oobe.defaultUid` in `/etc/wsl-distribution.conf`. In your case: 

```
[oobe]
defaultUid=1000
```

nzbr:
@craigloewen-msft It is a community project. Marking it as such explicitly would be fine by me - how do you recommend doing that? I could for example add something like `(Community)` or `(unofficial)` to the friendly name

@OneBlue NixOS is managed through a declarative configuration (usually in `/etc/nixos/configuration.nix`) written in the Nix langugage. All files produced by that configuration are stored in `/nix/store` (which is write protected) and symlinked to the correct places when applying the configuration. Everything related to WSL is handled by the modules from our repo which are imported by the configuration; the options are documented here: https://nix-community.github.io/NixOS-WSL/options.html#wslwslconf
The `user.default` setting is also defined by the config. We definitely needed it before there was official systemd support (because NixOS can't really run without it) but I'll take a look if it would be possible to remove it now

Also, thanks for fixing the warnings in the validation script!

craigloewen-msft:
@nzbr I would recommend naming it something like "NixOS (Community)" in the friendly name or NixOS-WSL would be fine too!

The other aspect here is we would need to make sure that this distro meets [the membership criteria for the security mailing list](https://oss-security.openwall.org/wiki/mailing-lists/distros#membership-criteria). This is a requirement for security purposes in case we need to put out early warnings, coordinate fixes, etc. I don't see NixOS on that list, is it realistic for this distro (and the maintainers) to meet the membership criteria and join that mailing list?

But luckily there are more options to distribute your WSL distro than just this list! You can also distribute your `.wsl` file directly on your web page as a download link that users can double click to install, or you could give command line instructions like `curl <urlTo.wsl> -o <distro.wsl>; wsl --install --from-file <distro.wsl>` to install. Given your project has a GitHub repo what might be easiest (and what I'd recommend) is just adding it as a 'release' under the GitHub project.

SuperSandro2000:
> is it realistic for this distro (and the maintainers) to meet the membership criteria and join that mailing list?

We do not have the infrastructure to prepare security fixes in advance in private which was I think the reason why we are not on it. We are going to recheck it we meet the patch delay requirement by the end of the month probably.

craigloewen-msft:
@nzbr  and @SuperSandro2000  I'm going to close this issue out for now while you're investigating (For our bookkeeping purposes). Please feel free to re-file it in the future! Thank you.

PR #13518, opened 2025-09-25
wsla: Implement writeable overlayfs support for GPU mounts

OneBlue:

This change adds a `Flags` argument to `WslMountGpuLibraries()` , allowing a writeable overlayfs to be mounted on top of the GPU mountpoint.

Note: This change makes the assumption that there is no usage of the previous `WslMountGpuLibraries()` API, since the COM interface change would break existing consumers. 

## PR Checklist


benhillis (approved):
:shipit:

PR #13537, opened 2025-09-29
Introduce a new kernel command line argument to collect hvsocket event logs during boot

OneBlue:

This will help root cause issues like #13497

## PR Checklist

### DIFF
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -3918,7 +3918,7 @@ void EnableDebugMode(const std::string& Mode)
             std::string line;
             while (std::getline(file, line))
             {
-                LOG_INFO("{}: {}", path, line);
+                LOG_INFO("{}", line);
             }
 
             LOG_ERROR("{}: closed", path);
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -4081,6 +4081,8 @@ int main(int Argc, char* Argv[])
     {
         LOG_ERROR("Running in debug mode: '{}'", debugMode);
         EnableDebugMode(debugMode);
+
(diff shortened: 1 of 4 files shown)


src/linux/init/main.cpp
```diff
@@ -3875,6 +3875,63 @@ Return Value:
...
+        }
+
+        // Relay logs to the host.
+        std::thread relayThread{[]() {
+            constexpr auto path = "/sys/kernel/debug/tracing/trace_pipe";
+            std::ifstream file(path);
+
+            if (!file)
+            {
+                LOG_ERROR("Failed to open {}, {}", path, errno);
```
benhillis, line 3914: What is the purpose of logging the entire sysfs path?

benhillis: just seems really long, maybe just a hardcoded string like "TRACE: {}" would be better?

OneBlue: Yeah that's probably not necessary. I'll just remove the prefix entirely

benhillis (approved):
LGTM!

src/linux/init/main.cpp
```diff
@@ -3875,6 +3875,63 @@ Return Value:
 
 int WslEntryPoint(int Argc, char* Argv[]);
 
+void EnableDebugMode(const std::string& Mode)
+{
+    if (Mode == "hvsocket")
```
benhillis, line 3880: We might want to think about how to handle multiple debug modes, maybe comma separated? This is fine for now though since there is only one.

OneBlue: Agreed. Probably something we'll do when we add the next one

Issue #13328, opened 2025-08-03
WSL2 CentOS 7 Failed to get D-Bus connection: Operation not permitted


gregbown:
**systemd fails to start in CentOS 7 image** after transferring image to new computer despite all the correct and identical configuration
[boot]
systemctl=true

The initial error was for this check
```
systemctl list-units --type service
Failed to get D-Bus connection: No such file or directory
```

Both original & new system are Windows 11Pro Version 24H2
**Exact same distro image, however; the failing case has WSL version: 2.5.9.0 & the working/original one has WSL version: 2.2.4.0**
Systemd still works perfectly with the same CentOS 7 image on the original computer
On the new system, I have a Rocky Linux 8 image with systemd working fine, but the imported CentOS 7 image, systemd does not work, neither does it work on a variety of CentOS images. . . see "Some additional testing" below

I configured my WSL2 for local debug output using below in my .wslconfig
[wsl2]
debugConsole=true
The logs showed the error
```
Failed to get D-Bus connection: Operation not permitted
```

Attached are logs collected using the recommended script
[WslNetworkingLogs-2025-08-02_22-41-59.zip](https://github.com/user-attachments/files/21564339/WslNetworkingLogs-2025-08-02_22-41-59.zip)

This seems similar to a closed issue https://github.com/microsoft/WSL/issues/13040

**Some additional testing**
I pulled/imported 2 other images from CentOS official docker, a CentOS 7.1, & CentOS 7,9, tried enabling systemd on them and got the exact same error after configuration
```
systemctl list-units --type service
Failed to get D-Bus connection: No such file or directory
```
I pulled/imported another Rocky Linux 8 from official Rocky docker, configured that for systemd and it worked perfectly
So something is weirdly broken ONLY for CentOS systemd, and ONLY on that computer OR the latest version of WSL ???

**Verified this is ONLY an ISSUE in WSL 2.5.9.0 & 2.6.0, + maybe a few more versions back**. 
I installed the same Centos7.9 on my old system that is running WSL 2.2.4.0 and . . .
systemd worked immediately after updating /etc/wsl.conf
```
[boot]
systemctl=true
```
**WSL version 2.6.0 does not fix this**
```
wsl --update --pre-release
```
After updating to the pre-release version 2.6.0, the error is now:
```
Failed to start the systemd user session for 'root'. See journalctl for more details
```
Here are the logs for startup on WSL version 2.6.0

[...]

PR #13546, opened 2025-09-30
Introduce a new wsl.conf config value to allow distributions to opt-in to cgroupv1 mounts

OneBlue:
…

This change introduces a new `/etc/wsl.conf` config value to allow a specific distribution to opt-in to unified cgroup mount instead of just cgroupv2:

```
[automount]
cgroups=v1|v2

```

While this config value only has two options, it's designed as an enum to allow for other values in the future. 
This will help solve issues like #13328, where systemd worked on earlier versions of WSL (before unified cgroup support was dropped). 

## PR Checklist


janderssonse:
What are the real life purposes here. Running old Linux distributions? Just curious.

craigloewen-msft (approved):

benhillis (approved):

craigloewen-msft (approved):

PR #13552, opened 2025-10-02
Implement clean instance terminate when using systemd

benhillis:
This change ensures that `systemctl poweroff` gets time to do its work before terminating the VM.

- [x] **Closes:** Link to issue #13435

### DIFF
--- a/src/linux/init/init.cpp
+++ b/src/linux/init/init.cpp
@@ -129,6 +129,8 @@ void InitEntryUtilityVm(wsl::linux::WslDistributionConfig& Config);
 
 void InitTerminateInstance(gsl::span<gsl::byte> Buffer, wsl::shared::SocketChannel& Channel, wsl::linux::WslDistributionConfig& Config);
 
+void InitTerminateInstanceInternal(wsl::linux::WslDistributionConfig& Config);
+
 void InstallSystemdUnit(const char* Path, const std::string& Name, const char* Content);
 
 int GenerateSystemdUnits(int Argc, char** Argv);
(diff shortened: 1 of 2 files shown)


src/linux/init/init.cpp
```diff
@@ -2661,10 +2652,56 @@ try
     }
 
     //
-    // Respond to the instance termination request.
+    // Attempt to stop the plan9 server, if it is not able to be stopped because of an
+    // in-use file, reply to the service that the instance could not be terminated.
     //
 
-    Channel.SendResultMessage<bool>(StopPlan9Server(Message->Force, Config));
+    if (!StopPlan9Server(Message->Force, Config))
```
OneBlue, line 2659: If the new contract is to close the channel on success, we should update the logic in WslCoreInstance|LxssInstance::RequestStop() not to log an error if the channel is closed. (ReceiveMessageOrClosed()) is probably the easiest way to do that

src/linux/init/init.cpp
```diff
@@ -2661,10 +2652,56 @@ try
...
+    {
+        Channel.SendResultMessage<bool>(false);
+        return;
+    }
+
```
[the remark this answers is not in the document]
benhillis: This is true, maybe it would be cleaner to actually receive the message and then try to read another message (which will fail). @OneBlue - what do you think.

src/linux/init/init.cpp
benhillis: Even better, I'll do that.

src/linux/init/init.cpp
```diff
@@ -2661,10 +2652,56 @@ try
...
+    InitTerminateInstanceInternal(Config);
+}
+CATCH_LOG();
+
+void InitTerminateInstanceInternal(wsl::linux::WslDistributionConfig& Config)
```
OneBlue: nit: `Config` could be const here

src/linux/init/init.cpp
```diff
@@ -2661,10 +2652,56 @@ try
...
+    None.
+
+--*/
+try
+{
+    if (Config.BootInit && !Config.BootStartWriteSocket)
+    {
+        THROW_LAST_ERROR_IF(UtilSetSignalHandlers(g_SavedSignalActions, false) < 0);
+
+        UtilExecCommandLine("systemctl poweroff", nullptr);
```
OneBlue: Looks like this changed from `poweroff` to `reboot` but I guess they both do the same thing ?

OneBlue: Also we might want to immediately shutdown if `systemctl poweroff` fails

OneBlue (approved):

OneBlue (approved):

PR #13553, opened 2025-10-02
Add defaultVhdType configuration setting to .wslconfig

magliocchetti:
## Summary

This PR implements a new `defaultVhdType` configuration setting for .wslconfig and improves how VHD size and type defaults are applied when creating new WSL 2 distributions. This addresses the feature request in issue #13529.

## Problem Statement

Currently, when creating a WSL 2 distribution with a fixed-size VHD, users must always specify both `--vhd-size` and `--fixed-vhd`, even if they have already configured a `defaultVhdSize` in their .wslconfig file:

```bash
wsl --install Ubuntu-24.04 --vhd-size 10GB --fixed-vhd
```

Additionally, there was no way to configure the default VHD type (dynamic vs. fixed) in .wslconfig, requiring users to always specify `--fixed-vhd` explicitly for every distribution they wanted to create with a fixed VHD.

## Solution

This PR adds:

1. **New `defaultVhdType` configuration setting** - Users can now specify their preferred VHD type in .wslconfig:
   ```ini
   [wsl2]
   defaultVhdSize=10GB
   defaultVhdType=fixed
   ```

2. **Improved command-line logic** - The `--vhd-size` parameter is now optional when `defaultVhdSize` is configured in .wslconfig, and the VHD type is determined by:
   - Explicit `--fixed-vhd` flag (highest priority)
   - `defaultVhdType` configuration setting (if no explicit flag)
   - Dynamic (backward compatible default)

3. **Simplified installation commands**:
   ```bash
   # With .wslconfig configured, both of these are now valid:
   wsl --install Ubuntu-24.04 --fixed-vhd  # Uses defaultVhdSize from config
   wsl --install Ubuntu-24.04               # Uses both defaults from config
   ```

### Configuration System
- Added `VhdType` enum with `Dynamic` and `Fixed` values to `WslCoreConfig.h`
- Added `VhdDefaultType` field to the `Config` struct (defaults to `Dynamic`)
- Added parsing for `wsl2.defaultVhdType` in config file parser
- Updated telemetry to track `VhdDefaultType` usage

### Command-Line Handling
- Removed strict validation requiring `--vhd-size` with `--fixed-vhd`
- Added comment explaining that the service layer will validate and apply defaults

### VHD Creation Logic
- Updated `LxssUserSession::RegisterDistribution` to apply `defaultVhdSize` when not explicitly provided
- Modified VHD type determination to consider both explicit flag and config default
- Maintains full backward compatibility

### API Interfaces
- Extended `WslCoreConfigInterface.h` with `VhdType` config entry and `VhdTypeConfiguration` enum
- Implemented Get/Set handlers in `WslCoreConfigInterface.cpp`
- Updated C# bindings in `LibWsl.cs` with renumbered enum values and new `VhdType` enum

### Testing
- Added comprehensive test `DefaultVhdTypeConfiguration` that validates:
  - Installation with `--fixed-vhd` uses `defaultVhdSize` from config
  - `defaultVhdType=fixed` creates Fixed VHDs
  - `defaultVhdType=dynamic` creates Dynamic VHDs  
  - Proper VHD type verification via PowerShell `Get-VHD`

### Documentation
- Updated `wsl-config.md` with new `defaultVhdType` setting documentation
- Included usage examples and value descriptions

## Testing

The implementation has been tested with:
- New unit test validating VHD type configuration behavior
- Backward compatibility verified (existing behavior unchanged when config not set)
- Both dynamic and fixed VHD creation tested
- Configuration file parsing validated

## Backward Compatibility

✅ This PR maintains full backward compatibility:
- Existing behavior unchanged when `defaultVhdType` is not configured (defaults to `dynamic`)
- Explicit command-line flags continue to work as before
- `--vhd-size` can still be specified explicitly and takes precedence

## Related Issue

Fixes #13529

### Configuration in `%USERPROFILE%\.wslconfig`:
```ini
[wsl2]
defaultVhdSize=10GB
defaultVhdType=fixed
```

### Installation commands:
```bash
# Uses both defaults from config (10GB fixed VHD)
wsl --install Ubuntu-24.04

# Explicit override of size but uses type from config
wsl --install Ubuntu-24.04 --vhd-size 20GB

# Explicit override forcing dynamic type
wsl --install Ubuntu-24.04 --vhd-size 20GB  # Omitting --fixed-vhd uses dynamic

# Using --fixed-vhd with default size from config
wsl --install Ubuntu-24.04 --fixed-vhd
```

## Checklist

- [x] Implementation complete
- [x] Unit tests added and passing
- [ ] Documentation updated
- [x] Backward compatibility maintained
- [x] Telemetry added
- [x] Configuration interfaces updated
- [x] C# bindings updated

### DIFF
--- a/src/windows/inc/WslCoreConfigInterface.h
+++ b/src/windows/inc/WslCoreConfigInterface.h
@@ -26,6 +26,7 @@ enum WslConfigEntry
     SwapSizeBytes,
     SwapFilePath,
     VhdSizeBytes,
+    VhdType,
     Networking,
     FirewallEnabled,
     IgnoredPorts,
@@ -65,6 +66,12 @@ enum MemoryReclaimConfiguration
     DropCache = 2
 };
 
+enum VhdTypeConfiguration
+{
+    Dynamic = 0,
+    Fixed = 1
+};
+
 typedef struct WslConfig* WslConfig_t;
 
 struct WslConfigSetting
@@ -78,6 +85,7 @@ struct WslConfigSetting
         bool BoolValue;
         enum NetworkingConfiguration NetworkingConfigurationValue;
         enum MemoryReclaimConfiguration MemoryReclaimModeValue;
+        enum VhdTypeConfiguration VhdTypeValue;
     };
 };
 
--- a/test/windows/UnitTests.cpp
+++ b/test/windows/UnitTests.cpp
@@ -174,7 +174,7 @@ class UnitTests
         WSL2_TEST_ONLY();
 
         // Test that defaultVhdType setting in .wslconfig works correctly
-        const auto testDistroName = L"TestDefaultVhdType";
+        const auto testDistroName = L"test_vhd_type";
         const auto configContent = L"[wsl2]\ndefaultVhdSize=10GB\ndefaultVhdType=fixed\n";
         
         WslConfigChange config(configContent);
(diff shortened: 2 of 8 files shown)


test/windows/UnitTests.cpp
```diff
@@ -169,6 +169,72 @@ class UnitTests
...
+        LxsstuLaunchWsl(std::format(L"--unregister {}", testDistroName));
+        
+        auto cleanup = wil::scope_exit([&] {
+            LxsstuLaunchWsl(std::format(L"--unregister {}", testDistroName));
+        });
+
+        // Install a distribution using --fixed-vhd without --vhd-size
+        // This should use the defaultVhdSize from config
+        auto [out, err] = LxsstuLaunchWslAndCaptureOutput(
+            std::format(L"--install Ubuntu-24.04 --name {} --no-launch --fixed-vhd", testDistroName));
```
OneBlue: We should avoid to depend on networking in the tests, since that can create unstable test results.

Instead of downloading ubuntu, let' use the test distro tar. Something like this should do: 

```
std::format(L"--install --from-file \"{}\" --no-launch --name test_vhd_flag --fixed-vhd", g_testDistroPath));

File: test/windows/UnitTests.cpp [hunk 2]
```diff
@@ -169,6 +169,72 @@ class UnitTests
...
+        WSL2_TEST_ONLY();
+
+        // Test that defaultVhdType setting in .wslconfig works correctly
+        const auto testDistroName = L"TestDefaultVhdType";
+        const auto configContent = L"[wsl2]\ndefaultVhdSize=10GB\ndefaultVhdType=fixed\n";
+        
+        WslConfigChange config(configContent);
+        
+        // Clean up any existing test distro
+        LxsstuLaunchWsl(std::format(L"--unregister {}", testDistroName));
```
OneBlue, line 183: Let's use a different name for the distribution we use for this test so we don't have to unregister the main one

test/windows/UnitTests.cpp
```diff
@@ -169,6 +169,72 @@ class UnitTests
...
+        // Install a distribution using --fixed-vhd without --vhd-size
+        // This should use the defaultVhdSize from config
+        auto [out, err] = LxsstuLaunchWslAndCaptureOutput(
+            std::format(L"--install Ubuntu-24.04 --name {} --no-launch --fixed-vhd", testDistroName));
+
+        // Verify installation succeeded
+        VERIFY_ARE_NOT_EQUAL(out.find(L"completed successfully"), std::wstring::npos);
+
+        // Get the VHD path for the test distribution
+        auto distroList = LxsstuEnumerateDistributions();
```
OneBlue: nit: it might be worth extracting this logic into a lambda. Something like: 

```
auto validateVhdType = [](const char* Expected){...};
```

So the test can just do
validateVhdType("Fixed");

src/windows/inc/WslCoreConfigInterface.h
```diff
@@ -26,6 +26,7 @@ enum WslConfigEntry
     SwapSizeBytes,
```
OneBlue, line 26: wslsettings.exe doesn't actually display this setting, so let's remove the changes in `libwsl` and `wslsettings`

OneBlue (changes requested):
Thank you for doing this ! Added a couple comments

OneBlue:
Closing since we haven't heard back from the author. Feel free to reopen in the future.

PR #13606, opened 2025-10-16
Update Microsoft.WSL.DeviceHost package to version 1.0.0-20251015.1

benhillis:
This new version of the WSL device host package contains support for ICMP packages with the virtioproxy networking mode.


benhillis:
Working on adding some more tests for virtioproxy networking mode, might be part of a different change.

OneBlue (approved):

PR #13623, opened 2025-10-21
wsla: Create wslaservice.exe

OneBlue:

This change creates a new executable: wslaservice.exe, and refactor shared logic between wslservice and wsla service

The change also rewires all the COM logic for the WSLA interfaces to wslaservice.exe and updates the MSI package to install and configure that new service. 

## PR Checklist


msipackage/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ set(OUTPUT_PACKAGE ${BIN}/wsl.msi)
 set(PACKAGE_WIX_IN ${CMAKE_CURRENT_LIST_DIR}/package.wix.in)
 set(PACKAGE_WIX ${BIN}/package.wix)
 set(CAB_CACHE ${BIN}/cab)
-set(BINARIES wsl.exe;wslg.exe;wslhost.exe;wslrelay.exe;wslservice.exe;wslserviceproxystub.dll;init;initrd.img;wslinstall.dll)
+set(BINARIES wsl.exe;wslg.exe;wslhost.exe;wslrelay.exe;wslservice.exe;wslserviceproxystub.dll;init;initrd.img;wslinstall.dll;wslaserviceproxystub.dll;wslaservice.exe)
```
kevpar, line 6: Is this temporary until we introduce a WSLA MSI package?

src/windows/common/wslutil.h
```diff
@@ -44,6 +44,7 @@ inline auto c_msixPackageFamilyName = L"MicrosoftCorporationII.WindowsSubsystemF
 inline auto c_githubUrlOverrideRegistryValue = L"GitHubUrlOverride";
 inline auto c_vhdFileExtension = L".vhd";
 inline auto c_vhdxFileExtension = L".vhdx";
+inline constexpr auto c_vmOwner = L"WSL"; // TODO-WSLA: Does this apply to WSLA ?
```
kevpar, line 47: I think we should include the application name in the VM owner string for WSLA.

src/windows/wslaservice/exe/ServiceMain.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include <ctime>
+
+using namespace wsl::windows::common::registry;
+using namespace wsl::windows::common::string;
+using namespace wsl::windows::common::wslutil;
+using namespace wsl::windows::policies;
+
+wil::unique_event g_networkingReady{wil::EventOptions::ManualReset};
+
+// Declare the LxssUserSession COM class.
```
kevpar: nit: Fix comment

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -0,0 +1,114 @@
+/*++
```
kevpar, line 1: Are these definitions moved from somewhere, or new as of this PR?

src/windows/wslaservice/exe/ServiceMain.cpp
```diff
@@ -0,0 +1,126 @@
...
+        return const_cast<LPWSTR>(L"WSLAService");
+    }
+
+    static void OnSessionChanged(DWORD eventType, DWORD sessionId);
+    HRESULT OnServiceStarting();
+    HRESULT ServiceStarted();
+    void ServiceStopped();
+
+private:
+    wil::unique_couninitialize_call m_coInit{false};
```
ptrivedi, line 57: I think we will need some kind of an Init()

src/windows/wslaservice/exe/ServiceMain.cpp
```diff
@@ -0,0 +1,126 @@
...
+    m_coInit = wil::CoInitializeEx(COINIT_MULTITHREADED);
+
+    return S_OK;
+}
+
+void WslaService::OnSessionChanged(DWORD eventType, DWORD sessionId)
+{
+    if (eventType == WTS_SESSION_LOGOFF)
+    {
+        // TODO-WSLA: Implement for WSLA
```
ptrivedi, line 98: We will need a similar TODOs for other methods in this class as well as in the class itself

src/windows/wslaservice/inc/wslaservice.idl
ptrivedi: Copied rather than moved... from wslservice.idl. Anticipate all this will be going through a bunch of cleanup and refactoring still

ptrivedi (approved):

src/windows/wslaservice/exe/ServiceMain.cpp
OneBlue: What do you mean by an `Init()` ? 

Service initialization is performed by `OnServiceStarting()` and `ServiceStarted()` in this class

src/windows/wslaservice/exe/ServiceMain.cpp
OneBlue: Which methods are you referring to ? All the other methods are implemented

msipackage/CMakeLists.txt
OneBlue: Yes. This change is already pretty big, and once complete it will unblock most of the other work items. 

I'm planning to work on the new MSI package next

src/windows/common/wslutil.h
OneBlue: For sure, but that's not wired yet. The objective of this change is to move the existing implementation to a new service. The plan is to actually implement the new API and logic in followup changes.

msipackage/CMakeLists.txt
OneBlue: Yes exactly

src/windows/wslaservice/exe/ServiceMain.cpp
ptrivedi: Init() or Initialize() or whatever you want to call it, but yes for some initialization stuff. Where would you call EventSourceRegister, or Plan9Redirector initializer etc.

src/windows/wslaservice/exe/ServiceMain.cpp
ptrivedi: Actually nm, they are. I missed that.

src/windows/wslaservice/exe/ServiceMain.cpp
OneBlue: We don't need either of those for WSLA (they only apply to WSL). 

The `initialize()` method in WslService.exe only exists as an easy way to "start" WSL if the GPO policies that controls whether WSL is enabled is changed after the service is started, but this doesn't apply to WSLA

ptrivedi: Why is something like RegisterEventSource() not needed? Because we don't want use Windows Event Log for WSLA?

OneBlue: As far as know, we don't have plans to write in the event log. WSL only uses this to write user warnings to the event log (See `ExecutionContext::EmitUserWarning()`)

ptrivedi: Then maybe we should remove it from wslservice as well?

OneBlue: I'm not following, why would we remove it from wslservice, since wslservice uses it ?

src/windows/wslaservice/inc/wslaservice.idl
kevpar: Why not remove it from the WSL side while we're at it?

OneBlue: Good point. Done in latest iteration.

PR #13638, opened 2025-10-24
Create structure for WSLASession COM class & implementation

OneBlue:

This change introduces a basic structure for the new WSLASession object 

## PR Checklist

### DIFF
--- a/src/windows/wslaservice/exe/WSLASession.cpp
+++ b/src/windows/wslaservice/exe/WSLASession.cpp
@@ -1,0 +1,27 @@
+/*++
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    WSLASession.coo
+
+Abstract:
+
+    This file contains the implementation of the WSLASession COM class.
+
+--*/
+
+#include "WSLASession.h"
+
+using wsl::windows::service::wsla::WSLASession;
+
+WSLASession::WSLASession(const WSLA_SESSION_SETTINGS& Settings) : m_displayName(Settings.DisplayName)
+{
+}
+
+HRESULT WSLASession::GetDisplayName(LPWSTR* DisplayName)
+{
+    *DisplayName = wil::make_unique_string<wil::unique_cotaskmem_string>(m_displayName.c_str()).release();
+    return S_OK;
+}
--- a/src/windows/wslaservice/exe/WSLASession.cpp
+++ b/src/windows/wslaservice/exe/WSLASession.cpp
@@ -4,7 +4,7 @@ Copyright (c) Microsoft. All rights reserved.
 
 Module Name:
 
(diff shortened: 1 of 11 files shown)


src/windows/wslaservice/exe/WSLAUserSession.h
```diff
@@ -28,14 +30,18 @@ class WSLAUserSessionImpl
...
 
 private:
     wil::unique_tokeninfo_ptr<TOKEN_USER> m_tokenInfo;
 
-    std::recursive_mutex m_virtualMachinesLock;
+    std::recursive_mutex m_lock;
     std::vector<WSLAVirtualMachine*> m_virtualMachines;
+
+    // TODO-WSLA: Consider using a weak_ptr to easily destroy when the last client reference is released.
+    std::vector<Microsoft::WRL::ComPtr<WSLASession>> m_sessions;
```
benhillis, line 44: >m_sessions [](http://example.com/codeflow?start=53&length=10)

What happens if we don't keep a reference to the com pointer in the service? My understanding was that the lifetime would be controlled by the client?

benhillis: We don't seem to be using the service version of the ptr anywhere else.

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+    This file contains the implementation of the WSLASession COM class.
+
+--*/
+
+#include "WSLASession.h"
+
+using wsl::windows::service::wsla::WSLASession;
+
+WSLASession::WSLASession(const WSLA_SESSION_SETTINGS& Settings) : m_displayName(Settings.DisplayName)
```
ptrivedi, line 19: I have a member variable for WSLA_SESSION_CONFIGURATION (aka WSLA_SESSION_SETTINGS). Fine for now.

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -0,0 +1,27 @@
+/*++
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    WSLASession.coo
```
ptrivedi:
```suggestion
    WSLASession.cpp
```

ptrivedi (approved):

ptrivedi:
Looks good. Let's submit this and iterate

src/windows/wslaservice/exe/WSLAUserSession.h
OneBlue: Yeah that's something I want to figure out later, this is just a placeholder for now. The lifetime management of the sessions needs more work for sure

PR #13643, opened 2025-10-28
fedora: Add Fedora 43 to the distribution list

jeremycline:

Six months have gone by, so it's time for another Fedora release. In addition to the general Fedora updates this release, the WSL image is now properly suffixed.

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -173,9 +173,22 @@
         ],
         "Fedora": [
             {
+                "Name": "FedoraLinux-43",
+                "FriendlyName": "Fedora Linux 43",
+                "Default": true,
+                "Amd64Url": {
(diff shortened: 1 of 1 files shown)


jeremycline:
Hmm, I thought the images should no longer include SELinux labels, but it seems I forgot to check that actually happened and they apparently still do. I guess I'll investigate that for F44.

OneBlue:
Change LGTM. The SELinux extended attributes aren't a blocker since they were already present on the previous version. 

FYI if you need an easy way to test this in the future you can run: 

```
pip install -r distributions/requirements.txt
python distributions/validate-modern.py --tar </path/to/tar>
```

OneBlue (approved):

craigloewen-msft (approved):

PR #13650, opened 2025-10-29
Draft: Initial rough ContainerManager 

ptrivedi:
## PR Checklist


CMakeLists.txt
```diff
@@ -388,6 +388,7 @@ include_directories(${WSLDEPS_SOURCE_DIR}/include/lxcore)
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src/shared/inc)
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src/windows/inc)
 include_directories(${CMAKE_CURRENT_BINARY_DIR}/src/windows/service/inc/${TARGET_PLATFORM}/${CMAKE_BUILD_TYPE})
+include_directories(${CMAKE_CURRENT_BINARY_DIR}/src/windows/wslaservice/inc/${TARGET_PLATFORM}/${CMAKE_BUILD_TYPE})
```
OneBlue, line 391: This part of the diff is probably a result of being out of sync. Could you merge the latest feature/wsl-for-apps in this branch ? That should make the diff cleaner

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -16,12 +16,12 @@ Module Name:
...
 
-WSLASession::WSLASession(const WSLA_SESSION_SETTINGS& Settings) : m_displayName(Settings.DisplayName)
+HRESULT WSLASession::GetDisplayName(LPWSTR* DisplayName)
 {
+    *DisplayName = wil::make_unique_string<wil::unique_cotaskmem_string>(m_sessionConfig.DisplayName).release();
+    return S_OK;
 }
 
-HRESULT WSLASession::GetDisplayName(LPWSTR* DisplayName)
+WSLASession::WSLASession(const WSLA_SESSION_CONFIGURATION& SessionConfiguration) : m_sessionConfig(SessionConfiguration)
```
OneBlue: FYI this makes a copy of `m_sessionConfig`, but that structure has pointers that are only valid while the COM call is in progress so in the future we'll need to make a deep copy of that structure

src/shared/configfile/configfile.h
```diff
@@ -20,7 +20,7 @@ Parses .gitconfig-style properties files.
 #ifdef WIN32
 
 #include <filesystem>
-#include "wslservice.h"
+#include "wslaservice.h"
```
OneBlue, line 23: This change and other similar ones are probably an artifact of the previous merge. Those should be removed (it's preventing the CI from building)

src/windows/wslaservice/exe/ContainerManager.h
```diff
@@ -0,0 +1,213 @@
...
+#include <vector>
+#include <string>
+#include <filesystem>
+#include <wil\resource.h>
+#include <map>
+#include <mutex>
+
+#include "WSLAVirtualMachine.h"
+
+namespace wsl::windows::service::wsla {
```
benhillis, line 13: In general we like our namespace names to match the file paths, so this would be wsl::windows::wslaservice

src/windows/wslaservice/exe/ContainerManager.h
```diff
@@ -0,0 +1,213 @@
...
+                {
+                    ss << " " << env;
+                } */
+
+                return ss.str();
+            }
+        };
+
+    private:
+        WSLAVirtualMachine* m_pVM;
```
benhillis, line 198: I worry a little bit about the lifetime of the VM class. The session is the owner of it, but also the owner of the container manager. Might be good to think about this relationship. Regardless, a raw c-style pointer is probably not the right data type since there's no guarantee around lifetime.

ptrivedi: Noted. This is just a placeholder for now, will need a revisit

OneBlue:
Closing since this has been merged via different PR

PR #13651, opened 2025-10-29
Implement DNS tunelling support for WSLA

OneBlue:

This change adds DNS tunnelling support to WSLA by implementing the `EnableDnsTunelling` option that the COM api already offers. 

## PR Checklist

### DIFF
--- a/src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
+++ b/src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
@@ -16,6 +16,7 @@ Abstract:
 #include "hcs_schema.h"
 #include "NatNetworking.h"
 #include "WSLAUserSession.h"
+#include "DnsResolver.h"
 
 using namespace wsl::windows::common;
 using helpers::WindowsBuildNumbers;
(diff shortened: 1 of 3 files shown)


src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -324,21 +325,33 @@ void WSLAVirtualMachine::ConfigureNetworking()
     else if (m_settings.NetworkingMode == WslNetworkingModeNAT)
     {
         // Launch GNS
+        // WSLA-TODO: Using fd=4 here seems to hang gns. There's probably a hardcoded file descriptor somewhere that's causing
```
benhillis, line 328: Could it be kmesg?

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -324,21 +325,33 @@ void WSLAVirtualMachine::ConfigureNetworking()
...
+        // so using 1000 for now.
+        std::vector<WSLA_PROCESS_FD> fds(1);
+        fds[0].Fd = 1000;
+        fds[0].Type = WslFdType::WslFdTypeDefault;
 
-        WSLA_PROCESS_FD fd{};
-        fd.Fd = 3;
-        fd.Type = WslFdType::WslFdTypeDefault;
+        std::vector<const char*> cmd{"/gns", LX_INIT_GNS_SOCKET_ARG, "1000"};
```
benhillis, line 334: nit, would be nice to have a single definition of the magic file descriptor numbers so only one place needs to be updated.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -324,21 +325,33 @@ void WSLAVirtualMachine::ConfigureNetworking()
...
+
+        // If DNS tunnelling is enabled, use an additional for its channel.
+        if (m_settings.EnableDnsTunneling)
```
benhillis, line 337: long term we probably don't want this to be a setting, unlike WSL I think we need to have a default config that is rock solid, rather than a bunch of options.

benhillis (approved):
Looks good, minor nits that can be cleaned up later if you want.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
OneBlue: I don't think so, that fd isn't hardcoded

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
OneBlue: Agreed. The current API will need to be rewritten either way so i'm just reusing the settings in the existing one to make testing easier

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
ptrivedi: To add, would it make sense define a starting fd in one place and then incrementally use as needed? Because I see a hardcoded 1001 being used in a couple of places below

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
OneBlue: Ok I figured it out. When we send the WSLA_ACCEPT with fd=4, we actually end up overriding g_logFd, and gns hangs when it writes logs. 

I'll merge this as-is for now, but I'll think of a redesign so the host doesn't need to hardcode fd's in a future change

PR #13655, opened 2025-10-31
wsla: Update the CreateProcess logic to support allocating file descr…

OneBlue:

This change adds support for allocating file descriptors from the guest side, instead of having to hardcode them from the host. 

This is done by using the special `-1` value to tell the guest logic to allocate a file descriptor and return it to the host via a RESULT_MESSAGE. 

## PR Checklist


src/linux/init/util.cpp
```diff
@@ -197,7 +197,7 @@ InteropServer::~InteropServer()
     Reset();
 }
 
-int UtilAcceptVsock(int SocketFd, sockaddr_vm SocketAddress, int Timeout)
+int UtilAcceptVsock(int SocketFd, sockaddr_vm SocketAddress, int Timeout, bool SetCloExec)
```
benhillis: Minor, might want to just pass the socket flags here instead of a bool.

src/linux/init/WSLAInit.cpp
```diff
@@ -85,10 +85,18 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_ACCEPT& M
 
     Channel.SendResultMessage<uint32_t>(SocketAddress.svm_port);
 
-    wil::unique_fd Socket{UtilAcceptVsock(ListenSocket.get(), SocketAddress, SESSION_LEADER_ACCEPT_TIMEOUT_MS)};
+    wil::unique_fd Socket{UtilAcceptVsock(ListenSocket.get(), SocketAddress, SESSION_LEADER_ACCEPT_TIMEOUT_MS, Message.Fd != -1)};
     THROW_LAST_ERROR_IF(!Socket);
 
-    THROW_LAST_ERROR_IF(dup2(Socket.get(), Message.Fd) < 0);
+    if (Message.Fd != -1)
```
benhillis, line 92: Is there any validation around the file descriptor numbers? What would happen if 0, 1, or 2 was specified?

src/linux/init/WSLAInit.cpp
```diff
@@ -85,10 +85,18 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_ACCEPT& M
...
+    {
+        THROW_LAST_ERROR_IF(dup2(Socket.get(), Message.Fd) < 0);
+    }
+    else
+    {
+        Channel.SendResultMessage<int32_t>(Socket.get());
+        Socket.release();
```
benhillis, line 99: Socket is about to leave scop anyway, release not needed.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -587,13 +601,21 @@ std::tuple<int32_t, int32_t, wsl::shared::SocketChannel> WSLAVirtualMachine::For
     return std::make_tuple(pid, ptyMaster, wsl::shared::SocketChannel{std::move(socket), std::to_string(pid), m_vmTerminatingEvent.get()});
 }
 
-wil::unique_socket WSLAVirtualMachine::ConnectSocket(wsl::shared::SocketChannel& Channel, int32_t Fd)
+std::pair<int, wil::unique_socket> WSLAVirtualMachine::ConnectSocket(wsl::shared::SocketChannel& Channel, int32_t Fd)
```
benhillis: minor, i think a simple struct would be better than a pair. I think it would make the code more reable if you could say socket.Fd and socket.Handle versus socket.first and socket.second

benhillis (changes requested):
just a couple tweaks

src/linux/init/WSLAInit.cpp
OneBlue: It is if Fd = 1, otherwise the socket we just connected would be closed when this method returns. This is not needed otherwise because it's dup'd

src/linux/init/WSLAInit.cpp
OneBlue: That would be OK. That's actually the mechanism we use when we create a user process (we connect sockets for all std handles)

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
OneBlue: Agreed, done in latest iteration.

benhillis (approved):

PR #13659, opened 2025-10-31
wsla: Add service API definition and stub methods

OneBlue:

This change adds a .idl definition of the service as well as stub implementation and a basic wiring smoke test.

## PR Checklist


src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -117,9 +204,41 @@ struct _WSLA_SESSION_SETTINGS {
...
+    HRESULT StopContainer([in] LPCSTR Name);
+    HRESULT DeleteContainer([in] LPCSTR Name);
+    HRESULT GetContainerState([in] LPCSTR Name, [out] enum WSLA_CONTAINER_STATE* State);
+
+    // Anonymous host port allocation.
+    HRESULT AllocateHostPort([in] LPCSTR Name, [in] USHORT ContainerPort, [out] USHORT* AllocatedHostPort);
+    HRESULT ReleaseHostPort([in] USHORT HostPort);
+
+    // Process management (nb passing ContainerName = NULL can be used to interact with VM level process for debugging).
+    HRESULT CreateContainerProcess([in, unique] LPCSTR ContainerName, [in] struct WSLA_PROCESS_OPTIONS* Options, [out] struct WSLA_CREATED_PROCESS* Process);
```
OneBlue: I've been doing a bit of thinking on whether we should use a COM object to represent a process and my current thought is that the only advantage in doing this would be to be able to track references to the process so we can kill it if all references are dropped. 

However, I'd say that having that behavior would be counter intuitive for users, since neither Windows nor Linux processes work that way. 

Open to thoughts and feedback on this though

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -117,9 +204,41 @@ struct _WSLA_SESSION_SETTINGS {
...
+
+    // Container management.
+    HRESULT CreateContainer([in] struct WSLA_CONTAINER_OPTIONS* Options);
+    HRESULT StartContainer([in] LPCSTR Name);
+    HRESULT StopContainer([in] LPCSTR Name);
+    HRESULT DeleteContainer([in] LPCSTR Name);
+    HRESULT GetContainerState([in] LPCSTR Name, [out] enum WSLA_CONTAINER_STATE* State);
+
+    // Anonymous host port allocation.
+    HRESULT AllocateHostPort([in] LPCSTR Name, [in] USHORT ContainerPort, [out] USHORT* AllocatedHostPort);
```
OneBlue: For now the API supports both allocating ports as a container starts (via WSLA_CONTAINER_OPTIONS::Ports) and allocating ports via anonymous bind using this method.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -103,12 +110,89 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+struct WSLA_CONTAINER_OPTIONS
+{
+    LPCWSTR Image;
+    LPCWSTR Name;
+    struct WSLA_PROCESS_OPTIONS* InitProcessOptions;
+    [unique, size_is(VolumesCount)] struct WSLA_VOLUME* Volumes;
+    ULONG VolumesCount;
+    [unique, size_is(PortsCount)] struct WSLA_PORT_MAPPING* Ports;
+    ULONG PortsCount;
+    DWORD Flags; // GPU, Privileged, ...
```
ptrivedi, line 169: GPU options will need to have a device list string

OneBlue: Agreed. This is a bit of placeholder for now because we need to some research into what format we should use to describe a GPU

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -103,12 +110,89 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+    ULONGLONG ShmSize;
+};
+
+enum WSLA_CONTAINER_STATE
+{
+    WslaContainerStateInvalid = 0,
+    WslaContainerStateCreated = 1,
+    WslaContainerStateRunning = 2,
+    WslaContainerStateExited = 3,
+    WslaContainerStateFailed = 4
```
ptrivedi: The wslaservice container states could map more closely to nerdctl states. This is an API design discussion and maybe needs a revisit. We could leverage the 'nerdctl ps -a' 's STATUS field or to be more detailed and accurate, we could leverage the 'nerdctl inspect ..' JSON or both to provide more granular and accurate container state information.  SDK can then wrap it to a broader state if necessary. 
For example, we should likely have a 'Deleting' or a 'Removing' state as well -- currently don't have any state mapping to nerdctl's 'Removing' state. We should also discuss if we want to report defunct/dead container via a  state, etc.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -103,12 +110,89 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+};
+
+enum WSLA_PROCESS_STATE
+{
+    WslaProcessStateUknonwn = 0,
```
ptrivedi:
```suggestion
    WslaProcessStateUnknown = 0,
```

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Agreed. This is most likely something that we'll need to update as we implement

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -117,9 +201,41 @@ struct _WSLA_SESSION_SETTINGS {
...
+    HRESULT DeleteContainer([in] LPCSTR Name);
+    HRESULT GetContainerState([in] LPCSTR Name, [out] enum WSLA_CONTAINER_STATE* State);
+
+    // Anonymous host port allocation.
+    HRESULT AllocateHostPort([in] LPCSTR Name, [in] USHORT ContainerPort, [out] USHORT* AllocatedHostPort);
+    HRESULT ReleaseHostPort([in] USHORT HostPort);
+
+    // Process management (nb passing ContainerName = NULL can be used to interact with VM level process for debugging).
+    HRESULT CreateContainerProcess([in, unique] LPCSTR ContainerName, [in] struct WSLA_PROCESS_OPTIONS* Options, [out] struct WSLA_CREATED_PROCESS* Process);
+    HRESULT GetContainerProcessState([in, unique] LPCWSTR ContainerName, [in] int Pid, [out] enum WSLA_PROCESS_STATE* State, [out] int* Code);
```
ptrivedi: nit: WSLA_PROCESS_STATE could be a struct and contain a State enum as well as process exit code

OneBlue: I thought about that, but ended up taking this path because I couldn't find a name for that struct that didn't collide with the enum. I'm open to suggestions though

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -103,12 +110,89 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
     LPCWSTR DisplayName;
     // Details TBD.
-} WSLA_SESSION_SETTINGS;
+};
+
+
+struct WSLA_REGISTRY_AUTHENTICATION_INFORMATION
+{
+    int Dummy; // Dummy value for the .idl file to compile.
+    // Details TBD.
```
ptrivedi:
```suggestion
    // TODO: Details TBD.
```

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -103,12 +110,89 @@ struct _VIRTUAL_MACHINE_SETTINGS {
 } VIRTUAL_MACHINE_SETTINGS;
 
 
-typedef
-struct _WSLA_SESSION_SETTINGS {
+
+struct WSLA_SESSION_SETTINGS {
     LPCWSTR DisplayName;
     // Details TBD.
```
ptrivedi:
```suggestion
    // TODO: Details TBD.
```

src/windows/wslaservice/inc/wslaservice.idl
ptrivedi: Hmm, what if someone leaves the WindowsPort field empty in the WSLA_PORT_MAPPING struct... does it become the anonymous bind situation then? What if we don't provide a special API to support anonymous bind, but simply have a GetHostPort() API?

OneBlue: > Hmm, what if someone leaves the WindowsPort field empty in the WSLA_PORT_MAPPING struct... does it become the anonymous bind situation then?

We could potentially design something like, but in the current API state, the caller would have no way to tell which port was actually allocated (which is the reason for the dedicated method)

> simply have a GetHostPort()

That's tricky because that method needs to "reserve" the port, but if it does, then the port isn't available to WSLA anymore.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -64,6 +60,17 @@ interface ITerminationCallback : IUnknown
...
 };
 
+[
+    uuid(5038842F-53DB-4F30-A6D0-A41B02C94AC1),
+    pointer_default(unique),
+    object
+]
+interface IProgressCallback : IUnknown
+{
+    HRESULT OnTermination(ULONG Progress, ULONG Total);
```
benhillis: What is progress / total for termination?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -103,12 +110,89 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
-struct _WSLA_SESSION_SETTINGS {
+
+struct WSLA_SESSION_SETTINGS {
     LPCWSTR DisplayName;
     // Details TBD.
-} WSLA_SESSION_SETTINGS;
+};
+
+
+struct WSLA_REGISTRY_AUTHENTICATION_INFORMATION
```
benhillis, line 121: What does this do?

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Discussed offline: Anonymous port binding will be a P1, not needed for public preview

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -117,9 +201,41 @@ struct _WSLA_SESSION_SETTINGS {
 ]
 interface IWSLASession : IUnknown
 {
+    // Image management.
+    HRESULT PullImage([in] LPCWSTR Image, [in, unique] struct WSLA_REGISTRY_AUTHENTICATION_INFORMATION* RegistryInformation, [in, unique] IProgressCallback* ProgressCallback);
+    HRESULT ListImages([out, size_is(, *Count)] struct WSLA_IMAGE_INFORMATION** Images, [out] ULONG* Count);
+    HRESULT DeleteImage([in] LPCWSTR Image);
+
+    // Container management.
+    HRESULT CreateContainer([in] struct WSLA_CONTAINER_OPTIONS* Options);
```
benhillis: Have we considered returning a "container" as its own distinct COM object then just having interface methods that do this stuff. This would avoid the need to pass around the right container name and enforcing unique names.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -117,9 +201,41 @@ struct _WSLA_SESSION_SETTINGS {
...
+    HRESULT AllocateHostPort([in] LPCSTR Name, [in] USHORT ContainerPort, [out] USHORT* AllocatedHostPort);
+    HRESULT ReleaseHostPort([in] USHORT HostPort);
+
+    // Process management (nb passing ContainerName = NULL can be used to interact with VM level process for debugging).
+    HRESULT CreateContainerProcess([in, unique] LPCSTR ContainerName, [in] struct WSLA_PROCESS_OPTIONS* Options, [out] struct WSLA_CREATED_PROCESS* Process);
+    HRESULT GetContainerProcessState([in, unique] LPCWSTR ContainerName, [in] int Pid, [out] enum WSLA_PROCESS_STATE* State, [out] int* Code);
+    HRESULT SignalContainerProcess([in, unique] LPCWSTR ContainerName, [in] int Pid, [in] int Signal);
+
+    // Disk management.
+    HRESULT FormatVirtualDisk([in] LPCWSTR Path);
```
benhillis, line 261: Is there a reason to split out create and format?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -102,12 +109,93 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+struct WSLA_PROCESS_OPTIONS
+{
+    LPCSTR Executable;
+    [unique] LPCSTR CurrentDirectory;
+    [size_is(CommandLineCount)] LPCSTR* CommandLine;
+    ULONG CommandLineCount;
+    [unique, size_is(EnvironmentCount)] LPCSTR* Environment;
+    ULONG EnvironmentCount;
+    [unique, size_is(EnvironmentCount)] WSLA_PROCESS_FD *Fds;
+    int FdsCount;
```
kevpar, line 145: What are these used for?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -102,12 +109,93 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+};
+
+struct WSLA_VOLUME
+{
+    LPCSTR WindowsPath;
+    LPCSTR LinuxPath;
```
kevpar: I'd suggest `HostPath` and `ContainerPath` instead here.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -102,12 +109,93 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+};
+
+struct WSLA_PORT_MAPPING
+{
+    USHORT WindowsPort;
```
kevpar: Suggest `HostPort` here

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -116,9 +204,41 @@ struct _WSLA_SESSION_SETTINGS {
 ]
 interface IWSLASession : IUnknown
 {
+    // Image management.
+    HRESULT PullImage([in] LPCWSTR Image, [in, unique] struct WSLA_REGISTRY_AUTHENTICATION_INFORMATION* RegistryInformation, [in, unique] IProgressCallback* ProgressCallback);
+    HRESULT ListImages([out, size_is(, *Count)] struct WSLA_IMAGE_INFORMATION** Images, [out] ULONG* Count);
```
kevpar, line 247: Does the callee allocate the image array? How does it get freed?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -116,9 +204,41 @@ struct _WSLA_SESSION_SETTINGS {
...
+    HRESULT DeleteImage([in] LPCWSTR Image);
+
+    // Container management.
+    HRESULT CreateContainer([in] struct WSLA_CONTAINER_OPTIONS* Options);
+    HRESULT StartContainer([in] LPCSTR Name);
+    HRESULT StopContainer([in] LPCSTR Name);
```
kevpar: I think this should signal and maybe timeout.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -116,9 +204,41 @@ struct _WSLA_SESSION_SETTINGS {
 ]
 interface IWSLASession : IUnknown
 {
+    // Image management.
+    HRESULT PullImage([in] LPCWSTR Image, [in, unique] struct WSLA_REGISTRY_AUTHENTICATION_INFORMATION* RegistryInformation, [in, unique] IProgressCallback* ProgressCallback);
+    HRESULT ListImages([out, size_is(, *Count)] struct WSLA_IMAGE_INFORMATION** Images, [out] ULONG* Count);
+    HRESULT DeleteImage([in] LPCWSTR Image);
+
+    // Container management.
+    HRESULT CreateContainer([in] struct WSLA_CONTAINER_OPTIONS* Options);
```
kevpar: Should we consider returning a container ID, which the client uses to identify the container to other functions, rather than using the container's display name?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -116,9 +204,41 @@ struct _WSLA_SESSION_SETTINGS {
...
+    HRESULT StartContainer([in] LPCSTR Name);
+    HRESULT StopContainer([in] LPCSTR Name);
+    HRESULT DeleteContainer([in] LPCSTR Name);
+    HRESULT GetContainerState([in] LPCSTR Name, [out] enum WSLA_CONTAINER_STATE* State);
+
+    // Anonymous host port allocation (P1).
+    HRESULT AllocateHostPort([in] LPCSTR Name, [in] USHORT ContainerPort, [out] USHORT* AllocatedHostPort);
```
kevpar: Is the intent that this function would first bind an arbitrary host port, and then forward it to the specified container and port?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -102,12 +109,93 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+    WslaProcessStateRunning = 1,
+    WslaProcessStateExited = 2,
+    WslaProcessStateSignalled = 3
+};
+
+struct WSLA_CREATED_PROCESS
+{
+    int Pid;
+    [unique, size_is(FdsCount)] ULONG* Fds;
+    ULONG FdsCount;
```
kevpar: What are these?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -102,12 +109,93 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+    ULONG ExitEvent;
```
kevpar: Is this intended to pass a Windows HANDLE? Is there a better type than ULONG we can use?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -127,6 +247,14 @@ interface IWSLASession : IUnknown
 interface IWSLAUserSession : IUnknown
```
kevpar, line 280: I do think we should rename this to `ISessionManager` or similar. Do you prefer to do that later?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -102,12 +109,93 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+
+enum WSLA_PROCESS_STATE
+{
+    WslaProcessStateUnknown = 0,
+    WslaProcessStateRunning = 1,
+    WslaProcessStateExited = 2,
+    WslaProcessStateSignalled = 3
+};
+
+struct WSLA_CREATED_PROCESS
```
kevpar: The container world generally refers to running a new process in an existing container as an "exec" (e.g. `docker exec` is the command to do so in Docker). I'd suggest we adopt the same terminology here (`struct WSLA_EXEC`, `CreateExec` or `ExecProcess`).

kevpar:
@OneBlue @ptrivedi would like to get your opinions on a possible alternate way of handling container stdio, more similar to how I've seen it done in other systems like HCS and containerd.

Rather than returning the underlying hvsocket handles for stdin/stdout/stderr, the host service would keep these cached internally and relay IO to/from pipes that can be returned to the client.

While this introduces some additional complexity and IO relaying overhead, it has some nice advantages:
- Clients don't need to keep the handles cached, they can drop them and come get them again later if they need them. If we want to support disconnected sessions in the future (client can create containers, exit, and then come back and open its session later) we will need this.
- Enables support for commands like `docker command attach`, which allows you to connect to a running container's stdio. Docker supports this even with multiple active stdio connections to the same container.
- Enables support for sending container output to a log and the client simultaneously.
- Makes debugging with wsladiag easier, since you can attach to a running container.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -102,12 +109,93 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+
+struct WSLA_IMAGE_INFORMATION
+{
+    LPWSTR Name;
+    LPWSTR Hash;
+    ULONGLONG Size;
+    ULONGLONG DownloadTimestamp;
+};
+
+struct WSLA_PROCESS_OPTIONS
```
kevpar, line 136: I think we should create separate options structs for a container and for an exec inside a container. There are some things that will only make sense to pass to a container.

One example is `docker run`'s separation of the `--entrypoint` flag and the `ARGS` that are passed at the end of the command line. Each OCI image has the option to define both an entrypoint and a cmd in its config. By default, if the image is run with no `--entrypoint` or `ARGS` given, the process that's run is the concatenation of the image's entrypoint and cmd. However, the user can specify `--entrypoint` and `ARGS` to override each of these, respectively.

For example, here are some entrypoint and cmd values in popular images:

ubuntu:latest
```json
"Entrypoint": null,
"Cmd": [
	"/bin/bash"
]
```

nginx:latest
```json
"Entrypoint": [
	"/docker-entrypoint.sh"
],
"Cmd": [
	"nginx",
	"-g",
	"daemon off;"
]
```

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -102,12 +109,93 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+    LPCWSTR Name;
+    struct WSLA_PROCESS_OPTIONS* InitProcessOptions;
+    [unique, size_is(VolumesCount)] struct WSLA_VOLUME* Volumes;
+    ULONG VolumesCount;
+    [unique, size_is(PortsCount)] struct WSLA_PORT_MAPPING* Ports;
+    ULONG PortsCount;
+    DWORD Flags; // GPU, Privileged, ...
+    // TODO: List specific GPU devices.
+    ULONGLONG ShmSize;
+};
```
yao-msft, line 173: I assume we can add more options based on customer need? Like the ulimit nofile we discussed.

src/windows/wslaservice/inc/wslaservice.idl
yao-msft: Curious what does this do internally. In initial implementation:
1. Does this just map to nerdctl run? And the later StartContainer just does no-op if it's already running? Or
2. This one just does some sanity check, and the StartContainer does the actual nerdctl run

src/windows/wslaservice/inc/wslaservice.idl
yao-msft: Question: does the public sdk need to use allocate ports explicitly? Or the public sdk would be using container options one always.

src/windows/wslaservice/inc/wslaservice.idl
benhillis: I would lean towards having a COM object I think.

src/windows/wslaservice/inc/wslaservice.idl
ptrivedi: This is for supporting custom/private registry for container images. These usually require authentication, but the details are still TBD because there's some research pending on how the auth works, where certs are configured, etc.

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: This isn't for termination, this is for image pull / import. The design behind this is to allow the app to display a progress bar

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: COM handles the array allocation for the caller, but the caller is responsible for freeing it. 

Here's an example on how this works in a similar WSL API: https://github.com/microsoft/WSL/blob/09c17eb6c6f7b678654bccd59ab7ad060dc2192e/src/windows/common/svccomm.cpp#L592C3-L593C133

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: This is indeed an event handle. There's a deep technical discussion (happy to discuss separately) as to why we sometimes have to use ULONG instead of handles, but for the purpose of the API discussion, this is a HANDLE indeed.

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: We do need to enforce unique names, because we want containers within a session to be able to refer to each other by names. 

I did consider having a dedicated COM object for containers, but when I designed it, it ended up adding an extra layer of object that the caller has to keep track off and required additional methods (like OpenContainer()) so that approach ended up being easier

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: I considered it, but given that names are required to be unique, I found that using the names made the API slightly nicer for the caller

OneBlue: This hasn't been decided yet, but it's most likely going to be one of those two in the first implementation yes

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: "Signal" as in which signal to send the init process ?

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Yes. We're keeping in in the .idl for now, but this will be a P1

OneBlue: In the long term we probably want to support both, the our initial implementation is only going to support explicit port bindings from `WSLA_CONTAINER_OPTIONS`

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: What's the thinking behind the COM object ? Would the process' lifetime be tied to it and so be killed when the COM object is destroyed, or would we need to offer an "OpenProcess" method ?

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: I don't have a strong opinion on this. This is something that @kevpar suggested.

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Agreed. That's a pretty big change, so let's do that later

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: These are the process' file descriptors, which can be handles, or linux files. 

Those are used for interactive and non interactive processes, and can be used to easily copy files from and to a container. For instance you can do `cat` with `stdin=/path/to/file` to easily get a file out of a container

src/windows/wslaservice/inc/wslaservice.idl
kevpar: I don't want to put things in the service side that don't need to be there. VHD creation can be done just fine from the SDK, but we do need the service to provide format functionality, since it has to be done inside a Linux VM.

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: They are the process's file descriptors, same as what we do for `WSLA_CONTAINER_OPTIONS` (see my answer there)

src/windows/wslaservice/inc/wslaservice.idl
kevpar: But isn't ULONG 32 bits, and HANDLE 64 bits?

OneBlue:
> Rather than returning the underlying hvsocket handles for stdin/stdout/stderr, the host service would keep these cached internally and relay IO to/from pipes that can be returned to the client.

I think that would be pretty useful when we'll get to the point of the container executable, but for WSLA I don't think it will be particularly useful to clients, since we don't plan to offer a way to "attach / detach" from a process. 

My thinking on this is that this is something that could implement later (probably via a new flag in WSLA_PROCESS_FD, and new API's to attach & detach from processes), because that would be quite a lot of work to implement today and we don't have a direct use case for this for our current GA plans.

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Yes. The objective of this change is to get an agreement on the overall API structure, there will be future changes to it as we move forward

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Interested. `HANDLE` are void pointers, so technically 64 bits on amd64, but 32 on x86. 

I can't find a definite answer to whether handles can actually use more than 64 bits, but given that `ULongToHandle` and `HandleToULong` are both explicitely defined when `_WIN64` is set, and that no "ULONGTOHANDLE64" or equivalent exists, I'm going to guess that HANDLES are probably expected to be 32 bits. 

For context, WSL uses ULONG <=> HANDLES equivalence in many places.

src/windows/wslaservice/inc/wslaservice.idl
kevpar: Right. For instance [`docker container stop`](https://docs.docker.com/reference/cli/docker/container/stop/) has `--signal` and `--timeout` options.

To get into details a bit, container runtimes will by default send SIGTERM, but this can be changed by specifying `StopSignal` in your [image config](https://github.com/opencontainers/image-spec/blob/main/config.md). The user can also override this by passing `--signal` when stopping the container.

The goal of sending something like SIGTERM is to give the container a chance to exit gracefully. If the container does not exit before the timeout is hit, SIGKILL is sent to hard kill it.

We have some choices in how we should support these options. We could rename StopContainer to SignalContainer, have it return immediately after delivering the signal, and make the SDK responsible for waiting on the event handle until a timeout and then calling SignalContainer again with SIGKILL. Or we could make StopContainer itself do the wait.

src/windows/wslaservice/inc/wslaservice.idl
kevpar: Okay, I assume COM provides a standard free function for the caller to use in this case.

src/windows/wslaservice/inc/wslaservice.idl
kevpar: I started an email thread to understand if casting HANDLE to ULONG is safe to do.

src/windows/wslaservice/inc/wslaservice.idl
kevpar: I think it's just a typo that the method is called `OnTermination`, then?

src/windows/wslaservice/inc/wslaservice.idl
kevpar: We can dig into this separately, but I think the SDK will be responsible for retrieving the right auth info, and just pass it as some sort of string blob to the host service.

kevpar:
> > Rather than returning the underlying hvsocket handles for stdin/stdout/stderr, the host service would keep these cached internally and relay IO to/from pipes that can be returned to the client.
> 
> I think that would be pretty useful when we'll get to the point of the container executable, but for WSLA I don't think it will be particularly useful to clients, since we don't plan to offer a way to "attach / detach" from a process.
> 
> My thinking on this is that this is something that could implement later (probably via a new flag in WSLA_PROCESS_FD, and new API's to attach & detach from processes), because that would be quite a lot of work to implement today and we don't have a direct use case for this for our current GA plans.

True, we probably don't need it for the GA scenarios, though it would make the diag tool a lot more useful. My concern with waiting, though, is that it might require us to change core pieces of the API that relate to IO handling, and I'd rather not have to make breaking changes like that later on.

OneBlue:
> True, we probably don't need it for the GA scenarios, though it would make the diag tool a lot more useful. My concern with waiting, though, is that it might require us to change core pieces of the API that relate to IO handling, and I'd rather not have to make breaking changes like that later on.

I'm hoping that we don't. The way the API is currently designed allows us to easily add enum values & new methods. 

Realistically we'll most likely need to do major API changes to support the standalone container executable, but I'm hoping that by the time we start that work, we'll have a clearer view on the end to end IO flow.

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Interesting. I wasn't aware of that. Maybe we could solve this by adding a flag that specifies whether `WSLA_PROCESS_OPTIONS` in the container options overrides the image's entrypoint, args or both.

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Did another pass and ended up removing that structure and replaced with `IWSLAProcess`

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Added a TODO for this in `WSLA_CONTAINER_OPTIONS`

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Added a TODO to capture this

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -102,12 +109,102 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+    WslaProcessStateExited = 2,
+    WslaProcessStateSignalled = 3
+};
+
+[
+    uuid(1AD163CD-393D-4B33-83A2-8A3F3F23E608),
+    pointer_default(unique),
+    object
+]
+interface IWSLAProcess : IUnknown
```
OneBlue, line 207: I did some more thinking and experimentation, and it turns out that having a process has multiple benefits: 

- It will easily allow us to have multiple people watching the same process
- It will allow us to have multiple implementation of that class in the future, potentially with a different IO relay logic
- It solves the "zombie" issue since now we'll now when callers drop the last reference to a given process

src/windows/wslaservice/inc/wslaservice.idl
ptrivedi: Tying a process' lifetime to an object's lifetime seems a bit strange in this context, mainly because there is also the container to which the process is tied. When the container stops/dies, the process is gone because container runtime manages the cgroups and the lifecycle of the process. It will send a signal to container's pid 1 process upon stopping.

src/windows/wslaservice/inc/wslaservice.idl
kevpar: `CreateContainer` -> `nerdctl create`
`StartContainer` -> `nerdctl start`

kevpar: `nerdctl run` is a shortcut for `nerdctl create` + `nerdctl start`

src/windows/wslaservice/inc/wslaservice.idl
kevpar: I thought the intent was WSLA returns a set of hvsocket handles to the client for stdio? The client passing in handles seems like the opposite of that. Will the host service relay the hvsockets to any handles that are passed here?

src/windows/wslaservice/inc/wslaservice.idl
kevpar: It does seem incongruous to use objects to track processes but not for containers. Should we reconsider using objects for containers as well?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -102,12 +109,102 @@ struct _VIRTUAL_MACHINE_SETTINGS {
...
+    [unique, size_is(EnvironmentCount)] LPCSTR* Environment;
+    ULONG EnvironmentCount;
+    [unique, size_is(EnvironmentCount)] WSLA_PROCESS_FD *Fds;
+    int FdsCount;
+};
+
+struct WSLA_VOLUME
+{
+    LPCSTR HostPath;
+    LPCSTR ContainerPathPath;
```
kevpar: Typo

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -116,9 +213,41 @@ struct _WSLA_SESSION_SETTINGS {
...
+    HRESULT PullImage([in] LPCWSTR Image, [in, unique] struct WSLA_REGISTRY_AUTHENTICATION_INFORMATION* RegistryInformation, [in, unique] IProgressCallback* ProgressCallback);
+    HRESULT ListImages([out, size_is(, *Count)] struct WSLA_IMAGE_INFORMATION** Images, [out] ULONG* Count);
+    HRESULT DeleteImage([in] LPCWSTR Image);
+
+    // Container management.
+    HRESULT CreateContainer([in] struct WSLA_CONTAINER_OPTIONS* Options);
+    HRESULT StartContainer([in] LPCSTR Name);
+    HRESULT StopContainer([in] LPCSTR Name, [in] int Signal, [in] ULONG TimeoutMs);
+    HRESULT DeleteContainer([in] LPCSTR Name);
+    HRESULT GetContainerState([in] LPCSTR Name, [out] enum WSLA_CONTAINER_STATE* State);
```
kevpar: Given this set of container APIs, it's not clear to me how the client would retrieve a container's stdio handles, or wait on the exit event, etc. Am I missing something?

src/windows/wslaservice/inc/wslaservice.idl
kevpar: I take that to mean you don't think we should use "Exec"?

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: This is now done in IWSLAProcess. See GetExitEvent() and GetStdHandle(). We can go over the flow in the API review meeting

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: That's the case. This is an input argument, the caller uses it to describe what kind of stdio handles it wants back from the service

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Updated the new interface to use Exec

src/windows/wslaservice/inc/wslaservice.idl
ptrivedi: @OneBlue for reference, in private preview we have 

    struct ContainerProcessOptions
    {
        std::vector<std::string> Args;
        std::vector<std::string> Envs;
    }

(If we had decided to support it, detached mode could also be here)

benhillis (approved):

benhillis:
Signing off so we can continue to make progress. I have some additional feedback I will follow up about offline.

PR #13664, opened 2025-11-04
Add a temporary flag to wsl.exe to create a WSLA shell

OneBlue:

This change introduces a new command line argument to wsl.exe: --wsla. 

This can be used to open a WSLA shell, which can be used for testing or debugging. Sample usage: 

```
$ wsl --wsla --vhd "C:\Users\<user>\AppData\Local\lxss\ext4.vhdx" --memory 256 --dns-tunelling --cpu 1
```

For convenience, wsl --wsla will use WSL's system distro VHD, so that `wsl --wsla` "just works". 

This will allow us to test & debug while we build the end to end flow. In the future, this logic should be moved to a dedicated executable, and use wsla's runtime VHD by default, once it will be available 

## PR Checklist

### DIFF
--- a/src/windows/common/WslClient.cpp
+++ b/src/windows/common/WslClient.cpp
@@ -1623,11 +1623,19 @@ int WslaShell(_In_ std::wstring_view commandLine)
     std::vector<ULONG> handles(fds.size());
 
     WSLA_CREATE_PROCESS_RESULT result{};
-    THROW_IF_FAILED(virtualMachine->CreateLinuxProcess(&processOptions, static_cast<ULONG>(fds.size()), fds.data(), handles.data(), &result));
+    auto createProcessResult =
+        virtualMachine->CreateLinuxProcess(&processOptions, static_cast<ULONG>(fds.size()), fds.data(), handles.data(), &result);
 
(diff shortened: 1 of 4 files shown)


src/windows/common/WslClient.cpp
```diff
@@ -1539,6 +1541,159 @@ int RunDebugShell()
...
+    THROW_IF_FAILED(virtualMachine->CreateLinuxProcess(&processOptions, static_cast<ULONG>(fds.size()), fds.data(), handles.data(), &result));
+
+    if (result.Errno != 0)
+    {
+        THROW_HR_WITH_USER_ERROR(E_FAIL, std::format(L"Failed to create process {}, errno = {}", shell.c_str(), result.Errno));
+    }
+
+    // Configure console for interactive usage.
+
+    HANDLE Stdout = GetStdHandle(STD_OUTPUT_HANDLE);
```
OneBlue, line 1643: In theory part of this could be factored with wslrelay's logic, but this is a temporary tool so I'd rather keep this simple

ptrivedi (approved):
Will do another pass after approval. Let's merge to unblock.

PR #13677, opened 2025-11-05
WSLAUserSession and WSLASession changes

ptrivedi:
* Add WSLASession creation and tracking in WSLAUserSession
* Wire up WSLASession to contain and create WSLAVirtualMachine

### DIFF
--- a/src/windows/wslaservice/exe/WSLASession.cpp
+++ b/src/windows/wslaservice/exe/WSLASession.cpp
@@ -13,15 +13,27 @@ Abstract:
 --*/
 
 #include "WSLASession.h"
+#include "WSLAUserSession.h"
 
 using wsl::windows::service::wsla::WSLASession;
 
-WSLASession::WSLASession(const WSLA_SESSION_SETTINGS& Settings) : m_displayName(Settings.DisplayName)
+WSLASession::WSLASession(const WSLA_SESSION_SETTINGS& Settings, WSLAUserSessionImpl& userSessionImpl, const VIRTUAL_MACHINE_SETTINGS& VmSettings) :
+    m_sessionSettings(Settings),
+    m_userSession(userSessionImpl),
+    m_virtualMachine(VmSettings, userSessionImpl.GetUserSid(), &userSessionImpl),
+    m_displayName(Settings.DisplayName)
 {
+    m_virtualMachine.Start();
 }
 
 HRESULT WSLASession::GetDisplayName(LPWSTR* DisplayName)
 {
     *DisplayName = wil::make_unique_string<wil::unique_cotaskmem_string>(m_displayName.c_str()).release();
+    return S_OK;
+}
+
+HRESULT WSLASession::GetVirtualMachine(IWSLAVirtualMachine** VirtualMachine)
+{
+    THROW_IF_FAILED(m_virtualMachine.QueryInterface(__uuidof(IWSLAVirtualMachine), (void**)VirtualMachine));
     return S_OK;
 }
--- a/src/windows/wslaservice/exe/WSLASession.h
+++ b/src/windows/wslaservice/exe/WSLASession.h
@@ -15,6 +15,7 @@ Abstract:
 #pragma once
 
 #include "wslaservice.h"
+#include "WSLAVirtualMachine.h"
 
 namespace wsl::windows::service::wsla {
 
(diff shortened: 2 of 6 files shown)


src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -13,15 +13,28 @@ Module Name:
...
+    m_userSession(userSessionImpl),
+    m_virtualMachine(wil::MakeOrThrow<WSLAVirtualMachine>(VmSettings, userSessionImpl.GetUserSid(), &userSessionImpl))
 {
+    m_virtualMachine->Start();
 }
 
 HRESULT WSLASession::GetDisplayName(LPWSTR* DisplayName)
 {
-    *DisplayName = wil::make_unique_string<wil::unique_cotaskmem_string>(m_displayName.c_str()).release();
+    *DisplayName = wil::make_unique_string<wil::unique_cotaskmem_string>(m_sessionSettings.DisplayName).release();
```
OneBlue: That won't work because the pointers in `m_sessionSettings` won't be valid when that method is called. 

We should probably make a copy of DisplayName here

src/windows/wslaservice/exe/WSLASession.h
```diff
@@ -15,18 +15,22 @@ Module Name:
...
-    WSLASession(const WSLA_SESSION_SETTINGS& Settings);
+    WSLASession(const WSLA_SESSION_SETTINGS& Settings, WSLAUserSessionImpl& userSessionImpl, const VIRTUAL_MACHINE_SETTINGS& VmSettings);
     IFACEMETHOD(GetDisplayName)(LPWSTR* DisplayName);
+    IFACEMETHOD(GetVirtualMachine)(IWSLAVirtualMachine** VirtualMachine);
 
 private:
-    std::wstring m_displayName;
+    WSLA_SESSION_SETTINGS m_sessionSettings;
+    WSLAUserSessionImpl& m_userSession;
+    Microsoft::WRL::ComPtr<WSLAVirtualMachine> m_virtualMachine;
```
OneBlue: That might cause some issues in the tests, because we're keeping a COM reference to `m_virtualMachine`, which can prevent the service from stopping.

I'm wondering if we can solve this by just not making a ComPtr, and just using a "regular" instance, and calling `QueryInterface` instead of `CopyTo` in GetVirtualMachine. 

Let me know if that makes sense, i'm worried that if we check this in, the service failing to stop is going to cause a bit of pain

ptrivedi: I forget why we ended up making it a ComPtr in the first place, but makes sense.

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -13,15 +13,27 @@ Module Name:
 --*/
 
 #include "WSLASession.h"
+#include "WSLAUserSession.h"
 
 using wsl::windows::service::wsla::WSLASession;
 
-WSLASession::WSLASession(const WSLA_SESSION_SETTINGS& Settings) : m_displayName(Settings.DisplayName)
+WSLASession::WSLASession(const WSLA_SESSION_SETTINGS& Settings, WSLAUserSessionImpl& userSessionImpl, const VIRTUAL_MACHINE_SETTINGS& VmSettings) :
+    m_sessionSettings(Settings),
```
OneBlue, line 21: nit: maybe for a future change. We probably shouldn't keep that structure as a field since some of its pointers will be invalid when the constructor returns. (and since it's not used in other places)

src/windows/wslaservice/exe/WSLAUserSession.h
```diff
@@ -30,15 +30,18 @@ class WSLAUserSessionImpl
...
+    HRESULT CreateSession(const WSLA_SESSION_SETTINGS* Settings, const VIRTUAL_MACHINE_SETTINGS* VmSettings, IWSLASession** WslaSession);
 
     void OnVmTerminated(WSLAVirtualMachine* machine);
 
 private:
     wil::unique_tokeninfo_ptr<TOKEN_USER> m_tokenInfo;
 
+    std::recursive_mutex m_wslaSessionsLock;
+    // TODO-WSLA: Consider using a weak_ptr to easily destroy when the last client reference is released.
+    std::vector<Microsoft::WRL::ComPtr<WSLASession>> m_wslaSessions;
```
OneBlue, line 42: I'm just realizing now that we also have a COM pointer here as well. I wonder how that didn't also cause a shutdown hang, but if it didn't, we're probably ok to merge as-is, and we can investigate this in a future change

OneBlue (approved):

src/windows/wslaservice/exe/WSLASession.cpp
ptrivedi: We will need to revisit other fields added to WSLA_SESSION_SETTINGS and deep copying etc., once this gets cleaned up more

src/windows/wslaservice/exe/WSLASession.cpp
ptrivedi: I did reply about this in an earlier comment. Having the struct here will need a revisit.

src/windows/wslaservice/exe/WSLAUserSession.h
ptrivedi: There is a TODO here anyway, and yes did not notice the shutdown hang with this being in there.

src/windows/wslaservice/exe/WSLASession.cpp
ptrivedi: Added a TODO around this as well

OneBlue (approved):

PR #13684, opened 2025-11-06
Add a --new-api temporary flag for WSLA debugging

ptrivedi:
* Add a --new-api flag to WslClient tool
* When the flag is specified, the newly wired codepath for WSLASession creation is followed
* Existing mode of WSLAUserSession creating VM will temporarily exist parallely until proper scaffolding cleanup

### DIFF
--- a/src/windows/common/WslClient.cpp
+++ b/src/windows/common/WslClient.cpp
@@ -1542,6 +1542,7 @@ int WslaShell(_In_ std::wstring_view commandLine)
     settings.NetworkingMode = WslNetworkingModeNAT;
     std::string shell = "/bin/bash";
     std::string fsType = "ext4";
+    bool newApi = false;
     bool help = false;
 
     ArgumentParser parser(std::wstring{commandLine}, WSL_BINARY_NAME);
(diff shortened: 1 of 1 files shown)


benhillis (approved):

OneBlue (approved):

src/windows/common/WslClient.cpp
```diff
@@ -1570,7 +1572,18 @@ int WslaShell(_In_ std::wstring_view commandLine)
     wsl::windows::common::security::ConfigureForCOMImpersonation(userSession.get());
 
     wil::com_ptr<IWSLAVirtualMachine> virtualMachine;
-    THROW_IF_FAILED(userSession->CreateVirtualMachine(&settings, &virtualMachine));
+    if (newApi)
+    {
+        WSLA_SESSION_SETTINGS sessionSettings{L"my-display-name"};
+        wil::com_ptr<IWSLASession> session;
```
yao-msft, line 1578: The session will go out of scope after this code block, what would happen to the vm created from the session?

ptrivedi: This is just a placeholder for now. We will hook up the VM termination to Session lifetime. Pierre's upcoming change will do that.

OneBlue: In theory it should, but currently WSLASession is pretty much never deleted because WSLAUserSession keeps a reference to all created sesssions. 

Let's look into fixing this once the new API structure is checked in

ptrivedi: Yes, I had the immediate next task as scaffolding cleanup the right way. But after the API review today, I feel like that's where we are going next anyway. But yes, to specifically answer Yao's question, the Sessins are not deleted

PR #13689, opened 2025-11-06
wslsettings: ensure selected setting is auto-expanded and selected

benhillis:
Implement keyboard focus management for SettingsExpander controls across settings pages. This resolves an accessibility issue reported internally.

### DIFF
--- a/src/windows/wslsettings/Helpers/RuntimeHelper.cs
+++ b/src/windows/wslsettings/Helpers/RuntimeHelper.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (C) Microsoft Corporation. All rights reserved.
 
+using CommunityToolkit.WinUI.Controls;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Input;
 using System.Reflection;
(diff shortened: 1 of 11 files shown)


src/windows/wslsettings/Helpers/RuntimeHelper.cs
```diff
@@ -47,4 +48,36 @@ public static void TryMoveFocusPreviousControl(Button? button)
...
+            });
+        }
+    }
+
+    public static void SetupExpanderFocusManagementByName(Microsoft.UI.Xaml.FrameworkElement parent, string expanderName, string textBoxName)
+    {
+        var expander = parent.FindName(expanderName) as Microsoft.UI.Xaml.FrameworkElement;
+        var textBox = parent.FindName(textBoxName) as Microsoft.UI.Xaml.Controls.Control;
+
+        if (expander != null && textBox != null)
```
OneBlue, line 82: Should we assert that those are non null here ? I'd expect that it should never happen

src/windows/wslsettings/Helpers/RuntimeHelper.cs
```diff
@@ -47,4 +48,36 @@ public static void TryMoveFocusPreviousControl(Button? button)
...
+        if (expander is CommunityToolkit.WinUI.Controls.SettingsExpander settingsExpander)
+        {
+            settingsExpander.RegisterPropertyChangedCallback(CommunityToolkit.WinUI.Controls.SettingsExpander.IsExpandedProperty, (sender, dp) =>
+            {
+                if (sender is CommunityToolkit.WinUI.Controls.SettingsExpander se && se.IsExpanded)
+                {
+                    System.EventHandler<object>? layoutHandler = null;
+                    layoutHandler = (s, e) =>
+                    {
+                        se.LayoutUpdated -= layoutHandler;
```
OneBlue, line 64: What's the reason for de-registering the callback here ?

src/windows/wslsettings/Views/Settings/FileSystemPage.xaml.cs
```diff
@@ -53,4 +56,15 @@ private void DefaultVHDSizeTextBox_TextChanged(object sender, TextChangedEventAr
         TextBox? textBox = sender as TextBox;
         ViewModel.SetDefaultVHDSize_ResetEnabled(textBox!.Text);
     }
+
+    private void OnPageLoaded(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
+    {
+        var expander = this.FindName("DefaultVHDSizeExpander") as SettingsExpander;
+        var textBox = this.FindName("DefaultVHDSizeTextBox") as TextBox;
+
+        if (expander != null && textBox != null)
```
OneBlue, line 69: We should probably assert that those are non null as well

src/windows/wslsettings/Views/Settings/OptionalFeaturesPage.xaml.cs
```diff
@@ -26,6 +26,14 @@ public OptionalFeaturesPage()
                 FrameworkElementAutomationPeer.FromElement(Settings_ErrorTryAgainLater).RaiseAutomationEvent(AutomationEvents.LiveRegionChanged);
             }
         });
+
+        this.Loaded += OnPageLoaded;
+    }
+
+    private void OnPageLoaded(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
+    {
+        RuntimeHelper.SetupExpanderFocusManagementByName(this, "SystemdSettingsExpander", "InitTextBox");
```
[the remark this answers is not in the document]
OneBlue: Indeed I can't find any references to those, are those from a previous iteration of the change maybe ?

src/windows/wslsettings/Helpers/RuntimeHelper.cs
benhillis: I assume so the handler doesn't invoke another callback.

OneBlue: Wouldn't want that though ? For instance if the window is resized ?

benhillis: No, This would repeatedly steal focus from the user during window resizes, animations, etc.

OneBlue (approved):

PR #13694, opened 2025-11-08
Implement WSLAProcess

OneBlue:

This pull requests starts moving WSLA to the new API. This PR focuses on IWSLAProcess and its implementation. 

The tests that could easily be ported to the new API have been updated, but others will need to be ported later when we make more progress on the Session API. 

## PR Checklist


src/linux/init/WSLAInit.cpp
```diff
@@ -636,11 +637,114 @@ void HandleMessage(wsl::shared::SocketChannel& Channel, LX_MESSAGE_TYPE Type, co
     }
 }
 
+void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_WATCH_PROCESSES& Message, const gsl::span<gsl::byte>& Buffer)
+{
+    // Create a signalfd to watch for SIGCHLD
+    sigset_t mask{};
+    sigemptyset(&mask);
+    sigaddset(&mask, SIGCHLD);
+    THROW_LAST_ERROR_IF(UtilSaveBlockedSignals(mask) < 0);
```
benhillis, line 646: Do we need to reset these when the method exits? #Resolved

src/linux/init/WSLAInit.cpp
```diff
@@ -636,11 +637,114 @@ void HandleMessage(wsl::shared::SocketChannel& Channel, LX_MESSAGE_TYPE Type, co
...
+                }
+                else
+                {
+                    LOG_ERROR("Received SIGCHLD for process that was neither signaled nor exited. Pid: {}, Status: {}", result, status)
+                }
+
+                Channel.SendMessage(message);
+            }
+
+            if (!exitedProcess)
```
benhillis, line 723: This is unreachable. #Resolved

OneBlue: It is if we hit the `break` in line 697

src/linux/init/WSLAInit.cpp
OneBlue: We shouldn't have to. This method isn't supposed to exit

benhillis: It does eventually through right?

src/windows/common/relay.cpp
```diff
@@ -923,4 +929,248 @@ try
...
+        {
+            while (m_handles[i]->GetState() == IOHandleStatus::Standby)
+            {
+                m_handles[i]->Schedule();
+            }
+        }
+
+        // Remove completed handles from m_handles.
+        m_handles.erase(
+            std::remove_if(
```
benhillis: std::erase_if #Resolved

src/linux/init/WSLAInit.cpp
OneBlue: It does, but only when the VM terminates and init exits

benhillis: does anything bad happen if this method returns and goes back to whatever thread is processing incoming message requests?

src/windows/common/relay.cpp
```diff
@@ -923,4 +929,248 @@ try
...
+    std::optional<std::chrono::steady_clock::time_point> deadline;
+
+    if (Timeout.has_value())
+    {
+        deadline = std::chrono::steady_clock::now() + Timeout.value();
+    }
+
+    // Run until all handles are completed.
+
+    bool updateHandles = true;
```
benhillis: never used #Resolved

src/windows/common/relay.cpp
```diff
@@ -923,4 +929,248 @@ try
...
+
+    while (!m_handles.empty())
+    {
+        // Schedule IO on each handle until all are either pending, or completed.
+        for (auto i = 0; i < m_handles.size(); i++)
+        {
+            while (m_handles[i]->GetState() == IOHandleStatus::Standby)
```
benhillis, line 955: Are you worried about this starving the other handles? #Resolved

src/windows/common/relay.cpp
```diff
@@ -923,4 +929,248 @@ try
...
+void ReadHandle::Collect()
+{
+    WI_ASSERT(State == IOHandleStatus::Pending);
+
+    // Transition back to standby
+    State = IOHandleStatus::Standby;
+
+    // Complete the read.
+    DWORD bytesRead{};
+    THROW_IF_WIN32_BOOL_FALSE(GetOverlappedResult(Handle.get(), &Overlapped, &bytesRead, false));
```
benhillis: Do we also need to handle ERROR_HANDLE_EOF or ERROR_BROKEN_HANDLE here? #Resolved

benhillis: And transition to completed?

src/windows/common/relay.cpp
OneBlue: Indeed that was from a previous implementation. Thank you

src/windows/common/relay.cpp
OneBlue: Not in this specific because that can only happen if the IO operation returns immediately, if not IO is immediately available, we get ERROR_IO_PENDING, which will allow us to move on the next handle.

src/windows/common/relay.cpp
OneBlue: Good point. Not sure if sockets can do that, but I'll add handling for it

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -314,8 +320,60 @@ void WSLAVirtualMachine::Start()
...
+                    try
+                    {
+                        e->OnTerminated(message->Signaled, message->Code);
+                    }
+                    CATCH_LOG();
+
+                    found = true;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
OneBlue: This is not true, the lock is held during the entire loop

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -640,69 +702,66 @@ void WSLAVirtualMachine::OpenLinuxFile(wsl::shared::SocketChannel& Channel, cons
...
+    _In_ const WSLA_PROCESS_OPTIONS& Options, int* Errno, const TPrepareCommandLine& PrepareCommandLine)
 {
+    auto setErrno = [Errno](int Error) {
+        if (Errno != nullptr)
+        {
+            *Errno = Error;
+        }
+    };
+
+    setErrno(-1);
```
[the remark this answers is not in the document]
OneBlue: This is meant as a catch-all and not supposed to actually be returned

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -101,6 +102,11 @@ WSLAVirtualMachine::~WSLAVirtualMachine()
         }
         CATCH_LOG()
     }
+
+    if (m_processExitThread.joinable())
+    {
+        m_processExitThread.join();
+    }
```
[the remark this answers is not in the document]
OneBlue: m_trackedProcesses are valid until the destructor returns

src/linux/init/WSLAInit.cpp
OneBlue: The only code path where that would happen would be when `Channel` is closed, and so if we went back to `ProcessMessages()` that method would just exit immediately

src/linux/init/WSLAInit.cpp
benhillis: Ok, previously I think we weren't setting exitProcess in the loop, or maybe i missed it.

benhillis (approved):

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -215,6 +157,55 @@ interface IWSLAProcess : IUnknown
...
+    BOOL EnableDnsTunneling;
+    BOOL EnableDebugShell;
+    BOOL EnableEarlyBootDmesg;
+    BOOL EnableGPU;
+} VIRTUAL_MACHINE_SETTINGS;
+
+
+
+struct WSLA_SESSION_SETTINGS {
+    LPCWSTR DisplayName;
```
JohnMcPMS, line 202: Not sure about the preferred process for raising this, but the "display name" is actually the "session key".  Typically one would tag something with "display" when it was used only for user presentation purposes, but this is in fact a unique value used to identify the session.  I would argue for changing the name of this value to `SessionKey` (at least semantically) and only if it makes sense adding a new `DisplayName` (again, semantically).

PR #13705, opened 2025-11-12
Add vmrs collection

kvega005:
This change allows WSLA service to collect a .vmrs file and crash log in the event of a guest kernel panic.

## PR Checklist

### DIFF
--- a/src/windows/common/hcs_schema.h
+++ b/src/windows/common/hcs_schema.h
@@ -22,6 +22,12 @@ Abstract:
         Json[#Value] = (Object).Value.value(); \
     }
 
+#define ASSIGN_IF_PRESENT(Json, Object, Value) \
+    if (Json.contains(#Value)) \
+    { \
+        (Object).Value = Json.at(#Value).get_to((Object).Value); \
+    }
+
 namespace wsl::windows::common::hcs {
 
 enum class ModifyRequestType
@@ -510,10 +516,8 @@ inline void to_json(nlohmann::json& j, const GuestErrorSaveReport& g)
 
 inline void from_json(const nlohmann::json& j, GuestErrorSaveReport& r)
 {
-    if (j.contains("SaveStateFile"))
-        r.SaveStateFile = j.at("SaveStateFile").get<std::wstring>();
-    if (j.contains("Status"))
-        r.Status = j.at("Status").get<int32_t>();
+    ASSIGN_IF_PRESENT(j, r, SaveStateFile);
+    ASSIGN_IF_PRESENT(j, r, Status);
 }
 
 struct CrashReport
@@ -531,10 +535,8 @@ inline void to_json(nlohmann::json& j, const CrashReport& c)
 
 inline void from_json(const nlohmann::json& j, CrashReport& c)
 {
-    if (j.contains("CrashLog"))
-        c.CrashLog = j.at("CrashLog").get<std::wstring>();
-    if (j.contains("GuestCrashSaveInfo"))
-        c.GuestCrashSaveInfo = j.at("GuestCrashSaveInfo").get<GuestErrorSaveReport>();
+    ASSIGN_IF_PRESENT(j, c, CrashLog);
+    ASSIGN_IF_PRESENT(j, c, GuestCrashSaveInfo);
 }
 
 enum class NotificationType
@@ -564,18 +566,12 @@ struct GuestCrashAttribution
 inline void to_json(nlohmann::json& j, const GuestCrashAttribution& g)
 {
     j = nlohmann::json::object();
-    if (g.CrashParameters.has_value())
-    {
-        j["CrashParameters"] = g.CrashParameters.value();
-    }
+    OMIT_IF_EMPTY(j, g, CrashParameters)
 }
 
 inline void from_json(const nlohmann::json& j, GuestCrashAttribution& g)
 {
-    if (j.contains("CrashParameters"))
-    {
-        g.CrashParameters = j.at("CrashParameters").get<std::vector<uint64_t>>();
-    }
+    ASSIGN_IF_PRESENT(j, g, CrashParameters);
 }
 
 // Attribution record (trimmed to GuestCrash only for now)
@@ -587,18 +583,12 @@ struct AttributionRecord
 inline void to_json(nlohmann::json& j, const AttributionRecord& a)
 {
     j = nlohmann::json::object();
-    if (a.GuestCrash.has_value())
-    {
-        j["GuestCrash"] = a.GuestCrash.value();
-    }
+    OMIT_IF_EMPTY(j, a, GuestCrash)
 }
 
 inline void from_json(const nlohmann::json& j, AttributionRecord& a)
 {
-    if (j.contains("GuestCrash"))
-    {
-        a.GuestCrash = j.at("GuestCrash").get<GuestCrashAttribution>();
-    }
+    ASSIGN_IF_PRESENT(j, a, GuestCrash);
 }
 
 struct SystemExitStatus
@@ -618,10 +608,8 @@ inline void to_json(nlohmann::json& j, const SystemExitStatus& s)
 inline void from_json(const nlohmann::json& j, SystemExitStatus& s)
 {
     s.Status = j.at("Status").get<int32_t>();
-    if (j.contains("ExitType"))
-        s.ExitType = j.at("ExitType").get<NotificationType>();
-    if (j.contains("Attribution"))
-        s.Attribution = j.at("Attribution").get<std::vector<AttributionRecord>>();
+    ASSIGN_IF_PRESENT(j, s, ExitType);
+    ASSIGN_IF_PRESENT(j, s, Attribution);
 }
 
 } // namespace wsl::windows::common::hcs
--- a/src/windows/common/hcs_schema.h
+++ b/src/windows/common/hcs_schema.h
@@ -439,6 +439,19 @@ struct Scsi
     NLOHMANN_DEFINE_TYPE_INTRUSIVE_ONLY_SERIALIZE(Scsi, Attachments);
 };
 
+struct DebugOptions
+{
+    std::optional<std::wstring> BugcheckSavedStateFileName;
+    std::optional<std::wstring> ShutdownOrResetSavedStateFileName;
+};
+
+inline void to_json(nlohmann::json& j, const DebugOptions& d)
+{
+    j = nlohmann::json::object();
+    OMIT_IF_EMPTY(j, d, BugcheckSavedStateFileName);
+    OMIT_IF_EMPTY(j, d, ShutdownOrResetSavedStateFileName);
+}
+
 struct Devices
 {
     std::optional<VirtioSerial> VirtioSerial;
@@ -467,8 +480,9 @@ struct VirtualMachine
     Chipset Chipset;
     Topology ComputeTopology;
     Devices Devices;
+    DebugOptions DebugOptions;
 
-    NLOHMANN_DEFINE_TYPE_INTRUSIVE_ONLY_SERIALIZE(VirtualMachine, StopOnReset, Chipset, ComputeTopology, Devices);
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE_ONLY_SERIALIZE(VirtualMachine, StopOnReset, Chipset, ComputeTopology, Devices, DebugOptions);
 };
 
 struct ComputeSystem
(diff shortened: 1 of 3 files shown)


ptrivedi:
LGTM

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1107,4 +1166,72 @@ try
...
+
+void WSLAVirtualMachine::CreateVmSavedStateFile()
+{
+    constexpr auto c_extension = L".vmrs";
+    constexpr auto c_prefix = L"saved-state-";
+    const auto filename = std::format(L"{}{}-{}{}", c_prefix, std::time(nullptr), m_vmIdString, c_extension);
+    auto savedStateFile = m_crashDumpFolder / filename;
+
+    std::error_code error;
+    std::filesystem::create_directories(m_crashDumpFolder, error);
```
benhillis: If you're just going to throw on error anyway, you can use the version of create directories that throws.

src/windows/common/hcs_schema.h
```diff
@@ -481,13 +495,135 @@ struct ComputeSystem
...
+    OMIT_IF_EMPTY(j, c, GuestCrashSaveInfo);
+}
 
-    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(CrashReport, CrashLog);
+inline void from_json(const nlohmann::json& j, CrashReport& c)
+{
+    if (j.contains("CrashLog"))
+        c.CrashLog = j.at("CrashLog").get<std::wstring>();
+    if (j.contains("GuestCrashSaveInfo"))
+        c.GuestCrashSaveInfo = j.at("GuestCrashSaveInfo").get<GuestErrorSaveReport>();
```
benhillis: @OneBlue - I would assume our format script would throw an error on these, do you know why it is not?

src/windows/common/hcs_schema.h
```diff
@@ -481,13 +495,135 @@ struct ComputeSystem
...
+
+struct GuestCrashAttribution
+{
+    std::optional<std::vector<uint64_t>> CrashParameters;
+};
+
+inline void to_json(nlohmann::json& j, const GuestCrashAttribution& g)
+{
+    j = nlohmann::json::object();
+    if (g.CrashParameters.has_value())
```
benhillis: I do not see this pattern elsewhere in the file, @OneBlue knows this stuff better than me, there is probably an easier way.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
benhillis: we also probably don't want to log the folder since it will contain PII (the user name).

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -254,12 +272,20 @@ void WSLAVirtualMachine::Start()
     hvSocketConfig.HvSocketConfig.DefaultConnectSecurityDescriptor = securityDescriptor;
     vmSettings.Devices.HvSocket = std::move(hvSocketConfig);
 
+    CreateVmSavedStateFile();
+    WI_ASSERT(!m_vmSavedStateFile.empty());
```
benhillis, line 293: Why do we need to create the file before there is a crash?

benhillis: Is this some HCS limitation?

benhillis: We have a TempFile helper you'll probably want to use instead to avoid having to manually clean this up.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1107,4 +1166,72 @@ try
...
+    wsl::windows::common::wslutil::EnforceFileLimit(m_crashDumpFolder.c_str(), MAX_VM_CRASH_FILES, pred);
+
+    {
+        wil::unique_handle file{CreateFileW(filePath.c_str(), GENERIC_WRITE, 0, nullptr, CREATE_NEW, FILE_ATTRIBUTE_TEMPORARY, nullptr)};
+        THROW_LAST_ERROR_IF(!file);
+    }
+    {
+        std::wofstream outputFile(filePath.wstring());
+        THROW_HR_IF(E_UNEXPECTED, !outputFile || !(outputFile << crashLog));
+    }
```
benhillis, line 1317: This is a little odd, there's definitely a cleaner way to do this.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1107,4 +1166,72 @@ try
...
-CATCH_RETURN();
\ No newline at end of file
+CATCH_RETURN();
+
+std::filesystem::path WSLAVirtualMachine::GetCrashDumpFolder(std::wstring vmIdString)
+{
+    auto userToken = wsl::windows::common::security::GetUserToken(TokenImpersonation);
+    auto runAsUser = wil::impersonate_token(userToken.get());
+
+    auto tempPath = wsl::windows::common::filesystem::GetTempFolderPath(userToken.get());
```
benhillis: Do you need to impersonate *and* pass the token?

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1107,4 +1166,72 @@ try
...
+    return tempPath / L"wsla-crashes";
+}
+
+void WSLAVirtualMachine::CreateVmSavedStateFile()
+{
+    constexpr auto c_extension = L".vmrs";
+    constexpr auto c_prefix = L"saved-state-";
```
benhillis: constant strings should just be part of the std::format to make this more readable.

benhillis (changes requested):
:clock1:

src/windows/common/hcs_schema.h
benhillis: (if statement with no braces)

benhillis:
Looking good, some minor comments, some follow-up needed from Pierre.

src/windows/wslaservice/exe/WSLAVirtualMachine.h
```diff
@@ -100,6 +105,12 @@ class DECLSPEC_UUID("0CFC5DC1-B6A7-45FC-8034-3FA9ED73CE30") WSLAVirtualMachine
     std::wstring m_debugShellPipe;
 
     wsl::windows::common::hcs::unique_hcs_system m_computeSystem;
+
+    std::filesystem::path m_vmSavedStateFile;
```
benhillis, line 116: WSL itself has an entire temp directory that is cleaned up when the VM exits. That might make this a bit cleaner and be useful in future cases where we need files to stick around for the lifetime of the VM.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
kvega005: It is an HCS limitation. You need to create the file first and grant appropriate permissions so that VMWP can write to it.

src/windows/common/hcs_schema.h
OneBlue: We most likely need to update our .clang-format to explicitly check for that

src/windows/common/hcs_schema.h
OneBlue: On the to_json direction that's the best way I found. 

There's probably a smarter way to make nlohmann_json handle std::optional<T> for us, but that's probably something that we'll want as a separate change

OneBlue: It should be possible rewrite those as: 

```
OMIT_IF_EMPTY(j, GuestCrashAttribution, CrashParameters):
```

though

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -411,24 +441,58 @@ void WSLAVirtualMachine::OnExit(_In_ const HCS_EVENT* Event)
...
+    const auto exitStatus = wsl::shared::FromJson<wsl::windows::common::hcs::SystemExitStatus>(Event->EventData);
+
+    WslVirtualMachineTerminationReason reason = WslVirtualMachineTerminationReasonUnknown;
+
+    if (exitStatus.ExitType.has_value())
     {
-        // TODO: parse json and give a better error.
-        WslVirtualMachineTerminationReason reason = WslVirtualMachineTerminationReasonUnknown;
-        if (Event->Type == HcsEventSystemExited)
+        switch (exitStatus.ExitType.value())
```
OneBlue, line 522: Looks like we're switching this from using `Event->Type` to using `Event->EventData->ExitType`, what's the reason for that ?

src/windows/common/hcs_schema.h
kvega005: I do not know why the format script did not throw an error, but I will fix.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -411,24 +441,58 @@ void WSLAVirtualMachine::OnExit(_In_ const HCS_EVENT* Event)
...
+void WSLAVirtualMachine::OnCrash(_In_ const HCS_EVENT* Event)
+{
+    WSL_LOG(
+        "WSLAGuestCrash",
+        TraceLoggingValue(Event->EventData, "details"),
+        TraceLoggingValue(static_cast<int>(Event->Type), "type"));
+
+    if (m_crashLogCaptured && m_vmSavedStateCaptured)
+    {
+        return;
```
OneBlue, line 552: Should we log something if we hit this block ? It doesn't feel right for OnCrash to be called twice

OneBlue: Unless we expect to be called once with `CrashLog` and another time when `GuestCrashSaveInfo` is captured ?

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
OneBlue: Is PII in non-telemetry logging an issue ? We log the full path on the WSL side

OneBlue: Also the `create_directories` logic is pretty much the same as the WSL one. I originally wrote it with a re-throw so that we would log the path on failure

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
kvega005: This is just a copy of [WslCoreVm.cpp](https://github.com/microsoft/WSL/blob/b2c28b92fe3153deb98bfc9e59bb09150fcef268/src/windows/service/exe/WslCoreVm.cpp#L2362C1-L2366C1), but I agree It can be cleaner. Should I change it in both places?

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1107,4 +1166,72 @@ try
...
+    }
+
+    auto pred = [&c_extension, &c_prefix](const auto& e) {
+        return WI_IsFlagSet(GetFileAttributes(e.path().c_str()), FILE_ATTRIBUTE_TEMPORARY) && e.path().has_extension() &&
+               e.path().extension() == c_extension && e.path().has_filename() &&
+               e.path().filename().wstring().find(c_prefix) == 0 && e.file_size() > 0;
+    };
+
+    const auto maxSavedSateFiles = MAX_VM_CRASH_FILES;
+    wsl::windows::common::wslutil::EnforceFileLimit(m_crashDumpFolder.c_str(), MAX_VM_CRASH_FILES + 1, pred);
```
OneBlue: `wsl::windows::common::wslutil::EnforceFileLimit` might not do what we want here, since it will remove previous dumps before an actual crash dump is captured, since we're just creating an empty file here. 

We might want to call this in `OnCrash` instead, or remove it entirely for this PR and iterate on the crash dump collection logic in a future change

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1107,4 +1166,72 @@ try
...
+
+    wil::unique_handle file{CreateFileW(savedStateFile.c_str(), GENERIC_WRITE, 0, nullptr, CREATE_NEW, FILE_ATTRIBUTE_TEMPORARY, nullptr)};
+    THROW_LAST_ERROR_IF(!file);
+
+    hcs::GrantVmAccess(m_vmIdString.c_str(), savedStateFile.c_str());
+    m_vmSavedStateFile = savedStateFile;
+}
+
+void WSLAVirtualMachine::WriteCrashLog(std::wstring crashLog)
```
OneBlue: nit: use `const std::wstring& crashLog` to avoid  copy here

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1107,4 +1166,72 @@ try
     RETURN_IF_FAILED(Mount("none", LibrariesMountPoint, "overlay", options.c_str(), Flags));
     return S_OK;
 }
-CATCH_RETURN();
\ No newline at end of file
+CATCH_RETURN();
+
+std::filesystem::path WSLAVirtualMachine::GetCrashDumpFolder(std::wstring vmIdString)
```
OneBlue: nit: I'd recommend removing `vmIdString` since we can use `m_vmIdString` directly

OneBlue (commented):
Overall change LGTM, added some minor comments

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
kvega005: Yes, this expects to be called first with CrashLog then when the vmrs file is written to disk, OnCrash is called again.

OneBlue: Ok that makes sense. In that should we should always be called either once, or twice, but never more than twice right ?

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
kvega005: That's why the predicate includes` e.file_size() > 0`, so we do not count empty dumps towards the file limit. But I agree, it makes sense to call this after the VMRS file is actually written, not when we create the file.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
kvega005: Removing.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
kvega005: We use the extension and prefix in the predicate for `EnforceFileLimit` below.

src/windows/wslaservice/exe/WSLAVirtualMachine.h
kvega005: I don't think we want the files to be deleted when the VM exists though. We want to keep the files around so users can look at logs in the case of kernel panics. Here I only delete the file if it never got written to by HCS. We can get a little smarter about how/where we keep the files later. For now, this is just the skeleton for getting the .vmrs files on a crash.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
kvega005: `s_OnExit`, is called many times. Both crash and graceful exits see `HcsEventSystemExited` event.

In the event of a crash we get the following Event-Types:
1. `HcsEventSystemCrashInitiated` - contains crash log
2. `HcsEventSystemCrashReport` - contains crash log and system information
3. `HcsEventSystemCrashReport` - contains GuestCrashSaveInfo
4. `HcsEventSystemExited`

In the event of a graceful exit, we only get:
1. `HcsEventSystemExited`

So the real question is, do we want to signal the `OnTermination` as soon as we see HcsEventSystemCrashInitiated, or do we want to wait for the full crash info to come through before we signal `OnTermination`?

If the former is preferred, I should change this code to just signal `OnTermination` in `s_OnExit` based on event type.

If it's the latter, once we have processed the crash info, we will get a `HcsEventSystemExited` and need to parse its data to see what type of exit occured. That's what this code does.

OneBlue: That makes sense, thank you. I think the latter makes the most sense. 

We should definitely add test coverage for those cases, but that can happen in a future change

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1107,4 +1169,71 @@ try
...
+        SAVED_STATE_FILE_PREFIX,
+        std::time(nullptr), 
+        m_vmIdString,
+        SAVED_STATE_FILE_EXTENSION);
+
+    auto savedStateFile = m_crashDumpFolder / filename;
+
+    wsl::windows::common::filesystem::EnsureDirectory(m_crashDumpFolder.c_str());
+
+    wil::unique_handle file{CreateFileW(savedStateFile.c_str(), GENERIC_WRITE, 0, nullptr, CREATE_NEW, FILE_ATTRIBUTE_TEMPORARY, nullptr)};
```
OneBlue, line 1274: Although we're creating an empty file in the user's temp folder, let's impersonate when we create the file just to be safe. 

(I'm not sure if a user could easily spoof their %TMP% folder to make us write to a file where the user doesn't actually have write access to)

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1107,4 +1169,71 @@ try
...
+
+    auto pred = [&c_extension, &c_prefix](const auto& e) {
+        return WI_IsFlagSet(GetFileAttributes(e.path().c_str()), FILE_ATTRIBUTE_TEMPORARY) && e.path().has_extension() &&
+               e.path().extension() == c_extension && e.path().has_filename() && e.path().filename().wstring().find(c_prefix) == 0;
+    };
+
+    wsl::windows::common::wslutil::EnforceFileLimit(m_crashDumpFolder.c_str(), MAX_VM_CRASH_FILES, pred);
+
+    {
+        std::wofstream outputFile(filePath.wstring());
```
OneBlue, line 1312: We should impersonate the user when we create this file as well

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -398,10 +426,14 @@ void WSLAVirtualMachine::ConfigureNetworking()
 
 void CALLBACK WSLAVirtualMachine::s_OnExit(_In_ HCS_EVENT* Event, _In_opt_ void* Context)
 {
-    if (Event->Type == HcsEventSystemExited || Event->Type == HcsEventSystemCrashInitiated || Event->Type == HcsEventSystemCrashReport)
+    if (Event->Type == HcsEventSystemExited)
```
OneBlue, line 498: Now that we have code paths that can throw exceptions here, it's probably a good idea to add a try / CATCH_LOG() at this method's scope. We'd probably crash the service if an exception escaped this callback

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
kvega005: With our current DebugOptions, yes. See this [comment](https://github.com/microsoft/WSL/pull/13705#discussion_r2520232522). OnCrash gets called with HcsEventSystemCrashInitiated and HcsEventSystemCrashReport.

There are other debug options which may cause OnCrash to be called with a different payload. For example, if we enable collecting vmwp dumps we will get a crash report event with a CrashReportProcessDump.

OneBlue: I see, thank you

OneBlue (approved):

src/windows/common/hcs_schema.h
benhillis: I would prefer we use the OMIT_IF_EMPTY macro like the rest of the code does.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1083,15 +1147,10 @@ try
...
-
 #ifdef WSL_GPU_LIB_PATH
-
     auto packagedLibPath = std::filesystem::path(TEXT(WSL_GPU_LIB_PATH));
-
 #else
-
     auto packagedLibPath = wslutil::GetBasePath() / L"lib";
-
 #endif
```
benhillis, line 1241: don't modify this

benhillis (changes requested):

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
kvega005: Woops! I think this was the CLANG formatter tool. Undoing.

benhillis (approved):
:shipit:

PR #13712, opened 2025-11-12
.clang-format: add InsertBraces: true and minor fix to FormatSource.ps1

benhillis:
Update .clang-format to enforce braces after if statements. Also while testing this I noticed a small issue with the source formatting script that was causing it to not work when run directly from powershell.


benhillis:
Wild, I'm not sure what happened there I swear I built it locally - must have forgotten.

OneBlue (approved):

PR #13717, opened 2025-11-14
Add MSI for WSLA

florelis:
* Added a new CMake target to create a separate .msi for WSLA. The `.wix.in` is mostly extracted from the existing one.
* Renamed the existing `package.wix.in` to match the folder name, to try and make it more clear which `.wix` file is for what. (I had originally also named the WSLA as `package`, but that causes a conflict in the output folder. Since they need to have distinct names, it seemed better to have them match the folder structure.)
* Updated the test scripts to also install the WSLA MSI

Pending:
* Some components are duplicated in the two MSIs. I'm not sure if that's what we want, or if WSLA should depend on WSL.
* Update the docs.
* Maybe extract all the common pieces of `CMakeLists.txt` and share it for both packages?
* Update to give the MSI its own version independent of the main one


tools/deploy/deploy-to-host.ps1
```diff
@@ -12,34 +14,47 @@ param (
...
-    "/i"
-    $PackagePath
-    "/qn"
-    "/norestart"
-)
-
-if ($MsiArgs)
-{
-    $MSIArguments += $MsiArgs
+foreach ($t in $Target) {
```
OneBlue, line 58: I'd recommend having a script parameter to set if we want to install WSL, WSLA, or both. That way we can run the tests quickly when we iterate on a specific project

OneBlue: (Same for below script)

tools/test/test-setup.ps1
```diff
@@ -59,7 +59,7 @@ if ($Package) {
             )
 
         $exitCode = (Start-Process -Wait "msiexec.exe" -ArgumentList $MSIArguments -NoNewWindow -PassThru).ExitCode
-        if ($exitCode -Ne 0)
+        if ($exitCode -NotIn (0, 1605))
```
OneBlue, line 64: What's the reason for this change ?

CMakeLists.txt
```diff
@@ -220,7 +220,7 @@ if (${WSL_BUILD_THIN_PACKAGE})
 endif ()
 
 string(REPLACE "/Zi" "" CMAKE_CXX_FLAGS_DEBUG ${CMAKE_CXX_FLAGS_DEBUG})  # make sure /Zi is removed from the debug flags
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj /W3 /WX /ZH:SHA_256")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj /W3 /WX /ZH:SHA_256 /Zm1000")
```
OneBlue, line 223: Interesting, does this make the precompiled header build faster ?

wslamsi/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
+set(BIN ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${CMAKE_BUILD_TYPE})
+set(OUTPUT_PACKAGE ${BIN}/wsla.msi)
+set(PACKAGE_WIX_IN ${CMAKE_CURRENT_LIST_DIR}/wslamsi.wix.in)
+set(PACKAGE_WIX ${BIN}/wslamsi.wix)
+set(CAB_CACHE ${BIN}/cab)
+set(BINARIES wslaservice.exe;wslaserviceproxystub.dll;wsladiag.exe)
```
OneBlue: `wslrelay.exe` should also be there

wslamsi/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+    DEPENDS ${PACKAGE_WIX} ${BINARIES_DEPENDENCIES} # Make sure the package is rebuilt if any of the binaries or resources change
+)
+
+add_custom_target(wslamsi DEPENDS ${OUTPUT_PACKAGE})
+set_target_properties(wslamsi PROPERTIES EXCLUDE_FROM_ALL FALSE SOURCES ${PACKAGE_WIX_IN})
+add_dependencies(wslamsi wslaservice wslaserviceproxystub wsladiag)
+
+set_source_files_properties(${OUTPUT_PACKAGE} PROPERTIES GENERATED TRUE)
+
+if (DEFINED WSL_POST_BUILD_COMMAND)
```
OneBlue, line 42: This is also set in the wsl.msi target, we might want to have another "post build" command specific to WSLA, so the same command doesn't run twice here

OneBlue (commented):
Change looks great ! 

To make this work in the tests, we'll need to: 

- Capture the new MSI in [TestGroup.xml](https://github.com/microsoft/WSL/blob/master/cloudtest/TestGroup.xml.in)

- Update the install logic in [test-setup.ps1](https://github.com/microsoft/WSL/blob/master/tools/test/test-setup.ps1) 

We'll probably going to need a WSLA specific registry key (like what we do for WSL in `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Lxss\MSI`)  where the store the MSI product code and installed version (so we can uninstall it easily in scripts)

tools/test/test-setup.ps1
OneBlue: I guess it's because the script currently tries to uninstall the WSL MSI package. We should update that once we have a registry key that stores the WSLA MSI product code and uninstall&  install both package in this script

CMakeLists.txt
florelis: I don't think it makes it any faster. I added it because my computer was running out of memory frequently.

tools/deploy/deploy-to-host.ps1
florelis: The script has a parameter for that. It defaults to installing both, but you can also install only one.

tools/test/test-setup.ps1
florelis: Maybe I am not holding things right, but many times I got into the state of running the `test-setup.ps1` with one of the MSIs not being installed, so uninstalling it failed.

But now that I think about it, I guess the registry key shouldn't be there in that case because the uninstall should have removed it

wslamsi/CMakeLists.txt
florelis: The command is basically calling `deploy-to-host.ps1`, with the additional parameters set here. The additional parameters specify which of the two to install.

CMakeLists.txt
OneBlue: @johnstep  / @ptrivedi : This will probably fix the out of memory issue that you were seeing

PR #13718, opened 2025-11-14
Add targets to nuget package

florelis:
Added `.targets` files to the WSLA NuGet package so that MSBuild projects using it don't need to manually consume the contents (i.e., set the include directory, list it as a library dependency, copy the library to the output directory). The `.targets` files and other contents of the package (readme) are organized with the file structure that NuGet expects; to match this I also renamed the readme for the plugin API. The targets are mostly copied from one of the packages from winget-cli.


nuget/.gitignore
```diff
@@ -0,0 +1,2 @@
+!*/build
+!*.targets
```
OneBlue: We could probably remove the `build` and `.targets` from the root .gitignore. The first one isn't used for sure, and the second can most likely be specialized to `directory.build.targets` instead, which should solve this

OneBlue (approved):
LGTM. We'll probably need to make changes to header & library names when we make more progress on the SDK, but this is great starting point

yao-msft:
There was an empty `_._` file added, is that intentional?

nuget/.gitignore
florelis: Removed this file and those two lines from the root .gitignore. I did not see git pick up a `directory.build.targets` or any other `.targets` file.

florelis:
> There was an empty `_._` file added, is that intentional?

It was intentional to prevent this warning when packing the nuget:
> NU5127: This package does not contain a lib/ or ref/ folder, and will therefore be treated as compatible for all frameworks. Since framework specific files were found under the build/ directory for net, consider creating the following empty files to correctly narrow the compatibility of the package:
> -lib/net/_._

I was thinking that we could delete the file once we have the wrapper for .NET, but on second thought we would probably forget to delete it and the warning doesn't break anything, so I removed it.

PR #13719, opened 2025-11-14
wsla: Add support for creating processes via WSLAProcessLauncher from the service

OneBlue:

This change implements `WSLAProcessLauncher` , which enables process creation from within the service, with the same API as its client counterpart. 

Unfortunately adding tests there is not possible, since that code must run in the service

## PR Checklist


src/windows/wslaservice/exe/ServiceProcessLauncher.cpp
```diff
@@ -0,0 +1,60 @@
...
+{
+    process.CopyTo(m_process.GetAddressOf());
+}
+
+wil::unique_handle ServiceRunningProcess::GetStdHandle(int Index)
+{
+    return std::move(Get().GetStdHandle(Index));
+}
+
+wil::unique_event ServiceRunningProcess::GetExitEvent()
```
benhillis, line 33: Since this is just reaching into the process to get the handle, should this be a method on the process instead?

OneBlue: It could be as well, but that specific method is there so that `RunningWSLAProcess::WaitAndCaptureOutput()` can generically get the exit HANDLE, regardless of whether it's running from the client or service side, which allows us to factor the `WaitAndCaptureOutput()` logic for both in one place

benhillis (approved):

PR #13723, opened 2025-11-14
Add test wsla rootfs initialization in wsl client

yao-msft:
Add test wsla rootfs initialization in wsl client

## PR Checklist

Add initialization support for earlier preview test rootfs. Use --wsla-test-rootfs and --container-vhd to invoke this path.

## Validation Steps Performed
Verified containerd started. nerdctl works. Can pull down hello-world image and run it.


src/windows/common/WslClient.cpp
```diff
@@ -1554,13 +1556,15 @@ int WslaShell(_In_ std::wstring_view commandLine)
     parser.AddArgument(Integer(settings.MemoryMb), L"--memory");
     parser.AddArgument(Integer(settings.CpuCount), L"--cpu");
     parser.AddArgument(Utf8String(fsType), L"--fstype");
+    parser.AddArgument(testRootfs, L"--wsla-test-rootfs");
+    parser.AddArgument(containerRootVhd, L"--container-vhd");
```
OneBlue, line 1561: We could use only `containerRootVhd` and take that path as long as it's not empty to simplify things a bit

src/windows/common/WslClient.cpp
```diff
@@ -1583,11 +1587,36 @@ int WslaShell(_In_ std::wstring_view commandLine)
...
+        THROW_IF_FAILED(virtualMachine->Mount(diskDevice.get(), "/lsw", "squashfs", "ro", WslMountFlagsChroot | WslMountFlagsWriteableOverlayFs));
+
+        wsl::windows::common::WSLAProcessLauncher mountProcessLauncher{shell, {shell, "/etc/lsw-mount.sh"}};
+        auto mountProcess = mountProcessLauncher.Launch(*session);
+        THROW_HR_IF(E_FAIL, mountProcess.WaitAndCaptureOutput().Code != 0);
+
+        // mount container data disk
+        wil::unique_cotaskmem_ansistring containerVhdDevice;
+        ULONG containerVhdLun{};
+        THROW_IF_FAILED(virtualMachine->AttachDisk(containerRootVhd.c_str(), false, &containerVhdDevice, &containerVhdLun));
```
OneBlue: We'll probably want to create the VHD if it doesn't exist, for convenience

src/windows/common/WslClient.cpp
```diff
@@ -1554,13 +1556,15 @@ int WslaShell(_In_ std::wstring_view commandLine)
...
+    parser.AddArgument(testRootfs, L"--wsla-test-rootfs");
+    parser.AddArgument(containerRootVhd, L"--container-vhd");
     parser.AddArgument(help, L"--help");
     parser.Parse();
 
     if (help)
     {
         const auto usage = std::format(
-            LR"({} --wsla [--vhd </path/to/vhd>] [--shell </path/to/shell>] [--memory <memory-mb>] [--cpu <cpus>] [--dns-tunneling] [--fstype <fstype>] [--new-api] [--help])",
+            LR"({} --wsla [--vhd </path/to/vhd>] [--shell </path/to/shell>] [--memory <memory-mb>] [--cpu <cpus>] [--dns-tunneling] [--fstype <fstype>] [--wsla-test-rootfs] [--container-vhd </path/to/vhd>] [--help])",
```
OneBlue: Now that we support the actual rootfs vhd with containerd, we should probably add the nuget package to the project and use that VHD by default. 

The easiest way I can think of doing this would be to propagate the nuget package path via a macro, and default the VHD path to that value (like we do for WSL_SYSTEM_DISTRO_PATH today)

OneBlue:
FYI we'll probably to merge with https://github.com/microsoft/WSL/pull/13726 , since it will pass the vhd & vhdtype directly to CreateSession()

yao-msft:
> FYI we'll probably to merge with #13726 , since it will pass the vhd & vhdtype directly to CreateSession()

Just to confirm, I should move the logic to CreateSession() as well after merging that pr?

OneBlue:
> > FYI we'll probably to merge with #13726 , since it will pass the vhd & vhdtype directly to CreateSession()
> 
> Just to confirm, I should move the logic to CreateSession() as well after merging that pr?

If by logic you mean the call to `/etc/lsw-init.sh`, I think we should leave it on the client side for now, since it's going to interact with tests. 

In the long term I think we'll eventually need this in the service though

yao-msft:
> > > FYI we'll probably to merge with #13726 , since it will pass the vhd & vhdtype directly to CreateSession()
> > 
> > 
> > Just to confirm, I should move the logic to CreateSession() as well after merging that pr?
> 
> If by logic you mean the call to `/etc/lsw-init.sh`, I think we should leave it on the client side for now, since it's going to interact with tests.
> 
> In the long term I think we'll eventually need this in the service though

Ok, then I'll leave them in client for this pr.

CMakeLists.txt
```diff
@@ -373,6 +374,7 @@ endif()
 
 if (DEFINED WSL_DEV_BINARY_PATH) # Development shortcut to make the package smaller
     add_compile_definitions(WSL_SYSTEM_DISTRO_PATH="${WSL_DEV_BINARY_PATH}/system.vhd"
+                            WSLA_TEST_DISTRO_PATH="${WSL_DEV_BINARY_PATH}/wslatestrootfs.vhd"
```
OneBlue: I'd recommend using `${WSLA_TEST_DISTRO_SOURCE_DIR}/wslatestrootfs.vhd` directly here, so we can skip the copy in `UserConfig.cmake` (that way this will work even if `WSL_DEV_BINARY_PATH` is not set

OneBlue: And if do that, let's always set that variable, outside of the `IF` here

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -379,7 +379,11 @@ void WSLAVirtualMachine::ConfigureMounts()
         MountGpuLibraries("/usr/lib/wsl/lib", "/usr/lib/wsl/drivers", WSLAMountFlagsNone);
     }
 
-    // TODO: Mount storage VHD here.
+    if (m_settings.ContainerRootVhd) // TODO: re-think how container root settings should work at the session level API.
+    {
+        auto [_, containerRootDevice] = AttachDisk(m_settings.ContainerRootVhd, false);
+        Mount(m_initChannel, containerRootDevice.c_str(), "/root", "ext4", "rw", 0);
```
OneBlue, line 399: This will most likely fail if the VHD was just created (and so isn't formatted yet). We might want to add a flag to tell the service to format the VHD, so things work "automatically"

OneBlue (approved):

PR #13728, opened 2025-11-15
deps: update a number of NuGet packages to the latest available versions

benhillis:
This updates most NuGet packages to the latest available versions.


nuget.config
```diff
@@ -19,11 +19,11 @@
     <packageSources>
         <clear />
         <add key="WSL" value="https://pkgs.dev.azure.com/shine-oss/wsl/_packaging/WslDependencies/nuget/v3/index.json" />
+        <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
OneBlue: Unfortunately this would be an SFI violation so we can't have nuget.org directly in nuget.config

OneBlue (changes requested):

OneBlue (approved):

PR #13735, opened 2025-11-17
Remove the old API - phase 2: Remove the GPU methods

OneBlue:

This changes updates the WSLASesssion logic to always mount the GPU folders in a constant path (if enabled) and removes the old GPU mountpoints APIs

## PR Checklist

This change removes tghe 


src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1243,9 +1245,7 @@ try
     }
 
     Mount(m_initChannel, "none", LibrariesMountPoint, "overlay", options.c_str(), Flags);
-    return S_OK;
 }
-CATCH_RETURN();
```
benhillis, line 1248: Should this be catch_log? Unclear where will handle the exception.

OneBlue: An exception thrown here would get all the way to WSLAUserSession::CreateSession(), which is what we want, since the VM failing to start should fail the session creation

OneBlue: This change removes the try / catch because the WSLAVirtualMachine method isn't a COM API anymore

benhillis (approved):

PR #13741, opened 2025-11-18
Add crash dump collection

kvega005:
Copies the WSL feature that collects user-mode process crash dumps to WSLA. 

## PR Checklist


src/linux/init/WSLAInit.cpp
```diff
@@ -796,6 +820,18 @@ int WSLAEntryPoint(int Argc, char* Argv[])
         return -1;
     }
 
+    // Enable crash dump collection if requested.
+    if (getenv(WSL_ENABLE_CRASH_DUMP_ENV))
```
OneBlue: We originally did this on the WSL codepath because this behavior was user configurable, but for WSLA I'd be of the opinion to simply always enable it

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -317,6 +327,12 @@ void WSLAVirtualMachine::Start()
     auto socket = wsl::windows::common::hvsocket::Accept(listenSocket.get(), m_settings.BootTimeoutMs, m_vmTerminatingEvent.get());
     m_initChannel = wsl::shared::SocketChannel{std::move(socket), "mini_init", m_vmTerminatingEvent.get()};
 
+    auto crashDumpSocket = wsl::windows::common::hvsocket::Listen(runtimeId, LX_INIT_UTILITY_VM_CRASH_DUMP_PORT);
```
OneBlue, line 324: I'd recommend binding the socket and starting the thread before we accept the first hvsocket connection from init. That way if init crashes early during boot, we should be able to collect a process dump

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1325,4 +1341,58 @@ void WSLAVirtualMachine::OnProcessReleased(int Pid)
...
+
+            constexpr auto dumpExtension = ".dmp";
+            constexpr auto dumpPrefix = "wsl-crash";
+
+            auto filename = std::format("{}-{}-{}-{}-{}{}", dumpPrefix, message.Timestamp, message.Pid, process, message.Signal, dumpExtension);
+
+            std::replace_if(filename.begin(), filename.end(), [](auto e) { return !std::isalnum(e) && e != '.' && e != '-'; }, '_');
+
+            auto fullPath = m_crashDumpFolder / filename;
+
```
OneBlue, line 1362: nit: We probably want to log an ETL event when we receive a crash (similar to what WSL does) so the file can be easily found on the machine. Something like: 

```
            WSL_LOG(
                "WSLALinuxCrash",
                TraceLoggingValue(fullPath.c_str(), "FullPath"),
                TraceLoggingValue(message.Pid, "Pid"),
                TraceLoggingValue(message.Signal, "Signal"),
                TraceLoggingValue(process, "process"));
```

OneBlue (commented):
Change LGTM. Few minor comments

OneBlue:
Also it would probably a good idea to add test coverage for this. To to that, we could write a test case that essentially does [this](https://github.com/microsoft/WSL/blob/f0d257f760f7aaf1614a47fb49e93b40a288eddf/test/windows/WSLATests.cpp#L901C11-L909C55) but instead of sending SIGKILL, it could send SIGSEGV, and validate that the dump is collected where we expect it to be.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
kvega005: Makes sense.

test/windows/WSLATests.cpp
```diff
@@ -969,4 +969,59 @@ class WSLATests
...
+
+        auto crashDumpsDir = std::filesystem::temp_directory_path() / "wsla-crashes";
+        VERIFY_IS_TRUE(std::filesystem::exists(crashDumpsDir));
+
+        // Dumps files are named with the format: wsl-crash-<sessionId>-<pid>-<processname>-<code>.dmp
+        // Check if file exists in crash dumps directory matching the pattern.
+        bool dumpFound = false;
+        std::string expectedPattern = std::format("wsl-crash-*-{}-_usr_bin_cat-11.dmp", processId);
+
+        for (const auto& entry : std::filesystem::directory_iterator(crashDumpsDir))
```
OneBlue: Unfortunately this might create a race condition, since writing the crash dump on the Windows side isn't guaranteed to be completed when `process.WaitAndCaptureOutput()` returns. 

To work around this (and some other small things), I would recommend: 

- Before running the tests, grab the list of dump inside `crashDumpsDir`
- Use `wsl::shared::retry::RetryWithTimeout()` and expect a file that wasn't there before to match `expectedPattern`
- Have a scope_exit() to delete the dump when the test case completes so we don't leak it

OneBlue (approved):

OneBlue (approved):

PR #13746, opened 2025-11-19
Fix invalid arg exception thrown when OnVmTerminated is invoked by WSLAVirtualMachine destructor 

kvega005:
This PR fixes an exception that could occur when removing stale virtual machine references in WSLAUserSessionImpl::OnVmTerminated. Previously, calling erase on an empty std::vector<WSLAVirtualMachine*> could trigger a debug assertion in MSVC STL due to invalid iterator bounds.

Why is m_virtualMachines empty during WSLAVirtualMachine destruction?
This can happen if:
- The WSLAUserSessionImpl object itself is being destroyed, which also destroys its member containers (like m_virtualMachines).
-  During destruction, the WSLAVirtualMachine destructor calls OnVmTerminated, but by this point, m_virtualMachines may already be empty (or even partially destroyed), so the erase operation is acting on an empty vector.

This exception consistently reproduces when stopping WSLAService via `net stop`.

## PR Checklist

### DIFF
--- a/src/windows/wslaservice/exe/WSLAUserSession.cpp
+++ b/src/windows/wslaservice/exe/WSLAUserSession.cpp
@@ -42,8 +42,7 @@ void WSLAUserSessionImpl::OnVmTerminated(WSLAVirtualMachine* machine)
     // Remove any stale VM reference.
     if (!m_virtualMachines.empty())
     {
-        auto pred = [machine](const auto* e) { return machine == e; };
-        m_virtualMachines.erase(std::remove_if(m_virtualMachines.begin(), m_virtualMachines.end(), pred), m_virtualMachines.end());
+        std::erase_if(m_virtualMachines, [machine](const auto* e) { return machine == e; });
     }
 }
 
--- a/src/windows/wslaservice/exe/WSLAUserSession.cpp
+++ b/src/windows/wslaservice/exe/WSLAUserSession.cpp
@@ -38,10 +38,13 @@ WSLAUserSessionImpl::~WSLAUserSessionImpl()
 void WSLAUserSessionImpl::OnVmTerminated(WSLAVirtualMachine* machine)
 {
     std::lock_guard lock(m_lock);
-    auto pred = [machine](const auto* e) { return machine == e; };
 
     // Remove any stale VM reference.
-    m_virtualMachines.erase(std::remove_if(m_virtualMachines.begin(), m_virtualMachines.end(), pred), m_virtualMachines.end());
+    if (!m_virtualMachines.empty())
+    {
+        auto pred = [machine](const auto* e) { return machine == e; };
+        m_virtualMachines.erase(std::remove_if(m_virtualMachines.begin(), m_virtualMachines.end(), pred), m_virtualMachines.end());
+    }
 }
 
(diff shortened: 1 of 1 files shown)


src/windows/wslaservice/exe/WSLAUserSession.cpp
```diff
@@ -38,10 +38,13 @@ WSLAUserSessionImpl::~WSLAUserSessionImpl()
...
 {
     std::lock_guard lock(m_lock);
-    auto pred = [machine](const auto* e) { return machine == e; };
 
     // Remove any stale VM reference.
-    m_virtualMachines.erase(std::remove_if(m_virtualMachines.begin(), m_virtualMachines.end(), pred), m_virtualMachines.end());
+    if (!m_virtualMachines.empty())
+    {
+        auto pred = [machine](const auto* e) { return machine == e; };
+        m_virtualMachines.erase(std::remove_if(m_virtualMachines.begin(), m_virtualMachines.end(), pred), m_virtualMachines.end());
```
benhillis: NIT: std::erase_if

src/windows/wslaservice/exe/WSLAUserSession.cpp
```diff
@@ -38,10 +38,13 @@ WSLAUserSessionImpl::~WSLAUserSessionImpl()
 void WSLAUserSessionImpl::OnVmTerminated(WSLAVirtualMachine* machine)
 {
     std::lock_guard lock(m_lock);
-    auto pred = [machine](const auto* e) { return machine == e; };
 
     // Remove any stale VM reference.
-    m_virtualMachines.erase(std::remove_if(m_virtualMachines.begin(), m_virtualMachines.end(), pred), m_virtualMachines.end());
+    if (!m_virtualMachines.empty())
```
OneBlue, line 43: Thank you for doing this ! This is liveness issue that I mentioning in the sync this morning. 

Unfortunately with the current state of the code, `OnVmTerminated ` can be invoked `m_virtualMachines` is already destroyed, so checking if it's empty could also lead to a crash.

I'm currently working on redesigning vm termination logic to solve this, but if we want a quick fix, we could simply have this method return without doing anything, that should avoid crashes for now

kvega005: Experimentally, checking empty has not led to crash yet, although it could. Whereas, not checking empty led to a crash 100% of the time on my device, and it causes issues where things are not being cleaned up properly. So, I added this as a temporary measure just to relieve that annoyance.

I considered a few alternatives. For example, we could add a ClearUserSession() to WSLAVirtualMachine, which gets called in the WSLAUserSessionImpl destructor.

We do definitely need a better long-term solution. Ultimately, we will need an event handling model where the User Session can hook/unhook to events that are fired by the Virtual Machine without having to supply the Virtual Machine with a pointer to user session.

This may be applicable: [Event Handling in COM](https://learn.microsoft.com/en-us/cpp/cpp/event-handling-in-com?view=msvc-170)

OneBlue: Here's the change with the ownership rework: https://github.com/microsoft/WSL/pull/13747 

This gets rid of the old CreateVirtualMachine() API (which is what OnVmTerminated() was originally added for) and fixes a few other issues. 

This still needs a bit more work to support our long term design, but it should resolve this issue

PR #13753, opened 2025-11-20
Implement WSLASession::FormatVirtualDisk

OneBlue:
## PR Checklist

### DIFF
--- a/src/windows/wslaservice/exe/WSLASession.cpp
+++ b/src/windows/wslaservice/exe/WSLASession.cpp
@@ -16,6 +16,7 @@ Abstract:
 #include "WSLASession.h"
 #include "WSLAUserSession.h"
 #include "WSLAContainer.h"
+#include "ServiceProcessLauncher.h"
 
 using wsl::windows::service::wsla::WSLASession;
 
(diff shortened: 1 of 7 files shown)


src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -102,9 +103,27 @@ try
...
+    // Attach the disk to the VM.
+    auto [lun, device] = m_virtualMachine->AttachDisk(Path, false);
+
+    // N.B. DetachDisk calls sync() before detaching.
+    auto detachDisk = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [this, lun]() { m_virtualMachine->DetachDisk(lun); });
+
+    // Format it to ext4.
+    constexpr auto mkfsPath = "/usr/sbin/mkfs.ext4";
+    ServiceProcessLauncher launcher(mkfsPath, {mkfsPath, device});
+    auto result = launcher.Launch(*m_virtualMachine).WaitAndCaptureOutput();
```
benhillis: I feel like it is a bit odd for the launcher to take the VM as a parameter.

benhillis: the Vm already has a createlinuxprocess api, why can't that be used directly?

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -102,9 +103,27 @@ try
 CATCH_RETURN();
 
 HRESULT WSLASession::FormatVirtualDisk(LPCWSTR Path)
```
benhillis, line 130: We have to be careful with this path, in fact we probably need to make sure it's an absolute path since the calling process and service will have different working directories.

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -102,9 +103,27 @@ try
...
+try
 {
-    return E_NOTIMPL;
+    std::lock_guard lock{m_lock};
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_virtualMachine.has_value());
+
+    // Attach the disk to the VM.
```
benhillis: Might be nice to clarify via the comment that AttachDisk does the permission check.

benhillis (changes requested):
Two nits and one thing that needs to be changed (enforce absolute paths)

src/windows/wslaservice/exe/WSLASession.cpp
OneBlue: WSLAVirtualMachine::CreateLinuxProcess is the lower level API. This is what it looks like: 

```
    Microsoft::WRL::ComPtr<WSLAProcess> CreateLinuxProcess(_In_ const WSLA_PROCESS_OPTIONS& Options, [...]);
```

To call it, the caller needs to initialize a WSLA_PROCESS_OPTIONS structure, which includes the command line array, env array and file descriptor array, which is pretty heavy.  On top of this, we get an WSLAProcess pointer back, and from there we'd have to read the stdhandles and wait for the process to exit, which is also pretty heavy.

To make this easier, we have a "higher level" API (*ProcessLauncher) that handles all of this for us, so we can create a process and get its output in a few lines, like in this specific example. 

The reason for passing the VM in `Launch()` is that the long term plan is to have another overload for that method that will take a WSLAContainer instead, so we can all linux process being created, and waited for in one single place, with a single implementation of the IO consumption logic.

benhillis: ok that's fair - I'm not 100% sold on having the class get passed in as an argument, but we can adjust later if I come up with something better.

benhillis (approved):
Thanks for making the changes!

PR #13755, opened 2025-11-20
Fix service crash when collecting a linux crash dump when maxCrashDumpCount is set to 0

OneBlue:

This change fixes a potential service crash that can hit when a linux crash dump is collected when `wsl2.maxCrashDumpCount` is set to zero and the crash folder doesn't contain any crash files. 

## PR Checklist

### DIFF
--- a/src/windows/common/wslutil.cpp
+++ b/src/windows/common/wslutil.cpp
@@ -594,6 +594,8 @@ std::wstring wsl::windows::common::wslutil::DownloadFile(std::wstring_view Url, 
 
 void wsl::windows::common::wslutil::EnforceFileLimit(LPCWSTR Path, size_t Limit, const std::function<bool(const std::filesystem::directory_entry&)>& pred)
 {
+    WI_ASSERT(Limit > 0);
+
     std::map<std::filesystem::file_time_type, std::filesystem::path> files;
     for (auto const& e : std::filesystem::directory_iterator{Path})
     {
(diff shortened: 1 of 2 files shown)


src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -1272,7 +1272,10 @@ void WslCoreVm::CollectCrashDumps(wil::unique_socket&& listenSocket) const
                        e.path().filename().string().find(dumpPrefix) == 0;
             };
 
-            wsl::windows::common::wslutil::EnforceFileLimit(m_vmConfig.CrashDumpFolder.c_str(), m_vmConfig.MaxCrashDumpCount, pred);
+            if (m_vmConfig.MaxCrashDumpCount > 0)
+            {
+                wsl::windows::common::wslutil::EnforceFileLimit(m_vmConfig.CrashDumpFolder.c_str(), m_vmConfig.MaxCrashDumpCount, pred);
```
benhillis: Can we push the 0 limit behavior into the EnforceFileLimit helper to just do nothing?

src/windows/common/wslutil.cpp
```diff
@@ -594,6 +594,8 @@ std::wstring wsl::windows::common::wslutil::DownloadFile(std::wstring_view Url,
 
 void wsl::windows::common::wslutil::EnforceFileLimit(LPCWSTR Path, size_t Limit, const std::function<bool(const std::filesystem::directory_entry&)>& pred)
 {
+    WI_ASSERT(Limit > 0);
```
benhillis: In other words, change this assert to if (limit <= 0) return;

benhillis (approved):

PR #13758, opened 2025-11-21
Implement Create Container 

ptrivedi:


src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -53,3 +64,68 @@ try
...
+    ServiceProcessLauncher launcher(nerdctlPath, args, {}, common::ProcessFlags::None);
+    for (size_t i = 0; i < containerOptions.InitProcessOptions.FdsCount; i++)
+    {
+        launcher.AddFd(containerOptions.InitProcessOptions.Fds[i]);
+    }
+
+    return wil::MakeOrThrow<WSLAContainer>(&parentVM, launcher.Launch(parentVM));
+}
+
+std::vector<std::string> WSLAContainer::prepareNerdctlRunCommand(const WSLA_CONTAINER_OPTIONS& options)
```
yao-msft: nit:
```suggestion
std::vector<std::string> WSLAContainer::PrepareNerdctlRunCommand(const WSLA_CONTAINER_OPTIONS& options)
```

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -18,6 +18,15 @@ Module Name:
 
 using wsl::windows::service::wsla::WSLAContainer;
 
+const std::string nerdctlPath = "/usr/bin/nerdctl";
```
yao-msft, line 21: nit: and similar for below
```suggestion
constexpr std::string_view c_nerdctlPath = "/usr/bin/nerdctl"sv;
```

test/windows/WSLATests.cpp
```diff
@@ -411,7 +411,7 @@ class WSLATests
             auto [hresult, _, process] = launcher.LaunchNoThrow(*session);
             VERIFY_ARE_EQUAL(hresult, expectedError);
 
-            return process;
+            return std::move(process);
```
yao-msft, line 452: nit: no need to move

OneBlue: We actually do since processes are not copyable anymore

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: Ended up using a constexpr const char*, since that makes it easier to add to the command line vectors

OneBlue:
Closing since this has been merged via different PR

PR #13760, opened 2025-11-21
cleanup: VirtioNetworking refactoring

benhillis:
This pull request refactors how `VirtioNetworking` is constructed and integrated with `WslCoreVm`, simplifying its initialization and improving code clarity. The main changes involve replacing the previous configuration object with explicit parameters and callbacks, removing chained setter methods, and moving callback logic into dedicated handler methods within `WslCoreVm`.

### DIFF
--- a/src/windows/service/exe/VirtioNetworking.h
+++ b/src/windows/service/exe/VirtioNetworking.h
@@ -17,11 +17,13 @@ class VirtioNetworking : public INetworkingEngine
 {
 public:
     VirtioNetworking(GnsChannel&& gnsChannel, bool enableLocalhostRelay);
+    VirtioNetworking(
+        GnsChannel&& gnsChannel,
+        bool enableLocalhostRelay,
+        AddGuestDeviceRoutine addGuestDeviceRoutine,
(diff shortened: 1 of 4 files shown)


src/windows/service/exe/VirtioNetworking.h
```diff
@@ -9,20 +9,21 @@
 
 namespace wsl::core {
 
-using AddGuestDeviceRoutine = std::function<GUID(const GUID& clsid, const GUID& deviceId, PCWSTR tag, PCWSTR options)>;
+using AddGuestDeviceCallback = std::function<GUID(const GUID& clsid, const GUID& deviceId, PCWSTR tag, PCWSTR options)>;
```
OneBlue, line 12: nit: That's not introduced by this change, but if we need 3 callbacks to go into the VM, I think it gets to the point where having an interface for it will make things a bit cleaner.

OneBlue (approved):

src/windows/service/exe/VirtioNetworking.h
benhillis: yes that would be a nice followup.

PR #13762, opened 2025-11-21
wsla: add support for ARM64

benhillis:
This change will get WSLA working for ARM64.


OneBlue (approved):
LGTM. We have no ARM64 test coverage right now so I wouldn't be surprised if there are a bunch of other similar things that we need to do for ARM64 to boot correctly

PR #13764, opened 2025-11-22
WSLA: Implement virtioproxy networking mode

benhillis:
This change implements the VirtioProxy networking mode for WSLA. This required moving the VirtioProxy and GnsChannel classes into the common library (with NatNetworking). I also included the cleanup change to VirtioNetworking that was just merged to main in this PR.

DNS tunneling is currently not supported with this networking mode, so an error is returned. With this change I have updated the default networking move of `wsl.exe --wsla` to also use VirtioProxy networking mode.

### DIFF
--- a/src/windows/common/WslClient.cpp
+++ b/src/windows/common/WslClient.cpp
@@ -1547,7 +1547,7 @@ int WslaShell(_In_ std::wstring_view commandLine)
     settings.DisplayName = L"WSLA";
     settings.MemoryMb = 1024;
     settings.BootTimeoutMs = 30000;
-    settings.NetworkingMode = WSLANetworkingMode::WSLANetworkingModeNAT;
+    settings.NetworkingMode = WSLANetworkingModeNAT;
     std::wstring containerRootVhd;
     bool help = false;
 
--- a/src/windows/common/WslClient.cpp
+++ b/src/windows/common/WslClient.cpp
@@ -1547,7 +1547,7 @@ int WslaShell(_In_ std::wstring_view commandLine)
     settings.DisplayName = L"WSLA";
     settings.MemoryMb = 1024;
     settings.BootTimeoutMs = 30000;
-    settings.NetworkingMode = WSLANetworkingModeNAT;
+    settings.NetworkingMode = WSLANetworkingMode::WSLANetworkingModeNAT;
     std::wstring containerRootVhd;
     bool help = false;
 
@@ -1557,6 +1557,7 @@ int WslaShell(_In_ std::wstring_view commandLine)
     parser.AddArgument(reinterpret_cast<bool&>(settings.EnableDnsTunneling), L"--dns-tunneling");
     parser.AddArgument(Integer(settings.MemoryMb), L"--memory");
     parser.AddArgument(Integer(settings.CpuCount), L"--cpu");
+    parser.AddArgument(Integer(reinterpret_cast<int&>(settings.NetworkingMode)), L"--networking-mode");
     parser.AddArgument(Utf8String(fsType), L"--fstype");
     parser.AddArgument(containerRootVhd, L"--container-vhd");
     parser.AddArgument(help, L"--help");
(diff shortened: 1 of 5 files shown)


src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -513,15 +530,62 @@ void WSLAVirtualMachine::ConfigureNetworking()
             wil::unique_socket{(SOCKET)process->GetStdHandle(gnsChannelFd).release()},
             config,
             dnsChannelFd != -1 ? wil::unique_socket{(SOCKET)process->GetStdHandle(dnsChannelFd).release()} : wil::unique_socket{});
+    }
+    else if (m_settings.NetworkingMode == WSLANetworkingModeVirtioProxy)
+    {
+        // Launch GNS
+        std::vector<WSLA_PROCESS_FD> fds(1);
```
OneBlue: We should launch the port relay like we do in NAT mode (LaunchPortRelay();) so port mapping works the same way between NAT & virtio

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -513,15 +530,62 @@ void WSLAVirtualMachine::ConfigureNetworking()
...
+        options.Executable = "/init";
+        options.Fds = fds.data();
+        options.FdsCount = static_cast<DWORD>(fds.size());
+
+        // Because the file descriptors numbers aren't known in advance, the command line needs to be generated after the file
+        // descriptors are allocated.
+
+        std::string socketFdArg;
+        int gnsChannelFd = -1;
+        auto prepareCommandLine = [&](const auto& sockets) {
```
OneBlue: Ah I should wire this in the ProcessLauncher util, so we can use it here too

src/windows/common/WslClient.cpp
```diff
@@ -1557,6 +1557,7 @@ int WslaShell(_In_ std::wstring_view commandLine)
     parser.AddArgument(reinterpret_cast<bool&>(settings.EnableDnsTunneling), L"--dns-tunneling");
     parser.AddArgument(Integer(settings.MemoryMb), L"--memory");
     parser.AddArgument(Integer(settings.CpuCount), L"--cpu");
+    parser.AddArgument(Integer(reinterpret_cast<int&>(settings.NetworkingMode)), L"--networking-mode");
```
OneBlue, line 1560: nit: Could we have a `--virtionet` instead ? Having to lookup the enum values feels a bit user unfriendly

OneBlue: That would also remove the need for the below switch

benhillis: We could, but I thought it might be nice to test no networking as well. This is just a temporary test entrypoint anyway, so I thought this was ok.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -513,15 +529,48 @@ void WSLAVirtualMachine::ConfigureNetworking()
...
+            gnsChannelFd = sockets[0].Fd;
+            socketFdArg = std::to_string(gnsChannelFd);
+            cmd.emplace_back(socketFdArg.c_str());
+            options.CommandLine = cmd.data();
+            options.CommandLineCount = static_cast<DWORD>(cmd.size());
+        };
+
+        auto process = CreateLinuxProcess(options, nullptr, prepareCommandLine);
```
ptrivedi: There is potential to avoid code duplication for some of the code leading up to here between NAT and Virtio modes

src/windows/wslaservice/exe/WSLAVirtualMachine.h
```diff
@@ -18,6 +18,8 @@ Module Name:
 #include "Dmesg.h"
 #include "WSLAApi.h"
 #include "WSLAProcess.h"
+#include "GuestDeviceManager.h"
+#include "DnsResolver.h"
```
ptrivedi: It would be good to make these alphabetical, at least for the newer WSLA code

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -513,15 +529,48 @@ void WSLAVirtualMachine::ConfigureNetworking()
...
+        auto gnsChannel = wsl::core::GnsChannel(wil::unique_socket{(SOCKET)process->GetStdHandle(gnsChannelFd).release()});
```
ptrivedi: Where does Stop() on gnsChannel get called?

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
benhillis: yeah true, I'll clean this up.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
benhillis: Looks like currently virtio networking doesn't set this event, I'll look at this in a follow up.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
benhillis: I moved things around a bit more to share the create process stuff so this will be easier to do.

OneBlue (approved):

PR #13792, opened 2025-11-26
test: refactor NetworkingTests to take much less time.

benhillis:
While looking at adding some additional tests for VirtioProxy networking mode, I noticed a lot of problems with the existing networking test class. This change is some minor updates that will:

1. Reduce the number of VMs that need to be created, this will improve runtime substantially
2. Improve the handling of WSL1 tests so the entire class is skipped instead of each test method individually

### DIFF
--- a/test/windows/NetworkTests.cpp
+++ b/test/windows/NetworkTests.cpp
@@ -280,21 +280,20 @@ class NetworkTests
         std::wstring VmCreatorId;
     };
 
-    GUID AdapterId;
+    GUID m_adapterId;
+    std::optional<WslConfigChange> m_config;
 
     TEST_CLASS_SETUP(TestClassSetup)
     {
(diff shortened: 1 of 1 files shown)


test/windows/NetworkTests.cpp
```diff
@@ -1893,23 +1847,13 @@ class NetworkTests
             TEST_METHOD_PROPERTY(L"Data:NetConfig", L"{1, 2, 3, 4}")
         END_TEST_METHOD_PROPERTIES()
 
-        // All networking modes for both WSL1/2 are expected to support TCP/IPv4 host to guest loopback by default.
+        // All networking modes for are expected to support TCP/IPv4 host to guest loopback by default.
```
Brian-Perkins:
```suggestion
        // All networking modes are expected to support TCP/IPv4 host to guest loopback by default.
```

PR #13791, opened 2025-11-26
Initial implementation for CreateContainer()

OneBlue:
## Summary of the Pull Request

This change introduces an initial "native" implementation of CreateContainer(), along with --container and --debug-shell options for wsl --wsla. 

A lot still needs to be done in WSLAContainer, but this is a good starting point.

This change is based on @ptrivedi's initial CreateContainer PR: https://github.com/microsoft/WSL/pull/13758 

## PR Checklist

- [ ] **Closes:** Link to issue #xxx
- [ ] **Communication:** I've discussed this with core contributors already. If work hasn't been agreed, this work might be rejected
- [X] **Tests:** Added/updated if needed and all pass
- [ ] **Localization:** All end user facing strings can be localized
- [ ] **Dev docs:** Added/updated if needed
- [ ] **Documentation updated:** If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/wsl/) and link it here: #xxx

PR #13814, opened 2025-12-02
Refactor the session creation logic to match the new API 

OneBlue:

** This change depends on: https://github.com/microsoft/WSL/pull/13791 **

This change refactors the session creation logic to: 

- Make the storage disk creation easier
- Get rid of the old VM settings
- Run the init script from within the service
- Simplify testing

After this change, running wsl --wsla will open a shell in a VM already configured to run `nerdctl`. `--storage` can be passed to enable on disk storage, otherwise a tmpfs will be used to allow container creation. 

## PR Checklist


src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -54,6 +116,71 @@ WSLASession::~WSLASession()
...
+        });
+
+        auto result =
+            wil::ResultFromException([&]() { diskDevice = m_virtualMachine->AttachDisk(m_storageVhdPath.c_str(), false).second; });
+
+        if (FAILED(result))
+        {
+            THROW_HR_IF_MSG(
+                result,
+                result != HRESULT_FROM_WIN32(ERROR_PATH_NOT_FOUND) && result != HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND),
```
benhillis: why are we ignoring these failures?

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -54,6 +116,71 @@ WSLASession::~WSLASession()
...
+        std::filesystem::path storagePath{Settings.StoragePath};
+        THROW_HR_IF_MSG(E_INVALIDARG, !storagePath.is_absolute(), "Storage path is not absolute: %ls", storagePath.c_str());
+
+        m_storageVhdPath = storagePath / "storage.vhdx";
+
+        std::string diskDevice;
+        std::optional<ULONG> diskLun{};
+        bool vhdCreated = false;
+
+        auto deleteVhdOnFailure = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
```
benhillis: impersonate when deleting the file

src/windows/wslaservice/exe/WSLASession.cpp
benhillis: I see, still seems odd to try to attach a disk that doesn't exist.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -209,14 +200,14 @@ void WSLAVirtualMachine::Start()
     kernelCmdLine += L" hv_utils.timesync_implicit=1";
 
     wil::unique_handle dmesgOutput;
-    if (m_settings.DmesgOutput != 0)
+    if (m_settings.DmesgHandle)
     {
-        dmesgOutput.reset(wsl::windows::common::wslutil::DuplicateHandleFromCallingProcess(ULongToHandle(m_settings.DmesgOutput)));
+        dmesgOutput = std::move(m_settings.DmesgHandle);
```
benhillis: can this whole condition be removed and just always move the dmesg handle? won't harm anythigng if it's empty.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -310,6 +301,7 @@ void WSLAVirtualMachine::Start()
     WSL_LOG("CreateWSLAVirtualMachine", TraceLoggingValue(json.c_str(), "json"));
 
     m_computeSystem = hcs::CreateComputeSystem(m_vmIdString.c_str(), json.c_str());
+    m_running = true;
```
benhillis: Running is a poor name since the VM has not started yet.

src/windows/common/WslClient.cpp
```diff
@@ -1587,46 +1582,42 @@ int WslaShell(_In_ std::wstring_view commandLine)
...
-    WSLA_SESSION_SETTINGS sessionSettings{L"WSLA Test Session"};
     wil::com_ptr<IWSLASession> session;
-    settings.RootVhd = vhd.c_str();
-    settings.RootVhdType = fsType.c_str();
+
+    if (!rootVhdOverride.empty())
+    {
+        sessionSettings.RootVhdOverride = rootVhdOverride.c_str();
+
+        if (rootVhdTypeOverride.empty())
```
ptrivedi, line 1594: nit: check this before sessionSettings.RootVhdOverride assignment

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -54,6 +116,71 @@ WSLASession::~WSLASession()
...
+            // Then format it.
+            Ext4Format(diskDevice);
+        }
+
+        // Mount the device to /root.
+        m_virtualMachine->Mount(diskDevice.c_str(), "/root", "ext4", "", 0);
+
+        deleteVhdOnFailure.release();
+    }
+    else
```
ptrivedi: nit: early return / guard clausing would be nice for readability's sake here -- keeps the big if block flat and at zero-indentation level

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -33,28 +33,14 @@ constexpr auto SAVED_STATE_FILE_EXTENSION = L".vmrs";
 constexpr auto SAVED_STATE_FILE_PREFIX = L"saved-state-";
 constexpr auto RECEIVE_TIMEOUT = 30 * 1000;
 
-WSLAVirtualMachine::WSLAVirtualMachine(const VIRTUAL_MACHINE_SETTINGS& Settings, PSID UserSid, WSLAUserSessionImpl* Session) :
-    m_settings(Settings), m_userSid(UserSid)
+WSLAVirtualMachine::WSLAVirtualMachine(WSLAVirtualMachine::Settings&& Settings, PSID UserSid, WSLAUserSessionImpl* Session) :
```
ptrivedi: Remove WSLAUserSession references from WSLAVirtualMachine

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
ptrivedi: Does it belong after StartComputeSystem?

(where, oh where, in this 200+ line function shall we set it?) :p

ptrivedi: Does it belong after StartComputeSystem?

(200+ line function, phew :) )

src/windows/wslaservice/exe/WSLASession.cpp
OneBlue: Good catch thank you

src/windows/wslaservice/exe/WSLASession.cpp
OneBlue: I thought about that vs explicitly checking if the file exists, my thinking was that they're equivalent, but in the case where the VHD does exists, this approach allows us to avoid a file access to validate that the VHD exists

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
OneBlue: Good point, it can

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
OneBlue: Indeed. Moved after StartComputeSystem

ptrivedi (approved):

PR #13817, opened 2025-12-02
WSLA: Add wsladiag tool and improve session enumeration (ListSessions + display name fix)

beena352:
This PR introduces the initial version of the wsladiag diagnostic tool and improves the WSLA service’s session enumeration. The changes simplify session management, remove unnecessary allocations, fix display-name handling, and ensure the MSI correctly packages the new tool.
Key Changes
WSLA Service

Added CopyDisplayName method to WSLASession to remove COM string allocations.

Updated WSLAUserSessionImpl::ListSessions to use fixed-size buffers and avoid CoTaskMemFree logic.

Removed unused GetDisplayName IDL/COM method.

Cleaned up session tracking (m_sessions) by removing old m_wslaSessions references.

Added missing includes and declaration for the updated telemetry provider.

Fixed telemetry provider registration name mismatch (WslaTelemetryProvider).

wsladiag Tool

Added a new wsladiag executable under src/windows/wsladiag.

Introduced minimal wsladiag.cpp entry point and CMake integration.

Registered the new target in the root CMakeLists.

Updated MSI packaging to include wsladiag.exe.

MSI / Packaging

Updated msipackage/CMakeLists.txt to add wsladiag.exe to the binary list.

Updated package.wix.in to install wsladiag.exe into the WSL installation directory.


src/windows/wsladiag/wsladiag.cpp
```diff
@@ -0,0 +1,132 @@
...
+        {
+            wslutil::PrintMessage(std::format(L"Found {} WSLA session{}:\n", sessions.size(), sessions.size() > 1 ? L"s" : L""), stdout);
+
+            wslutil::PrintMessage(L"ID\tCreator PID\tDisplay Name\n", stdout);
+            wslutil::PrintMessage(L"--\t-----------\t------------\n", stdout);
+
+             for (const auto& session : sessions)
+            {
+                const auto* displayName = session.DisplayName;
+                if (displayName[0] == L'\0')
```
OneBlue: We should need a special block for this. (We also probably shouldn't allow empty strings as DisplayName, but we can do that in a followup change)

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -54,10 +60,12 @@ WSLASession::~WSLASession()
...
 }
 
-HRESULT WSLASession::GetDisplayName(LPWSTR* DisplayName)
+void WSLASession::CopyDisplayName(
+    _Out_writes_z_(bufferLength) PWSTR buffer,
+    size_t bufferLength) const
 {
-    *DisplayName = wil::make_unique_string<wil::unique_cotaskmem_string>(m_displayName.c_str()).release();
-    return S_OK;
+    WI_ASSERT(m_displayName.size() + 1 <= bufferLength);
```
OneBlue: I'd recommend throwing an exception here, so the service doesn't crash if we ever hit this

OneBlue (changes requested):
This PR should be targeted to `feature/wsl-for-apps` (it's currently targeting `master`)

Let's retarget it and I'll have a deeper look

beena352:
Implemented both suggested changes:
• Removed the special-case for empty DisplayName in wsladiag
• Updated CopyDisplayName to use THROW_HR_IF(E_BOUNDS, …)

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
OneBlue, line 16: The changes in that file probably come from the feature branch.  I'd recommend merging the latest `feature/wsl-for-apps` to make the diff cleaner

OneBlue (approved):
LGTM ! Feel to publish the PR. 

Next step would be to add test coverage, but that can be done in a followup change

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -294,15 +294,14 @@ interface IWSLASession : IUnknown
...
     HRESULT GetVirtualMachine([out] IWSLAVirtualMachine **VirtualMachine);
 }
 
 struct WSLA_SESSION_INFORMATION
 {
-    ULONG Id;
+    ULONG SessionId;
     DWORD CreatorPid;
-    LPSTR DisplayName;
+    wchar_t DisplayName[256];
```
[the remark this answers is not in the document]
beena352: Good suggestion, for now I’m keeping the buffer as-is to match existing patterns in the service. Validation can be added in a follow-up if needed

src/windows/wslaservice/exe/WSLAUserSession.cpp
```diff
@@ -80,6 +81,26 @@ HRESULT WSLAUserSessionImpl::OpenSessionByName(LPCWSTR DisplayName, IWSLASession
...
 
+HRESULT wsl::windows::service::wsla::WSLAUserSessionImpl::ListSessions(_Out_ WSLA_SESSION_INFORMATION** Sessions, _Out_ ULONG* SessionsCount)
+{
+    std::lock_guard lock(m_wslaSessionsLock);
+    auto output = wil::make_unique_cotaskmem<WSLA_SESSION_INFORMATION[]>(m_sessions.size());
+
+    size_t index = 0;
+    for (auto* session : m_sessions)
+    {
+        output[index].SessionId = session->GetId();
```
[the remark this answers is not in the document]
beena352: CreatorPid is not tracked yet, so I’m leaving it at 0 for now. We can add a TODO or populate it in a follow-up when that feature exists

msipackage/CMakeLists.txt
```diff
@@ -39,15 +39,15 @@ add_custom_command(
...
+add_dependencies(msipackage wsl wslg wslservice wslhost wslrelay wslserviceproxystub init initramfs wslinstall msixgluepackage wslaservice wslaserviceproxystub wsladiag)
 
 if (WSL_BUILD_WSL_SETTINGS)
     add_dependencies(msipackage wslsettings libwsl)
 endif()
 
 set_source_files_properties(${OUTPUT_PACKAGE} PROPERTIES GENERATED TRUE)
 
-if (DEFINED WSL_POST_BUILD_COMMAND)
+if (DEFINED WSL_POST_BUILD_COMMAND AND NOT "${WSL_POST_BUILD_COMMAND}" STREQUAL "")
```
OneBlue, line 50: What's the reason for that change ?

src/windows/wslaservice/exe/WSLASession.h
```diff
@@ -23,11 +23,21 @@ class DECLSPEC_UUID("4877FEFC-4977-4929-A958-9F36AA1892A4") WSLASession
...
     ~WSLASession();
 
-    IFACEMETHOD(GetDisplayName)(LPWSTR* DisplayName) override;
-    const std::wstring& DisplayName() const;
+    ULONG GetId() const noexcept
+    {
+        return m_id;
+    }
+
+    const std::wstring& GetDisplayName() const noexcept
```
OneBlue: Unless we have a specific reason to rename this method and move it to header, I'd prefer to keep as it is today (that will also reduce the changes of conflicts with other changes)

beena352: Got it, thanks for pointing that out! I’ll revert the method name and keep it in the current location to avoid unnecessary conflicts.

msipackage/CMakeLists.txt
OneBlue: Looks like this changed has been merged. Could you give context on the reason behind this change ?

beena352: Thanks for asking! The earlier version of this check used only DEFINED WSL_POST_BUILD_COMMAND. I had temporarily added an empty-string guard while testing a local scenario, but that extra check isn’t needed for the normal build flow.

I’ve created a follow-up draft PR to revert this and restore the original simpler condition for consistency.

PR #13819, opened 2025-12-03
Add WSLA-specific class registration for virtio networking

benhillis:
This change updates WSLA to use its own registration of wsldevicehost.dll and updates the virtio networking code to support running with either the WSL or WSLA versions of the class.

Currently since the WSLA service is part of the WSL MSI things were working fine, however when there is a separate package for the WSLA runtime a separate version of the wsldevicehost registration will be needed.


src/windows/common/VirtioNetworking.cpp
```diff
@@ -114,13 +114,12 @@ try
 
     THROW_IF_WIN32_ERROR(NotifyNetworkConnectivityHintChange(&VirtioNetworking::OnNetworkConnectivityChange, this, true, &m_networkNotifyHandle));
 }
-CATCH_LOG()
```
benhillis, line 117: This was swallowing failures, I think it's probably better for initialize to throw instead of fail silently and leave the VM with partially configured networking.

OneBlue (approved):

PR #13821, opened 2025-12-04
Implement container state management 

OneBlue:

This change adds logic to keep track of running containers and their state within a WSLASession

## PR Checklist


src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -26,6 +26,37 @@ static std::vector<std::string> defaultNerdctlRunArgs{//"--pull=never", // TODO:
...
+    {
+        // TODO: empty string is returned while the container image is still downloading.
+        // Remove this logic once the image pull is separated from container creation.
+        if (status != "created" && status != "")
+        {
+            THROW_HR_MSG(
+                E_UNEXPECTED, "Unexpected nerdctl status '%hs', for container '%hs'", status.value_or("<empty>").c_str(), m_name.c_str());
+        }
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
benhillis, line 52: why the sleep?

OneBlue: It's to avoid burning CPU while we loop

yao-msft, line 52: Do we consider to use "nerdctl events" in the future which could listen to container state changes instead of polling in loop?

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -26,6 +26,37 @@ static std::vector<std::string> defaultNerdctlRunArgs{//"--pull=never", // TODO:
...
+
+        status = GetNerdctlStatus();
```
yao-msft, line 54: By the way, since we are separating container creation and running, we could improve our code by separating into 2 calls "nerdctl create" then "nerdctl start". I think that may solve some of our port mapping timing problems we faced earlier when we use "nerdctl run"

ptrivedi: We will have to look into the options for create and start combination. nerdctl run seems to have a much richer set of options, we will have to see how it translates to create+start

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: That's a great idea. If that's OK by you, I'd like to merge the change as-is since it will unblock various things, but I'll study using nerdctl events in a followup PR

src/windows/wslaservice/exe/WSLAContainer.cpp
yao-msft: "The nerdctl create command similar to nerdctl run -d except the container is never started. You can then use the nerdctl start <container_id> command to start the container at any point."
This is from nerdctl command reference doc. I 'think' create and run should have very similar options, except it just does not start the main thread automatically.

src/windows/wslaservice/exe/WSLAContainer.cpp
yao-msft: For sure, my comment is not blocking :)

src/windows/wslaservice/exe/WSLAContainer.cpp
yao-msft: "The nerdctl create command similar to nerdctl run -d except the container is never started. You can then use the nerdctl start <container_id> command to start the container at any point."
This is from nerdctl command reference doc. I 'think' create and run should have very similar options, except it just does not start the main thread automatically.

ptrivedi: https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-nerdctl-run

https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-nerdctl-start

Will need some looking into

yao-msft: I meant "run" vs "create" + "start".  run and create should be similar.

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -312,4 +353,15 @@ try
     m_virtualMachine.Reset();
     return S_OK;
 }
-CATCH_RETURN();
\ No newline at end of file
+CATCH_RETURN();
+
+void WSLASession::ClearDeletedContainers()
+{
+    std::lock_guard lock{m_lock};
```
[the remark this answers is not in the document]
benhillis: Was this where we were considering using weak refs?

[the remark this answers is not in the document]
OneBlue: No, because we don't want containers to be deleted when the caller drops their reference to it. 

The weak references were for the WSLAUserSession <-> WSLASession relationship

ptrivedi (approved):
Approving this to unblock further changes.

As discussed, 
1) we will be looking at splitting "nerdctl run..." into "nerdctl create + start"
2) avoiding polling for container status by moving to "nerdctl events"

PR #13827, opened 2025-12-04
Draft public sdk api for wsla

yao-msft:
## PR Checklist


src/windows/wslasdk/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+set(SOURCES wsla.cpp)
+set(HEADERS wsla.h)
+
+add_library(wslasdk SHARED ${SOURCES} ${HEADERS} wslasdk.def)
```
OneBlue, line 4: I was originally of having the methods inside the `wslaclient` target, since it's essentially going to be the same thing conceptually. 

I don't mind renaming it though, but if we do that let's delete the old target

src/windows/wslasdk/wsla.h
```diff
@@ -0,0 +1,223 @@
...
+    WSLA_CONTAINER_STATE_RUNNING = 2,
+    WSLA_CONTAINER_STATE_EXITED = 3,
+    WSLA_CONTAINER_STATE_FAILED = 4,
+} WSLA_CONTAINER_STATE;
+
+STDAPI WslaGetContainerState(_In_ WslaRuntimeContainer container, _Out_ WSLA_CONTAINER_STATE* state);
+
+// Container Process
+
+STDAPI WslaCreateContainerProcess(_In_ WslaRuntimeContainer container, _In_ const WSLA_CONTAINER_PROCESS_OPTIONS* options, _Out_ WSLA_CONTAINER_PROCESS* process);
```
OneBlue, line 186: We might want to follow the COM logic a bit closer to make the integration easier. 

Something like: 

```
WslaGetContainerInitProcess(WslaRuntimeContainer container, WSLA_CONTAINER_PROCESS* process);
```

might be easier to implement

yao-msft: This one was the "nerdctl exec". I think you mean splitting CreateNewContainer into 2 methods? That works for me too.

src/windows/wslasdk/wsla.h
```diff
@@ -0,0 +1,223 @@
...
+    WslaContainerImageProgressCallback progressCallback;
+    PVOID progressCallbackContext;
+} WLSA_IMPORT_CONTAINER_IMAGE_OPTIONS;
+
+STDAPI WslaImportContainerImage(_In_ WslaSession session, _In_ const WLSA_PULL_CONTAINER_IMAGE_OPTIONS* options);
+
+typedef struct WSLA_CONTAINER_IMAGE_INFO
+{
+    PCSTR repository;
+    PCSTR tag;
```
yao-msft, line 99: Maybe just the imageName and use char[]

src/windows/wslasdk/wsla.h
ptrivedi: I am a bit confused by your comment too @OneBlue. I think the API above would map to WSLAContainerImpl::Exec in the backend.
I was also thinking we may not need something like GetContainerInitProcess in the public-facing API at all...?

src/windows/wslasdk/wslasdk.def
```diff
@@ -0,0 +1,22 @@
+LIBRARY wslasdk
+
+EXPORTS
+    WslaCanRun
+    WslaGetVersion
+    WslaInstallWithDependencies
+    WslaCreateSession
+    WslaReleaseSession
+    WslaPullContainerImage
+    WslaImportContainerImage
```
ptrivedi, line 10: Does this WslaImportContainerImage here correspond to LoadImage? Load and Import are a bit different.

src/windows/wslasdk/wslasdk.def
```diff
@@ -0,0 +1,22 @@
...
+    WslaListContainerImages
+    WslaDeleteContainerImage
+    WslaCreateNewContainer
+    WslaStartContainer
+    WslaStopContainer
+    WslaDeleteContainer
+    WslaRestartContainer
+    WslaGetContainerState
```
ptrivedi, line 18: What about inspection of an image?

PR #13828, opened 2025-12-04
Add --debug-shell support to wsladiag and integrate session opening + process launcher

beena352:
This PR introduces initial support for the new --debug-shell option in wsladiag, enabling developers to launch and debug WSLA sessions more easily.

Key changes

Added command-line parsing for --debug-shell <SessionName>.

Wired session opening through IWSLAUserSession::OpenSessionByName when a session name is provided.

Added fallback to CreateSession when no existing session name is passed.

Integrated process launcher support to start the debug shell in the associated virtual machine.

Performed minor formatting and whitespace cleanup across touched files.

Notes

This is a minimal first implementation of the debug-shell workflow.

Follow-up improvements (error handling, structured output, alignment with service changes) will come in future PRs.

### DIFF
--- a/msipackage/CMakeLists.txt
+++ b/msipackage/CMakeLists.txt
@@ -47,7 +47,7 @@ endif()
 
 set_source_files_properties(${OUTPUT_PACKAGE} PROPERTIES GENERATED TRUE)
 
-if (DEFINED WSL_POST_BUILD_COMMAND)
+if (DEFINED WSL_POST_BUILD_COMMAND AND NOT "${WSL_POST_BUILD_COMMAND}" STREQUAL "")
     add_custom_command(
         TARGET msipackage
         POST_BUILD
--- a/msipackage/CMakeLists.txt
+++ b/msipackage/CMakeLists.txt
@@ -1,29 +1,16 @@
 set(BIN ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${CMAKE_BUILD_TYPE})
-
-# For pipeline builds, use PACKAGE_INPUT_DIR if specified (contains signed binaries)
-# For local builds, use bin directory
-if(DEFINED PACKAGE_INPUT_DIR)
-    message(STATUS "Using signed binaries from ${PACKAGE_INPUT_DIR} for MSI packaging")
-else()
-    set(PACKAGE_INPUT_DIR ${BIN})
-endif()
-
 set(OUTPUT_PACKAGE ${BIN}/wsl.msi)
 set(PACKAGE_WIX_IN ${CMAKE_CURRENT_LIST_DIR}/package.wix.in)
 set(PACKAGE_WIX ${BIN}/package.wix)
 set(CAB_CACHE ${BIN}/cab)
-set(WINDOWS_BINARIES wsl.exe;wslg.exe;wslhost.exe;wslrelay.exe;wslservice.exe;wslserviceproxystub.dll;wslinstall.dll;wslaservice.exe;wsladiag.exe)
+set(BINARIES wsl.exe;wslg.exe;wslhost.exe;wslrelay.exe;wslservice.exe;wslserviceproxystub.dll;init;initrd.img;wslinstall.dll;wslaserviceproxystub.dll;wslaservice.exe;wsladiag.exe)
+
 if (WSL_BUILD_WSL_SETTINGS)
-    list(APPEND WINDOWS_BINARIES "wslsettings/wslsettings.dll;wslsettings/wslsettings.exe;libwsl.dll")
+    list(APPEND BINARIES_DEPENDENCIES "wslsettings/wslsettings.dll;wslsettings/wslsettings.exe;libwsl.dll")
 endif()
 
 set(BINARIES_DEPENDENCIES)
-foreach(binary ${WINDOWS_BINARIES})
-    list(APPEND BINARIES_DEPENDENCIES "${PACKAGE_INPUT_DIR}/${binary}")
-endforeach()
-
-set(LINUX_BINARIES init;initrd.img)
-foreach(binary ${LINUX_BINARIES})
+foreach(binary ${BINARIES})
     list(APPEND BINARIES_DEPENDENCIES "${BIN}/${binary}")
 endforeach()
 
--- a/msipackage/CMakeLists.txt
+++ b/msipackage/CMakeLists.txt
@@ -47,7 +47,7 @@ endif()
 
 set_source_files_properties(${OUTPUT_PACKAGE} PROPERTIES GENERATED TRUE)
 
-if (DEFINED WSL_POST_BUILD_COMMAND AND NOT "${WSL_POST_BUILD_COMMAND}" STREQUAL "")
+if (DEFINED WSL_POST_BUILD_COMMAND)
     add_custom_command(
         TARGET msipackage
         POST_BUILD
--- a/msipackage/CMakeLists.txt
+++ b/msipackage/CMakeLists.txt
@@ -1,16 +1,29 @@
 set(BIN ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${CMAKE_BUILD_TYPE})
+
+# For pipeline builds, use PACKAGE_INPUT_DIR if specified (contains signed binaries)
+# For local builds, use bin directory
+if(DEFINED PACKAGE_INPUT_DIR)
+    message(STATUS "Using signed binaries from ${PACKAGE_INPUT_DIR} for MSI packaging")
+else()
+    set(PACKAGE_INPUT_DIR ${BIN})
+endif()
+
 set(OUTPUT_PACKAGE ${BIN}/wsl.msi)
 set(PACKAGE_WIX_IN ${CMAKE_CURRENT_LIST_DIR}/package.wix.in)
 set(PACKAGE_WIX ${BIN}/package.wix)
 set(CAB_CACHE ${BIN}/cab)
-set(BINARIES wsl.exe;wslg.exe;wslhost.exe;wslrelay.exe;wslservice.exe;wslserviceproxystub.dll;init;initrd.img;wslinstall.dll;wslaserviceproxystub.dll;wslaservice.exe;wsladiag.exe)
-
+set(WINDOWS_BINARIES wsl.exe;wslg.exe;wslhost.exe;wslrelay.exe;wslservice.exe;wslserviceproxystub.dll;wslinstall.dll;wslaservice.exe;wsladiag.exe)
 if (WSL_BUILD_WSL_SETTINGS)
-    list(APPEND BINARIES_DEPENDENCIES "wslsettings/wslsettings.dll;wslsettings/wslsettings.exe;libwsl.dll")
+    list(APPEND WINDOWS_BINARIES "wslsettings/wslsettings.dll;wslsettings/wslsettings.exe;libwsl.dll")
 endif()
 
 set(BINARIES_DEPENDENCIES)
-foreach(binary ${BINARIES})
+foreach(binary ${WINDOWS_BINARIES})
+    list(APPEND BINARIES_DEPENDENCIES "${PACKAGE_INPUT_DIR}/${binary}")
+endforeach()
+
+set(LINUX_BINARIES init;initrd.img)
+foreach(binary ${LINUX_BINARIES})
     list(APPEND BINARIES_DEPENDENCIES "${BIN}/${binary}")
 endforeach()
 
--- a/src/windows/wsladiag/wsladiag.cpp
+++ b/src/windows/wsladiag/wsladiag.cpp
@@ -43,134 +43,169 @@ static int ReportError(const std::wstring& context, HRESULT hr)
     return 1;
 }
 
-static void PrintVerbose(bool verbose, std::wstring_view msg)
-{
-    if (verbose)
-    {
-        wslutil::PrintMessage(std::wstring(msg), stdout);
-    }
-}
-
-// Function to handle the --shell argument (TTY-backed interactive shell).
+// Handler for `wsladiag shell <SessionName>` (TTY-backed interactive shell).
 static int RunShellCommand(const std::wstring& sessionName, bool verbose)
 {
-    PrintVerbose(verbose, std::format(L"[diag] shell='{}'\n", sessionName));
+    const auto log = [&](std::wstring_view msg) {
+        if (verbose)
+        {
+            wslutil::PrintMessage(std::wstring(msg), stdout);
+        }
+    };
 
-    try
-    {
-        wil::com_ptr<IWSLAUserSession> userSession;
-        THROW_IF_FAILED(CoCreateInstance(__uuidof(WSLAUserSession), nullptr, CLSCTX_LOCAL_SERVER, IID_PPV_ARGS(&userSession)));
-        wsl::windows::common::security::ConfigureForCOMImpersonation(userSession.get());
+    log(std::format(L"[diag] shell='{}'\n", sessionName));
 
-        wil::com_ptr<IWSLASession> session;
-        THROW_IF_FAILED(userSession->OpenSessionByName(sessionName.c_str(), &session));
-        PrintVerbose(verbose, L"[diag] OpenSessionByName succeeded\n");
+    wil::com_ptr<IWSLAUserSession> userSession;
+    THROW_IF_FAILED(CoCreateInstance(__uuidof(WSLAUserSession), nullptr, CLSCTX_LOCAL_SERVER, IID_PPV_ARGS(&userSession)));
+    wsl::windows::common::security::ConfigureForCOMImpersonation(userSession.get());
 
-        // Console size for TTY.
-        CONSOLE_SCREEN_BUFFER_INFO info{};
-        THROW_LAST_ERROR_IF(!GetConsoleScreenBufferInfo(GetStdHandle(STD_OUTPUT_HANDLE), &info));
-        const ULONG rows = static_cast<ULONG>(info.srWindow.Bottom - info.srWindow.Top + 1);
-        const ULONG cols = static_cast<ULONG>(info.srWindow.Right - info.srWindow.Left + 1);
+    wil::com_ptr<IWSLASession> session;
+    THROW_IF_FAILED(userSession->OpenSessionByName(sessionName.c_str(), &session));
+    log(L"[diag] OpenSessionByName succeeded\n");
 
-        std::string shell = "/bin/sh";
+    // Console size for TTY.
+    CONSOLE_SCREEN_BUFFER_INFO info{};
+    THROW_LAST_ERROR_IF(!GetConsoleScreenBufferInfo(GetStdHandle(STD_OUTPUT_HANDLE), &info));
+    const ULONG rows = static_cast<ULONG>(info.srWindow.Bottom - info.srWindow.Top + 1);
+    const ULONG cols = static_cast<ULONG>(info.srWindow.Right - info.srWindow.Left + 1);
 
-        // Launch with terminal fds (PTY).
-        wsl::windows::common::WSLAProcessLauncher launcher{shell, {shell, "-i"}, {"TERM=xterm-256color"}, wsl::windows::common::ProcessFlags::None};
+    const std::string shell = "/bin/sh";
 
-        launcher.AddFd(WSLA_PROCESS_FD{.Fd = 0, .Type = WSLAFdTypeTerminalInput});
-        launcher.AddFd(WSLA_PROCESS_FD{.Fd = 1, .Type = WSLAFdTypeTerminalOutput});
-        launcher.AddFd(WSLA_PROCESS_FD{.Fd = 2, .Type = WSLAFdTypeTerminalControl});
-        launcher.SetTtySize(rows, cols);
+    // Launch with terminal fds (PTY).
+    wsl::windows::common::WSLAProcessLauncher launcher{
+        shell, {shell, "--login"}, {"TERM=xterm-256color"}, wsl::windows::common::ProcessFlags::None};
 
-        PrintVerbose(verbose, L"[diag] launching shell process...\n");
-        auto process = launcher.Launch(*session);
-        PrintVerbose(verbose, L"[diag] shell launched (TTY)\n");
+    launcher.AddFd(WSLA_PROCESS_FD{.Fd = 0, .Type = WSLAFdTypeTerminalInput});
+    launcher.AddFd(WSLA_PROCESS_FD{.Fd = 1, .Type = WSLAFdTypeTerminalOutput});
+    launcher.AddFd(WSLA_PROCESS_FD{.Fd = 2, .Type = WSLAFdTypeTerminalControl});
+    launcher.SetTtySize(rows, cols);
 
-        auto ttyIn = process.GetStdHandle(0);
-        auto ttyOut = process.GetStdHandle(1);
-        auto ttyControl = process.GetStdHandle(2);
+    log(L"[diag] launching shell process...\n");
+    auto process = launcher.Launch(*session);
+    log(L"[diag] shell launched (TTY)\n");
 
-        // Console handles.
-        wil::unique_hfile conin{CreateFileW(
-            L"CONIN$", GENERIC_READ | GENERIC_WRITE, FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, 0, nullptr)};
-        wil::unique_hfile conout{CreateFileW(
-            L"CONOUT$", GENERIC_READ | GENERIC_WRITE, FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, 0, nullptr)};
-        THROW_LAST_ERROR_IF(!conin);
-        THROW_LAST_ERROR_IF(!conout);
+    auto ttyIn = process.GetStdHandle(0);
+    auto ttyOut = process.GetStdHandle(1);
+    auto ttyControl = process.GetStdHandle(2);
 
-        const HANDLE consoleIn = conin.get();
-        const HANDLE consoleOut = conout.get();
+    // Console handles.
+    wil::unique_hfile conin{
+        CreateFileW(L"CONIN$", GENERIC_READ | GENERIC_WRITE, FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, 0, nullptr)};
+    wil::unique_hfile conout{
+        CreateFileW(L"CONOUT$", GENERIC_READ | GENERIC_WRITE, FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, 0, nullptr)};
+    THROW_LAST_ERROR_IF(!conin);
+    THROW_LAST_ERROR_IF(!conout);
 
-        // Save/restore console state.
-        DWORD originalInMode{};
-        DWORD originalOutMode{};
-        UINT originalOutCP = GetConsoleOutputCP();
-        UINT originalInCP = GetConsoleCP();
+    const HANDLE consoleIn = conin.get();
+    const HANDLE consoleOut = conout.get();
 
-        THROW_LAST_ERROR_IF(!GetConsoleMode(consoleIn, &originalInMode));
-        THROW_LAST_ERROR_IF(!GetConsoleMode(consoleOut, &originalOutMode));
+    // Save/restore console state.
+    DWORD originalInMode{};
+    DWORD originalOutMode{};
 
-        auto restoreConsole = wil::scope_exit([&] {
-            SetConsoleMode(consoleIn, originalInMode);
-            SetConsoleMode(consoleOut, originalOutMode);
-            SetConsoleOutputCP(originalOutCP);
-            SetConsoleCP(originalInCP);
-        });
+    THROW_LAST_ERROR_IF(!GetConsoleMode(consoleIn, &originalInMode));
+    THROW_LAST_ERROR_IF(!GetConsoleMode(consoleOut, &originalOutMode));
 
-        // Console mode for interactive terminal.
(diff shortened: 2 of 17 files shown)


src/windows/wsladiag/wsladiag.cpp
```diff
@@ -44,15 +47,20 @@ int wsladiag_main(std::wstring_view commandLine)
...
+    parser.AddArgument(debugShell, L"--debug-shell");
+
     parser.Parse();
 
     auto printUsage = []() {
         wslutil::PrintMessage(
             L"wsladiag - WSLA diagnostics tool\n"
             L"Usage:\n"
             L"  wsladiag --list    List WSLA sessions\n"
+            L"  wsladiag --debug-shell <SessionName>  Open a debug shell in an existing WSLA session\n"
```
OneBlue: `debug-shell` is a term that we used in the temporary wsl.exe tool, for this I'd recommend just using `shell` to make things easier

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -64,6 +72,59 @@ int wsladiag_main(std::wstring_view commandLine)
         return 0;
     }
 
+    if (!debugShell.empty())
```
OneBlue: Given that this argument is probably going to grow in complexity, I'd recommend creating a dedicated function to handle it

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -64,6 +72,59 @@ int wsladiag_main(std::wstring_view commandLine)
...
+            wsl::windows::common::WSLAProcessLauncher launcher{shell, {shell}, {"TERM=xterm-256color"}};
+            launcher.AddFd(WSLA_PROCESS_FD{.Fd = 0, .Type = WSLAFdTypeTerminalInput});
+            launcher.AddFd(WSLA_PROCESS_FD{.Fd = 1, .Type = WSLAFdTypeTerminalOutput});
+            launcher.AddFd(WSLA_PROCESS_FD{.Fd = 2, .Type = WSLAFdTypeTerminalControl});
+
+            wslutil::PrintMessage(L"[diag] launching shell process...\n", stdout);
+
+            auto process = launcher.Launch(*session);
+
+            wslutil::PrintMessage(std::format(L"Successfully opened debug shell session handle for '{}'.\n", debugShell), stdout);
```
OneBlue: LGTM so far. Here we'll want to actually relay the process' output to the terminal, like [we do in wslclient.cpp](https://github.com/microsoft/WSL/blob/3148403fbe5a186bc2178ea960d538d27b51b86c/src/windows/common/WslClient.cpp#L1669C4-L1730C94)

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -64,6 +72,59 @@ int wsladiag_main(std::wstring_view commandLine)
...
+
+            return 0;
+        }
+        catch (...)
+        {
+            const auto hr = wil::ResultFromCaughtException();
```
OneBlue: nit: let's factor error handling between all the arguments

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -64,6 +72,59 @@ int wsladiag_main(std::wstring_view commandLine)
...
+            wsl::windows::common::security::ConfigureForCOMImpersonation(userSession.get());
+
+            wil::com_ptr<IWSLASession> session;
+
+            // Open the existing session by name/ID from the --debug-shell argument.
+            THROW_IF_FAILED(userSession->OpenSessionByName(debugShell.c_str(), &session));
+
+            wslutil::PrintMessage(L"[diag] OpenSessionByName succeeded\n", stdout);
+
+            std::string shell = "/bin/bash";
```
OneBlue: We should probably default to `/bin/sh` since I suspect that bash won't be in our runtime image

msipackage/CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ endif()
 
 set_source_files_properties(${OUTPUT_PACKAGE} PROPERTIES GENERATED TRUE)
 
-if (DEFINED WSL_POST_BUILD_COMMAND)
+if (DEFINED WSL_POST_BUILD_COMMAND AND NOT "${WSL_POST_BUILD_COMMAND}" STREQUAL "")
```
OneBlue: This is probably coming from a merge issue, but let's remove this

OneBlue (commented):
LGTM so far.

We might want to discuss the format that we're going to take the command line arguments to differentiate between the different modes. 

I can think of two approaches: 

1)

```
wsl --list-sessions
wsl --shell <session> [additional args>
``` 

2)

```
wsl list-sessions
wsl shell <session> [ additional args>
```

I have a preference for 2) since that matches what nerdctl does and what users coming from container cli backgrounds will expect. 

Probably worth discussing in our next sync

src/windows/wslaservice/exe/WSLAUserSession.cpp
```diff
@@ -65,13 +65,13 @@ HRESULT WSLAUserSessionImpl::CreateSession(const WSLA_SESSION_SETTINGS* Settings
 
 HRESULT WSLAUserSessionImpl::OpenSessionByName(LPCWSTR DisplayName, IWSLASession** Session)
 {
-    std::lock_guard lock(m_lock);
+    std::lock_guard lock(m_wslaSessionsLock);
```
OneBlue, line 68: This part of the diff seems unrelated, let's refresh the branch to clean it up (same for the CMake change)

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -44,16 +175,21 @@ int wsladiag_main(std::wstring_view commandLine)
...
+
     parser.Parse();
 
     auto printUsage = []() {
         wslutil::PrintMessage(
             L"wsladiag - WSLA diagnostics tool\n"
             L"Usage:\n"
-            L"  wsladiag --list    List WSLA sessions\n"
-            L"  wsladiag --help    Show this help",
+            L"  wsladiag --list                  List WSLA sessions\n"
```
OneBlue: This can be done in a followup PR, but thinking through this I think splitting the argument parser into multiple verbs like: 

```
wsladbg list
wsladbg shell
wsldbg stop
[...]
```

Is probably the best approach going forward, since the command line logic is likely to grow in complexity

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,147 @@ Module Name:
...
+        wslutil::PrintMessage(L"[diag] OpenSessionByName succeeded\n", stdout);
+
+        std::string shell = "/bin/sh";
+
+        WSLAProcessLauncher launcher{shell, {shell}, {"TERM=xterm-256color"}};
+        launcher.AddFd(WSLA_PROCESS_FD{.Fd = 0, .Type = WSLAFdTypeTerminalInput});
+        launcher.AddFd(WSLA_PROCESS_FD{.Fd = 1, .Type = WSLAFdTypeTerminalOutput});
+        launcher.AddFd(WSLA_PROCESS_FD{.Fd = 2, .Type = WSLAFdTypeTerminalControl});
+
+        wslutil::PrintMessage(L"[diag] launching shell process...\n", stdout);
```
OneBlue: I imagine that those are for debugging, but once this ready to be merged I'd have to preference to either remove them, or put them behind a flag like `--verbose` to make the command's output easier to read

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,168 @@ Module Name:
...
+    if (verbose)
+    {
+        wslutil::PrintMessage(std::wstring(msg), stdout);
+    }
+}
+
+// Function to handle the --shell argument (TTY-backed interactive shell).
+static int RunShellCommand(const std::wstring& sessionName, bool verbose)
+{
+    PrintVerbose(verbose, std::format(L"[diag] shell='{}'\n", sessionName));
```
ptrivedi: nit: personal preference but would be nicer to use lambda for this one

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,168 @@ Module Name:
...
+            SetConsoleMode(consoleOut, originalOutMode);
+            SetConsoleOutputCP(originalOutCP);
+            SetConsoleCP(originalInCP);
+        });
+
+        // Console mode for interactive terminal.
+        DWORD inMode = originalInMode;
+        WI_SetAllFlags(inMode, ENABLE_WINDOW_INPUT | ENABLE_VIRTUAL_TERMINAL_INPUT);
+        WI_ClearAllFlags(inMode, ENABLE_ECHO_INPUT | ENABLE_LINE_INPUT | ENABLE_INSERT_MODE);
+        WI_SetFlag(inMode, ENABLE_PROCESSED_INPUT);
```
ptrivedi: Why is this flag being set separately here? Also, do we want to set this or unset it?

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,168 @@ Module Name:
...
+        THROW_LAST_ERROR_IF(!conout);
+
+        const HANDLE consoleIn = conin.get();
+        const HANDLE consoleOut = conout.get();
+
+        // Save/restore console state.
+        DWORD originalInMode{};
+        DWORD originalOutMode{};
+        UINT originalOutCP = GetConsoleOutputCP();
+        UINT originalInCP = GetConsoleCP();
```
ptrivedi: nit: I think these UINT declarations can go

beena352: I’ll move the UINT declarations closer to where they’re used for consistency and readability.

src/windows/wsladiag/wsladiag.cpp
beena352: It’s set separately because it controls console-level processing, unlike the other raw-input flags. We want it enabled so Ctrl+C behaves correctly in the interactive shell.

src/windows/wsladiag/wsladiag.cpp
beena352: Sure, I can switch this to a small lambda to keep the logging local.

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,168 @@ Module Name:
...
+
+        // It will keep terminal control socket alive.
+        auto exitEvent = wil::unique_event(wil::EventOptions::ManualReset);
+        wsl::shared::SocketChannel controlChannel{wil::unique_socket{(SOCKET)ttyControl.release()}, "TerminalControl", exitEvent.get()};
+
+        // Relay console -> tty input.
+        std::thread inputThread([&] {
+            try
+            {
+                wsl::windows::common::relay::InterruptableRelay(consoleIn, ttyIn.get());
```
OneBlue: We should use `wsl::windows::common::relay::StandardInputRelay` here since we want this to behave like a tty.

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,168 @@ Module Name:
...
+        // Console size for TTY.
+        CONSOLE_SCREEN_BUFFER_INFO info{};
+        THROW_LAST_ERROR_IF(!GetConsoleScreenBufferInfo(GetStdHandle(STD_OUTPUT_HANDLE), &info));
+        const ULONG rows = static_cast<ULONG>(info.srWindow.Bottom - info.srWindow.Top + 1);
+        const ULONG cols = static_cast<ULONG>(info.srWindow.Right - info.srWindow.Left + 1);
+
+        std::string shell = "/bin/sh";
+
+        // Launch with terminal fds (PTY).
+        wsl::windows::common::WSLAProcessLauncher launcher{shell, {shell, "-i"}, {"TERM=xterm-256color"}, wsl::windows::common::ProcessFlags::None};
```
OneBlue: What's the reason for passing `-i` here ?

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,168 @@ Module Name:
...
+        DWORD outMode = originalOutMode;
+        WI_SetAllFlags(outMode, ENABLE_PROCESSED_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING | DISABLE_NEWLINE_AUTO_RETURN);
+        THROW_IF_WIN32_BOOL_FALSE(SetConsoleMode(consoleOut, outMode));
+
+        THROW_LAST_ERROR_IF(!SetConsoleOutputCP(CP_UTF8));
+        THROW_LAST_ERROR_IF(!SetConsoleCP(CP_UTF8));
+
+        // It will keep terminal control socket alive.
+        auto exitEvent = wil::unique_event(wil::EventOptions::ManualReset);
+        wsl::shared::SocketChannel controlChannel{wil::unique_socket{(SOCKET)ttyControl.release()}, "TerminalControl", exitEvent.get()};
```
OneBlue: This channel needs to written to whenever the terminal is resized (otherwise the linux terminal size isn't updated). 

Here's an example on how to implement it: https://github.com/microsoft/WSL/blob/170bb8f3cc2ea63cf20926c9acad8b24d858b787/src/windows/common/WslClient.cpp#L1701 

To test it, you can run `top` inside the shell and verify that i's currently resized when the terminal window is resized

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,168 @@ Module Name:
...
+            }
+        });
+
+        // Relay tty output -> console (blocks until output ends).
+        wsl::windows::common::relay::InterruptableRelay(ttyOut.get(), consoleOut);
+
+        process.GetExitEvent().wait();
+        auto [code, signalled] = process.GetExitState();
+
+        const std::wstring shellW{shell.begin(), shell.end()};
```
OneBlue: This isn't needed since since::format can handle narrow string directly

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,168 @@ Module Name:
...
+
+        // Relay console -> tty input.
+        std::thread inputThread([&] {
+            try
+            {
+                wsl::windows::common::relay::InterruptableRelay(consoleIn, ttyIn.get());
+            }
+            catch (...)
+            {
+                exitEvent.SetEvent();
```
OneBlue: What's the reason for setting the `exitEvent` here ?

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -38,50 +190,55 @@ int wsladiag_main(std::wstring_view commandLine)
...
 
-    // Command-line parsing using ArgumentParser.
     ArgumentParser parser(std::wstring{commandLine}, L"wsladiag");
 
     bool help = false;
     bool list = false;
+    bool verbose = false;
+    std::wstring shellSession;
 
     parser.AddArgument(list, L"--list");
```
OneBlue: Given that the arguments are mutually exclusive, I'd recommend turning them into positional arguments and having one method per argument.

Something like:  https://github.com/microsoft/WSL/blob/c6ac7433b1a4f1a918ae14e74546474b461458f8/src/linux/init/wslinfo.cpp#L64 

would be a good example. 

I'd also be of the opinion to remove the `--` prefix to match docker & nerdctl's command line parsing logic (but that can be done in a followup change)

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,168 @@ Module Name:
...
+        auto [code, signalled] = process.GetExitState();
+
+        const std::wstring shellW{shell.begin(), shell.end()};
+        wslutil::PrintMessage(std::format(L"{} exited with: {}{}\n", shellW, code, signalled ? L" (signalled)" : L""), stdout);
+
+        return 0;
+    }
+    catch (...)
+    {
+        const auto hr = wil::ResultFromCaughtException();
```
OneBlue: I'd recommend having error handling at the `main` level, so we can have only one error handler for the entire program

OneBlue (changes requested):

src/windows/wsladiag/wsladiag.cpp
beena352: We could rely on the PTY alone, but keeping -i makes the behavior explicit and consistent across environments.

OneBlue: Ok I see. For this specifically I'd recommend doing the same thing we do since https://github.com/microsoft/WSL/pull/13920/changes and pass `--login` so the shell init scripts are loaded (see PR for more context)

src/windows/wsladiag/wsladiag.cpp
beena352: This is just to let the rest of the code know that input has stopped or something went wrong, so everything can shut down cleanly instead of hanging.

OneBlue: True, but since we don't pass it on line 159, if something went wrong on the stdin thread, the main thread would still hang

beena352: Yes, setting exitEvent won’t stop the stdout relay unless it’s passed in. I’ll switch the stdout relay to the InterruptableRelay(..., exitEvent) overload so we don’t hang if the stdin relay fails.

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,203 @@ Module Name:
...
+        THROW_IF_WIN32_BOOL_FALSE(GetConsoleScreenBufferInfoEx(consoleOut, &infoEx));
+
+        WSLA_TERMINAL_CHANGED message{};
+        message.Columns = infoEx.srWindow.Right - infoEx.srWindow.Left + 1;
+        message.Rows = infoEx.srWindow.Bottom - infoEx.srWindow.Top + 1;
+
+        controlChannel.SendMessage(message);
+    };
+
+    updateTerminalSize();
```
OneBlue: What's the reason for manually calling this here ? The initial tty size should be set when we launch the process

beena352: Yeah, the initial size is already set via SetTtySize at launch. I’ve removed the manual call and now only send size updates on resize.

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -38,97 +225,64 @@ int wsladiag_main(std::wstring_view commandLine)
...
     ArgumentParser parser(std::wstring{commandLine}, L"wsladiag");
 
     bool help = false;
-    bool list = false;
+    bool verbose = false;
+    std::wstring verb;
+    std::wstring shellSession;
+
+    parser.AddPositionalArgument(verb, 0);         // "list" or "shell"
+    parser.AddPositionalArgument(shellSession, 1); // session name for "shell"
```
OneBlue, line 268: Sorry if my previous comment was unclear. My recommendation is for each method to have its own parsing. For instance something like: 

```
std::wstring verb;
parser.AddPositionalArgument(verb, 0); 
parser.Parse();

if (verb == "list")
{
    List(commandLine);
}
else if (verb = "shell")
{
    Shell(commandLine);
}
else
{
    wslutil::PrintMessage(std::format("Unknown command: '{}'", verb);
}
```

That way each command can do its own parsing, for instance:

```
int Shell(const std::wstring& commandLine)
{
    std::wstring sessionName;
    std::string shell = "/bin/bash";

    ArgumentParser parser(std::wstring{commandLine}, L"wsladiag", 2); // Start index needs to be to 2, so we ignore the verb

 parser.AddPositionalArgument(sessionName, 0)
 parser.AddArgument(Utf8String(shell), L"--shell");

   [...]
}

```

That way each sub-command can do its own parsing.

To do this nicely though, we should probably add a flag to the parser, so we can avoid throwing an exception when unknown arguments are thrown [here](https://github.com/microsoft/WSL/blob/d48d374c8cba41dca6d041f3274c807015001d2e/src/shared/inc/CommandLine.h#L407) 

The easiest way would to add a parameter to the constructor that's unset by default, but when set skips the above error

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,203 @@ Module Name:
...
+        SetConsoleMode(consoleOut, originalOutMode);
+        SetConsoleOutputCP(originalOutCP);
+        SetConsoleCP(originalInCP);
+    });
+
+    // Console mode for interactive terminal.
+    DWORD inMode = originalInMode;
+    WI_SetAllFlags(inMode, ENABLE_WINDOW_INPUT | ENABLE_VIRTUAL_TERMINAL_INPUT);
+    WI_ClearAllFlags(inMode, ENABLE_ECHO_INPUT | ENABLE_LINE_INPUT | ENABLE_INSERT_MODE);
+    WI_SetFlag(inMode, ENABLE_PROCESSED_INPUT);
```
OneBlue: What's the reason for explicitly setting `ENABLE_PROCESSED_INPUT`  for the tty input mode ?

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -38,97 +225,64 @@ int wsladiag_main(std::wstring_view commandLine)
...
-        // No recognized command → show usage
-        printUsage();
-        return 0;
+        return RunListCommand(verbose);
     }
 
-    // --list: Call WSLA service COM interface to retrieve and display sessions.
-
-    try
+    if (verb == L"shell")
```
OneBlue: nit: else if

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -38,97 +225,64 @@ int wsladiag_main(std::wstring_view commandLine)
...
-            for (const auto& session : sessions)
-            {
-                const auto* displayName = session.DisplayName;
-
-                wslutil::PrintMessage(std::format(L"{}\t{}\t\t{}\n", session.SessionId, session.CreatorPid, displayName), stdout);
-            }
-        }
-
-        return 0;
+        return RunShellCommand(shellSession, verbose);
```
OneBlue, line 313: we should have a `else()` block do display an error like `unknown command: {command}` if the verb isn't known

src/windows/wsladiag/wsladiag.cpp
beena352: Thanks, that makes sense. I’ll change wsladiag_main to only parse the verb and dispatch to List(...) / Shell(...), and then each subcommand will parse its own arguments using the parser offset to skip the verb.

OneBlue (changes requested):
Thank you for the updated, added a few comments. 

Feel to do the command line parsing changes in a followup PR if that makes things easier

src/windows/wsladiag/wsladiag.cpp
beena352: We currently keep ENABLE_PROCESSED_INPUT enabled so the Windows console continues to handle Ctrl+C / Ctrl+Break correctly, while the other flags are cleared to get more TTY-like input behavior. We could also clear this flag to make the input path more purely raw, but that would change how Ctrl+C is handled, happy to adjust based on what behavior we want here.

OneBlue: So if we did that, that would cause wsladiag.exe to exit when `ctrl-c` is pressed, which is not the behavior that we want. We only want wsladig.exe to exit once the Linux process has exited 

Overall the shell behavior should be about the same as what we see in wsl.exe / wsl.exe --wsla

beena352:
Thanks! I’ll keep this PR focused on the TTY shell changes, and I’ll do the verb-based/per-command parsing refactor in a follow-up PR.

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -120,7 +120,10 @@ static int RunShellCommand(const std::wstring& sessionName, bool verbose)
     DWORD inMode = originalInMode;
     WI_SetAllFlags(inMode, ENABLE_WINDOW_INPUT | ENABLE_VIRTUAL_TERMINAL_INPUT);
     WI_ClearAllFlags(inMode, ENABLE_ECHO_INPUT | ENABLE_LINE_INPUT | ENABLE_INSERT_MODE);
-    WI_SetFlag(inMode, ENABLE_PROCESSED_INPUT);
+
+    // We clear ENABLE_PROCESSED_INPUT and install a Ctrl handler so Ctrl+C/Ctrl+Break
+    // are forwarded to the Linux TTY instead of terminating wsladiag, matching wsl.exe behavior.
+    WI_ClearFlag(inMode, ENABLE_PROCESSED_INPUT);
```
OneBlue: Let's include this flag in the above line like we do in WslClient.cpp today so it's a bit more readable

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -130,6 +133,13 @@ static int RunShellCommand(const std::wstring& sessionName, bool verbose)
     THROW_LAST_ERROR_IF(!SetConsoleOutputCP(CP_UTF8));
     THROW_LAST_ERROR_IF(!SetConsoleCP(CP_UTF8));
 
+    auto ctrlHandler = [](DWORD ctrlType) -> BOOL {
+        return (ctrlType == CTRL_C_EVENT || ctrlType == CTRL_BREAK_EVENT) ? TRUE : FALSE;
+    };
+
+    THROW_IF_WIN32_BOOL_FALSE(SetConsoleCtrlHandler(ctrlHandler, TRUE));
+    auto removeCtrlHandler = wil::scope_exit([&] { SetConsoleCtrlHandler(ctrlHandler, FALSE); });
```
OneBlue: What's the reason for adding this ?

beena352: This is to make sure wsladiag.exe itself doesn’t exit when Ctrl+C / Ctrl+Break is pressed. We want Ctrl+C to be forwarded to the Linux process through the TTY, and for wsladiag to only exit once the Linux process has exited, matching the behavior of wsl.exe.

src/windows/wsladiag/wsladiag.cpp
beena352: Sure.

src/windows/wsladiag/wsladiag.cpp
OneBlue: That surprises me because we don't set this in wsl.exe. 

If ctrl-C is hit while wsladiag.exe is running, does the callback actually get invoked ?

beena352: I added the handler initially to be safe around Ctrl+C behavior. After testing with ENABLE_PROCESSED_INPUT cleared and comparing with wsl.exe, the handler isn’t invoked and isn’t needed, so I’ve removed it.

beena352:
Tested locally :

- `wsladiag list` to verify session listing and column alignment
- `wsladiag shell <valid session>` to verify shell launch
- `wsladiag shell <invalid session>` to verify the "session not found" error
- `wsladiag shell <valid session> <extra args>` to verify usage is shown for invalid usage

All scenarios behaved as expected.

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -10,16 +10,235 @@ Module Name:
...
+
+    wil::com_ptr<IWSLAUserSession> userSession;
+    THROW_IF_FAILED(CoCreateInstance(__uuidof(WSLAUserSession), nullptr, CLSCTX_LOCAL_SERVER, IID_PPV_ARGS(&userSession)));
+    wsl::windows::common::security::ConfigureForCOMImpersonation(userSession.get());
+
+    wil::com_ptr<IWSLASession> session;
+    HRESULT hr = userSession->OpenSessionByName(sessionName.c_str(), &session);
+    if (FAILED(hr))
+    {
+        if (hr == HRESULT_FROM_WIN32(ERROR_NOT_FOUND) || hr == HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND) ||
```
OneBlue: `ERROR_NOT_FOUND` is the only error that we expect if the session doesn't exist so let's only handle that one. Any other error code would be unexpected

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -199,18 +199,42 @@ static int RunListCommand(bool /*verbose*/)
     if (sessions.size() == 0)
     {
         wslutil::PrintMessage(L"No WSLA sessions found.\n", stdout);
+        return 0;
     }
-    else
+
+    wslutil::PrintMessage(std::format(L"Found {} WSLA session{}:\n\n", sessions.size(), sessions.size() > 1 ? L"s" : L""), stdout);
```
OneBlue: `PrintMessage()` already prints a newline at the end of the message, so let's only have one here so we don't have 2 empty lines after this (applies in all calls to `PrintMessage()` in this file)

OneBlue (approved):
Change LGTM, thank you for the making the changes ! 

Looking through it, there are a couple followups that I'd recommend doing in future PR's: 

1) Splitting the argument parsing per verb (like we discussed on this PR)

2) Adding test coverage. Our tests are under `test/windows`.  [This file](https://github.com/microsoft/WSL/blob/master/test/windows/SimpleTests.cpp) contains good examples of the test cases we'd want for wsladiag.exe (essentially try out all the arguments, and check that the behavior & outputs are what we expect) 

2) Using localized messages 
      Our localized strings are stored in [localization/strings/en-US](https://github.com/microsoft/WSL/blob/master/localization/strings/en-US/Resources.resw) 

     Once added, they can be accessed via the `Localization` class, [like this for instance](https://github.com/microsoft/WSL/blob/77a35b9a0c2d1b527be502c33288a85c0bc0c46d/src/windows/common/WslClient.cpp#L1932C31-L1932C58) 


3) Using exception for error messages. To do that, we'd need to setup an `ExecutionContext`, like [this](https://github.com/microsoft/WSL/blob/77a35b9a0c2d1b527be502c33288a85c0bc0c46d/src/windows/common/WslClient.cpp#L1820), and we can then get the error message [like this](https://github.com/microsoft/WSL/blob/77a35b9a0c2d1b527be502c33288a85c0bc0c46d/src/windows/common/WslClient.cpp#L1905)

PR #13831, opened 2025-12-04
Add a utility Wait() method with timeout to WSLAProcessLauncher

ptrivedi:
## PR Checklist

### DIFF
--- a/src/windows/common/WSLAProcessLauncher.h
+++ b/src/windows/common/WSLAProcessLauncher.h
@@ -47,6 +47,7 @@ public:
     DEFAULT_MOVABLE(RunningWSLAProcess);
 
     ProcessResult WaitAndCaptureOutput(DWORD TimeoutMs = INFINITE, std::vector<std::unique_ptr<relay::OverlappedIOHandle>>&& ExtraHandles = {});
+    std::pair<int, bool> Wait(DWORD TimeoutMs = INFINITE);
     virtual wil::unique_handle GetStdHandle(int Index) = 0;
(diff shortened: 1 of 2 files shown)


OneBlue (approved):
LGTM

src/windows/common/WSLAProcessLauncher.h
```diff
@@ -47,6 +47,7 @@ class RunningWSLAProcess
     DEFAULT_MOVABLE(RunningWSLAProcess);
 
     ProcessResult WaitAndCaptureOutput(DWORD TimeoutMs = INFINITE, std::vector<std::unique_ptr<relay::OverlappedIOHandle>>&& ExtraHandles = {});
+    std::pair<int, bool> Wait(DWORD TimeoutMs = INFINITE);
```
yao-msft, line 50: nit: should we reuse ProcessResult for return?

btw, just curious, in which scenarios would this be preferred over WaitAndCaptureOutput?

OneBlue: We can't reuse it because we don't have the process output. 

This is useful in scenarios where we send a signal to a process that we don't own the IO handles (like the init process of a container when it's being terminated for instance)

yao-msft: Oh I see. The ProcessResult could still be reused I think, just the Output map will be empty. But I'm fine with either way.

OneBlue: My preference would be not to do that, because that could lead to scenarios where it could give the impression that the process returned no output.

PR #13832, opened 2025-12-05
Localization change from build: 135163088

OneBlue:


benhillis:
Not sure why these are being changed to English, if this doesn't resolve soon I will open a ticket with the localization team.

PR #13834, opened 2025-12-05
test: fix and issue with the test-setup.ps1 script that was preventing it from being run from some version of powershell

benhillis:
This change does two things:
1. Rewrote the certificate import logic to try -AllowUnsigned first, avoiding PowerShell module conflicts that were causing installation failures.
2. Changed Get-AppPackageLog -All to only show the last 10 entries from the past 5 minutes, making errors visible. Previously the entire log was flooding the output.

### DIFF
--- a/tools/test/test-setup.ps1
+++ b/tools/test/test-setup.ps1
@@ -71,23 +71,41 @@ if ($Package) {
     try {
         if ($AllowUnsigned)
         {
-            # unfortunately -AllowUnsigned isn't supported on vb so we need to manually import the certificate and trust it.
-            (Get-AuthenticodeSignature $Package).SignerCertificate | Export-Certificate -FilePath private-wsl.cert | Out-Null
-            try
-            {
-                Import-Certificate -FilePath .\private-wsl.cert -CertStoreLocation Cert:\LocalMachine\Root | Out-Null
+            # Try to add with -AllowUnsigned first (supported in newer PowerShell)
+            try {
+                Add-AppxPackage $Package -AllowUnsigned -ErrorAction Stop
             }
-            finally
-            {
-                Remove-Item -Path .\private-wsl.cert
+            catch {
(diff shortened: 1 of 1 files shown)


tools/test/test-setup.ps1
```diff
@@ -71,23 +71,45 @@ if ($Package) {
...
+                        try {
+                            Import-Certificate -FilePath $certPath -CertStoreLocation Cert:\LocalMachine\Root | Out-Null
+                        }
+                        finally {
+                            Remove-Item -Path $certPath -ErrorAction SilentlyContinue
+                        }
+                    }
+                }
+                catch {
+                    Write-Warning "Could not import certificate: $_"
```
OneBlue: Should we just throw that exception to fail the script here ?

tools/test/test-setup.ps1
```diff
@@ -71,23 +71,45 @@ if ($Package) {
...
             }
-            finally
-            {
-                Remove-Item -Path .\private-wsl.cert
+            catch {
+                # Fallback: manually import the certificate and trust it
+                try {
+                    $cert = (Get-AuthenticodeSignature -LiteralPath $Package).SignerCertificate
+                    if ($cert) {
+                        $certPath = Join-Path $env:TEMP "private-wsl-$([guid]::NewGuid()).cert"
```
OneBlue: What's the reason for having a GUID in the certificate file name ?

OneBlue (commented):
Out of curiosity, what's the context on the module conflict ?

tools/test/test-setup.ps1
benhillis: Yeah good call, let me figure out a better way to solve this.

benhillis:
> Out of curiosity, what's the context on the module conflict ?

I've had issues running the test script from the powershell prompt in vscode. This made the issue go away.

OneBlue (approved):

PR #13835, opened 2025-12-05
Add back WSL2_VM_ID environment variable to the system distro.

benhillis:
This change adds back the VM ID environment variable to the system distro (WSLg runtime environment). Turns out weston has a dependency on this environment variable: https://github.com/microsoft/weston-mirror/blob/2318fecaeac1f1a2d5a7a042c34d931c71dae04c/rdprail-shell/shell.h#L56

### DIFF
--- a/src/linux/init/init.cpp
+++ b/src/linux/init/init.cpp
@@ -2258,7 +2258,16 @@ Return Value:
     if (Value != nullptr)
     {
         Config.VmId = Value;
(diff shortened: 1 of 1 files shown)


hideyukn88 (approved):

OneBlue (approved):
LGTM. We should also consider adding test coverage for the wslg shortcut creation logic (can be done in  followup PR though)

PR #13836, opened 2025-12-05
Clean up localhost relay implementation to not rely on procfs parsing.

benhillis:
This change vastly simplifies the logic around localhost relay to not manually parse the /proc/net/tcp files. This has better performance and allows us to use existing functionality that is used for mirrored networking.

### DIFF
--- a/src/linux/init/localhost.cpp
+++ b/src/linux/init/localhost.cpp
@@ -12,6 +12,8 @@
 #include <netinet/ip.h>
 #include <sys/syscall.h>
 #include <linux/unistd.h>
+#include <linux/sock_diag.h>
+#include <linux/inet_diag.h>
 #include <lxwil.h>
 #include <linux/if_tun.h>
 
@@ -21,6 +23,8 @@
 #include "SecCompDispatcher.h"
 #include "seccomp_defs.h"
 #include "CommandLine.h"
+#include "NetlinkChannel.h"
+#include "NetlinkTransactionError.h"
 
 #define TCP_LISTEN 10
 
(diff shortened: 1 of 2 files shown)


src/linux/init/localhost.cpp
```diff
@@ -145,80 +149,60 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
+            auto onMessage = [&](const NetlinkResponse& response) {
+                for (const auto& e : response.Messages<inet_diag_msg>(SOCK_DIAG_BY_FAMILY))
+                {
+                    const auto* payload = e.Payload();
+                    sockaddr_storage sock{};
+                    auto* ipv6 = reinterpret_cast<sockaddr_in6*>(&sock);
+                    ipv6->sin6_family = AF_INET6;
+                    static_assert(sizeof(ipv6->sin6_addr.s6_addr32) == sizeof(payload->id.idiag_src));
+                    memcpy(ipv6->sin6_addr.s6_addr32, payload->id.idiag_src, sizeof(ipv6->sin6_addr.s6_addr32));
+                    ipv6->sin6_port = ntohs(payload->id.idiag_sport);
```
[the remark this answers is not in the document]
benhillis: this seems right but I will also need to update the windows side of the code, taking a look.

src/linux/init/localhost.cpp
```diff
@@ -145,80 +149,60 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
-            sockets.emplace_back(sock);
+            message.sdiag_family = AF_INET6;
+            auto transaction = channel.CreateTransaction(message, SOCK_DIAG_BY_FAMILY, NLM_F_DUMP);
+            transaction.Execute(onMessage);
         }
     }
+    catch (const NetlinkTransactionError& e)
+    {
+        // Log but don't fail - network state might be temporarily unavailable
+        LOG_ERROR("Failed to query listening sockets via sock_diag: {}", e.what());
```
OneBlue, line 204: I have a distant memory that netlink will occasionally return EAGAIN. We should probably not log an error if we get that specific error so we don't flood dmesg

benhillis: Good suggestion.

benhillis: I have run this a bunch and haven't seen any EAGAIN errors.

OneBlue: Well I can't find any reference to that error in the code, so I'm probably misremembering here

OneBlue (approved):

PR #13843, opened 2025-12-06
cleanup: refactor svccommio class

benhillis:
This helper class has always been a bit of a mess, this change reimplements the class using RAII classes and reduces complexity and code length.

### DIFF
--- a/src/windows/common/svccommio.cpp
+++ b/src/windows/common/svccommio.cpp
@@ -138,11 +138,8 @@ ConsoleOutput::ConsoleOutput(wil::unique_hfile ConsoleHandle, DWORD SavedMode) :
 
 ConsoleOutput::~ConsoleOutput()
 {
-    if (m_ConsoleHandle)
-    {
-        TrySetConsoleMode(m_ConsoleHandle.get(), m_SavedMode);
-        LOG_IF_WIN32_BOOL_FALSE(SetConsoleOutputCP(m_SavedCodePage));
-    }
+    TrySetConsoleMode(m_ConsoleHandle.get(), m_SavedMode);
+    LOG_IF_WIN32_BOOL_FALSE(SetConsoleOutputCP(m_SavedCodePage));
 }
 
 SvcCommIo::SvcCommIo()
--- a/src/windows/common/svccommio.cpp
+++ b/src/windows/common/svccommio.cpp
@@ -25,46 +25,76 @@ bool IsConsoleHandle(_In_ HANDLE Handle)
     return GetFileType(Handle) == FILE_TYPE_CHAR && GetConsoleMode(Handle, &Mode);
 }
 
-void TrySetConsoleMode(_In_ HANDLE Handle, _In_ DWORD Mode)
+void ChangeConsoleMode(_In_ HANDLE Handle, _In_ DWORD Mode)
 {
+    //
+    // Use the invalid parameter error code to detect the v1 console that does
+    // not support the provided mode. This can be improved in the future when
+    // a more elegant solution exists.
+    //
+    // N.B. Ignore failures setting the mode if the console has already
+    //      disconnected.
+    //
+
     if (!SetConsoleMode(Handle, Mode))
     {
-        const auto Error = GetLastError();
-        if (Error != ERROR_INVALID_PARAMETER && Error != ERROR_PIPE_NOT_CONNECTED)
+        // DISABLE_NEWLINE_AUTO_RETURN is not supported everywhere, if the flag was present fall back and try again.
+        if (WI_IsFlagSet(Mode, DISABLE_NEWLINE_AUTO_RETURN))
         {
-            LOG_IF_WIN32_ERROR(Error);
+            if (SetConsoleMode(Handle, WI_ClearFlag(Mode, DISABLE_NEWLINE_AUTO_RETURN)))
+            {
+                return;
+            }
+        }
+
+        switch (GetLastError())
+        {
+        case ERROR_INVALID_PARAMETER:
+            THROW_HR(WSL_E_CONSOLE);
+
+        case ERROR_PIPE_NOT_CONNECTED:
+            break;
+
+        default:
+            THROW_LAST_ERROR();
         }
     }
 }
+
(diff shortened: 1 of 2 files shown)


src/windows/common/svccommio.cpp
```diff
@@ -18,265 +18,165 @@ Module Name:
...
+        return nullptr;
     }
 
-    if (_stdHandlesInfo.IsConsoleOutput)
+    DWORD Mode;
+    if (GetConsoleMode(ConsoleHandle.get(), &Mode))
     {
-        _stdHandles.StdOut.Handle = LXSS_HANDLE_USE_CONSOLE;
-        _stdHandles.StdOut.HandleType = LxssHandleConsole;
+        return std::unique_ptr<ConsoleOutput>(new ConsoleOutput(std::move(ConsoleHandle), Mode));
```
OneBlue: nit: I'd have a preference towards either having `ConsoleOutput` be default constructible / movable, or using an std::optional in `SvcCommIo` so we can avoid having to allocate a `std::unique_ptr` here

src/windows/common/svccommio.cpp
```diff
@@ -18,265 +18,165 @@ Module Name:
...
-        NewConsoleInputCP = CP_UTF8;
-        THROW_LAST_ERROR_IF(!::SetConsoleCP(NewConsoleInputCP));
-    }
-
-    bool RestoreMode = false;
-    bool RestoreCp = false;
-    auto RestoreOutput = wil::scope_exit([&] {
-        if (RestoreMode)
+        const auto Error = GetLastError();
+        if (Error != ERROR_INVALID_PARAMETER && Error != ERROR_PIPE_NOT_CONNECTED)
```
OneBlue: Looks like currently throw `WSL_E_CONSOLE` if we hit this, what's the reason for the behavior change for that error code  ?

jobayer110 (commented):
H

src/windows/common/svccommio.cpp
```diff
@@ -45,238 +61,142 @@ void ChangeConsoleMode(_In_ HANDLE File, _In_ DWORD ConsoleMode)
...
+    // Set UTF-8 code page.
+    LOG_IF_WIN32_BOOL_FALSE(SetConsoleCP(CP_UTF8));
+}
 
-    IsConsoleError = FALSE;
-    if ((GetFileType(StdHandlesInfo.ErrorHandle) == FILE_TYPE_CHAR) &&
-        (GetConsoleScreenBufferInfo(StdHandlesInfo.ErrorHandle, &ScreenBufferInfo)))
+ConsoleInput::~ConsoleInput()
+{
+    if (m_Handle)
```
OneBlue: Is it actually possible for this handle to be unset ? If not, I'd remove the if because it gives the impression that his class is moveable, but it actually isn't (same for ConsoleOutput)

src/windows/common/svccommio.cpp
```diff
@@ -45,238 +61,139 @@ void ChangeConsoleMode(_In_ HANDLE File, _In_ DWORD ConsoleMode)
...
 }
 
-COORD
-wsl::windows::common::SvcCommIo::GetWindowSize() const
+ConsoleOutput::~ConsoleOutput()
 {
-    CONSOLE_SCREEN_BUFFER_INFOEX Info{};
-    Info.cbSize = sizeof(Info);
-    if (_stdHandlesInfo.IsConsoleOutput)
+    if (m_ConsoleHandle)
```
OneBlue: nit: this if can also be removed

OneBlue (approved):

jobayer110 (commented):
###

jobayer110:
H

src/windows/common/svccommio.cpp
```diff
@@ -45,238 +61,136 @@ void ChangeConsoleMode(_In_ HANDLE File, _In_ DWORD ConsoleMode)
...
-    if (_stdHandlesInfo.IsConsoleError)
-    {
-        _stdHandles.StdErr.Handle = LXSS_HANDLE_USE_CONSOLE;
-        _stdHandles.StdErr.HandleType = LxssHandleConsole;
-    }
+    return std::nullopt;
 }
 
-wsl::windows::common::SvcCommIo::~SvcCommIo()
+ConsoleOutput::ConsoleOutput(wil::unique_hfile ConsoleHandle, DWORD SavedMode) :
```
OneBlue: I'm surprised that this works since `unique_hfile` isn't copyable. (I'd expect something like `wil::unique_hfile&& ConsoleHandle)` here

src/windows/common/svccommio.cpp
```diff
@@ -45,238 +61,136 @@ void ChangeConsoleMode(_In_ HANDLE File, _In_ DWORD ConsoleMode)
...
+            static_cast<short>(Info.srWindow.Right - Info.srWindow.Left + 1),
+            static_cast<short>(Info.srWindow.Bottom - Info.srWindow.Top + 1)};
     }
 
-    if (_stdHandlesInfo.IsConsoleInput != FALSE)
-    {
-        ChangeConsoleMode(_stdHandlesInfo.InputHandle, _stdHandlesInfo.SavedInputMode);
-        SetConsoleCP(_stdHandlesInfo.SavedInputCP);
-    }
+    return {80, 24}; // Default size if no console
```
OneBlue: nit: I think we should fail if we hit this, since that would mean that we try to access the console size when there's none.

src/windows/common/svccommio.hpp
```diff
@@ -14,39 +14,66 @@ Module Name:
...
+    UINT m_SavedCodePage = 0;
+};
 
-namespace wsl::windows::common {
 class SvcCommIo
 {
 public:
     SvcCommIo();
-    ~SvcCommIo();
+    ~SvcCommIo() = default;
```
OneBlue: nit: this can be removed

src/windows/common/svccommio.cpp
benhillis: I worry about introducing a new failure path, I'll add a log.

OneBlue (approved):

PR #13844, opened 2025-12-06
Use nerdctl events to track container state changes 

OneBlue:
## PR Checklist

### DIFF
--- a/src/windows/wslaservice/exe/ContainerEventTracker.cpp
+++ b/src/windows/wslaservice/exe/ContainerEventTracker.cpp
@@ -42,6 +42,7 @@ void ContainerEventTracker::ContainerTrackingReference::Reset()
     {
         m_tracker->UnregisterContainerStateUpdates(m_id);
         m_tracker = nullptr;
+        m_id = {};
     }
 }
 
@@ -52,7 +53,7 @@ ContainerEventTracker::ContainerTrackingReference::~ContainerTrackingReference()
 
 ContainerEventTracker::ContainerEventTracker(WSLAVirtualMachine& virtualMachine)
 {
-    ServiceProcessLauncher launcher{"/usr/bin/nerdctl", {"/usr/bin/nerdctl", "events", "--format", "{{json .}}"}, {}, common::ProcessFlags::Stdout};
+    ServiceProcessLauncher launcher{nerdctlPath, {nerdctlPath, "events", "--format", "{{json .}}"}, {}, common::ProcessFlags::Stdout};
 
     // Redirect stderr to /dev/null to avoid pipe deadlocks.
     launcher.AddFd({.Fd = 2, .Type = WSLAFdTypeLinuxFileOutput, .Path = "/dev/null"});
--- a/src/windows/wslaservice/exe/ContainerEventTracker.cpp
+++ b/src/windows/wslaservice/exe/ContainerEventTracker.cpp
@@ -24,11 +24,6 @@ ContainerEventTracker::ContainerTrackingReference::ContainerTrackingReference(Co
 {
 }
 
-ContainerEventTracker::ContainerTrackingReference::ContainerTrackingReference(ContainerEventTracker::ContainerTrackingReference&& other)
-{
-    (*this) = std::move(other);
-}
-
 ContainerEventTracker::ContainerTrackingReference& ContainerEventTracker::ContainerTrackingReference::operator=(ContainerEventTracker::ContainerTrackingReference&& other)
 {
     Reset();
@@ -36,6 +31,7 @@ ContainerEventTracker::ContainerTrackingReference& ContainerEventTracker::Contai
     m_tracker = other.m_tracker;
 
     other.m_tracker = nullptr;
+    other.m_id = {};
 
     return *this;
 }
--- a/src/windows/wslaservice/exe/ContainerEventTracker.cpp
+++ b/src/windows/wslaservice/exe/ContainerEventTracker.cpp
@@ -1,3 +1,16 @@
+/*++
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    ContainerEventTracker.cpp
+
+Abstract:
+
+    Contains the implementation of ContainerEventTracker.
+
+--*/
 #include "precomp.h"
 #include "ContainerEventTracker.h"
 #include "WSLAVirtualMachine.h"
--- a/src/windows/wslaservice/exe/ContainerEventTracker.cpp
+++ b/src/windows/wslaservice/exe/ContainerEventTracker.cpp
@@ -1,0 +1,185 @@
+#include "precomp.h"
+#include "ContainerEventTracker.h"
+#include "WSLAVirtualMachine.h"
+#include <nlohmann/json.hpp>
+
+using wsl::windows::service::wsla::ContainerEventTracker;
+using wsl::windows::service::wsla::WSLAVirtualMachine;
+
+ContainerEventTracker::ContainerTrackingReference::ContainerTrackingReference(ContainerEventTracker* tracker, size_t id) :
+    m_tracker(tracker), m_id(id)
+{
+}
+
+ContainerEventTracker::ContainerTrackingReference::ContainerTrackingReference(ContainerEventTracker::ContainerTrackingReference&& other)
+{
+    (*this) = std::move(other);
+}
+
+ContainerEventTracker::ContainerTrackingReference& ContainerEventTracker::ContainerTrackingReference::operator=(ContainerEventTracker::ContainerTrackingReference&& other)
+{
+    m_id = other.m_id;
+    m_tracker = other.m_tracker;
+
+    other.m_tracker = nullptr;
+
+    return *this;
+}
+
+void ContainerEventTracker::ContainerTrackingReference::Reset()
+{
+    if (m_tracker != nullptr)
+    {
+        m_tracker->UnregisterContainerStateUpdates(m_tracker != nullptr);
+        m_tracker = nullptr;
+    }
+}
+
+ContainerEventTracker::ContainerTrackingReference::~ContainerTrackingReference()
+{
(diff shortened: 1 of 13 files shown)


src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -26,46 +26,76 @@ static std::vector<std::string> defaultNerdctlRunArgs{//"--pull=never", // TODO:
...
 
-HRESULT WSLAContainer::Start()
-{
-    return E_NOTIMPL;
+    WSL_LOG(
+        "ContainerEvent",
+        TraceLoggingValue(m_name.c_str(), "Name"),
+        TraceLoggingValue(m_id.c_str(), "Id"),
+        TraceLoggingValue((int)event, "Event"));
 }
```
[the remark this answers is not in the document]
OneBlue: `m_startedEvent` is immutable so this isn't an issue. Acquiring the lock in OnEvent() could cause deadlocks

src/windows/wslaservice/exe/ContainerEventTracker.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+using wsl::windows::service::wsla::ContainerEventTracker;
+using wsl::windows::service::wsla::WSLAVirtualMachine;
+
+ContainerEventTracker::ContainerTrackingReference::ContainerTrackingReference(ContainerEventTracker* tracker, size_t id) :
+    m_tracker(tracker), m_id(id)
+{
+}
+
+ContainerEventTracker::ContainerTrackingReference::ContainerTrackingReference(ContainerEventTracker::ContainerTrackingReference&& other)
```
yao-msft: nit: = default ?

src/windows/wslaservice/exe/ContainerEventTracker.cpp
```diff
@@ -0,0 +1,194 @@
...
+    (*this) = std::move(other);
+}
+
+ContainerEventTracker::ContainerTrackingReference& ContainerEventTracker::ContainerTrackingReference::operator=(ContainerEventTracker::ContainerTrackingReference&& other)
+{
+    Reset();
+    m_id = other.m_id;
+    m_tracker = other.m_tracker;
+
+    other.m_tracker = nullptr;
```
yao-msft, line 33: Do we need to set other.m_id to some other value?

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -26,46 +26,79 @@ static std::vector<std::string> defaultNerdctlRunArgs{//"--pull=never", // TODO:
...
+    wait.AddHandle(std::make_unique<common::relay::EventHandle>(m_startedEvent.get(), [&]() { wait.Cancel(); }));
+    wait.Run({});
 
-        status = GetNerdctlStatus();
+    if (!m_startedEvent.is_signaled())
+    {
+        auto status = GetNerdctlStatus();
+
+        THROW_HR_IF_MSG(
+            E_FAIL,
```
yao-msft, line 77: nit: should we use custom failure error code

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -166,19 +210,25 @@ Microsoft::WRL::ComPtr<WSLAContainer> WSLAContainer::Create(const WSLA_CONTAINER
...
     {
-        launcher.AddFd(containerOptions.InitProcessOptions.Fds[i]);
+        id.pop_back();
     }
 
-    return wil::MakeOrThrow<WSLAContainer>(&parentVM, launcher.Launch(parentVM), containerOptions.Name, containerOptions.Image);
+    return wil::MakeOrThrow<WSLAContainer>(&parentVM, containerOptions, std::move(id), eventTracker);
 }
 
 std::vector<std::string> WSLAContainer::PrepareNerdctlRunCommand(const WSLA_CONTAINER_OPTIONS& options, std::vector<std::string>&& inputOptions)
```
yao-msft: nit:
```suggestion
std::vector<std::string> WSLAContainer::PrepareNerdctlCreateCommand(const WSLA_CONTAINER_OPTIONS& options, std::vector<std::string>&& inputOptions)
```

yao-msft (approved):

src/windows/wslaservice/exe/ContainerEventTracker.cpp
OneBlue: We don't necessarily need to since we use `m_tracker` to determine if we need to release the callback, but added logic to reset `m_id` for consistency

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: Agreed in principle, although right now we don't have the capacity to do that since we have no simple way to know why nerdctl actually failed. 

This is something that we should study in a future change though, maybe by adding logic to detect what caused the nerdctl failure

src/windows/wslaservice/exe/ContainerEventTracker.cpp
```diff
@@ -0,0 +1,190 @@
...
+    return *this;
+}
+
+void ContainerEventTracker::ContainerTrackingReference::Reset()
+{
+    if (m_tracker != nullptr)
+    {
+        m_tracker->UnregisterContainerStateUpdates(m_id);
+        m_tracker = nullptr;
+    }
```
ptrivedi, line 46: reset m_id here as well

src/windows/wslaservice/exe/ContainerEventTracker.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+ContainerEventTracker::ContainerTrackingReference::~ContainerTrackingReference()
+{
+    Reset();
+}
+
+ContainerEventTracker::ContainerEventTracker(WSLAVirtualMachine& virtualMachine)
+{
+    ServiceProcessLauncher launcher{"/usr/bin/nerdctl", {"/usr/bin/nerdctl", "events", "--format", "{{json .}}"}, {}, common::ProcessFlags::Stdout};
```
ptrivedi: nit: declare nerdctlPath somewhere and use

src/windows/wslaservice/exe/ContainerEventTracker.cpp
```diff
@@ -0,0 +1,190 @@
...
+                pendingBuffer.insert(pendingBuffer.end(), begin, end);
+
+                if (!pendingBuffer.empty()) // nerdctl inserts empty lines between events, skip those.
+                {
+                    OnEvent(pendingBuffer);
+                }
+
+                pendingBuffer.clear();
+
+                begin = end + 1;
```
ptrivedi, line 151: Haven't looked very closely, but something to check.. Could we get into a situation where begin is beyond end and we call pendingBuffer.insert with that?

OneBlue: We can only be in this loop if `end` is not at the end of the vector, so `end + 1` cannot be passed the vector's end.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -26,46 +26,79 @@ static std::vector<std::string> defaultNerdctlRunArgs{//"--pull=never", // TODO:
...
+    return m_image;
+}
+
+void WSLAContainer::Start(const WSLA_CONTAINER_OPTIONS& Options)
+{
+    std::lock_guard lock{m_lock};
+
+    THROW_HR_IF_MSG(
+        HRESULT_FROM_WIN32(ERROR_INVALID_STATE),
+        m_state != WslaContainerStateCreated,
```
ptrivedi, line 51: You can call start on an Exited container

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -100,25 +133,31 @@ WSLA_CONTAINER_STATE WSLAContainer::State() noexcept
...
 
     return S_OK;
 }
+CATCH_RETURN();
 
 HRESULT WSLAContainer::GetInitProcess(IWSLAProcess** Process)
 try
 {
-    return m_containerProcess.Get().QueryInterface(__uuidof(IWSLAProcess), (void**)Process);
+    std::lock_guard lock{m_lock};
```
ptrivedi, line 155: This is now a recursive_mutex, merge should bring that in

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: True, but that method isn't meant by called by users, it's only called as part of Create(). 

We could add support for restarting stopped containers, but that would required an API change so it's probably a bigger discussion

test/windows/WSLATests.cpp
```diff
@@ -1193,7 +1201,7 @@ class WSLATests
 
             // Create a stuck container.
             WSLAContainerLauncher launcher(
-                "debian:latest", "test-container-1", "/bin/cat", {}, {}, ProcessFlags::Stdin | ProcessFlags::Stdout | ProcessFlags::Stderr);
+                "debian:latest", "test-container-1", "sleep", {"sleep", "99999"}, {}, ProcessFlags::Stdout | ProcessFlags::Stderr);
```
ptrivedi, line 1205: Should also add tests for Start() being called on Exited containers

OneBlue: Calling Start() on a an exited container isn't currently supported

src/windows/wslaservice/exe/WSLAContainer.cpp
ptrivedi: I see. Yes, we could discuss API changes in that case. For completeness sake, there's start and restart. Start can be called on Created or Exited containers, Restart on Running containers

ptrivedi (approved):

PR #13851, opened 2025-12-08
Localization change from build: 135341677

OneBlue:


benhillis:
I have filed an issue with our localization team.

PR #13857, opened 2025-12-09
Implement Stop Container

ptrivedi:


src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -69,9 +69,40 @@ HRESULT WSLAContainer::Start()
...
+
+    /* 'nerdctl stop ...'
+     *   returns success and <containerId> on stdout if the container is running or already stopped
+     *   returns error "No such container: <containerId>" on stderr if the container is in 'Created' state or does not exist
+     *
+     * For our case, we consider stopping a container that is not running or does not exist as a no-op and return success.
+     * TODO: Discuss and return stdout/stderr or corresponding HRESULT from nerdctl stop for better diagnostics.
+     */
+
+    if (m_state == WslaContainerStateExited)
```
OneBlue: We should call `State()` here so the state is updated if the init process has returned

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -69,9 +69,40 @@ HRESULT WSLAContainer::Start()
...
+        m_state != WslaContainerStateRunning,
+        "No such running or exited container '%hs', state: %i",
+        m_name.c_str(),
+        m_state);
+    ServiceProcessLauncher launcher(nerdctlPath, {nerdctlPath, "stop", m_name});
+    // TODO: Figure out how we want to handle custom signals and timeout values.
+    // nerdctl stop has a --time and a --signal option that can be used
+    // By default, it uses SIGTERM and a default timeout of 10 seconds.
+    auto result = launcher.Launch(*m_parentVM).Wait();
+    THROW_HR_IF(E_FAIL, result.first != 0);
```
OneBlue: We should log the command output when this fails to make debugging easier: 

```
    THROW_HR_IF_MSG(E_FAIL, result.Code != 0, "%hs", launcher.FormatResult(result).c_str());
```

test/windows/WSLATests.cpp
```diff
@@ -1274,9 +1303,16 @@ class WSLATests
                 {{"exited-container", "debian:latest", WslaContainerStateExited},
                  {"test-unique-name", "debian:latest", WslaContainerStateExited}});
 
+            // Verify that calling Stop() on exited containers is a no-op and state remains as WslaContainerStateExited.
+            VERIFY_SUCCEEDED(container.Get().Stop(15, 1000));
```
OneBlue: This is not wired yet, but I'd recommend putting a longer timeout so the tests don't fail if the test VM is slow

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -69,9 +69,40 @@ HRESULT WSLAContainer::Start()
...
+        HRESULT_FROM_WIN32(ERROR_INVALID_STATE),
+        m_state != WslaContainerStateRunning,
+        "No such running or exited container '%hs', state: %i",
+        m_name.c_str(),
+        m_state);
+    ServiceProcessLauncher launcher(nerdctlPath, {nerdctlPath, "stop", m_name});
+    // TODO: Figure out how we want to handle custom signals and timeout values.
+    // nerdctl stop has a --time and a --signal option that can be used
+    // By default, it uses SIGTERM and a default timeout of 10 seconds.
+    auto result = launcher.Launch(*m_parentVM).Wait();
```
OneBlue: We should pass `TimeoutMs` to Wait() here so we fail if nerdctl doesn't return within the timeout

ptrivedi: Yes, of course, great catch! Typo from changing around between passing time and signal nerdctl args, process signals, etc. Pushing the correct change.

src/windows/wslaservice/exe/WSLAContainer.cpp
ptrivedi: Can't use current FormatResult directly because result is not ProcessResult. Will fix though

src/windows/wslaservice/exe/WSLAContainer.cpp
ptrivedi: Can't call here, it locks m_lock. Will move it before lock

OneBlue: True, the easiest fix would be to switch `m_lock` into a `std::recursive_mutex`

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: Oh thank you. I didn't actually catch this: We need to WaitAndCaptureOutput() here (otherwise we might deadlock if nerdctl writes on its stdio fds). That would also allow us to actually log the errors here

src/windows/wslaservice/exe/WSLAContainer.cpp
ptrivedi: Easiest, maybe, but not the most efficient. recursive_mutexes typically have more overhead. I've added a StateNoLock(), or the other option would be to pass a bool into State() to indicate if lock is already held. I don't have a strong preference for either of the latter approaches, but would like to avoid recursive_mutex if possible

OneBlue (approved):

PR #13860, opened 2025-12-09
Add tests for WSLA session lookup and listing

beena352:
This PR adds initial automated test coverage for WSLA session operations.
Specifically, it adds a new test that verifies:

Sessions created through CreateSession are discoverable via OpenSessionByName

The correct session is returned when matching by DisplayName

OpenSessionByName returns ERROR_NOT_FOUND when the name does not exist

These tests use the existing WSLA testing infrastructure (GetDefaultSessionSettings, CreateSession, WSL2_TEST_ONLY) and follow the same style and patterns as the current WSLA test suite.

### DIFF
--- a/test/windows/WSLATests.cpp
+++ b/test/windows/WSLATests.cpp
@@ -177,6 +177,60 @@ class WSLATests
         }
     }
 
+    TEST_METHOD(ListSessionsReturnsSessionWithDisplayName)
+    {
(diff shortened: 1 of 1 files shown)


OneBlue (approved):
LGTM. Long term we'll probably to add tests that run `wsladbg.exe` and validate its output, but this is great start.

PR #13861, opened 2025-12-09
cleanup: remove stale version of wsl.wprp

benhillis:
noticed we have two versions of the trace profile, this removes the old one.


jobayer1V (commented):
110230

OneBlue (approved):

PR #13862, opened 2025-12-10
Implement WSLAContainer::Exec()

OneBlue:

This change implements WSLAContainer::Exec() and adds test coverage for it.

## PR Checklist

### DIFF
--- a/src/windows/common/WSLAProcessLauncher.cpp
+++ b/src/windows/common/WSLAProcessLauncher.cpp
@@ -164,18 +164,6 @@ RunningWSLAProcess::ProcessResult RunningWSLAProcess::WaitAndCaptureOutput(DWORD
     return result;
 }
 
-ClientRunningWSLAProcess WSLAProcessLauncher::Launch(IWSLASession& Session)
-{
-    auto [hresult, error, process] = LaunchNoThrow(Session);
-    if (FAILED(hresult))
-    {
-        auto commandLine = wsl::shared::string::Join(m_arguments, ' ');
-        THROW_HR_MSG(hresult, "Failed to launch process: %hs (commandline: %hs). Errno = %i", m_executable.c_str(), commandLine.c_str(), error);
-    }
-
-    return std::move(process.value());
-}
-
 std::tuple<HRESULT, int, std::optional<ClientRunningWSLAProcess>> WSLAProcessLauncher::LaunchNoThrow(IWSLASession& Session)
 {
     auto [options, commandLine, env] = CreateProcessOptions();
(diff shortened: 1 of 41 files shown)


src/windows/common/relay.hpp
benhillis, line 193: Are Handle and OwnedHandle mutually exclusive?

OneBlue: HANDLE is always set, OwnedHandle is only set when the event is owned by that class

benhillis (approved):

src/windows/common/WSLAProcessLauncher.cpp
```diff
@@ -169,25 +169,31 @@ RunningWSLAProcess::ProcessResult RunningWSLAProcess::WaitAndCaptureOutput(DWORD
...
     {
-        auto commandLine = wsl::shared::string::Join(m_arguments, ' ');
-        THROW_HR_MSG(hresult, "Failed to launch process: %hs (commandline: %hs). Errno = %i", m_executable.c_str(), commandLine.c_str(), error);
+        return std::make_tuple(result, error, std::optional<ClientRunningWSLAProcess>());
     }
 
-    return std::move(process.value());
+    wsl::windows::common::security::ConfigureForCOMImpersonation(process.get());
+
+    return {S_OK, 0, ClientRunningWSLAProcess{std::move(process), std::move(m_fds)}};
```
ptrivedi, line 186: Should just return what CreateRootNamespaceProcess returned for result and error instead of overriding

OneBlue: `CreateRootNamespaceProcess()` returns an HRESULT and a ` IWSLAProcess`, which doesn't match what we want to return here

ptrivedi: I meant for result and error instead of masking it. Similar to line 181

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -196,35 +197,93 @@ CATCH_RETURN();
 HRESULT WSLAContainer::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess** Process, int* Errno)
 try
 {
-    // auto process = wil::MakeOrThrow<WSLAProcess>();
+    *Errno = -1;
+    THROW_HR_IF_MSG(E_INVALIDARG, Options->Executable != nullptr, "Executable must be null");
+
+    std::lock_guard lock{m_lock};
```
ptrivedi, line 203: Is this the C++17 thing of auto-deducing template args? Not seeing the `std::lock_guard<std::recursive_mutex> lock{m_lock};` here

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -196,35 +197,93 @@ CATCH_RETURN();
...
     }
 
+    return {hasStdin, hasTty};
+}
+
+void WSLAContainer::AddEnvironmentVariables(std::vector<std::string>& args, const WSLA_PROCESS_OPTIONS& options)
+{
+    for (ULONG i = 0; i < options.EnvironmentCount; i++)
+    {
+        THROW_HR_IF_MSG(E_INVALIDARG, options.Environment[i][0] == '-', "Invalid environment string: %hs", options.Environment[i]);
```
ptrivedi, line 276: Is it safe to assume that options.Environment has been sanity checked by the time we get here that options.Environment[i][0] access cannot be a cause of any crash?

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: Yes, C++ can auto deduct template arguments (since C++ 11 IIRC)

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: Yes, COM guarantees that there are at least `EnvironmentCount` entries in `Environment`. It would still be good to add test coverage for it though, I can do that in a future PR

src/windows/common/WSLAProcessLauncher.cpp
OneBlue: Ah I see. Yes I'd be OK with either, this should be equivalent

PR #13871, opened 2025-12-10
Initial Image pull, load and import support

yao-msft:
## PR Checklist


src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -19,6 +19,36 @@ Module Name:
 #include "ServiceProcessLauncher.h"
 #include "WslCoreFilesystem.h"
 
+namespace
+{
+    constexpr const char* nerdctlPath = "/usr/bin/nerdctl";
+
+    HRESULT TransferFileContent(HANDLE from, HANDLE to)
```
OneBlue: FYI we already have a helper to do this: https://github.com/microsoft/WSL/blob/25572cc168b5a54e18a63104d5cebc478fd6677d/src/windows/common/relay.hpp#L39 

Also whenever we call it we should make sure to pass the VM termination event ass `ExitHandle` so the relay is interrupted when vm shuts down

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -213,15 +228,67 @@ void WSLASession::CopyDisplayName(_Out_writes_z_(bufferLength) PWSTR buffer, siz
...
+    return S_OK;
 }
+CATCH_RETURN();
 
-HRESULT WSLASession::ImportImage(ULONG Handle, LPCWSTR Image, IProgressCallback* ProgressCallback)
+HRESULT WSLASession::LoadImage(ULONG ImageHandle, IProgressCallback* ProgressCallback)
+try
 {
-    return E_NOTIMPL;
+    HANDLE imageFileHandle = ULongToHandle(ImageHandle);
```
OneBlue: The handle needs to opened in the current process to read from it. 

We can do that with: 

```
auto localHandle = wsl::windows::common::wslutil::DuplicateHandleFromCallingProcess(ImageHandle);
```

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -213,15 +228,67 @@ void WSLASession::CopyDisplayName(_Out_writes_z_(bufferLength) PWSTR buffer, siz
...
+    RETURN_HR_IF(E_INVALIDARG, INVALID_HANDLE_VALUE == imageFileHandle);
+
+    // Directly invoking "nerdctl load" will immediately return with failure
+    // "stdin is empty and input flag is not specified".
+    // TODO: Change the workaround when nerdctl has a fix.
+    ServiceProcessLauncher launcher{"/bin/sh", {"/bin/sh", "-c", "cat | /usr/bin/nerdctl load"}, {}, ProcessFlags::Stdin | ProcessFlags::Stdout | ProcessFlags::Stderr};
+    auto loadProcess = launcher.Launch(*m_virtualMachine.Get());
+
+    auto loadProcessStdin = loadProcess.GetStdHandle(0);
+    RETURN_IF_FAILED(TransferFileContent(imageFileHandle, loadProcessStdin.get()));
```
OneBlue: We should also consume nerctl's stdout & stderr output while we write its stdin or we could end up in a pipe deadlock. 

The nicest way to do this would be to use an `ExtraHandle`, like [this example](https://github.com/microsoft/WSL/blob/14875a2d09b8bcbb38c96f109e28468afdfc8b54/test/windows/WSLATests.cpp#L914C11-L923C91) 

We'll probably need to create a new class to handle writing from a handle to another using Overlapped IO though for this to work nicely. I'm happy to help with that if you want

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -19,6 +19,12 @@ Module Name:
 #include "ServiceProcessLauncher.h"
 #include "WslCoreFilesystem.h"
 
+namespace {
+constexpr const char* nerdctlPath = "/usr/bin/nerdctl";
```
OneBlue: nit: this is already available as `wsl::windows::service::wsla::nerdctlPath`

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -292,15 +298,71 @@ try
...
+    ServiceProcessLauncher launcher{
+        "/bin/sh", {"/bin/sh", "-c", "cat | /usr/bin/nerdctl load"}, {}, ProcessFlags::Stdin | ProcessFlags::Stdout | ProcessFlags::Stderr};
+    auto loadProcess = launcher.Launch(*m_virtualMachine.Get());
+
+    auto loadProcessStdin = loadProcess.GetStdHandle(0);
+    wsl::windows::common::relay::InterruptableRelay(
+        imageFileHandle, loadProcessStdin.get(), m_sessionTerminatingEvent.get(), 4 * 1024 * 1024 /* 4MB buffer */);
+    loadProcessStdin.reset();
+
+    auto result = loadProcess.WaitAndCaptureOutput();
```
OneBlue, line 338: Unfortunately this can also lead to a pipe deadlock if `nerdctl` writes on stdout / stderr and fills the socket buffer. The easiest way to solve this would be use a `wsl::windows::common::relay::MultiHandleWait` and create a new `OverlappedIOHandle` class that allows us to relay a handle to the process's stdin. 

`[WriteHandle](wsl::windows::common::relay::MultiHandleWait)` would be the closest example to what we'll need here

yao-msft: Currently nerdctl only outputs very small message (like the imported image id), it's quite unlikely that it'll fill the buffer. Is it critical to make the change with this pr?

OneBlue: It's definitely unlikely, but if that did happen, (for instance if there was an error with lots of details) we'd essentially deadlock, since we'll need to hold `m_lock` here

I'd be OK fixing this in a followup PR though

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -292,15 +298,71 @@ try
...
+    return S_OK;
 }
+CATCH_RETURN();
 
-HRESULT WSLASession::ImportImage(ULONG Handle, LPCWSTR Image, IProgressCallback* ProgressCallback)
+HRESULT WSLASession::LoadImage(ULONG ImageHandle, IProgressCallback* ProgressCallback)
+try
 {
-    return E_NOTIMPL;
+    HANDLE imageFileHandle = wsl::windows::common::wslutil::DuplicateHandleFromCallingProcess(ULongToHandle(ImageHandle));
```
OneBlue: We should acquire `m_lock`  before accessing `m_virtualMachine`

OneBlue: (same for below methods)

tools/test/pack-test-data.ps1
```diff
@@ -0,0 +1,30 @@
+<#
```
OneBlue, line 1: Could we also add the logic that actually creates the test data ? That would make it easier to update it in the future. 

The easiest way would be to just call into `wsl.exe` (for this script, we could just assume that `docker` is available in the default WSL distro).

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -292,15 +293,84 @@ try
...
+CATCH_RETURN();
+
+HRESULT WSLASession::ImportImage(ULONG ImageHandle, LPCSTR ImageName, IProgressCallback* ProgressCallback)
+try
+{
+    UNREFERENCED_PARAMETER(ProgressCallback);
+
+    wil::unique_handle imageFileHandle{wsl::windows::common::wslutil::DuplicateHandleFromCallingProcess(ULongToHandle(ImageHandle))};
+    RETURN_HR_IF(E_INVALIDARG, INVALID_HANDLE_VALUE == imageFileHandle.get());
+    RETURN_HR_IF_NULL(E_POINTER, ImageName);
```
OneBlue, line 353: nit: This check should be at the beginning of the method, before we duplicate the HANDLE

test/windows/WSLATests.cpp
```diff
@@ -269,6 +271,71 @@ class WSLATests
...
+        auto session = CreateSession(settings);
+
+        VERIFY_SUCCEEDED(session->PullImage("hello-world", nullptr, nullptr));
+
+        // Verify that the image is in the list of images.
+        WSLAProcessLauncher launcher("/usr/bin/nerdctl", {"/usr/bin/nerdctl", "images"});
+        auto listImagesResult = launcher.Launch(*session).WaitAndCaptureOutput();
+        VERIFY_ARE_EQUAL(0, listImagesResult.Code);
+        VERIFY_IS_TRUE(listImagesResult.Output[1].find("hello-world") != std::string::npos);
+    }
```
OneBlue, line 291: nit: I'd recommend adding test coverage for negative paths like: 

- If the image URI is invalid 
- If the HANDLE isn't a valid file (we could open something like `wsl.exe` and just pass this through to check that we behave correctly) 
- If the image name is already in use

OneBlue (approved):
Change looks good, added a few minor comments that can be followed up with in a future PR

test/windows/WSLATests.cpp
yao-msft: Yes, I'll add them in follow up pr.

PR #13880, opened 2025-12-11
WSLA: Add virtiofs support for mounted Windows paths

benhillis:
This change adds a new feature flag to allow opting into VirtioFs for WSLA. There are a few other minor changes included:
1. Change the signature of MountWindowsFolderImpl to not pass a boolean in addition to flags
2. Fix a string format print issue in UtilMount
3. Resolve an issue in 9p windows folder mounts where the readonly flag was not being passed correctly

I have added a flag to 'wsl.exe --wsla' to test this as well as automated test coverage.

### DIFF
--- a/src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
+++ b/src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
@@ -1196,7 +1196,7 @@ try
         else
         {
             const bool admin = wsl::windows::common::security::IsTokenElevated(userToken.get());
-            const auto deviceId = m_guestDeviceManager->AddGuestDevice(
+            m_guestDeviceManager->AddGuestDevice(
                 VIRTIO_FS_DEVICE_ID,
                 admin ? WSLA_VIRTIO_FS_ADMIN_CLASS_ID : WSLA_VIRTIO_FS_CLASS_ID,
                 shareName.c_str(),
@@ -1204,8 +1204,6 @@ try
                 WindowsPath,
                 VIRTIO_FS_FLAGS_TYPE_FILES,
                 userToken.get());
-
-            // TODO: store deviceId so it can be removed later.
         }
 
         m_mountedWindowsFolders.emplace(LinuxPath, shareName);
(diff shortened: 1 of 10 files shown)


src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1151,64 +1154,101 @@ CATCH_RETURN();
...
+                LX_INIT_UTILITY_VM_PLAN9_PORT,
+                flags,
+                userToken.get());
+        }
+        else
+        {
+            const bool admin = wsl::windows::common::security::IsTokenElevated(userToken.get());
+            m_guestDeviceManager->AddGuestDevice(
+                VIRTIO_FS_DEVICE_ID,
+                admin ? WSLA_VIRTIO_FS_ADMIN_CLASS_ID : WSLA_VIRTIO_FS_CLASS_ID,
```
OneBlue, line 1201: What do the different class ids actually do for virtiofs ? It makes to me to split those two in WSL since we have two mount namespaces, but in WSLA the VM is only associated to a single user token, could we get always use the same class to simplify things a bit ?

OneBlue (approved):
One small question, but LGTM. 

We'll probably need to add support to dynamically remove the shares before shipping so they don't accumulate over time as containers start and stop, but this is great for starting to test with virtiofs

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
benhillis: Yeah, the ClassId determines which dllhost process gets the share. For WSL we have separate dllhosts, one that runs as admin, one that runs as non-admin. If that's not a requirement for WSLA we can only have a single server, but here we're impersonating the client to get the token so I think it would be safer to leave this as-is for now.

OneBlue: Ok I think this is fine for now, we should probably investigate whether we can use only one before shipping though, that way we could have only one COM registration

ChangmingTian:
I'd like to try this new feature
My WSL version is 2.7.0.0 pre-release
"wsl --wsla" not work
How can I get the virtiofs feature work?

PR #13881, opened 2025-12-12
Integrate the new runtime VHD

OneBlue:

This PR updates the rootfs nuget package to use something closer to our final rootfs.

This change also removes the `/etc/lsw-init.sh` logic and implement proper synchronization on containerd's startup

## PR Checklist


src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -39,27 +41,33 @@ WSLASession::WSLASession(ULONG id, const WSLA_SESSION_SETTINGS& Settings, WSLAUs
...
+    ServiceProcessLauncher launcher{
+        "/usr/bin/containerd",
+        {"/usr/bin/containerd"},
+        {{"PATH=/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/sbin"}},
+        common::ProcessFlags::Stdout | common::ProcessFlags::Stderr};
+    m_containerdThread = std::thread(&WSLASession::MonitorContainerd, this, launcher.Launch(*m_virtualMachine.Get()));
+
+    // Wait for containerd to be ready before.
+    // TODO: Configurable timeout.
+    THROW_WIN32_IF_MSG(ERROR_TIMEOUT, !m_containerdReadyEvent.wait(10 * 1000), "Timed out waiting for containerd to start");
```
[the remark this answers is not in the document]
OneBlue: That's not true. Loglines are not dropped, they're on the socket until the monitoring thread reads them

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -116,12 +124,20 @@ WSLASession::~WSLASession()
...
 
+    m_sessionTerminatingEvent.SetEvent();
+
+    // N.B. The containerd thread can only run if the VM is running.
+    if (m_containerdThread.joinable())
+    {
+        m_containerdThread.join();
+    }
+
     if (m_virtualMachine)
```
benhillis, line 135: why unmount instead of just terminating the VM?

benhillis (approved):

src/windows/wslaservice/exe/WSLASession.cpp
OneBlue: This is to make sure that all IO is correctly flushed when the VM terminates and has the benefit of guaranteeing that if the caller restarts a session immediately after, the VHD will be mountable right away

PR #13887, opened 2025-12-12
virtiofs: implement best-effort removal of flexiov devices

benhillis:
This change implements best effort removal of flexiov devices so these devices will currently accumulate while the VM is running, which is not ideal. This is because HCS is currently missing the plumbing for this which I am working on. When that support is ready, the RemoveDevice call will light up without any future changes to the WSL codebase.

### DIFF
--- a/src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
+++ b/src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
@@ -1214,15 +1214,7 @@ try
         {
             auto mountInfo = mountIt->second;
             m_mountedWindowsFolders.erase(mountIt);
-
-            if (!FeatureEnabled(WslaFeatureFlagsVirtioFs))
-            {
-                hcs::RemovePlan9Share(m_computeSystem.get(), mountInfo.ShareName.c_str(), LX_INIT_UTILITY_VM_PLAN9_PORT);
-            }
-            else if (mountInfo.InstanceId.has_value())
-            {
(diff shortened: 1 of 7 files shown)


src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -1206,16 +1204,25 @@ try
...
         {
-            hcs::RemovePlan9Share(m_computeSystem.get(), shareName.c_str(), LX_INIT_UTILITY_VM_PLAN9_PORT);
+            auto mountInfo = mountIt->second;
+            m_mountedWindowsFolders.erase(mountIt);
+
+            if (!FeatureEnabled(WslaFeatureFlagsVirtioFs))
+            {
+                hcs::RemovePlan9Share(m_computeSystem.get(), mountInfo.ShareName.c_str(), LX_INIT_UTILITY_VM_PLAN9_PORT);
+            }
+            else if (mountInfo.InstanceId.has_value())
```
OneBlue: nit: This logic is in two places, we might want to have something like: 

```
void DisconnectShare(const MountedFolderInfo&);
```

Also if virtiofs is enabled, `InstanceId` should always be set, so we should assert if it's not

OneBlue (approved):

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
benhillis: good call adding a helper, I like that.

OneBlue (approved):

PR #13888, opened 2025-12-12
Windows folders can be mounted to WSLA containers

kvega005:
- Adds support to mount windows folders to WSLA containers.
- Ensures that containers are deleted when the reference to the container object is dropped (i.e. the WSLA session exits).

## PR Checklist


test/windows/WSLATests.cpp
```diff
@@ -1407,11 +1271,77 @@ class WSLATests
...
+                "debian:latest", "test-unique-name", "echo", {"OK"}, {}, ProcessFlags::Stdin | ProcessFlags::Stdout | ProcessFlags::Stderr);
 
             auto result = otherLauncher.Launch(*session).GetInitProcess().WaitAndCaptureOutput();
             VERIFY_ARE_EQUAL(result.Output[1], "OK\n");
             VERIFY_ARE_EQUAL(result.Code, 0);
         }
     }
-};
+
+    TEST_METHOD(ContainerVolume)
```
benhillis: Similar to what I did for the mountwindowsfolder test, create virtiofs and non virtiofs variations of this using a helper.

src/windows/common/WSLAContainerLauncher.cpp
```diff
@@ -53,6 +53,16 @@ WSLAContainerLauncher::WSLAContainerLauncher(
 {
 }
 
+void wsl::windows::common::WSLAContainerLauncher::AddVolume(const std::wstring& HostPath, const std::string& ContainerPath, bool ReadOnly)
```
OneBlue, line 57: This API might be a bit tricky, because we capture the reference via `c_str()`, but don't actually make a copy, so if the caller did this: 

```
launcher.AddVolume("foo", "bar", true);
```

Both pointers would be invalid after `AddVolume()` returns.

To fix this we should either: 

```
- Make `HostPath` and `ContainerPath` char pointers (to make it explicit to the caller)
- Or have an array of `std::string` where we copy the strings to make sure they exist until the object is destroyed
```

I have a slight preference for the latter since it's a bit more user friendly, but I'm OK with either

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -240,27 +176,63 @@ Microsoft::WRL::ComPtr<WSLAContainer> WSLAContainer::Create(
...
-    while (!id.empty() && (id.back() == '\n'))
+    for (ULONG i = 0; i < Options.VolumesCount; i++)
     {
-        id.pop_back();
+        try
+        {
+            const WSLA_VOLUME& volume = Options.Volumes[i];
+            std::string parentVMPath = std::format("/mnt/wsla/{}/volumes/{}", Options.Name, i);
+
+            THROW_IF_FAILED(parentVM.MountWindowsFolder(volume.HostPath, parentVMPath.c_str(), volume.ReadOnly));
```
OneBlue: nit: I'd recommend doing something like: 

```
THROW_IF_FAILED_MSG([...], "Failed to mount %ls -> %hs", volume.HostPath, parentVMPath.c_str());
```

To make debugging easier if this fails

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -294,8 +266,27 @@ std::vector<std::string> WSLAContainer::PrepareNerdctlCreateCommand(const WSLA_C
         args.push_back(options.InitProcessOptions.Executable);
     }
 
+    
+    if (!volumes.empty())
```
OneBlue: nit: this `if` can be removed

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -17,179 +17,120 @@ Module Name:
...
 
 // Constants for required default arguments for "nerdctl run..."
 static std::vector<std::string> defaultNerdctlRunArgs{//"--pull=never", // TODO: Uncomment once PullImage() is implemented.
                                                       "--net=host", // TODO: default for now, change later
                                                       "--ulimit",
                                                       "nofile=65536:65536"};
 
-WSLAContainer::WSLAContainer(WSLAVirtualMachine* parentVM, const WSLA_CONTAINER_OPTIONS& Options, std::string&& Id, ContainerEventTracker& tracker) :
-    m_parentVM(parentVM), m_name(Options.Name), m_image(Options.Image), m_id(std::move(Id))
+WSLAContainer::WSLAContainer(WSLAVirtualMachine* parentVM, ServiceRunningProcess&& containerProcess, const char* name, const char* image, std::vector<VolumeMountInfo> volumes) :
```
OneBlue: nit: Let's use `std::vector<VolumeMountInfo>&& volumes` here so we actually move the vector

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -294,8 +266,27 @@ std::vector<std::string> WSLAContainer::PrepareNerdctlCreateCommand(const WSLA_C
         args.push_back(options.InitProcessOptions.Executable);
     }
 
+    
+    if (!volumes.empty())
+    {
+        for (const auto& volume : volumes)
+        {
+            std::string mountArg = "-v";
```
OneBlue: nit: this could be simplified to: 

```
args.emplace_back(std::format("-v{}:{}{}", volume.ParentVMPath, volume.ContainerPath, volume.ReadOnly ? ":ro", "");
```

(and also avoids copying `mountArg`)

test/windows/WSLATests.cpp
```diff
@@ -1407,11 +1271,77 @@ class WSLATests
...
+
+            ValidateProcessOutput(process, {{1, "OK\n"}});
+
+            VERIFY_ARE_EQUAL(container.State(), WslaContainerStateExited);
+            VERIFY_SUCCEEDED(container.Get().Delete());
+        }
+
+        // Validate that the volumes are not mounted after container exits.
+        ExpectMount(session.get(), std::format("/mnt/wsla/{}/volumes/{}", containerName, 0), {});
+        ExpectMount(session.get(), std::format("/mnt/wsla/{}/volumes/{}", containerName, 1), {});
```
OneBlue, line 1742: nit: We might want to add a test case that validates that if we pass multiple volumes, but the last one fails, we correctly unmount everything (this will give us coverage for the catch block in `WSLAContainer::MountVolumes()`)

OneBlue (commented):
Change LGTM, added a few minor comments

src/windows/common/WSLAContainerLauncher.cpp
kvega005: Yeah, I was thinking of making this change because I ran into this exact problem while writing the test code. I can implement it in this PR.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -279,6 +294,46 @@ void WSLAContainer::AddEnvironmentVariables(std::vector<std::string>& args, cons
...
+    return mountedVolumes;
+}
+
+void wsl::windows::service::wsla::WSLAContainer::UnmountVolumes(const std::vector<VolumeMountInfo>& volumes, WSLAVirtualMachine& parentVM)
+{
+    for (const auto& volume : volumes)
+    {
+        try
+        {
+            parentVM.UnmountWindowsFolder(volume.ParentVMPath.c_str());
```
OneBlue: nit: `LOG_IF_FAILED` since `UnmountWindowsFolder()` doesn't throw on error (probably something we should fix later). 

Also `UnmountWindowsFolder()` doesn't delete the folder after unmounting, we might want to add a flag to do that since in this specific case we know that we created the mount folder. 

That can be done in a followup change though

OneBlue (approved):
LGTM. One minor comment, but feel free to resolve it in a future change

PR #13890, opened 2025-12-12
cleanup: switch drvfs.cpp to use common MountUtil helper

benhillis:
The drvfs.cpp version of MountWithRetry was almost identical to the version in util.cpp. This change removes the duplicate mount retry helper.

### DIFF
--- a/src/linux/init/drvfs.cpp
+++ b/src/linux/init/drvfs.cpp
@@ -247,7 +247,11 @@ try
     //
 
     int Result = access(Target, F_OK);
-    if (Result == 0)
+    if (Result < 0)
+    {
+        LOG_STDERR(errno);
+    }
+    else
     {
         auto Parsed = mountutil::MountParseFlags(Options);
         Result = UtilMount(Source, Target, FsType, Parsed.MountFlags, Parsed.StringOptions.c_str(), std::chrono::seconds{2});
@@ -255,15 +259,7 @@ try
 
     if (ExitCode)
     {
-        if (Result < 0)
-        {
-            LOG_STDERR(errno);
-            *ExitCode = c_exitCodeMountFail;
-        }
-        else
-        {
-            *ExitCode = 0;
-        }
+        *ExitCode = Result < 0 ? c_exitCodeMountFail : 0;
     }
 
     return Result;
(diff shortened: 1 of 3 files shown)


src/linux/init/drvfs.cpp
```diff
@@ -246,52 +242,15 @@ Return Value:
...
+
+    int Result = access(Target, F_OK);
+    if (Result == 0)
     {
-        errno = wil::ResultFromCaughtException();
-        auto parsed = mountutil::MountParseFlags(Options);
-        LOG_ERROR("mount({}, {}, {}, {:#x}, {}) failed: {}", Source, Target, FsType, parsed.MountFlags, parsed.StringOptions.c_str(), strerror(errno));
-        Result = -1;
+        auto Parsed = mountutil::MountParseFlags(Options);
+        int Result = UtilMount(Source, Target, FsType, Parsed.MountFlags, Parsed.StringOptions.c_str(), std::chrono::seconds{2});
```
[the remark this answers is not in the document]
benhillis: nice catch, this was an intentional change since there is a message logged in kmesg when trying to mount a tag that does not exist.

src/linux/init/drvfs.cpp
```diff
@@ -246,52 +242,15 @@ Return Value:
...
-            });
-
-        Result = 0;
-    }
-    catch (...)
+    //
+    // Verify the target directory exists before mounting.
+    //
+
+    int Result = access(Target, F_OK);
```
OneBlue, line 249: What's the reason for the manual access check ?

src/linux/init/util.cpp
```diff
@@ -1720,7 +1722,23 @@ Return Value:
                 TimeoutSeconds.value(),
                 [&]() {
                     errno = wil::ResultFromCaughtException();
-                    return errno == ENOENT || errno == ENXIO || errno == EIO;
+
+                    // Generic device not ready errors
+                    if (errno == ENXIO || errno == EIO || errno == ENOENT)
```
OneBlue, line 1727: In the previous implementation we only retry on `ENOENT` if the target path doesn't exist, but with change we'll always going to create the folder, which is probably not what we want (at least for `mount.drvfs`). 

One way we could factor these two mount logic would be to have a flag on whether the target folder should be created, and a `std::function<bool(int)>` so the caller can override the retry logic (it could be defaulted to the logic we currently have, which retries if `errno == ENXIO || errno == EIO || errno == ENOENT`

OneBlue (approved):

OneBlue (approved):

PR #13918, opened 2025-12-15
WSLA: Invoke WSLAContainer callbacks without the callback lock held

benhillis:
This pull request improves thread safety and reduces the risk of deadlocks by restructuring how locks are held during callback invocation and resource cleanup in container event and state management code. Invoking callbacks while holding the lock of the callback manager is a dangerous pattern that we appear to be trying to work around by using `std::recursive_mutex` which may not be needed after this change.

**Thread safety and deadlock avoidance:**

* In `ContainerEventTracker::OnEvent`, callbacks are now copied while holding the lock, but are invoked after releasing the lock to prevent deadlocks if callbacks attempt to register or unregister themselves.
* In `WSLAContainer::State`, the destruction of the container process is deferred until after the lock is released, ensuring that the destructor does not run while the lock is held, which could otherwise cause deadlocks.

### DIFF
--- a/src/windows/wslaservice/exe/ContainerEventTracker.cpp
+++ b/src/windows/wslaservice/exe/ContainerEventTracker.cpp
@@ -112,15 +112,25 @@ void ContainerEventTracker::OnEvent(const std::string& event)
     auto containerIdIt = innerEvent.find("container_id");
     THROW_HR_IF_MSG(E_INVALIDARG, containerIdIt == innerEvent.end(), "Failed to parse json: %hs", innerEventJson.c_str());
 
-    std::lock_guard lock{m_lock};
-
     std::string containerId = containerIdIt->get<std::string>();
-    for (const auto& e : m_callbacks)
+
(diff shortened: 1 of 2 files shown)


src/windows/wslaservice/exe/ContainerEventTracker.cpp
```diff
@@ -112,16 +112,26 @@ void ContainerEventTracker::OnEvent(const std::string& event)
...
+            {
+                callbacksToInvoke.push_back(e.Callback);
+            }
         }
     }
+
+    // Invoke callbacks outside the lock
+    for (const auto& callback : callbacksToInvoke)
+    {
+        callback(it->second);
```
OneBlue, line 133: Unfortunately this won't work because it could mean that we call a callback after `UnregisterContainerStateUpdates()` returns, which could mean that we'll call into deleted classes.

If we refactor this we need to make sure that callbacks can't be called after unregistered

OneBlue (changes requested):

PR #13922, opened 2025-12-16
Implement container port mapping support

OneBlue:
## PR Checklist

### DIFF
--- a/src/windows/common/WSLAContainerLauncher.h
+++ b/src/windows/common/WSLAContainerLauncher.h
@@ -57,6 +57,7 @@ public:
 private:
     std::string m_image;
     std::string m_name;
+    std::vector<WSLA_PORT_MAPPING> m_ports;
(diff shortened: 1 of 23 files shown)


src/windows/common/WSLAContainerLauncher.h
```diff
@@ -57,6 +57,7 @@ class WSLAContainerLauncher : private WSLAProcessLauncher
 private:
     std::string m_image;
     std::string m_name;
+    std::vector<WSLA_PORT_MAPPING> m_ports;
```
benhillis, line 61: do we need a lock to protect this member?

OneBlue: We don't, this class isn't meant to be used from different threads

OneBlue:
I messed up the history. I'll close and reopen a new PR for that change.

PR #13931, opened 2025-12-17
WSLA: Always user the token of the user that created the VM

benhillis:
Currently there are a few paths that use COM impersonation to get the token of the calling user. For WSLA we want to always use the token of the user that created the VM.


src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -46,6 +46,7 @@ WSLAVirtualMachine::WSLAVirtualMachine(WSLAVirtualMachine::Settings&& Settings,
 
     m_vmIdString = wsl::shared::string::GuidToString<wchar_t>(m_vmId, wsl::shared::string::GuidToStringFlags::Uppercase);
     m_userToken = wsl::windows::common::security::GetUserToken(TokenImpersonation);
+    m_virtioFsClassId = wsl::windows::common::security::IsTokenElevated(m_userToken.get()) ? WSLA_VIRTIO_FS_ADMIN_CLASS_ID : WSLA_VIRTIO_FS_CLASS_ID;
```
benhillis, line 49: We likely only need a single virtiofs class ID, but we can revisit that later.

OneBlue (approved):

PR #13934, opened 2025-12-18
Implement container port mapping support

OneBlue:

This change adds support for port mapping, both in bridge & host network modes

## PR Checklist

- [] **Closes:** Link to issue #xxx

### DIFF
--- a/src/shared/inc/lxinitshared.h
+++ b/src/shared/inc/lxinitshared.h
@@ -284,6 +284,11 @@ Abstract:
     }
 
 //
+// Definitions used by WSLA
+//
+
+constexpr auto c_ephemeralPortRange = std::pair<uint16_t, uint16_t>(10000, 20000);
+
 // The types of messages that can be sent to init and mini init.
 //
 
--- a/src/windows/wslaservice/exe/WSLAContainer.cpp
+++ b/src/windows/wslaservice/exe/WSLAContainer.cpp
@@ -18,17 +18,131 @@ Abstract:
 
 using wsl::windows::service::wsla::VolumeMountInfo;
 using wsl::windows::service::wsla::WSLAContainer;
+using wsl::windows::service::wsla::WSLAVirtualMachine;
 
 // Constants for required default arguments for "nerdctl create..."
 static std::vector<std::string> defaultNerdctlCreateArgs{//"--pull=never", // TODO: Uncomment once PullImage() is implemented.
                                                          "--ulimit",
                                                          "nofile=65536:65536"};
 
+static std::vector<std::string> defaultNerdctlEnv{"PATH=/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/sbin"};
+
+namespace {
+auto ProcessPortMappings(const WSLA_CONTAINER_OPTIONS& options, WSLAVirtualMachine& vm, std::vector<std::string>& args)
+{
+    THROW_HR_IF_MSG(
+        E_INVALIDARG,
+        options.PortsCount > 0 && options.ContainerNetwork.ContainerNetworkType == WSLA_CONTAINER_NETWORK_NONE,
+        "Port mappings are not supported without networking");
+
+    // Validate that port mappings are valid.
+    // N.B. If a host port is duplicated, MapPort() will fail later.
+    for (ULONG i = 0; i < options.PortsCount; i++)
+    {
+        const auto& port = options.Ports[i];
+        THROW_HR_IF_MSG(E_INVALIDARG, port.Family != AF_INET && port.Family != AF_INET6, "Invalid family for port mapping %i: %i", i, port.Family);
+    }
+
+    // Generate Windows <-> VM port mappings depending on the networking mode.
+    // N.B. pointers are used so the vectors are still available if the errorCleanup is executed.
+    auto vmPorts = std::make_shared<std::set<uint16_t>>();
+    auto mappedPorts = std::make_shared<std::vector<WSLAContainer::PortMapping>>();
+
+    auto errorCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [mappedPorts = mappedPorts, vmPorts = vmPorts, &vm]() {
+        if (!vmPorts->empty())
+        {
+            LOG_IF_FAILED(wil::ResultFromException([&]() { vm.ReleasePorts(*vmPorts); }));
+        }
+
+        for (const auto& e : *mappedPorts)
+        {
+            if (e.MappedToHost)
+            {
+                LOG_IF_FAILED_MSG(
+                    vm.MapPort(e.Family, e.VmPort, e.HostPort, true),
+                    "Failed to unmap port (family=%i, guestPort=%u, hostPort=%u)",
+                    e.Family,
+                    e.VmPort,
+                    e.HostPort);
+            }
+        }
+    });
+
+    if (options.ContainerNetwork.ContainerNetworkType == WSLA_CONTAINER_NETWORK_BRIDGE)
+    {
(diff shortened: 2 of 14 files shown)


src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -47,6 +167,24 @@ WSLAContainer::~WSLAContainer()
...
     m_trackingReference.Reset();
+
+    // Release port mappings.
+    std::set<uint16_t> allocatedGuestPorts;
+    for (const auto& e : m_mappedPorts)
+    {
+        WI_ASSERT(e.MappedToHost);
+
+        LOG_IF_FAILED_MSG(
+            m_parentVM->MapPort(e.Family, e.HostPort, e.VmPort, true),
```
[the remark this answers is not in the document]
OneBlue: That's not true. Here's the method signature: 

```
HRESULT WSLAVirtualMachine::MapPort(_In_ int Family, _In_ short WindowsPort, _In_ short LinuxPort, _In_ BOOL Remove)
```

src/shared/inc/lxinitshared.h
```diff
@@ -284,6 +284,11 @@ Module Name:
     }
 
 //
+// Definitions used by WSLA
+//
+
+constexpr auto c_ephemeralPortRange = std::pair<uint16_t, uint16_t>(10000, 20000);
```
ptrivedi, line 290: nit: Is this too limited a range to leave for the OS? Should we just aim at using what k8s nodeport uses (30000-32767 I think?) and leave the OS default of 32768-6xxxx as it is?

OneBlue: I'd rather explicitly set it so we can be guaranteed that ephemeral ports won't collide with ours. 10K ports feels like a pretty reasonable amount. I also wouldn't mind moving the range

test/windows/WSLATests.cpp
```diff
@@ -1674,6 +1655,198 @@ class WSLATests
...
+
+            // Validate that the port cannot be reused while the container is running.
+            WSLAContainerLauncher subLauncher(
+                "python:3.12-alpine", "test-ports-2", {}, {"python3", "-m", "http.server"}, {"PYTHONUNBUFFERED=1"}, Mode);
+
+            subLauncher.AddPort(1234, 8000, AF_INET);
+            auto [hresult, newContainer] = subLauncher.LaunchNoThrow(*session);
+            VERIFY_ARE_EQUAL(hresult, HRESULT_FROM_WIN32(ERROR_ALREADY_EXISTS));
+
+            VERIFY_SUCCEEDED(container.Get().Stop(9, 0));
```
ptrivedi, line 1760: Do we unmap the host port upon container delete? Don't think I saw that. We should unmap host port on Delete and verify that here

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -351,10 +489,13 @@ Microsoft::WRL::ComPtr<WSLAContainer> WSLAContainer::Create(
 
     AddEnvironmentVariables(inputOptions, containerOptions.InitProcessOptions);
 
+    // TODO: Rethink command line generation logic.
+    auto [mappedPorts, errorCleanup] = ProcessPortMappings(containerOptions, parentVM, inputOptions);
```
ptrivedi, line 496: I just thought of something. The ports are not allocated/mapped in the host's networking stack until the container is 'started'.
When I do nerdctl create on a container, the port bindings will still show null. Technically, this means that you can 'nerdctl create' two containers with the same port mapping

test/windows/WSLATests.cpp
OneBlue: We unmap when the container instance is destroyed: https://github.com/microsoft/WSL/blob/484d7316b625b9f6abdc63cbd9a52ba115be120b/src/windows/wslaservice/exe/WSLAContainer.cpp#L169 

Which happens either when the container is deleted, or when the VM terminates. We could do this directly in delete(), but I'm going to make changes the to WSLAContainer lifetime logic, so I'd rather wait for that change

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: This is probably not something that we want to support though, even if nerdctl allows it, allowing multiple container to bind the same port feels pretty counter intuitive at the API level

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -351,10 +489,13 @@ Microsoft::WRL::ComPtr<WSLAContainer> WSLAContainer::Create(
...
+
```
ptrivedi, line 497: Can we add some TODOs here to 
1) Determine if the port mapping should be done in Create() or Start()
2) Test out port mapping scenarios with docker/nerdctl and understand some of the behavior differences
3) Unmapping of ports in Delete()
4) Determine if we want to unmap ports on Stop() depending on what shakes out of 1) and 2)

test/windows/WSLATests.cpp
```diff
@@ -1674,6 +1655,198 @@ class WSLATests
...
+            auto container = launcher.Launch(*session);
+            auto initProcess = container.GetInitProcess();
+            auto stdoutHandle = initProcess.GetStdHandle(1);
+
+            // Wait for the container bind() to be completed.
+            WaitForOutput(stdoutHandle.get(), "Serving HTTP on 0.0.0.0 port 8000");
+
+            ExpectHttpResponse(L"http://127.0.0.1:1234", 200);
+            ExpectHttpResponse(L"http://[::1]:1234", {});
+
```
ptrivedi, line 1751: In addition to this, would also be good to have the test check PortBindings in the inspect output to ensure the mapping. We could also ensure unmapping after Delete() when we have that

ptrivedi (approved):
As per discussion, since you are planning to address some of the comments already and some in a follow-up PR, I will go ahead and approve this to unblock.

src/windows/wslrelay/localhost.cpp
```diff
@@ -583,8 +580,16 @@ void wsl::windows::wslrelay::localhost::RunWSLAPortRelay(const GUID& VmId, uint3
...
-                ports.emplace(key, CreatePortListener(message->WindowsPort, message->LinuxPort, RelayPort, message->AddressFamily));
-                update = true;
+                try
+                {
+                    ports.emplace(key, CreatePortListener(message->WindowsPort, message->LinuxPort, RelayPort, message->AddressFamily));
+                    update = true;
+                }
+                catch (...)
+                {
+                    result = wil::ResultFromCaughtException();
```
benhillis, line 590: I don't think this result variable is getting used anywhere, or at least I can't follow where.

benhillis: well I see where it's getting used, but isn't it getting overwritten after the loop restarts?

OneBlue: This is used to return the error via the scope_exit on line 547. It's inside the scope of the loop: 


```
        HRESULT result = E_UNEXPECTED;
        auto sendResponse = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
            WSL_LOG(
                "PortMapping",
                TraceLoggingValue(result, "Result"),
                TraceLoggingValue(message->WindowsPort, "WindowsPort"),
                TraceLoggingValue(message->LinuxPort, "LinuxPort"),
                TraceLoggingValue(message->Stop, "Remove"));

            THROW_LAST_ERROR_IF(!WriteFile(GetStdHandle(STD_OUTPUT_HANDLE), &result, sizeof(result), nullptr, nullptr));
        });
```

PR #13939, opened 2025-12-19
cleanup: add scsi disks during VM creation

benhillis:
This change moves from "hot adding" the system distro and kernel modules VHDs during runtime, to adding them to the initial creation JSON of the VM. This will have marginally better performance but also simplifies the flow and removes error paths.

### DIFF
--- a/src/windows/service/exe/WslCoreVm.cpp
+++ b/src/windows/service/exe/WslCoreVm.cpp
@@ -1730,7 +1730,7 @@ std::wstring WslCoreVm::GenerateConfigJson()
         return lun;
     };
 
-    if (m_systemDistroDeviceType = LxMiniInitMountDeviceTypeLun)
+    if (m_systemDistroDeviceType == LxMiniInitMountDeviceTypeLun)
     {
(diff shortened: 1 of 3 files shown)


src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -1729,9 +1714,33 @@ std::wstring WslCoreVm::GenerateConfigJson()
...
+        disk.ReadOnly = true;
+        disk.SupportCompressedVolumes = true;
+        disk.AlwaysAllowSparseFiles = true;
+        disk.SupportEncryptedFiles = true;
+        scsiController.Attachments[std::to_string(lun)] = std::move(disk);
+        m_attachedDisks.emplace(AttachedDisk{DiskType::VHD, path, false}, DiskState{lun, {}, {}});
+        return lun;
+    };
+
+    if (m_systemDistroDeviceType = LxMiniInitMountDeviceTypeLun)
```
OneBlue: This should be a `==` .

I'm actually surprised the compiler didn't issue a warning for this

OneBlue (approved):

PR #13943, opened 2025-12-19
WSLA: Switch VHD boot disk approach and support PMEM for VHD disks

benhillis:
This PR introduces support for attaching VHDs (Virtual Hard Disks) as persistent memory (PMEM) devices in WSL, controlled by a new feature flag. It refactors how virtual disks are attached and mounted, adds new schema definitions for PMEM devices, and simplifies the mounting logic by integrating it directly into the VM creation JSON.

### DIFF
--- a/src/windows/common/hcs_schema.h
+++ b/src/windows/common/hcs_schema.h
@@ -458,6 +458,53 @@ inline void to_json(nlohmann::json& j, const DebugOptions& d)
     OMIT_IF_EMPTY(j, d, ShutdownOrResetSavedStateFileName);
 }
 
+enum VirtualPMemImageFormat
+{
+    Vhdx,
+    Vhd1
+};
+
+NLOHMANN_JSON_SERIALIZE_ENUM(
+    VirtualPMemImageFormat,
+    {
+        {VirtualPMemImageFormat::Vhdx, "Vhdx"},
+        {VirtualPMemImageFormat::Vhd1, "Vhd1"},
+    })
+
+struct VirtualPMemDevice
+{
+    std::wstring HostPath;
+    bool ReadOnly;
+    VirtualPMemImageFormat ImageFormat;
+    // uint64_t SizeBytes;
+    // std::map<uint64_t, VirtualPMemMapping> Mappings;
+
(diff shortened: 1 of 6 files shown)


OneBlue (commented):
Change looks good, I'd be curious to see what the performance difference is with pmem vs scsi before making it the default though 

Also do we know if there are versions of Windows that don't support it ?

src/windows/common/hcs_schema.h
```diff
@@ -458,6 +458,53 @@ inline void to_json(nlohmann::json& j, const DebugOptions& d)
     OMIT_IF_EMPTY(j, d, ShutdownOrResetSavedStateFileName);
 }
 
+enum VirtualPMemImageFormat
+{
+    Vhdx,
+    Vhd1
+};
```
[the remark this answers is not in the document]
benhillis: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/WSL/pull/13943#discussion_r2636223380)

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -272,17 +272,74 @@ void WSLAVirtualMachine::Start()
...
+    auto kernelModulesPath = std::filesystem::path(TEXT(WSL_KERNEL_MODULES_PATH));
+
+#else
+
+    auto kernelModulesPath = basePath / L"tools" / L"modules.vhd";
+
+#endif
+
+    // Initialize the boot VHDs.
+    std::pair<std::optional<ULONG>, std::optional<std::string>> rootVhd;
```
OneBlue: nit: Since only one of the two value is set, I'd recommend using a `std::variant` for this so we can't use an uninitialized/default value by accident

benhillis: @copilot - implement this suggestion.

benhillis:
> Change looks good, I'd be curious to see what the performance difference is with pmem vs scsi before making it the default though
> 
> Also do we know if there are versions of Windows that don't support it ?

I've validated that PMEM is supported all the way back to Win 10.

OneBlue (approved):

PR #13965, opened 2025-12-23
Refactor wsladiag CLI parsing and fix shell PTY handling

beena352:
Summary of the Pull Request:
This PR refactors wsladiag command-line argument handling to use per-verb parsing and fixes multiple correctness and robustness issues in the interactive shell (PTY) path. It improves live terminal resize behavior, simplifies process lifecycle handling, and makes console cleanup more diagnosable, while preserving existing behavior by default.

Detailed Description:
Argument parsing

Refactors argument handling to be per-verb (list, shell), allowing each command to define and validate its own arguments.

Adds an opt-in relaxed mode to ArgumentParser to allow top-level verb dispatch without throwing on subcommand arguments. The default parsing behavior remains unchanged.

This makes the CLI easier to extend as additional wsladiag commands are added.

Shell / PTY handling

Fixes TTY size initialization by passing rows and columns in the correct order.

Restores and correctly handles live terminal resize using the TerminalControl FD.

Validation Steps Performed

Ran wsladiag list to verify session enumeration.

Ran wsladiag shell <SessionName> to verify interactive shell launch.

Verified interactive input/output behavior inside the shell.

Verified live terminal resize by resizing the console window while running commands such as top.

Verified clean process exit without hangs and correct console state restoration.

### DIFF
--- a/src/shared/inc/CommandLine.h
+++ b/src/shared/inc/CommandLine.h
@@ -307,8 +307,8 @@ class ArgumentParser
 public:
 #ifdef WIN32
 
-    ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1, bool IgnoreUnknownArgs = false) :
-        m_startIndex(StartIndex), m_name(Name), m_stopUnknownArgs(IgnoreUnknownArgs)
+    ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1, bool stopUnknownArgs = false) :
+        m_startIndex(StartIndex), m_name(Name), m_stopUnknownArgs(stopUnknownArgs)
     {
         m_argv.reset(CommandLineToArgvW(std::wstring(CommandLine).c_str(), &m_argc));
         THROW_LAST_ERROR_IF(!m_argv);
@@ -316,8 +316,8 @@ public:
 
 #else
 
-    ArgumentParser(int argc, const char* const* argv, bool IgnoreUnknownArgs = false) :
-        m_argc(argc), m_argv(argv), m_startIndex(1), m_stopUnknownArgs(IgnoreUnknownArgs)
+    ArgumentParser(int argc, const char* const* argv, bool stopUnknownArgs = false) :
+        m_argc(argc), m_argv(argv), m_startIndex(1), m_stopUnknownArgs(stopUnknownArgs)
     {
     }
 
@@ -402,7 +402,7 @@ public:
                         const TChar* value = nullptr;
                         if (e.Positional)
                         {
-                            value = m_argv[i]; // Positional arguments directly receive arvg[i]
+                            value = m_argv[i]; // Positional arguments directly receive argv[i]
                         }
                         else if (i + 1 < m_argc)
                         {
--- a/src/shared/inc/CommandLine.h
+++ b/src/shared/inc/CommandLine.h
@@ -307,8 +307,8 @@ class ArgumentParser
 public:
 #ifdef WIN32
 
-    ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1, bool stopUnknownArgs = false) :
-        m_startIndex(StartIndex), m_name(Name), m_stopUnknownArgs(stopUnknownArgs)
+    ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1, bool ignoreUnknownArgs = false) :
+        m_startIndex(StartIndex), m_name(Name), m_ignoreUnknownArgs(ignoreUnknownArgs)
     {
         m_argv.reset(CommandLineToArgvW(std::wstring(CommandLine).c_str(), &m_argc));
         THROW_LAST_ERROR_IF(!m_argv);
@@ -316,8 +316,8 @@ public:
 
 #else
 
-    ArgumentParser(int argc, const char* const* argv, bool stopUnknownArgs = false) :
-        m_argc(argc), m_argv(argv), m_startIndex(1), m_stopUnknownArgs(stopUnknownArgs)
+    ArgumentParser(int argc, const char* const* argv, bool ignoreUnknownArgs = false) :
+        m_argc(argc), m_argv(argv), m_startIndex(1), m_ignoreUnknownArgs(ignoreUnknownArgs)
     {
     }
 
@@ -431,7 +431,7 @@ public:
 
             if (!foundMatch)
             {
-                if (m_stopUnknownArgs)
+                if (m_ignoreUnknownArgs)
                 {
                     break;
                 }
@@ -542,7 +542,7 @@ private:
 
     int m_startIndex{};
     const TChar* m_name{};
-    bool m_stopUnknownArgs{false};
+    bool m_ignoreUnknownArgs{false};
 };
 } // namespace wsl::shared
 
--- a/src/shared/inc/CommandLine.h
+++ b/src/shared/inc/CommandLine.h
@@ -307,7 +307,8 @@ class ArgumentParser
 public:
 #ifdef WIN32
 
-    ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1) : m_startIndex(StartIndex), m_name(Name)
+    ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1, bool IgnoreUnknownArgs = false) :
+        m_startIndex(StartIndex), m_name(Name), m_ignoreUnknownArgs(IgnoreUnknownArgs)
     {
         m_argv.reset(CommandLineToArgvW(std::wstring(CommandLine).c_str(), &m_argc));
         THROW_LAST_ERROR_IF(!m_argv);
@@ -315,7 +316,8 @@ public:
 
 #else
 
-    ArgumentParser(int argc, const char* const* argv) : m_argc(argc), m_argv(argv), m_startIndex(1)
+    ArgumentParser(int argc, const char* const* argv, bool IgnoreUnknownArgs = false) :
+        m_argc(argc), m_argv(argv), m_startIndex(1), m_ignoreUnknownArgs(IgnoreUnknownArgs)
     {
     }
 
@@ -429,6 +431,11 @@ public:
 
             if (!foundMatch)
             {
+                if (m_ignoreUnknownArgs)
+                {
+                    break;
+                }
+
                 THROW_USER_ERROR(wsl::shared::Localization::MessageInvalidCommandLine(m_argv[i], m_name ? m_name : m_argv[0]));
             }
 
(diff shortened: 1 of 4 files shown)


src/windows/wsladiag/wsladiag.cpp
```diff
@@ -42,9 +42,35 @@ static int ReportError(const std::wstring& context, HRESULT hr)
...
+    parser.AddPositionalArgument(sessionName, 0);
+    parser.AddArgument(verbose, L"--verbose", L'v');
+
+    try
+    {
+        parser.Parse();
+    }
+    catch (...)
+    {
+        const auto hr = wil::ResultFromCaughtException();
```
OneBlue: I'd recommend letting the exception go up and handle the error at the `main()` level, so we can have one error handler for all arguments

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -42,9 +42,35 @@ static int ReportError(const std::wstring& context, HRESULT hr)
...
+        if (hr == E_INVALIDARG)
+        {
+            return 1;
+        }
+        throw;
+    }
+
+    if (sessionName.empty())
```
OneBlue, line 51: We should throw an error here since that `sessionName` being empty would mean that the command line was invalid

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -139,23 +166,46 @@ static int RunShellCommand(const std::wstring& sessionName, bool verbose)
     THROW_LAST_ERROR_IF(!SetConsoleOutputCP(CP_UTF8));
     THROW_LAST_ERROR_IF(!SetConsoleCP(CP_UTF8));
 
-    // Keep terminal control socket alive.
     auto exitEvent = wil::unique_event(wil::EventOptions::ManualReset);
-    wsl::shared::SocketChannel controlChannel{wil::unique_socket{(SOCKET)ttyControl.release()}, "TerminalControl", exitEvent.get()};
+
+    std::optional<wsl::shared::SocketChannel> controlChannel;
+    try
```
OneBlue: What's the reason for adding a try/catch here ?

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -178,16 +228,19 @@ static int RunShellCommand(const std::wstring& sessionName, bool verbose)
     // Relay tty output -> console (blocks until output ends).
     wsl::windows::common::relay::InterruptableRelay(ttyOut.get(), consoleOut, exitEvent.get());
 
+    // Output relay finished - signal input thread to stop and wait for process exit
+    exitEvent.SetEvent();
```
OneBlue: what's the reason for adding this ?

src/shared/inc/CommandLine.h
```diff
@@ -307,15 +307,17 @@ class ArgumentParser
...
+        m_startIndex(StartIndex), m_name(Name), m_ignoreUnknownArgs(IgnoreUnknownArgs)
     {
         m_argv.reset(CommandLineToArgvW(std::wstring(CommandLine).c_str(), &m_argc));
         THROW_LAST_ERROR_IF(!m_argv);
     }
 
 #else
 
-    ArgumentParser(int argc, const char* const* argv) : m_argc(argc), m_argv(argv), m_startIndex(1)
+    ArgumentParser(int argc, const char* const* argv, bool IgnoreUnknownArgs = false) :
```
OneBlue: nit: I'd recommend naming this argument something like `stopOnUnknownArgs`

src/windows/wsladiag/wsladiag.cpp
beena352: there isn’t a strong reason to special-case this. I’ve removed the try/catch and the optional and now treat TerminalControl as required; any failure will propagate and be handled by the existing top-level error handling.

src/windows/wsladiag/wsladiag.cpp
beena352: Yeah, the input relay is already stopped and joined via the existing scope_exit cleanup, so this extra signal was redundant. I’ve removed it and kept a single shutdown path.

src/shared/inc/CommandLine.h
OneBlue: Let's also rename the constructor argument to match the same name

OneBlue: Also stopUnknownArgs -> stopOnUnknownArgs

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -149,13 +165,14 @@ static int RunShellCommand(const std::wstring& sessionName, bool verbose)
         THROW_IF_WIN32_BOOL_FALSE(GetConsoleScreenBufferInfoEx(consoleOut, &infoEx));
 
         WSLA_TERMINAL_CHANGED message{};
-        message.Columns = infoEx.srWindow.Right - infoEx.srWindow.Left + 1;
-        message.Rows = infoEx.srWindow.Bottom - infoEx.srWindow.Top + 1;
+        message.Columns = static_cast<unsigned short>(infoEx.srWindow.Right - infoEx.srWindow.Left + 1);
```
OneBlue: What's the reason for this change ?

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -329,6 +326,11 @@ int wmain(int, wchar_t**)
     catch (...)
     {
         const auto hr = wil::ResultFromCaughtException();
+        if (hr == E_INVALIDARG)
```
OneBlue: `E_INVALIDARG` can be thrown for other reasons for I would not recommend having a specific catch block here. 

Once we have ExecutionContext based errors, we can use those to show command line parsing specific errors

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -42,9 +43,24 @@ static int ReportError(const std::wstring& context, HRESULT hr)
...
 
-// Handler for `wsladiag shell <SessionName>` (TTY-backed interactive shell).
-static int RunShellCommand(const std::wstring& sessionName, bool verbose)
+// Handler for `wsladiag shell <SessionName> [--verbose]` command - launches TTY-backed interactive shell.
+static int RunShellCommand(std::wstring_view commandLine)
 {
+    std::wstring sessionName;
+    bool verbose = false;
+
+    ArgumentParser parser(std::wstring{commandLine}, L"wsladiag", 2); // Skip "wsladiag.exe shell" to parse shell-specific args
```
ptrivedi: The third argument should be true or false for stopUnknownArgs flag

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -42,9 +43,24 @@ static int ReportError(const std::wstring& context, HRESULT hr)
...
+    parser.AddPositionalArgument(sessionName, 0);
+    parser.AddArgument(verbose, L"--verbose", L'v');
```
ptrivedi, line 47: Does this mean that verbose only applies to shell option? For session listing, looks like we took out the verbose flag?

src/shared/inc/CommandLine.h
```diff
@@ -307,15 +307,17 @@ class ArgumentParser
...
     {
         m_argv.reset(CommandLineToArgvW(std::wstring(CommandLine).c_str(), &m_argc));
         THROW_LAST_ERROR_IF(!m_argv);
     }
 
 #else
 
-    ArgumentParser(int argc, const char* const* argv) : m_argc(argc), m_argv(argv), m_startIndex(1)
+    ArgumentParser(int argc, const char* const* argv, bool stopUnknownArgs = false) :
+        m_argc(argc), m_argv(argv), m_startIndex(1), m_stopUnknownArgs(stopUnknownArgs)
```
ptrivedi: It would be more intuitive to have ignoreUnknownArgs, in line with other command parsers

beena352: I originally named it ignoreUnknownArgs, then changed it to stopOnUnknownArgs based on earlier review feedback to reflect the behavior more precisely. Given the preference for consistency with other command parsers, I’m happy to switch it back to ignoreUnknownArgs.

src/windows/wsladiag/wsladiag.cpp
beena352: earlier in the refactor --verbose was only wired up for shell as part of the per-verb parsing change. Based on this feedback, I’ve now explicitly added --verbose support back to list as well, so it’s handled consistently per-command.

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -42,9 +43,24 @@ static int ReportError(const std::wstring& context, HRESULT hr)
...
+    ArgumentParser parser(std::wstring{commandLine}, L"wsladiag", 2, false); // Skip "wsladiag.exe shell" to parse shell-specific args
+    parser.AddPositionalArgument(sessionName, 0);
+    parser.AddArgument(verbose, L"--verbose", L'v');
+
+    parser.Parse();
+
+    if (sessionName.empty())
+    {
+        THROW_HR_WITH_USER_ERROR(
+            E_INVALIDARG, wsl::shared::Localization::MessageMissingArgument(L"<SessionName>", L"wsladiag shell"));
```
ptrivedi, line 54: Why throw here and PrintMessage&return in other error return paths, @OneBlue ?

OneBlue: That's the pattern we usually use for command line related errors in WSL. 

It works nicely with localized strings and allows us to factor all the error handling in one place

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -240,8 +272,20 @@ static int RunListCommand(bool /*verbose*/)
     return 0;
 }
 
+static void PrintUsage()
```
OneBlue, line 288: FYI this is probably going to conflict with the localization PR

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -257,67 +301,95 @@ int wsladiag_main(std::wstring_view commandLine)
...
+    if (disableWarnings == nullptr || strcmp(disableWarnings, "1") != 0)
+    {
+        warningsFile = stderr;
+    }
 
-    parser.AddPositionalArgument(verb, 0);         // "list" or "shell"
-    parser.AddPositionalArgument(shellSession, 1); // session name for "shell"
-    parser.AddArgument(help, L"--help", L'h');
-    parser.AddArgument(verbose, L"--verbose", L'v');
+    std::optional<wsl::windows::common::ExecutionContext> context;
```
OneBlue: This doesn't need to be an std::optional since it's always set to the same value

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -257,67 +301,95 @@ int wsladiag_main(std::wstring_view commandLine)
...
+    context.emplace(wsl::windows::common::Context::Wsl, warningsFile);
```
OneBlue: We should add a new `Wsla` entrypoint for this

OneBlue (approved):

PR #13971, opened 2025-12-23
diagnostics: improve logging of hcs helper utilities on debug builds

benhillis:
This change adds additional debug-only tracing to help diagnose HCS / HNS issues.

### DIFF
--- a/src/windows/common/hcs.cpp
+++ b/src/windows/common/hcs.cpp
@@ -98,11 +98,11 @@ wsl::windows::common::hcs::unique_hcs_system wsl::windows::common::hcs::CreateCo
     return system;
 }
 
-std::vector<std::string> wsl::windows::common::hcs::GetProcessorFeatures()
+const std::vector<std::string>& wsl::windows::common::hcs::GetProcessorFeatures()
 {
     static std::vector<std::string> g_processorFeatures;
     static std::once_flag flag;
-    std::call_once(flag, [&]() {
+    std::call_once(flag, []() {
         ExecutionContext context(Context::HCS);
 
         wil::unique_cotaskmem_string result;
@@ -117,7 +117,7 @@ std::vector<std::string> wsl::windows::common::hcs::GetProcessorFeatures()
             THROW_HR_MSG(static_cast<HRESULT>(response.Error->Error), "%hs", response.Error->ErrorMessage.c_str());
         }
 
-        g_processorFeatures = std::move(response.Response.ProcessorFeatures);
+        g_processorFeatures = response.Response.ProcessorFeatures;
     });
 
     return g_processorFeatures;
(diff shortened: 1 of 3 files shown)


benhillis:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/microsoft/WSL/pull/13971#pullrequestreview-3609408775)

benhillis:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/microsoft/WSL/pull/13971#pullrequestreview-3609509074)

src/windows/common/hcs.cpp
```diff
@@ -98,32 +100,39 @@ wsl::windows::common::hcs::unique_hcs_system wsl::windows::common::hcs::CreateCo
...
-    {
-        THROW_HR_MSG(static_cast<HRESULT>(response.Error->Error), "%hs", response.Error->ErrorMessage.c_str());
-    }
+        const auto& response = properties.PropertyResponses.at(c_processorCapabilities);
+        if (response.Error)
+        {
+            THROW_HR_MSG(static_cast<HRESULT>(response.Error->Error), "%hs", response.Error->ErrorMessage.c_str());
+        }
+
+        g_processorFeatures = std::move(response.Response.ProcessorFeatures);
```
OneBlue: nit: this doesn't actually move the structure since `response` is a const reference

src/windows/common/hcs.cpp
```diff
@@ -98,32 +100,39 @@ wsl::windows::common::hcs::unique_hcs_system wsl::windows::common::hcs::CreateCo
 
 std::vector<std::string> wsl::windows::common::hcs::GetProcessorFeatures()
 {
```
OneBlue: now that we're keeping a static copy, we could return a const reference to avoid copying the vector every call

src/windows/common/hcs.cpp
```diff
@@ -98,32 +100,39 @@ wsl::windows::common::hcs::unique_hcs_system wsl::windows::common::hcs::CreateCo
...
+        g_processorFeatures = std::move(response.Response.ProcessorFeatures);
+    });
 
-    return response.Response.ProcessorFeatures;
+    return g_processorFeatures;
 }
 
 wsl::shared::hns::HNSEndpoint wsl::windows::common::hcs::GetEndpointProperties(HCN_ENDPOINT Endpoint)
 {
+    WSL_LOG_DEBUG("HcsGetEndpointProperties");
```
OneBlue, line 128: nit: I'm not sure how much we want to log this since it doesn't carry any specific information, and the below logline will display the full failure context

OneBlue: (Same for below calls)

benhillis: The idea is on debug builds it will give a better idea of the sequence of calls, not sure what fails.

benhillis: Im playing with this a bit more, will adjust.

benhillis: These are only logged on debug builds, I think it's fine to have them, the goal is to have a clear picture of what we are doing with HCS.

OneBlue (approved):

PR #13972, opened 2025-12-23
WSLA: Add support for Windows 10

benhillis:
We have not yet determined if we will support Windows 10 for WSLA, but in the meantime this will ensure that we are able to run to test additional features and determine feasibility. This required two changes:
1. Disabling of the .vmrs dump collection on Win10
2. Disabling of Hyper-V firewall support on Win10 for NAT networking

### DIFF
--- a/src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
+++ b/src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
@@ -120,8 +120,8 @@ WSLAVirtualMachine::~WSLAVirtualMachine()
             WI_ASSERT(std::filesystem::is_empty(m_vmSavedStateFile));
             std::filesystem::remove(m_vmSavedStateFile);
         }
+        CATCH_LOG()
     }
-    CATCH_LOG()
 
     if (m_processExitThread.joinable())
     {
--- a/src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
+++ b/src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
@@ -112,10 +112,10 @@ WSLAVirtualMachine::~WSLAVirtualMachine()
         CATCH_LOG()
     }
 
-    try
+    if (!m_vmSavedStateFile.empty() && !m_vmSavedStateCaptured)
     {
         // If the VM did not crash, the saved state file should be empty, so we can remove it.
-        if (!m_vmSavedStateFile.empty() && !m_vmSavedStateCaptured)
+        try
         {
             WI_ASSERT(std::filesystem::is_empty(m_vmSavedStateFile));
             std::filesystem::remove(m_vmSavedStateFile);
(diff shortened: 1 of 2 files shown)


src/windows/wslaservice/exe/WSLAVirtualMachine.h
```diff
@@ -149,7 +149,6 @@ class DECLSPEC_UUID("0CFC5DC1-B6A7-45FC-8034-3FA9ED73CE30") WSLAVirtualMachine
 
     GUID m_vmId{};
     std::wstring m_vmIdString;
-    wsl::windows::common::helpers::WindowsVersion m_windowsVersion = wsl::windows::common::helpers::GetWindowsVersion();
```
benhillis, line 152: There is no reason to have a member variable, this is a global static in the helper that is initialized the first time the version functions are called.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -518,11 +527,11 @@ void WSLAVirtualMachine::ConfigureNetworking()
         // TODO: refactor this to avoid using wsl config
         static wsl::core::Config config(nullptr);
 
-        // TODO-WSLA: Implement firewall logic
-        /*if (!wsl::core::MirroredNetworking::IsHyperVFirewallSupported(config))
+        // Disable HyperV Firewall if not supported
```
benhillis, line 530: This may not be perfect, but it's better than before.

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -162,12 +173,8 @@ void WSLAVirtualMachine::Start()
...
     //
     // N.B. Cold discard hint size should be a multiple of the fault cluster shift size.
-    //
-    // N.B. This is only done on builds that have the fix for the VID deadlock on partition teardown.
-    if ((m_windowsVersion.BuildNumber >= WindowsBuildNumbers::Germanium) ||
-        (m_windowsVersion.BuildNumber >= WindowsBuildNumbers::Cobalt && m_windowsVersion.UpdateBuildRevision >= 2360) ||
-        (m_windowsVersion.BuildNumber >= WindowsBuildNumbers::Iron && m_windowsVersion.UpdateBuildRevision >= 1970) ||
-        (m_windowsVersion.BuildNumber >= WindowsBuildNumbers::Vibranium_22H2 && m_windowsVersion.UpdateBuildRevision >= 3393))
+    const auto windowsVersion = wsl::windows::common::helpers::GetWindowsVersion();
+    if (windowsVersion.BuildNumber >= WindowsBuildNumbers::Germanium)
```
OneBlue, line 177: nit: I'd prefer to keep th previous checks to avoid hitting the deadlock if we ever run on older builds

OneBlue (approved):

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
benhillis: The relaxing of this check means we will fall back to the “safe but less memory efficient” behavior more often. It doesn’t prevent the deadlock.

PR #13975, opened 2025-12-23
tests: add Wsladiag CLI tests

beena352:
## Summary of the Pull Request:
This PR adds automated test coverage for the wsladiag CLI tool under test/windows.
The new tests exercise the supported commands and common error paths to ensure argument parsing, exit codes, and output behavior remain correct.

- Specifically, the tests cover:
- --help and -h behavior
- list command output (both with and without active sessions)
- shell command error paths (missing session name, invalid session name, verbose vs non-verbose)
- Unknown and empty commands

## PR Checklist

A new test file (WsladiagTests.cpp) is added to the Windows test suite. The tests invoke wsladiag.exe directly and validate:

- Correct exit codes for valid and invalid commands
- Whether output is written to stdout vs stderr as expected
- That help text and error messages are shown consistently
- That non-interactive error paths (e.g., invalid session name) fail fast and do not hang

The tests intentionally avoid interactive shell success cases and distro-dependent scenarios to keep the suite stable and fast. Assertions use substring matching rather than exact formatting to reduce flakiness.

## Validation Steps Performed
Built the solution and confirmed wsladiag.exe is generated correctly
- Ran the new test suite locally and verified all tests pass
- Tested edge cases manually (e.g., running wsladiag with no arguments, invalid commands)
- Verified test behavior in both scenarios: when WSLA sessions exist and when none exist
- Confirmed tests fail appropriately when wsladiag.exe is not available

### DIFF
--- a/src/shared/inc/CommandLine.h
+++ b/src/shared/inc/CommandLine.h
@@ -307,13 +307,9 @@ class ArgumentParser
 public:
 #ifdef WIN32
 
-<<<<<<< HEAD
     ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1, bool ignoreUnknownArgs = false) :
         m_parseIndex(StartIndex), m_name(Name), m_ignoreUnknownArgs(ignoreUnknownArgs)
-=======
-    ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1, bool IgnoreUnknownArgs = false) :
-        m_startIndex(StartIndex), m_name(Name), m_ignoreUnknownArgs(IgnoreUnknownArgs)
->>>>>>> ea162030 (Localize Wsladiag tests)
+
     {
         m_argv.reset(CommandLineToArgvW(std::wstring(CommandLine).c_str(), &m_argc));
         THROW_LAST_ERROR_IF(!m_argv);
@@ -321,13 +317,9 @@ public:
 
 #else
 
-<<<<<<< HEAD
     ArgumentParser(int argc, const char* const* argv, bool ignoreUnknownArgs = false) :
         m_argc(argc), m_argv(argv), m_parseIndex(1), m_ignoreUnknownArgs(ignoreUnknownArgs)
-=======
-    ArgumentParser(int argc, const char* const* argv, bool IgnoreUnknownArgs = false) :
-        m_argc(argc), m_argv(argv), m_startIndex(1), m_ignoreUnknownArgs(IgnoreUnknownArgs)
->>>>>>> ea162030 (Localize Wsladiag tests)
+
     {
     }
 
--- a/src/shared/inc/CommandLine.h
+++ b/src/shared/inc/CommandLine.h
@@ -307,8 +307,13 @@ class ArgumentParser
 public:
 #ifdef WIN32
 
+<<<<<<< HEAD
     ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1, bool ignoreUnknownArgs = false) :
         m_parseIndex(StartIndex), m_name(Name), m_ignoreUnknownArgs(ignoreUnknownArgs)
+=======
+    ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1, bool IgnoreUnknownArgs = false) :
+        m_startIndex(StartIndex), m_name(Name), m_ignoreUnknownArgs(IgnoreUnknownArgs)
+>>>>>>> ea162030 (Localize Wsladiag tests)
     {
         m_argv.reset(CommandLineToArgvW(std::wstring(CommandLine).c_str(), &m_argc));
         THROW_LAST_ERROR_IF(!m_argv);
@@ -316,8 +321,13 @@ public:
 
 #else
 
+<<<<<<< HEAD
     ArgumentParser(int argc, const char* const* argv, bool ignoreUnknownArgs = false) :
         m_argc(argc), m_argv(argv), m_parseIndex(1), m_ignoreUnknownArgs(ignoreUnknownArgs)
+=======
+    ArgumentParser(int argc, const char* const* argv, bool IgnoreUnknownArgs = false) :
+        m_argc(argc), m_argv(argv), m_startIndex(1), m_ignoreUnknownArgs(IgnoreUnknownArgs)
+>>>>>>> ea162030 (Localize Wsladiag tests)
     {
     }
 
--- a/src/windows/wsladiag/wsladiag.cpp
+++ b/src/windows/wsladiag/wsladiag.cpp
@@ -65,11 +65,8 @@ static int ReportError(const std::wstring& context, HRESULT hr)
     return 1;
 }
 
-<<<<<<< HEAD
 // Handler for `wsladiag shell <SessionName>` command.
-=======
-// Handler for `wsladiag shell <SessionName> [--verbose]` command - launches TTY-backed interactive shell.
->>>>>>> ea162030 (Localize Wsladiag tests)
+
 static int RunShellCommand(std::wstring_view commandLine)
 {
     std::wstring sessionName;
--- a/src/windows/wsladiag/wsladiag.cpp
+++ b/src/windows/wsladiag/wsladiag.cpp
@@ -727,6 +727,7 @@ int wmain(int, wchar_t**)
     wsl::windows::common::EnableContextualizedErrors(false);
 
     ExecutionContext context{Context::WslaDiag};
+
     int exitCode = 1;
     HRESULT result = S_OK;
 
--- a/src/windows/wsladiag/wsladiag.cpp
+++ b/src/windows/wsladiag/wsladiag.cpp
@@ -65,7 +65,11 @@ static int ReportError(const std::wstring& context, HRESULT hr)
     return 1;
 }
 
+<<<<<<< HEAD
 // Handler for `wsladiag shell <SessionName>` command.
+=======
+// Handler for `wsladiag shell <SessionName> [--verbose]` command - launches TTY-backed interactive shell.
+>>>>>>> ea162030 (Localize Wsladiag tests)
 static int RunShellCommand(std::wstring_view commandLine)
 {
     std::wstring sessionName;
@@ -727,7 +731,6 @@ int wmain(int, wchar_t**)
     wsl::windows::common::EnableContextualizedErrors(false);
 
     ExecutionContext context{Context::WslaDiag};
-
     int exitCode = 1;
     HRESULT result = S_OK;
 
--- a/test/windows/WsladiagTests.cpp
+++ b/test/windows/WsladiagTests.cpp
@@ -1,0 +1,134 @@
+/*++
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    WsladiagTests.cpp
+
+Abstract:
+
+    This file contains smoke tests for wsladiag.
+
+--*/
+
+#include "precomp.h"
+#include "Common.h"
+#include <format>
+
+namespace WsladiagTests {
+class WsladiagTests
+{
+    WSL_TEST_CLASS(WsladiagTests)
+
+    TEST_CLASS_SETUP(TestClassSetup)
+    {
+        VERIFY_ARE_EQUAL(LxsstuInitialize(FALSE), TRUE);
+        return true;
+    }
+
+    TEST_CLASS_CLEANUP(TestClassCleanup)
+    {
+        LxsstuUninitialize(FALSE);
+        return true;
+    }
+
+    static std::wstring BuildWsladiagCmd(const std::wstring& args)
+    {
+        const auto exePath = wsl::windows::common::wslutil::GetBasePath() / L"wsladiag.exe";
+        VERIFY_IS_TRUE(std::filesystem::exists(exePath));
+
+        const auto exe = exePath.wstring();
+        return args.empty() ? std::format(L"\"{}\"", exe) : std::format(L"\"{}\" {}", exe, args);
+    }
+
+    static std::tuple<std::wstring, std::wstring, int> RunWsladiag(const std::wstring& args)
+    {
+        auto cmd = BuildWsladiagCmd(args);
+        return LxsstuLaunchCommandAndCaptureOutputWithResult(cmd.data());
+    }
+
+    TEST_METHOD(List_ShowsSessionsOrNoSessions)
+    {
+        auto [out, err, code] = RunWsladiag(L"list");
+        VERIFY_ARE_EQUAL(0, code);
+        VERIFY_ARE_EQUAL(L"", err);
+
+        const bool noSessions = (out.find(L"No WSLA sessions found.") != std::wstring::npos);
+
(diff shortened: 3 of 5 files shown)


test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+    TEST_METHOD(List_ShowsSessionsOrNoSessions)
+    {
+        auto [out, err, code] = RunWsladiag(L"list");
+        VERIFY_ARE_EQUAL(0, code);
+        VERIFY_ARE_EQUAL(L"", err);
+
+        const bool noSessions = (out.find(L"No WSLA sessions found.") != std::wstring::npos);
+
+        const bool hasTable = (out.find(L"WSLA session") != std::wstring::npos) && (out.find(L"ID") != std::wstring::npos) &&
```
OneBlue: I'd recommend having a utility method to verify the complete output. Something like:

```
ValidateOutput("list", "No WSLA sessions found.\r\n");
```

For instance. That would make writing future tests easier

OneBlue: (Same for below tests)

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+    TEST_CLASS_CLEANUP(TestClassCleanup)
+    {
+        LxsstuUninitialize(FALSE);
+        return true;
+    }
+
+    static std::wstring BuildWsladiagCmd(const std::wstring& args)
+    {
+        const auto exePath = wsl::windows::common::wslutil::GetBasePath() / L"wsladiag.exe";
```
OneBlue: `GetBasePath()` won't work here since it will look up the current process's executable folder. 

To get the installation folder where wsladiag.exe is, use : `wsl::windows::common::wslutil::GetMsiPackagePath()`

OneBlue (changes requested):

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -329,6 +326,11 @@ int wmain(int, wchar_t**)
     catch (...)
     {
         const auto hr = wil::ResultFromCaughtException();
+        if (hr == E_INVALIDARG)
```
OneBlue: I would recommend against this, because the service API might return that, and in that case we wouldn't want to show the usage. 

To show command line parsing errors, I would recommend using something similar to this: https://github.com/microsoft/WSL/blob/cec99af70a0c3a5043dd2f8cddab6cb1e7d224d4/src/windows/common/WslClient.cpp#L1905 

For the user errors to work, we'd need to explicitely enable them, like this: 

```
    wsl::windows::common::EnableContextualizedErrors(false);
```

And then create an execution context that we initialize in main (probably with new entrypoint value for wsladiag): 

```
wsl::windows::common::ExecutionContext context{Context::WslaDiag};
```

Then we can use `context` to actually get the error string when an exception is thrown

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+namespace WsladiagTests {
+class WsladiagTests
+{
+    WSL_TEST_CLASS(WsladiagTests)
+
+    // Initialize the tests
+    TEST_CLASS_SETUP(TestClassSetup)
+    {
+        VERIFY_ARE_EQUAL(LxsstuInitialize(FALSE), TRUE);
```
OneBlue: This is specific to WSL and isn't needed for WSLA tests

OneBlue: Same for `        LxsstuUninitialize(FALSE);` below

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,166 @@
...
+    static std::tuple<std::wstring, std::wstring, int> RunWsladiag(const std::wstring& args)
+    {
+        auto cmd = BuildWsladiagCmd(args);
+        return LxsstuLaunchCommandAndCaptureOutputWithResult(cmd.data());
+    }
+
+    // Validate that list command output shows either no sessions message or session table
+    static void ValidateListOutput(const std::wstring& out)
+    {
+        const bool noSessions = out.find(L"No WSLA sessions found.") != std::wstring::npos;
```
OneBlue: Sorry if I was unclear in my previous comment. What I was suggesting was to have a method that takes the expected output as a parameter and do the validation. Something like: 

```
void ValidateWslaDiagOutput(const std::string& Cmd, const std::string& ExpectedOutput);
```

So we can something like: 

```
ValidateWslaDiagOutput("list", "<expected output>");

```

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+        VERIFY_ARE_EQUAL(errH, errLong);
+        ValidateUsage(errH);
+    }
+
+    // Test that unknown commands show error message and usage
+    TEST_METHOD(UnknownCommand_ShowsUsage)
+    {
+        auto [out, err, code] = RunWsladiag(L"blah");
+        VERIFY_ARE_NOT_EQUAL(0, code);
```
OneBlue: I'd recommend validating that the error code is 1 explicitly, since that's what we expect

OneBlue (changes requested):

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -257,78 +289,68 @@ int wsladiag_main(std::wstring_view commandLine)
...
+        PrintUsage();
         return 1;
     }
 }
 
 int wmain(int, wchar_t**)
 {
+    wsl::windows::common::EnableContextualizedErrors(false);
+
+    std::optional<ExecutionContext> context;
```
OneBlue: nit: `context` is always set to the same value, so it doesn't need to be an `std::optional`

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,159 @@
...
+        ValidateUsage(errH);
+    }
+
+    // Test that unknown commands show error message and usage
+    TEST_METHOD(UnknownCommand_ShowsError)
+    {
+        auto [out, err, code] = RunWsladiag(L"blah");
+        VERIFY_ARE_EQUAL(1, code);
+
+        const std::wstring combined = out + err;
```
OneBlue: nit: I'd recommend validating both fds separately. For instance error messages should always be on stderr, not stdout

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,159 @@
...
+    static std::tuple<std::wstring, std::wstring, int> RunWsladiag(const std::wstring& args)
+    {
+        auto cmd = BuildWsladiagCmd(args);
+        return LxsstuLaunchCommandAndCaptureOutputWithResult(cmd.data());
+    }
+
+    static void ValidateWslaDiagOutput(const std::wstring& cmd, const std::wstring& expectedSubstring)
+    {
+        auto [out, err, code] = RunWsladiag(cmd);
+        const std::wstring combined = out + err;
```
OneBlue: Same here, I'd recommend having two parameters, one for each output so we can validate them separately. 

I'll also recommend validating the entire strings, that way we'll catch if anything unexpected is printed

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -257,78 +289,67 @@ int wsladiag_main(std::wstring_view commandLine)
...
-        }
-        throw;
-    }
+    parser.Parse(); // Let exceptions propagate to wmain for centralized handling
 
     if (help || verb.empty())
     {
-        printUsage();
+        PrintUsage();
         return 0;
```
ptrivedi, line 698: Using early return (no else block if there's a return in 'if' block) pattern here might be better for readability and reduced indentation

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -42,9 +45,23 @@ static int ReportError(const std::wstring& context, HRESULT hr)
...
+
+    ArgumentParser parser(std::wstring{commandLine}, L"wsladiag", 2); // Skip "wsladiag.exe shell" to parse shell-specific args
+    parser.AddPositionalArgument(sessionName, 0);
+    parser.AddArgument(verbose, L"--verbose", L'v');
+
+    parser.Parse();
+
+    if (sessionName.empty())
+    {
+        THROW_HR(E_INVALIDARG);
```
ptrivedi: We should do a wslutil::PrintMessage here and return 1

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -240,6 +261,17 @@ static int RunListCommand(bool /*verbose*/)
...
 
+static void PrintUsage()
+{
+    wslutil::PrintMessage(
+        L"wsladiag - WSLA diagnostics tool\n"
+        L"Usage:\n"
+        L"  wsladiag list\n"
+        L"  wsladiag shell <SessionName> [--verbose]\n"
+        L"  wsladiag --help\n",
+        stderr);
```
ptrivedi: The usage string could be made into a const std::string or a constexpr std::string_view. Then you can use that in the ValidateUsage() method in the WsladiagTests to verify if Usage string was returned as expected by the help case or an error case.

kevpar:
@beena352 please write descriptive titles for your PRs. Don't include the branch name in the title.

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,164 @@
...
+    L"wsladiag - WSLA diagnostics tool\r\n"
+    L"Usage:\r\n"
+    L"  wsladiag list\r\n"
+    L"  wsladiag shell <SessionName> [--verbose]\r\n"
+    L"  wsladiag --help\r\n";
+
+namespace WsladiagTests {
+class WsladiagTests
+{
+    WSL_TEST_CLASS(WsladiagTests)
```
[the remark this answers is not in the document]
kevpar: @OneBlue does BinaryUnderTest actually do something useful in WSL? Most docs I see on it seem related to internal infra.

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+    // Use localized usage text at runtime
+    static std::wstring GetUsageText()
+    {
+        return std::wstring(wsl::shared::Localization::MessageWsladiagUsage());
+    }
+
+    // Test that wsladiag list command shows either sessions or "no sessions" message
+    TEST_METHOD(List_ShowsSessionsOrNoSessions)
```
kevpar: I think it's better to explicitly test that we get no list sessions output here, and then have another test to explicitly validate that `list` shows sessions when it should.

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+        VERIFY_IS_TRUE(noSessions || hasTable);
+    }
+
+    // Validate that usage information contains expected command descriptions
+    static void ValidateUsage(const std::wstring& err)
+    {
+        const std::wstring nerr = NormalizeForCompare(err);
+        const std::wstring usage = NormalizeForCompare(GetUsageText());
+        VERIFY_IS_TRUE(nerr.find(usage) != std::wstring::npos);
```
kevpar: Might make more sense to check if the strings are exactly equal here.

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,179 @@
...
+        auto [out, err, code] = RunWsladiag(L"shell");
+        VERIFY_ARE_EQUAL(1, code);
+        VERIFY_ARE_EQUAL(L"", out);
+        const std::wstring missingArgMsg =
+            std::wstring(wsl::shared::Localization::MessageMissingArgument(L"<SessionName>", L"wsladiag shell"));
+        VERIFY_IS_TRUE(NormalizeForCompare(err).find(NormalizeForCompare(missingArgMsg)) != std::wstring::npos);
+    }
+
+    // Test shell command with invalid session name (silent mode)
+    TEST_METHOD(Shell_InvalidSessionName_Silent)
```
kevpar: What does "silent mode" mean here?

test/windows/WsladiagTests.cpp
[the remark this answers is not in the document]
OneBlue: I've always assumed that TE would watch said binaries in case they crash and fail the test case if that happens, but I'm not actually not sure that it is the case

test/windows/WsladiagTests.cpp
beena352: by silent mode, I meant the default non-verbose behavior. I’ll rename this to “no verbose output” to make it clearer.

test/windows/WsladiagTests.cpp
beena352: Sounds good, I’ll update the test to check for exact string equality here.

src/shared/inc/CommandLine.h
```diff
@@ -309,6 +309,7 @@ class ArgumentParser
 
     ArgumentParser(const std::wstring& CommandLine, LPCWSTR Name, int StartIndex = 1, bool ignoreUnknownArgs = false) :
         m_parseIndex(StartIndex), m_name(Name), m_ignoreUnknownArgs(ignoreUnknownArgs)
+
```
kevpar, line 312: Are these new lines intentional?

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -66,6 +66,7 @@ static int ReportError(const std::wstring& context, HRESULT hr)
 }
 
 // Handler for `wsladiag shell <SessionName>` command.
+
```
kevpar, line 69: Same here. Seems like something is inserting extra newlines?

test/windows/WsladiagTests.cpp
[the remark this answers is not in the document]
kevpar: @beena352 for completeness, we should specify TEST_CLASS_PROPERTY(L"BinaryUnderTest", L"wsladiag.exe") here.

test/windows/WsladiagTests.cpp
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    // Test that wsladiag list command "no sessions" message
+    TEST_METHOD(List_NoSessions)
+    {
+        auto [stdOut, stdErr, exitCode] = RunWsladiag(L"list");
+
+        VERIFY_ARE_EQUAL(0, exitCode);
+        VERIFY_ARE_EQUAL(L"", stdErr);
+        VERIFY_ARE_EQUAL(std::wstring(wsl::shared::Localization::MessageWslaNoSessionsFound()) + L"\r\n", stdOut);
```
kevpar, line 86: Should we use ValidateWsladiagOutput here as well?

src/shared/inc/CommandLine.h
beena352: Not intentional. I started editing this section but ended up not needing the change, and I missed removing the extra newline. I’ll clean it up.

src/windows/wsladiag/wsladiag.cpp
beena352: Not intentional. I started editing this section but ended up not needing the change, and I missed removing the extra newline. I’ll clean it up.

benhillis (changes requested):
wsladiag.exe is being replace with wslc.exe. I don't think it's a good use of time to add tests for something we are removing soon.

beena352:
closing this PR as we are to moving to wslc.exe.

PR #13978, opened 2025-12-24
Switch from nerdctl to dockerd  

OneBlue:

This change switches the WSLAService logic to use the dockerd API via its unix socket HTTP endpoint 

## PR Checklist


src/linux/init/WSLAInit.cpp
```diff
@@ -182,6 +182,106 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_OPEN& Mes
...
+    sendResult.reset();
+
+    // Relay data between the two sockets.
+    pollfd pollDescriptors[2];
+    pollDescriptors[0].fd = socket.get();
+    pollDescriptors[0].events = POLLIN;
+    pollDescriptors[1].fd = Channel.Socket();
+    pollDescriptors[1].events = POLLIN;
+
+    std::vector<gsl::byte> relayBuffer;
```
benhillis, line 222: Have you tried tuning the buffer size?

src/windows/common/relay.cpp
```diff
@@ -1221,16 +1314,256 @@ void WriteHandle::Collect()
...
+            ActiveHandle = &WriteStderr;
+        }
+        else
+        {
+            THROW_HR_MSG(E_UNEXPECTED, "Unexpected Docker IO multiplexed header fd: %u", header->Fd);
+        }
+
+        // Consume the header.
+        PendingBuffer.erase(PendingBuffer.begin(), PendingBuffer.begin() + sizeof(MultiplexedHeader));
+    }
```
benhillis, line 1610: these handle wrapper tests would be nice to have some unit tests

src/linux/init/WSLAInit.cpp
OneBlue: Not particularly. `UtilReadBuffer` will resize the vector to the default (INITIAL_MESSAGE_BUFFER_SIZE), which seems to be working well. 

This is probably part of the things that we'll have to benchmark and tune up

src/windows/common/WslClient.cpp
```diff
@@ -1657,70 +1673,149 @@ int WslaShell(_In_ std::wstring_view commandLine)
...
+        io.AddHandle(std::make_unique<RelayHandle>(process->GetStdHandle(1), GetStdHandle(STD_OUTPUT_HANDLE)));
+        io.AddHandle(std::make_unique<RelayHandle>(process->GetStdHandle(2), GetStdHandle(STD_ERROR_HANDLE)));
 
+        io.Run({});
+    }
+    else
     {
-        // Create a thread to relay stdin to the pipe.
+        // Save original console modes so they can be restored on exit.
+        DWORD OriginalInputMode{};
```
benhillis, line 1753: This is almost identical to the svccomio class. I don't like that class name much, but we should probably have a single utility class for setting and restoring console mode.

src/windows/inc/docker_schema.h
```diff
@@ -0,0 +1,153 @@
...
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    docker_schema.h
+
+Abstract:
+
+    JSON schema for the docker API.
```
benhillis, line 11: Where is this from? Maybe a link in the comments.

src/windows/wslaservice/exe/ContainerEventTracker.cpp
```diff
@@ -83,83 +81,113 @@ ContainerEventTracker::~ContainerEventTracker()
 void ContainerEventTracker::OnEvent(const std::string& event)
 {
     // TODO: log session ID
-    WSL_LOG("NerdCtlEvent", TraceLoggingValue(event.c_str(), "Data"));
```
benhillis, line 86: if you haven't, probably worth doing a search for "nerdctl" and removing any refs.

src/windows/wslaservice/exe/DockerHTTPClient.cpp
```diff
@@ -0,0 +1,447 @@
...
+    This file contains the implementation of the Docker HTTP client.
+
+--*/
+
+#include "precomp.h"
+
+#include "DockerHTTPClient.h"
+
+namespace http = boost::beast::http;
+using boost::beast::http::verb;
```
benhillis, line 31: this is our first usage of boost, are we sure we're ok using it?

src/windows/wslaservice/exe/DockerHTTPClient.cpp
```diff
@@ -0,0 +1,447 @@
...
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    DockerHTTPClient.cpp
+
+Abstract:
+
+    This file contains the implementation of the Docker HTTP client.
```
benhillis, line 11: I think this entire file would benefit from some more comments.

src/windows/common/WslClient.cpp
OneBlue: I agree. All this method is meant to be replaced by `wsladiag.exe` in the long run so that where we should do this refactoring.

src/windows/inc/docker_schema.h
OneBlue: It's from: https://docs.docker.com/reference/api/engine/version/v1.52/#tag/Container/operation/ContainerCreate 

I'll add a link in the header comment

src/windows/wslaservice/exe/DockerHTTPClient.cpp
OneBlue: I think so, the boost license is essentially MIT, but more permissive. 

I'll add a tracking entry in our dependency manifest though

src/windows/common/relay.cpp
OneBlue: That's a good point. Let me see if I can think of an easy way to add test coverage for those

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -387,7 +387,27 @@ void WSLAVirtualMachine::ConfigureMounts()
...
+        "perf_event",
+        "net_prio",
+        "hugetlb",
+        "pids",
+        "rdma"};
+
+    for (const auto* e : cgroups)
+    {
+        Mount(m_initChannel, nullptr, std::format("/sys/fs/cgroup/{}", e).c_str(), "cgroup", e, 0);
+    }
```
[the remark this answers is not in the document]
benhillis: Interesting, docker requires v1?

benhillis (commented):
looking good, minor comments.

src/linux/init/WSLAInit.cpp
```diff
@@ -182,6 +182,106 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_OPEN& Mes
...
+        {
+            auto bytesRead = UtilReadBuffer(pollDescriptors[0].fd, relayBuffer);
+            if (bytesRead < 0)
+            {
+                LOG_ERROR("read failed {}", errno);
+                break;
+            }
+            else if (bytesRead == 0)
+            {
+                // Unix socket has been closed.
```
ptrivedi, line 238: where is close() on this socket being called?

src/linux/init/WSLAInit.cpp
```diff
@@ -182,6 +182,106 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_OPEN& Mes
...
+    std::vector<gsl::byte> relayBuffer;
+    while (true)
+    {
+        auto result = poll(pollDescriptors, COUNT_OF(pollDescriptors), -1);
+        THROW_LAST_ERROR_IF(result < 0);
+
+        if (pollDescriptors[0].revents & (POLLIN | POLLHUP | POLLERR))
+        {
+            auto bytesRead = UtilReadBuffer(pollDescriptors[0].fd, relayBuffer);
+            if (bytesRead < 0)
```
ptrivedi, line 231: Assuming that UtilReadBuffer takes care of things like EWOULDBLOCK or EAGAIN returns

OneBlue: It does. See: https://github.com/microsoft/WSL/blob/6f47fa9b202fe41704e3c559b38b33a64aaa1387/src/linux/init/util.cpp#L2267

src/linux/init/WSLAInit.cpp
OneBlue: `socket` is closed automatically when it goes out of scope  and `Channel` is closed by `closeChannel()`

src/windows/common/relay.hpp
```diff
@@ -159,6 +159,60 @@ enum class IOHandleStatus
...
+    {
+        return Handle;
+    }
+
+    void Reset()
+    {
+        if (OnClose != nullptr)
+        {
+            OnClose();
+            OnClose = {};
```
ptrivedi: should set this to nullptr to be consistent with the nullptr check above

src/windows/common/relay.hpp
```diff
@@ -159,6 +159,60 @@ enum class IOHandleStatus
...
+    HandleWrapper(HANDLE handle, std::function<void()>&& OnClose = []() {}) : Handle(handle), OnClose(std::move(OnClose))
+    {
+    }
+
+    HandleWrapper(
+        wil::unique_hfile&& handle, std::function<void()>&& OnClose = []() {}) :
+        OwnedHandle(handle.release()), Handle(OwnedHandle.get()), OnClose(std::move(OnClose))
+    {
+    }
+
```
ptrivedi, line 188: copy semantics should be disabled for this object (deleted copy constructor and assignment operator)

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -432,6 +432,26 @@ void WSLAVirtualMachine::Start()
     Mount(m_initChannel, nullptr, "/sys/fs/cgroup", "cgroup2", "", 0);
```
benhillis: I think this was not merged correctly, I think you want to mount a tmpfs instead of cgroup2.

src/windows/common/relay.cpp
```diff
@@ -967,7 +971,7 @@ bool MultiHandleWait::Run(std::optional<std::chrono::milliseconds> Timeout)
         // Remove completed handles from m_handles.
         std::erase_if(m_handles, [&](const auto& e) { return e->GetState() == IOHandleStatus::Completed; });
 
-        if (m_handles.empty())
+        if (m_handles.empty() || m_cancel)
```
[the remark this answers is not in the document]
OneBlue: Yes, that's by design

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -511,6 +597,10 @@ CATCH_RETURN();
 
 void WSLASession::OnUserSessionTerminating()
 {
+    // m_sessionTerminatingEvent is always valid, so it can be signalled with the lock.
+    // This allows a sesssion to be unblocked if a stuck operation is holding the lock.
+    m_sessionTerminatingEvent.SetEvent();
```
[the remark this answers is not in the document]
OneBlue: That's fine. The design here is to unblock any pending operation that might be holding the lock

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
OneBlue: Good catch, thank you

src/windows/common/relay.cpp
OneBlue: Added in latest iteration. I actually found a few issues writing tests !

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -471,150 +509,120 @@ void wsl::windows::service::wsla::WSLAContainerImpl::UnmountVolumes(const std::v
...
+    catch (const DockerHTTPException& e)
     {
-        args.push_back("--");
-
-        for (ULONG i = 0; i < options.InitProcessOptions.CommandLineCount; i++)
-        {
-            args.push_back(options.InitProcessOptions.CommandLine[i]);
-        }
+        // TODO: propagate error message to caller.
+        THROW_HR_MSG(E_FAIL, "Failed to create container: %hs ", e.what());
```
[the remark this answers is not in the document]
OneBlue: That's the only thing we can do until we do an API change

test/windows/WSLATests.cpp
```diff
@@ -52,7 +52,9 @@ class WSLATests
         testVhd = std::filesystem::path{vhdPath} / "ext4.vhdx";
         storagePath = std::filesystem::current_path() / "test-storage";
 
-        WslShutdown();
+        auto session = CreateSession();
+        VERIFY_SUCCEEDED(session->PullImage("debian:latest", nullptr, nullptr));
+        VERIFY_SUCCEEDED(session->PullImage("python:3.12-alpine", nullptr, nullptr));
```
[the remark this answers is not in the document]
OneBlue: That's not true. VERIFY_SUCEEDED does that

benhillis (approved):
Looks good to me, there are some things that can be done as a follow-up but since this is a large change we should get this in and iterate.

PR #14028, opened 2026-01-06
WSLA: Use latest avaialble schema version when creating the VM

benhillis:
This change updates VM creation to use the latest available schema version provided by the host OS. This ensures that we request a schema version that is supported by HCS running on the host.


OneBlue (changes requested):
I don't think we should do this unless there's a definite benefit in using the latest version of the schema.

If there's ever a breaking change in the HCS schema (which is how I interpret the major version field), that would could break us

benhillis:
> I don't think we should do this unless there's a definite benefit in using the latest version of the schema.
> 
> If there's ever a breaking change in the HCS schema (which is how I interpret the major version field), that would could break us

From chatting with the HCS folks it looks like you're right, major version could technically introduce breaking changes.

PR #14029, opened 2026-01-06
WSLA: Update logging infra to only log to hvc1 if available

benhillis:
This change gets rid of an error in kmesg about /dev/hvc1 if it is not present (which is it not for WSLA), and updates the GNS error logging macro to log directly to kmesg instead of the telemetry channel if it is not present.

### DIFF
--- a/src/linux/init/common.h
+++ b/src/linux/init/common.h
@@ -137,15 +137,23 @@ auto LogImpl(int fd, const std::format_string<Args...>& format, Args&&... args)
     }
 
 #define GNS_LOG_INFO(str, ...) \
-    if (g_TelemetryFd != -1) { \
-        LogImpl(g_TelemetryFd, "{}: {} - " str "\n", g_threadName.c_str(), __FUNCTION__, ##__VA_ARGS__); \
+    { \
+        if (g_TelemetryFd != -1) \
+        { \
+            LogImpl(g_TelemetryFd, "{}: {} - " str "\n", g_threadName.c_str(), __FUNCTION__, ##__VA_ARGS__); \
+        } \
     }
 
 #define GNS_LOG_ERROR(str, ...) \
-    if (g_TelemetryFd != -1) { \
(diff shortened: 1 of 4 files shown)


src/linux/init/common.h
```diff
@@ -138,12 +138,22 @@ auto LogImpl(int fd, const std::format_string<Args...>& format, Args&&... args)
 
 #define GNS_LOG_INFO(str, ...) \
     { \
-        LogImpl(g_TelemetryFd, "{}: {} - " str "\n", g_threadName.c_str(), __FUNCTION__, ##__VA_ARGS__); \
+        if (g_TelemetryFd != -1) \
+        { \
+            LogImpl(g_TelemetryFd, "{}: {} - " str "\n", g_threadName.c_str(), __FUNCTION__, ##__VA_ARGS__); \
+        } \
```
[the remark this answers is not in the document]
benhillis: this was by design

src/linux/init/main.cpp
```diff
@@ -1643,11 +1643,12 @@ Return Value:
     }
 
     // Initialize logging to the hvc console device responsible for logging telemetry.
+    // If the device is not present, error messages will be logged to kmesg.
     if (UtilIsUtilityVm())
     {
         devicePath = DEVFS_PATH "/" LX_INIT_HVC_TELEMETRY;
         g_TelemetryFd = TEMP_FAILURE_RETRY(open(devicePath, (O_WRONLY | O_CLOEXEC)));
-        if (g_TelemetryFd < 0)
+        if (g_TelemetryFd < 0 && errno != ENODEV)
```
OneBlue, line 1651: I have a small preference towards passing an argument on whether this should be done or not. That will make it easier to catch if this ever gets broken

OneBlue (approved):

src/linux/init/main.cpp
benhillis: I looked into doing this, but it's kind of difficult now because WSL and WSLA share an entrypoint. If we change that in the future, yeah we can do this.

PR #14032, opened 2026-01-07
Update eLxr distribution name

rfu-cn:
Keep command-line install/unregister eLxr operations consistent

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -201,7 +201,7 @@
         ],
         "eLxr": [
             {
-                "Name": "eLxr-12",
+                "Name": "eLxr",
                 "FriendlyName": "eLxr 12.12.0.0 GNU/Linux",
                 "Default": true,
                 "Amd64Url": {


OneBlue:
@rfu-cn: what's the background for this change ? 

To give some context, since `elxr-12` is the default, users can use `wsl --install elxr` or `wsl --install elxr-12` interchangeably right now (this is the pattern that other distribution uses)

rfu-cn:
> @rfu-cn: what's the background for this change ?
> 
> To give some context, since `elxr-12` is the default, users can use `wsl --install elxr` or `wsl --install elxr-12` interchangeably right now (this is the pattern that other distribution uses)

Hi @OneBlue,
Thank you very much for your patience and for taking the time to explain this.
While wsl --install elxr and wsl --install elxr-12 both work correctly at the moment, I noticed a small inconsistency during removal. After installation, running:
wsl --terminate eLxr
wsl --unregister eLxr

results in the following error:

There is no distribution with the supplied name.
Error code: Wsl/Service/WSL_E_DISTRO_NOT_FOUND

Only the commands using eLxr-12 work as expected:
wsl --terminate eLxr-12
wsl --unregister eLxr-12

Since there is currently only a single eLxr version, I think removing the -12 suffix may make the user experience simpler and avoid potential confusion. Using "eLxr" alone should be unambiguous at this stage.
If additional versions are added in the future, I can then introduce version-specific suffixes for each item as appropriate.
I apologize for not considering this more carefully in the initial submission, and I appreciate your understanding and guidance.

BR/FuRong

DoctorWho8:
I was able to install the created distribution from the tar file stored elsewhere. I can now log into it via the terminal app.

rfu-cn:
> I was able to install the created distribution from the tar file stored elsewhere. I can now log into it via the terminal app.

Hi @DoctorWho8 
Thank you for verifying this.

I agree that the generated rootfs itself appears to be working correctly.

The behavior I am observing seems limited to the wsl --install / item-based installation path. In that case, although both elxr and elxr-12 can be used for installation, the distribution ends up being registered under the name elxr-12. As a result, management commands such as wsl --terminate and wsl --unregister only succeed when using that name.

Given that there is currently only a single eLxr item, I was wondering whether it might be preferable to register the distribution simply as eLxr for now, to keep installation and management commands consistent. Version details could still be reflected in the friendly name and package version.

Apologies for not considering this more carefully in the initial submission, and thank you for your guidance.

DoctorWho8:
> > I was able to install the created distribution from the tar file stored elsewhere. I can now log into it via the terminal app.
> 
> Hi @DoctorWho8 Thank you for verifying this.
> 
> I agree that the generated rootfs itself appears to be working correctly.
> 
> The behavior I am observing seems limited to the wsl --install / item-based installation path. In that case, although both elxr and elxr-12 can be used for installation, the distribution ends up being registered under the name elxr-12. As a result, management commands such as wsl --terminate and wsl --unregister only succeed when using that name.
> 
> Given that there is currently only a single eLxr item, I was wondering whether it might be preferable to register the distribution simply as eLxr for now, to keep installation and management commands consistent. Version details could still be reflected in the friendly name and package version.
> 
> Apologies for not considering this more carefully in the initial submission, and thank you for your guidance.

Hello!
I agree. More work is needed. But it looks promising.

OneBlue (approved):
Ok thank you for the explanation. LGTM then

craigloewen-msft (approved):

PR #14035, opened 2026-01-07
Fix dmesg warning from enable crash dump collection

kvega005:
This change eliminates an unnecessary write to `/proc/sys/kernel/core_pattern` in WSLAInit after the chroot operation. Crash dump collection is already configured at the `WSLAEntryPoint` before chroot, so the post-chroot write adds no value.

After chroot, /proc is not mounted in the new root, causing the write to fail. The failed write previously logged an error, creating noise without functional impact.

Additionally, this PR moves the WSLAEnableCrashDumpCollection call to later in WSLAEntryPoint so that logging is fully initialized when crash dump collection is enabled.

## PR Checklist

### DIFF
--- a/src/linux/init/WSLAInit.cpp
+++ b/src/linux/init/WSLAInit.cpp
@@ -855,7 +855,7 @@ int WSLAEntryPoint(int Argc, char* Argv[])
     }
 
     //
-    // Enable dump colleciton when processes crash.
+    // Enable dump collection when processes crash.
     //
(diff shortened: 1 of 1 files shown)


OneBlue (approved):
LGTM. Let's just fix the typo that copilot pointed out and we're good to merge this

benhillis:
Looks like the test isn't passing.

benhillis (approved):

OneBlue (approved):

PR #14037, opened 2026-01-07
Add inventory-as-code metadata

florelis:
See https://aka.ms/inventory-as-code


OneBlue (approved):

benhillis (approved):

florelis:
Turns out this isn't needed because inventory-as-code doesn't support GitHub.com repos...
GitHub repos need to be inventoried using an internal portal instead, even though the docs don't mention it

PR #14042, opened 2026-01-08
TEST: Add testcases to validate bridged networking connectivity

benhillis:
This will help ensure bridged networking doesn't get broken by future system distro vhd changes, for example.

### DIFF
--- a/test/windows/NetworkTests.cpp
+++ b/test/windows/NetworkTests.cpp
@@ -4486,6 +4486,12 @@ class BridgedTests
         WSL2_TEST_ONLY();
         WINDOWS_11_TEST_ONLY();
 
+        if (!NetworkTests::HostHasInternetConnectivity(AF_INET) && !NetworkTests::HostHasInternetConnectivity(AF_INET6))
+        {
+            LogSkipped("Host does not have internet connectivity. Skipping...");
+            return;
+        }
+
         m_config->Update(LxssGenerateTestConfig({.networkingMode = wsl::core::NetworkingMode::Bridged, .vmSwitch = L"Default Switch"}));
 
         // Verify that we have a working connection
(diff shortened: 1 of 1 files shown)


test/windows/NetworkTests.cpp
```diff
@@ -4480,6 +4480,33 @@ class BridgedTests
         auto [out, _] = LxsstuLaunchWslAndCaptureOutput(L"cat /proc/sys/net/ipv6/conf/all/disable_ipv6");
         VERIFY_ARE_EQUAL(L"0\n", out);
     }
+
+    TEST_METHOD(SmokeTest)
+    {
+        WSL2_TEST_ONLY();
+        WINDOWS_11_TEST_ONLY();
+
+        m_config->Update(LxssGenerateTestConfig({.networkingMode = wsl::core::NetworkingMode::Bridged, .vmSwitch = L"Default Switch"}));
```
OneBlue, line 4495: Unfortunately I'm not sure that we want those to run by default, because those rely on a DHCP server running on the default switch network.

We can see what will happen in the pipeline, but I'm not sure that those will be reliable enough to be enabled by default

benhillis: Ok, they pass locally so I can make them debug only or something.

benhillis: They're passing in the pipeline, and from my research the default hyper-v switch has a DHCP server. I think we can enable these.

OneBlue: Ok in that case this is probably OK. We can always revisit if we see failures in the future

test/windows/NetworkTests.cpp
```diff
@@ -4480,6 +4480,33 @@ class BridgedTests
...
+
+    TEST_METHOD(SmokeTest)
+    {
+        WSL2_TEST_ONLY();
+        WINDOWS_11_TEST_ONLY();
+
+        m_config->Update(LxssGenerateTestConfig({.networkingMode = wsl::core::NetworkingMode::Bridged, .vmSwitch = L"Default Switch"}));
+
+        // Verify that we have a working connection
+        NetworkTests::GuestClient(L"tcp-connect:bing.com:80");
```
OneBlue, line 4498: nit: we should probably only do this of the host has either ipv4 or ipv6 connectivity

OneBlue (approved):

OneBlue (approved):

PR #14048, opened 2026-01-09
 diagnostics: combine wprp files in a single one with profiles

benhillis:
This pull refactors all the separate .wprp files into a single one that will be easier to maintain.


diagnostics/collect-wsl-logs.ps1
```diff
@@ -11,32 +11,46 @@ Set-StrictMode -Version Latest
...
+    if (Test-Path "$PSScriptRoot/wsl.wprp")
+    {
+        Write-Host -ForegroundColor Yellow "Using local copy of wsl.wprp"
+        Copy-Item "$PSScriptRoot/wsl.wprp" $wprpFile
     }
-    catch {
-        throw
+    else
+    {
+        Write-Host -ForegroundColor Yellow "wsl.wprp not found in the current directory. Downloading it from GitHub."
```
OneBlue: I'd have a small opinion not to print anything here, since that's the default usecase and user can be easily confused by what we print on stdout. 

Ideally I'd prefer to keep the minimal output that we have today unless something goes wrong

OneBlue (approved):

OneBlue (approved):

PR #14049, opened 2026-01-10
Update to Azure Linux 3

benhillis:
This pull request updates system distro VHD used by WSL to Azure Linux 3. This required updates to the DHCP client implementation in the Linux initialization code, switching from `dhclient` to `dhcpcd` .

* Updated the `Microsoft.WSLg` package version from `1.0.71` to `1.0.72` in `packages.config`.
* Replaced all references to `dhclient` with `dhcpcd` in `src/linux/init/main.cpp`.

### DIFF
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -1252,6 +1252,7 @@ Return Value:
 
         std::string Config = std::format(
             "option subnet_mask, routers, broadcast, domain_name, domain_name_servers, domain_search, host_name, interface_mtu\n"
+            "noarp\n"
             "timeout {}\n",
             DhcpTimeout);
 
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -1251,8 +1251,7 @@ Return Value:
         //
 
         std::string Config = std::format(
-            "option subnet_mask, routers, domain_name, domain_name_servers, domain_search, host_name, interface_mtu\n"
-            "noarp\n"
+            "option subnet_mask, routers, broadcast, domain_name, domain_name_servers, domain_search, host_name, interface_mtu\n"
             "timeout {}\n",
             DhcpTimeout);
 
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -72,8 +72,9 @@ Abstract:
 #define CROSS_DISTRO_SHARE_PATH "/mnt/wsl"
 #define DEVFS_PATH "/dev"
 #define DEVNULL_PATH DEVFS_PATH "/null"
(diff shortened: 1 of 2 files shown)


src/linux/init/main.cpp
```diff
@@ -1243,50 +1245,29 @@ Return Value:
...
         //
 
         std::string Config = std::format(
-            "request subnet-mask, broadcast-address, routers,"
-            "domain-name, domain-name-servers, domain-search, host-name,"
-            "interface-mtu;\n"
-            "timeout {};\n",
+            "option subnet_mask, routers, domain_name, domain_name_servers, domain_search, host_name, interface_mtu\n"
+            "noarp\n"
+            "timeout {}\n",
```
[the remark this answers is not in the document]
benhillis: Nope, we want it to write resolv.conf.

src/linux/init/main.cpp
```diff
@@ -1243,50 +1245,29 @@ Return Value:
...
-        // Write the dhclient.conf config file.
+        // Write the dhcpcd.conf config file.
         //
 
         std::string Config = std::format(
-            "request subnet-mask, broadcast-address, routers,"
-            "domain-name, domain-name-servers, domain-search, host-name,"
-            "interface-mtu;\n"
-            "timeout {};\n",
+            "option subnet_mask, routers, domain_name, domain_name_servers, domain_search, host_name, interface_mtu\n"
```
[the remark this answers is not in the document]
benhillis: @OneBlue - we probably want this? I dropped this by accident but not sure if we need it.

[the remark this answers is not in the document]
OneBlue: I'd keep it. I think it's pretty rare for DHCP servers to have custom broadcast addresses, but doesn't hurt to have request it

src/linux/init/main.cpp
```diff
@@ -1243,50 +1245,29 @@ Return Value:
...
+            "noarp\n"
```
OneBlue: What's the thinking for disabling ARP here ?

benhillis: Reading through the options it sounded like we wouldn’t need this, but I’ll think about it a little more

benhillis: Looks like this is actually required, dhcpcd fails if this option is not present, I'm not getting any errors in dmesg it is just timing out. I'm assuming maybe the hyper-v default switch doesn't support arp?

OneBlue: Weird, the documentation says that this is just an ARP based check for IP duplication after DHCP. It probably doesn't matter too much for our usecase though

OneBlue (approved):

PR #14057, opened 2026-01-13
Localization change from build: 137487010

OneBlue:


benhillis (approved):

saydemr:
hi @benhillis & @OneBlue, Copilot correctly identifies these two errors above. Would that require separate issue or could you quickly fix those as well?

OneBlue:
> hi @benhillis & @OneBlue, Copilot correctly identifies these two errors above. Would that require separate issue or could you quickly fix those as well?

@saydemr: A separate issue would be the best way for us to track this.

PR #14062, opened 2026-01-14
Initial delete container image

yao-msft:
Initial delete container image implementation

## PR Checklist

## Validation Steps Performed
Manual validation and unit tests


src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -427,10 +427,22 @@ try
...
 CATCH_RETURN();
 
-HRESULT WSLASession::DeleteImage(LPCWSTR Image)
+HRESULT WSLASession::DeleteImage(LPCSTR Image, BOOL Force)
+try
 {
-    return E_NOTIMPL;
+    RETURN_HR_IF_NULL(E_POINTER, Image);
+
+    auto [repo, tag] = ParseImage(Image);
```
OneBlue: Looks like this is unused

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -427,10 +427,22 @@ try
...
+
+    std::lock_guard lock{m_lock};
+
+    auto deletedImages = m_dockerClient->DeleteImage(Image, !!Force);
```
OneBlue: We should handle DockerHTTPException here and return the proper error codes & message when possible (for instance when the image doesn't exist, we should  return `WSLA_E_IMAGE_NOT_FOUND`

Here's an example of HTTP error handling: https://github.com/microsoft/WSL/blob/f90c8e8e52ff0ea269ad37b25f706aa16ccc6147/src/windows/wslaservice/exe/WSLASession.cpp#L528C4-L551C6

src/windows/wslaservice/exe/DockerHTTPClient.cpp
```diff
@@ -77,6 +77,12 @@ std::vector<docker_schema::Image> DockerHTTPClient::ListImages()
     return Transaction<docker_schema::EmptyRequest, std::vector<docker_schema::Image>>(verb::get, "http://localhost/images/json");
 }
 
+std::vector<docker_schema::DeletedImage> wsl::windows::service::wsla::DockerHTTPClient::DeleteImage(const char* Name, bool Force)
```
OneBlue: Do we expect `name` to be a tag, an id, or both ?

yao-msft: Docker api said it could be either, should I rename the variable to be ImageNameOrId?

OneBlue: Either that or a comment make sense imo

test/windows/WSLATests.cpp
```diff
@@ -283,7 +283,14 @@ class WSLATests
             tags.push_back(e.Image);
         }
 
-        VERIFY_IS_TRUE(std::ranges::find(tags, Image) != tags.end());
+        if (Present)
+        {
+            VERIFY_IS_TRUE(std::ranges::find(tags, Image) != tags.end());
```
benhillis, line 289: nit: could use VERIFY_IS_EQUAL

yao-msft: This is to find an image name from a list. I'm not sure VERIFY_IS_EQUAL is the better usage here. Or maybe I'm missing something?

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -490,10 +490,47 @@ try
...
+        else if (e.StatusCode() == 409)
+        {
+            THROW_WIN32_MSG(ERROR_SHARING_VIOLATION, "%hs", errorMessage.c_str());
+        }
+        else
+        {
+            THROW_HR_MSG(E_FAIL, "%hs", errorMessage.c_str());
+        }
+
+        THROW_HR_IF_MSG(E_FAIL, deletedImages.empty(), "Failed to delete image: %hs", Image);
```
OneBlue: This THROW is unreachable

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -490,10 +490,47 @@ try
...
+    try
+    {
+        deletedImages = m_dockerClient->DeleteImage(Image, !!Force);
+    }
+    catch (const DockerHTTPException& e)
+    {
+        std::string errorMessage;
+        if ((e.StatusCode() >= 400 && e.StatusCode() < 500))
+        {
+            errorMessage = e.DockerMessage<docker_schema::ErrorResponse>().message;
```
OneBlue, line 518: We should return this error message back to the caller, like this: https://github.com/microsoft/WSL/blob/fafb09c53d1a64be2edd7b1cea108949a528aedf/src/windows/wslaservice/exe/WSLASession.cpp#L343C9-L352C1

test/windows/WSLATests.cpp
```diff
@@ -384,6 +391,38 @@ class WSLATests
...
+        ExpectImagePresent(*session, "hello-world:latest");
+
+        VERIFY_SUCCEEDED(session->DeleteImage("hello-world:latest", TRUE));
+
+        // Verify that the image is no longer in the list of images.
+        ExpectImagePresent(*session, "hello-world:latest", false);
+
+        // Test delete failed if image not exists.
+        VERIFY_ARE_EQUAL(WSLA_E_IMAGE_NOT_FOUND, session->DeleteImage("hello-world:latest", TRUE));
+    }
```
OneBlue, line 440: nit: I'd recommend adding test coverage for trying to delete an image that's in use with and without force

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -490,10 +490,47 @@ try
 }
 CATCH_RETURN();
 
-HRESULT WSLASession::DeleteImage(LPCWSTR Image)
+HRESULT WSLASession::DeleteImage(LPCSTR Image, BOOL Force)
```
OneBlue: We should probably the deleted / untagged image ID's, since the CLI will need it to print them on stdout

src/windows/wslaservice/exe/WSLASession.cpp
yao-msft: Does that mean all runtime apis will have ErrorInfo object returned?

src/windows/wslaservice/exe/WSLASession.cpp
OneBlue: (still unreachable)

yao-msft: Fixed in latest commit

test/windows/WSLATests.cpp
```diff
@@ -383,6 +390,54 @@ class WSLATests
...
+        WSLA_DELETE_IMAGE_OPTIONS options{};
+        options.Image = "alpine:latest";
+        options.Force = FALSE;
+        wil::unique_cotaskmem_array_ptr<WSLA_DELETED_IMAGE_INFORMATION> deletedImages;
+
+        VERIFY_ARE_EQUAL(
+            HRESULT_FROM_WIN32(ERROR_SHARING_VIOLATION),
+            session->DeleteImage(&options, deletedImages.addressof(), deletedImages.size_address<ULONG>(), nullptr));
+
+        // Force should suuceed.
```
OneBlue, line 428: nit: typo

test/windows/WSLATests.cpp
```diff
@@ -383,6 +390,54 @@ class WSLATests
...
+        options.Force = TRUE;
+        VERIFY_SUCCEEDED(session->DeleteImage(&options, deletedImages.addressof(), deletedImages.size_address<ULONG>(), nullptr));
+        VERIFY_IS_TRUE(deletedImages.size() > 0);
+        VERIFY_IS_TRUE(std::strlen(deletedImages[0].Image) > 0);
+
+        // Verify that the image is no longer in the list of images.
+        ExpectImagePresent(*session, "alpine:latest", false);
+
+        // Test delete failed if image not exists.
```
OneBlue, line 437: nit: ` image does not exist`

test/windows/WSLATests.cpp
```diff
@@ -383,6 +390,54 @@ class WSLATests
         VERIFY_ARE_EQUAL(0, result.Code);
         VERIFY_IS_TRUE(result.Output[1].find("Hello from Docker!") != std::string::npos);
     }
+    TEST_METHOD(DeleteImage)
```
OneBlue, line 393: nit: I'll recommend adding tests to cover `NoPrune` as well. We can import an image, get its hash, and then call Delete(NoPrune = true) and validate that the image is untagged, but that the layer with that hash it still present.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -297,6 +297,26 @@ interface IWSLAContainer : IUnknown
...
+struct WSLA_DELETED_IMAGE_INFORMATION
+{
+    char Image[WSLA_MAX_IMAGE_NAME_LENGTH + 1];
+    WSLADeletedImageType Type;
+};
+
+struct WSLA_DELETE_IMAGE_OPTIONS
+{
+    LPCSTR Image; // Image can be ID or Repo:Tag.
+    BOOL Force;
```
OneBlue, line 315: nit: I would recommend having a `Flags` field here. That will make this easy to extend in the future if we need to

OneBlue (approved):
Change looks good ! A few minor comments and suggestions

yao-msft:
I'll address the nits in next pr (completing listimage api)

PR #14067, opened 2026-01-15
Redesign WSLA Session management 

OneBlue:

This change redesigns our per user session management logic into a global SessionManager.

This allows sessions to be interacted across Windows users.

## PR Checklist


test/windows/Common.cpp
```diff
@@ -1328,20 +1329,26 @@ void StopWslaService()
     StopService(service.get());
 }
 
-wil::unique_handle GetNonElevatedToken()
+wil::unique_handle GetNonElevatedToken(TOKEN_TYPE Type)
```
OneBlue, line 1332: Big thanks to @johnstep for his help in getting this right

src/windows/wslaservice/exe/WSLASessionManager.cpp
```diff
@@ -0,0 +1,197 @@
...
+    Version->Revision = WSL_PACKAGE_VERSION_REVISION;
+}
+
+WSLASessionManagerImpl::CallingProcessTokenInfo WSLASessionManagerImpl::GetCallingProcessTokenInfo()
+{
+    const wil::unique_handle userToken = wsl::windows::common::security::GetUserToken(TokenImpersonation);
+
+    auto tokenInfo = wil::get_token_information<TOKEN_USER>(userToken.get());
+
+    auto elevated = wil::test_token_membership(userToken.get(), SECURITY_NT_AUTHORITY, SECURITY_BUILTIN_DOMAIN_RID, DOMAIN_ALIAS_RID_ADMINS);
```
benhillis, line 141: have we determined this is better than the other elevation check we are doing elsewhere?

test/windows/Common.cpp
benhillis: I'd prefer we consolidate our code that is handling admin / non-admin tokens into a common helper function used everywhere. Is that possible?

benhillis: nit: Type = TokenPrimary as default param to avoid churn.

test/windows/WSLATests.cpp
```diff
@@ -90,8 +90,8 @@ class WSLATests
 
     wil::com_ptr<IWSLASession> CreateSession(const WSLA_SESSION_SETTINGS& sessionSettings = GetDefaultSessionSettings(), WSLASessionFlags Flags = WSLASessionFlagsNone)
     {
-        wil::com_ptr<IWSLAUserSession> userSession;
-        VERIFY_SUCCEEDED(CoCreateInstance(__uuidof(WSLAUserSession), nullptr, CLSCTX_LOCAL_SERVER, IID_PPV_ARGS(&userSession)));
+        wil::com_ptr<IWSLASessionManager> userSession;
```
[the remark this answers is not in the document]
benhillis: agree

benhillis (commented):
Couple small comments, looking good though.

src/windows/wslaservice/exe/WSLASessionManager.cpp
OneBlue: I think so. I spent quite a bit of time trying to switch WSL to this check, but doing so breaks the tests because there doesn't seem to be a way for a non-SYSTEM process to generate a non-elevated primary token that can be used in WSL.

I think this is something we should fix, but it's not related to this change, so this is something that we should investigate as a followup on the WSL side. I'm meeting with a security token expert later today to see if they have pointers on how to easily solve this.

benhillis (approved):

benhillis (approved):

PR #14069, opened 2026-01-15
Enable recovery of Docker containers when restarting a WSLA session from persistent storage.

kvega005:
Enable recovery of Docker containers when restarting a WSLA session from persistent storage.
- Added ListContainers endpoint to WSLA Docker HTTP client.
- On session startup, query Docker for existing containers and reconstruct WSLAContainerImpl objects.
- Refactored container construction to accept explicit state, name, and TTY parameters; introduced Open method for recovery.
- Added functional tests verifying the WSLA session can recover containers from previous sessions and delete the containers.

**Notes**
- Running recovered containers remains unsupported until metadata persistence (volumes, ports, mounts) is implemented in a follow-up PR.
- Current scope focuses on discovery and lifecycle control for containers created in previous sessions.

## PR Checklist


src/windows/inc/docker_schema.h
```diff
@@ -151,6 +151,48 @@ struct StartExec
...
+    Paused,
+    Restarting,
+    Exited,
+    Removing,
+    Dead,
+    Unknown
+};
+
+NLOHMANN_JSON_SERIALIZE_ENUM(ContainerState, {
+    {ContainerState::Unknown, nullptr},
```
OneBlue: I'm not sure that we want `nullptr` here since strings are `std::string` under the hood, and a null pointer will usually cause a crash there

test/windows/WSLATests.cpp
```diff
@@ -2354,6 +2354,62 @@ class WSLATests
...
+
+            // Try to open the container from the previous session
+            wil::com_ptr<IWSLAContainer> recoveredContainer;
+            VERIFY_SUCCEEDED(session->OpenContainer(containerName.c_str(), &recoveredContainer));
+
+            // Verify container state
+            WSLA_CONTAINER_STATE state{};
+            VERIFY_SUCCEEDED(recoveredContainer->GetState(&state));
+            VERIFY_ARE_EQUAL(state, WslaContainerStateExited);
+
```
OneBlue, line 2443: We should validate that we're able to Start() the recovered container again here (or we could do that in a followup PR if we need to write the TTY flag in first)

OneBlue (approved):
Change looks good, a couple minor comments

test/windows/WSLATests.cpp
kvega005: I think if we dont have ports or volumes, start can still work, but I intentionally left the Start functionality for a follow-up PR. I always set the state to WslaContainerStateExited in the `WSLAContainer::Open` function so that it's clear that starting a container that we didn't create in this session is not yet supported.

src/windows/wslaservice/exe/DockerHTTPClient.cpp
```diff
@@ -72,6 +72,12 @@ std::vector<docker_schema::Image> DockerHTTPClient::ListImages()
     return Transaction<docker_schema::EmptyRequest, std::vector<docker_schema::Image>>(verb::get, "http://localhost/images/json");
 }
 
+std::vector<docker_schema::ContainerInfo> DockerHTTPClient::ListContainers(bool all)
+{
+    auto url = all ? "http://localhost/containers/json?all=true" : "http://localhost/containers/json";
```
kevpar: >auto url = all ? "http://localhost/containers/json?all=true" : "http://localhost/containers/json"; [](http://example.com/codeflow?start=4&length=98)

Suggest `std::format("http://localhost/containers/json?all={}", all ? "true" : "false")` instead.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -122,35 +122,53 @@ auto ProcessPortMappings(const WSLA_CONTAINER_OPTIONS& options, WSLAVirtualMachi
...
+std::string ExtractContainerName(const std::vector<std::string>& names, const std::string& id)
+{
+    if (names.empty())
+    {
+        return id;
+    }
+
+    // Docker container names have a leading '/', strip it.
+    std::string name = names[0];
+    if (!name.empty() && name[0] == '/')
```
kevpar, line 134: >if (!name.empty() && name[0] == '/') [](http://example.com/codeflow?start=4&length=36)

If `names[0]` happens to be an empty string, we will return it as-is. I assume it's not expected that we would ever get this returned, so maybe add a comment clarifying that expectation, if we aren't checking for it?

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -181,15 +199,12 @@ WSLAContainerImpl::~WSLAContainerImpl()
...
-    // restarting a session from a persisted storage.
-
-    if (m_state == WslaContainerStateExited)
+    // Stop running containers.
+    if (m_state == WslaContainerStateRunning)
     {
         try
         {
-            Delete();
+            Stop(WSLASignalSIGKILL, stopTimeout);
```
kevpar, line 207: >Stop(WSLASignalSIGKILL, stopTimeout); [](http://example.com/codeflow?start=12&length=37)

Is it safe to assume this destructor only runs when the runtime is shutting down?

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -607,13 +622,72 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
...
+        if (port.PublicPort != 0)
+        {
+            ports.push_back({port.PublicPort, port.PublicPort, port.PrivatePort, AF_INET, false});
+        }
+    }
+
+    // Create a WSLAContainerImpl directly without going through Create().
+    // TODO: Recover TTY from metadata.
+    // TODO: Recover volumes from metadata.
+    // TODO: Recover container state from Docker.
```
kevpar, line 668: If we already need to recover some pieces of container configuration from host-side metadata, should we just keep everything we need to rehydrate host-side?

src/windows/wslaservice/exe/WSLAContainer.cpp
kvega005: The destructor runs when the session object that owns the container is destroyed.

src/windows/wslaservice/exe/WSLAContainer.cpp
kvega005: It's possible, and we will discuss that for the next step of this PR.

src/windows/common/WSLAContainerLauncher.cpp
```diff
@@ -22,13 +22,25 @@ RunningWSLAContainer::RunningWSLAContainer(wil::com_ptr<IWSLAContainer>&& Contai
...
+}
+
 IWSLAContainer& RunningWSLAContainer::Get()
 {
     return *m_container;
 }
 
 void RunningWSLAContainer::Reset()
 {
+    if (m_container && m_deleteOnClose)
```
OneBlue, line 37: nit: At some point we'll need a flag like that at the WSLAContainer level (mostly for the `run --rm` usecase). 

This can be done in a future change though

OneBlue (approved):

PR #14073, opened 2026-01-16
virtiofs: add support for mounting directories (not just full volumes)

benhillis:
The primary behavior change introduced by this PR is that virtiofs can now mount sub-directories instead of just the entire path. This is achieved by using a GUID tag for the vitiofs share and introducing a query method to ask the wsl service for the source of a virtiofs share, instead of relying on parsing the /proc/pid/mountinfo file.

### DIFF
--- a/src/linux/init/drvfs.cpp
+++ b/src/linux/init/drvfs.cpp
@@ -298,70 +298,12 @@ try
     {
         return MountFilesystem(DRVFS_FS_TYPE, Source, Target, Options, ExitCode);
     }
-
-    // Use virtiofs if the source of the mount is the root of a drive; otherwise, use 9p.
-    if (WSL_USE_VIRTIO_FS(Config))
+    else if (WSL_USE_VIRTIO_FS(Config))
     {
-        if (wsl::shared::string::IsDriveRoot(Source))
-        {
-            return MountVirtioFs(Source, Target, Options, Admin, Config, ExitCode);
-        }
-
-        LOG_WARNING("virtiofs is only supported for mounting full drives, using 9p to mount {}", Source);
+        return MountVirtioFs(Source, Target, Options, Admin, Config, ExitCode);
     }
 
-    //
-    // Check if the path is a UNC path.
-    //
-
-    const char* Plan9Source;
-    std::string UncSource;
-    if ((strlen(Source) >= PLAN9_UNC_PREFIX_LENGTH) && ((Source[0] == '/') || (Source[0] == '\\')) &&
-        ((Source[1] == '/') || (Source[1] == '\\')))
-    {
-        UncSource = PLAN9_UNC_TRANSLATED_PREFIX;
-        UncSource += &Source[PLAN9_UNC_PREFIX_LENGTH];
-        Plan9Source = UncSource.c_str();
-    }
-    else
-    {
-        Plan9Source = Source;
-    }
-
-    //
-    // Check whether to use the elevated or regular 9p server.
-    //
-
-    bool Elevated = Admin.has_value() ? Admin.value() : IsDrvfsElevated();
-
-    //
-    // Initialize mount options.
-    //
-
-    auto Plan9Options = std::format("{};path={}", PLAN9_ANAME_DRVFS, Plan9Source);
-
-    //
-    // N.B. The cache option is added to the start of this so if the user
-    //      specifies one explicitly, it will override the default.
-    //
-
-    std::string MountOptions = "cache=mmap,";
-    auto ParsedOptions = ConvertDrvfsMountOptionsToPlan9(Options ? Options : "", Config);
-    Plan9Options += ParsedOptions.first;
-    MountOptions += ParsedOptions.second;
-
-    //
-    // Append the 9p mount options to the end of the other mount options and perform the mount operation.
-    //
-
-    MountOptions += Plan9Options;
-
-    if (MountPlan9Filesystem(Source, Target, MountOptions.c_str(), Elevated, Config, ExitCode) < 0)
-    {
-        return -1;
-    }
-
-    return 0;
+    return MountPlan9(Source, Target, Options, Admin, Config, ExitCode);
 }
 CATCH_RETURN_ERRNO()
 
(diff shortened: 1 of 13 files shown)


src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -2141,23 +2141,21 @@ std::wstring WslCoreVm::AddVirtioFsShare(_In_ bool Admin, _In_ PCWSTR Path, _In_
...
     std::wstring tag;
     VirtioFsShare key(sharePath.c_str(), Options, Admin);
     if (!m_virtioFsShares.contains(key))
     {
-        // Generate a new tag for the share.
-        tag = Admin ? TEXT(LX_INIT_DRVFS_ADMIN_VIRTIO_TAG) : TEXT(LX_INIT_DRVFS_VIRTIO_TAG);
-        tag += sharePath[0];
-        tag += std::to_wstring(m_virtioFsShares.size());
+        // Generate a new tag for the share. The tag contains a prefix, the share path, and a unique index.
+        tag = std::format(
```
benhillis: This is the primary behavior difference.

test/windows/UnitTests.cpp
```diff
@@ -878,11 +878,6 @@ class UnitTests
         }
     }
 
-    TEST_METHOD(WslPath)
```
benhillis, line 881: This test was moved to DrvFs tests so different cross-os FS servers can be validated instead of just the default.

src/linux/init/drvfs.cpp
```diff
@@ -527,11 +553,10 @@ try
 
     gsl::span<gsl::byte> ResponseSpan;
     const auto& Response = Channel.Transaction<LX_INIT_ADD_VIRTIOFS_SHARE_MESSAGE>(AddShare.Span(), &ResponseSpan);
-
     if (Response.Result != 0)
     {
-        LOG_ERROR("Add virtiofs share for {} failed {}", Source, Response.Result);
-        return -1;
+        LOG_WARNING("Add virtiofs share for {} failed {}, falling back to Plan9", Source, Response.Result);
+        return MountPlan9(Source, Target, Options, Admin, Config, ExitCode);
```
OneBlue, line 560: Oh I never realized that we actually allowed both plan9 & vritiofs when virtiofs was enabled

benhillis: yeah before it was only if we tried to mount something that wasn't the root of the drive, now it's just in failure cases - for example mounting SMB or a FAT share. We should debug that, but the result is us using virtiofs for much more than we were previously.

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -2136,22 +2133,27 @@ std::wstring WslCoreVm::AddVirtioFsShare(_In_ bool Admin, _In_ PCWSTR Path, _In_
     WI_ASSERT(Admin == wsl::windows::common::security::IsTokenElevated(UserToken));
 
     // Ensure that the path has a trailing path separator.
-    std::wstring sharePath{Path};
-    if (sharePath.back() != L'\\')
+    std::wstring sharePath(Path);
+    if (!sharePath.ends_with(L'\\') && !sharePath.ends_with(L'/'))
```
OneBlue, line 2137: Do we still need this if we canonicalize the path later ?

test/linux/unit_tests/lxtfs.c
```diff
@@ -294,7 +294,7 @@ Return Value:
             snprintf(
                 Plan9Options,
                 sizeof(Plan9Options),
-                "aname=drvfs;path=%s%s;symlinkroot=/mnt/,cache=5,access=client,msize=65536,trans=fd,rfd=4,wfd=4",
+                "aname=drvfs;path=%s%s;symlinkroot=/mnt/,cache=5,access=client,msize=65536,trans=fd,rfd=*,wfd=*",
```
OneBlue, line 297: What's the reason for this change ? Are we allocating more fds that cause those to be different ?

OneBlue: Ok found the answer is the new file

OneBlue (approved):
LGTM, one minor comment

src/linux/init/drvfs.cpp
```diff
@@ -452,10 +396,95 @@ Return Value:
...
+
+--*/
+
+try
+{
+    //
+    // Check if the path is a UNC path.
+    //
+
+    const char* Plan9Source;
```
ptrivedi, line 438: nit: initialize to nullptr

src/linux/init/drvfs.cpp
```diff
@@ -452,10 +396,95 @@ Return Value:
...
+    auto ParsedOptions = ConvertDrvfsMountOptionsToPlan9(Options ? Options : "", Config);
+    Plan9Options += ParsedOptions.first;
+    MountOptions += ParsedOptions.second;
+
+    //
+    // Append the 9p mount options to the end of the other mount options and perform the mount operation.
+    //
+
+    MountOptions += Plan9Options;
+    if (MountPlan9Share(Source, Target, MountOptions.c_str(), Elevated, Config, ExitCode) < 0)
```
ptrivedi, line 479: nit: could just    return MountPlan9Share(...);

src/linux/init/drvfs.cpp
```diff
@@ -452,10 +396,95 @@ Return Value:
...
+    if ((strlen(Source) >= PLAN9_UNC_PREFIX_LENGTH) && ((Source[0] == '/') || (Source[0] == '\\')) &&
+        ((Source[1] == '/') || (Source[1] == '\\')))
+    {
+        UncSource = PLAN9_UNC_TRANSLATED_PREFIX;
+        UncSource += &Source[PLAN9_UNC_PREFIX_LENGTH];
+        Plan9Source = UncSource.c_str();
+    }
+    else
+    {
+        Plan9Source = Source;
```
ptrivedi, line 449: This means Plan9Source has the potential of being null / empty. Could this cause issues with the "path=" option being set up on line 462?

src/windows/service/exe/WslCoreVm.cpp
benhillis: I think so. This will allow us to reuse shares more often rather than creating new ones.

PR #14074, opened 2026-01-16
WSLA: Switch to using the system distro vhd.

benhillis:
This changes moves WSLA to use the WSL system distro vhd instead of its own. This will allow for lower disk and memory usage for the product.

### DIFF
--- a/src/windows/wslaservice/exe/WSLAContainer.cpp
+++ b/src/windows/wslaservice/exe/WSLAContainer.cpp
@@ -597,7 +597,9 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
     {
         // TODO: UDP support
         // TODO: Investigate ipv6 support.
-        auto& portEntry = request.HostConfig.PortBindings[std::format("{}/tcp", e.ContainerPort)];
+        auto portKey = std::format("{}/tcp", e.ContainerPort);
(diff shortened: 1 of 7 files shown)


test/windows/WSLATests.cpp
```diff
@@ -313,7 +313,8 @@ class WSLATests
 
         {
             std::string expectedError =
-                "pull access denied for does-not, repository does not exist or may require 'docker login'";
```
benhillis, line 316: Error message changed in the v25 version of docker we are now using.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -597,7 +597,9 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
     {
         // TODO: UDP support
         // TODO: Investigate ipv6 support.
-        auto& portEntry = request.HostConfig.PortBindings[std::format("{}/tcp", e.ContainerPort)];
+        auto portKey = std::format("{}/tcp", e.ContainerPort);
```
benhillis, line 600: This was not needed in the newer version of docker we were using which had this patch:
https://github.com/moby/moby/pull/50874

OneBlue (approved):

PR #14077, opened 2026-01-17
Implement container logs support

OneBlue:
## PR Checklist


test/windows/WSLATests.cpp
```diff
@@ -284,13 +284,11 @@ class WSLATests
...
-        if (Present)
+        auto found = std::ranges::find(tags, Image) != tags.end();
+        if (Present != found)
         {
-            VERIFY_IS_TRUE(std::ranges::find(tags, Image) != tags.end());
-        }
-        else
-        {
-            VERIFY_IS_TRUE(std::ranges::find(tags, Image) == tags.end());
+            LogError("Image presence check failed for image: %hs, images: %hs", Image, wsl::shared::string::Join(tags, ',').c_str());
```
OneBlue, line 313: This made debugging a test issue easier, so keeping it

src/windows/common/CMakeLists.txt
```diff
@@ -132,7 +132,7 @@ set(HEADERS
 
 include_directories(${CMAKE_SOURCE_DIR}/src/windows/wslaclient)
 add_library(common STATIC ${SOURCES} ${HEADERS})
-add_dependencies(common wslserviceidl localization wslservicemc wslinstalleridl wslaserviceproxystub)
+add_dependencies(common wslserviceidl wslaserviceidl localization wslservicemc wslinstalleridl wslaserviceproxystub)
```
OneBlue, line 135: Caught this dependency issue while developing

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -633,6 +638,46 @@ void WSLAContainerImpl::Inspect(LPSTR* Output)
...
+
+    if (m_tty)
+    {
+        // For tty processes, simply relay the HTTP chunks.
+        auto [ttyRead, ttyWrite] = common::wslutil::OpenAnonymousPipe(0, true, true);
+
+        auto handle = std::make_unique<RelayHandle<HTTPChunkBasedReadHandle>>(std::move(socket), std::move(ttyWrite));
+        m_logsRelay.AddHandle(std::move(handle));
+
+        *Stdout = HandleToULong(common::wslutil::DuplicateHandleToCallingProcess(ttyRead.get()));
```
kevpar, line 737: Would it be better for the caller to provide a handle instead of receiving one? That way, if they want to route the logs to a file, they could just give a file handle.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -291,6 +297,7 @@ interface IWSLAContainer : IUnknown
     HRESULT GetInitProcess([out] IWSLAProcess** Process);
     HRESULT Exec([in] const struct WSLA_PROCESS_OPTIONS* Options, [out] IWSLAProcess** Process, [out] int* Errno);
     HRESULT Inspect([out] LPSTR* Output);
+    HRESULT Logs([in] WSLALogsFlags Flags, [out] ULONG* Stdout, [out] ULONG* Stderr, [in] ULONGLONG Since, [in] ULONGLONG Until, [in] ULONGLONG Tail);
```
kevpar, line 300: Is it valid to only provide a value for one of Stdout/Stderr?

src/windows/wslaservice/exe/LogsRelay.h
```diff
@@ -0,0 +1,40 @@
...
+    void AddHandle(std::unique_ptr<common::relay::OverlappedIOHandle>&& Handle);
+
+private:
+    void StartRelayThread();
+    void StopRelayThread();
+    void Run();
+
+    std::thread m_thread;
+    windows::common::relay::MultiHandleWait m_io;
+    wil::unique_event m_stopEvent{wil::EventOptions::ManualReset};
```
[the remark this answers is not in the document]
OneBlue: That's by design. LogsRelay is protected by the process' mutex and should not be accessed concurrently

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: I considered this, but decided not to (at least in the first implementation) because: 

1) The caller needs to know when the relaying is over. We could in theory make that work by making Logs() block until all the IO is flushed, but then we'd need extra logic, probably via an event HANDLE to cancel in case the client receives ctrl-c 

2) We use overlapped IO, and not all handles support it. If the caller passes a handle that doesn't support overlapped IO, we could end up blocked on a ReadFile(), which could deadlock the service 

Given that Logs() isn't expected to have a high data throughput, I'm not too worried about its performance, but  open to revisiting if that changes.

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: It's not. The client is expected to pass addresses for both of those. 

In the case of a TTY process though, only `Stdout` will be returned, since the docker API only gives us one stream back

test/windows/Common.cpp
```diff
@@ -2604,6 +2604,41 @@ std::string ReadToString(SOCKET Handle)
...
+    }
+
+    return output;
+}
+
+void VerifyPatternMatch(const std::string& Content, const std::string& Pattern)
+{
+    if (!PathMatchSpecA(Content.c_str(), Pattern.c_str()))
+    {
+        std::wstring message = std::format(L"Output: '{}' didn't match pattern: '{}'", Content, Pattern);
```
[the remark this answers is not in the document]
OneBlue: That's incorrect. This is done automatically

src/windows/wslaservice/exe/WSLAContainer.cpp
ptrivedi: For container export / image save cases, my changes so far have caller providing the handle because the output is tar content that needs to go into a tar file. @OneBlue let's discuss some more on this today

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -466,6 +467,43 @@ static int Pull(std::wstring_view commandLine)
     return 0;
 }
 
+static int Logs(std::wstring_view commandLine)
```
benhillis, line 470: we're continuing to add to wsladiag, and wslc is basically a copy paste of this. not something we need to address in this pr, but this sucks.

benhillis (approved):

src/windows/wsladiag/wsladiag.cpp
OneBlue: I agree. This was just for me to test tty scenarios. Long term this will need to move to wsl.exe

PR #14103, opened 2026-01-23
Rethink WSLAContainer tracking to use Ids instead of names

OneBlue:

This change redesigns the WSLASession logic to keep of containers based on Ids instead of names. 

Benefits: 

- Names are no longer required when creating containers
- Open() will match containers based on id, names, or id prefixes, similar to what containers tool do

## PR Checklist


src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -38,6 +38,24 @@ std::pair<std::string, std::string> ParseImage(const std::string& Input)
...
+        {
+            return false;
+        }
+
+        Name++;
+        length++;
+    }
+
+    return length > 0 && length <= WSLA_MAX_CONTAINER_NAME_LENGTH;
+}
```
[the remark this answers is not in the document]
OneBlue: Fixed by using the STL version

src/windows/common/WSLAContainerLauncher.cpp
```diff
@@ -112,7 +128,12 @@ std::pair<HRESULT, std::optional<RunningWSLAContainer>> WSLAContainerLauncher::C
 {
     WSLA_CONTAINER_OPTIONS options{};
     options.Image = m_image.c_str();
```
benhillis, line 130: do we need similar handling of the image?

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -665,16 +689,39 @@ try
 }
 CATCH_RETURN();
 
-HRESULT WSLASession::OpenContainer(LPCSTR Name, IWSLAContainer** Container)
+HRESULT WSLASession::OpenContainer(LPCSTR Id, IWSLAContainer** Container)
 try
 {
-    // TODO: Rethink name / id usage here.
+    THROW_HR_IF_MSG(E_INVALIDARG, !IsContainerNameValid(Id), "Invalid container name: %hs", Id);
```
benhillis, line 695: The param is called Id but the checks refer to it as name, this is a little confusing.

benhillis (approved):
minor comments

src/windows/wslaservice/exe/WSLASession.cpp
OneBlue: Yeah that's a bit weird. It can be both, but I want to check for invalid names here because otherwise it could be use to break URL when we call InspectContainer() below. 

I'm open to suggestions to make this more clear

src/windows/common/WSLAContainerLauncher.cpp
OneBlue: We don't, `Image` is required to create a container

PR #14105, opened 2026-01-24
Localization change from build: 138252891

OneBlue:


benhillis:
I was getting sick of the warnings about all the missing strings.

benhillis (approved):

OneBlue:
> I was getting sick of the warnings about all the missing strings.

Thank you ! I was actually considering disabling those on the feature branch, but this is a better solution

PR #14107, opened 2026-01-24
Optimize the WSLA unit tests

OneBlue:

This change updates the WSLA test logic to reuse the same session whenever possible, significantly speeding up the tests

## PR Checklist

### DIFF
--- a/src/shared/inc/lxinitshared.h
+++ b/src/shared/inc/lxinitshared.h
@@ -1536,7 +1536,7 @@ struct WSLA_ERROR
 {
     static inline auto Type = LxMessageWSLAError;
     MESSAGE_HEADER Header;
-    int Errno;
+    int Errno{};
     PRETTY_PRINT(FIELD(Header), FIELD(Errno));
 };
 
@@ -1547,7 +1547,7 @@ struct WSLA_GET_DISK_RESULT
     DECLARE_MESSAGE_CTOR(WSLA_GET_DISK_RESULT);
 
     MESSAGE_HEADER Header;
-    unsigned int Result;
+    unsigned int Result{};
     char Buffer[];
 
     PRETTY_PRINT(FIELD(Header), FIELD(Result), FIELD(Buffer));
@@ -1561,7 +1561,7 @@ struct WSLA_GET_DISK
     DECLARE_MESSAGE_CTOR(WSLA_GET_DISK);
 
     MESSAGE_HEADER Header;
-    unsigned int ScsiLun;
+    unsigned int ScsiLun{};
 
     PRETTY_PRINT(FIELD(Header), FIELD(ScsiLun));
 };
@@ -1569,8 +1569,8 @@ struct WSLA_GET_DISK
 struct WSLA_MOUNT_RESULT
 {
     static inline auto Type = LxMessageWSLAMountResult;
-    MESSAGE_HEADER Header;
-    int Result;
+    MESSAGE_HEADER Header{};
+    int Result{};
 
     PRETTY_PRINT(FIELD(Header), FIELD(Result));
 };
@@ -1582,12 +1582,12 @@ struct WSLA_MOUNT
 
     DECLARE_MESSAGE_CTOR(WSLA_MOUNT);
 
-    MESSAGE_HEADER Header;
-    unsigned int SourceIndex;
-    unsigned int DestinationIndex;
-    unsigned int TypeIndex;
-    unsigned int OptionsIndex;
-    unsigned int Flags;
+    MESSAGE_HEADER Header{};
+    unsigned int SourceIndex{};
+    unsigned int DestinationIndex{};
+    unsigned int TypeIndex{};
+    unsigned int OptionsIndex{};
+    unsigned int Flags{};
 
     enum MountType : uint8_t
     {
@@ -1671,10 +1671,10 @@ struct WSLA_TTY_RELAY
     DECLARE_MESSAGE_CTOR(WSLA_TTY_RELAY);
 
     MESSAGE_HEADER Header;
-    int32_t TtyMaster;
-    int32_t TtyInput;
-    int32_t TtyOutput;
-    int32_t TtyControl;
+    int32_t TtyMaster{};
+    int32_t TtyInput{};
+    int32_t TtyOutput{};
+    int32_t TtyControl{};
 
     PRETTY_PRINT(FIELD(Header), FIELD(TtyMaster), FIELD(TtyInput), FIELD(TtyOutput), FIELD(TtyControl));
 };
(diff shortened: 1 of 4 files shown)


src/shared/inc/lxinitshared.h
```diff
@@ -1536,7 +1536,7 @@ struct WSLA_ERROR
 {
     static inline auto Type = LxMessageWSLAError;
     MESSAGE_HEADER Header;
-    int Errno;
+    int Errno{};
```
OneBlue, line 1539: While testing this, I discovered an uninitialized field in the WSLA_PROCESS_EXITED message, which could incorrectly report that a process was signaled. 

Moving forward, let's just always initialize all message fields to prevent this

test/windows/WSLATests.cpp
```diff
@@ -844,16 +819,16 @@ class WSLATests
...
 
-        auto session = CreateSession(settings);
+        auto session = networkingMode != WSLANetworkingModeNAT ? CreateSession(settings) : defaultSession;
 
         // Install socat in the container.
         //
         // TODO: revisit this in the future to avoid pulling packages from the network.
         auto installSocat = WSLAProcessLauncher("/bin/sh", {"/bin/sh", "-c", "tdnf install socat -y"}).Launch(*session);
-        ValidateProcessOutput(installSocat, {});
+        ValidateProcessOutput(installSocat, {}, 0, 300 * 1000);
```
OneBlue, line 831: @benhillis : I've been hitting hangs here, so adding a timeout so any failure in the CI is obvious

test/windows/WSLATests.cpp
```diff
@@ -51,13 +52,13 @@ class WSLATests
         auto distroKey = OpenDistributionKey(LXSS_DISTRO_NAME_TEST_L);
 
         auto vhdPath = wsl::windows::common::registry::ReadString(distroKey.get(), nullptr, L"BasePath");
-        testVhd = std::filesystem::path{vhdPath} / "ext4.vhdx";
+        auto testVhd = std::filesystem::path{vhdPath} / "ext4.vhdx";
         storagePath = std::filesystem::current_path() / "test-storage";
 
-        auto session = CreateSession();
+        defaultSession = CreateSession(GetDefaultSessionSettings(testSessionName, true, WSLANetworkingModeNAT));
```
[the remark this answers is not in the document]
OneBlue: That's by design, to make the tests faster

test/windows/WSLATests.cpp
```diff
@@ -1256,7 +1226,7 @@ class WSLATests
             VERIFY_ARE_EQUAL(process.Get().GetStdHandle(3, reinterpret_cast<ULONG*>(&dummyHandle)), E_INVALIDARG);
 
             // Validate that the process object correctly handle requests after the VM has terminated.
-            session.reset();
+            CloseTestSession();
```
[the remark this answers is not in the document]
OneBlue: That's by design, this is just there to force the session to be recreated

test/windows/WSLATests.cpp
```diff
@@ -1763,7 +1718,7 @@ class WSLATests
             VERIFY_SUCCEEDED(container.Get().Delete());
 
             // Terminate the session
-            session.reset();
+            CloseTestSession();
```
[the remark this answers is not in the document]
OneBlue: That's also be design to force the test session to be recreated immediately

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -553,7 +553,7 @@ void WSLAVirtualMachine::ConfigureNetworking()
     if (m_settings.NetworkingMode == WSLANetworkingModeNAT)
     {
         // TODO: refactor this to avoid using wsl config
-        static wsl::core::Config config(nullptr);
+        wsl::core::Config config(nullptr);
```
OneBlue, line 556: This was causing all VM's to try to use the same IP address. We should refactor this in the long run, but for now fixed by making sure that configs aren't shared accross sessions

test/windows/WSLATests.cpp
```diff
@@ -97,21 +97,38 @@ class WSLATests
...
+        settings.StoragePath = enableStorage ? storagePath.c_str() : nullptr;
         settings.MaximumStorageSizeMb = 1000; // 1GB.
-        settings.NetworkingMode = WSLANetworkingModeNAT;
+        settings.NetworkingMode = networkingMode;
 
         return settings;
     }
 
-    wil::com_ptr<IWSLASession> CreateSession(const WSLA_SESSION_SETTINGS& sessionSettings = GetDefaultSessionSettings(), WSLASessionFlags Flags = WSLASessionFlagsNone)
+    auto CloseTestSession()
```
benhillis: Since this creates a new line, should this be called ResetTestSession?

test/windows/WSLATests.cpp
```diff
@@ -97,21 +97,38 @@ class WSLATests
...
+    {
+        defaultSession.reset();
```
benhillis: why the reset and scope exit? why can't we just create a newline that overwrites the old one?

test/windows/WSLATests.cpp
```diff
@@ -42,7 +42,8 @@ class WSLATests
     WSL_TEST_CLASS(WSLATests)
     wil::unique_couninitialize_call coinit = wil::CoInitializeEx();
     WSADATA Data;
-    std::filesystem::path testVhd;
+    wil::com_ptr<IWSLASession> defaultSession;
```
benhillis: nit: it's a little hard inside a method what is a local versus what is a class member variable?

test/windows/WSLATests.cpp
OneBlue: That's because some test cases need to open the same storage VHD, so they need a scope where the default session is not running, so they do: 

```
{
   auto restoreDefaultSession = CloseTestSession();

   // Run the test
}

```

test/windows/WSLATests.cpp
OneBlue: That's a good point. I'll rename with a prefix to make it explicit

test/windows/WSLATests.cpp
```diff
@@ -40,24 +40,24 @@ extern bool g_fastTestRun;
...
-        THROW_IF_WIN32_ERROR(WSAStartup(MAKEWORD(2, 2), &Data));
+        THROW_IF_WIN32_ERROR(WSAStartup(MAKEWORD(2, 2), &m_wsadata));
 
         auto distroKey = OpenDistributionKey(LXSS_DISTRO_NAME_TEST_L);
 
         auto vhdPath = wsl::windows::common::registry::ReadString(distroKey.get(), nullptr, L"BasePath");
         storagePath = std::filesystem::current_path() / "test-storage";
 
-        defaultSession = CreateSession(GetDefaultSessionSettings(testSessionName, true, WSLANetworkingModeNAT));
+        m_defaultSession = CreateSession(Getm_defaultSessionSettings(c_testSessionName, true, WSLANetworkingModeNAT));
```
benhillis: nit: find / replace changed the method name.

benhillis (approved):

PR #14112, opened 2026-01-27
Implement container attach 

OneBlue:

This change adds support for attaching to a running container, as well as a simple attach implementation for wsladiag.exe 

## PR Checklist

### DIFF
--- a/src/windows/wsladiag/wsladiag.cpp
+++ b/src/windows/wsladiag/wsladiag.cpp
@@ -529,8 +529,6 @@ static void RelayNonTtyProcess(HANDLE Stdin, HANDLE Stdout, HANDLE Stderr)
 
 static int InteractiveShell(ClientRunningWSLAProcess&& Process, bool Tty)
 {
-    auto exitEvent = Process.GetExitEvent();
-
     if (Tty)
     {
         RelayInteractiveTty(Process, Process.GetStdHandle(WSLAFDTty).get());
--- a/src/windows/wsladiag/wsladiag.cpp
+++ b/src/windows/wsladiag/wsladiag.cpp
@@ -453,79 +453,95 @@ static int Logs(std::wstring_view commandLine)
     return 0;
 }
 
+static void RelayInteractiveTty(ClientRunningWSLAProcess& Process, HANDLE Tty, bool triggerRefresh = false)
+{
+    // Configure console for interactive usage.
+    wsl::windows::common::ConsoleState console;
+
+    if (triggerRefresh)
+    {
+        // In the case of an Attach, force a terminal resize to force the tty to refresh its display.
+        // The docker client uses the same trick.
+
+        auto size = console.GetWindowSize();
+
+        LOG_IF_FAILED(Process.Get().ResizeTty(size.Y + 1, size.X + 1));
+        LOG_IF_FAILED(Process.Get().ResizeTty(size.Y, size.X));
+    }
+
+    wil::unique_event exitEvent(wil::EventOptions::ManualReset);
+
+    // Create a thread to relay stdin to the pipe.
+    std::thread inputThread([&]() {
+        auto updateTerminal = [&console, &Process]() {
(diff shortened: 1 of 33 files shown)


src/windows/wslaservice/exe/LogsRelay.cpp
```diff
@@ -25,13 +25,25 @@ LogsRelay::~LogsRelay()
 }
 
 void LogsRelay::AddHandle(std::unique_ptr<common::relay::OverlappedIOHandle>&& Handle)
```
OneBlue, line 27: This class will be renamed to `ContainerRelay` in a future refactoring

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -453,79 +453,93 @@ static int Logs(std::wstring_view commandLine)
...
-        io.AddHandle(std::make_unique<RelayHandle<ReadHandle>>(Process.GetStdHandle(2), GetStdHandle(STD_ERROR_HANDLE)));
-        io.AddHandle(std::make_unique<EventHandle>(exitEvent.get()));
-
-        io.Run({});
+    // Required because ReadFile() blocks if stdin is a tty.
+    if (wsl::windows::common::wslutil::IsInteractiveConsole())
+    {
+        // TODO: Will output CR instead of LF's which can confuse the linux app.
+        // Consider a custom relay logic to fix this.
+        inputThread = std::thread{
```
benhillis, line 515: maybe in a future change the io class can add a method that does this?

benhillis: Or a new handle type that creates a thread, or maybe a flag to an existing class?

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -622,6 +636,47 @@ static int Run(std::wstring_view commandLine)
...
+
+    wil::com_ptr<IWSLAProcess> process;
+    THROW_IF_FAILED(container->GetInitProcess(&process));
+
+    ClientRunningWSLAProcess runningProcess(std::move(process), {});
+
+    wil::unique_handle stdinLogs;
+    wil::unique_handle stdoutLogs;
+    wil::unique_handle stderrLogs;
+    THROW_IF_FAILED(container->Attach((ULONG*)&stdinLogs, (ULONG*)&stdoutLogs, (ULONG*)&stderrLogs));
```
benhillis, line 662: I'm surprised the cast is needed.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -294,6 +294,7 @@ typedef char WSLAContainerId[WSLA_CONTAINER_ID_LENGTH + 1] ;
 ]
 interface IWSLAContainer : IUnknown
 {
+    HRESULT Attach([out] ULONG* StdIn, [out] ULONG* StdOut, [out] ULONG* StdErr);
```
benhillis, line 297: CreateLxProcess uses output HANDLE* in the idl, why can't we do the same here?

benhillis (approved):
Couple non-blocking questions.

src/windows/wsladiag/wsladiag.cpp
OneBlue: Agreed, there's a task for that

src/windows/wsladiag/wsladiag.cpp
OneBlue: It is because HANDLE* is 8 bytes but ULONG* is 4

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: That's because we return different types of handles depending on the type of process (either hvsocket, or pipes), and COM doesn't support that. 

We might be able to clean this up a bit when we refactor the API by having a custom HANDLE typedef

PR #14115, opened 2026-01-27
Resolve minor issues with notice.txt generation

benhillis:
This PR resolves some minor issues I noticed while looking at the notice file generation logs.

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -22,12 +22,37 @@
         },
         {
             "component": {
-                "type": "other",
-                "other": {
-                    "name": "libarchive",
-                    "version": "3.7.7",
-                    "downloadUrl": "https://github.com/libarchive/libarchive/releases/download/v3.7.7/libarchive-3.7.7.tar.gz",
-                    "hash": "sha1:918692098b11db61aff23684ab04f375e4a68f69"
+                "type": "git",
+                "git": {
+                    "repositoryUrl": "https://github.com/libarchive/libarchive",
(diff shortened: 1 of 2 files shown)


packages.config
```diff
@@ -12,7 +12,7 @@
   <package id="Microsoft.NETCore.App.Runtime.win-x64" version="10.0.0" />
   <package id="Microsoft.RemoteDesktop.Client.MSRDC.SessionHost" version="1.2.6676" />
   <package id="Microsoft.Taef" version="10.100.251104001" targetFramework="native" />
-  <package id="Microsoft.Windows.ImplementationLibrary" version="1.0.251108.1" targetFramework="native" />
+  <package id="Microsoft.Windows.ImplementationLibrary" version="1.0.260126.7" targetFramework="native" />
```
benhillis, line 15: might have to push this to the feed, we will see.

cgmanifest.json
```diff
@@ -22,14 +22,39 @@
...
+            "component": {
+                "type": "git",
+                "git": {
+                    "repositoryUrl": "https://github.com/microsoft/WindowsAppSDK",
+                    "commitHash": "0cddda05034d6aaceb5bfd88ab00a0269345737d"
+                }
+            }
+        },
+        {
+            "component": {
```
OneBlue, line 51: Do we need to explicitly reference those ? I expected CG to pick those up from `packages.config`

benhillis: yeah I thought so too, but they were flagged in the notice pipeline as needing to be added manually.

benhillis:
I need to look at this more...

PR #14117, opened 2026-01-27
Rework the process & container API

OneBlue:

The objective of this change is to simplify the APIs for process & container creation.

As part of that, this change adds support and test coverage for: 

- Custom stop signals
- Default signal & timeouts in WSLAContainer::Stop()
- Custom working directories
- Custom hostnames & domainanmes
- Custom users

This change also removes support for custom FDs (including opening files as fds) since it's unused 

## PR Checklist

### DIFF
--- a/src/linux/init/WSLAInit.cpp
+++ b/src/linux/init/WSLAInit.cpp
@@ -300,7 +300,7 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_TTY_RELAY
 
     pollfd pollDescriptors[3];
 
-    pollDescriptors[0].fd = Message.TtyInput;
+    pollDescriptors[0].fd = Message.Socket;
     pollDescriptors[0].events = POLLIN;
     pollDescriptors[1].fd = Message.TtyMaster;
     pollDescriptors[1].events = POLLIN;
@@ -402,7 +402,7 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_TTY_RELAY
                 break;
             }
 
-            bytesWritten = UtilWriteBuffer(Message.TtyOutput, buffer.data(), bytesRead);
+            bytesWritten = UtilWriteBuffer(Message.Socket, buffer.data(), bytesRead);
             if (bytesWritten < 0)
             {
                 LOG_ERROR("write failed {}", errno);
@@ -437,8 +437,7 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_TTY_RELAY
     }
 
     // Shutdown sockets and tty
-    UtilSocketShutdown(Message.TtyInput, SHUT_WR);
-    UtilSocketShutdown(Message.TtyOutput, SHUT_WR);
+    UtilSocketShutdown(Message.Socket, SHUT_WR);
 }
 
 void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_FORK& Message, const gsl::span<gsl::byte>& Buffer)
--- a/src/windows/common/WslClient.cpp
+++ b/src/windows/common/WslClient.cpp
@@ -34,7 +34,6 @@ using wsl::windows::common::WSLAProcessLauncher;
 using namespace wsl::windows::common;
 using namespace wsl::shared;
 using namespace wsl::windows::common::distribution;
-using wsl::windows::common::ProcessFlags;
 
 static bool g_promptBeforeExit = false;
 
@@ -1534,10 +1533,8 @@ int WslaShell(_In_ std::wstring_view commandLine)
     std::string shell = "/bin/bash";
     std::string cmd;
 
-    std::string containerImage;
     bool help = false;
     bool noTty = false;
-    bool exec = false;
     std::wstring debugShell;
 
     std::wstring storagePath;
@@ -1557,11 +1554,8 @@ int WslaShell(_In_ std::wstring_view commandLine)
     parser.AddArgument(Utf8String(rootVhdTypeOverride), L"--fstype");
     parser.AddArgument(storagePath, L"--storage");
     parser.AddArgument(Integer(reinterpret_cast<int&>(sessionSettings.NetworkingMode)), L"--networking-mode");
-    parser.AddArgument(Utf8String(containerImage), L"--image");
     parser.AddArgument(debugShell, L"--debug-shell");
     parser.AddArgument(noTty, L"--no-tty");
-    parser.AddArgument(Utf8String(cmd), L"--cmd");
-    parser.AddArgument(exec, L"--exec");
     parser.AddArgument(help, L"--help");
     parser.Parse();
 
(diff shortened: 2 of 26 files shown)


src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -353,7 +339,7 @@ interface IWSLASession : IUnknown
     HRESULT ListContainers([out, size_is(, *Count)] struct WSLA_CONTAINER** Images, [out] ULONG* Count);
 
     // Create a process at the VM level. This is meant for debugging.
-    HRESULT CreateRootNamespaceProcess([in] const struct WSLA_PROCESS_OPTIONS* Options, [out] IWSLAProcess** Process, [out] int* Errno);
+    HRESULT CreateRootNamespaceProcess([in] LPCSTR Executable, [in] const struct WSLA_PROCESS_OPTIONS* Options, [out] IWSLAProcess** Process, [out] int* Errno);
```
kevpar, line 376: Why do we take `Executable` here but not for `IWSLAContainer::Exec`?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -173,24 +154,29 @@ struct WSLA_CONTAINER_NETWORK
...
+    WSLAContainerFlagsRm = 1,
+    WSLAContainerFlagsGpu = 2,
+    WSLAContainerFlagsAttach = 4,
+} WSLAContainerFlags;
 
 struct WSLA_CONTAINER_OPTIONS
 {
     LPCSTR Image;
     [unique] LPCSTR Name;
+    WSLAStringArray Entrypoint;
```
kevpar, line 202: I assume this will be passed to docker as the Entrypoint value for the container. Is the idea that InitProcessOptions.CommandLine will be Cmd?

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -173,24 +154,29 @@ struct WSLA_CONTAINER_NETWORK
...
-enum WSLA_CONTAINER_FLAGS
+typedef enum _WSLAContainerFlags
 {
-    WSLA_CONTAINER_FLAG_ENABLE_GPU = 1
-} ;
-
+    WSLAContainerFlagsNone = 0,
+    WSLAContainerFlagsRm = 1,
+    WSLAContainerFlagsGpu = 2,
+    WSLAContainerFlagsAttach = 4,
```
kevpar: What does this do?

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Exec doesn't support explicitly specifying the executable (see: https://docs.docker.com/reference/api/engine/version/v1.52/#tag/Exec/operation/ContainerExec)

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Yes exactly

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Rm controls the AutoDelete behavior, Gpu enables GPU support in the container, and attach creates an initial attached relay (needed for non-detached scenarios, otherwise output can be lost between the start() and attach() calls)

src/windows/wslaservice/inc/wslaservice.idl
kevpar: Ah, just a variance in launching directly vs through Docker, I see.

src/windows/wslaservice/inc/wslaservice.idl
kevpar: Shouldn't the Attach option be something specified on Start? Seems odd to put it on Create.

OneBlue: It could also be on start() yeah, having it here has the benefit of having all flags in one place, but I'm OK with either

kevpar: IMO Start is better, if there's no technical reason we can't

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -131,16 +100,28 @@ struct WSLA_IMAGE_INFORMATION
...
 {
-    [unique] LPCSTR Executable;
     [unique] LPCSTR CurrentDirectory;
-    [size_is(CommandLineCount)] LPCSTR* CommandLine;
-    ULONG CommandLineCount;
-    [unique, size_is(EnvironmentCount)] LPCSTR* Environment;
-    ULONG EnvironmentCount;
-    [unique, size_is(FdsCount)] WSLA_PROCESS_FD *Fds;
-    int FdsCount;
+    [unique] LPCSTR User;
```
ptrivedi, line 152: what about the executable/entrypoint name?

src/windows/wslaservice/inc/wslaservice.idl
ptrivedi, line 202: Okay I see the Entrypoint was moved to here.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -32,6 +32,19 @@ using namespace wsl::windows::common::docker_schema;
 
 namespace {
 
+std::vector<std::string> StringArrayToVector(const WSLAStringArray& array)
+{
```
ptrivedi, line 36: Hmm, why not use vector.assign() instead of this helper? We will have to check if assign() does a reserve() first, if being able to reserve() was one of the reasons for this helper.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -593,39 +570,23 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
...
-        request.Entrypoint = std::vector<std::string>{containerOptions.InitProcessOptions.Executable};
-    }
-
-    for (DWORD i = 0; i < containerOptions.InitProcessOptions.EnvironmentCount; i++)
-    {
-        request.Env.push_back(containerOptions.InitProcessOptions.Environment[i]);
-    }
+    request.Cmd = StringArrayToVector(containerOptions.InitProcessOptions.CommandLine);
+    request.Entrypoint = StringArrayToVector(containerOptions.Entrypoint);
+    request.Env = StringArrayToVector(containerOptions.InitProcessOptions.Environment);
```
ptrivedi, line 616: see comment above where StringArrayToVector helper is defined

src/windows/wslaservice/exe/WSLAVirtualMachine.cpp
```diff
@@ -529,25 +526,25 @@ void WSLAVirtualMachine::ConfigureNetworking()
...
     auto prepareCommandLine = [&](const auto& sockets) {
+
         gnsChannelFd = sockets[0].Fd;
         socketFdArg = std::to_string(gnsChannelFd);
         cmd.emplace_back(socketFdArg.c_str());
 
         if (sockets.size() > 1)
         {
-            dnsChannelFd = sockets[1].Fd;
+            dnsChannelFd = sockets[0].Fd;
```
ptrivedi: sockets[0].Fd is set to the gnsChannelFd above

ptrivedi (approved):

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: I found the format of having to call assign with beginning and end a bit heavy (and easy to get wrong), this guaranteed that things are done correctly once. We can probably use assign() in that method though

src/windows/wslaservice/inc/wslaservice.idl
OneBlue: Ok. I'll remove this flag from this enum for now. I'll do the Start() changes in a future change

src/linux/init/WSLAInit.cpp
```diff
@@ -300,7 +300,7 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_TTY_RELAY
 
     pollfd pollDescriptors[3];
 
-    pollDescriptors[0].fd = Message.TtyInput;
+    pollDescriptors[0].fd = Message.Socket;
```
OneBlue, line 303: Switching to using the same socket for tty input & input to match the container & exec processes. It also simplifies things a bit

src/windows/common/WslClient.cpp
```diff
@@ -1630,78 +1624,10 @@ int WslaShell(_In_ std::wstring_view commandLine)
     Info.cbSize = sizeof(Info);
     THROW_IF_WIN32_BOOL_FALSE(::GetConsoleScreenBufferInfoEx(Stdout, &Info));
 
-    if (containerImage.empty())
```
OneBlue, line 1633: Removing support for this since wsladiag can do run now

test/windows/WSLATests.cpp
```diff
@@ -681,111 +675,6 @@ class WSLATests
         ValidateNetworking(WSLANetworkingModeVirtioProxy);
     }
```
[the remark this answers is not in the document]
OneBlue: Support for this is being removed

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -82,18 +58,42 @@ typedef enum _WSLAFD
...
 } WSLAFD;
 
-typedef enum _WSLAFDFlags
-{
-    WSLAFDFlagsNone = 0,
-    WSLAFDFlagsLogs = 1, // Replay previous logs
-    WSLAFDFlagsStream = 2, // Stream new logs.
-} WSLAFDFlags;
-
 typedef enum _WSLASignal
```
benhillis, line 61: nice addition, maybe these shouldn't be WSLA specific though?

benhillis (approved):
Nice simplification, much cleaner.

PR #14118, opened 2026-01-28
Persist port mapping and volume information for containers across WSLA sessions

kvega005:
Added support to persist WSLA container metadata (port mappings, volume mounts, TTY setting) in a Docker label. This enables running containers with their full configuration when opening them from previous sessions.

### Changes

__New Container Metadata Schema__ (`docker_schema.h`)

- Added `WSLAContainerMetadataLabel` constant for the label key

- Added `ContainerMetadata` struct with JSON schema for storing:

  - Volume mounts (host path, VM path, container path, read-only flag)
  - Port mappings (host port, VM port, container port, family, mapped-to-host flag)
  - TTY setting

__Container Creation__ (`WSLAContainer.cpp`)

- On `Create()`, serialize metadata to JSON and store in Docker label

__Container Recovery__ (`WSLAContainer.cpp`)

- On `Open()`, read metadata from Docker label
- Re-mount volumes and re-map ports using stored configuration

__Tests__ (`WSLATests.cpp`)

- Added `ContainerVolumeAndPortRecoveryFromStorage` test that:

  - Creates a container with volumes and ports
  - Recovers it in a new session and verifies port/volume work
  - Verifies session creation succeeds when host folder for volume is missing

###

## PR Checklist


src/windows/wslaservice/exe/WSLAContainerMetadata.h
```diff
@@ -0,0 +1,61 @@
...
+{
+    std::string HostPath;
+    std::string ParentVMPath;
+    std::string ContainerPath;
+    bool ReadOnly{};
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(WSLAVolumeMount, HostPath, ParentVMPath, ContainerPath, ReadOnly);
+};
+
+struct ContainerMetadata
```
OneBlue: For something like this I'd recommend doing this: 

```
struct VersionedContainerMetaData
{
    std::optional<ContainerMetadataV1> v1;
    // in the future, we can do: 
    // std::optional<ContainerMetadataV2> v2;
};

```

That way we can easily add a V2 later without breaking the schema

test/windows/WSLATests.cpp
```diff
@@ -2543,6 +2543,78 @@ class WSLATests
...
+            VERIFY_SUCCEEDED(container.Get().Stop(WSLASignalSIGKILL, 0));
+            VERIFY_SUCCEEDED(container.Get().Delete());
+        }
+
+        // Delete the host folder to simulate volume folder being missing on recovery
+        cleanup.reset();
+
+        // Create a new session - this should succeed even though the volume folder is gone
+        auto session = CreateSession(GetDefaultSessionSettings(L"recovery-test-vp", true, WSLANetworkingModeNAT));
+        VERIFY_FAILED(session->OpenContainer(containerName.c_str(), nullptr));
```
OneBlue: nit: I recommend using `VERIFY_ARE_EQUAL` here so we can validate that the error code is what we expect

test/windows/WSLATests.cpp
```diff
@@ -2543,6 +2543,78 @@ class WSLATests
...
+    }
+
```
OneBlue, line 2754: nit: I'd recommend adding a test case of a container that has an invalid / missing tag, just to make sure that we handle that case correctly.

Long term we should add a mechanism to reports warnings for to the caller, but for now just ignoring the container should be OK

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -147,7 +147,7 @@ struct WSLA_PROCESS_OPTIONS
 
 struct WSLA_VOLUME
 {
-    LPCWSTR HostPath;
+    LPCSTR HostPath;
```
OneBlue: I would recommend keeping this as a wide string at the COM API level, since we've designed the API to use wide strings for Windows path, and convert internally when we generate the metadata json

test/windows/WSLATests.cpp
OneBlue: Actually given that we're passing nullptr here, this might be failing with E_POINTER

OneBlue (commented):
Change looks good, added a couple minor comments

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -654,41 +721,34 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Open(
...
+        metadataIt == dockerContainer.Labels.end(),
+        "Cannot open WSLA container %hs: missing WSLA metadata label",
+        dockerContainer.Id.c_str());
+
+    auto metadata = wsl::shared::FromJson<ContainerMetadata>(metadataIt->second.c_str());
+
+    // TODO: Offload volume mounting and port mapping to the Start() method so that its still possible
+    // to open containers that are not running.
+    MountVolumes(metadata.Volumes, parentVM);
+    auto [vmPorts, errorCleanup] = MapPorts(metadata.Ports, parentVM);
```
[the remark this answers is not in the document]
kvega005: Mount volumes handle its own error cleanup. Unmounting previous volumes if one fails to mount.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -145,6 +206,25 @@ std::string ExtractContainerName(const std::vector<std::string>& names, const st
     return name;
 }
 
+WSLAContainerMetadataV1 ParseContainerMetadata(const std::string& json)
+{
+    auto wrapper = wsl::shared::FromJson<WSLAContainerMetadata>(json.c_str());
+
+    WI_ASSERT(wrapper.Version == WSLAContainerMetadataV1::Version);
```
OneBlue: I would recommend removing the `Version` field here, so that if we ever add a V2, the downgrade path won't break. 

The thinking would be that if we ever add a V2 field, the future version of the service will write both the v1 and v2 entries, so older versions of the service will still be able to function

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -145,6 +206,25 @@ std::string ExtractContainerName(const std::vector<std::string>& names, const st
...
+    WI_ASSERT(wrapper.V1.has_value());
```
OneBlue: I'd recommend `THROW_HR_IF(E_UNEXPECTED, !wrapper.V1.has_value())` since this might able if we ever decide to drop V1 support at some point, and if it does we should avoid crashing older versions

OneBlue (approved):

benhillis (dismissed):
I think some of these copilot comments are legit.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -716,42 +790,35 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Open(
...
-    for (const auto& port : dockerContainer.Ports)
-    {
-        if (port.PublicPort != 0)
-        {
-            ports.push_back({port.PublicPort, port.PublicPort, port.PrivatePort, AF_INET, false});
-        }
-    }
+    auto metadataIt = dockerContainer.Labels.find(WSLAContainerMetadataLabel);
+    THROW_HR_IF_MSG(
+        E_INVALIDARG,
```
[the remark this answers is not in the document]
kvega005: Today, WSLASession::OpenContainer throws E_UNEXPECTED if the container exists in docker, but it's not in our list.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -716,42 +790,35 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Open(
...
+        metadataIt == dockerContainer.Labels.end(),
+        "Cannot open WSLA container %hs: missing WSLA metadata label",
+        dockerContainer.Id.c_str());
+
+    auto metadata = ParseContainerMetadata(metadataIt->second.c_str());
+
+    // TODO: Offload volume mounting and port mapping to the Start() method so that its still possible
+    // to open containers that are not running.
```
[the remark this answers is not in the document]
kvega005: This is planned for a follow-up PR.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -722,16 +713,20 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
 
     for (ULONG i = 0; i < containerOptions.VolumesCount; i++)
     {
-        const WSLA_VOLUME& volume = containerOptions.Volumes[i];
-        std::string parentVMPath = std::format("/mnt/wsla/{}/volumes/{}", containerOptions.Name, i);
+        GUID volumeId;
+        THROW_IF_FAILED(CoCreateGuid(&volumeId));
+
+        auto parentVMPath = std::format("/mnt/{}", wsl::shared::string::GuidToString<char>(volumeId));
```
OneBlue, line 730: nit: I liked having the container name as prefix, but now that it might be empty, we could use the container id (since it's always guaranteed to be set)

OneBlue (approved):

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -85,59 +89,140 @@ auto ProcessPortMappings(const WSLA_CONTAINER_OPTIONS& options, WSLAVirtualMachi
...
+                WI_ASSERT(allocatedIt != allocatedPorts.end());
+                port.VmPort = *allocatedIt;
+                vmPorts->insert(*allocatedIt);
+                ++allocatedIt;
+            }
         }
     }
-    else if (options.ContainerNetwork.ContainerNetworkType == WSLA_CONTAINER_NETWORK_HOST)
+
+    // Allocate VM ports for host mode (VmPort != 0).
```
ptrivedi: Not sure what is meant by "host mode" here. But it looks from the diff like this is taking the pre-determined VmPorts, checking for conflicts, then mapping the vm ports to the windows host

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -85,59 +89,140 @@ auto ProcessPortMappings(const WSLA_CONTAINER_OPTIONS& options, WSLAVirtualMachi
...
+WSLA_CONTAINER_STATE DockerStateToWSLAState(ContainerState state)
+{
+    switch (state)
+    {
+    case ContainerState::Created:
+        return WSLA_CONTAINER_STATE::WslaContainerStateCreated;
+    case ContainerState::Running:
+    case ContainerState::Restarting:
+        return WSLA_CONTAINER_STATE::WslaContainerStateRunning;
+    case ContainerState::Paused:
```
ptrivedi: Should Paused state be clubbed with Created instead of Exited? I believe on both those types of states one could all container Start and be able to get a running container.

src/windows/wslaservice/exe/WSLAContainer.cpp
kvega005: The problem is at this point I don't think we have a container Id available yet. We get the ID after making the create request. I can change that if we move the logic over to the Start method in a follow-up PR.

OneBlue: Ah that's a good point. Agreed let's just do that as part of the change that moves resource allocation to Start()

test/windows/WSLATests.cpp
```diff
@@ -2485,6 +2485,100 @@ class WSLATests
...
+            VERIFY_ARE_EQUAL(result.Code, 0L);
+        }
+
+        {
+            auto session = CreateSession(GetDefaultSessionSettings(L"persistence-invalid-metadata", true));
+
+            // Try to open the container - this should fail due to missing metadata.
+            wil::com_ptr<IWSLAContainer> container;
+            auto hr = session->OpenContainer("test-invalid-metadata", &container);
+            VERIFY_ARE_EQUAL(hr, E_UNEXPECTED);
```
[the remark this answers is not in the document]
OneBlue: That's not true: 

```
 auto metadataIt = dockerContainer.Labels.find(WSLAContainerMetadataLabel);
    THROW_HR_IF_MSG(
        E_INVALIDARG);

```

OneBlue (approved):

src/windows/wslaservice/exe/WSLAContainer.cpp
kvega005: Host mode means: WSLA_CONTAINER_NETWORK_HOST. In this network mode, we use the same port for both the VM and container. So we just need to allocate the port on the parent vm. I will try to clarify this in the comment.

src/windows/wslaservice/exe/WSLAContainer.cpp
kvega005: We do not really support pausing containers today and have not looked into what we need to resume a paused container, so I figured it's probably not worth handling in this PR.

ptrivedi: Maybe leave a TODO here to revisit. I think a container in 'Restarting' state should also likely not be considered 'Running'. Though not sure what could cause those containers to go into these states if we don't support these operations (e.g. a container with a suspended process would be in Paused state)... in which case, maybe those should be considered invalid states?

kvega005: Thats a good point, we can consider them invalid states for now.

PR #14119, opened 2026-01-28
Move from shipping the initrd to generating it during install.

benhillis:
I think it's pretty wasteful to ship both the init binary and the initrd.img when we can easily create the initrd.img dynamically.

This change implements this by generating the initrd.img during installation. I timed this on my dev box and this takes around 50ms, so very quick. This allows us to remove the dependency on perl in our build system and reduces package size.


src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -175,6 +175,9 @@ void WslCoreVm::Initialize(const GUID& VmId, const wil::shared_handle& UserToken
     // Query the Windows version.
     m_windowsVersion = wsl::windows::common::helpers::GetWindowsVersion();
 
+    // Ensure initrd.img exists in the tools folder.
+    m_initrdPath = wsl::windows::common::filesystem::EnsureCpioInitrd(m_rootFsPath);
```
OneBlue: Could we do this during installation instead of during the first launch ? We could have a custom action to generate this

OneBlue: That would give us the best of both worlds: not initrd in the package, and no delay during the first boot

OneBlue: And it would completely the remove for the need for the locking logic

OneBlue: Oh and actually we need to be absolutely certain that the initrd is updated when the package is updated, which the current approach doesn't do since the file would be left behind (I think MSI will leave the folder if it contains a file that it doesn't recognize)

benhillis: That is what the update in the WIX file is for it will remove the file. I wrote a test that validate this.

benhillis: Good call on moving the logic to the installer. I will do that.

src/windows/common/filesystem.cpp
```diff
@@ -1115,6 +1115,126 @@ void wsl::windows::common::filesystem::UpdateInit(_In_ PCWSTR BasePath, _In_ ULO
...
+        // Write padding to align to 4 bytes
+        if (headerPadding > 0)
+        {
+            char padding[4] = {0};
+            THROW_IF_WIN32_BOOL_FALSE(WriteFile(destFile.get(), padding, static_cast<DWORD>(headerPadding), &bytesWritten, nullptr));
+        }
+    };
+
+    // Write init file header
+    writeCpioHeader(initSize, SourcePath.filename().string().c_str());
```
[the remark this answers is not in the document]
benhillis: This was intentional.

src/windows/common/filesystem.cpp
```diff
@@ -1115,6 +1115,126 @@ void wsl::windows::common::filesystem::UpdateInit(_In_ PCWSTR BasePath, _In_ ULO
...
+{
+    static wil::srwlock s_initrdLock;
+
+    auto initPath = ToolsPath / L"init";
+    auto initrdPath = ToolsPath / LXSS_VM_MODE_INITRD_NAME;
+
+    auto lock = s_initrdLock.lock_exclusive();
+    if (!FileExists(initrdPath.c_str()))
+    {
+        CreateCpioInitrd(initPath, initrdPath);
```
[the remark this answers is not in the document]
benhillis: Good call. Will add this.

src/windows/common/filesystem.cpp
```diff
@@ -1115,6 +1115,116 @@ void wsl::windows::common::filesystem::UpdateInit(_In_ PCWSTR BasePath, _In_ ULO
...
+        // Write padding to align to 4 bytes
+        if (headerPadding > 0)
+        {
+            char padding[4] = {0};
+            THROW_IF_WIN32_BOOL_FALSE(WriteFile(destFile.get(), padding, static_cast<DWORD>(headerPadding), &bytesWritten, nullptr));
+        }
+    };
+
+    // Write init file header
+    writeCpioHeader(initSize, SourcePath.filename().string().c_str());
```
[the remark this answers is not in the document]
benhillis: This was intentional.

src/windows/wslinstall/DllMain.cpp
```diff
@@ -874,6 +872,157 @@ extern "C" UINT __stdcall UnregisterLspCategories(MSIHANDLE install)
...
+    });
+
+    auto writeCpioHeader = [&destFile](DWORD fileSize, PCSTR name) {
+        // CPIO newc header: magic(6) + 13 fields of 8 hex chars each = 110 bytes
+        constexpr size_t headerSize = 110;
+        const auto nameLen = strlen(name) + 1;
+        const auto headerPadding = (4 - ((headerSize + nameLen) % 4)) % 4;
+
+        // Get current time for mtime
+        const auto mtime = static_cast<DWORD>(time(nullptr));
```
OneBlue: nit: we should probably cast to ULONGLONG here to make sure we don't force the timestamp to 32 bites

src/windows/wslinstall/DllMain.cpp
```diff
@@ -874,6 +872,157 @@ extern "C" UINT __stdcall UnregisterLspCategories(MSIHANDLE install)
...
+
+        char header[headerSize + 1];
+        sprintf_s(
```
OneBlue: nit: I have a small preference over using std::format for something like this. We could also simplify it a bit: 

I also recommend locking down the locale, given how critical this method will be during the install process 

```

auto header = std::format(std::locale::classic(), "070701{:08}{:08}", 0, fileSize, ....)";

```

src/windows/wslinstall/DllMain.cpp
```diff
@@ -874,6 +872,157 @@ extern "C" UINT __stdcall UnregisterLspCategories(MSIHANDLE install)
...
+
+    // Pad the archive to 512-byte boundary
+    constexpr DWORD archiveBlockSize = 512;
+    LARGE_INTEGER currentPos{};
+    THROW_IF_WIN32_BOOL_FALSE(SetFilePointerEx(destFile.get(), {}, &currentPos, FILE_CURRENT));
+
+    const auto archivePadding = (archiveBlockSize - (currentPos.LowPart % archiveBlockSize)) % archiveBlockSize;
+    if (archivePadding > 0)
+    {
+        auto paddingBuffer = std::make_unique<char[]>(archivePadding);
```
OneBlue: nit: an std::vector would be a bit cleaner (since it's directly initialized to zero)

src/windows/wslinstall/DllMain.cpp
```diff
@@ -874,6 +872,157 @@ extern "C" UINT __stdcall UnregisterLspCategories(MSIHANDLE install)
     return NOERROR;
 }
 
+void CreateCpioInitrd(_In_ const std::filesystem::path& SourcePath, _In_ const std::filesystem::path& DestPath)
```
OneBlue: nit: I think it would be a good idea to move to method to `wslutil` so we can have unit tests that validate that it behaves correctly on different file sizes (to make sure that the padding logic works correctly

benhillis: not a bad iea.

src/windows/wslinstall/DllMain.cpp
benhillis: won't the printf below do that anyway?

src/windows/wslinstall/DllMain.cpp
benhillis: ok, I'll consider this. The benefit of the printf approach is no allocation is needed.

src/windows/wslinstall/DllMain.cpp
benhillis: yep fair, or even a stack buffer since it's small, let me think about it.

src/windows/wslinstall/DllMain.cpp
OneBlue: I don't think this since we write the integer as a string

benhillis: We need to make sure we don't print any more than 8 chars for the timestamp, ULONGLONG would print more if the value was 0x1ffffffff or bigger.

src/windows/wslinstall/DllMain.cpp
benhillis: %08X is locale-independent, I think this is fine.

src/windows/wslinstall/DllMain.cpp
OneBlue: Oh, that's bad news for the linux boot process. 

TBH we could probably just hardcode this timestamp anyway, it's not really used for anything

OneBlue (approved):

PR #14121, opened 2026-01-28
Implement image save and export container

ptrivedi:


src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -515,6 +515,110 @@ void WSLASession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request,
     }
 }
 
+HRESULT WSLASession::ExportContainer(ULONG OutHandle, LPCSTR ContainerID, IProgressCallback* ProgressCallback, WSLA_ERROR_INFO* Error)
+try
+{
+    UNREFERENCED_PARAMETER(ProgressCallback);
+    RETURN_HR_IF_NULL(E_POINTER, ContainerID);
+    std::lock_guard lock{m_lock};
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_dockerClient.has_value());
```
benhillis, line 513: nit: we tend to like to have blank lines after throw or return statements to make things more readable.

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -515,6 +515,110 @@ void WSLASession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request,
...
+    std::pair<uint32_t, std::unique_ptr<DockerHTTPClient::HTTPRequestContext>>& RequestCodePair, ULONG OutputHandle, WSLA_ERROR_INFO* Error)
+{
+    wil::unique_handle imageFileHandle{wsl::windows::common::wslutil::DuplicateHandleFromCallingProcess(ULongToHandle(OutputHandle))};
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_dockerClient.has_value());
+    relay::MultiHandleWait io;
+    std::optional<boost::beast::http::status> importResult;
+    auto onCompleted = [&]() {
+        io.Cancel();
+        WSL_LOG("OnCompletedCalledForExport", TraceLoggingValue("OnCompletedCalledForExport", "Content"));
+    };
```
benhillis, line 532: newline after closing braces.

benhillis: should add this to the clang formatter...

src/windows/wslaservice/exe/WSLASession.cpp
ptrivedi: Looks like before and after

src/windows/wslaservice/exe/WSLASession.cpp
ptrivedi: too many newlines! but done

src/windows/wslaservice/exe/DockerHTTPClient.cpp
```diff
@@ -501,5 +512,16 @@ std::pair<uint32_t, wil::unique_socket> DockerHTTPClient::SendRequest(
         }
     }
 
-    return {parser.get().result_int(), wil::unique_socket{context->stream.release()}};
+    return {parser.get().result_int(), std::move(context)};
```
OneBlue: Unfortunately we can't call the parser on the stream like this, because this will use boost's built-in reader to feed data into the parser, and that might read past the HTTP header and consume bytes of the body. 

That's why we have [this logic](https://github.com/microsoft/WSL/blob/535759e0d596c8a5f1ab8336c581da4f60cd0905/src/windows/wslaservice/exe/DockerHTTPClient.cpp#L432) in place when we need the socket to be returned 

If we both need to know the HTTP result, and read bytes from the body, `DockerHttpResponseHandle` would probably be our best call. 

We might need to adapt RelayHandle to support that though

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -758,6 +758,11 @@ void WSLAContainerImpl::Inspect(LPSTR* Output)
     }
 }
 
+void WSLAContainerImpl::GetID(LPSTR* Output)
```
OneBlue, line 792: We can remove those since they're already implemented in https://github.com/microsoft/WSL/blob/535759e0d596c8a5f1ab8336c581da4f60cd0905/src/windows/wslaservice/exe/WSLAContainer.cpp#L873C1-L874C1

test/windows/WSLATests.cpp
```diff
@@ -467,6 +467,113 @@ class WSLATests
...
+                imageTar.c_str(), GENERIC_WRITE | GENERIC_READ, FILE_SHARE_READ, nullptr, TRUNCATE_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr)};
+            VERIFY_IS_FALSE(INVALID_HANDLE_VALUE == imageTarFileHandle.get());
+            LARGE_INTEGER fileSize{};
+            VERIFY_IS_TRUE(GetFileSizeEx(imageTarFileHandle.get(), &fileSize));
+            VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, false);
+            WSLA_ERROR_INFO errorInfo{};
+            VERIFY_SUCCEEDED(m_defaultSession->SaveImage(HandleToULong(imageTarFileHandle.get()), "hello-world:latest", nullptr, &errorInfo));
+            VERIFY_ARE_EQUAL(0 == errorInfo.UserErrorMessage, true);
+            VERIFY_IS_TRUE(GetFileSizeEx(imageTarFileHandle.get(), &fileSize));
+            VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, true);
```
OneBlue, line 500: I would recommend something like calling `file` or `tar tf` to make sure that the tar is valid, or maybe try to import the image back

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -344,13 +345,15 @@ interface IWSLASession : IUnknown
...
+    HRESULT SaveImage([in] ULONG OutputHandle, [in] LPCSTR ImageNameOrID, [ in, unique ] IProgressCallback * ProgressCallback, [in, out, unique] WSLA_ERROR_INFO * ErrorInfo);
     HRESULT ListImages([out, size_is(, *Count)] struct WSLA_IMAGE_INFORMATION** Images, [out] ULONG* Count);
     HRESULT DeleteImage([in] const struct WSLA_DELETE_IMAGE_OPTIONS* Options, [out, size_is(, *Count)] struct WSLA_DELETED_IMAGE_INFORMATION** DeletedImages, [out] ULONG* Count, [in, out, unique] WSLA_ERROR_INFO* ErrorInfo);
 
     // Container management.
     HRESULT CreateContainer([in] const struct WSLA_CONTAINER_OPTIONS* Options, [out] IWSLAContainer** Container, [in, out, unique] WSLA_ERROR_INFO* ErrorInfo);
     HRESULT OpenContainer([in] LPCSTR Id, [out] IWSLAContainer** Container);
     HRESULT ListContainers([out, size_is(, *Count)] struct WSLA_CONTAINER** Images, [out] ULONG* Count);
+    HRESULT ExportContainer([in] ULONG OutputHandle, [in] LPCSTR ContainerID, [ in, unique ] IProgressCallback * ProgressCallback, [in,out,unique] WSLA_ERROR_INFO * ErrorInfo);
```
OneBlue: I recommend implementing this method at the `WSLAContainer` level, that way we can lock it while the container export is in progress

test/windows/WSLATests.cpp
```diff
@@ -467,6 +467,113 @@ class WSLATests
...
+        {
+            std::filesystem::path imageTar = std::filesystem::path{g_testDataPath} / L"HelloWorldExportError.tar";
+            wil::unique_handle contTarFileHandle{CreateFileW(
+                imageTar.c_str(), GENERIC_WRITE | GENERIC_READ, FILE_SHARE_READ, nullptr, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr)};
+            VERIFY_IS_FALSE(INVALID_HANDLE_VALUE == contTarFileHandle.get());
+            LARGE_INTEGER fileSize{};
+            VERIFY_IS_TRUE(GetFileSizeEx(contTarFileHandle.get(), &fileSize));
+            VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, false);
+            WSLA_ERROR_INFO errorInfo{};
+            VERIFY_FAILED(m_defaultSession->ExportContainer(HandleToULong(contTarFileHandle.get()), "dummy", nullptr, &errorInfo));
```
OneBlue: I recommend validating the error code here, so we know that we're correctly handling the error (although if we move this method to WSLAContainer that will solve this since Open() will fail on invalid name / id)

OneBlue (changes requested):

src/windows/wslaservice/exe/DockerHTTPClient.cpp
OneBlue: Actually, I read the diff wrong, please ignore this comment :)

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -515,6 +515,110 @@ void WSLASession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request,
...
+        // Save failed, parse the error message.
+        auto error = wsl::shared::FromJson<docker_schema::ErrorResponse>(errorJson.c_str());
+        if (Error != nullptr)
+        {
+            Error->UserErrorMessage = wil::make_unique_ansistring<wil::unique_cotaskmem_ansistring>(error.message.c_str()).release();
+        }
+        THROW_HR_MSG(E_FAIL, "Image save failed: %hs", error.message.c_str());
+    }
+}
+
```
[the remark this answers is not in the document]
ptrivedi: @copilot: will address this in a follow-up PR

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -344,13 +345,15 @@ interface IWSLASession : IUnknown
     HRESULT PullImage([in] LPCSTR ImageUri, [in, unique] const struct WSLA_REGISTRY_AUTHENTICATION_INFORMATION* RegistryAuthenticationInformation, [in, unique] IProgressCallback* ProgressCallback, [in, out, unique] WSLA_ERROR_INFO* ErrorInfo);
     HRESULT LoadImage([in] ULONG ImageHandle, [in, unique] IProgressCallback* ProgressCallback, [in] ULONGLONG ContentLength);
     HRESULT ImportImage([in] ULONG ImageHandle, [in] LPCSTR ImageName, [in, unique] IProgressCallback* ProgressCallback, [in] ULONGLONG ContentLength);
+    HRESULT SaveImage([in] ULONG OutputHandle, [in] LPCSTR ImageNameOrID, [ in, unique ] IProgressCallback * ProgressCallback, [in, out, unique] WSLA_ERROR_INFO * ErrorInfo);
```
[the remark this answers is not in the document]
ptrivedi: @copilot : fix this for me

src/windows/wslaservice/inc/wslaservice.idl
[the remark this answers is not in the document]
ptrivedi: @copilot: and this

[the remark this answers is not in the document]
ptrivedi: @copilot: also this

test/windows/WSLATests.cpp
ptrivedi: Error code doesn't bubble up to here. We are verifying the error message. This is the same model as is used in ImportImage etc. 
Not opposed to moving it to WSLAContainer though

OneBlue: The error code should definitely get returned to the client (that's the HRESULT that VERIFY_FAILED() is looking at), but yeah this doesn't apply if we move all of this to WSLAContainer, which I think would be the best approach here

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -302,6 +302,7 @@ interface IWSLAContainer : IUnknown
     HRESULT GetInitProcess([out] IWSLAProcess** Process);
     HRESULT Exec([in] const struct WSLA_PROCESS_OPTIONS* Options, [out] IWSLAProcess** Process, [out] int* Errno);
     HRESULT Inspect([out] LPSTR* Output);
+    HRESULT GetID([out] LPSTR * Id);
```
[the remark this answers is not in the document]
ptrivedi: Doing this in followup PR

src/windows/wslaservice/inc/wslaservice.idl
ptrivedi: Taking care of this in a follow-up PR

src/windows/wslaservice/exe/WSLAContainer.cpp
ptrivedi: Yes doing this in a followup PR

src/windows/wslaservice/exe/DockerHTTPClient.cpp
```diff
@@ -513,5 +523,16 @@ std::pair<uint32_t, wil::unique_socket> DockerHTTPClient::SendRequest(
...
 }
+
+std::pair<uint32_t, wil::unique_socket> DockerHTTPClient::SendRequest(
```
OneBlue: I think we should remove this, since ExportContainer() only uses the socket and none of the other fields in the context, we can directly use [this method](https://github.com/microsoft/WSL/blob/fc3dfa7c526dd9efd0fa1c1b9e42196f2f0956ba/src/windows/wslaservice/exe/DockerHTTPClient.cpp#L424C1-L425C1)

test/windows/WSLATests.cpp
ptrivedi: Fixed this, but let's talk around the SendRequest method signatures and around returning result() vs result_int()

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -514,6 +514,129 @@ void WSLASession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request,
...
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_dockerClient.has_value());
+
+    auto retVal = m_dockerClient->SaveImage(ImageNameOrID);
+    SaveImageImpl(retVal, OutHandle, Error);
+    return S_OK;
+}
+CATCH_RETURN();
+
+void WSLASession::SaveImageImpl(std::pair<uint32_t, wil::unique_socket>& SocketCodePair, ULONG OutputHandle, WSLA_ERROR_INFO* Error)
```
OneBlue, line 587: nit: This can probably be factored with ExportContainerImpl(), but that can be done in the followup PR

test/windows/WSLATests.cpp
```diff
@@ -465,6 +465,157 @@ class WSLATests
...
+                imageTar.c_str(), GENERIC_WRITE | GENERIC_READ, FILE_SHARE_READ, nullptr, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr)};
+            VERIFY_IS_FALSE(INVALID_HANDLE_VALUE == contTarFileHandle.get());
+            LARGE_INTEGER fileSize{};
+            VERIFY_IS_TRUE(GetFileSizeEx(contTarFileHandle.get(), &fileSize));
+            VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, false);
+            WSLA_ERROR_INFO errorInfo{};
+            VERIFY_ARE_EQUAL(m_defaultSession->ExportContainer(HandleToULong(contTarFileHandle.get()), "dummy", nullptr, &errorInfo), WSLA_E_CONTAINER_NOT_FOUND);
+            VERIFY_IS_NOT_NULL(errorInfo.UserErrorMessage);
+            std::string errMsg = errorInfo.UserErrorMessage;
+            LogInfo("Error message: %hs", errMsg.c_str());
```
OneBlue, line 609: nit: this can probably be removed (looks like a debug print)

OneBlue (approved):
Couple minor comments, can be done in a followup PR

src/windows/wslaservice/exe/WSLASession.cpp
ptrivedi: Yes planning on this

PR #14126, opened 2026-01-29
cleanup: remove duplicated code from wslc.exe

benhillis:
Multiples times Pierre or I have had to modify copy / pasted code in wsladiag.exe and wslc.exe, this is really wasteful. For now just leave wslc.exe as a stub that can be implemented piece-by-piece.


OneBlue (commented):
LGTM. This can be copied from wsladiag.exe again if / when needed

benhillis:
> LGTM. This can be copied from wsladiag.exe again if / when needed

Agreed, it's very silly to be building the same binary twice and having to update each when we change the interface etc.

OneBlue (approved):

PR #14128, opened 2026-01-30
Refactor container IO handling 

OneBlue:

This changes moves all the asynchronous IO happening under a WSLASession to a single thread, with a unified interface to schedule IO. 

This change also adds support for `wsladiag run -d` and adds test coverage to validate that containers can be started without attached IO, and attached later 

## PR Checklist


src/shared/inc/CommandLine.h
```diff
@@ -44,7 +44,7 @@ struct Argument
     bool Positional;
 };
 
-template <typename T, T Flag>
+template <auto Flag, typename T = std::remove_reference_t<decltype(Flag)>>
```
OneBlue, line 47: This was bothering me, but I finally found a way to automatically detect the flag type here

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -195,6 +195,15 @@ typedef enum _WSLAContainerFlags
 
 cpp_quote("DEFINE_ENUM_FLAG_OPERATORS(WSLAContainerFlags);")
 
+
+typedef enum _WSLAContainerStartFlags
+{
+    WSLAContainerStartFlagsNone = 0,
+    WSLAContainerStartFlagsAttach = 1, // Attach stdio handles on start.
```
OneBlue, line 202: @kevpar: FYI this is the "attach on start" flag that we talked about

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -500,24 +500,27 @@ static void RelayNonTtyProcess(wil::unique_handle&& Stdin, wil::unique_handle&&
...
-        // TODO: Will output CR instead of LF's which can confuse the linux app.
-        // Consider a custom relay logic to fix this.
-        inputThread = std::thread{[&]() {
-            try
-            {
-                wsl::windows::common::relay::InterruptableRelay(GetStdHandle(STD_INPUT_HANDLE), Stdin.get(), exitEvent.get());
-            }
-            CATCH_LOG();
+        // Required because ReadFile() blocks if stdin is a tty.
+        if (wsl::windows::common::wslutil::IsInteractiveConsole())
```
benhillis, line 506: I'm wondering if it would make sense to add this functionality to the Io class (which could also keep track of the optional input thread).

src/windows/wslaservice/exe/ContainerEventTracker.cpp
```diff
@@ -53,36 +53,41 @@ ContainerEventTracker::ContainerTrackingReference::~ContainerTrackingReference()
...
+    m_sessionId(sessionId)
 {
-    auto socket = dockerClient.MonitorEvents();
-    m_thread = std::thread([socket = std::move(socket), this]() mutable { Run(std::move(socket)); }
+    auto onChunk = [this](const gsl::span<char>& buffer) {
+        if (!buffer.empty()) // docker inserts empty lines between events, skip those.
+        {
+            try
+            {
+                OnEvent(std::string{buffer.begin(), buffer.end()});
```
benhillis: we need to copy the string? can't we use std::string_view?

src/windows/wsladiag/wsladiag.cpp
OneBlue: Potentially, I'd want to do that in a followup change, since this PR is already pretty big

benhillis (approved):
minor comments

src/windows/wslaservice/exe/ContainerEventTracker.cpp
OneBlue: Good call, fixed (turns out there's a way to log a non-null terminated string to ETL)

benhillis (approved):

PR #14131, opened 2026-01-30
WSLA: Change default networking mode for WSLA to VirtioProxy

benhillis:
This change updates the default networking mode for WSLA to VirtioProxy in both test entrypoints and the WSLA unit tests. This is the expected shipping configuration of WSLA and all there are no known issues blocking the feature from being enabled.

A few more improvements including IPV6 support are coming soon.

### DIFF
--- a/src/windows/wsladiag/wsladiag.cpp
+++ b/src/windows/wsladiag/wsladiag.cpp
@@ -284,7 +284,7 @@ static wil::com_ptr<IWSLASession> OpenCLISession()
     settings.BootTimeoutMs = 30 * 1000;
     settings.StoragePath = dataFolder.c_str();
     settings.MaximumStorageSizeMb = 10000; // 10GB.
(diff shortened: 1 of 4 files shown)


src/windows/wsladiag/wsladiag.cpp
```diff
@@ -284,7 +284,7 @@ static wil::com_ptr<IWSLASession> OpenCLISession()
     settings.BootTimeoutMs = 30 * 1000;
     settings.StoragePath = dataFolder.c_str();
     settings.MaximumStorageSizeMb = 10000; // 10GB.
-    settings.NetworkingMode = WSLANetworkingModeNAT;
+    settings.NetworkingMode = WSLANetworkingModeVirtioProxy;
```
benhillis, line 287: I did not modify wslc.exe because I have another change that removes this code.

OneBlue (approved):

PR #14132, opened 2026-01-30
Add WSLA runtime inspect schema and mapping

beena352:
Update WSLAContainer::Inspect to return WSLA-schema JSON instead of raw Docker inspect JSON, preserve WSLA host-side port mappings when Docker reports empty bindings, and report host-side mount paths from WSLA state.

## PR Checklist

Inspect now returns WSLA-schema JSON instead of Docker’s raw inspect output. Docker inspect data is mapped through the WSLA runtime schema before being serialized. Port bindings are added without overwriting existing WSLA host-side mappings, and mounts use WSLA host paths. This only affects the inspect output and does not change container runtime behavior.


src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +554,106 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
 
+wsl::windows::service::wsla::WSLAContainerImpl::WslaInspectContainer wsl::windows::service::wsla::WSLAContainerImpl::BuildInspectContainer(
+    const wsl::windows::service::wsla::WSLAContainerImpl::DockerInspectContainer& dockerInspect)
+{
+    WslaInspectContainer wslaInspect{};
+
+    wslaInspect.Id = dockerInspect.Id;
+    wslaInspect.Name = dockerInspect.Name;
+
+    // Remove leading '/' from Docker container names.
```
kvega005, line 679: There is already an ExtractContainerName function in this file. Perhaps you could alter it to fit this purpose, so we don't duplicate that logic.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +554,106 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
     }
 }
 
+wsl::windows::service::wsla::WSLAContainerImpl::WslaInspectContainer wsl::windows::service::wsla::WSLAContainerImpl::BuildInspectContainer(
```
OneBlue: FYI: We can omit the full namespace here, because we have a `using` for WSLAContainerImpl

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +554,106 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+            portBinding.HostPort = std::to_string(e.HostPort);
+
+            if (!hasBinding(portBindings, portBinding.HostIp, portBinding.HostPort))
+            {
+                portBindings.push_back(std::move(portBinding));
+            }
+        }
+
+        // Include Docker configured port bindings.
+        for (const auto& [containerPort, bindings] : dockerInspect.HostConfig.PortBindings)
```
OneBlue: I would recommend directly reading from `m_mappedPorts` and generating the list from there, that should be simpler than having to read the bindings returned by docker and mapping those to our port mappings

test/windows/WSLATests.cpp
```diff
@@ -1728,7 +1713,7 @@ class WSLATests
             VERIFY_ARE_EQUAL(container.State(), WslaContainerStateRunning);
 
             auto details = container.Inspect();
-            VERIFY_ARE_EQUAL(details.HostConfig.NetworkMode, "host");
+            VERIFY_ARE_EQUAL(details.NetworkMode, "host");
```
OneBlue: I would recommend matching docker's JSON structure and keeping the `HostConfig` field here

OneBlue (commented):
This approach LGTM, added a few minor comments

src/windows/wslaservice/exe/WSLAContainer.cpp
kvega005: I think this would be misleading. 

The `HostPort` returned from `dockerInspect.HostConfig.PortBindings` is not the same as the `hostPort` in `m_mappedPorts`. From the docker's perspective the host is actually the WSLA VM. The host in `m_mappedPorts` is the actual Windows host.

kvega005: @OneBlue Do we need this code? Should we just return the port bindings in m_mappedPorts? Those are the only ones that have information about the real host port which is the only thing WSLA users care about. 

Or do we want to create a separate field called something like VMPort and leave the HostPort entry null or empty for the ports that don't have host bindings?

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +554,106 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+        for (const auto& e : m_mappedPorts)
+        {
+            auto portKey = std::format("{}/tcp", e.ContainerPort);
+            auto& portBindings = wslaInspect.Ports[portKey];
+
+            wsla_schema::InspectPortBinding portBinding{};
+            portBinding.HostIp = defaultHostIp(e.Family);
+            portBinding.HostPort = std::to_string(e.HostPort);
+
+            if (!hasBinding(portBindings, portBinding.HostIp, portBinding.HostPort))
```
kvega005: I believe m_mappedPorts is guaranteed not to have duplicate entries. If it did, MapPorts() would fail.

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: Agreed ! That was I was referring to this in [this comment](https://github.com/microsoft/WSL/pull/14132) 

I don't think we should return `VmPort` at all. These are implementation details that the caller shouldn't know about

src/windows/wslaservice/exe/WSLAContainer.cpp
beena352: Since docker inspect only returns a single Name (not a list), I kept the local handling here and didn’t reuse ExtractContainerName.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +554,58 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+    wslaInspect.Name = dockerInspect.Name;
+
+    // Remove leading '/' from Docker container names.
+    if (!wslaInspect.Name.empty() && wslaInspect.Name[0] == '/')
+    {
+        wslaInspect.Name = wslaInspect.Name.substr(1);
+    }
+
+    wslaInspect.Created = dockerInspect.Created;
+    wslaInspect.Image = !dockerInspect.Config.Image.empty() ? dockerInspect.Config.Image : dockerInspect.Image;
```
OneBlue: Is it possible for `dockerInspect.Config.Image` to be empty here ?

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +554,58 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+
+    // Map container state.
+    wslaInspect.State.Status = dockerInspect.State.Status;
+    wslaInspect.State.Running = dockerInspect.State.Running;
+    wslaInspect.State.ExitCode = dockerInspect.State.ExitCode;
+    wslaInspect.State.StartedAt = dockerInspect.State.StartedAt;
+    wslaInspect.State.FinishedAt = dockerInspect.State.FinishedAt;
+
+    wslaInspect.NetworkMode = dockerInspect.HostConfig.NetworkMode;
+    wslaInspect.HostConfig.NetworkMode = dockerInspect.HostConfig.NetworkMode;
```
OneBlue, line 695: I'd recommend only exposing one of those two fields

src/windows/wslaservice/exe/WSLAContainer.h
```diff
@@ -102,9 +103,12 @@ class WSLAContainerImpl
         DockerHTTPClient& DockerClient);
 
 private:
+    using DockerInspectContainer = wsl::windows::common::docker_schema::InspectContainer;
+    using WslaInspectContainer = wsl::windows::common::wsla_schema::InspectContainer;
```
OneBlue: I'd recommend not having a `using` here since this type is only used once in this file, moving the `using` to the .cpp file would be a bit cleaner imo

test/windows/WSLATests.cpp
```diff
@@ -1800,6 +1785,67 @@ class WSLATests
...
+            VERIFY_IS_FALSE(details.Id.empty());
+            VERIFY_IS_FALSE(details.Name.empty());
+            VERIFY_IS_FALSE(details.Image.empty());
+
+            VERIFY_IS_FALSE(details.State.Running);
+            VERIFY_ARE_EQUAL(details.State.ExitCode, 0);
+
+            VERIFY_SUCCEEDED(container.Get().Delete());
+        }
+    }
```
OneBlue, line 2143: We should add test coverage for volumes as well. Creating a container with both read-only and read-write volumes would be a good way to check that we handle both correctly

src/windows/wslaservice/exe/WSLAContainer.cpp
beena352: Yes, in Docker inspect output, Config.Image can be empty in some cases, so this falls back to dockerInspect.Image.

OneBlue: Weird, do we know what can cause that ?

src/windows/wslaservice/exe/WSLAContainer.cpp
beena352: Agreed, I’ll keep NetworkMode under HostConfig and remove the root-level field to avoid duplication.

src/windows/wslaservice/exe/WSLAContainer.h
beena352: Sure, we'll move it to .cpp file.

src/windows/wslaservice/exe/WSLAContainer.cpp
beena352: Yeah, I don’t have a specific case in mind. I’ll switch to always using dockerInspect.Image.

test/windows/WSLATests.cpp
beena352: Agreed, I’ll add volume coverage for both cases.

src/windows/inc/wsla_schema.h
```diff
@@ -0,0 +1,72 @@
...
+
+#pragma once
+
+#include "JsonUtils.h"
+
+namespace wsl::windows::common::wsla_schema {
+
+struct InspectPortBinding
+{
+    // HostIp is not currently populated by WSLAContainer.cpp and will remain empty.
```
OneBlue: For now we always default to 127.0.0.1, we could default initialize HostIp to that

src/windows/inc/docker_schema.h
```diff
@@ -102,13 +102,54 @@ struct CreateContainer
...
+    std::string Type;
+    std::string Source;
+    std::string Destination;
+    bool RW{};
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(InspectMount, Type, Source, Destination, RW);
+};
+
+struct NetworkSettings
+{
```
kvega005: I don't think we are using this anymore since we are getting the port info from `m_mappedPorts` in WSLAContainer.cpp

src/windows/inc/wsla_schema.h
```diff
@@ -0,0 +1,72 @@
...
+
+--*/
+
+#pragma once
+
+#include "JsonUtils.h"
+
+namespace wsl::windows::common::wsla_schema {
+
+struct InspectPortBinding
```
kvega005, line 21: I don't think we need this anymore since we are no longer using the port info from docker.

src/windows/inc/wsla_schema.h
```diff
@@ -0,0 +1,72 @@
...
+    // HostIp is not currently populated by WSLAContainer.cpp and will remain empty.
+    std::string HostIp;
+    std::string HostPort;
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(InspectPortBinding, HostIp, HostPort);
+};
+
+struct InspectMount
+{
+    std::string Type;
```
kvega005, line 33: We do not need this either. We are using the mount info already cached in WSLAContainer

test/windows/WSLATests.cpp
```diff
@@ -1800,6 +1785,104 @@ class WSLATests
...
+
+            VERIFY_ARE_EQUAL(details.HostConfig.NetworkMode, "host");
+            VERIFY_IS_TRUE(details.State.Running);
+            VERIFY_IS_FALSE(details.State.Status.empty());
+
+            // Verify port mappings are populated.
+            VERIFY_IS_FALSE(details.Ports.empty());
+
+            auto& p8000 = details.Ports.at("8000/tcp");
+            VERIFY_IS_TRUE(p8000.size() >= 1);
```
kvega005: nit: this should be exactly equal to 1 since we only added 1 port.

test/windows/WSLATests.cpp
```diff
@@ -1800,6 +1785,104 @@ class WSLATests
...
+            auto cleanup = wil::scope_exit([&]() {
+                std::error_code ec;
+                std::filesystem::remove_all(testFolder, ec);
+                std::filesystem::remove_all(testFolderReadOnly, ec);
+            });
+
+            WSLAContainerLauncher launcher(
+                "debian:latest", "test-container-inspect", {"sleep", "99999"}, {}, WSLA_CONTAINER_NETWORK_TYPE::WSLA_CONTAINER_NETWORK_HOST);
+
+            launcher.AddPort(1234, 8000, AF_INET);
```
kvega005, line 2083: Might be a good idea to test the scenario where we bind multiple host ports to the same container port.

test/windows/WSLATests.cpp
```diff
@@ -1800,6 +1785,104 @@ class WSLATests
...
+                {
+                    VERIFY_IS_FALSE(mount.ReadWrite);
+                    foundReadOnly = true;
+                }
+            }
+            VERIFY_IS_TRUE(foundReadWrite);
+            VERIFY_IS_TRUE(foundReadOnly);
+
+            VERIFY_SUCCEEDED(container.Get().Stop(WSLASignalSIGKILL, 0));
+            VERIFY_SUCCEEDED(container.Get().Delete());
```
kvega005, line 2111: Probably a good idea to define a helper similar to ExpectMount, but that just compares details.Mounts with some expected vector.

This would help us verify that the output has exactly the mounts we expect and nothing more.

test/windows/WSLATests.cpp
```diff
@@ -1800,6 +1785,104 @@ class WSLATests
...
+            auto container = launcher.Launch(*m_defaultSession);
+
+            auto details = container.Inspect();
+
+            VERIFY_IS_FALSE(details.Id.empty());
+            VERIFY_IS_FALSE(details.Name.empty());
+            VERIFY_IS_FALSE(details.Image.empty());
+
+            VERIFY_ARE_EQUAL(details.HostConfig.NetworkMode, "host");
+            VERIFY_IS_TRUE(details.State.Running);
```
kvega005, line 2100: I think the way this test is written, if there are duplicate entries for ports or mounts the test will still pass. We probably do not want that. You should include an additional guard that verifies the outputs are the exact size we expect and doesn't have any duplicate entries.

src/windows/inc/docker_schema.h
beena352: Will update it.

src/windows/inc/wsla_schema.h
beena352: we'll update it.

src/windows/inc/wsla_schema.h
beena352: we'll update it.

test/windows/WSLATests.cpp
beena352: Thanks! we'll update it.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +557,57 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+    wslaInspect.State.ExitCode = dockerInspect.State.ExitCode;
+    wslaInspect.State.StartedAt = dockerInspect.State.StartedAt;
+    wslaInspect.State.FinishedAt = dockerInspect.State.FinishedAt;
+
+    wslaInspect.HostConfig.NetworkMode = dockerInspect.HostConfig.NetworkMode;
+
+    // Map WSLA port mappings (Windows host ports only). HostIp is not surfaced.
+    for (const auto& e : m_mappedPorts)
+    {
+        auto portKey = std::format("{}/tcp", e.ContainerPort);
```
ptrivedi, line 703: @OneBlue: not sure how I missed this, but do we have protocol in port mapping? I know we have family

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +557,57 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+
+    // Map volume mounts using WSLA's host-side data.
+    wslaInspect.Mounts.reserve(m_mountedVolumes.size());
+    for (const auto& volume : m_mountedVolumes)
+    {
+        wsla_schema::InspectMount mountInfo{};
+        mountInfo.Type = "bind";
+        mountInfo.Source = wsl::shared::string::WideToMultiByte(volume.HostPath);
+        mountInfo.Destination = volume.ContainerPath;
+        mountInfo.ReadWrite = !volume.ReadOnly;
```
ptrivedi, line 720: @OneBlue @kvega005 : assuming we don't support bind propagation type? Should we add a hardcoded default for that?

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +557,57 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+
+    wslaInspect.Created = dockerInspect.Created;
+    wslaInspect.Image = dockerInspect.Image;
+
+    // Map container state.
+    wslaInspect.State.Status = dockerInspect.State.Status;
+    wslaInspect.State.Running = dockerInspect.State.Running;
+    wslaInspect.State.ExitCode = dockerInspect.State.ExitCode;
+    wslaInspect.State.StartedAt = dockerInspect.State.StartedAt;
+    wslaInspect.State.FinishedAt = dockerInspect.State.FinishedAt;
```
ptrivedi, line 693: Are we not doing pid, error? Also, what about Entrypoint?

src/windows/wslaservice/exe/WSLAContainer.cpp
beena352: Good catch. Our PortMapping only tracks address family (Family) and doesn’t include protocol, so adding "/tcp" there is misleading. I’ll update the inspect output to not imply a protocol.

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: I don't think pid has a lot of value here, since it's going to be a VM PID that the caller can't interact with

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: Not really since the underlying mount is a plan9 mount of the Windows folder.  One thing we should probably do here is to have different types for Windows mounts and VHD mounts, but we can design that when we implement VHD volumes. 

For now I think "bind" is acceptable (with a TODO)

src/windows/wslaservice/exe/WSLAContainer.cpp
OneBlue: We don't right now, but we should add support for UDP binds at some point. I'd recommend keep the hardcoded /tcp for now with a TODO so this doesn't break when we add UDP support

beena352: Quick follow-up after re-checking the CreateContainer path: we currently hard-code {port}/tcp there (with a TODO for UDP support). So while PortMapping doesn’t explicitly model protocol, WSLA effectively assumes TCP today. I’ll keep the /tcp suffix in inspect for consistency and add a comment clarifying the current TCP-only behavior.

src/windows/inc/wsla_schema.h
```diff
@@ -0,0 +1,73 @@
...
+
+namespace wsl::windows::common::wsla_schema {
+
+struct InspectPortBinding
+{
+    // WSLA always binds to localhost. Included for Docker API compatibility.
+    std::string HostIp = "127.0.0.1";
+    std::string HostPort;
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(InspectPortBinding, HostIp, HostPort);
```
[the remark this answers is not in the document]
beena352: make sense. I'll update it.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -663,6 +667,70 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+    {
+        wslaInspect.Name = wslaInspect.Name.substr(1);
+    }
+
+    wslaInspect.Created = dockerInspect.Created;
+
+    // Prefer the user-friendly image name/tag from Config.Image, falling back to the image ID.
+    if (!dockerInspect.Config.Image.empty())
+    {
+        wslaInspect.Image = dockerInspect.Config.Image;
```
OneBlue: Are there situations in which this field is not set by docker ?

test/windows/WSLATests.cpp
```diff
@@ -1974,6 +1974,168 @@ class WSLATests
         }
     }
 
+    TEST_METHOD(ContainerInspect)
+    {
+        WSL2_TEST_ONLY();
+        SKIP_TEST_ARM64();
+
+        // Helper to verify port mappings.
+        auto expectPorts = [&](const auto& actualPorts, const std::map<std::string, std::set<std::string>>& expectedPorts) {
```
OneBlue, line 2013: Given that this is coming from JSON, one way we could simplify this would be to serialize both to string and compare the string values (JSON serialization should be alphabetical)

OneBlue: (Same for mounts below)

test/windows/WSLATests.cpp
```diff
@@ -1974,6 +1974,168 @@ class WSLATests
...
+            launcher.AddPort(1236, 8001, AF_INET);
+            launcher.AddVolume(testFolder.wstring(), "/test-volume", false);
+            launcher.AddVolume(testFolderReadOnly.wstring(), "/test-volume-ro", true);
+
+            auto container = launcher.Launch(*m_defaultSession);
+            auto details = container.Inspect();
+
+            // Verify basic container metadata is present.
+            VERIFY_IS_FALSE(details.Id.empty());
+            VERIFY_IS_FALSE(details.Name.empty());
```
OneBlue: I recommend validating that the name is actually what we expect here

OneBlue: (same for image, status and other deterministic fields)

src/windows/wslaservice/exe/WSLAContainer.cpp
beena352: No, the fallback isn’t needed here.

test/windows/WSLATests.cpp
beena352: I took a look and agree that JSON string comparison could work for basic shape validation. Here we followed existing test patterns using simple std::map-based comparisons to keep checks order-independent and to validate WSLA-specific invariants for example HostIp == 127.0.0.1 that would be harder to express with raw JSON string comparison. Happy to simplify if needed.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +557,60 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+    wslaInspect.Name = dockerInspect.Name;
+
+    // Remove leading '/' from Docker container names.
+    if (!wslaInspect.Name.empty() && wslaInspect.Name[0] == '/')
+    {
+        wslaInspect.Name = wslaInspect.Name.substr(1);
+    }
+
+    wslaInspect.Created = dockerInspect.Created;
+    wslaInspect.Image = dockerInspect.Image;
```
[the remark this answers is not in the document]
kvega005: @beena352, is this a valid callout?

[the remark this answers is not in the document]
beena352: I tried this earlier, but per feedback we’re assuming dockerInspect.Config.Image is always populated for our use cases, so I switched back.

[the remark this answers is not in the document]
OneBlue: What about using `m_image` here ? We know that it's always set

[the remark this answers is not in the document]
beena352: Yes, I missed that. we can use m_image here since it’s always set. I’ll switch to that.

test/windows/WSLATests.cpp
beena352: Added.

test/windows/WSLATests.cpp
beena352: Updated.

src/windows/wslaservice/exe/WSLAContainer.cpp
beena352: Added TODO comment.

test/windows/WSLATests.cpp
beena352: Updated the code for deterministic fields.

src/windows/wslaservice/exe/WSLAContainer.cpp
beena352: Fixed.

src/windows/wslaservice/exe/WSLAContainer.cpp
beena352: Thanks! updated.

test/windows/WSLATests.cpp
beena352: Fixed.

src/windows/inc/wsla_schema.h
```diff
@@ -0,0 +1,73 @@
...
+    std::string FinishedAt;
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(InspectState, Status, Running, ExitCode, StartedAt, FinishedAt);
+};
+
+struct InspectHostConfig
+{
+    std::string NetworkMode;
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(InspectHostConfig, NetworkMode);
```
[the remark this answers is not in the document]
beena352: Updated.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +557,60 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+    wslaInspect.State.StartedAt = dockerInspect.State.StartedAt;
+    wslaInspect.State.FinishedAt = dockerInspect.State.FinishedAt;
+
+    wslaInspect.HostConfig.NetworkMode = dockerInspect.HostConfig.NetworkMode;
+
+    // Map WSLA port mappings (Windows host ports only). HostIp is not surfaced.
+    for (const auto& e : m_mappedPorts)
+    {
+        // TODO: UDP support
+        // TODO: Investigate ipv6 support.
```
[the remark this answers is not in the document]
beena352: Done.

src/windows/inc/wsla_schema.h
beena352: Yes, I’ve updated it to default initialize HostIp to 127.0.0.1

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -553,6 +557,60 @@ void WSLAContainerImpl::Exec(const WSLA_PROCESS_OPTIONS* Options, IWSLAProcess**
...
+    // Map container state.
+    wslaInspect.State.Status = dockerInspect.State.Status;
+    wslaInspect.State.Running = dockerInspect.State.Running;
+    wslaInspect.State.ExitCode = dockerInspect.State.ExitCode;
+    wslaInspect.State.StartedAt = dockerInspect.State.StartedAt;
+    wslaInspect.State.FinishedAt = dockerInspect.State.FinishedAt;
+
+    wslaInspect.HostConfig.NetworkMode = dockerInspect.HostConfig.NetworkMode;
+
+    // Map WSLA port mappings (Windows host ports only). HostIp is not surfaced.
```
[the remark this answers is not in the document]
beena352: Done.

test/windows/WSLATests.cpp
beena352: Done, added a helper to validate details.Mounts against the expected mounts.

src/windows/wslaservice/exe/WSLAContainer.cpp
beena352: Fixed.

src/windows/inc/docker_schema.h
```diff
@@ -103,13 +103,46 @@ struct CreateContainer
...
 struct InspectContainer
 {
     std::string Id;
     std::string Name;
+    std::string Created;
+    std::string Image;
+    ContainerInspectState State;
+    ContainerConfig Config;
     HostConfig HostConfig;
+    std::vector<InspectMount> Mounts;
```
OneBlue: Is this still used from the docker schema side ? If not let's remove it

OneBlue (approved):
LGTM. One minor comment

src/windows/inc/docker_schema.h
beena352: No, It's not used from docker schema side. I'll remove it. Thanks!

PR #14133, opened 2026-01-30
DNS: Simplify dns logic and switch to using 'search' instead of 'domain' which is obsolete.

benhillis:
This PR refactors HostDnsInfo to make GetDnsSettings a static method, removing unnecessary state (cached adapter addresses, mutex, UpdateNetworkInformation(), and the DnsUpdateHelper class) since the cache was being populated and immediately consumed without reuse. It also standardizes on the search directive for DNS suffixes instead of the obsolete domain directive, per resolv.conf(5).

Additionally the option to supply a header for /etc/resolv.conf was added to virtioproxy and NAT networking modes.

### DIFF
--- a/src/windows/common/VirtioNetworking.cpp
+++ b/src/windows/common/VirtioNetworking.cpp
@@ -14,11 +14,12 @@ using wsl::core::VirtioNetworking;
 static constexpr auto c_loopbackDeviceName = TEXT(LX_INIT_LOOPBACK_DEVICE_NAME);
 
 VirtioNetworking::VirtioNetworking(
-    GnsChannel&& gnsChannel, bool enableLocalhostRelay, std::shared_ptr<GuestDeviceManager> guestDeviceManager, wil::shared_handle userToken) :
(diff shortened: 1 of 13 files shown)


src/windows/common/VirtioNetworking.cpp
```diff
@@ -14,11 +14,12 @@ using wsl::core::VirtioNetworking;
 static constexpr auto c_loopbackDeviceName = TEXT(LX_INIT_LOOPBACK_DEVICE_NAME);
 
 VirtioNetworking::VirtioNetworking(
-    GnsChannel&& gnsChannel, bool enableLocalhostRelay, std::shared_ptr<GuestDeviceManager> guestDeviceManager, wil::shared_handle userToken) :
+    GnsChannel&& gnsChannel, bool enableLocalhostRelay, LPCWSTR dnsOptions, std::shared_ptr<GuestDeviceManager> guestDeviceManager, wil::shared_handle userToken) :
```
benhillis, line 17: @OneBlue - I'm open to other options, but basically I just don't want the WSL header in /etc/resolv.conf when we are running WSLA.

OneBlue (approved):

PR #14137, opened 2026-01-31
Correctly handle empty arguments when creating processes

OneBlue:

This change updates the init protocol to correctly handle empty strings in array. 

This change also test coverage for empty arguments at the VM, container and exec level

## PR Checklist

### DIFF
--- a/src/shared/inc/message.h
+++ b/src/shared/inc/message.h
@@ -147,28 +147,34 @@ public:
         WriteString(Index, wsl::shared::string::WideToMultiByte(String));
     }
 
-    // TODO: This design doesn't allow empty strings
+    // Write an array of strings.
+    // Each field is prefixed with its size as int32_t, and the array ends with a -1 terminator.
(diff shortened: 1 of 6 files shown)


src/shared/inc/message.h
```diff
@@ -147,28 +147,34 @@ class MessageWriter
...
-            totalSize += size + 1;
+            totalSize += strlen(String[i]) + sizeof(int32_t);
         }
 
         auto span = InsertBuffer(Index, totalSize);
-
         auto it = span.begin();
+
+        auto insertSize = [&](int32_t size) {
+            it = std::copy(reinterpret_cast<const std::byte*>(&size), reinterpret_cast<const std::byte*>(&size) + sizeof(size), it);
```
benhillis, line 164: wouldn't using gsl::copy be better here?

OneBlue: I don't think so, because gsl::copy doesn't return the end iterator, so we have to manually increase it after it returns

benhillis (approved):

PR #14139, opened 2026-01-31
Redesign Accept() logic to differenciate between cancellation and errors

OneBlue:
## PR Checklist

### DIFF
--- a/src/windows/common/socket.cpp
+++ b/src/windows/common/socket.cpp
@@ -17,20 +17,25 @@ Abstract:
 #include "socket.hpp"
 #pragma hdrstop
 
-void wsl::windows::common::socket::Accept(
-    _In_ SOCKET ListenSocket, _In_ SOCKET Socket, _In_ int Timeout, _In_opt_ HANDLE ExitHandle, _In_ const std::source_location& Location)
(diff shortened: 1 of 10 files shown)


src/windows/common/socket.cpp
```diff
@@ -17,20 +17,25 @@ Module Name:
 #include "socket.hpp"
 #pragma hdrstop
 
-void wsl::windows::common::socket::Accept(
-    _In_ SOCKET ListenSocket, _In_ SOCKET Socket, _In_ int Timeout, _In_opt_ HANDLE ExitHandle, _In_ const std::source_location& Location)
+bool wsl::windows::common::socket::CancellableAccept(
```
OneBlue, line 20: @benhillis: Let me know what you think of this. 


The idea is to allow the caller to differentiate between: "Something went wrong", and "the Accept() was cancelled".

This will be good because we have a tendency to log a bunch of errors when the VM terminates because all the pending accept() are cancelled. 

If you're OK with this approach, I'll bring the change to `master` first

OneBlue:
Done in: https://github.com/microsoft/WSL/pull/14161

PR #14147, opened 2026-02-02
WSLA: Update GetExitEvent, ImportImage, LoadImage, and Logs to use system handle marshalling.

benhillis:
This change updates the WSLA interface to use system handle marshalling when possible. I did not update any of the methods where the handle type can vary (GetStdHandle, Attach, WarningsPipe), but we can consider that in a future change.

### DIFF
--- a/src/windows/wslaservice/exe/WSLAContainer.cpp
+++ b/src/windows/wslaservice/exe/WSLAContainer.cpp
@@ -875,7 +875,7 @@ void WSLAContainerImpl::Inspect(LPSTR* Output)
     }
 }
 
-void WSLAContainerImpl::Logs(WSLALogsFlags Flags, ULONG* Stdout, ULONG* Stderr, ULONGLONG Since, ULONGLONG Until, ULONGLONG Tail)
+void WSLAContainerImpl::Logs(WSLALogsFlags Flags, HANDLE* Stdout, HANDLE* Stderr, ULONGLONG Since, ULONGLONG Until, ULONGLONG Tail)
 {
     std::lock_guard lock(m_lock);
 
@@ -897,7 +897,7 @@ void WSLAContainerImpl::Logs(WSLALogsFlags Flags, ULONG* Stdout, ULONG* Stderr, 
         auto handle = std::make_unique<RelayHandle<HTTPChunkBasedReadHandle>>(std::move(socket), std::move(ttyWrite));
         m_ioRelay.AddHandle(std::move(handle));
 
-        *Stdout = HandleToULong(common::wslutil::DuplicateHandleToCallingProcess(ttyRead.get()));
+        *Stdout = ttyRead.release();
     }
     else
     {
@@ -910,8 +910,8 @@ void WSLAContainerImpl::Logs(WSLALogsFlags Flags, ULONG* Stdout, ULONG* Stderr, 
 
         m_ioRelay.AddHandle(std::move(handle));
 
-        *Stdout = HandleToULong(common::wslutil::DuplicateHandleToCallingProcess(stdoutRead.get()));
-        *Stderr = HandleToULong(common::wslutil::DuplicateHandleToCallingProcess(stderrRead.get()));
+        *Stdout = stdoutRead.release();
+        *Stderr = stderrRead.release();
     }
 }
 
@@ -1016,12 +1016,10 @@ try
 }
 CATCH_RETURN();
 
-HRESULT WSLAContainer::Logs(WSLALogsFlags Flags, ULONG* Stdout, ULONG* Stderr, ULONGLONG Since, ULONGLONG Until, ULONGLONG Tail)
+HRESULT WSLAContainer::Logs(WSLALogsFlags Flags, HANDLE* Stdout, HANDLE* Stderr, ULONGLONG Since, ULONGLONG Until, ULONGLONG Tail)
 {
-    RETURN_HR_IF(E_POINTER, Stdout == nullptr || Stderr == nullptr);
-
-    *Stdout = 0;
-    *Stderr = 0;
+    *Stdout = nullptr;
+    *Stderr = nullptr;
 
     return CallImpl(&WSLAContainerImpl::Logs, Flags, Stdout, Stderr, Since, Until, Tail);
 }
--- a/src/windows/wslaservice/exe/WSLASession.cpp
+++ b/src/windows/wslaservice/exe/WSLASession.cpp
@@ -450,7 +450,7 @@ void WSLASession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request,
     }
 }
 
-HRESULT WSLASession::ExportContainer(ULONG OutHandle, LPCSTR ContainerID, IProgressCallback* ProgressCallback, WSLA_ERROR_INFO* Error)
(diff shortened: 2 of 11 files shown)


src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -322,7 +322,7 @@ interface IWSLAContainer : IUnknown
     HRESULT GetInitProcess([out] IWSLAProcess** Process);
     HRESULT Exec([in] const struct WSLA_PROCESS_OPTIONS* Options, [out] IWSLAProcess** Process, [out] int* Errno);
     HRESULT Inspect([out] LPSTR* Output);
-    HRESULT Logs([in] WSLALogsFlags Flags, [out] ULONG* Stdout, [out] ULONG* Stderr, [in] ULONGLONG Since, [in] ULONGLONG Until, [in] ULONGLONG Tail);
+    HRESULT Logs([in] WSLALogsFlags Flags, [out, system_handle(sh_pipe)] HANDLE* Stdout, [out, system_handle(sh_pipe)] HANDLE* Stderr, [in] ULONGLONG Since, [in] ULONGLONG Until, [in] ULONGLONG Tail);
```
OneBlue, line 387: Does this work for tty processes as well ? For TTY processes we return an hvsocket

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -362,8 +362,8 @@ interface IWSLASession : IUnknown
 
     // Image management.
     HRESULT PullImage([in] LPCSTR ImageUri, [in, unique] const struct WSLA_REGISTRY_AUTHENTICATION_INFORMATION* RegistryAuthenticationInformation, [in, unique] IProgressCallback* ProgressCallback, [in, out, unique] WSLA_ERROR_INFO* ErrorInfo);
-    HRESULT LoadImage([in] ULONG ImageHandle, [in, unique] IProgressCallback* ProgressCallback, [in] ULONGLONG ContentLength);
-    HRESULT ImportImage([in] ULONG ImageHandle, [in] LPCSTR ImageName, [in, unique] IProgressCallback* ProgressCallback, [in] ULONGLONG ContentLength);
+    HRESULT LoadImage([in, system_handle(sh_file)] HANDLE ImageHandle, [in, unique] IProgressCallback* ProgressCallback, [in] ULONGLONG ContentLength);
```
OneBlue, line 435: FYI I was planning to change this into a returned handle

OneBlue (commented):
I think using handle marshalling for exit events is a good idea, but I'd have the opinion to leave the other ones as ULONG*, so we don't prevent ourselves from changing their types in the future (also some methods can return different types based on the type of process)

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -797,7 +797,7 @@ void WSLAContainerImpl::Logs(WSLALogsFlags Flags, ULONG* Stdout, ULONG* Stderr,
         auto handle = std::make_unique<RelayHandle<HTTPChunkBasedReadHandle>>(std::move(socket), std::move(ttyWrite));
         m_logsRelay.AddHandle(std::move(handle));
 
-        *Stdout = HandleToULong(common::wslutil::DuplicateHandleToCallingProcess(ttyRead.get()));
+        *Stdout = ttyRead.release();
```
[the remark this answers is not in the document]
benhillis: this is correct.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -810,8 +810,8 @@ void WSLAContainerImpl::Logs(WSLALogsFlags Flags, ULONG* Stdout, ULONG* Stderr,
 
         m_logsRelay.AddHandle(std::move(handle));
 
-        *Stdout = HandleToULong(common::wslutil::DuplicateHandleToCallingProcess(stdoutRead.get()));
-        *Stderr = HandleToULong(common::wslutil::DuplicateHandleToCallingProcess(stderrRead.get()));
+        *Stdout = stdoutRead.release();
+        *Stderr = stderrRead.release();
```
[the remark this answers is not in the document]
benhillis: this is correct.

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -493,7 +491,7 @@ void WSLASession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request,
     auto onCompleted = [&]() { io.Cancel(); };
 
     io.AddHandle(std::make_unique<relay::RelayHandle<relay::ReadHandle>>(
-        common::relay::HandleWrapper{std::move(imageFileHandle)}, common::relay::HandleWrapper{Request.stream.native_handle()}));
+        common::relay::HandleWrapper{InputHandle}, common::relay::HandleWrapper{Request.stream.native_handle()}));
```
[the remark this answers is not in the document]
benhillis: Is this true? let me double check the HandleWrapper doesn't close the handle.

src/windows/wslaservice/inc/wslaservice.idl
benhillis: That's also fine, let's just use handle marshalling.

src/windows/wslaservice/exe/WSLASession.cpp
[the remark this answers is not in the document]
benhillis: Yeah this isn't true, when passing in a HANDLE HandleWrapper does not take ownership.

src/windows/wslaservice/inc/wslaservice.idl
benhillis: stdout and stderr are always pipes, see WSLAContainerImpl::Logs

OneBlue: Ah that's true, I forgot that we had to relay the chunks

benhillis:
> I think using handle marshalling for exit events is a good idea, but I'd have the opinion to leave the other ones as ULONG*, so we don't prevent ourselves from changing their types in the future (also some methods can return different types based on the type of process)

I disagree, I think we should use handle marshalling everywhere possible to avoid strange behavior when the callers specify unexpected handle types. If we decide we need more flexibility we can add it.

OneBlue:
> I disagree, I think we should use handle marshalling everywhere possible to avoid strange behavior when the callers specify unexpected handle types. If we decide we need more flexibility we can add it.

I think this would be OK as long as it doesn't lock us down in the future. If let's say we ship with a `sh_pipe` in the .IDL, and then in the future we decide that we want to return an `hvsocket`, will the client COM side accept the new handle type ? If so, I think this is OK

benhillis:
> > I disagree, I think we should use handle marshalling everywhere possible to avoid strange behavior when the callers specify unexpected handle types. If we decide we need more flexibility we can add it.
> 
> I think this would be OK as long as it doesn't lock us down in the future. If let's say we ship with a `sh_pipe` in the .IDL, and then in the future we decide that we want to return an `hvsocket`, will the client COM side accept the new handle type ? If so, I think this is OK

That's a good question. I'd argue that either way that would be an API breaking change.

PR #14160, opened 2026-02-04
Enable detection cmake script for Visual Studio 2026 and pre-release versions of Visual Studio

andysterland:
Updates to enable the creation of the .slnx file with Visual Studio 2026 and preview versions of Visual Studio (either 2022 or 2026).

## PR Checklist

- [x] **Closes:** Link to issue #14159
- [ ] **Communication:** Not discussed.
- [ ] **Tests:** None added.
- [x] **Localization:** No user facing strings changed
- [x] **Dev docs:** The getting started doc doesn't explicitly mention 2026 or 2022 so no need to update.

Update the args for calling `vswhere` to:
- Changed the, exclusive, max version to 19.0 from 18.0 so VS2026 installs will be searched for
- Added the arg `--prerelease` so that prerelease versions of Visual Studio are included in the search for VS installs, so preview and insiders installs will be found.
- Added the arg `--latest` so that the latest version of VS is returned assuming the developer wants to use the latest version of VS  this is optional but it seemed better to be explicit about which VS version will be selected if multiple are available.  

## Validation Steps Performed
Ran `cmake .`  in a fresh repo with VS2026 installed.


CMakeLists.txt
```diff
@@ -263,7 +263,7 @@ endif()
 # Determine the Visual Studio installation directory which contains LLVM tools
 # N.B. The version is set to VS2022 to ensure local runs match pipeline behavior
 execute_process(
-    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -version "[17.0,18.0)" -products * -property installationPath
+    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -version "[17.0,19.0)" -products * -property installationPath -latest -prerelease
```
OneBlue: I don't think we want `-latest` here since we want to use the VS 2022 version of llvm-tools to match what currently runs in the pipeline

benhillis: Agreed, the different versions of clang format cause discrepancies.

andysterland: Removed

OneBlue (commented):
Change LGTM. It just needs to be merged with latest `master`

benhillis:
> Change LGTM. It just needs to be merged with latest `master`

resolved.

OneBlue (approved):

PR #14163, opened 2026-02-05
Users can set and retrieve labels for containers

kvega005:
This PR adds the ability to set custom labels when creating a WSLA container and retrieve them later via the GetLabels API.

### Changes
**API Changes:**
- Added Labels and LabelsCount fields to WSLA_CONTAINER_OPTIONS to allow specifying key-value labels at container creation time
- Added GetLabels method to IWSLAContainer interface to retrieve labels for an existing container

**Implementation:**
- WSLAContainerImpl::Create: Processes user-provided labels and passes them to the Docker container creation request, while also storing them in the container instance for later retrieval
- WSLAContainerImpl::Open: Extracts and preserves user labels (excluding internal WSLA metadata) when recovering containers from storage
- WSLAContainerImpl::GetLabels: Returns the stored labels as a COM-friendly array

**Testing:**
Added ContainerLabels test case that:
- Creates a container with two custom labels (key1=value1, key2=value2)
- Verifies labels can be retrieved after container creation
- Verifies labels persist after container restart/recovery from storage
- Tests error handling for null keys and null values

## PR Checklist

### DIFF
--- a/src/windows/wslaservice/exe/WSLAContainer.cpp
+++ b/src/windows/wslaservice/exe/WSLAContainer.cpp
@@ -270,6 +270,7 @@ WSLAContainerImpl::WSLAContainerImpl(
     std::string&& Image,
     std::vector<WSLAVolumeMount>&& volumes,
     std::vector<WSLAPortMapping>&& ports,
+    std::map<std::string, std::string>&& labels,
     std::function<void(const WSLAContainerImpl*)>&& onDeleted,
     ContainerEventTracker& EventTracker,
     DockerHTTPClient& DockerClient,
@@ -283,6 +284,7 @@ WSLAContainerImpl::WSLAContainerImpl(
     m_id(std::move(Id)),
     m_mountedVolumes(std::move(volumes)),
     m_mappedPorts(std::move(ports)),
+    m_labels(std::move(labels)),
     m_comWrapper(wil::MakeOrThrow<WSLAContainer>(this, std::move(onDeleted))),
     m_dockerClient(DockerClient),
     m_eventTracker(EventTracker),
@@ -755,6 +757,15 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
         portEntry.emplace_back(common::docker_schema::PortMapping{.HostIp = "127.0.0.1", .HostPort = std::to_string(e.VmPort)});
     }
 
+    std::map<std::string, std::string> labels;
+    for (ULONG i = 0; i < containerOptions.LabelsCount; i++)
+    {
+        const auto& label = containerOptions.Labels[i];
+        THROW_HR_IF_MSG(E_INVALIDARG, label.Key == nullptr || label.Value == nullptr, "Label key and value cannot be null");
+
+        labels[label.Key] = label.Value;
+    }
+
     // Build WSLA metadata to store in a label for recovery on Open().
     WSLAContainerMetadataV1 metadata;
     metadata.InitProcessFlags = containerOptions.InitProcessOptions.Flags;
@@ -762,6 +773,7 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
     metadata.Ports = ports;
 
     request.Labels[WSLAContainerMetadataLabel] = SerializeContainerMetadata(metadata);
+    request.Labels.insert(labels.begin(), labels.end());
 
     // Send the request to docker.
     auto result =
@@ -774,6 +786,7 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
         std::move(std::string(containerOptions.Image)),
         std::move(volumes),
         std::move(ports),
+        std::move(labels),
         std::move(OnDeleted),
         EventTracker,
         DockerClient,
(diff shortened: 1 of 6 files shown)


src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -755,13 +757,23 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
         portEntry.emplace_back(common::docker_schema::PortMapping{.HostIp = "127.0.0.1", .HostPort = std::to_string(e.VmPort)});
     }
 
+    std::map<std::string, std::string> labels;
+    for (ULONG i = 0; i < containerOptions.LabelsCount; i++)
+    {
+        const auto& label = containerOptions.Labels[i];
+        THROW_HR_IF_MSG(E_INVALIDARG, label.Key == nullptr || label.Value == nullptr, "Label key and value cannot be null");
```
OneBlue: nit: I recommend logging the index to make debugging potential failures here easier

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -755,13 +757,23 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
         portEntry.emplace_back(common::docker_schema::PortMapping{.HostIp = "127.0.0.1", .HostPort = std::to_string(e.VmPort)});
     }
 
+    std::map<std::string, std::string> labels;
+    for (ULONG i = 0; i < containerOptions.LabelsCount; i++)
+    {
+        const auto& label = containerOptions.Labels[i];
```
OneBlue, line 763: I recommend adding a checks for: 

- The caller trying to set: WSLAContainerMetadataLabel
- The caller having two labels with the same key (since that second would just override the first)

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -1036,4 +1053,36 @@ try
...
+    }
+
+    auto labelsArray = wil::make_unique_cotaskmem<WSLA_LABEL_INFORMATION[]>(m_labels.size());
+    ULONG i = 0;
+    for (const auto& [key, value] : m_labels)
+    {
+        labelsArray[i].Key = wil::make_unique_ansistring<wil::unique_cotaskmem_ansistring>(key.c_str()).release();
+        labelsArray[i].Value = wil::make_unique_ansistring<wil::unique_cotaskmem_ansistring>(value.c_str()).release();
+        i++;
+    }
```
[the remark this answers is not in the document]
OneBlue: Agreed with this. I recommend using a wil::scope_exit_log to cleanup if an exception is thrown while we iterate on the labels

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -1036,4 +1053,36 @@ try
...
+
+    *Labels = labelsArray.release();
+}
+
+HRESULT WSLAContainer::GetLabels(WSLA_LABEL_INFORMATION** Labels, ULONG* Count)
+try
+{
+    RETURN_HR_IF(E_POINTER, Labels == nullptr || Count == nullptr);
```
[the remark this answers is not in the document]
OneBlue: Agreed with this

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -171,6 +171,18 @@ struct WSLA_PORT_MAPPING
     int Family;
 };
 
+struct WSLA_LABEL
+{
+    [string] LPCSTR Key;
+    [string] LPCSTR Value;
+};
+
+struct WSLA_LABEL_INFORMATION
```
OneBlue, line 180: Could we reuse the same structure for both calls ? (I think it would be OK to use const strings both as input and output here)

test/windows/WSLATests.cpp
```diff
@@ -3024,6 +3024,83 @@ class WSLATests
...
+            label.Key = "key";
+            label.Value = nullptr;
+
+            options.Labels = &label;
+            options.LabelsCount = 1;
+
+            wil::com_ptr<IWSLAContainer> container;
+            auto hr = session->CreateContainer(&options, &container, nullptr);
+            VERIFY_ARE_EQUAL(hr, E_INVALIDARG);
+        }
```
OneBlue, line 3114: I recommend adding test coverage for: 

- Caller trying to set the same label key twice
- Caller trying to use our metadata key (both when creating a container, and when retrieving labels) 
- Trying to set a label with a large value (since label values don't have size limits from what I can see)

OneBlue (commented):
Change looks great, couple minor comments

src/windows/wslaservice/inc/wslaservice.idl
kvega005: I spent some time looking into this problem because it feels really clunky to use two different structs.

We can do this, but the problem with using const strings for both is that the caller cannot call CoTaskMemFree on a const string. COM expects the callee to allocate and write to the output buffer (and the caller to free it).

We could use non-const strings for both but using non‑const strings (LPSTR) for input would weaken const‑correctness and make ownership unclear. It blurs the contract: who owns the memory and who can modify it.

The alternative is to use `BSTR`. I did not do this initially because nothing else in the IDL uses BSTR, so I was not sure if there is a specific reason why we are not using it.

OneBlue: Ah I see that makes sense. BSTRs are wide strings so don't work well for linux scenarios since we use utf8 there. 

In that case I having two structures is the simplest solution

OneBlue (approved):

PR #14171, opened 2026-02-06
Initial Support for WSLC Commands

AmelBawa-msft:

_⚠️ The code changes in this PR are still in a construction phase, but the current quality is sufficient to unblock the team while new functionalities are being implemented._

### 🙌 Demo
```
[Window 1]
wslc container --help
wslc container run --help
wslc container run -it --name demo ubuntu:latest /bin/bash

[Window 2]
wslc container exec demo ls -lt
wslc container list --help
wslc container list
wslc container list -aq
wslc container inspect demo
wslc container kill demo

[Window 1]
# Verify container is not attached anymore
wslc container delete demo
wslc container list -a
```

### 🦞  Full implementation
Run `--help` for each command and subcommand for the full list of supported options.

## PR Checklist


src/windows/wslc/SessionModel.cpp
```diff
@@ -0,0 +1,18 @@
...
+SessionOptions SessionOptions::Default()
+{
+    // TODO: Have a configuration file for those.
+    auto dataFolder = std::filesystem::path(wsl::windows::common::filesystem::GetLocalAppDataPath(nullptr)) / "wsla";
+    SessionOptions options{};
+    options.DisplayName = L"wsla-cli";
+    options.CpuCount = 4;
+    options.MemoryMb = 2048;
+    options.BootTimeoutMs = 30 * 1000;
+    options.StoragePath = dataFolder.c_str();
```
JohnMcPMS: Dangling pointer to local; a suggestion on a pattern to fix it: https://github.com/microsoft/WSL/compare/user/amelbawa/init...JohnMcPMS:amir-suggest?expand=1

src/windows/wslc/ICommand.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+#include <CommandLine.h>
+#include "SessionService.h"
+
+#define CMD_IF_HELP_PRINT_HELP() if (m_help) { PrintHelp(); return 0; }
+#define CMD_ARG_REQUIRED(arg, msg) if (arg.empty()) { wslutil::PrintMessage(msg, stderr); PrintHelp(); return E_INVALIDARG; }
+#define CMD_ARG_ARRAY_REQUIRED(argArray, msg) if (argArray.empty()) { wslutil::PrintMessage(msg, stderr); PrintHelp(); return E_INVALIDARG; }
```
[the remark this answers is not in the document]
AmelBawa-msft: Unless necessary, I will wait until @dkbennett PR which implements superior handling for arguments overall.

OneBlue (commented):
FYI this branch needs to be rebased (we've gotten rid of WSLAErrorDetails, the error messages are now captured through the ExecutionContext). 

[Here's an example of how to retrieve the error message](https://github.com/microsoft/WSL/blob/901e9d2e38b707ed277a28e78a1a67c7ef4c9c05/src/windows/wsladiag/wsladiag.cpp#L772C8-L776C125 )

src/windows/wslc/ImageCommand.cpp
```diff
@@ -0,0 +1,89 @@
...
+    CMD_IF_HELP_PRINT_HELP();
+    auto session = m_sessionService.CreateSession();
+    wslc::services::ImageService imageService;
+    auto images = imageService.List(session);
+    if (m_format == "json")
+    {
+        for (const models::ImageInformation& image : images)
+        {
+            wprintf(L"%hs", wsl::shared::ToJson(image).c_str());
+        }
```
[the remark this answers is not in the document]
AmelBawa-msft: By design

src/windows/wslc/ConsoleService.cpp
```diff
@@ -0,0 +1,126 @@
+/*++
```
benhillis, line 1: Is this all copy / pasted from somewhere else?

benhillis: if this is used / usable in multiple places we should put it in common.

src/windows/wslc/ContainerCommand.cpp
```diff
@@ -0,0 +1,332 @@
...
+    }
+}
+
+int ContainerRunCommand::ExecuteInternal(std::wstring_view commandLine, int parserOffset)
+{
+    CMD_IF_HELP_PRINT_HELP();
+    CMD_ARG_REQUIRED(m_image, L"Error: image name is required.");
+    auto session = m_sessionService.CreateSession();
+    m_options.Arguments = Arguments();
+    m_options.Name = GetContainerName(m_options.Name);
```
OneBlue: I don't recommend generating a GUID if no name is passed. If the caller doesn't pass a name, let's just not give one to the service (the service will generate a name)

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+static void SetContainerTTYOptions(WSLA_PROCESS_OPTIONS& options)
+{
+    if (WI_IsFlagSet(options.Flags, WSLAProcessFlagsTty))
+    {
+        HANDLE Stdout = GetStdHandle(STD_OUTPUT_HANDLE);
+        HANDLE Stdin = GetStdHandle(STD_INPUT_HANDLE);
+        CONSOLE_SCREEN_BUFFER_INFOEX info{};
+        info.cbSize = sizeof(info);
+        THROW_IF_WIN32_BOOL_FALSE(::GetConsoleScreenBufferInfoEx(Stdout, &info));
```
OneBlue: This can fail if stdout is not a tty. We might want to add special handling to default to stdin, and show an error if neither stdin not stdout is a tty

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,222 @@
...
+        options.TtyColumns = info.srWindow.Right - info.srWindow.Left + 1;
+        options.TtyRows = info.srWindow.Bottom - info.srWindow.Top + 1;
+    }
+}
+
+static void SetContainerArguments(WSLA_PROCESS_OPTIONS& options, const std::vector<std::string>& args, std::vector<const char*>& argsStorage)
+{
+    argsStorage.clear();
```
OneBlue: nit: This could be replaced by wsl::shared::string::StringPointersFromArray()

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,222 @@
...
+        argsStorage.push_back(arg.c_str());
+    }
+    options.CommandLine = {.Values = argsStorage.data(), .Count = static_cast<ULONG>(argsStorage.size())};
+}
+
+static void CreateInternal(
+    Session& session,
+    IWSLAContainer** container,
+    WSLA_CONTAINER_OPTIONS& containerOptions,
+    std::string image,
```
OneBlue: nit: `const std::string&` to avoid the copy

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,222 @@
...
+        imageService.Pull(session, image, callback);
+        result = session.Get()->CreateContainer(&containerOptions, container);
+    }
+
+    THROW_IF_FAILED(result);
+}
+
+static void StopInternal(IWSLAContainer& container, int signal, ULONG timeout = StopContainerOptions::DefaultTimeout)
+{
+    THROW_IF_FAILED(container.Stop(static_cast<WSLASignal>(signal), timeout)); // TODO: Error message
```
OneBlue, line 97: The frontend shouldn't set the default for either the signal or the timeout (those are handled by the service). 

Use `WSLASignalNone` for the default signal and -1 for the default stop timeout

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,222 @@
...
+CreateContainerResult ContainerService::Create(Session& session, std::string image, ContainerCreateOptions runOptions, IProgressCallback* callback)
+{
+    wil::com_ptr<IWSLAContainer> container;
+    WSLA_CONTAINER_OPTIONS containerOptions{};
+    CreateInternal(session, &container, containerOptions, image, runOptions, callback);
+
+    wil::unique_cotaskmem_ansistring output;
+    THROW_IF_FAILED(container->Inspect(&output));
+
+    auto inspect = wsl::shared::FromJson<InspectContainer>(output.get());
```
OneBlue: This can be simplified by using WSLAContainer::GetId() so we can avoid the JSON parsing here

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,222 @@
...
+            ClientRunningWSLAProcess(std::move(process), containerOptions.InitProcessOptions.Flags));
+    }
+
+    WSLAContainerId containerId{};
+    THROW_IF_FAILED(container->GetId(containerId));
+    PrintMessage(L"%hs\n", stdout, containerId);
+    return 0;
+}
+
+CreateContainerResult ContainerService::Create(Session& session, std::string image, ContainerCreateOptions runOptions, IProgressCallback* callback)
```
OneBlue: nit: take string by const references (applies to other places in the PR)

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,222 @@
...
+    StopInternal(*container, signal);
+}
+
+void ContainerService::Delete(Session& session, std::string id, bool force)
+{
+    wil::com_ptr<IWSLAContainer> container;
+    THROW_IF_FAILED(session.Get()->OpenContainer(id.c_str(), &container));
+    if (force)
+    {
+        StopInternal(*container, WSLASignalSIGKILL, StopContainerOptions::DefaultTimeout);
```
OneBlue: The frontend shouldn't manually stop the container when force is set (the service does that)

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,222 @@
...
+    return result;
+}
+
+int ContainerService::Exec(Session& session, std::string id, ExecContainerOptions options)
+{
+    wil::com_ptr<IWSLAContainer> container;
+    THROW_IF_FAILED(session.Get()->OpenContainer(id.c_str(), &container));
+
+    // Set up the options for the process to be created
+    WSLA_PROCESS_OPTIONS processOptions{};
```
OneBlue: I recommend using `WSLAProcessLauncher` to create exec processes (the overload that takes an IWSLAContainer will launch an exec) and directly return a ClientRunningWSLAProcess

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,222 @@
...
+    argsStorage.clear();
+    argsStorage.reserve(args.size());
+    for (const auto& arg : args)
+    {
+        argsStorage.push_back(arg.c_str());
+    }
+    options.CommandLine = {.Values = argsStorage.data(), .Count = static_cast<ULONG>(argsStorage.size())};
+}
+
+static void CreateInternal(
```
OneBlue: I recommend reusing `WSLAContainerLauncher` here

src/windows/wslc/SessionModel.cpp
```diff
@@ -0,0 +1,43 @@
...
+    options.m_sessionSettings.CpuCount = 4;
+    options.m_sessionSettings.MemoryMb = 2048;
+    options.m_sessionSettings.BootTimeoutMs = 30 * 1000;
+    options.StoragePath(std::move(dataFolder));
+    options.m_sessionSettings.MaximumStorageSizeMb = 10000; // 10GB.
+    options.m_sessionSettings.NetworkingMode = WSLANetworkingModeNAT;
+    return options;
+}
+
+void SessionOptions::StoragePath(std::filesystem::path path)
```
OneBlue: nit: const ref here

src/windows/wslc/SessionModel.h
```diff
@@ -0,0 +1,46 @@
...
+
+    This file contains the SessionModel definition
+
+--*/
+#pragma once
+
+#include <wslservice.h>
+#include <wslaservice.h>
+
+namespace wsl::windows::wslc::models {
```
OneBlue, line 19: This helper looks like it just wraps and returns the COM ptr, we can probably remove it

src/windows/wslc/SessionModel.h
```diff
@@ -0,0 +1,46 @@
...
+struct SessionOptions
+{
+    static SessionOptions Default();
+
+    void StoragePath(std::filesystem::path path);
+
+    operator const WSLA_SESSION_SETTINGS*() const;
+
+private:
+    WSLA_SESSION_SETTINGS m_sessionSettings;
```
OneBlue: I recommend using `WSLA_SESSION_SETTINGS m_sessionSettings{}` to make sure that everything is always default-initialized (this has bitten us in the past)

src/windows/wslc/SessionService.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "precomp.h"
+#include "SessionService.h"
+#include <wslaservice.h>
+
+namespace wsl::windows::wslc::services {
+using namespace wsl::windows::wslc::models;
+
+DEFINE_ENUM_FLAG_OPERATORS(WSLASessionFlags);
+
+Session SessionService::CreateSession(std::optional<SessionOptions> options)
```
OneBlue: nit: const ref

src/windows/wslc/SessionService.cpp
```diff
@@ -0,0 +1,36 @@
...
+{
+    SessionOptions sessionOptions = options.has_value() ? options.value() : SessionOptions::Default();
+    const WSLA_SESSION_SETTINGS* settings = sessionOptions;
```
OneBlue: This is a bit confusing, I would recommend having a `Get()` method or similar and avoiding the conversion operator here

src/windows/wslc/ShellService.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    wil::unique_cotaskmem_array_ptr<WSLA_SESSION_INFORMATION> sessions;
+    THROW_IF_FAILED(sessionManager->ListSessions(&sessions, sessions.size_address<ULONG>()));
+    for (size_t i = 0; i < sessions.size(); ++i)
+    {
+        const auto& current = sessions[i];
+        SessionInformation info;
+        info.CreatorPid = current.CreatorPid;
+        info.SessionId = current.SessionId;
+        info.DisplayName = current.DisplayName ? current.DisplayName : L"";
```
OneBlue: DisplayName should never be null so we can remove that default to empty string here

OneBlue (commented):
Change looks good overall. From a high level: 

- We should let the service pick the defaults for things like stop signal / stop timeouts and such 

- We should pass most object by const reference to avoid copies 

- Moving forward, we should avoid implementing multiple commands at once in the same PR, because it makes it difficult to review & merge (but this is fine for this once, I'm going to prioritize reviewing it quickly so we can merge it before other API changes get merged)

src/windows/wslc/ICommand.h
```diff
@@ -0,0 +1,81 @@
...
+        wsl::windows::common::wslutil::PrintMessage(msg, stderr); \
+        PrintHelp(); \
+        return E_INVALIDARG; \
+    }
+#define CMD_ARG_ARRAY_REQUIRED(argArray, msg) \
+    if (argArray.empty()) \
+    { \
+        wsl::windows::common::wslutil::PrintMessage(msg, stderr); \
+        PrintHelp(); \
+        return E_INVALIDARG; \
```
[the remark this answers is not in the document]
AmelBawa-msft: Command files are throw-away to get the client unblocked

src/windows/wslc/ImageCommand.cpp
```diff
@@ -0,0 +1,182 @@
...
+    CMD_IF_HELP_PRINT_HELP();
+    auto session = m_sessionService.CreateSession();
+    services::ImageService imageService;
+    auto images = imageService.List(session);
+    if (m_format == "json")
+    {
+        for (const models::ImageInformation& image : images)
+        {
+            wprintf(L"%hs", wsl::shared::ToJson(image).c_str());
+        }
```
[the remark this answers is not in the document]
AmelBawa-msft: by design

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,222 @@
...
+        THROW_IF_FAILED(container->GetInitProcess(&process));
+
+        ConsoleService consoleService;
+        return consoleService.AttachToCurrentConsole(
+            ClientRunningWSLAProcess(std::move(process), containerOptions.InitProcessOptions.Flags));
+    }
+
+    WSLAContainerId containerId{};
+    THROW_IF_FAILED(container->GetId(containerId));
+    PrintMessage(L"%hs\n", stdout, containerId);
```
benhillis: you're using a mix of std::format and snprintf style prints.

src/windows/wslc/main.cpp
```diff
@@ -15,17 +15,78 @@ Module Name:
...
+
+protected:
+    int ExecuteInternal(std::wstring_view commandLine, int parserOffset = 0) override
+    {
+        if (m_subverb == m_image.Name())
+        {
+            return m_image.Execute(commandLine, parserOffset + 1);
+        }
+
+        if (m_subverb == m_container.Name())
```
benhillis, line 66: this pattern is duplicated in many places, I think we should be able to use templates and avoid huge if statements.

benhillis (commented):
few comments.

src/windows/wslc/ContainerCommand.cpp
AmelBawa-msft: Moved the code to the service, so the client just passes the data as is 🙌

src/windows/wslc/ContainerService.cpp
AmelBawa-msft: Thanks for the review. Can you please elaborate on this comment? Are you referring to the runtime API?

OneBlue: Yes. The runtime API already decides what the default values are, so the CLI doesn't need to have a default timeout. 

See: https://github.com/microsoft/WSL/blob/cb2595c3f18ff6cf9f5fb29c594aa44de4a1c7b7/src/windows/wslasession/WSLAContainer.cpp#L518C6-L529C10

AmelBawa-msft: Gotcha! So I would just need to use -1 for timeout right? For signal, would delete use the kill signal?

src/windows/wslc/ConsoleService.cpp
AmelBawa-msft: This is copied from wsladiag, which I believe will be deleted in favor of wslc

src/windows/wslc/SessionModel.h
AmelBawa-msft: Your right! I wanted to have a consistent design that wraps COM types for the client, which (I think) wouldn't need to know if it's COM or native type. However, that's probably the simplest use case for the wrapper 😁 But I will consider your suggestion as I continue adding features 😊

src/windows/wslc/ContainerService.cpp
AmelBawa-msft: Right, I also noticed this as I copied from wsladiag. Once we have the frontend CLI in place, I believe we'll be using a different message reporter and hopefully eliminate inconsistency moving forward.

src/windows/wslc/main.cpp
AmelBawa-msft: This temp code is just to get me unblocked as I wait for the more robust CLI structure to be merged (after this PR) 😊

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,218 @@
...
+{
+    // Create the container
+    auto runningContainer = CreateInternal(session, image, runOptions, callback);
+    runningContainer.SetDeleteOnClose(false);
+    auto& container = runningContainer.Get();
+
+    // Start the created container
+    WSLAContainerStartFlags startFlags{};
+    WI_SetFlagIf(startFlags, WSLAContainerStartFlagsAttach, !runOptions.Detach);
+    THROW_IF_FAILED(container.Start(startFlags)); // TODO: Error message
```
OneBlue, line 110: The error message will be automatically propagated via the ExecutionContext once https://github.com/microsoft/WSL/pull/14199 is merged

src/windows/wslc/SessionModel.h
```diff
@@ -0,0 +1,43 @@
...
+    This file contains the SessionModel definition
+
+--*/
+#pragma once
+
+#include <wslservice.h>
+#include <wslaservice.h>
+
+namespace wsl::windows::wslc::models {
+struct Session
```
OneBlue, line 20: nit: this struct doesn't add any functionally on top of `m_session` so I recommend using ` wil::com_ptr<IWSLASession>` directly

src/windows/wslc/ContainerService.cpp
```diff
@@ -0,0 +1,218 @@
...
+        if (consoleOutput && tryGetConsoleInfo(consoleOutput.get(), info))
+        {
+            options.TtyColumns = info.srWindow.Right - info.srWindow.Left + 1;
+            options.TtyRows = info.srWindow.Bottom - info.srWindow.Top + 1;
+            return;
+        }
+    }
+
+    PrintMessage(L"error: --tty requires stdin or stdout to be a console", stderr);
+    THROW_HR(E_FAIL);
```
OneBlue, line 64: FYI this is one of the areas where we can use: 

`THROW_HR_WITH_USER_ERROR(E_FAIL, L"error: --tty requires stdin or stdout to be a console");`

(Of course in practice we'd want the error to be localized). 

But let's do that in a future change

OneBlue (approved):
LGTM, thank you for doing this ! 

Feel free to merge and resolve the outstanding comment in a followup PR

PR #14172, opened 2026-02-06
Merge the WSL Container SDK DLL project into the main feature branch

1wizkid:
this adds a new project src/windows/wslcsdk and the necessary .h and cpp files to build the WSLCSDK.DLL.


benhillis:
I've updated the target branch.

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+    return E_NOTIMPL;
+}
+
+
+STDAPI WslcSessionTerminate(_In_ WslcSession session)
+{
+    UNREFERENCED_PARAMETER(session);
+    return E_NOTIMPL;
+}
+STDAPI WslcContainerSettingsNetworkingMode(_In_ WslcContainerSettings containerSettings,
```
OneBlue: nit: I'd rename to WslcContainerSettingsSetNetworkingMode (same for below method) to follow the get / set naming

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+STDAPI WslcSessionSettingsVHD(_In_ WslcSessionSettings sessionSettings,
+                              _In_ WSLC_VHD_REQUIREMENTS WslcVHDRequirements)
+{
+    UNREFERENCED_PARAMETER(WslcVHDRequirements);
+    UNREFERENCED_PARAMETER(sessionSettings);
+    return E_NOTIMPL;
+}
+STDAPI WslcContainerSettingsHostName(_In_ WslcContainerSettings containerSettings,
+                                     _In_ const PCSTR hostName)
```
OneBlue: nit: `const` is not needed here (already in PCSTR)`

OneBlue: Also I recommend using the long pointer version (LPCSTR) since that's what modern Windows APIs do

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+--*/
+#include "precomp.h"
+
+#include "wslcsdk.h"
+#include "wslcsdkprivate.h"
+
+
+// SESSION DEFINITIONS
+STDAPI WslcSessionInitSettings(_In_ PCWSTR storagePath,
```
OneBlue: This method can't fail, since I recommend returning the sessions by copy to simplify things. This also remove the need for the "release" method for settings

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+}
+
+STDAPI WslcContainerSettingsFlags(_In_ WslcContainerSettings containerSettings,
+                                  _In_ WSLC_CONTAINER_FLAGS flags)
+{
+    UNREFERENCED_PARAMETER(flags);
+    UNREFERENCED_PARAMETER(containerSettings);
+    return E_NOTIMPL;
+}
+STDAPI WslcContainerSettingsRuntimeName(_In_ WslcContainerSettings containerSettings,
```
OneBlue: I'd just use `WslcContainerSettingsSetName` (runtime is implied)

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+STDAPI WslcContainerSettingsVolume(_In_ WslcContainerSettings containerSettings,
+                                   _In_ const WSLC_CONTAINER_VOLUME* volumes)
+{
+    UNREFERENCED_PARAMETER(volumes);
+    UNREFERENCED_PARAMETER(containerSettings);
+    return E_NOTIMPL;
+}
+
+STDAPI WslcContainerExecProcess(_In_ WslcContainer container,
```
OneBlue: I recommend just `WslcContainerExec`  here

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+STDAPI WslcContainerGetState(_In_ WslcContainer container,
+                             _Out_ WSLC_CONTAINER_STATE* state)
+{
+    UNREFERENCED_PARAMETER(container);
+    UNREFERENCED_PARAMETER(state);
+    return E_NOTIMPL;
+}
+
+STDAPI WslcContainerStop(_In_ WslcContainer container)
```
OneBlue: This method should take stop signal and timeout

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+{
+    UNREFERENCED_PARAMETER(container);
+    return E_NOTIMPL;
+}
+
+STDAPI WslcContainerDelete(_In_ WslcContainer container)
```
OneBlue: We should have a set of flags here (to know when to do a forced delete for instance)

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+}
+
+// PROCESS DEFINITIONS
+
+STDAPI WslcProcessInitSettings(_Out_ WslcProcessSettings* processSettings)
+{
+    UNREFERENCED_PARAMETER(processSettings);
+    return E_NOTIMPL;
+}
+STDAPI WslcProcessSettingsEntryPoint(_In_ WslcProcessSettings processSettings,
```
OneBlue: Entrypoints only apply to containers, not to execs. So I recommend moving the entrypoint to the container settings

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+{
+    UNREFERENCED_PARAMETER(currentDirectory);
+    UNREFERENCED_PARAMETER(processSettings);
+    return E_NOTIMPL;
+}
+
+// OPTIONAL PROCESS SETTINGS
+
+STDAPI WslcProcessSettingsCmdLineArgs(WslcProcessSettings processSettings,
+                                     PWCHAR const* argv,
```
OneBlue: Arguments should be utf8 (char)

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+                                     size_t argc)
+{
+    UNREFERENCED_PARAMETER(argv);
+    UNREFERENCED_PARAMETER(argc);
+    UNREFERENCED_PARAMETER(processSettings);
+    return E_NOTIMPL;
+}
+
+STDAPI WslcProcessSettingsEnvVariables(_In_ WslcProcessSettings processSettings,
+                                       _In_ PWCHAR const* key_value,
```
OneBlue: Environment should be utf8 (char) as well

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+                         _In_ WSLC_PROCESS_SIGNAL signal)
+{
+    UNREFERENCED_PARAMETER(process);
+    UNREFERENCED_PARAMETER(signal);
+    return E_NOTIMPL;
+}
+
+STDAPI WslcProcessSettingsIoCallback(_In_ WslcProcessSettings processSettings,
+                                     _In_ WSLC_PROCESS_IO_HANDLE ioHandle,
+                                     _In_ WslcStdIOCallback stdIOCallback)
```
OneBlue: We should pass an arbitrary void pointer here (for context)

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+STDAPI WslcSessionImageLoad(_In_ WslcSession session,
+                            _In_ const WSLC_LOAD_CONTAINER_IMAGE_OPTIONS* options)
+{
+    UNREFERENCED_PARAMETER(session);
+    UNREFERENCED_PARAMETER(options);
+    return E_NOTIMPL;
+}
+
+STDAPI WslcSessionImageDelete(_In_ WslcSession session,
+                              _In_reads_(32) const UINT8* sha256)
```
OneBlue: I'd rename `sha256` to NameOrId, also let's make it a char *

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+{
+    UNREFERENCED_PARAMETER(session);
+    UNREFERENCED_PARAMETER(options);
+    return E_NOTIMPL;
+}
+
+// INSTALL
+
+STDAPI WslcCanRun(_Out_ BOOL* canRun,
+                  _Out_ WSLC_COMPONENT_FLAGS missingComponents)
```
OneBlue: `missingComponents` should be a pointer

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,513 @@
...
+#include <windows.h>
+#include <stdint.h>
+#include <winsock2.h>
+
+
+EXTERN_C_START
+
+//Session values
+#define WSLC_SESSION_OPTIONS_SIZE 80
+#define WSLC_SESSION_OPTIONS_ALIGNMENT 8
```
OneBlue, line 25: I would recommend removing the explicit alignment for simplicity. We can have a static_assert inside the .cpp file to validate that sizes match

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,513 @@
...
+    WSLA_NetworkingModeBridged,    
+} WSLC_ContainerNetworkingMode;
+
+typedef enum WSLC_VhdType
+{
+    WSLA_VhdTypeDynamic = 0, // Expanding VHDX (default)
+    WSLA_VhdTypeFixed,       // Fully allocated VHDX
+} WSLC_VhdType;
+
+typedef struct WSLC_VHD_REQUIREMENTS
```
OneBlue: I recommend using the same casing for all enum and structures (WSLCVhdRequirements) 

Same for below enums

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,513 @@
...
+
+typedef enum WSLC_PORT_PROTOCOL
+{
+    WslcProtocol_TCP = 0,
+    WslcProtocol_UDP = 1
+} WSLC_PORT_PROTOCOL;
+
+typedef struct WSLC_CONTAINER_PORT_MAPPING
+{
+    _In_ UINT16 windowsPort;          // Port on Windows host
```
OneBlue: Let's use uint16_t to match the rest of the API

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,513 @@
...
+// Notes:
+//   - WSLC frees or reuses the buffer immediately after the callback returns.
+//   - The callback must return promptly; long-running operations may block
+//     WSLC's internal I/O processing.
+//   - The buffer is not null-terminated; it is a raw byte sequence.
+//
+typedef __callback VOID(CALLBACK WslcStdIOCallback)(_In_reads_bytes_(dataSize) const BYTE* data, _In_ UINT32 dataSize, _In_opt_ PVOID context);
+typedef enum WSLC_PROCESS_IO_HANDLE
+{
+    stdIn = 0,
```
OneBlue: This will conflicts with predefined std handles. I recommend prefixing the enum values with the enum type name (for all enums)

src/windows/WslcSDK/WslcsdkPrivate.h
```diff
@@ -0,0 +1,93 @@
...
+#pragma once
+#include <windows.h>
+#include "wslcsdk.h"
+#include <stdint.h>
+#include <wil/com.h>      // COM helpers
+//#include <wil/resource.h> // handle wrappers
+//#include <wil/result.h>   // error handling
+
+// SESSION DEFINITIONS
+typedef struct WSLC_SESSION_OPTIONS_INTERNAL
```
OneBlue, line 24: Let's make this a C++ header (no need for typedef struct)

src/windows/WslcSDK/WslcsdkPrivate.h
```diff
@@ -0,0 +1,93 @@
...
+    uint64_t memoryMb;
+    uint32_t timeoutMS;
+
+    WSLC_VHD_REQUIREMENTS vhdRequirements;
+    WSLC_SESSION_FLAGS flags;
+    WslcSessionTerminationCallback terminationCallback;
+    PVOID terminationCallbackContext;
+} WSLC_SESSION_OPTIONS_INTERNAL;
+
+typedef char WSLC_SESSION_OPTIONS_INTERNAL_SIZE_ASSERT[sizeof(WSLC_SESSION_OPTIONS_INTERNAL)];//size = 80
```
OneBlue: I recommend static_assert instead of this

OneBlue (changes requested):
Thank you for doing this. Added a couple comments, most of the copilot comments seem to be accurate

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+    UNREFERENCED_PARAMETER(containerSettings);
+    return E_NOTIMPL;
+}
+
+STDAPI WslcSessionSettingsDisplayName(_In_ WslcSessionSettings sessionSettings,
+                                      _In_ PCWSTR displayName)
+{
+    UNREFERENCED_PARAMETER(displayName);
+    UNREFERENCED_PARAMETER(sessionSettings);
+    return E_NOTIMPL;
```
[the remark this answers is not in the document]
1wizkid: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/WSL/pull/14172#discussion_r2776380966)

src/windows/WslcSDK/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+    wslcsdk.h
+    wslcsdkprivate.h
+)
+
+add_library(wslcsdk SHARED ${SOURCES} ${HEADERS} wslcsdk.def)
+set_target_properties(wslcsdk PROPERTIES EXCLUDE_FROM_ALL FALSE)
+add_dependencies(wslcsdk wslaserviceidl)
+target_link_libraries(wslcsdk ${COMMON_LINK_LIBRARIES} legacy_stdio_definitions common)
+target_precompile_headers(wslcsdk REUSE_FROM common)
+set_target_properties(wslcsdk PROPERTIES FOLDER windows)
```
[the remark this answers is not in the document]
1wizkid: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/WSL/pull/14172#discussion_r2776380974)

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,467 @@
...
+{
+
+    
+    // demo test code to show how to cast the opaque struct to the internal struct and set some values. The real implementation would have actual logic here.
+    WSLC_SESSION_OPTIONS_INTERNAL* sessionSettingsinternal = (WSLC_SESSION_OPTIONS_INTERNAL*)sessionSettings;
+
+    UNREFERENCED_PARAMETER(storagePath);
+    UNREFERENCED_PARAMETER(cpuCount);
+    UNREFERENCED_PARAMETER(memoryMb);
+    //UNREFERENCED_PARAMETER(sessionSettings);
```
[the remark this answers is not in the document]
1wizkid: Made it a comment

src/windows/WslcSDK/wslcsdk.cpp
1wizkid: Changed all the Set() API's to have Set in the name

src/windows/WslcSDK/wslcsdk.cpp
1wizkid: Removed EntryPoint reference

src/windows/WslcSDK/WslcsdkPrivate.h
```diff
@@ -0,0 +1,93 @@
...
+    WslcSDKPrivate.h
+
+Abstract:
+
+    This file contains the private WSL Container SDK definitions.
+
+--*/
+#pragma once
+#include <windows.h>
+#include "wslcsdk.h"
```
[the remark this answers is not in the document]
1wizkid: fixed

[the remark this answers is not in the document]
1wizkid: fixed

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,534 @@
...
+typedef enum WSLC_SESSION_TERMINATION_REASON
+{
+    WSLC_SESSION_TERMINATION_REASON_UNKNOWN = 0,
+    WSLC_SESSION_TERMINATION_REASON_SHUTDOWN = 1,
+    WSLC_SESSION_TERMINATION_REASON_CRASHED = 2,
+} WSLC_SESSION_TERMINATION_REASON;
+
+typedef __callback VOID(CALLBACK *WslcSessionTerminationCallback)(_In_ WSLC_SESSION_TERMINATION_REASON reason, _In_opt_ PVOID context);
+
+STDAPI_(void) WslcSessionInitSettings(_In_ PCWSTR storagePath,
```
JohnMcPMS: Should return `HRESULT`; internal memory allocations may fail, etc.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,534 @@
...
+                                      _In_ uint32_t cpuCount,
+                                      _In_ uint64_t memoryMb,
```
JohnMcPMS: Per discussion, there is a desire to have defaults for CPU and memory, which likely means moving to their own setter (although since 0 is not a reasonable value for either it could be the way to request the default).  To me that also means needing getters because the caller has not provided these values.

src/windows/WslcSDK/wslcsdk.cpp
1wizkid: Flags added. We can update as needed with supported parameters

src/windows/WslcSDK/wslcsdk.cpp
1wizkid: updated to reflect that

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,534 @@
...
+
+STDAPI WslcContainerCreate(_In_ WslcContainerSettings* containerSettings,
+                           _Out_ WslcContainer* container,
+                           _Outptr_opt_result_z_ PWSTR* errorMessage);
+
+STDAPI WslcContainerStart(_In_ WslcContainer container);
+
+//OPTIONAL CONTAINER SETTINGS
+STDAPI WslcContainerSettingsSetName(_In_ WslcContainerSettings* containerSettings,
+                                        _In_ PCSTR name);
```
JohnMcPMS:
```suggestion
                                        _In_ PCWSTR name);
```
Unless this is intentionally UTF-8, or there are technical reasons for otherwise restricting to ASCII.

src/windows/WslcSDK/wslcsdk.cpp
1wizkid: Updated apis

src/windows/WslcSDK/wslcsdk.cpp
1wizkid: updated to reflect that

src/windows/WslcSDK/wslcsdk.cpp
1wizkid: corrected

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,534 @@
...
+
+EXTERN_C_START
+
+//Session values
+#define WSLC_SESSION_OPTIONS_SIZE 80
+#define WSLC_SESSION_OPTIONS_ALIGNMENT 8
+
+typedef struct WslcSessionSettings
+{
+    __declspec(align(WSLC_SESSION_OPTIONS_ALIGNMENT)) BYTE _opaque[WSLC_SESSION_OPTIONS_SIZE];
```
JohnMcPMS, line 29: I don't understand why these aren't just HANDLEs as well.  Are we just trying to save on an allocation with the expectation that callers create them on the stack?  It seems like a hyper optimization on something that likely isn't happening regularly.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,534 @@
...
+                                     _In_ const PCSTR hostName);
+
+STDAPI WslcContainerSettingsSetDomainName(_In_ WslcContainerSettings* containerSettings,
+                                       _In_ const PCSTR domainName);
+
+
+STDAPI WslcContainerSettingsSetFlags(_In_ WslcContainerSettings* containerSettings,
+                                  _In_ WSLC_CONTAINER_FLAGS flags);
+
+STDAPI WslcContainerSettingsSetPortMapping(_In_ WslcContainerSettings* containerSettings,
```
JohnMcPMS: Presumably no ability to remove mappings ("Don't add them if you don't want them")?

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,534 @@
...
+
+typedef enum WSLC_VhdType
+{
+    WSLC_VhdTypeDynamic = 0, // Expanding VHDX (default)
+    WSLC_VhdTypeFixed,       // Fully allocated VHDX
+} WSLC_VhdType;
+
+typedef struct WSLC_VHD_REQUIREMENTS
+{
+    _In_z_ PCWSTR path;      // Full path to the VHD/VHDX file
```
OneBlue: Path is already passed when the session settings are created, it doesn't need to be inside that structure

src/windows/WslcSDK/wslcsdk.h
JohnMcPMS: Is it expected that the settings objects are destroyed (at least internally) when their respective object's Create is called?  Given the current API, it is either that or no internal allocations are allowed (and given that paths are being stored and none of them are even MAX_PATH sized, it cannot be that option).

I think we need a method to release/destroy the settings.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,534 @@
...
+                         _In_ WSLC_PROCESS_SIGNAL signal,
+                         _In_ uint32_t timeoutMS);
+
+typedef enum WslcDeleteContainerFlags
+{
+    DELETE_FLAG_NONE = 0,
+    DELETE_FLAG_FORCE = 0x01
+} WslcDeleteContainerFlags;
+
+STDAPI WslcContainerDelete(_In_ WslcContainer container,
```
JohnMcPMS: The container operations are `Create`, `Start`, `Stop`, `Delete`, `Release`.  If there are differences between `Delete` and `Release` in terms of lasting state, do we need a method to create a container object from an existing container item?

src/windows/WslcSDK/wslcsdk.h
1wizkid: I think what you are suggesting is to not check the alignment directly but just include in a single value. I'd like to leave this for now until I can consider longer term compiler compatibility, etc. with not calling it out directly,

src/windows/WslcSDK/wslcsdk.h
1wizkid: That is a good point. Lets get through all these changes first and then I'll make a revision to settle on one format.

1wizkid (commented):
Updates made.

JohnMcPMS (commented):
I want to voice that I believe we should be using MIDL3 to author this API definition (and thus COM to implement it).  Many of the structs defined here are not extensible and will require `2` variants to support any future additions.  Any higher level language projections will need to be written manually, despite likely looking like the projections that would be generated by their language specific libraries.

It would be far easier to write this C ABI wrapper around the above mentioned COM API than it will be to create even one of the higher level projection copies.  But I wouldn't do that unless we have customers calling for it explicitly in great numbers; anyone using C can still use COM.

src/windows/WslcSDK/wslcsdk.h
1wizkid: We started as handles, but ultimately did not like the fact we then had to expose another release() api for each of these.

src/windows/WslcSDK/wslcsdk.h
1wizkid: Create/Start/Stop/Delete have actual implications with the runtime (Ie: they map to a runtime call). Release() we are just freeing any sdk related memory we allocated that was needed to service that object.

src/windows/WslcSDK/wslcsdk.h
1wizkid: The data in the settings struct is only used once when they call Create() and pass it in. We use it to send the right creation parameters to the runtime and return object handle that represents what the runtime created. After that call concludes, we don't use that struct any more. Any changes they make to it, etc. have no effect.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,534 @@
...
+
+//OPTIONAL PROCESS SETTINGS
+
+STDAPI WslcProcessSettingsSetExecutable(_In_ WslcProcessSettings* processSettings,
+                                     _In_ const PCSTR entryPoint);
+
+STDAPI WslcProcessSettingsSetCurrentDirectory(_In_ WslcProcessSettings* processSettings,
+                                           _In_ const PCSTR currentDirectory);
+
+STDAPI WslcSessionSettingsSetDisplayName(_In_ WslcSessionSettings* sessionSettings,
```
JohnMcPMS: Move with other session settings.

src/windows/WslcSDK/wslcsdk.h
JohnMcPMS: If the SDK requires the caller to Delete, does that imply that the runtime would preserve the container item across sessions / caller processes?  Or is it for intra-caller process container management only?

src/windows/WslcSDK/wslcsdk.h
JohnMcPMS: So the other option that I didn't list was that the settings objects are non-owning, which sounds like the case here.

JohnMcPMS: Better reason, invalid arguments.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+typedef struct WslcProcessSettings
+{
+    __declspec(align(WSLC_CONTAINER_PROCESS_OPTIONS_ALIGNMENT)) BYTE _opaque[WSLC_CONTAINER_PROCESS_OPTIONS_SIZE];
+} WslcProcessSettings;
+
+DECLARE_HANDLE(WslcProcess);
+
+typedef enum WslcContainerNetworkingMode
+{
+    WSLC_NetworkingModeNone = 0, // No networking / isolated
```
JohnMcPMS: I find it awkward that the enum values are not of the form with the enum name prefixing them (precisely prefixing them with the same casing, etc.).  It makes it harder to type the values for `EnumName` when they are `ENUM_NAME_VALUE_NAME` instead of the more natural `EnumNameValueName`.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+} WslcSessionFlags;
+
+typedef enum WslcSessionTerminationReason
+{
+    WSLC_SESSION_TERMINATION_REASON_UNKNOWN = 0,
+    WSLC_SESSION_TERMINATION_REASON_SHUTDOWN = 1,
+    WSLC_SESSION_TERMINATION_REASON_CRASHED = 2,
+} WslcSessionTerminationReason;
+
+typedef __callback VOID(CALLBACK* WslcSessionTerminationCallback)(_In_ WslcSessionTerminationReason reason, _In_opt_ PVOID context);
```
JohnMcPMS: The runtime callback includes a `LPCWSTR Details`; are we ok with that being lost to the SDK caller?

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+{
+    _In_z_ PCWSTR path;      // Full path to the VHD/VHDX file
+    _In_ UINT64 sizeInBytes; // Desired size (for create/expand)
+    _In_ WslcVhdType type;   // Dynamic / Fixed
+} WslcVhdRequirements;
+
+typedef enum WslcSessionFlags
+{
+    WSLC_SESSION_FLAG_NONE = 0x00000000,
+    WSLC_SESSION_FLAG_ENABLE_GPU = 0x00000001
```
JohnMcPMS, line 76: There are many more flags in the runtime, is this list intentionally limited?

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+{
+    WSLC_VHDTYPEDYNAMIC = 0, // Expanding VHDX (default)
+    WSLC_VHDTYPEFIXED,       // Fully allocated VHDX
+} WslcVhdType;
+
+typedef struct WslcVhdRequirements
+{
+    _In_z_ PCWSTR path;      // Full path to the VHD/VHDX file
+    _In_ UINT64 sizeInBytes; // Desired size (for create/expand)
+    _In_ WslcVhdType type;   // Dynamic / Fixed
```
JohnMcPMS: For session settings, you are using the type value but the only runtime session setting it maps to is actually the `filesystemtype` to the linux `mount` function.  It is not clear how to map dynamic and fixed to values like `ext4` and `tmpfs`.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+
+typedef __callback VOID(CALLBACK* WslcSessionTerminationCallback)(_In_ WslcSessionTerminationReason reason, _In_opt_ PVOID context);
+
+STDAPI WslcSessionInitSettings(_In_ PCWSTR storagePath, _Out_ WslcSessionSettings* sessionSettings);
+
+STDAPI WslcSessionCreate(_In_ WslcSessionSettings* sessionSettings, _Out_ WslcSession* session);
+
+// OPTIONAL SESSION SETTINGS
+STDAPI WslcSessionSettingsSetCpuCount(_In_ WslcSessionSettings* sessionSettings, _In_ uint32_t cpuCount);
+STDAPI WslcSessionSettingsSetMemory(_In_ WslcSessionSettings* sessionSettings, _In_ uint64_t memoryMb);
```
JohnMcPMS: The runtime session setting is a 32-bit value, so we can't actually send in 64 bits for now.

For practical purposes, a 32-bit value gets you 4 PB of memory.  I doubt that this will be a limiting factor for the API any time soon (hopefully we are all retired by the time someone complains about it).

At the same time, we can also just error if given a value that is too large to represent.

src/windows/WslcSDK/WslcsdkPrivate.h
```diff
@@ -0,0 +1,99 @@
...
+
+    uint32_t cpuCount;
+    uint64_t memoryMb;
+    uint32_t timeoutMS;
+
+    WslcVhdRequirements vhdRequirements;
+    WslcSessionFlags flags;
+    WslcSessionTerminationCallback terminationCallback;
+    PVOID terminationCallbackContext;
+} WSLC_SESSION_OPTIONS_INTERNAL;
```
JohnMcPMS, line 37: There are no user controls over the following available runtime settings:

1. MaximumStorageSizeMb :: Related to the storage path; while there is a value in the VHD requirements, it doesn't seem related as this value is for use at all times while the other VHD requirements are debugging related.
2. NetworkingMode :: Available runtime values are NAT and VirtIO
3. DmesgOutput :: Appears to be a handle that I assume gets debug message written
4. Session Flags :: These include Persistent and OpenExisting

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+typedef enum WslcImageProgressStatus
+{
+    WSLC_IMAGE_PROGRESS_UNKNOWN = 0,
+    WSLC_IMAGE_PROGRESS_PULLING,     // "Pulling fs layer"
+    WSLC_IMAGE_PROGRESS_WAITING,     // "Waiting"
+    WSLC_IMAGE_PROGRESS_DOWNLOADING, // "Downloading"
+    WSLC_IMAGE_PROGRESS_VERIFYING,   // "Verifying Checksum"
+    WSLC_IMAGE_PROGRESS_EXTRACTING,  // "Extracting"
+    WSLC_IMAGE_PROGRESS_COMPLETE     // "Pull complete"
+} WslcImageProgressStatus;
```
JohnMcPMS, line 329: The runtime provides these as strings, directly from the engine progress. The engine might be revved independently from the SDK, leading to a skew on the mapping of these strings to the enum.  Is the value of mapping them for the caller greater than the risk of mapping them to Unknown if they change?

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+// options struct � typedef is a pointer type and _In_opt_ is valid
+typedef struct WslcPullImageOptions
+{
+    _In_z_ PCSTR uri;
+    WslcContainerImageProgressCallback progressCallback;
+    PVOID progressCallbackContext;
+    _In_opt_ const WslcResgistryAuthenticationInformation* authInfo;
+} WslcPullImageOptions;
+
+STDAPI WslcSessionImagePull(_In_ WslcSession session, _In_ const WslcPullImageOptions* options, _Outptr_opt_result_z_ PWSTR* errorMessage);
```
JohnMcPMS, line 355: Needs comment on caller ownership and `CoTaskMemFree` call for `errorMessage` if provided.

src/windows/WslcSDK/wslcsdk.h
OneBlue: What I'm suggesting is this: 

```
// Public API header
struct WslcSessionSettings
{
    [...]
};

// SDK implementation

static_assert(sizeof(WslcSessionSettings), sizeof(WslcSessionSettingsPrivate));
```
This will guarantee that the size matches between the public header and implementation. 

AFAIK Alignment doesn't matter here because the caller isn't actually accessing fields inside the structure, only the SDK implementation is

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+
+typedef enum WslcVhdType
+{
+    WSLC_VHDTYPEDYNAMIC = 0, // Expanding VHDX (default)
+    WSLC_VHDTYPEFIXED,       // Fully allocated VHDX
+} WslcVhdType;
+
+typedef struct WslcVhdRequirements
+{
+    _In_z_ PCWSTR path;      // Full path to the VHD/VHDX file
```
OneBlue: This can be removed since it's already passed to `WslcSessionInitSettings`

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+typedef enum WslcSessionTerminationReason
+{
+    WSLC_SESSION_TERMINATION_REASON_UNKNOWN = 0,
+    WSLC_SESSION_TERMINATION_REASON_SHUTDOWN = 1,
+    WSLC_SESSION_TERMINATION_REASON_CRASHED = 2,
+} WslcSessionTerminationReason;
+
+typedef __callback VOID(CALLBACK* WslcSessionTerminationCallback)(_In_ WslcSessionTerminationReason reason, _In_opt_ PVOID context);
+
+STDAPI WslcSessionInitSettings(_In_ PCWSTR storagePath, _Out_ WslcSessionSettings* sessionSettings);
```
OneBlue, line 88: Given that this method cannot fail, I recommend returning by copy here: 

```
WslcSessionSettings WslcSessionInitSettings(_In_ PCWSTR storagePath);
```

This will simplify things for the caller and make it clear that this structure doesn't need to be explicitly released

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+
+STDAPI WslcSessionCreate(_In_ WslcSessionSettings* sessionSettings, _Out_ WslcSession* session);
+
+// OPTIONAL SESSION SETTINGS
+STDAPI WslcSessionSettingsSetCpuCount(_In_ WslcSessionSettings* sessionSettings, _In_ uint32_t cpuCount);
+STDAPI WslcSessionSettingsSetMemory(_In_ WslcSessionSettings* sessionSettings, _In_ uint64_t memoryMb);
+STDAPI WslcSessionSettingsSetTimeout(_In_ WslcSessionSettings* sessionSettings, uint32_t timeoutMS);
+
+STDAPI WslcSessionSettingsSetVHD(_In_ WslcSessionSettings* sessionSettings, _In_ WslcVhdRequirements* vhdRequirements);
```
OneBlue: nit: `const WslcVhdRequirements*`

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+STDAPI WslcContainerSettingsSetHostName(_In_ WslcContainerSettings* containerSettings, _In_ PCSTR hostName);
+
+STDAPI WslcContainerSettingsSetDomainName(_In_ WslcContainerSettings* containerSettings, _In_ PCSTR domainName);
+
+STDAPI WslcContainerSettingsSetFlags(_In_ WslcContainerSettings* containerSettings, _In_ WslcContainerFlags flags);
+
+STDAPI WslcContainerSettingsSetPortMapping(_In_ WslcContainerSettings* containerSettings, _In_ const WslcContainerPortMapping* portMappings);
+
+// Add the container volume to the volumes array
+STDAPI WslcContainerSettingsSetVolume(_In_ WslcContainerSettings* containerSettings, _In_ const WslcContainerVolume* volumes);
```
OneBlue: I'd rename to `WslcContainerSettingsAddVolume` since multiple volumes can be added

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+} WslcDeleteContainerFlags;
+
+STDAPI WslcContainerDelete(_In_ WslcContainer container, _In_ WslcDeleteContainerFlags flags);
+
+// PROCESS DEFINITIONS
+STDAPI WslcProcessInitSettings(_Out_ WslcProcessSettings* processSettings);
+
+// OPTIONAL PROCESS SETTINGS
+
+STDAPI WslcProcessSettingsSetExecutable(_In_ WslcProcessSettings* processSettings, _In_ const PCSTR entryPoint);
```
OneBlue: nit: `const` is redundant here

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+    _Out_ WslcImageProgressDetail detail;
+} WslcImageProgressMessage;
+
+typedef struct WslcResgistryAuthenticationInformation
+{
+    // TBD
+} WslcResgistryAuthenticationInformation;
+
+// pointer-to-function typedef (unambiguous)
+typedef VOID(CALLBACK* WslcContainerImageProgressCallback)(const WslcImageProgressMessage* progress, PVOID context);
```
OneBlue: I recommend `void` instead of `VOID` here

src/windows/WslcSDK/wslcsdk.h
OneBlue, line 355: I recommend passing the auth info as a separate argument to `WslcSessionImagePull` instead of a field via `WslcPullImageOptions` (that will avoid having a structure's field pointing to another structure) and make the default-to-null logic more clear IMO

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+    WslcContainerImageProgressCallback progressCallback;
+    PVOID progressCallbackContext;
+} WslcLoadImageOptions;
+
+STDAPI WslcSessionImageLoad(_In_ WslcSession session, _In_ const WslcLoadImageOptions* options);
+
+typedef struct WslcImageInfo
+{
+    // we should expose this
+    PCSTR repository; // e.g., "ubuntu"
```
OneBlue: I recommend settings fixed sizes for `tag` and `repository` as well (this will make this API much nicer to use)

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+
+typedef struct WslcVersion
+{
+    UINT32 major;
+    UINT32 minor;
+    UINT32 revision;
+} WslcVersion;
+STDAPI WslcGetVersion(_Out_writes_(1) WslcVersion* version);
+
+typedef __callback VOID(CALLBACK* WslcInstallCallback)(_In_ WslcComponentFlags component, _In_ UINT32 progress, _In_ UINT32 total, _In_opt_ PVOID context);
```
OneBlue: `VOID` -> `void`

OneBlue (changes requested):
Thank you for the updates ! 

Added  a couple comments. 

Overall we're pretty close, I would recommend having a pass on the API header to : 

- make sure that the types we use are consistent (UINT32 vs uint32_t, VOID vs void, etc) 
- Make sure that enum values are prefixed by the enum type (with the same casing) 
- All operations that can return errors from the docker API can return the error string to the caller

src/windows/WslcSDK/wslcsdk.h
1wizkid: I've added the static asserts to the WslcsdkPrivate.h file

src/windows/WslcSDK/WslcsdkPrivate.h
1wizkid: will file a bug for this

src/windows/WslcSDK/WslcsdkPrivate.h
1wizkid: fixed.

src/windows/WslcSDK/wslcsdk.h
1wizkid: Yes, non-owing. We process in create and then disgard usage.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+{
+    _Out_ PCSTR id;                       // layer ID or digest
+    _Out_ WslcImageProgressStatus status; // "Downloading", "Extracting", etc.
+    _Out_ WslcImageProgressDetail detail;
+} WslcImageProgressMessage;
+
+typedef struct WslcResgistryAuthenticationInformation
+{
+    // TBD
+} WslcResgistryAuthenticationInformation;
```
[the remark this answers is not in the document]
1wizkid: fixed type

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -0,0 +1,406 @@
...
+#include "wslcsdk.h"
+#include "wslcsdkprivate.h"
+
+// SESSION DEFINITIONS
+STDAPI WslcSessionInitSettings(_In_ PCWSTR storagePath, _Out_ WslcSessionSettings* sessionSettings)
+{
+
+    // demo test code to show how to cast the opaque struct to the internal struct and set some values. The real implementation would have actual logic here.
+    // WSLC_SESSION_OPTIONS_INTERNAL* sessionSettingsinternal = (WSLC_SESSION_OPTIONS_INTERNAL*)sessionSettings;
+    sessionSettings = NULL;
```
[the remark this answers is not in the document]
1wizkid: this is just temp code to mark the var as being used. Will get replaced.

src/windows/WslcSDK/wslcsdk.h
1wizkid: Once the session ends, the container is no longer available for reuse. The cannot restart a session and then access the container.

src/windows/WslcSDK/wslcsdk.h
1wizkid: Noted.

src/windows/WslcSDK/wslcsdk.h
1wizkid: Could have sometime of invalid parameter issue. Would prefer to keep the syntax like the other ones.

src/windows/WslcSDK/wslcsdk.h
1wizkid: I do agree this is a good idea. However, given we don't have the final definition yet of how this would work, I'm going to leave as is, and will incorporate this if final design does not require a struct.

src/windows/WslcSDK/wslcsdk.h
1wizkid: upated

src/windows/WslcSDK/wslcsdk.h
JohnMcPMS: Alignment matters to the caller because they are allocating the memory for the structure.

src/windows/WslcSDK/wslcsdk.h
JohnMcPMS: There may be a need/desire to validate `storagePath` in some way in the future.  I don't think it wise to change the standard API pattern here because it cannot currently fail.

OneBlue: That would be OK. We can decide to change the SDK API in a future update, since it's compiled within the app

src/windows/WslcSDK/wslcsdk.h
OneBlue: I don't think it does if the caller doesn't actually access the fields (especially if we create the structure by having N chars / bytes inside of it). 

For instance, the PEB structure has opaque fields, and it's declared like this (without explicit alignment indication): 

```
ypedef struct _PEB {
    BYTE Reserved1[2];
    BYTE BeingDebugged;
    BYTE Reserved2[1];
    PVOID Reserved3[2];
    PPEB_LDR_DATA Ldr;
    PRTL_USER_PROCESS_PARAMETERS ProcessParameters;
    PVOID Reserved4[3];
    PVOID AtlThunkSListPtr;
    PVOID Reserved5;
    ULONG Reserved6;
    PVOID Reserved7;
    ULONG Reserved8;
    ULONG AtlThunkSListPtr32;
    PVOID Reserved9[45];
    BYTE Reserved10[96];
    PPS_POST_PROCESS_INIT_ROUTINE PostProcessInitRoutine;
    BYTE Reserved11[128];
    PVOID Reserved12[1];
    ULONG SessionId;
} PEB, *PPEB;
```

src/windows/WslcSDK/wslcsdk.h
1wizkid: We will figure it out.

src/windows/WslcSDK/wslcsdk.h
JohnMcPMS: Does the public consumer of a PEB create a PEB for the system to read using its private definition?  If anyone reads the fields and they are not aligned because the caller just allocated however it wanted (and it matters to the processor), an exception will occur.

OneBlue: I don't think that's possible for s structure only containing an array of 1 byte elements, but I guess explicitly setting the alignment doesn't hurt. Let's merge as-is and we can revisit later if needed

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+{
+    WSLC_PROCESS_STATE_UNKNOWN = 0,
+    WSLC_PROCESS_STATE_CREATED = 1,
+    WSLC_PROCESS_STATE_RUNNING = 2,
+    WSLC_PROCESS_STATE_EXITED = 3
+} WslcProcessState;
+
+STDAPI WslcProcessGetState(_In_ WslcProcess process, _Out_ WslcProcessState* state);
+
+STDAPI WslcProcessGetExitCode(_In_ WslcProcess process, _Out_ PINT32 exitCode);
```
OneBlue, line 303: nit: int32_t*

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -0,0 +1,443 @@
...
+    WslcContainerImageProgressCallback progressCallback;
+    PVOID progressCallbackContext;
+} WslcLoadImageOptions;
+
+STDAPI WslcSessionImageLoad(_In_ WslcSession session, _In_ const WslcLoadImageOptions* options);
+
+typedef struct WslcImageInfo
+{
+    // we should expose this
+    PCSTR repository;
```
OneBlue, line 379: I thought I already had a comment for that, but I recommend using fixed size arrays here to make it nicer to the caller

OneBlue (approved):
A couple minor outstanding comments, but feel free to fix in a followup PR

src/windows/WslcSDK/wslcsdk.h
yao-msft: nit: INT32* per win32 api design guideline, same for other basic types across this file.

https://www.osgwiki.com/wiki/Win32_API_Design_Guidelines#Basic_Data_Types

PR #14180, opened 2026-02-08
build: ensure wsla deps are declared in build-stage.yml and msi package dependencies

benhillis:


benhillis:
Included these fixes in https://github.com/microsoft/WSL/pull/14185

PR #14186, opened 2026-02-09
Minor follow-up initrd feedback items

benhillis:
Some minor feedback items that came up during the initrd change that I wanted to wait until after merging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@ tmp
 *.user
 *.csproj
 *.vcxproj
-*.targets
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -10,7 +10,6 @@ tmp
 *.user
```
benhillis, line 10: unrelated but I keep getting annoyed that the test-storage directory isn't in the gitignore in the main branch.

OneBlue (approved):

PR #14187, opened 2026-02-09
Add Clang requirement to vswhere usage

JohnMcPMS:
Add a requirement that the VS product have Clang installed when detecting with vswhere.

## PR Checklist

If multiple products in the version range are installed, the invocation of vswhere will output multiple lines with each install path, breaking the rest of the cmake setup.  This change adds a requirement that the product have Clang, which is what the later `if (NOT EXISTS ${LLVM_INSTALL_DIR})` is testing for (although even that check is incorrect as this directory can exist without actually containing clang.exe due to alternate components).

Also adds the `LLVM_INSTALL_DIR` that does not exist to the error message, making it easier to see why it was unhappy.

## Validation Steps Performed
Was able to build with both VS 2022 Enterprise and the VS Build tools products installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,8 +260,9 @@ endif()
 
 # Determine the Visual Studio installation directory which contains LLVM tools
 # N.B. The version is set to VS2022 to ensure local runs match pipeline behavior
+# Require that Clang be installed in the product to potentially de-deduplicate
 execute_process(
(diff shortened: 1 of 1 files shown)


OneBlue (approved):
Nice, thank you for doing this !

benhillis (approved):

PR #14188, opened 2026-02-09
Add experimental image build support

johnstep:
Add experimental image build support via the Docker Engine API using the legacy builder. This allows multi-stage builds, but not parallel builds or build mounts.


src/linux/init/WSLAInit.cpp
```diff
@@ -230,6 +230,25 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_UNIX_CONN
     pollDescriptors[1].events = POLLIN;
 
     std::vector<gsl::byte> relayBuffer;
+
+    auto writeAll = [&](int fd, auto bytesRead) -> bool {
```
OneBlue: nit: We could use `UtilWriteBuffer` for this

src/windows/common/helpers.cpp
```diff
@@ -708,3 +708,81 @@ bool wsl::windows::common::helpers::TryAttachConsole()
...
+
+    wil::unique_hfile stderrRead, stderrWrite;
+    THROW_IF_WIN32_BOOL_FALSE(CreatePipe(&stderrRead, &stderrWrite, &attributes, 0));
+    THROW_IF_WIN32_BOOL_FALSE(SetHandleInformation(stderrRead.get(), HANDLE_FLAG_INHERIT, 0));
+
+    STARTUPINFOW startupInfo{
+        .cb = sizeof(startupInfo), .dwFlags = STARTF_USESTDHANDLES, .hStdOutput = stderrWrite.get(), .hStdError = stderrWrite.get()};
+
+    wil::unique_process_information processInfo;
+    THROW_IF_WIN32_BOOL_FALSE(CreateProcessW(
```
OneBlue: nit: I recommend using the `wsl::windows::common::SubProcess` class for creating child processes.

OneBlue: `RunAndCaptureOutput()` will also capture all the process's output for us without risking a pipe deadlock

src/windows/common/helpers.cpp
```diff
@@ -708,3 +708,81 @@ bool wsl::windows::common::helpers::TryAttachConsole()
 
     return ReopenStdHandles();
 }
+
+void wsl::windows::common::helpers::CreateDockerContextTarArchive(_In_ const std::filesystem::path& sourceDir, _In_ const std::filesystem::path& outputPath)
+{
+    auto absoluteSourceDir = std::filesystem::absolute(sourceDir);
```
OneBlue: nit: It might be a good idea to assign a job object to the child with `JOB_OBJECT_TERMINATE_AT_END_OF_JOB` to make sure that we don't leave the process behind if the parent exits

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -409,6 +409,94 @@ static int Pull(std::wstring_view commandLine)
...
+    std::string dockerfilePath;
+
+    parser.AddPositionalArgument(contextPath, 0);
+    parser.AddArgument(Utf8String{tag}, L"--tag", 't');
+    parser.AddArgument(Utf8String{dockerfilePath}, L"--file", 'f');
+
+    parser.Parse();
+    THROW_HR_IF(E_INVALIDARG, contextPath.empty());
+
+    std::filesystem::path inputPath(contextPath);
```
OneBlue: nit We can use `AbsolutePath` to directly get a std::filesystem::path assigned. 

Example usage: https://github.com/microsoft/WSL/blob/0e263df6f36c71ee8cdb84406620458047316ba8/src/windows/common/WslClient.cpp#L324

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -409,6 +409,94 @@ static int Pull(std::wstring_view commandLine)
...
+    auto tempDir = std::filesystem::temp_directory_path();
+    auto tempTarPath = tempDir / std::format(L"wsla-build-context-{}.tar", GetCurrentProcessId());
+
+    auto cleanupTar = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&] {
+        std::error_code ec;
+        std::filesystem::remove(tempTarPath, ec);
+    });
+
+    wprintf(L"Creating build context...\n");
+    wsl::windows::common::helpers::CreateDockerContextTarArchive(inputPath, tempTarPath);
```
OneBlue: Given that we're creating a file, closing it and then reopening it, instead I would recommend: 

- Creating the tar file HANDLE with exclusive read & write 
- Calling tar with the file handle as stdout
- Keeping the handle opened after tar returns
- Moving the file pointer back to the beginning
- Passing that handle to the service


This guarantees that no one else can modify the file after tar has exited 


Feel free to consider this as non-blocking though (can be done in a followup change)

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -364,15 +364,25 @@ try
...
 
-        auto parsed = wsl::shared::FromJson<docker_schema::CreateImageProgress>(contentString.c_str());
+        std::optional<docker_schema::CreateImageProgress> parsed;
+        try
+        {
+            parsed = wsl::shared::FromJson<docker_schema::CreateImageProgress>(contentString.c_str());
+        }
+        catch (...)
+        {
+            // Ignore JSON parsing errors for pull output - Docker may emit non-JSON lines
```
OneBlue: Out of curiosity, what do the non-json lines look like ?

src/windows/wslaservice/exe/DockerHTTPClient.cpp
```diff
@@ -130,6 +130,26 @@ std::unique_ptr<DockerHTTPClient::HTTPRequestContext> DockerHTTPClient::PullImag
...
+    url.SetParameter("rm", "1");
+    url.SetParameter("forcerm", "1");
+
+    if (Tag.has_value())
+    {
+        url.SetParameter("t", Tag.value());
+    }
+    if (DockerfilePath.has_value())
+    {
+        url.SetParameter("dockerfile", DockerfilePath.value());
```
OneBlue, line 146: From reading the doc it's looking like dockerfile can be omitted if `remote` is specified. Is this something that we're planning to support ?

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -418,6 +428,141 @@ try
...
+
+        std::optional<docker_schema::BuildProgress> parsed;
+        try
+        {
+            parsed = wsl::shared::FromJson<docker_schema::BuildProgress>(contentString.c_str());
+        }
+        catch (...)
+        {
+            // If JSON parsing fails, log but don't display
+            LOG_CAUGHT_EXCEPTION();
```
OneBlue: I think we should fail if we ever hit this (the exception should come up all the way to this method and fail the call)

src/windows/common/helpers.cpp
```diff
@@ -708,3 +708,81 @@ bool wsl::windows::common::helpers::TryAttachConsole()
...
+
+            // Strip trailing directory slash.
+            if (line.back() == '/')
+            {
+                line.pop_back();
+            }
+
+            if (!line.empty())
+            {
+                commandLine += std::format(L" --exclude \"{}\"", std::wstring(line.begin(), line.end()));
```
OneBlue: Ah it's too bad that we do have the parse the Dockerfile ourselves. 

One thing I'll recommend would be to move the parsing logic to its own method so that way we can easily unit test it

OneBlue (commented):
Change looks great ! Couple minor comments

test/windows/WSLATests.cpp
```diff
@@ -477,6 +477,352 @@ class WSLATests
...
+
+        auto contextDir = std::filesystem::current_path() / "build-context-large";
+        std::filesystem::create_directories(contextDir);
+        auto cleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
+            std::error_code ec;
+            std::filesystem::remove_all(contextDir, ec);
+        });
+
+        static constexpr int fileSizeMb = 1024;
+
```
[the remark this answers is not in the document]
johnstep: This is intentional and should only take about 30 seconds. It will be revisited if needed.

src/windows/wslaservice/exe/WSLASession.cpp
johnstep: I made a mistake here. As far as I can tell, it should always be a self-contained JSON blob.

src/windows/wslaservice/exe/DockerHTTPClient.cpp
johnstep: This parameter is only needed if the Dockerfile has a different name or is not at the root of the build context. We should support remote build contexts; the changes will be relatively small but will be done in a future PR with tests.

src/windows/common/helpers.cpp
johnstep: This is a great idea. I will implement it in an additional PR.

src/windows/wsladiag/wsladiag.cpp
johnstep: This is an awesome approach, done!

src/windows/common/helpers.cpp
```diff
@@ -708,3 +708,99 @@ bool wsl::windows::common::helpers::TryAttachConsole()
...
+
+wil::unique_hfile wsl::windows::common::helpers::CreateDockerContextTarArchive(_In_ const std::filesystem::path& sourceDir)
+{
+    auto absoluteSourceDir = std::filesystem::absolute(sourceDir);
+
+    SECURITY_ATTRIBUTES attributes{.nLength = sizeof(attributes), .bInheritHandle = TRUE};
+
+    auto tempPath = std::filesystem::temp_directory_path() / std::format(L"wsla-build-context-{}.tar", GetCurrentProcessId());
+    wil::unique_hfile tarFile{CreateFileW(
+        tempPath.c_str(), GENERIC_READ | GENERIC_WRITE, 0, &attributes, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL | FILE_FLAG_DELETE_ON_CLOSE, nullptr)};
```
OneBlue: nit: We might want to set `FILE_ATTRIBUTE_TEMPORARY` here

src/windows/common/helpers.cpp
```diff
@@ -708,3 +708,99 @@ bool wsl::windows::common::helpers::TryAttachConsole()
...
+
+    wil::unique_hfile stderrRead, stderrWrite;
+    THROW_IF_WIN32_BOOL_FALSE(CreatePipe(&stderrRead, &stderrWrite, &attributes, 0));
+    THROW_IF_WIN32_BOOL_FALSE(SetHandleInformation(stderrRead.get(), HANDLE_FLAG_INHERIT, 0));
+
+    auto workingDir = absoluteSourceDir.wstring();
+    SubProcess process(tarExecutable.c_str(), commandLine.c_str());
+    process.SetWorkingDirectory(workingDir.c_str());
+    process.SetFlags(CREATE_NO_WINDOW);
+    process.SetStdHandles(nullptr, tarFile.get(), stderrWrite.get());
```
OneBlue, line 782: Because we wait for the process to exit before reading from the stderr pipe, this can lead to a pipe deadlock if the child process fills the pipe buffer before exiting. 

I can think of two ways to fix this: 

1) Using a MultiHandleWait to wait for either the process to exit or IO being available on the pipe

2) Modify RunAndCaptureOutput() to have a flag to only read stderr and skip stdout (since we pass a custom handle here) 


This can be done in a followup change though

src/windows/wslaservice/exe/WSLASession.cpp
```diff
@@ -352,6 +352,122 @@ try
...
+        common::relay::HandleWrapper{std::move(contextFileHandle)}, common::relay::HandleWrapper{requestContext->stream.native_handle()}));
+
+    io.AddHandle(std::make_unique<relay::EventHandle>(m_sessionTerminatingEvent.get(), [&]() { THROW_HR(E_ABORT); }));
+
+    io.AddHandle(std::make_unique<DockerHTTPClient::DockerHttpResponseHandle>(
+        *requestContext, std::move(onHttpResponse), std::move(onChunk), std::move(onCompleted)));
+
+    io.Run({});
+
+    THROW_HR_IF(E_ABORT, m_sessionTerminatingEvent.is_signaled());
```
OneBlue: nit: This shouldn't be needed since we already have logic to throw E_ABORT if that event is signaled

OneBlue (approved):
LGTM. Couple minor comments, feel free to merge and fix them in a future PR

src/windows/common/helpers.cpp
johnstep: Sounds good. I will handle this in a future PR.

benhillis (approved):

PR #14189, opened 2026-02-09
Followup changes to address todos in delete and list

yao-msft:
Followup changes to address todos in delete and list

## PR Checklist

## Validation Steps Performed
Added more tests.


src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -122,13 +122,34 @@ struct WSLA_REGISTRY_AUTHENTICATION_INFORMATION
...
+    LONGLONG Created; // Unix timestamp
+    char ParentId[256];
+};
+
+typedef enum _WSLAListImagesFlags
+{
+    WSLAListImagesFlagsNone = 0,
+    WSLAListImagesFlagsAll = 1,           // Show all images (default hides intermediate images)
+    WSLAListImagesFlagsDigests = 2,       // Include digest information
+    WSLAListImagesFlagsDanglingTrue = 4,  // Show only dangling images (untagged)
```
OneBlue, line 133: nit: We could simplify this a bit by having: 

```
WSLAListImagesFlagsShowDangling,
WSLAListImagesFlagsShowNonDangling,
```

And having the caller set those by default, that way we don't have to do any special logic for it in the service

yao-msft: Is this comment just about renaming the flag?

src/windows/wslasession/DockerHTTPClient.cpp
```diff
@@ -160,9 +160,47 @@ void DockerHTTPClient::TagImage(const std::string& Id, const std::string& Repo,
...
+    }
+
+    if (filters.since.has_value())
+    {
+        filtersJson["since"] = nlohmann::json::array({filters.since.value()});
+    }
+
+    if (filters.dangling.has_value())
+    {
+        filtersJson["dangling"] = nlohmann::json::array({filters.dangling.value() ? "true" : "false"});
```
OneBlue, line 190: Does this need to be an array or could we just write the true|false directly ?

src/windows/wslasession/WSLASession.cpp
```diff
@@ -637,37 +637,172 @@ void WSLASession::SaveImageImpl(std::pair<uint32_t, wil::unique_socket>& SocketC
...
+    }
+
+    std::vector<docker_schema::Image> images;
+    try
+    {
+        images = m_dockerClient->ListImages(all, digests, filters);
+    }
+    catch (const DockerHTTPException& e)
+    {
+        std::string errorMessage;
```
OneBlue: This is something that has changed since this PR was created, now we should just do: 

```
    try
    {
        images = m_dockerClient->ListImages(all, digests, filters);
    }
    CATCH_AND_THROW_DOCKER_USER_ERROR("Failed to list images");
```

src/windows/wslasession/WSLASession.cpp
```diff
@@ -637,37 +637,172 @@ void WSLASession::SaveImageImpl(std::pair<uint32_t, wil::unique_socket>& SocketC
...
     for (const auto& e : images)
     {
-        // TODO: download_timestamp
-        for (const auto& tag : e.RepoTags)
+        // Build a map from repo name to digest for this image
+        // RepoDigests format: "repo@sha256:digest"
+        std::map<std::string, std::string> repoToDigest;
+        for (const auto& repoDigest : e.RepoDigests)
+        {
+            size_t atPos = repoDigest.find('@');
```
OneBlue, line 740: Do we expect that the '@' could ever be absent ? If so we should probably throw

test/windows/WSLATests.cpp
```diff
@@ -388,26 +388,316 @@ class WSLATests
...
 
-        // TODO: Add more test coverage once ListImages() is fully implemented.
-
-        // Validate that images with multiple tags are correctly returned.
+        // Setup: Ensure debian:latest is available
         ExpectImagePresent(*m_defaultSession, "debian:latest");
 
-        ExpectCommandResult(m_defaultSession.get(), {"/usr/bin/docker", "tag", "debian:latest", "debian:test-list-images"}, 0);
+        // Create additional tags for testing
+        ExpectCommandResult(m_defaultSession.get(), {"/usr/bin/docker", "tag", "debian:latest", "debian:test-tag1"}, 0);
```
OneBlue: nit: We can use `TagImage()` now

test/windows/WSLATests.cpp
```diff
@@ -388,26 +388,316 @@ class WSLATests
...
+
+            // Test 'since' filter - images created after debian
+            {
+                WSLA_LIST_IMAGES_OPTIONS options{};
+                options.Flags = WSLAListImagesFlagsNone;
+                options.Since = debianId.c_str();
+
+                wil::unique_cotaskmem_array_ptr<WSLA_IMAGE_INFORMATION> images;
+                VERIFY_SUCCEEDED(m_defaultSession->ListImages(&options, images.addressof(), images.size_address<ULONG>()));
+                VERIFY_IS_TRUE(images.size() > 0);
```
OneBlue, line 586: nit: we should probably validate that we see an images that was created after this time, and not debian:latest (assuming the filter is exclusive)

test/windows/WSLATests.cpp
```diff
@@ -388,26 +388,316 @@ class WSLATests
...
+                LogInfo("Images since debian: %zu", images.size());
```
OneBlue: nit: This looks like a debug print, we might want to remove it (and the ones below)

test/windows/WSLATests.cpp
```diff
@@ -388,26 +388,316 @@ class WSLATests
...
+                std::string imageName = image.Image;
+                VERIFY_ARE_EQUAL(imageName, std::string("<none>:<none>"));
+                LogInfo("Dangling image: %hs (hash: %hs)", imageName.c_str(), image.Hash);
+            }
+
+            // List non-dangling images
+            options.Flags = WSLAListImagesFlagsDanglingFalse;
+            wil::unique_cotaskmem_array_ptr<WSLA_IMAGE_INFORMATION> nonDanglingImages;
+            VERIFY_SUCCEEDED(m_defaultSession->ListImages(&options, nonDanglingImages.addressof(), nonDanglingImages.size_address<ULONG>()));
+
```
OneBlue, line 661: nit: We should validate that we received at least one image here

OneBlue (commented):

src/windows/wslasession/DockerHTTPClient.cpp
yao-msft: The api spec says it has to be an array (I guess to be consistent with other filters)

src/windows/wslasession/WSLASession.cpp
yao-msft: Changed to throw

OneBlue (approved):

PR #14191, opened 2026-02-10
Use ISupportErrorInfo to return error messages to the caller

OneBlue:

This change switches to using ISupportErrorInfo to return error strings back to the caller without having to explicitly return the strings through the COM API. 

This change also propagates the source location of where the error was thrown to make debugging easier.

Advantages of this method: 

- Existing macros can be reused without specific handling for WSLA (THROW_HR, THROW_HR_WITH_USER_ERROR)
- The root error printing logic is the same between WSL & WSLA 

This change only implements the new error model in WSLASession. I'll do the other classes in followup changes to make this easier to review 

## PR Checklist


test/windows/WSLATests.cpp
```diff
@@ -542,13 +564,9 @@ class WSLATests
...
             VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, false);
-            WSLA_ERROR_INFO errorInfo{};
-            VERIFY_FAILED(m_defaultSession->SaveImage(HandleToULong(imageTarFileHandle.get()), "hello-wld:latest", nullptr, &errorInfo));
-            VERIFY_IS_NOT_NULL(errorInfo.UserErrorMessage);
-            std::string errMsg = errorInfo.UserErrorMessage;
-            VERIFY_IS_TRUE(errMsg.find("reference does not exist") != std::string::npos);
-            CoTaskMemFree(errorInfo.UserErrorMessage);
-            errorInfo.UserErrorMessage = nullptr;
+            VERIFY_FAILED(m_defaultSession->SaveImage(HandleToULong(imageTarFileHandle.get()), "hello-wld:latest", nullptr));
+            ValidateCOMErrorMessage(L"reference does not exist");
```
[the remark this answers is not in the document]
OneBlue: That's by design. We want to validate the expect error message

src/windows/common/ExecutionContext.cpp
```diff
@@ -362,24 +386,60 @@ ServiceExecutionContext::~ServiceExecutionContext()
     }
 }
 
+COMServiceExecutionContext::COMServiceExecutionContext() : ExecutionContext(Empty)
+{
+}
+
+COMServiceExecutionContext::~COMServiceExecutionContext()
+try
```
benhillis, line 394: Maybe we should have each of these calls log instead of throw? as it currently is if one throws none of the others will run.

benhillis (approved):

src/windows/common/ExecutionContext.cpp
OneBlue: That's pretty much what we would want since anything throwing here would be very unexpected

PR #14192, opened 2026-02-11
wsla cleanup: move source files from wslaservice -> wslasession

benhillis:
This change is some follow-up to the previous change that split wslaservice into two processes. Primarily this is just moving source files to the correct place, but I also noticed some unused link libs, headers, etc.


src/windows/wslaservice/exe/CMakeLists.txt
```diff
@@ -1,18 +1,17 @@
...
-    )
+    WSLASessionManagerFactory.cpp)
 
 set(HEADERS
     WSLASessionManager.h
     WSLASessionManagerFactory.h
     HcsVirtualMachine.h)
 
-include_directories(${CMAKE_SOURCE_DIR}/src/windows/wslaclient)
+include_directories(${CMAKE_SOURCE_DIR}/src/windows/wslasession)
```
benhillis: might not be needed

OneBlue: Yeah I'd expect that this shouldn't be needed, do we need it for specific header except the IDL ones ?

OneBlue (approved):

PR #14196, opened 2026-02-11
Add WSLAContainer flag to auto delete container after exit

kvega005:

Add support for Docker-style auto-remove (Rm) flag for containers
- Persist and restore the WSLAContainerFlagsRm flag in container metadata.
- Automatically delete containers with the Rm flag after they stop, using a detached thread.
- Add tests to verify auto-remove behavior and flag persistence across sessions.

## PR Checklist

## Validation Steps Performed


src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -496,6 +498,14 @@ void WSLAContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCod
         }
 
         m_processes.clear();
+
+        // If the Rm flag is set, delete the container.
+        // This needs to be done in a detached thread because it will trigger deletion of this object.
+        if (WI_IsFlagSet(m_containerFlags, WSLAContainerFlagsRm))
+        {
+            auto comWrapper = m_comWrapper;
+            std::thread([comWrapper]() { LOG_IF_FAILED(comWrapper->Delete()); }).detach();
```
OneBlue: This is definitely tricky because we're indeed inside the object that we want to delete. 

Here's an idea that might help simplify this: 

Instead of registering the event callback in WSLAContainerImpl, we register the callback at the WSLAContainer level, and then we have something like this: 

```
void WSLAContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCode)
{
        if (event == ContainerEvent::Stop)
        {
              auto [lock, impl] = LockImpl();

             impl->OnStopped();

             if (WI_IsFlagSet(impl->Flags(), WSLAContainerFlagsRm))
             {
                  impl->Delete();
                  m_onDeleted(impl);
             }
        }
}
```` 

We'd have to add a OnStopped() & Flags() methods, but that should allow us to completely avoid the "I need to delete myself" problem

Comment (kvega005) [reply src/windows/wslaservice/exe/WSLAContainer.cpp]: Had to do a bit of refactoring to avoid lock ordering issues and potential deadlocks. The OnDeleted callback now returns ownership of the WSLAContainerImpl. This needs to be destroyed outside of the  WSLAContainer lock.

File: src/windows/wslaservice/exe/WSLAContainer.cpp [hunk 2]
```diff
@@ -1071,14 +1062,50 @@ HRESULT WSLAContainer::Inspect(LPSTR* Output)
...
+    {
+        auto [lock, impl] = LockImpl();
+
+        WSL_LOG(
+            "ContainerEvent",
+            TraceLoggingValue(impl->Name().c_str(), "Name"),
+            TraceLoggingValue(impl->ID().c_str(), "Id"),
+            TraceLoggingValue((int)event, "Event"));
+
+        if (event == ContainerEvent::Stop)
```
OneBlue: nit: I'd recommend doing an moving the logging and this check in the beginning of this method so we can skip acquiring the lock if this isn't a stop event

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -339,6 +339,9 @@ WSLAContainerImpl::~WSLAContainerImpl()
         CATCH_LOG();
     }
 
+    // Unmount volumes.
+    UnmountVolumes(m_mountedVolumes, *m_parentVM);
```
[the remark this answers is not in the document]
OneBlue: Agreed with this. Do we need to explicitly unmount in the destructor ? We should only need to do it when it's deleted, which Delete() already does

test/windows/WSLATests.cpp
```diff
@@ -3543,4 +3543,74 @@ class WSLATests
...
+            // A container with the Rm flag is deleted asynchronously after it's stopped.
+            // This retry loop verifies the container is deleted within one second.
+            wsl::shared::retry::RetryWithTimeout<void>(
+                [&]() {
+                    wil::com_ptr<IWSLAContainer> container;
+                    auto hr = m_defaultSession->OpenContainer(containerName, &container);
+                    THROW_HR_IF(E_FAIL, hr != HRESULT_FROM_WIN32(ERROR_NOT_FOUND));
+                },
+                std::chrono::milliseconds{100},
+                std::chrono::seconds{1});
```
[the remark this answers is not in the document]
OneBlue: Agree I'd recommend increasing the timeout to a minute so it doesn't risk failing in CI (which can pretty slow sometimes)

test/windows/WSLATests.cpp
```diff
@@ -3543,4 +3543,74 @@ class WSLATests
...
+        // Test that a container with the Rm flag is automatically deleted on Stop().
+        {
+            WSLAContainerLauncher launcher("debian:latest", "test-auto-remove", {"/bin/cat"}, {}, {}, WSLAProcessFlagsStdin);
+            launcher.SetContainerFlags(WSLAContainerFlagsRm);
+
+            auto container = launcher.Launch(*m_defaultSession);
+            auto process = container.GetInitProcess();
+
+            VERIFY_ARE_EQUAL(container.State(), WslaContainerStateRunning);
+            VERIFY_SUCCEEDED(container.Get().Stop(WSLASignalSIGKILL, 0));
```
OneBlue: nit: It would be nice to have special logic in WSLAContainer() to synchronously perform the container deleting if the rm flag is set, that way the caller is guaranteed that the container is deleted when Stop() returns successfully (feel free to do this in a future change though)

test/windows/WSLATests.cpp
```diff
@@ -3543,4 +3543,74 @@ class WSLATests
...
+                WSLAContainerLauncher launcher("debian:latest", "test-auto-remove", {"/bin/cat"}, {}, {}, WSLAProcessFlagsStdin);
+                launcher.SetContainerFlags(WSLAContainerFlagsRm);
+
+                auto container = launcher.Create(*m_defaultSession);
+                container.SetDeleteOnClose(false);
+
+                ResetTestSession();
+            }
+
+            auto container = OpenContainer(m_defaultSession.get(), "test-auto-remove");
```
OneBlue, line 4262: We might need to add the container flags to `ContainerInfo` for the flags to be correctly re-populated when we recover containers from storage

src/windows/wslaservice/exe/WSLAContainer.cpp
kvega005: I was thinking about this last night. Right now we need to acquire the lock for logging as well. But we can store the name and ID in the com wrapper, so we don't need to call out to the implementation for logging.

src/windows/wslaservice/exe/WSLAContainer.cpp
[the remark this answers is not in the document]
kvega005: We do not. this is a remnant of when I was trying to use the docker auto-remove flag. In that case the container could be deleted without going through `Delete()`, but I ultimately decided against that approach for a few different reasons. I just forgot to remove this.

test/windows/WSLATests.cpp
kvega005: I did think about that earlier and ran into the problem of `impl` deleting itself, but now that we are implementing some of this in the com wrapper, I can revisit.

test/windows/WSLATests.cpp
kvega005: We store the process and container flags in the container metadata. I figured in the future we might have more info in these flags that would be important to recover but are not necessarily important to docker.

src/windows/wslaservice/exe/WSLAContainer.cpp
```diff
@@ -1071,14 +1062,50 @@ HRESULT WSLAContainer::Inspect(LPSTR* Output)
...
 
+void WSLAContainer::OnEvent(ContainerEvent event, std::optional<int> exitCode)
+{
+    // The impl must be destroyed outside COMImplClass::m_lock to avoid a lock ordering
+    // inversion. The destruction chain (~WSLAContainerImpl -> ~ContainerTrackingReference ->
+    // UnregisterContainerStateUpdates) acquires ContainerEventTracker::m_lock, which the IORelay
+    // thread (our caller) may already hold. Destroying under COMImplClass::m_lock would create:
+    //   IORelay: ContainerEventTracker::m_lock -> COMImplClass::m_lock
+    //   Destroy: COMImplClass::m_lock -> ContainerEventTracker::m_lock
+    std::unique_ptr<WSLAContainerImpl> implGuard;
```
OneBlue: Unfortunately I just realized that this isn't going to work because that means that WSLAContainerImpl could outlive WSLASession, which could lead to undefined behavior since we access things like the VM ports in its destructor. 

Here's another approach that I think should work: We can completely remove the concept of WSLAContainer having a callback to WSLASession, and instead we add a `Deleted` state to `WSLA_CONTAINER_STATE`. 

When the container is deleted (either via Delete, or via an event), we moved into the Deleted state, in which nothing can happen to the container. 

Then from WSLASession, whenever when list the containers (either for List() or Open()), we delete any container that's in the Deleted state

To make this clean we'll probably to have something similar to WSLASession with "Terminate()" method that releases all the resources so the container doesn't maintain mounts or port mappings when it's in the deleted state. 

Let me know if that makes sense, and sorry for suggesting an approach that didn't end up working

test/windows/WSLATests.cpp
```diff
@@ -4084,4 +4084,70 @@ class WSLATests
...
+        WSL2_TEST_ONLY();
+        SKIP_TEST_ARM64();
+
+        // Test that a container with the Rm flag is automatically deleted on Stop().
+        {
+            WSLAContainerLauncher launcher("debian:latest", "test-auto-remove", {"/bin/cat"}, {}, {}, WSLAProcessFlagsStdin);
+            launcher.SetContainerFlags(WSLAContainerFlagsRm);
+
+            auto container = launcher.Launch(*m_defaultSession);
+            auto process = container.GetInitProcess();
```
OneBlue: nit: `process` is unused in this scope

test/windows/WSLATests.cpp
```diff
@@ -4084,4 +4084,70 @@ class WSLATests
...
+
+            VERIFY_ARE_EQUAL(container.State(), WslaContainerStateRunning);
+            VERIFY_SUCCEEDED(container.Get().Stop(WSLASignalSIGKILL, 0));
+
+            // verifyContainerDeleted("test-auto-remove");
```
OneBlue: nit: We should remove that comment since it refers to a previous version of the tests

test/windows/WSLATests.cpp
```diff
@@ -4084,4 +4084,70 @@ class WSLATests
...
+            VERIFY_SUCCEEDED(container.Get().Start(WSLAContainerStartFlagsNone));
+            VERIFY_SUCCEEDED(container.Get().Stop(WSLASignalSIGKILL, 0));
+
+            // verifyContainerDeleted("test-auto-remove");
+            VERIFY_ARE_EQUAL(container.Get().Delete(), RPC_E_DISCONNECTED);
+
+            wil::com_ptr<IWSLAContainer> notFound;
+            VERIFY_ARE_EQUAL(m_defaultSession->OpenContainer("test-auto-remove", &notFound), HRESULT_FROM_WIN32(ERROR_NOT_FOUND));
+        }
+    }
```
OneBlue, line 4272: nit: I think it would also be a good idea to add test coverage for the "stop detection" code path. 

To do that we could create a container with the AutoRm flag that just does "echo foo", and validate that after the init process has exited, the container gets deleted (we'll probably have to check with a timeout since that won't be synchronous though, but we can use `wsl::shared::retry::RetryWithTimeout` for that)

OneBlue (approved):
LGTM, minor comments in the tests. 

This turned out to be more complex problem than we thought. Thank you for doing this !

PR #14197, opened 2026-02-11
Tag Image

yao-msft:
## PR Checklist


test/windows/WSLATests.cpp
```diff
@@ -827,6 +827,184 @@ class WSLATests
...
+                {
+                    latestHash = image.Hash;
+                }
+                else if (std::strcmp(image.Image, "debian:test-tag") == 0)
+                {
+                    testTagHash = image.Hash;
+                }
+            }
+
+            VERIFY_IS_FALSE(latestHash.empty());
```
OneBlue, line 884: nit: We should put this in  wil::scope_exit_log() so it's cleaned up even if the test fails

OneBlue: (same for below cleanups)

test/windows/WSLATests.cpp
```diff
@@ -827,6 +827,184 @@ class WSLATests
...
+        // Negative test: Tag a non-existent image.
+        {
+            WSLA_TAG_IMAGE_OPTIONS options{};
+            options.Image = "nonexistent:notfound";
+            options.Repo = "test";
+            options.Tag = "fail";
+
+            HRESULT hr = m_defaultSession->TagImage(&options);
+
+            VERIFY_ARE_EQUAL(WSLA_E_IMAGE_NOT_FOUND, hr);
```
OneBlue: nit: We should also validate the error message (or lack of)

test/windows/WSLATests.cpp
```diff
@@ -827,6 +827,184 @@ class WSLATests
...
+        {
+            WSLA_TAG_IMAGE_OPTIONS options{};
+            options.Image = "debian:latest";
+            options.Repo = "test";
+            options.Tag = "invalid tag";
+
+            HRESULT hr = m_defaultSession->TagImage(&options);
+
+            VERIFY_ARE_EQUAL(HRESULT_FROM_WIN32(ERROR_BAD_ARGUMENTS), hr);
+        }
```
OneBlue, line 996: I recommend adding coverage for trying to overwrite a tag that already exists

test/windows/WSLATests.cpp
```diff
@@ -827,6 +827,184 @@ class WSLATests
...
 
+    TEST_METHOD(TagImage)
+    {
+        WSL2_TEST_ONLY();
+
+        // Positive test: Tag an existing image with a new tag in the same repository.
+        {
+            ExpectImagePresent(*m_defaultSession, "debian:latest");
+
+            WSLA_TAG_IMAGE_OPTIONS options{};
```
OneBlue: nit: To make this a bit more concise, we could have lambda like this: 

```
auto expectTagResult = [&](const char* Image, const char* Repo, const char* Tag, HRESULT expectedResult)
{
[...]
}
```
That way the tests can just do: 

```
expectTagResult("debian:latest, "my-repo", "my-tag", S_OK);
```

test/windows/WSLATests.cpp
OneBlue: (Same for other negative tests)

OneBlue (commented):
LGTM, minor comments

test/windows/WSLATests.cpp
```diff
@@ -826,6 +826,176 @@ class WSLATests
...
+
+                wil::unique_cotaskmem_array_ptr<WSLA_DELETED_IMAGE_INFORMATION> deletedImages;
+                VERIFY_SUCCEEDED(
+                    m_defaultSession->DeleteImage(&deleteOptions, deletedImages.addressof(), deletedImages.size_address<ULONG>()));
+
+                ExpectImagePresent(*m_defaultSession, "test:duplicate-tag", false);
+            });
+
+            VERIFY_SUCCEEDED(runTagImage("debian:latest", "test", "duplicate-tag"));
+            VERIFY_SUCCEEDED(runTagImage("debian:latest", "test", "duplicate-tag"));
```
OneBlue, line 963: Does do actually do anything or is it a no-op since the underlying image is the same ?

OneBlue: If so, we might want to add a test that overwrites the tag with another image, so we can validate that the tag operation actually happened

OneBlue (approved):
LGTM, one minor test comment

PR #14198, opened 2026-02-12
Improved VirtioProxy networking and VirtioFS posix compatibility

benhillis:
This change updates WSL to use a new version of the Microsoft.WSL.DeviceHost package (version 1.1.14-0) which contains the following changes:
1. DNS tunneling support for virtio proxy networking
3. UDP timeout support for virtio proxy networking
4. Improvements to virtiofs posix compatibility

### DIFF
--- a/src/windows/common/VirtioNetworking.cpp
+++ b/src/windows/common/VirtioNetworking.cpp
@@ -11,14 +11,15 @@ using namespace wsl::shared;
 using namespace wsl::windows::common::stringify;
 using wsl::core::VirtioNetworking;
 
+static constexpr auto c_eth0DeviceName = L"eth0";
 static constexpr auto c_loopbackDeviceName = TEXT(LX_INIT_LOOPBACK_DEVICE_NAME);
 
 VirtioNetworking::VirtioNetworking(
-    GnsChannel&& gnsChannel, bool enableLocalhostRelay, LPCWSTR dnsOptions, std::shared_ptr<GuestDeviceManager> guestDeviceManager, wil::shared_handle userToken) :
+    GnsChannel&& gnsChannel, VirtioNetworkingFlags flags, LPCWSTR dnsOptions, std::shared_ptr<GuestDeviceManager> guestDeviceManager, wil::shared_handle userToken) :
     m_guestDeviceManager(std::move(guestDeviceManager)),
     m_userToken(std::move(userToken)),
     m_gnsChannel(std::move(gnsChannel)),
-    m_enableLocalhostRelay(enableLocalhostRelay),
+    m_flags(flags),
     m_dnsOptions(dnsOptions)
 {
 }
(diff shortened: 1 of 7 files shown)


OneBlue (approved):

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -576,8 +576,10 @@ void WslCoreVm::Initialize(const GUID& VmId, const wil::shared_handle& UserToken
             }
             else if (m_vmConfig.NetworkingMode == NetworkingMode::VirtioProxy)
             {
+                wsl::core::VirtioNetworkingFlags flags = wsl::core::VirtioNetworkingFlags::None;
+                WI_SetFlagIf(flags, wsl::core::VirtioNetworkingFlags::LocalhostRelay, m_vmConfig.EnableLocalhostRelay);
```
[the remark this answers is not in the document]
benhillis: this is intentional. there is currently an issue with dns over tcp, needs a new devicehost.

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -1929,7 +1931,9 @@ bool WslCoreVm::InitializeDrvFsLockHeld(_In_ HANDLE UserToken)
 
 bool WslCoreVm::IsDnsTunnelingSupported() const
 {
-    WI_ASSERT(m_vmConfig.NetworkingMode == NetworkingMode::Nat || m_vmConfig.NetworkingMode == NetworkingMode::Mirrored);
+    WI_ASSERT(
+        m_vmConfig.NetworkingMode == NetworkingMode::Nat || m_vmConfig.NetworkingMode == NetworkingMode::Mirrored ||
+        m_vmConfig.NetworkingMode == NetworkingMode::VirtioProxy);
```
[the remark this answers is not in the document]
benhillis: This is also intentional, this same IsDnsTunnelingSupported check will be used when the dns over tcp issue is fixed.

src/windows/common/VirtioNetworking.cpp
```diff
@@ -251,57 +178,128 @@ int VirtioNetworking::ModifyOpenPorts(_In_ PCWSTR tag, _In_ const SOCKADDR_INET&
...
 {
-    auto lock = m_lock.lock_exclusive();
-    UpdateMtu();
+    // Query current networking information before acquiring the lock.
+    m_networkSettings = GetHostEndpointSettings();
 
-    // Check for DNS changes and send update if needed.
-    auto currentDns = networking::HostDnsInfo::GetDnsSettings(networking::DnsSettingsFlags::IncludeVpn);
-    if (currentDns != m_trackedDnsSettings)
+    // TODO: Determine gateway MAC address
```
OneBlue, line 190: What do we need the gateway MAC address for here ?

src/windows/common/VirtioNetworking.cpp
```diff
@@ -251,57 +178,128 @@ int VirtioNetworking::ModifyOpenPorts(_In_ PCWSTR tag, _In_ const SOCKADDR_INET&
...
+    auto client_ip = m_networkSettings->PreferredIpAddress.AddressString;
+    if (!client_ip.empty())
     {
-        m_trackedDnsSettings = currentDns;
-        SendDnsUpdate(currentDns);
+        if (device_options.tellp() > 0)
+        {
+            device_options << L";";
+        }
+        device_options << L"client_ip=" << client_ip;
```
OneBlue, line 199: Does this IP address represent the IP address that the client process will use when sending traffic on the behalf of the VM ? 

If so, would setting it bypass routes, for instance if there's a specific VPN route that uses a different source ip address ?

src/windows/common/VirtioNetworking.cpp
```diff
@@ -251,57 +178,128 @@ int VirtioNetworking::ModifyOpenPorts(_In_ PCWSTR tag, _In_ const SOCKADDR_INET&
...
+    }
+
+    std::wstring default_route = m_networkSettings->GetBestGatewayAddressString();
+    if (!default_route.empty())
+    {
+        if (device_options.tellp() > 0)
+        {
+            device_options << L";";
+        }
+        device_options << L"gateway_ip=" << default_route;
```
OneBlue, line 218: Same question for the gateway, is this a catch-all gateway that will bypass the windows routing table ?

src/windows/common/VirtioNetworking.cpp
```diff
@@ -251,57 +178,128 @@ int VirtioNetworking::ModifyOpenPorts(_In_ PCWSTR tag, _In_ const SOCKADDR_INET&
...
+        {
+            m_adapterId = m_guestDeviceManager->AddGuestDevice(
+                VIRTIO_NET_DEVICE_ID, VIRTIO_NET_CLASS_ID, c_eth0DeviceName, nullptr, newDeviceOptions.c_str(), 0, m_userToken.get());
+        }
+        else
+        {
+            const auto server = m_guestDeviceManager->GetRemoteFileSystem(VIRTIO_NET_CLASS_ID, c_defaultDeviceTag);
+            if (server)
+            {
+                LOG_IF_FAILED(server->AddSharePath(c_eth0DeviceName, newDeviceOptions.c_str(), 0));
```
OneBlue, line 252: I'm guessing AddSharePath() is how we change the device settings ?

src/windows/common/VirtioNetworking.cpp
benhillis: I don't know, this is an existing comment.

src/windows/common/VirtioNetworking.cpp
benhillis: This sets the IP address of the guest adapter to match the ip address of the default adapter on the host. in the future we could technically implement a mirrored virtio net where multiple adapters exist in the vm.

src/windows/common/VirtioNetworking.cpp
benhillis: no, this is used for is telling virtio net what address it should intercept DNS packets for. those then use the same apis that dns tunnling use.

src/windows/common/VirtioNetworking.cpp
benhillis: yeah, same way we configure the port binds.

benhillis: Brian overloaded that API to do a bunch of stuff.

OneBlue (approved):

PR #14207, opened 2026-02-12
test: introduce a separate test runner for WSLA tests

benhillis:
This should speed up validation by allowing the WSLA tests to run in parallel with the WSL1 and WSL2 tests.


test/windows/Common.cpp
```diff
@@ -58,7 +58,8 @@ static HANDLE g_OriginalStderr;
 static BOOL g_RelogEverything = TRUE;
 static bool g_LogDmesgAfterEachTest = false;
 static PTP_TIMER g_WatchdogTimer;
-static BOOL g_VmMode;
+static BOOL g_VmMode = FALSE;
```
OneBlue: nit: I'd have a small preference for a tri-state enum here

PR #14208, opened 2026-02-12
Fix potential long delay in wslaservice termination

benhillis:
This change resolves a potential long delay when stopping the wslaservice. Previously, the GetState method would acquire the session lock, which could be locked behind a long-running operation. Instead, check the signal status of the terminating event which can be safely done without the lock held.

### DIFF
--- a/src/windows/wslasession/WSLASession.cpp
+++ b/src/windows/wslasession/WSLASession.cpp
@@ -1012,6 +1012,7 @@ try
         m_virtualMachine.reset();
     }
 
+    m_terminated = true;
     return S_OK;
 }
 CATCH_RETURN();
@@ -1079,7 +1080,7 @@ void WSLASession::OnContainerDeleted(const WSLAContainerImpl* Container)
 
 HRESULT WSLASession::GetState(_Out_ WSLASessionState* State)
 {
-    *State = m_sessionTerminatingEvent.is_signaled() ? WSLASessionStateTerminated : WSLASessionStateRunning;
+    *State = m_terminated ? WSLASessionStateTerminated : WSLASessionStateRunning;
     return S_OK;
 }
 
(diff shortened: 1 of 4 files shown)


src/windows/wslasession/WSLASession.cpp
```diff
@@ -315,14 +315,12 @@ try
 
     auto onCompleted = [&]() { io.Cancel(); };
 
-    io.AddHandle(std::make_unique<relay::EventHandle>(m_sessionTerminatingEvent.get(), [&]() { THROW_HR(E_ABORT); }));
```
benhillis, line 318: duplicated code, not needed.

src/windows/wslasession/WSLASession.cpp
```diff
@@ -315,14 +315,12 @@ try
...
     io.AddHandle(std::make_unique<relay::EventHandle>(m_sessionTerminatingEvent.get(), [&]() { THROW_HR(E_ABORT); }));
     io.AddHandle(std::make_unique<DockerHTTPClient::DockerHttpResponseHandle>(
         *requestContext, std::move(onHttpResponse), std::move(onChunk), std::move(onCompleted)));
 
     io.Run({});
 
-    THROW_HR_IF(E_ABORT, m_sessionTerminatingEvent.is_signaled());
```
benhillis, line 325: This would never be true, if this was signaled the lambda above would have fired and thrown.

src/windows/wslasession/WSLASession.cpp
```diff
@@ -1081,8 +1079,7 @@ void WSLASession::OnContainerDeleted(const WSLAContainerImpl* Container)
 
 HRESULT WSLASession::GetState(_Out_ WSLASessionState* State)
 {
-    std::lock_guard lock{m_lock};
-    *State = m_virtualMachine ? WSLASessionStateRunning : WSLASessionStateTerminated;
+    *State = m_sessionTerminatingEvent.is_signaled() ? WSLASessionStateTerminated : WSLASessionStateRunning;
```
OneBlue: Unfortunately that might cause a race condition with the storage VHD. 

If we consider the session terminated as soon as this event is set, that would mean that the session manager would allow another sesssion with the same name to be created before the storage folder is unmounted, which could cause CreateSession() failures

benhillis: good call.

OneBlue (approved):

PR #14211, opened 2026-02-12
Partial SDK implementation

JohnMcPMS:
Implements enough of the SDK for the sample demo to operate.

## PR Checklist

Implements the functions called by the sample demo (as well as a few more trivial functions).  These are implemented as best as possible, but there are several TODO callouts for updates that are likely needed to the SDK API surface to account for disparities with the runtime API.  I intentionally did not make those changes to the SDK in this PR to limit the friction so that it might get merged sooner.  I plan to make a follow-up PR with those changes that can have more/longer discussions.

Additionally, tests are not added for the SDK with this PR at this time.  Again, a fast follow-up PR is expected to start adding those tests.  Recommendation for new or existing test project are welcome.

Finally, I left the changes that were merged in #14187 to discuss how mainline changes should make it to the feature branch (as I'm dependent on this change for my machine setup to build WSL).  Happy to remove it late in the PR process so I can keep building without constant git juggling.

## Validation Steps Performed
Sample demo runs successfully and can be run multiple times.


src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -15,64 +15,228 @@ Module Name:
...
 
-    // demo test code to show how to cast the opaque struct to the internal struct and set some values. The real implementation would have actual logic here.
-    // WSLC_SESSION_OPTIONS_INTERNAL* sessionSettingsinternal = (WSLC_SESSION_OPTIONS_INTERNAL*)sessionSettings;
-    sessionSettings = NULL;
-    UNREFERENCED_PARAMETER(storagePath);
-    // UNREFERENCED_PARAMETER(sessionSettings);
-    return E_NOTIMPL;
+    *internalType = {};
+
+    internalType->storagePath = storagePath;
```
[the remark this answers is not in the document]
JohnMcPMS: Except null means "use `tmpfs`".

src/windows/WslcSDK/WslcsdkPrivate.h
```diff
@@ -76,24 +80,50 @@ typedef struct WSLC_CONTAINER_OPTIONS_INTERNAL
...
+#define WSLC_GET_INTERNAL_TYPE(_input_) WSLC_GET_INTERNAL_TYPE_NAMED(_input_, internalType)
+
+// Returns an error on null input then converts to a unique_ptr local named `internalType`.
+// Use for Release functions to clean up the implementation object on return.
+#define WSLC_GET_INTERNAL_TYPE_FOR_RELEASE(_input_) \
+    RETURN_HR_IF_NULL(E_POINTER, _input_); \
+    std::unique_ptr<std::remove_pointer_t<decltype(GetInternalType(_input_))>> internalType \
+    { \
+        GetInternalType(_input_) \
+    }\
```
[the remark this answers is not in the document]
JohnMcPMS: Obviously the suggestion is short sighted, but leaving this open for WSL maintainers feedback on style.  I intentionally leave off the final semicolon as I always want one at the invocation site.

src/windows/WslcSDK/ProgressCallback.cpp
```diff
@@ -0,0 +1,74 @@
...
+    if (m_callback)
+    {
+        WslcImageProgressMessage message{};
+
+        message.id = Id;
+        message.status = ConvertStatus(Status);
+        message.detail.current = Current;
+        message.detail.total = Total;
+
+        m_callback(&message, m_context);
```
OneBlue: We should return the caller's hresult back to the service here

src/windows/WslcSDK/ProgressCallback.h
```diff
@@ -0,0 +1,32 @@
...
+
+    // IProgressCallback
+    HRESULT STDMETHODCALLTYPE OnProgress(LPCSTR Status, LPCSTR Id, ULONGLONG Current, ULONGLONG Total) override;
+
+    // Creates a ProgressCallback if the options provides a callback.
+    static winrt::com_ptr<ProgressCallback> CreateIf(const WslcPullImageOptions* options);
+
+private:
+    WslcContainerImageProgressCallback m_callback;
+    PVOID m_context;
```
OneBlue: nit: I recommend always default initializing class fields (this has bitten us in the past):

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -15,64 +15,228 @@ Module Name:
...
+// Default to 1 GB
+constexpr UINT64 s_DefaultStorageSize = 1000 * 1000 * 1000;
+
+WSLAFeatureFlags ConvertFlags(WslcSessionFlags flags)
+{
+    WSLAFeatureFlags result = WslaFeatureFlagsNone;
+
+    if (WI_IsFlagSet(flags, WSLC_SESSION_FLAG_ENABLE_GPU))
+    {
+        result |= WslaFeatureFlagsGPU;
```
OneBlue: Instead of manually converting the flags, I recommend static_asserting that the values are the same as the COM API, and casting. 

The COM API flags value are guaranteed to be constant across versions

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -103,78 +276,164 @@ STDAPI WslcContainerSettingsSetDomainName(_In_ WslcContainerSettings* containerS
...
+
+    const WSLC_CONTAINER_PROCESS_OPTIONS_INTERNAL* initProcessOptions = internalContainerSettings->initProcessOptions;
+    if (initProcessOptions)
+    {
+        containerOptions.InitProcessOptions.CurrentDirectory = initProcessOptions->currentDirectory;
+        // TODO: Runtime needs an update to take in LPCSTR const*
+        containerOptions.InitProcessOptions.CommandLine.Values = const_cast<LPCSTR*>(initProcessOptions->commandLine);
+        containerOptions.InitProcessOptions.CommandLine.Count = initProcessOptions->commandLineCount;
+        containerOptions.InitProcessOptions.Environment.Values = const_cast<LPCSTR*>(initProcessOptions->environment);
+        containerOptions.InitProcessOptions.Environment.Count = initProcessOptions->environmentCount;
```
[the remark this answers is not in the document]
JohnMcPMS: Leaving open for discussion from WSL maintainers on updating runtime API.

src/windows/WslcSDK/WslcsdkPrivate.h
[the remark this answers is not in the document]
OneBlue: Having the variable name hidden in the macro feels a bit difficult to read. Instead could we something like:

```
template <typename T>
auto GetInternalType(T* ptr)
{
    THROW_HR_IF_NULL(ptr);

    return std::unique_ptr<std::remove_pointer_t<decltype(GetInternalType(ptr))>>{GetInternalType(Ptr)};
}
```

That way the caller explicitly assigns the internal pointer: 

```
auto internal= GetInternalType(...);
```

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -15,64 +15,228 @@ Module Name:
...
+    wil::com_ptr<IWSLASessionManager> sessionManager;
+    RETURN_IF_FAILED(CoCreateInstance(__uuidof(WSLASessionManager), nullptr, CLSCTX_LOCAL_SERVER, IID_PPV_ARGS(&sessionManager)));
+    wsl::windows::common::security::ConfigureForCOMImpersonation(sessionManager.get());
+
+    auto result = std::make_unique<WslcSessionImpl>();
+    WSLA_SESSION_SETTINGS runtimeSettings{};
+    runtimeSettings.DisplayName = internalType->displayName;
+    runtimeSettings.StoragePath = internalType->storagePath;
+    // TODO: Is this the intended use for vhdRequirements.sizeInBytes?
+    runtimeSettings.MaximumStorageSizeMb = internalType->vhdRequirements.sizeInBytes / (1000 * 1000);
```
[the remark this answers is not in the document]
OneBlue: I recommend dividing by `_1MB` here instead

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -103,78 +276,164 @@ STDAPI WslcContainerSettingsSetDomainName(_In_ WslcContainerSettings* containerS
...
 
 STDAPI WslcSessionSettingsSetTerminateCallback(
     _In_ WslcSessionSettings* sessionSettings, _In_opt_ WslcSessionTerminationCallback terminationCallback, _In_opt_ PVOID terminationContext)
 {
-    UNREFERENCED_PARAMETER(terminationCallback);
-    UNREFERENCED_PARAMETER(terminationContext);
-    UNREFERENCED_PARAMETER(sessionSettings);
-    return E_NOTIMPL;
+    WSLC_GET_INTERNAL_TYPE(sessionSettings);
+
```
OneBlue, line 301: nit: We should fail with E_INVALIDARG if the callback is non null and the context is non-null

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -103,78 +276,164 @@ STDAPI WslcContainerSettingsSetDomainName(_In_ WslcContainerSettings* containerS
...
 
 STDAPI WslcContainerInitSettings(_In_ PCSTR imageName, _Out_ WslcContainerSettings* containerSettings)
 {
-    UNREFERENCED_PARAMETER(imageName);
-    UNREFERENCED_PARAMETER(containerSettings);
-    return E_NOTIMPL;
+    WSLC_GET_INTERNAL_TYPE(containerSettings);
+    RETURN_HR_IF_NULL(E_POINTER, imageName);
+
+    *internalType = {};
```
OneBlue, line 349: This feels a bit dangerous because `containerSettings` is uninitialized memory, so if `internalType` happened to have a field with a destructor, like an std::string , this could lead to undefined behavior. 

To fix this we could use an in-place constructor on `internalType`, which will guarantee that the object will only be constructed, but that nothing will be destructed

src/windows/WslcSDK/WslcsdkPrivate.h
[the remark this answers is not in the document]
OneBlue: That would also have the benefit of removing the need for the *_NAMED version

CMakeLists.txt
```diff
@@ -275,8 +275,9 @@ endif()
 
 # Determine the Visual Studio installation directory which contains LLVM tools
 # N.B. The version is set to VS2022 to ensure local runs match pipeline behavior
+# Require that Clang be installed in the product to potentially de-deduplicate
 execute_process(
-    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -version "[17.0,18.0)" -products * -property installationPath
+    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -version "[17.0,18.0)" -products * -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath
```
OneBlue: FYI this should be brought to this branch via https://github.com/microsoft/WSL/pull/14212

OneBlue (commented):
Change looks good, mostly minor comments

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -15,64 +15,228 @@ Module Name:
...
 #include "wslcsdk.h"
 #include "WslcsdkPrivate.h"
+#include "ProgressCallback.h"
+#include "TerminationCallback.h"
+#include "wslutil.h"
+
+
+namespace {
+constexpr uint32_t s_DefaultCPUCount = 2;
+constexpr uint32_t s_DefaultMemoryMB = 2000;
```
1wizkid: Should this be a bit larger as we start out CPU=4 and Memory = 4000? Also, will we ultimately need to compute these values using machine configuration information for example, or just hard code initial settings?

src/windows/WslcSDK/ProgressCallback.cpp
JohnMcPMS: The SDK function is void return.  Do you expect you would take action based on the return from a progress callback (maybe a cheap cancellation mechanism)?

benhillis (dismissed):
Looks like your PR from your fork isn't triggering checks. I'm not sure why, in the meantime you might want to push a user branch to the main WSL repo

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: In this and other cases they are not the same right now.  Obviously, that could be changed.

To the root of the feedback, I don't want to expose runtime options that are not exposed in the SDK.  Having a mask for that is an alternative though.

If that is the preferred route I can update both.

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: I think this comment is not quite right.  `RETURN_HR_IF(E_INVALIDARG, callback != null && context != null)`?

My assumption is that you meant `RETURN_HR_IF(E_INVALIDARG, callback != null && context == null)`, but I don't think that it should be a requirement to provide a context.

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: The caller is already allocating the memory for these settings without any ability to destroy them.  This setup requires no destructors are allowed.  I added a `static_assert(std::is_trivial_v<T>, ...);` for all of the internal structs.

src/windows/WslcSDK/wslcsdk.cpp
OneBlue: That's a good point. I'd have a preference for a mask, since that would it easy to see which flag we allow and won't require manually mapping everything

src/windows/WslcSDK/wslcsdk.cpp
OneBlue: Sorry what I actually meant was "We should fail if the callback is null and the context is non-null", so: 

```
RETURN_HR_IF(E_INVALIDARG, callback == null && context != null)
```

Because if the caller did that, `context` would never be used, so that would most likely be a mistake

src/windows/WslcSDK/ProgressCallback.cpp
OneBlue: Ah I didn't catch that in the SDK PR. 

I don't expect the SDK to do anything about it, but the service will cancel the operation the callback returns an error

src/windows/WslcSDK/wslcsdk.cpp
OneBlue: Ok asserting that the structure we manually are trivial should be a good mitigation for this

benhillis:
> Looks like your PR from your fork isn't triggering checks. I'm not sure why, in the meantime you might want to push a user branch to the main WSL repo

seems to be working now.

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -15,64 +15,213 @@ Module Name:
...
+        {
+            *errorMessage = wil::make_unique_string<wil::unique_cotaskmem_string>(errorInfo->Message.get()).release();
+        }
+    }
+}
+} // namespace
 
 // SESSION DEFINITIONS
 STDAPI WslcSessionInitSettings(_In_ PCWSTR storagePath, _Out_ WslcSessionSettings* sessionSettings)
 {
```
OneBlue, line 89: Given that WSL's codebase is heavily exception based, I would recommend having a try / CATCH_RETURN() for each API entrypoint. 

Doing this would also allow us to move `WSLC_GET_INTERNAL_TYPE` into a method, since it can then threw if the pointer is null

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -234,31 +488,46 @@ STDAPI WslcContainerGetState(_In_ WslcContainer container, _Out_ WslcContainerSt
...
-    UNREFERENCED_PARAMETER(container);
-    UNREFERENCED_PARAMETER(signal);
-    UNREFERENCED_PARAMETER(timeoutMS);
-    return E_NOTIMPL;
+    auto internalType = WSLC_GET_INTERNAL_TYPE(container);
+    RETURN_HR_IF_NULL(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), internalType->container);
+
+    auto convertedSignal = ConvertSignal(signal);
+    RETURN_HR_IF(E_INVALIDARG, !convertedSignal);
```
OneBlue: nit: once we have exception handlers for every API calls, ConvertSignal can directly throw the E_INVALIDARG

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -73,9 +73,11 @@ typedef struct WslcVhdRequirements
 typedef enum WslcSessionFlags
 {
     WSLC_SESSION_FLAG_NONE = 0x00000000,
-    WSLC_SESSION_FLAG_ENABLE_GPU = 0x00000001
+    WSLC_SESSION_FLAG_ENABLE_GPU = 0x00000004
 } WslcSessionFlags;
 
+DEFINE_ENUM_FLAG_OPERATORS(WslcSessionFlags);
```
OneBlue, line 79: This should be under an `#ifdef __cplusplus` so we don't break C compilers here 

(As a TODO should probably have this header included by a .c file to validate that we're compatible with C)

src/windows/WslcSDK/WslcsdkPrivate.h
```diff
@@ -76,24 +85,47 @@ typedef struct WSLC_CONTAINER_OPTIONS_INTERNAL
...
+
 struct WslcProcessImpl
 {
     wil::com_ptr<IWSLAProcess> process;
 };
+
+WslcProcessImpl* GetInternalType(WslcProcess handle);
+
+// Converts to the internal type and returns an error on null input.
+#define WSLC_GET_INTERNAL_TYPE(_input_) \
```
OneBlue: Following up on the previous comment about exception, I'd have a preference to make this a templated method so it's bit easier to follow what's going on

OneBlue (commented):
Change LGTM, couple minor comments

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: I think we have a schism on exception use, probably better served over voice at a later time.

But I can agree with these cases well enough 😉

src/windows/WslcSDK/wslcsdk.h
JohnMcPMS: The definition of this macro is guarded already to do nothing outside of `#ifdef __cplusplus`.

OneBlue (approved):

PR #14218, opened 2026-02-13
Switch to BuildKit via the Docker client

johnstep:
Replace the tar-based Docker Engine API image build with BuildKit by mounting the context directory into the VM and invoking docker build directly with BuildKit enabled.


src/windows/wslasession/WSLASession.cpp
```diff
@@ -350,116 +350,115 @@ try
...
 
-    io.AddHandle(std::make_unique<relay::RelayHandle<relay::ReadHandle>>(
-        common::relay::HandleWrapper{std::move(contextFileHandle)}, common::relay::HandleWrapper{requestContext->stream.native_handle()}));
+    io.AddHandle(std::make_unique<relay::LineBasedReadHandle>(buildProcess.GetStdHandle(2), captureOutput, false));
 
     io.AddHandle(std::make_unique<relay::EventHandle>(m_sessionTerminatingEvent.get(), [&]() { THROW_HR(E_ABORT); }));
 
-    io.AddHandle(std::make_unique<DockerHTTPClient::DockerHttpResponseHandle>(
-        *requestContext, std::move(onHttpResponse), std::move(onChunk), std::move(onCompleted)));
+    io.AddHandle(std::make_unique<relay::EventHandle>(buildProcess.GetExitEvent(), [&]() { io.Cancel(); }));
```
OneBlue: I wouldn't recommend cancelling on the exit event since the exit event might be called before the std handles are closed. I think just waiting on the ExitEvent() before calling GetExitCode() would be right approach here

src/windows/wslasession/WSLASession.cpp
```diff
@@ -350,116 +350,115 @@ try
...
 CATCH_RETURN();
 
-HRESULT WSLASession::BuildImage(ULONG ContextHandle, ULONGLONG ContentLength, LPCSTR DockerfilePath, LPCSTR ImageTag, IProgressCallback* ProgressCallback)
+HRESULT WSLASession::BuildImage(LPCWSTR ContextPath, LPCSTR DockerfilePath, LPCSTR ImageTag, IProgressCallback* ProgressCallback)
 try
 {
     COMServiceExecutionContext context;
 
-    RETURN_HR_IF(E_INVALIDARG, ContextHandle == 0);
+    RETURN_HR_IF_NULL(E_POINTER, ContextPath);
```
OneBlue, line 358: nit: We might want to validate that both `ContextPath` is an absolute path, otherwise that could lead to tricky behavior

src/windows/wslasession/WSLASession.cpp
```diff
@@ -350,116 +350,115 @@ try
...
-    std::optional<std::string> dockerfilePath;
-    if (DockerfilePath != nullptr && *DockerfilePath != '\0')
-    {
-        dockerfilePath = std::string(DockerfilePath);
-    }
+    std::lock_guard lock{m_lock};
+
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_virtualMachine);
+
+    static constexpr char mountPath[] = "/tmp/wsla-build-context";
```
OneBlue: I don't recommend using a hardcoded path for this. Right now I see that we're holding the lock for the entire BuildImageCall(), but long term we should probably not do that and support multiple builds in parallel. 

For this context path I would recommend doing something similar to [this](https://github.com/microsoft/WSL/blob/2a4d75635eb30d1b86f475c600e74cbc79d752c0/src/windows/wslasession/WSLAContainer.cpp#L787C7-L790C103)

src/windows/wslasession/WSLASession.cpp
```diff
@@ -350,116 +350,115 @@ try
...
     if (ImageTag != nullptr && *ImageTag != '\0')
     {
-        tag = std::string(ImageTag);
+        buildArgs.push_back("-t");
+        buildArgs.push_back(ImageTag);
     }
+    if (DockerfilePath != nullptr && *DockerfilePath != '\0')
+    {
+        buildArgs.push_back("-f");
+        buildArgs.push_back(std::string(mountPath) + "/" + std::string(DockerfilePath));
```
OneBlue: If `DockerFilePath` is expect to be just a filename, we should add validation for that, otherwise the caller might be able to pass something like:

```
../../../escape-to-root
```

or 

```
/dev/urandom
```

Which could lead to some weird behaviors

src/windows/wslasession/WSLASession.cpp
```diff
@@ -350,116 +350,115 @@ try
...
+            if (!vertex.error.empty() && reportedErrors.insert(vertex.digest).second)
+            {
+                allOutput.append(vertex.error).append("\n");
+                reportProgress(vertex.error + "\n");
+            }
         }
     };
 
-    auto onCompleted = [&]() { io.Cancel(); };
+    // Drain stdout to prevent blocking; --progress=rawjson writes progress to stderr.
```
OneBlue: What kind of content do we get on stdout ? Error messages ? Trying to follow what `allOutput` will look like on a success & failure cases

src/windows/wslasession/WSLASession.cpp
johnstep: The client converts it to an absolute path first, and `MountWindowsFolderImpl` returns an error if it is not.

src/windows/wslasession/WSLASession.cpp
johnstep: I updated the comment to explain; there is no stdout content on error, and the image ID on success.

src/windows/wslasession/WSLASession.cpp
johnstep: This was updated to open the file and pass a handle to the server, which relays it to stdin. Support for reading the Dockerfile from stdin was also added, along with a new test.

src/linux/init/WSLAInit.cpp
```diff
@@ -666,7 +665,13 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_SIGNAL& M
 
 void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_UNMOUNT& Message, const gsl::span<gsl::byte>& Buffer)
 {
-    Channel.SendResultMessage<int32_t>(umount(Message.Buffer) == 0 ? 0 : errno);
+    auto result = umount(Message.Buffer) == 0 ? 0 : errno;
+    if (result == 0)
+    {
+        rmdir(Message.Buffer);
+    }
```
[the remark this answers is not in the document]
johnstep: This is unlikely and not actionable by the user, so there is questionable benefit in reporting it.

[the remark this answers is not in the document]
OneBlue: I recommend returning the value from `rmdir` as `result` here. This would be mostly useful to detect invalid operations where we unmount a folder and somehow someone writes to it before we delete the directory

src/windows/wslasession/WSLASession.cpp
```diff
@@ -350,116 +350,135 @@ try
...
-            // If the status code is an error, then this is an error message, not a progress update.
-            errorJson.append(Content.data(), Content.size());
+            if (pendingJson.empty() || pendingJson[0] != '{')
+            {
+                allOutput.append(pendingJson).append("\n");
+                pendingJson.clear();
+            }
+
             return;
         }
```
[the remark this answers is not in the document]
johnstep: This could possibly happen if the process terminates early but the build will still fail with an error.

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -453,22 +453,27 @@ static int Build(std::wstring_view commandLine)
...
-    LARGE_INTEGER fileSize{};
-    THROW_IF_WIN32_BOOL_FALSE(GetFileSizeEx(tarFile.get(), &fileSize));
-
-    wslutil::PrintMessage(std::format(L"Sending build context ({} bytes)...\n", fileSize.QuadPart), stdout);
+    wil::unique_hfile dockerfileHandle;
+    if (dockerfilePath == "-")
+    {
+        HANDLE stdinDup;
+        THROW_IF_WIN32_BOOL_FALSE(DuplicateHandle(
+            GetCurrentProcess(), GetStdHandle(STD_INPUT_HANDLE), GetCurrentProcess(), &stdinDup, 0, TRUE, DUPLICATE_SAME_ACCESS));
```
OneBlue: nit: We can get rid of `stdinDup` by doing: 

        THROW_IF_WIN32_BOOL_FALSE(DuplicateHandle(
            GetCurrentProcess(), GetStdHandle(STD_INPUT_HANDLE), GetCurrentProcess(), &dockerfileHandle, 0, TRUE, DUPLICATE_SAME_ACCESS));

src/windows/wsladiag/wsladiag.cpp
```diff
@@ -453,22 +453,27 @@ static int Build(std::wstring_view commandLine)
...
+        dockerfileHandle.reset(stdinDup);
+    }
+    else if (!dockerfilePath.empty())
+    {
+        SECURITY_ATTRIBUTES attributes{.nLength = sizeof(attributes), .bInheritHandle = TRUE};
+        auto dockerfileFullPath = std::filesystem::absolute(inputPath / wsl::shared::string::MultiByteToWide(dockerfilePath));
```
OneBlue: Given that we are opening the file inside `wsladiag` directly, we don't actually need to go through the absolute path, we directly pass the user's path to CreateFile()

src/windows/wslasession/WSLASession.cpp
```diff
@@ -350,116 +350,136 @@ try
...
-            try
-            {
-                errorMessage = wsl::shared::FromJson<docker_schema::ErrorResponse>(errorMessage.c_str()).message;
-            }
-            catch (...)
-            {
-                LOG_CAUGHT_EXCEPTION();
-            }
-        }
+    buildProcess.GetExitEvent().wait();
```
OneBlue: nit: Could be simplified to: 

```
int exitCode = buildProcess.Wait();
```

test/windows/WSLATests.cpp
```diff
@@ -505,18 +505,25 @@ class WSLATests
 
     HRESULT BuildImageFromContext(const std::filesystem::path& contextDir, const char* imageTag, const char* dockerfilePath = nullptr)
     {
-        auto tarFile = wsl::windows::common::helpers::CreateDockerContextTarArchive(contextDir);
+        wil::unique_hfile dockerfileHandle;
+        if (dockerfilePath != nullptr)
+        {
+            SECURITY_ATTRIBUTES attributes{.nLength = sizeof(attributes), .bInheritHandle = TRUE};
```
OneBlue: nit: I don't think that handle needs to be inheritable here

src/windows/wslasession/WSLASession.cpp
```diff
@@ -350,116 +350,136 @@ try
...
 
-    auto onChunk = [&](const gsl::span<char>& Content) {
-        if (buildResult.has_value() && buildResult.value() != boost::beast::http::status::ok)
+        pendingJson.append(line);
+
+        if (!nlohmann::json::accept(pendingJson))
         {
-            // If the status code is an error, then this is an error message, not a progress update.
-            errorJson.append(Content.data(), Content.size());
+            if (pendingJson.empty() || pendingJson[0] != '{')
```
OneBlue, line 426: It's not great that docker doesn't output error messages on stderr, but I guess as long as the first char is always a `{` and that the json is always on a single line we should be OK

OneBlue (approved):
LGTM, minor comments

src/windows/wslasession/WSLASession.cpp
johnstep: Unfortunately, the JSON objects span multiple lines in the BuildKit being used by this version of Docker. Multiple lines are handled by multiple `accept` calls until the object is valid. The following change to BuildKit switches to a single line: https://github.com/moby/buildkit/pull/4770/files

PR #14224, opened 2026-02-18
SDK surface updates

JohnMcPMS:
SDK API surface updates found while implementing the simple demo portions.

## PR Checklist

Changes to the SDK API surface from implementing the simple demo APIs.  There are both missing options and some mismatches between the runtime API and SDK API that are resolved.

## Validation Steps Performed
None at the moment, next PR will be initializing the test infrastructure for SDK.

### DIFF
--- a/src/windows/WslcSDK/wslcsdk.cpp
+++ b/src/windows/WslcSDK/wslcsdk.cpp
@@ -30,13 +30,13 @@ constexpr UINT64 s_DefaultStorageSize = 1000 * 1000 * 1000;
 #define WLSC_FLAG_VALUE_ASSERT(_wlsc_name_, _wsla_name_) \
     static_assert(_wlsc_name_ == _wsla_name_, "Flag values differ: " #_wlsc_name_ " != " #_wsla_name_);
 
-template<typename Flags>
+template <typename Flags>
 struct FlagsTraits
 {
     static_assert(false, "Flags used without traits defined.");
 };
 
-template<>
+template <>
 struct FlagsTraits<WslcSessionFeatureFlags>
 {
     using WslaType = WSLAFeatureFlags;
@@ -72,7 +72,7 @@ struct FlagsTraits<WslcContainerStartFlags>
     WLSC_FLAG_VALUE_ASSERT(WSLC_CONTAINER_START_FLAG_ATTACH, WSLAContainerStartFlagsAttach);
 };
 
-template<typename Flags>
+template <typename Flags>
 FlagsTraits<Flags>::WslaType ConvertFlags(Flags flags)
 {
     using traits = FlagsTraits<Flags>;
--- a/src/windows/WslcSDK/wslcsdk.cpp
+++ b/src/windows/WslcSDK/wslcsdk.cpp
@@ -27,13 +27,23 @@ constexpr ULONG s_DefaultBootTimeout = 300000;
 // Default to 1 GB
 constexpr UINT64 s_DefaultStorageSize = 1000 * 1000 * 1000;
 
-WSLAFeatureFlags ConvertFlags(WslcSessionFlags flags)
+WSLAFeatureFlags ConvertFlags(WslcSessionFeatureFlags flags)
 {
-    static_assert(WSLC_SESSION_FLAG_ENABLE_GPU == WslaFeatureFlagsGPU, "Session GPU flag values differ.");
+    static_assert(WSLC_SESSION_FEATURE_FLAG_ENABLE_GPU == WslaFeatureFlagsGPU, "Session GPU feature flag values differ.");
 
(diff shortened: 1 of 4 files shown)


src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -27,25 +27,56 @@ constexpr ULONG s_DefaultBootTimeout = 300000;
...
 {
-    static_assert(WSLC_CONTAINER_FLAG_AUTO_REMOVE == WSLAContainerFlagsRm, "Container auto remove flag values differ.");
-    static_assert(WSLC_CONTAINER_FLAG_ENABLE_GPU == WSLAContainerFlagsGpu, "Container GPU flag values differ.");
+    using WslaType = WSLAContainerFlags;
+    constexpr static WslcContainerFlags Mask = WSLC_CONTAINER_FLAG_AUTO_REMOVE | WSLC_CONTAINER_FLAG_ENABLE_GPU;
+    WLSC_FLAG_VALUE_ASSERT(WSLC_CONTAINER_FLAG_AUTO_REMOVE, WSLAContainerFlagsRm);
+    WLSC_FLAG_VALUE_ASSERT(WSLC_CONTAINER_FLAG_ENABLE_GPU, WSLAContainerFlagsGpu);
     // TODO: Are these the same flags?
-    // static_assert(WSLC_CONTAINER_FLAG_PRIVILEGED == WSLAContainerFlagsInit, "Container privileged flag values differ.");
+    // WLSC_FLAG_VALUE_ASSERT(WSLC_CONTAINER_FLAG_PRIVILEGED, WSLAContainerFlagsInit);
```
OneBlue: Nope these are completely different. We might want to update the values to match here (I don't think we plan to expose init in the SDK)

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -52,6 +52,13 @@ typedef struct WslcProcessSettings
 
 DECLARE_HANDLE(WslcProcess);
 
+typedef enum WslcSessionNetworkingMode
+{
+    WSLC_SESSION_NETWORKING_MODE_NONE = 0, // No networking / isolated
+    WSLC_SESSION_NETWORKING_MODE_NAT = 1,
+    WSLC_SESSION_NETWORKING_MODE_VIRT_IO_PROXY = 2
```
OneBlue: I don't think we want to expose virtio proxy vs NAT here. I would recommend just hardcoding WSLC_SESSION_NETWORKING_MODE_VIRT_IO_PROXY for now

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -70,10 +70,19 @@ typedef struct WslcVhdRequirements
...
+    WSLC_SESSION_FEATURE_FLAG_ENABLE_GPU = 0x00000004
+} WslcSessionFeatureFlags;
+
+DEFINE_ENUM_FLAG_OPERATORS(WslcSessionFeatureFlags);
+
 typedef enum WslcSessionFlags
 {
     WSLC_SESSION_FLAG_NONE = 0x00000000,
-    WSLC_SESSION_FLAG_ENABLE_GPU = 0x00000004
+    WSLC_SESSION_FLAG_PERSISTENT = 0x00000001,
```
OneBlue: Hmm I'm not sure that we want to offer persistent sessions in the SDK. Do you think there's a scenario for this @craigloewen-msft  ?

OneBlue (approved):
LGTM. One design question regarding persistent sessions, but we can resolve this in a future PR. Feel free to merge and update once we know whether we want to support persistent sessions in the SDK or not

src/windows/WslcSDK/wslcsdk.h
JohnMcPMS: Per internal discussion, removing this.  Leaving previously existing flags with their new name as we may decide to add these back at some future time.

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -84,13 +105,16 @@ void GetErrorInfoFromCOM(PWSTR* errorMessage)
 } // namespace
 
 // SESSION DEFINITIONS
-STDAPI WslcSessionInitSettings(_In_ PCWSTR storagePath, _Out_ WslcSessionSettings* sessionSettings)
+STDAPI WslcSessionInitSettings(_In_ PCWSTR identifier, _In_opt_ PCWSTR storagePath, _Out_ WslcSessionSettings* sessionSettings)
```
OneBlue: `storagePath` shouldn't be optional here. The runtime does support using a tmpfs for storage as a convenience feature during development, but I don't think we want to offer it to the SDK (mainly because it's very easy to run out of memory and get the VM to a broken state that way)

OneBlue (approved):

PR #14225, opened 2026-02-18
Initial support for port mapping in CLI

AmelBawa-msft:
## 🦞 Port mapping parser
Parser support examples:
#### ✅ Valid (parsing)
```bash
 --- IP not specified ---
-p 8080:80
-p 443:443
-p 80
-p 80/tcp
-p 53/udp

# --- IPv4 specified ---
-p 127.0.0.1:8080:80
-p 0.0.0.0:8080:80
-p 192.168.1.50:8080:80
-p 127.0.0.1:5353:5353/udp
-p 127.0.0.1::80
-p 127.0.0.1::53/udp

# --- IPv6 specified ---
-p '[::1]:8080:80'
-p '[::]:8080:80'
-p '[2001:db8::10]:8080:80'
-p '[::1]:5353:5353/udp'
-p '[::1]::80'
-p '[::]::53/udp'

# --- Protocol not specified (defaults to tcp) ---
-p 9090:90
-p 127.0.0.1:9090:90
-p '[::1]:9090:90'

# --- Both tcp and udp (separate publishes) ---
-p 53:53/tcp
-p 53:53/udp
-p 5000:5000/tcp
-p 5000:5000/udp

# --- Port ranges ---
-p 8000-8005:8000-8005
-p 127.0.0.1:9000-9003:9000-9003
-p '[::1]:7000-7002:7000-7002'
-p 10000-10010:10000-10010/tcp
-p 20000-20002:20000-20002/udp
```
#### ❌ Invalid (parsing)
```bash
# Malformed IP/container combinations
-p 127.0.0.1:80
-p '[::1]:8080'

# Port range mismatches
-p 8000-8005:8000-8006
-p 8000-8005:8000
-p 8000:8000-8005

# Bad protocol
-p 8080:80/icmp
-p 8080:80/udpp

# IPv6 without brackets
-p ::1:8080:80
```

## 🦐 Full example
```bash
# Example 1: HOST_PORT:CONTAINER_PORT
# (host port) 8080 => (container port) 80
wslc container run -p 8080:80 -it --name demo nginx:alpine
```

## PR Checklist


dkbennett:
I have a meta comment before we go too far down the argument validation path. We have two types of arguments in the CLI:
1) Arguments directly consumed by the CLI to change its behavior (progress settings, output, logging, settings specific to the CLI, etc)
2) Arguments that are generally passed on to the service or another component, such as the operating system.

@OneBlue recommended that our general approach for argument validation for arguments to the service is that we should pass the argument along to the service with as little modification as possible, and if it's bad input the service will return error messages that we can surface to the user. I completely agree with this approach and believe it should apply to everything not just the service.

I would prefer not to have rigid validation of arguments in the CLI _**unless**_ it's a type 1 argument specific to the CLI or some minimum parsing needs to be done in order to put the data into a form the service can consume (converting a string to two integers, for example).

The port validation is something I would expect the service to handle and I am generally concerned about the CLI parsing the port data argument differently from the service.  I am strongly in favor of only validating the bare minimum needed to make it valid input to the service and let the service tell us if its inputs were invalid.

So the general guidance for argument validation should be: "**Is something else already validating or going to validate this?**" If the answer is yes, then the CLI should only validate it enough to get it to that validator and then rely on that validator to do its job correctly.

AmelBawa-msft:
@dkbennett great observation and feedback, thank you! I have a couple of questions before marking this PR as ready, one of which you touched on, and another related to what we currently support.

This PR contains the necessary pieces to light up the feature, but I’ve held off on marking it ready to avoid duplicating service logic and to clarify how certain parts of the syntax should be handled. Some aspects may require follow-up work with the runtime team. I will sync up today with the team for suggestions on the next step.

Excited to see this feature finally available 🚀🙌

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -26,6 +28,46 @@ using wsl::windows::common::docker_schema::InspectContainer;
 using wsl::windows::common::wslutil::PrintMessage;
 using namespace wsl::windows::wslc::models;
 
+static inline int ResolveOrAllocatePort(PublishPort port, int offset)
```
AmelBawa-msft: @OneBlue should resolving ephemeral host port (e.g. ask windows to give us a port, which is implemented below) be part of the service?

For example, from the client (CLI), would it make sense to do:
```
container.AddPort(/*host_port*/ 0, 8080, ...);
```

And then the service implementation of `AddPort` would reason about the input (`0`) and ask windows to allocate a port for us?

OneBlue: Yes that should be done by the service. If the CLI binds a port like it's done below, then the service won't be able to use it. 

If the CLI closes the bound socket just before calling the service, then that creates a race condition because someone else could bind that socket before the service does

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+namespace wsl::windows::wslc::models {
+
+PublishPort::PortRange PublishPort::PortRange::ParsePortPart(const std::string& portPart)
+{
+    static auto parsePort = [](const std::string& value, const char* errorMessage) -> int {
+        try
+        {
+            size_t idx = 0;
+            int port = std::stoi(value, &idx, 10);
```
OneBlue: I recommend using `stoul ` and failing if the value is  >= 2 ^16, since port numbers are unsigned 16 bits integers

src/windows/wslc/services/ContainerService.cpp
AmelBawa-msft: Great! That totally makes sense, I had fun working on this, but would prefer to make sure it's implemented in the appropriate layer 😆 Do you suggest passing 0 for port from the CLI in order to ask for a port on the service end?

Also, I noticed the __host IP__ and __protocol__ are not being consumed by the service API in 
```c++
void AddPort(uint16_t WindowsPort, uint16_t ContainerPort, int Family);
```
Any suggestion if that bit is implemented yet?

For instance:

```bash
wslc container run -p 8080:80 ➡️ container.AddPortW(8080, 80,AF_INET /** or AF_INET6*/);
wslc container run -p 127.0.0.1:8080:80 ➡️ ❓
wslc container run -p 8080:80/udp ➡️ ❓
```

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+namespace wsl::windows::wslc::models {
+
+PublishPort::PortRange PublishPort::PortRange::ParsePortPart(const std::string& portPart)
+{
+    static auto parsePort = [](const std::string& value, const char* errorMessage) -> uint16_t {
+        try
+        {
+            // Ensure the value is not empty and contains only digits before parsing
+            if (value.empty() || !std::all_of(value.begin(), value.end(), ::isdigit))
```
OneBlue, line 28: I recommend calling std::stoul directly and catching exceptions that it might throw directly

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+    if (!hostIpPart.empty() && hostIpPart.front() == '[' && hostIpPart.back() == ']')
+    {
+        auto address = hostIpPart.substr(1, hostIpPart.size() - 2);
+        IN6_ADDR v6{};
+        if (inet_pton(AF_INET6, address.c_str(), &(v6)) == 1)
+        {
+            return PublishPort::IPAddress(v6);
+        }
+
+        THROW_HR_WITH_USER_ERROR(E_INVALIDARG, "Invalid IPv6 address specified in port mapping.");
```
OneBlue: I recommend including the argument value in the error message

src/windows/wslc/services/ContainerModel.cpp
AmelBawa-msft: `std::stoul` unfortunately will successfully parse the following (invalid) strings:
- "80 " => 80
- "127.0.0.1" => 127

I added those edge cases to the UT

src/windows/wslc/services/ContainerService.cpp
AmelBawa-msft: ephemeral port, host ip are not currently supported

CC: @OneBlue

src/windows/wslc/services/ContainerModel.h
```diff
@@ -66,4 +67,141 @@ struct ContainerInformation
...
+        }
+
+        static IPAddress ParseHostIP(const std::string& hostIpPart);
+        bool IsLoopback() const;
+        bool IsAllInterfaces() const;
+        std::string ToString() const;
+
+    private:
+        bool m_isIPv6 = false;
+        std::array<uint8_t, 16> m_bytes{};
```
OneBlue: I don't recommend parsing the IP addresses in the CLI. The service will take the IP addresses as strings and parse them, so we can just forward the unparsed IP addresses to the service

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -85,6 +87,36 @@ static wsl::windows::common::RunningWSLAContainer CreateInternal(
...
+                    "Port mappings with ephemeral host ports, specific host IPs, or UDP protocol are not currently supported");
+            }
+        }
+
+        auto containerPort = portMapping.ContainerPort();
+        for (uint16_t i = 0; i < containerPort.Count(); ++i)
+        {
+            auto currentContainerPort = static_cast<uint16_t>(containerPort.Start() + i);
+            auto currentHostPort = static_cast<uint16_t>(portMapping.HostPort().Start() + i);
+            containerLauncher.AddPort(currentHostPort, currentContainerPort, AF_INET);
```
OneBlue, line 116: This is assuming that the port binding is ipv4 (AF_INET), but if the user passed something like `[::]:80:80`, we should use ipv6 (AF_INET6)

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -191,5 +318,83 @@ class WSLCE2EContainerCreateTests
...
+           << L"while (my $c = $s->accept) {"     //
+           << L"  print $c '" << message << L"';" //
+           << L"  close $c;"                      //
+           << L"}";                               //
+        return ss.str();
+    }
+
+    uint16_t GetFreePort()
+    {
+        // Note: This function finds a free port by binding to port 0, which
```
OneBlue: Instead of doing this I would recommend just hardcoding a port. We commonly use '1234' in the tests in other places (that way we know that as long nothing binds that port, the tests can't fail)

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -191,5 +318,83 @@ class WSLCE2EContainerCreateTests
...
+            THROW_HR_IF(E_FAIL, response.empty());
+            return wsl::shared::string::MultiByteToWide(response);
+        };
+
+        return wsl::shared::retry::RetryWithTimeout<std::wstring>(readOnce, std::chrono::milliseconds(500), std::chrono::minutes(1), []() {
+            return wil::ResultFromCaughtException() == E_FAIL;
+        });
+    }
+
+    std::wstring GetPerlHttpServerCode(std::wstring message, uint16_t port)
```
OneBlue: I recommend reusing the python http server logic instead of using perl here : https://github.com/microsoft/WSL/blob/e91443592d6320d5de9a5f8db4e86b7b2f8ffe38/test/windows/WSLATests.cpp#L3806-L3807

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -191,5 +318,83 @@ class WSLCE2EContainerCreateTests
                 << L"  -h,--help         Shows help about the selected command\r\n\r\n";
         return options.str();
     }
+
+    std::wstring ReadFromLoopback(uint16_t port)
```
OneBlue: I recommend reusing https://github.com/microsoft/WSL/blob/e91443592d6320d5de9a5f8db4e86b7b2f8ffe38/test/windows/WSLATests.cpp#L3741-L3742 instead (We can move that method to Common.cpp if needed)

src/windows/wslc/services/ContainerModel.h
AmelBawa-msft: Is this already implemented in the service?

nit: It feels somewhat awkward to split the parsing of the port mapping value between the client and the service, with each handling only part of it

src/windows/wslc/services/ContainerService.cpp
AmelBawa-msft: I had this as an if/else for ipv4/6, but I reduced it to just ipv4 since we decided to not allow the user to pass a host ip as part of the input for now

- #14433

src/windows/wslc/services/ContainerModel.h
OneBlue: I'm working on it right now, but it's not available yet. You can expect that the service will take a string for the IP address

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
AmelBawa-msft: Found this method duplicated in both files WSLATest and Common. I removed the WSLATest copy

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+namespace wsl::windows::wslc::models {
+
+PublishPort::PortRange PublishPort::PortRange::ParsePortPart(const std::string& portPart)
+{
+    static auto parsePort = [](const std::string& value, const std::string& errorMessage) -> uint16_t {
+        try
+        {
+            // Ensure the value is not empty and contains only digits before parsing
+            if (value.empty() || !std::all_of(value.begin(), value.end(), ::isdigit))
```
OneBlue, line 28: We don't need to manually check this, since `std::stoul` will throw on exception

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+PublishPort::IPAddress PublishPort::IPAddress::ParseHostIP(const std::string& hostIpPart)
+{
+    // Check if it's an IPv6 address (enclosed in square brackets)
+    if (!hostIpPart.empty() && hostIpPart.front() == '[' && hostIpPart.back() == ']')
+    {
+        auto address = hostIpPart.substr(1, hostIpPart.size() - 2);
+        IN6_ADDR v6{};
+        if (inet_pton(AF_INET6, address.c_str(), &(v6)) == 1)
```
OneBlue: The CLI doesn't need to parse this since this will be done in the service (same for IPV4)

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+        if (inet_pton(AF_INET, hostIpPart.c_str(), &(v4)) == 1)
+        {
+            return PublishPort::IPAddress(v4);
+        }
+
+        THROW_HR_WITH_USER_ERROR(E_INVALIDARG, std::format("Invalid IPv4 address specified in port mapping: '{}'.", hostIpPart));
+    }
+}
+
+bool PublishPort::IPAddress::IsAllInterfaces() const
```
OneBlue: Looks like this method isn't used

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+        if (m_bytes[i] != 0)
+        {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+std::string PublishPort::IPAddress::ToString() const
```
OneBlue: I recommend storing the IP address as strings so we don't have to do any of this in the CLI

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -85,6 +87,36 @@ static wsl::windows::common::RunningWSLAContainer CreateInternal(
...
+            // https://github.com/microsoft/WSL/issues/14433
+            // The following scenarios are currently not implemented:
+            // - Ephemeral host port mappings
+            // - Host port mappings with a specific host IP
+            // - Host port mappings with UDP protocol
+            if (portMapping.HostPort().IsEphemeral() || portMapping.HostIP().has_value() ||
+                portMapping.PortProtocol() == PublishPort::Protocol::UDP)
+            {
+                THROW_HR_WITH_USER_ERROR(
+                    E_FAIL,
```
OneBlue: nit: I recommend `HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED)` for errors like this one

test/windows/Common.cpp
```diff
@@ -2824,41 +2824,55 @@ std::filesystem::path GetTestImagePath(std::string_view imageName)
...
+                VERIFY_ARE_EQUAL(result, HRESULT_FROM_WIN32(WININET_E_INVALID_SERVER_RESPONSE));
+                return;
+            }
+
+            // Throw so RetryWithTimeout can decide whether to retry.
+            THROW_HR(result);
         }
+    };
+
+    if (retry)
```
OneBlue, line 2856: Ports should always be bound synchronously, so there shouldn't be a need for retries. Was this for a specific test failure ?

src/windows/wslc/services/ContainerModel.cpp
AmelBawa-msft: std::stoul unfortunately will successfully parse the following (invalid) strings:

"80 " => 80
"127.0.0.1" => 127
I added those edge cases to the UT

src/windows/wslc/services/ContainerModel.cpp
AmelBawa-msft: Will remove. I added a couple methods that are unused now because I had in mind we had several not yet implemented featuers (addressed in #14469). But I wasn't sure what the API would look like back then

src/windows/wslc/services/ContainerModel.cpp
AmelBawa-msft: We use this information to determine if the IP is v4 or v6. But if the service will handle this, we can possibly reduce this to checking for just `[` and `]`?

src/windows/wslc/services/ContainerModel.cpp
OneBlue: Ah that's a good point. We could probably fix this by throwing if `pos` isn't equal to the size of the string when `std::stoul()` returns. That should cover all cases

src/windows/wslc/services/ContainerModel.cpp
OneBlue: Yeah I think just relying on the bracket syntax is the right move for now

AmelBawa-msft: One interesting use case I came across is:

```
::1:8080:80
```

After the refactor PR is merged, the CLI will pass:

```
AddPort(8080, 80, AF_INET, "::1")
```

Notice that AF_INET is used instead of AF_INET6 because square brackets were not provided. I’m wondering whether this scenario will be handled correctly on the service side.

test/windows/Common.cpp
AmelBawa-msft: From running the e2e, for example:
```
wslc container run -d -p 8080:8080 python:... python3 -m http.server",
```

It takes a couple milliseconds before being able to connect and receive data.

Here are the logs:
```
StartGroup: WSLCE2ETests::WSLCE2EContainerCreateTests::WSLCE2E_Container_Run_Port_TCP
Verify: AreEqual(*expected.Stderr, *Stderr)
Verify: AreEqual(*expected.ExitCode, *ExitCode)
Sending request to: http://127.0.0.1:1234 <--------------------------------------------------------
Sending request to: http://127.0.0.1:1234 <--------------------------------------------------------
Verify: AreEqual(static_cast<int>(response.StatusCode()), expectedCode.value())
Verify: AreEqual(*expected.Stderr, *Stderr)
Verify: AreEqual(*expected.ExitCode, *ExitCode)
Verify: AreEqual(1u, stdoutLines.size())
Verify: AreEqual(1u, inspectData.size())
Verify: IsTrue(inspectContainer.Ports.contains(portKey))
Verify: AreEqual(1u, portBindings.size())
Verify: AreEqual(std::to_string(HostTestPort1), portBindings[0].HostPort)
Verify: AreEqual("127.0.0.1", portBindings[0].HostIp)
EndGroup: WSLCE2ETests::WSLCE2EContainerCreateTests::WSLCE2E_Container_Run_Port_TCP [Passed]
Verify: AreEqual(*expected.Stderr, *Stderr)
Verify: AreEqual(*expected.ExitCode, *ExitCode)
Verify: AreEqual(*expected.Stdout, *Stdout)
Verify: AreEqual(*expected.Stderr, *Stderr)
Verify: AreEqual(*expected.ExitCode, *ExitCode)
Verify: AreEqual(*expected.Stdout, *Stdout)
Verify: AreEqual(*expected.Stderr, *Stderr)
Verify: AreEqual(*expected.ExitCode, *ExitCode)

StartGroup: WSLCE2ETests::WSLCE2EContainerCreateTests::WSLCE2E_Container_Run_PortMultipleMappings
Verify: AreEqual(*expected.Stderr, *Stderr)
Verify: AreEqual(*expected.ExitCode, *ExitCode)
Sending request to: http://127.0.0.1:1234<--------------------------------------------------------
Sending request to: http://127.0.0.1:1234<--------------------------------------------------------
Sending request to: http://127.0.0.1:1234<--------------------------------------------------------
Verify: AreEqual(static_cast<int>(response.StatusCode()), expectedCode.value())
Sending request to: http://127.0.0.1:1235<--------------------------------------------------------
Verify: AreEqual(static_cast<int>(response.StatusCode()), expectedCode.value())
EndGroup: WSLCE2ETests::WSLCE2EContainerCreateTests::WSLCE2E_Container_Run_PortMultipleMappings [Passed]
```

OneBlue: Oh I see what's happening. When the container starts, the python process needs a bit of time to actually perform the port mapping. 

To synchronize this, after starting the container we should wait for python to say that the port has been bound. In the case of wslc you can probably do that by reading stdout and wait for the same output as the WSLATest class does.

[here's an example](https://github.com/microsoft/WSL/blob/0b1c0299250d40e19f79459a6515703c95064b85/test/windows/WSLATests.cpp#L3855-L3856) 

Doing this will probably be easier once @dkbennett's test changes are in though, so let's merge as-is and fix this as a followup ? (I have a preference for not having retries on operations that are synchronous, since that could potentially hide race conditions)

AmelBawa-msft: Added this suggestion here:
- https://github.com/microsoft/WSL/issues/14471

OneBlue (approved):

PR #14228, opened 2026-02-18
Initial support for volume mounting in CLI

AmelBawa-msft:
## 🦞 Volume mounting parser
- Added support for container volume option

## 🦐 Full example
```bash
# -v <host path>:<container path>[:mode]
wslc container run -v "C:/src/wsl/bin/x64/Debug/volume:/usr/share/nginx/html" -it --name demo2 nginx:alpine
```

## PR Checklist


src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+--*/
+
+#include <precomp.h>
+#include "ContainerModel.h"
+
+namespace wsl::windows::wslc::models {
+static bool IsWindowsDriveColon(const std::string& s, size_t index, size_t tokenStart)
+{
+    if (s[index] != ':' || index != tokenStart + 1 || !std::isalpha((unsigned char)s[tokenStart]) || index + 1 >= s.size())
```
OneBlue: I don't think we can make the assumption that a colon is a windows drive if the previous char is a letter. Instead I would recommend: 

Looking at the last ':' in the string, extract the last part of the string based on that. Then:

1) If the last part is "ro", "readonly" or "rw", apply the read-only flag, and look for the previous ':' in the string to split the host & container paths

2) Otherwise, use that colon to split between host & container paths

That way we don't have to worry about ':' being part of a Windows path

src/windows/wslc/services/ContainerModel.h
```diff
@@ -64,4 +65,30 @@ struct ContainerInformation
...
+    constexpr bool IsReadOnly() const
+    {
+        return m_mode == "ro";
+    }
+    static VolumeMount Parse(const std::string& value);
+
+private:
+    std::string m_hostPath;
+    std::string m_containerPath;
+    std::string m_mode;
```
OneBlue: I recommend just storing whether the mount is read-only instead of the mode string, that way we can do all the parsing in one place

src/windows/wslc/services/ContainerModel.h
```diff
@@ -64,4 +65,30 @@ struct ContainerInformation
...
+        return m_mode;
+    }
+    constexpr bool IsReadOnly() const
+    {
+        return m_mode == "ro";
+    }
+    static VolumeMount Parse(const std::string& value);
+
+private:
+    std::string m_hostPath;
```
OneBlue: m_hostPath should be a wide string (I would recommend working on a wide string, and only converting m_containerPath so we don't need a double conversion)

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -0,0 +1,83 @@
...
+    parts.push_back(current);
+    return parts;
+}
+
+VolumeMount VolumeMount::Parse(const std::string& value)
+{
+    auto parts = SplitVolumeValue(value);
+    if (parts.size() < 2 || parts.size() > 3)
+    {
+        THROW_HR_WITH_USER_ERROR(E_INVALIDARG, "Volume mount value must be in the format <host path>:<container path>[:mode]");
```
OneBlue: I recommend writing the input volume spec as part of the error message, that usually makes the UX better for the user, something like: 

"Invalid volume specifications: '{value}'. Expected format: <host path>:<container path>[:mode]"

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -80,7 +80,17 @@ static wsl::windows::common::RunningWSLAContainer CreateInternal(
     WI_SetFlagIf(processFlags, WSLAProcessFlagsStdin, options.Interactive);
     WI_SetFlagIf(processFlags, WSLAProcessFlagsTty, options.TTY);
     wsl::windows::common::WSLAContainerLauncher containerLauncher(
-        image, options.Name, options.Arguments, {}, WSLAContainerNetworkTypeHost, processFlags);
+    image, options.Name, options.Arguments, {}, WSLAContainerNetworkTypeHost, processFlags);
+
+    // Add volumes if specified
+    for (const auto& volumeSpec : options.Volumes)
+    {
+        auto volume = VolumeMount::Parse(volumeSpec);
```
dkbennett, line 92: Recommend adding VolumeMount::Parse to the ArgumentValidation.cpp so if there is an error we find it earlier before sessions are opened and any work is done, and can report it as an argument error.  See the Signal ArgType as an example.

src/windows/wslc/services/ContainerModel.cpp
AmelBawa-msft: I originally had something like the above in mind, but I quickly realized a lot of edge cases would break it:

| Input | Host path | Container path | Mode |
|------|-----------|----------------|------|
| `C:\hostPath` | `C` | `\hostPath` | `false` |
| `C:\hostPath:/containerPath:invalid_mode` | `C:\hostPath:/containerPath` | `invalid_mode` | `false` |
| `C:\hostPath:/containerPath:ro:extra` | `C:\hostPath:/containerPath:ro` | `extra` | `false` |
| `C:\hostPath:ro` | `C` | `\hostPath` | `true` |

After adding a check for a potential Windows path, the heuristic ends up doing a pretty good job of dissecting the volume input

AmelBawa-msft: Discussed offline, summary in this issue:
- https://github.com/microsoft/WSL/issues/14432

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        vm.m_containerPath = WideToMultiByte(value.substr(splitColon + 1, lastColon - splitColon - 1));
+    }
+    else
+    {
+        vm.m_containerPath = WideToMultiByte(lastToken);
+    }
+
+    vm.m_hostPath = value.substr(0, splitColon);
```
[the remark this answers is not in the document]
AmelBawa-msft: https://github.com/microsoft/WSL/issues/14432

AmelBawa-msft:
@OneBlue, a few thoughts on this PR:

- As we discussed earlier, the parser here is intentionally simple and lightweight, and we were planning to potentially enhance it in issue #14432.
- We also talked about the service being responsible for validating input, with the CLI simply passing arguments through as-is. However, I noticed that for this command, Docker validation only happens when the container is started, not when it is created. For example, if we run wslc run -v <invalid/unsupported>, the error only appears when attempting to start the container. This suggests that the container can still be created successfully even with invalid input.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -25,21 +25,33 @@ class WSLCE2EContainerCreateTests
...
     {
         EnsureContainerDoesNotExist(WslcContainerName);
+        EnsureImageIsDeleted(AlpineImage);
+        EnsureImageIsDeleted(DebianImage);
         return true;
     }
 
     TEST_METHOD_SETUP(TestMethodSetup)
     {
+        VolumeTestFile1 = wsl::windows::common::filesystem::GetTempFilename();
```
OneBlue, line 42: I recommend using a file in the current directory instead (if the tests crash, they could leak temporary files)

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +109,242 @@ class WSLCE2EContainerCreateTests
...
+        auto result = RunWslc(std::format(
+            L"container run --name {} --volume \"{}:/data\" {} sh -c \"echo -n 'WSLC Volume Test' > /data/{}\"",
+            WslcContainerName,
+            hostDirectory.wstring(),
+            AlpineImage.NameAndTag(),
+            fileName));
+        result.Verify({.Stdout = L"", .Stderr = L"", .ExitCode = S_OK});
+
+        // Read all file content
+        std::wifstream in(VolumeTestFile1);
```
OneBlue: nit: We can use `ReadFileContent()` for this

OneBlue (approved):
LGTM, minor comments

PR #14230, opened 2026-02-19
Fix leaked tracking reference when a process outlives its container

OneBlue:

This change fixes a potential crash that can be caused by a `ContainerTrackingReference` outliving the session if a process reference is kept after a container is released. 

## PR Checklist

### DIFF
--- a/src/windows/wslasession/WSLAProcessControl.cpp
+++ b/src/windows/wslasession/WSLAProcessControl.cpp
@@ -92,11 +92,16 @@ int DockerContainerProcessControl::GetPid() const
 
 void DockerContainerProcessControl::OnContainerReleased()
 {
-    std::lock_guard lock{m_lock};
-    m_container = nullptr;
+    {
+        std::lock_guard lock{m_lock};
+
+        WI_ASSERT(m_container != nullptr);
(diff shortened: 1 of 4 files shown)


src/windows/wslasession/WSLAProcessControl.cpp
```diff
@@ -153,10 +162,18 @@ void DockerExecProcessControl::OnEvent(ContainerEvent Event, std::optional<int>
...
+        m_container = nullptr;
+    }
+
+    // N.B. The caller might keep a reference to the process even after the container is released.
+    // If that happens, make sure that the state tracking can't outlive the session.
+    // This is safe to call without the lock because removing the tracking reference is protected by the event tracker lock.
 
-    WI_ASSERT(m_container != nullptr);
-    m_container = nullptr;
+    m_trackingReference.Reset();
```
ptrivedi, line 176: Hmm, interesting. Curious about how this issue manifested. Was this causing a crash?

OneBlue: It was. That's how I discovered the issue

ptrivedi (approved):

PR #14231, opened 2026-02-19
Unshare the FD table when creating threads in init to avoid fd leaks

OneBlue:

This change solves a potential FD leak that can happen if multiple WSLA shells are open at the same time. 

The leak happens when threads inside of init create children processes, which inherit all fd from their parents by default. If that happens while an hvsocket is opened, that socket is inherited in the child, which will never close it, causing a hang

## PR Checklist


src/linux/init/WSLAInit.cpp
```diff
@@ -447,13 +447,20 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_FORK& Mes
     std::promise<pid_t> childPid;
 
     {
-        auto childLogic = [ListenSocket = std::move(ListenSocket), &SocketAddress, &Channel, &Message, &childPid]() mutable {
+        auto childLogic = [ListenSocket = wil::unique_fd{ListenSocket.get()}, &SocketAddress, &Channel, &Message, &childPid]() mutable {
```
johnstep: I believe that the process and pty cases will now result in a double close because the lambda's destructor will close the captured `ListenSocket`, while the destructor of the outer `ListenSocket` will also close it. `UtilCreateChildProcess` takes a forwarding reference, so the lambda is not actually moved.

Maybe just capture the fd value directly and then close it directly within the lambda?
```C++
450:         auto childLogic = [listenFd = ListenSocket.get(), &SocketAddress, &Channel, &Message, &childPid]() mutable {
...
466:             wil::unique_fd ProcessSocket{UtilAcceptVsock(listenFd, SocketAddress, SESSION_LEADER_ACCEPT_TIMEOUT_MS)};
...
469:             close(listenFd);
```

src/linux/init/WSLAInit.cpp
```diff
@@ -447,13 +447,20 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_FORK& Mes
...
             if (Message.ForkType == WSLA_FORK::Process || Message.ForkType == WSLA_FORK::Pty)
             {
                 Channel.Close();
             }
 
+            if (Message.ForkType == WSLA_FORK::Thread)
+            {
+                // If this is a thread, detach from the process' fd table.
+                // This prevents other threads from creating child processes that could inherit fds that this thread could create.
+                THROW_LAST_ERROR_IF(unshare(CLONE_FILES) < 0);
```
ptrivedi, line 463: This means that going forward, any new files opened or closed by this thread won't affect the parent and vice versa. Could there be unintended consequences to this?

OneBlue: That's essentially what already happens if we fork() with the Process/Pty flags instead. It's because forking with the Thread flag shares the parent's fd table that can have handle leaks

src/linux/init/WSLAInit.cpp
OneBlue: That's a good point. Fixed in the latest iteration by only "assigning" the FD once we have the guarantee that we have our own FD table

src/linux/init/WSLAInit.cpp
```diff
@@ -446,13 +446,25 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_FORK& Mes
...
+            if (Message.ForkType == WSLA_FORK::Thread)
+            {
+                // If this is a thread, detach from the process' fd table.
+                // This prevents other threads from creating child processes that could inherit fds that this thread could create.
+                // N.B. This needs to happen before childPid is signalled to ensure that ListenSocket() is not closed by the parent before getting duplicated in the child's fd table.
+                THROW_LAST_ERROR_IF(unshare(CLONE_FILES) < 0);
+            }
+
+            // ListenSocket should only be assigned after this thread is guaranteed to have its own fd table (either via unshare() or a child process).
+            ListenSocket.reset(ListenSocketFd);
```
benhillis, line 467: This doesn’t seem right to me. Both this thread and the parent have this file descriptor.

johnstep: I believe this is correct since the file descriptor here is in a separate table from the parent.

johnstep (approved):

src/linux/init/WSLAInit.cpp
benhillis: yeah I couldn't quite follow the logic, this seems right to me.

benhillis (approved):

PR #14232, opened 2026-02-19
WSLC SDK Initial Tests

JohnMcPMS:
Tests for initial WLSC SDK implementation and placeholders for not-yet-implemented APIs.

## PR Checklist

Adds tests for the implemented portions of the WSLC SDK and placeholders that expect `E_NOTIMPL` results from the unimplemented functions.  Based on the WSLA tests as the SDK is a wrapper layer for it.

## Validation Steps Performed
Ran these tests.


test/windows/WslcSdkTests.cpp
```diff
@@ -0,0 +1,833 @@
...
+extern std::wstring g_testDataPath;
+extern bool g_fastTestRun;
+
+namespace {
+
+//
+// RAII guards for opaque WSLC handle types.
+//
+
+struct WslcSessionGuard
```
OneBlue: I recommend reusing `wil::unique_any` for this. For instance: 

```
void CloseSession(WslcSession session)
{
            WslcSessionTerminate(session);
            WslcSessionRelease(session);
}

using UniqueSession = wil::unique_any<WslcSession , decltype(CloseSession), CloseSession>;

```

The main benefit of this it has it already has .get(), .release() and move semantics implemented

test/windows/WslcSdkTests.cpp
```diff
@@ -0,0 +1,833 @@
...
+
+    WslcSessionGuard() = default;
+    WslcSessionGuard(const WslcSessionGuard&) = delete;
+    WslcSessionGuard& operator=(const WslcSessionGuard&) = delete;
+
+    ~WslcSessionGuard()
+    {
+        if (session)
+        {
+            WslcSessionTerminate(session);
```
OneBlue: I recommend validating the return value of these (even in the RAII helpers) since we're validating the SDK behavior here

test/windows/WslcSdkTests.cpp
```diff
@@ -0,0 +1,833 @@
...
+        {
+            WslcProcessRelease(process);
+        }
+    }
+
+    operator WslcProcess() const { return process; }
+};
+
+// Reads all data from a pipe handle until it closes.
+std::string ReadHandleToString(HANDLE handle)
```
OneBlue: I recommend reusing [this method](https://github.com/microsoft/WSL/blob/465f3f093e12345dcdef273e59bcc1235208a0ec/test/windows/Common.cpp#L2605)  from Common.h

test/windows/WslcSdkTests.cpp
```diff
@@ -0,0 +1,833 @@
...
+    // Acquire stdout / stderr pipe handles (caller owns these).
+    HANDLE rawStdout = nullptr;
+    THROW_IF_FAILED(WslcProcessGetIOHandles(process, WSLC_PROCESS_IO_HANDLE_STDOUT, &rawStdout));
+    wil::unique_handle ownedStdout(rawStdout);
+
+    HANDLE rawStderr = nullptr;
+    THROW_IF_FAILED(WslcProcessGetIOHandles(process, WSLC_PROCESS_IO_HANDLE_STDERR, &rawStderr));
+    wil::unique_handle ownedStderr(rawStderr);
+
+    // Read stdout / stderr concurrently so that full pipe buffers do not stall the process.
```
OneBlue, line 121: I recommend using a `MuliHandleWait` to read from all handles on the same thread. 

[Here's a similar usage example](https://github.com/microsoft/WSL/blob/465f3f093e12345dcdef273e59bcc1235208a0ec/test/windows/WSLATests.cpp#L3332C1-L3347C24) 

The main benefit of that approach is we don't have to worry about detaching / joining threads if something goes wrong.

test/windows/WslcSdkTests.cpp
```diff
@@ -0,0 +1,833 @@
...
+
+        WslcVhdRequirements vhdReqs{};
+        vhdReqs.sizeInBytes = 4096ull * 1024 * 1024; // 4 GB
+        vhdReqs.type = WSLC_VHD_TYPE_DYNAMIC;
+        VERIFY_SUCCEEDED(WslcSessionSettingsSetVHD(&sessionSettings, &vhdReqs));
+
+        VERIFY_SUCCEEDED(WslcSessionCreate(&sessionSettings, &m_defaultSession));
+
+        // Pull images required by the tests (no-op if already present).
+        for (const char* image : {"debian:latest", "hello-world:linux"})
```
OneBlue, line 182: Could we reuse the same storage folder that we use in `WSLATests` ? That would remove the number of pulls required (which can be an issue when developing since the default registry will rate limit at ~ 100 calls / 6 hours)

test/windows/WslcSdkTests.cpp
```diff
@@ -0,0 +1,833 @@
...
+    }
+
+    // -----------------------------------------------------------------------
+    // Container lifecycle tests
+    // -----------------------------------------------------------------------
+
+    TEST_METHOD(CreateContainer)
+    {
+        WSL2_TEST_ONLY();
+        SKIP_TEST_ARM64();
```
OneBlue: Some of our tests have this "skip ARM64" flag for legacy reasons, let's not add it to new tests (this will make arm64 testing easier later)

OneBlue (commented):
LGTM, minor comments

test/windows/WslcSdkTests.cpp
JohnMcPMS: I'm not seeing a `VERIFY_SUCCEEDED_LOG` type macro in TAEF.  I doubt that we want to terminate the process if these are happening during an exception in the test.  We could try/catch it, but would that actually cause the test to fail due to error logging?

In addition, these are attempting to do maximum cleanup for idempotency reasons.  Tests can and should call these cleanup functions and verify the output as part of running the main test body.

Do you want it to be a requirement that the runtime always returns a success on repeat calls to these cleanup functions?

test/windows/WslcSdkTests.cpp
```diff
@@ -0,0 +1,769 @@
...
+        VERIFY_ARE_EQUAL(future.get(), WSLC_SESSION_TERMINATION_REASON_SHUTDOWN);
+    }
+
+    // -----------------------------------------------------------------------
+    // Image tests
+    // -----------------------------------------------------------------------
+
+    TEST_METHOD(PullImage)
+    {
+        WSL2_TEST_ONLY();
```
OneBlue, line 320: We might need to disable this test for now, or add handling for when we hit the API limit given that this is already hitting in our CI

test/windows/WslcSdkTests.cpp
```diff
@@ -0,0 +1,769 @@
...
+        VERIFY_SUCCEEDED(WslcSessionSettingsSetVHD(&sessionSettings, &vhdReqs));
+
+        VERIFY_SUCCEEDED(WslcSessionCreate(&sessionSettings, &m_defaultSession));
+
+        // Pull images required by the tests (no-op if already present).
+        for (const char* image : {"debian:latest", "hello-world:linux"})
+        {
+            WslcPullImageOptions pullOptions{};
+            pullOptions.uri = image;
+            PWSTR rawErrorMsg = nullptr;
```
OneBlue: nit: This can be simplified to: 

```
unique_cotaskmem_string errorMsg;
const auto hr = WslcSessionImagePull(m_defaultSession, &pullOptions, &errorMsg);

File: test/windows/WslcSdkTests.cpp [hunk 9]
```diff
@@ -0,0 +1,769 @@
...
+        VERIFY_SUCCEEDED(WslcContainerStart(container.get(), WSLC_CONTAINER_START_FLAG_NONE));
+
+        // Acquire and release the init process handle — we won't read its I/O.
+        {
+            UniqueProcess process;
+            VERIFY_SUCCEEDED(WslcContainerGetInitProcess(container.get(), &process));
+        }
+
+        // Stop the container gracefully (after the timeout).
+        VERIFY_SUCCEEDED(WslcContainerStop(container.get(), WSLC_SIGNAL_SIGTERM, 10));
```
OneBlue: I recommend sending SIGKILL so the test doesn't block for 10 seconds here

OneBlue (approved):
LGTM. Minor comments

test/windows/WslcSdkTests.cpp
JohnMcPMS: Does pull always cost us in the rate limit, or only when the image is net new?  These tests share the storage with the WSLA tests and should not be pulling any net new images over what it already does.

If the limit is an issue, have we considered moving to hosting our test containers in an artifact feed that we own?

OneBlue: >Does pull always cost us in the rate limit, or only when the image is net new? These tests share the storage with the WSLA tests and should not be pulling any net new images over what it already does.

That's a good question. I'm actually not sure.

> If the limit is an issue, have we considered moving to hosting our test containers in an artifact feed that we own?

We have, but for the short term @yao-msft is working on importing the test images via a nuget package. Long term I think a registry that we own would be the right path, but we'll need to authenticate to it, so that's a bigger effort

PR #14233, opened 2026-02-19
Create Container List command, tests, alias support

dkbennett:

First convergence PR, focusing on only the container list command and moving some files around.
List implementation shows general pattern for how others will be done.
Command alias support.

## PR Checklist

* Moved the earlier command implementation files that were all *Service or *Model to the \services folder. 
* Added Command support for aliases, with `container list` having the aliases `ls` and `ps` for parity with Docker CLI. Added tests for this.
* Moved the commonly used  session and container list functions into tasks and context data. Added tests for this.
* Renamed the help alias to '-h' to align with Docker

## Validation Steps Performed
* Ran tests, including new ones
* Verified list was outputting as expected, and the session was created and showed up in diag list (session list).
* Manually verified all arguments work as expected (Session argument is not yet implemented).


src/windows/wslc/commands/ContainerListCommand.cpp
```diff
@@ -0,0 +1,110 @@
...
+        for (const auto& container : containers)
+        {
+            PrintMessage(string::MultiByteToWide(container.Id));
+        }
+    }
+    else if (context.Args.Contains(ArgType::Format) && (string::IsEqual(context.Args.Get<ArgType::Format>(), L"json"), true))
+    {
+        auto json = ToJson(containers);
+        PrintMessage(string::MultiByteToWide(json));
+    }
```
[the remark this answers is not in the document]
dkbennett: Great catch copilot, I had put the comma for case insensitivity to true outside the check and it was still valid code.

src/windows/wslc/commands/ContainerListCommand.cpp
```diff
@@ -0,0 +1,110 @@
...
+using namespace wsl::shared;
+using namespace wsl::windows::wslc::execution;
+using namespace wsl::windows::wslc::task;
+
+namespace wsl::windows::wslc {
+// Container List Command
+std::vector<Argument> ContainerListCommand::GetArguments() const
+{
+    return {
+        Argument::Create(ArgType::ContainerId, std::nullopt, 25, L"Container names to list. Can specify multiple."),
```
OneBlue: Is 25 the maximum number of values ? What's the thinking for specifically picking 25 ?

src/windows/wslc/commands/ContainerListCommand.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    if (context.Args.Contains(ArgType::Quiet))
+    {
+        // Print only the container ids
+        for (const auto& container : containers)
+        {
+            PrintMessage(string::MultiByteToWide(container.Id));
+        }
+    }
+    else if (context.Args.Contains(ArgType::Format) && (string::IsEqual(context.Args.Get<ArgType::Format>(), L"json"), true))
```
OneBlue: We should compare `--format` with case sensitivity. If the caller passed `JSON`, then we should interepret as a template like docker does: 

```
$ docker container ls --format JSON
JSON
```

src/windows/wslc/commands/ContainerCommand.h
```diff
@@ -0,0 +1,49 @@
...
+--*/
+#pragma once
+#include "Command.h"
+
+namespace wsl::windows::wslc {
+// Root Container Command
+struct ContainerCommand final : public Command
+{
+    constexpr static std::wstring_view CommandName = L"container";
+    ContainerCommand(std::wstring parent) : Command(CommandName, parent)
```
OneBlue: `parent` should either be a const ref, or an rvalue ref to avoid copying the string (applies to other strings in this PR)

src/windows/wslc/core/Command.cpp
```diff
@@ -284,11 +297,20 @@ std::unique_ptr<Command> Command::FindSubCommand(Invocation& inv) const
 
     for (auto& command : commands)
     {
-        if (string::IsEqual(*itr, command->Name()))
+        if (string::IsEqual(*itr, command->Name(), true))
```
OneBlue: I don't think we should do case insensitive compares here

src/windows/wslc/core/Command.h
```diff
@@ -34,7 +34,10 @@ struct Command
     // The character used to split between commands and their parents in FullName.
     constexpr static wchar_t ParentSplitChar = L':';
 
-    Command(std::wstring_view name, std::wstring parent);
+    Command(std::wstring_view name, std::wstring parent) : Command(name, {}, parent)
+    {
+    }
+    Command(std::wstring_view name, std::vector<std::wstring_view> aliases, std::wstring parent);
```
OneBlue: `aliases` and `parent` are copied here

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -0,0 +1,47 @@
...
+    {
+        // TODO: Add session ID to the session options to open the specifed session.
+    }
+
+    context.Data.Add<Data::Session>(SessionService::CreateSession(options));
+}
+
+void GetContainers(CLIExecutionContext& context)
+{
+    THROW_HR_IF(E_NOT_SET, !context.Data.Contains(Data::Session));
```
OneBlue: Could we make the caller pass in the session instead ? That would give us a compile time guarantee that it's not missing

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -97,6 +97,24 @@ static void StopInternal(IWSLAContainer& container, int signal = WSLASignalNone,
...
+        return L"created";
+    case WSLA_CONTAINER_STATE::WslaContainerStateRunning:
+        return L"running";
+    case WSLA_CONTAINER_STATE::WslaContainerStateDeleted:
+        return L"stopped";
+    case WSLA_CONTAINER_STATE::WslaContainerStateExited:
+        return L"exited";
+    case WSLA_CONTAINER_STATE::WslaContainerStateInvalid:
+    default:
+        return L"invalid";
```
OneBlue: We should probably assert / fast_fail if we ever hit this

src/windows/wslc/commands/ContainerListCommand.cpp
dkbennett: Maximum number of container yes. I picked it arbitrarily as a large enough number that I think it will work fine for now. We can certainly talk about the limits we want later, its very easy to change this. We can make it infinite if we want, though I'd prefer limited that we can create a stress test for that large enough that nobody would conceivably hit it. If you want it bigger I can make it bigger but this is super easy to change or to make infinite if we really want.

src/windows/wslc/commands/ContainerListCommand.cpp
dkbennett: confirmed with design, we are intentionally case sensitive so I am changing that and adding test cases to ensure we dont regress on that later.

src/windows/wslc/tasks/ContainerTasks.cpp
dkbennett: Session is there, this is really just an assert to catch programming errors.

OneBlue: What I meant is that if a code path calls GetContainer() when `Data::Session` is unset, this will fail. If we pass in a session we can guarantee that the caller has opened the session before like: 

```
auto session = OpenSession();

GetContainers(context, session);
```

That way it's impossible to end up in a invalid state

src/windows/wslc/core/Command.cpp
dkbennett: This is a higher order design direction question that's also very easy to change later, so I want to leave it alone until we have confirmation on whether we should follow windows convention and be consistent with the operating system or be consistent with docker (which is not consistent with the operating system). Its easy enough to change true to false later if we decide it should be case sensitive for commands and arguments.

OneBlue: Agreed that this shouldn't be blocking for this PR, but if possible I'd like to get this right the first time on the first implementation so we don't need to update tests later. 

I have a strong opinion that case sensitive is the right approach, mostly because because there are case differencesin docker's command line (docker run -p and docker run -P do different things for instance). 

@craigloewen-msft: What do you think of this ? For context, WSL & docker's command line are case sensitive

src/windows/wslc/commands/RootCommand.cpp
```diff
@@ -14,19 +14,19 @@ Module Name:
...
-using namespace wsl::windows::common::wslutil;
 using namespace wsl::windows::wslc::execution;
 
 namespace wsl::windows::wslc {
 std::vector<std::unique_ptr<Command>> RootCommand::GetCommands() const
 {
     std::vector<std::unique_ptr<Command>> commands;
-    commands.reserve(2);
+    commands.reserve(3);
+    commands.push_back(std::make_unique<ContainerCommand>(FullName()));
```
AmelBawa-msft, line 27: Nit: Will we be adding all arguments here? If yes, I suggest that each command category (container, image, etc...) to have a method that each adds its own sub-commands just so it's easier to read, reduce code conflicts and size.

src/windows/wslc/core/ExecutionContextData.h
```diff
@@ -40,7 +43,8 @@ namespace details {
     {
     };
 
-    DEFINE_DATA_MAPPING(SessionId, std::wstring);
+    DEFINE_DATA_MAPPING(Session, wsl::windows::wslc::models::Session);
+    DEFINE_DATA_MAPPING(Containers, std::vector<wsl::windows::wslc::models::ContainerInformation>);
```
AmelBawa-msft, line 47: We may need to reason about this later on. ContainerInformation contains limited information about the container used in the List command. I don't have a better naming or suggestion for now, but now that we have the new CLI infra in place, I will try to better reason about it as I add support for other commands/options.

src/windows/wslc/services/ContainerService.h
AmelBawa-msft, line 1: I guess git did not detect this file as move?

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -0,0 +1,47 @@
...
+namespace wsl::windows::wslc::task {
+void CreateSession(CLIExecutionContext& context)
+{
+    std::optional<SessionOptions> options = std::nullopt;
+    if (context.Args.Contains(ArgType::SessionId))
+    {
+        // TODO: Add session ID to the session options to open the specified session.
+    }
+
+    context.Data.Add<Data::Session>(SessionService::CreateSession(options));
```
AmelBawa-msft, line 38: Nice! I like how things are getting together well with the code split, good example here 👍

AmelBawa-msft (approved):

src/windows/wslc/tasks/ContainerTasks.cpp
AmelBawa-msft: I could be wrong, but I think the intention of this file is to define methods that act as building blocks that assume that the caller is calling the building blocks in the right order. Similar to dependency injection in C#, you attempt to inject the session into your method trusting that the session has been previously registered.

Usage example:
```cpp

void ContainerListCommand::ExecuteInternal(CLIExecutionContext& context) const
{
    context << CreateSession << GetContainers;
    auto& containers = context.Data.Get<Data::Containers>();
    RenderOutput(containers);
}
```

I agree this model is inheritably (and by design) a runtime based, making it challenging to detect invalid code flows at compile time. Testing here is going to be a key to ensure all dependencies and the chain of tasks is not missing information.

dkbennett: Yes this is correct and the intent. We are intentionally not passing data around between these tasks. We are assuming the chain is constructed correctly and the throw is just there acting as an assert so we find it very quickly or if we have a bug in a previous method. I don't want to update the signature and constantly be passing arguments around and ending up with soup of data to manage, that's why we are using the context as the source of state. I understand that's a compile time detection tradeoff but prefer the simpler maintenance and cleaner code of the present design.

dkbennett: I am changing this to WI_ASSERT  which will compile out on release builds but otherwise catch programming errors if we get the composition incorrect.

src/windows/wslc/commands/RootCommand.cpp
dkbennett: It does, this is just because the root command has some commands from container also aliased to it. The container command laso has the container list defined under it. This is a clean way to do the aliasing so the same exact command is executed whether you do 'wslc ls' or 'wslc container ls'. Each command defines its subcommands in one place.

src/windows/wslc/core/ExecutionContextData.h
dkbennett: Names are easy to change as we need and add more data.

src/windows/wslc/services/ContainerService.h
dkbennett: No I changed the methods to static, that's why its listed. And added "ContainerStateToString()" to the containerservice.

src/windows/wslc/core/Command.cpp
dkbennett: Just feedback that we resolved this, case-sensitive we shall be, and I'm adding tests to ensure it.

src/windows/wslc/services/ContainerService.cpp
dkbennett: I updated this to a THROW_HR, if it ever does escape our observation and make it into the wild a throw hr would still be easy to find the problem without quite as jarring of a user experience as a fail fast.

src/windows/wslc/core/Command.h
dkbennett: changed parent to be a const reference, and aliases to be an rvalue reference, since aliases is always instantiated with a temporary vector, unnecessary copies removed!

src/windows/wslc/commands/ContainerCommand.h
dkbennett: changed parent and alias to remove unnecessary copies

src/windows/wslc/commands/RootCommand.cpp
dkbennett: Confirmed offline this concern has been resolved.

src/windows/wslc/commands/ContainerCommand.h
```diff
@@ -0,0 +1,49 @@
...
+
+protected:
+    void ExecuteInternal(CLIExecutionContext& context) const override;
+};
+
+// List Command
+struct ContainerListCommand final : public Command
+{
+    constexpr static std::wstring_view CommandName = L"list";
+    ContainerListCommand(std::wstring parent) : Command(CommandName, {L"ls", L"ps"}, parent)
```
OneBlue: Does this support aliasing outside the subcommand ? For instance here we'd want to support: 

```
wslc container list
wslc container ls|ps
wslc ps 
```

(this can be addressed in a future change)

src/windows/wslc/commands/ContainerListCommand.cpp
```diff
@@ -0,0 +1,114 @@
...
+    if (!context.Args.Contains(ArgType::All))
+    {
+        auto shouldRemove = [](const ContainerInformation& container) {
+            return container.State != WSLA_CONTAINER_STATE::WslaContainerStateRunning;
+        };
+        containers.erase(std::remove_if(containers.begin(), containers.end(), shouldRemove), containers.end());
+    }
+
+    // Filter by name if provided
+    if (context.Args.Contains(ArgType::ContainerId))
```
OneBlue: I'm just realizing this now, but what's the reason for this logic ? Docker doesn't seem to support anything like this. 

I also don't love the idea of the frontend doing matching on the container id's since the matching rules are not trivial

src/windows/wslc/commands/ContainerListCommand.cpp
```diff
@@ -0,0 +1,114 @@
...
+        for (const auto& container : containers)
+        {
+            PrintMessage(string::MultiByteToWide(container.Id));
+        }
+    }
+    // TODO: When we have more arguments that have CLI-specific validation, centralize the validation logic
+    // so they are validated prior to reaching command execution. We should check the format types during
+    // ArgumentValidation in the command and error if the user put in an invalid format type, then the
+    // command can safely assume the format type is valid here and doesn't need to check for it again.
+    else if (context.Args.Contains(ArgType::Format) && (string::IsEqual(context.Args.Get<ArgType::Format>(), L"json")))
```
OneBlue: We should fail if --format is present and its value is anything other than `json`. We should probably support templates at some point though

src/windows/wslc/core/Command.cpp
```diff
@@ -187,6 +189,17 @@ void Command::OutputHelp(const CommandException* exception) const
 
     infoOut << std::endl << std::endl;
 
+    if (!commandAliases.empty())
+    {
+        infoOut << Localization::WSLCCLI_AvailableCommandAliases();
+
+        for (const auto& commandAlias : commandAliases)
```
OneBlue: nit: We can use wsl::shared::string::Join() for this

src/windows/wslc/core/Command.h
```diff
@@ -52,6 +55,10 @@ struct Command
     {
         return m_fullName;
     }
+    const std::vector<std::wstring_view>& Aliases() const&
```
OneBlue: This this `&` meant to be here ?

dkbennett: nope, compiler didn't seem to care either, removed

src/windows/wslc/commands/ContainerCommand.h
dkbennett: Yes, it is the same command added to root and container, so `wslc list`, `wslc ls`, and `wslc ps` all work with this change.

src/windows/wslc/commands/ContainerListCommand.cpp
dkbennett: We followed the spec; if this needs to change we can adjust accordingly in a followup PR.

src/windows/wslc/commands/ContainerListCommand.cpp
dkbennett: Updated to add optional command argument validation with this command validating format is either 'json' or 'table' and updated the logic accordingly. Throws an exception if argument is bad, which results in a user friendly message and help display.

Edit: and added tests for this

src/windows/wslc/core/Command.cpp
dkbennett: Attempted to, but couldn't use the constexpr Join method into the output stream, so leaving it alone for now.

OneBlue (approved):

PR #14238, opened 2026-02-19
Move ExportContainer out of WSLASession into WSLAContainer

ptrivedi:


src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -557,6 +557,62 @@ void WSLAContainerImpl::Delete()
     m_state = WslaContainerStateDeleted;
 }
 
+void WSLAContainerImpl::Export(ULONG OutHandle)
+{
+    std::lock_guard<std::recursive_mutex> lock(m_lock);
```
OneBlue: nit: Could be simplified to:

```
    std::lock_guard lock(m_lock);
```

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -557,6 +557,62 @@ void WSLAContainerImpl::Delete()
...
+    }
+    CATCH_AND_THROW_DOCKER_USER_ERROR("Failed to export container '%hs'", m_id.c_str());
+
+    wil::unique_handle containerFileHandle{wsl::windows::common::wslutil::DuplicateHandleFromCallingProcess(ULongToHandle(OutHandle))};
+
+    wsl::windows::common::relay::MultiHandleWait io;
+
+    auto onCompleted = [&]() {
+        io.Cancel();
+        WSL_LOG("OnCompletedCalledForExport", TraceLoggingValue("OnCompletedCalledForExport", "Content"));
```
OneBlue: `TraceLoggingValue("OnCompletedCalledForExport", "Content")` looks like used to write a variable but now it's a constant string. 

We can probably get rid of `onCompleted` and do: 

```
        io.AddHandle(std::make_unique<RelayHandle<HTTPChunkBasedReadHandle>>(
                HandleWrapper{std::move(SocketCodePair.second)},
                HandleWrapper{std::move(containerFileHandle)}],
            MultiHandleWait::CancelOnCompleted));

```

Instead

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -557,6 +557,62 @@ void WSLAContainerImpl::Delete()
...
+    {
+        io.AddHandle(std::make_unique<ReadHandle>(HandleWrapper{std::move(SocketCodePair.second)}, std::move(accumulateError)));
+    }
+    else
+    {
+        io.AddHandle(std::make_unique<RelayHandle<HTTPChunkBasedReadHandle>>(
+            HandleWrapper{std::move(SocketCodePair.second)}, HandleWrapper{std::move(containerFileHandle), onCompleted}));
+    }
+
+    io.Run({});
```
OneBlue, line 598: We should cancel on session termination. now that #14239 is merged, the easiest way to do that is by getting the IO that way: 

````
    auto io = CreateIOContext();
```

This will automatically handle session & client process termination

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -557,6 +557,62 @@ void WSLAContainerImpl::Delete()
...
+        "Cannot export container '%hs', state: %i",
+        m_name.c_str(),
+        m_state);
+
+    std::pair<uint32_t, wil::unique_socket> SocketCodePair;
+    try
+    {
+        SocketCodePair = m_dockerClient.ExportContainer(m_id);
+    }
+    CATCH_AND_THROW_DOCKER_USER_ERROR("Failed to export container '%hs'", m_id.c_str());
```
OneBlue: No DockerHTTPException are thrown by `ExportContainer` so we can remove the try / catch here

test/windows/WSLATests.cpp
```diff
@@ -1130,50 +1131,53 @@ class WSLATests
...
-            ExpectImagePresent(*m_defaultSession, "hello-world:latest");
-            WSLAContainerLauncher launcher("hello-world:latest", "wsla-hello-world-container");
-            auto container = launcher.Launch(*m_defaultSession);
-            auto result = container.GetInitProcess().WaitAndCaptureOutput();
-            VERIFY_ARE_EQUAL(0, result.Code);
-            VERIFY_IS_TRUE(result.Output[1].find("Hello from Docker!") != std::string::npos);
-        }
 
-        // Try to export a non-existing container.
+        std::filesystem::path imageTarFile = std::filesystem::path{g_testDataPath} / L"HelloWorldExported.tar";
```
OneBlue: I recommend creating a sub scope for each test that way variable names can't collide across test cases

test/windows/WSLATests.cpp
```diff
@@ -1130,50 +1131,53 @@ class WSLATests
...
-            ValidateCOMErrorMessage(L"No such container: dummy");
+            ULONG outFile2 = HandleToULong(contTarFileHandle.get());
 
-            VERIFY_IS_TRUE(GetFileSizeEx(contTarFileHandle.get(), &fileSize));
-            VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, false);
+            container2.Get().Stop(WSLASignalSIGILL, 10);
+            container2.Get().Delete();
+            VERIFY_ARE_EQUAL(container2.Get().Export(outFile2), RPC_E_DISCONNECTED);
+
+            VERIFY_IS_TRUE(GetFileSizeEx(contTarFileHandle.get(), &fileSize3));
```
OneBlue: `contTarFileHandle` will likely be non-empty because of the previous test, I recommend creating a new file for this, or a pipe if we don't want to clean up the file once the test is done

src/windows/wslasession/WSLAContainer.cpp
ptrivedi: This will require some refactor as CreateIOContext is in WSLASession which is not available in WSLAContainer

test/windows/WSLATests.cpp
ptrivedi: There will be a pair each within the subscope. Can't subscope more than that before one block needs to use the container started by the previous block.

test/windows/WSLATests.cpp
ptrivedi: Not sure what you mean, containerTarFileHandle is being newly created above in this scope at line 1166

test/windows/WSLATests.cpp
```diff
@@ -1113,68 +1113,76 @@ class WSLATests
...
+                VERIFY_IS_TRUE(GetFileSizeEx(containerTarFileHandle.get(), &fileSize));
+                VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, true);
+            }
 
-            VERIFY_IS_TRUE(GetFileSizeEx(contTarFileHandle.get(), &fileSize));
-            VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, false);
+            // Load the exported container to verify it's valid.
+            {
+                std::filesystem::path containerTar = std::filesystem::path{g_testDataPath} / L"HelloWorldExported.tar";
+                wil::unique_handle containerTarFileHandle2{CreateFileW(
```
OneBlue: This can be renamed to ` containerTarFileHandle` now that the colliding variable is in a different scope (same for `imageTarFile3`, `fileSize3`, `container`, etc)

test/windows/WSLATests.cpp
OneBlue: Ok I figured it out, I mis-imprinted what the test does. 

Reading through it, I recommend changing: 

```
                VERIFY_ARE_EQUAL(fileSize3.QuadPart > 0, false);
```

Into:

```
                VERIFY_ARE_NOT_EQUAL(fileSize3.QuadPart, 0);
```

That should make this a bit easier to read

test/windows/WSLATests.cpp
```diff
@@ -1113,68 +1113,76 @@ class WSLATests
...
+                VERIFY_SUCCEEDED(container.Get().Export(HandleToULong(containerTarFileHandle.get())));
+                VERIFY_IS_TRUE(GetFileSizeEx(containerTarFileHandle.get(), &fileSize));
+                VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, true);
+            }
 
-            VERIFY_IS_TRUE(GetFileSizeEx(contTarFileHandle.get(), &fileSize));
-            VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, false);
+            // Load the exported container to verify it's valid.
+            {
+                std::filesystem::path containerTar = std::filesystem::path{g_testDataPath} / L"HelloWorldExported.tar";
```
OneBlue: This test should use a scope_exit to delete `HelloWorldExported.tar`  once done

test/windows/WSLATests.cpp
```diff
@@ -1113,68 +1113,76 @@ class WSLATests
...
+                // Verify that the image is in the list of images.
+                ExpectImagePresent(*m_defaultSession, "hello-world:latest");
+                WSLAContainerLauncher launcher("hello-world:latest", "wsla-hello-world-container");
+                auto container = launcher.Launch(*m_defaultSession);
+                auto result = container.GetInitProcess().WaitAndCaptureOutput();
+                VERIFY_ARE_EQUAL(0, result.Code);
+                VERIFY_IS_TRUE(result.Output[1].find("Hello from Docker!") != std::string::npos);
+
+                // Export the container to a tar file.
+                std::filesystem::path containerTar = std::filesystem::path{g_testDataPath} / L"HelloWorldExported.tar";
```
OneBlue: I don't recommend writing to `g_testDataPath` since that points to an installed nuget package. Instead writing to the current directory should be OK

ptrivedi: So does this mean that HelloWorldSaved. tar that is being used in LoadImage test would be available there?

OneBlue: It would not. Importing from that path is OK, but we shouldn't Export() to it (which is what this test case and the next one do)

test/windows/WSLATests.cpp
OneBlue: Looks like the latest iteration is missing a scope_exit to delete `HelloWorldExported.tar`

ptrivedi: SaveImage had it, Export was missing some. In the first subscope, we don't want HelloWorldExported.tar to be deleted, because the second subscope uses it. But added scope_exit for both HelloWorldExported.tar and HelloWorldExportError.tar in the second subscope. Good catch, thanks.

test/windows/WSLATests.cpp
```diff
@@ -1113,68 +1119,79 @@ class WSLATests
...
-            VERIFY_IS_FALSE(INVALID_HANDLE_VALUE == contTarFileHandle.get());
-            LARGE_INTEGER fileSize{};
-            VERIFY_IS_TRUE(GetFileSizeEx(contTarFileHandle.get(), &fileSize));
-            VERIFY_ARE_EQUAL(fileSize.QuadPart > 0, false);
+                std::filesystem::path imageTarFile = L"HelloWorldExportError.tar";
+                wil::unique_handle contTarFileHandle{CreateFileW(
+                    imageTarFile.c_str(), GENERIC_WRITE | GENERIC_READ, FILE_SHARE_READ, nullptr, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr)};
+                VERIFY_IS_FALSE(INVALID_HANDLE_VALUE == contTarFileHandle.get());
+                auto cleanfile =
+                    wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { LOG_IF_WIN32_BOOL_FALSE(DeleteFileW(imageTarFile.c_str())); });
```
OneBlue, line 1211: nit: I recommend moving this to the function's top scope, that way if something goes wrong in the previous scope, the file will be deleted

OneBlue (approved):
LGTM, one minor test comment. Feel free to merge once fixed

test/windows/WSLATests.cpp
ptrivedi: It is right after the CreateFile call though

OneBlue: Sorry wrong line, I was referring to the scope_exit for `HelloWorldExported.tar`

PR #14240, opened 2026-02-20
Converge Container Start/Stop/Run/Create into CLI model

dkbennett:

Porting over Create, Run, Stop, and Start container commands.
Added an argument validator for a common type (unint) for time and signal.
Added support for infinite limit on multi-input arguments.

## PR Checklist

* Migrated container commands Start, Stop, Create, and Run to the new model. The functionality of these should not be any different beyond extra argument validation.
* All arguments for these commands were added per the spec, many are not implemented but will get fleshed out over time and adjusted as needed.
* Added support for common argument validation, with unsigned integer being the first common validator added for --signal and --time arguments. This allows the commands to not duplicate validation logic and they can assume all relevant validation of the arguments has happened prior to execution. This also allows us to bubble up argument errors before starting execution flows and show the help.
* Added support for infinite values provided on arguments. A "NO_LIMIT" macro was added to represent this (-1 integer on limit). Updated tests and existing arguments as appropriate.
* Added CMake Unity build option to the wslc as a potential optimization in the future. It is not active and commented out. It's there for testing and finding what breaks when unity is enabled. We plan on using some form of Unity build at least for commands for full builds so we can have the performance benefit of large compilation units but have the incremental and maintainability value of more files.
* The PullImageProgress class and file was created from the previous ImageCommand.cpp. It is not new code, just refactored. It may show up in the diff as different but it was not changed from its previous form.

## Validation Steps Performed
* Unit tests added for command line validation and arguments.
* Manual validation for the commands themselves. Encountered some problems but the same problems were also happening with wsladiag so I do not believe these are regressions due to the migration to the new model. In any case these commands are still WIP and this update just opens the way to more iteration on them.


src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    default:
+        break;
+    }
+}
+} // namespace wsl::windows::wslc
+
+namespace wsl::windows::wslc::validation {
+
+void ValidateUInteger(const std::wstring& value, const std::wstring& argName)
```
OneBlue: Looks like this parsing is already done inside the command, could we only do it there so we don't parse the string to an integer twice ? 

Also for error handling, we can do: 

```
                THROW_USER_ERROR(wsl::shared::Localization::MessageInvalidInteger(input));
```

src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -34,16 +34,40 @@ Module Name:
...
+_(Pull,           "pull",                NO_ALIAS,          Kind::Value,       L"Image pull policy (always|missing|never) (default:never)") \
 _(Quiet,          "quiet",               L"q",              Kind::Flag,        L"Outputs the container IDs only") \
-_(Remove,         "remove",              L"rm",             Kind::Flag,        L"Remove the container after execution") \
+_(Remove,         "remove",              L"rm",             Kind::Flag,        L"Remove the container after it stops") \
+_(Scheme,         "scheme",              NO_ALIAS,          Kind::Value,       L"Use this scheme for registry connection") \
 _(SessionId,      "session",             NO_ALIAS,          Kind::Value,       Localization::WSLCCLI_SessionIdArgDescription()) \
-_(Verbose,        "verbose",             L"v",              Kind::Flag,        L"Output verbose details")
+_(Signal,         "signal",              L"s",              Kind::Value,       L"Signal to send (default: SIGKILL)") \
+_(Time,           "time",                L"t",              Kind::Value,       L"Time in seconds to wait before executing (default 5)") \
```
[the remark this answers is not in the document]
dkbennett: This is how it was spec'd, leaving it alone. These two should never be in the same command, and if we do it should be obvious and we'd fix it at that time.

src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -34,16 +34,40 @@ Module Name:
 // clang-format off
 #define WSLC_ARGUMENTS(_) \
 _(All,            "all",                 L"a",              Kind::Flag,        L"Show all regardless of state.") \
+_(Attach,         "attach",              L"a",              Kind::Flag,        Localization::WSLCCLI_AttachArgDescription()) \
```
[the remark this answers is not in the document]
dkbennett: As before, these should not be together and leaving the aliases as they are intended to be .

src/windows/wslc/services/PullImageCallback.cpp
```diff
@@ -0,0 +1,116 @@
...
+    {
+        line = std::format(L"{} '{}': {}s", status, id, current);
+    }
+    else
+    {
+        line = std::format(L"{} '{}'", status, id);
+    }
+
+    // Erase any previously written char on that line.
+    while (line.size() < info.dwSize.X)
```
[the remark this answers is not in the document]
dkbennett: This code is not changed and we plan on replacing it entirely in the near future, so I'm not going to spend time on it.

src/windows/wslc/services/PullImageCallback.cpp
```diff
@@ -0,0 +1,116 @@
...
+    THROW_IF_WIN32_BOOL_FALSE(GetConsoleScreenBufferInfo(GetStdHandle(STD_OUTPUT_HANDLE), &info));
+    return info;
+}
+
+std::wstring PullImageCallback::GenerateStatusLine(LPCSTR status, LPCSTR id, ULONGLONG current, ULONGLONG total, const CONSOLE_SCREEN_BUFFER_INFO& info)
+{
+    std::wstring line;
+    if (total != 0)
+    {
+        line = std::format(L"{} '{}': {}%", status, id, current * 100 / total);
```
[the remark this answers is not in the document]
dkbennett: This is temporary implementation prior to better unified progress reporting so leaving this alone. We will not ship with this.

src/windows/wslc/services/PullImageCallback.cpp
```diff
@@ -0,0 +1,116 @@
...
+    });
+}
+
+HRESULT PullImageCallback::OnProgress(LPCSTR status, LPCSTR id, ULONGLONG current, ULONGLONG total)
+{
+    try
+    {
+        if (id == nullptr || *id == '\0') // Print all 'global' statuses on their own line
+        {
+            wprintf(L"%hs\n", status);
```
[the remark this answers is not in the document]
dkbennett: Not changing PullImageCallback, it's a temporary progress solution that will soon be replaced.

src/windows/wslc/commands/ContainerStopCommand.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+namespace wsl::windows::wslc {
+// Container Stop Command
+std::vector<Argument> ContainerStopCommand::GetArguments() const
+{
+    return {
+        Argument::Create(ArgType::ContainerId, std::nullopt, NO_LIMIT),
+        Argument::Create(ArgType::All),
+        Argument::Create(ArgType::SessionId),
+        Argument::Create(ArgType::Signal, std::nullopt, std::nullopt, L"Signal to send (default: SIGTERM)"),
```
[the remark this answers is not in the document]
dkbennett: Per command is correct

src/windows/wslc/commands/ContainerStopCommand.cpp
```diff
@@ -0,0 +1,97 @@
...
+}
+
+void ContainerStopCommand::ExecuteInternal(CLIExecutionContext& context) const
+{
+    context << CreateSession;
+
+    auto containersToStop = context.Args.GetAll<ArgType::ContainerId>();
+    if (context.Args.Contains(ArgType::All))
+    {
+        // All overwrites any specified container IDs.
```
OneBlue: Shouldn't it be invalid to pass both container ids and --all ?

src/windows/wslc/commands/ContainerStopCommand.cpp
```diff
@@ -0,0 +1,97 @@
...
+        containersToStop.clear();
+        context << GetContainers;
+        const auto& allContainers = context.Data.Get<Data::Containers>();
+        for (const auto& container : allContainers)
```
OneBlue: If we want to support --all, we should implement it in the service because doing this on the frontend side creates multiple race conditions (containers being created after they're retried, or changing state after the CLI checks their state)

src/windows/wslc/services/PullImageCallback.h
```diff
@@ -0,0 +1,46 @@
...
+    NON_MOVABLE(ChangeTerminalMode);
+    ChangeTerminalMode(HANDLE console, bool cursorVisible);
+    ~ChangeTerminalMode();
+
+private:
+    HANDLE m_console{};
+    CONSOLE_CURSOR_INFO m_originalCursorInfo{};
+};
+
+class DECLSPEC_UUID("7A1D3376-835A-471A-8DC9-23653D9962D0") PullImageCallback
```
OneBlue, line 33: The original implementation of this class had a TODO comment: 

``` 
    // TODO: Handle terminal resizes.
```

I think we should keep it, because the progress bars will break during a terminal size

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -44,4 +83,58 @@ void GetContainers(CLIExecutionContext& context)
...
+
+    context.Data.Add<Data::RunContainerOptions>(std::move(options));
+}
+
+void CreateContainer(CLIExecutionContext& context)
+{
+    WI_ASSERT(context.Data.Contains(Data::Session));
+    WI_ASSERT(context.Args.Contains(ArgType::ImageId));
+    WI_ASSERT(context.Data.Contains(Data::CreateContainerOptions));
+    PullImageCallback callback;
```
OneBlue, line 120: We should only instantiate `callback` if we actually need to do a pull since it changes the terminal mode

OneBlue: (Same for below)

src/windows/wslc/arguments/ArgumentValidation.cpp
dkbennett: The difference is that this is happening during validation of arguments, before we ever even begin command execution. If there's invalid argument input at that time, we throw argument exception so we can have the help display path and the message indicates which argument had the data and offending data. Once we start execution we know the data is valid for what the CLI needs.

Yes we do call it twice for every valid integer, and that is not ideal, but I do not want to try to alter the data at validation time, and the code is simpler not having any further mechanism for storing or retrieving processed arguments when at this time we only have a couple integers. If we get more expensive validation or more complex evaluations then I think we'd solve that problem then. For now I'd like to keep it simple and don't think saving an integer parse is worth the trouble.

src/windows/wslc/commands/ContainerStopCommand.cpp
dkbennett: I'm going to just remove --all and then ask for clariification. If we do need to support it the CLI probably doesn't need to be the one doing the filtering or matching and that is best in the service as you say.

src/windows/wslc/commands/ContainerStopCommand.cpp
dkbennett: I can see it both ways. Silently ignoring when all is specified is one way to go, or we can make it mutually exclusive an an error, but we have other concerns with --all functionality here so I am going to remove it pending clarification.

src/windows/wslc/tasks/ContainerTasks.cpp
dkbennett: We should fix that in another PR. as that is how it worked in the code before and I'm just translating. Further refinement can happen after the migration to the new model is complete. This needs to be fixed in the container service code which I do not want to touch if I can help it. I want to get this ported over so work can continue.

src/windows/wslc/services/PullImageCallback.h
dkbennett: Re-added the comment from wsladiag.

src/windows/wslc/commands/ContainerStopCommand.cpp
AmelBawa-msft: @OneBlue could you please elaborate a bit more on this? I’m not seeing the race condition in this method.

That said, I believe the service should be responsible for ensuring thread safety so that the client doesn’t need to be aware of or manage those concerns.

From my perspective, handling --all on the client side with a simple loop makes sense. If we later decide to process it using multiple threads, the service layer should handle synchronization (e.g., with mutexes where appropriate) to ensure safe execution.

src/windows/wslc/arguments/ArgumentValidation.cpp
OneBlue: >  If there's invalid argument input at that time, we throw argument exception so we can have the help display path and the message indicates which argument had the data and offending data.

We can do that regardless of where the exception is thrown. 

Given that,  I would recommend something like this: 

```
template <typename T>
T IntegerFromString(const std::string& input, const std::string& argumentName)
{
    T value{};
    auto result = std::from_chars(input.c_str(), input.c_str() + utf8value.size(), value);


   if (result.ec == std::errc::invalid_argument)
   {
       THROW_USER_ERROR(std::format("Invalid integer '{}' for argument: '{}', input, ArgumentName);
   }

    return value;
}

void ContainerStopCommand::ExecuteInternal(CLIExecutionContext& context) const
{
[...]
        options.Signal = IntegerFromString<decltype(options.Signal)>(context.Args.Get<ArgType::Signal>(), "--signal");
```

Advantages of this approach: 

- The string is only parsed once
- The correct exception & error messages are guaranteed to be thrown (doesn't required the implementor to remember to add a validation in a different file)
- Works for all integers types (the current implementation is hardcoded for int, but in practice we'll need signed integers, 16, 32 and 64 bits integers)

src/windows/wslc/commands/ContainerStopCommand.cpp
OneBlue: > @OneBlue could you please elaborate a bit more on this? I’m not seeing the race condition in this method.

By the time we call `ContainerService::Stop()`, the state of the containers might be stale, since the containers are not locked and other callers can change their state. 

Operations that change the state of a container based on their current state should happen while the container is locked, which can only happen in the service

src/windows/wslc/arguments/ArgumentValidation.cpp
OneBlue: Overall I don't recommend separating validating from processing, because in my experience this almost always ends up duplicating the same logic in two different places, and if those two ever get out of sync, this will lead to bugs and unexpected behaviors

src/windows/wslc/tasks/ContainerTasks.cpp
OneBlue: Ok that's fair

src/windows/wslc/services/ContainerModel.h
```diff
@@ -17,8 +17,11 @@ Module Name:
 #include <wslservice.h>
 #include <wslaservice.h>
 #include <docker_schema.h>
+#include <unordered_map>
```
AmelBawa-msft: Nit: Can added `#include` be in the cpp file instead?

AmelBawa-msft (approved):

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <algorithm>
+
+namespace wsl::windows::wslc {
+// Common argument validation that occurs across multiple commands.
+void Argument::Validate(const ArgMap& execArgs) const
+{
+    switch (m_argType)
+    {
+    case ArgType::Signal:
+        validation::ValidateUInteger(execArgs.GetAll<ArgType::Signal>(), m_name);
```
OneBlue: Unfortunately we can't assume that signal will be an integer. For instance the user can do this: 

```
docker stop $(docker run -itd debian:latest) -s SIGKILL
```

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -14,19 +14,58 @@ Module Name:
...
+        {
+            options.TTY = true;
+        }
+
+        if (context.Args.Contains(ArgType::Interactive))
+        {
+            options.Interactive = true;
+        }
+
+        if (context.Args.Contains(ArgType::Command))
```
OneBlue, line 53: What's the thinking for having both `Command` and `ForwardArgs` here ? Shouldn't we only have `ForwardArgs` ?

dkbennett: Following the docker paradigm:
docker container run [OPTIONS] IMAGE [COMMAND] [ARG...]

We build help and documentation dynamically, and while functionally these may be represented as a single vector of arguments, that doesn't really change the fact that the command is a distinct argument vs its args.

src/windows/wslc/arguments/ArgumentValidation.cpp
dkbennett: That's ok, string name is actually how I had originally implemented it, but the source convergence code implemented it as an integer input only. I'm happy to make this change to allow dual inputs, but that is best as a bugfix improvement in a separate PR.

src/windows/wslc/arguments/ArgumentValidation.cpp
dkbennett: We believe there is value in the principle of separation of concerns. Validation of inputs, where we know they are incorrect, can be done prior to executing any potentially system-altering code rather than mid-execution, and it is good practice to do so. Mid execution errors cannot be entirely avoided but when we can avoid them we should.

While it may be less efficient to do so, the validation code and the conversion code could be shared to avoid duplication/out-of-sync issue, which is how we will implement it when the logic becomes more complex (due to having both string or integer input representing an integer). 

I want to leave this as-is in the interest of progressing the conversion work. This code needs to be adjusted in a subsequent PR as a bugfix for the string input possibility as well.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -0,0 +1,57 @@
...
+    for (const auto& value : values)
+    {
+        try
+        {
+            [[maybe_unused]] auto intValue = std::stoul(value);
+        }
+        catch (...)
+        {
+            throw ArgumentException(L"Invalid " + argName + L" argument value: " + value);
+        }
```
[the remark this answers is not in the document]
dkbennett: This validation is ok for now, it should be obvious to the user, but this is a good suggestion.

src/windows/wslc/commands/ContainerStopCommand.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+void ContainerStopCommand::ExecuteInternal(CLIExecutionContext& context) const
+{
+    context << CreateSession;
+
+    auto containersToStop = context.Args.GetAll<ArgType::ContainerId>();
+    StopContainerOptions options;
+    if (context.Args.Contains(ArgType::Signal))
+    {
+        options.Signal = std::stoul(context.Args.Get<ArgType::Signal>());
```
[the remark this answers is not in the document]
dkbennett: This is more of a problem with the service accepting a broader integer than it should rather than the conversion being incorrect. The enum is in fact unsigned, so this is correct. Fixing the service to accept ULONG instead.

src/windows/wslc/arguments/ArgumentValidation.cpp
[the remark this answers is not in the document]
dkbennett: Actually fixing this differently because of the timeout change to LONGLONG.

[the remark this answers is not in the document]
dkbennett: Fixed with a variation of OneBlue's suggestion for templated validation. This version uses the same conversion code to retrieve the value as used for validation of the value, only the validation step loops over the potential vector of arguments to validate. Individual retrieval of converted value can be called individually and is done so by the methods to get the correct integer type.

OneBlue (approved):
Change looks good. I still think that we avoid parsing non-string command line arguments twice, but that's not worth blocking the change over. 

Let's merge this and re-evaluate once we have most the API surface implemented & test coverage

src/windows/wslc/tasks/ContainerTasks.cpp
OneBlue: Turns out that this is what was hiding the cursor in containers shells, since the pull callback disables it.

PR #14250, opened 2026-02-23
Fix AttachDisk LUN counter leak on AddVhd failure

benhillis:
This PR resolves an issue where if LUN numbers would be wasted if adding the VHD to the VM fails. It also adjusts the LUN allocation logic to support reuse of LUN numbers of disks are removed from the VM.

### DIFF
--- a/src/windows/wslaservice/exe/HcsVirtualMachine.cpp
+++ b/src/windows/wslaservice/exe/HcsVirtualMachine.cpp
@@ -193,8 +193,8 @@ HcsVirtualMachine::HcsVirtualMachine(_In_ const WSLA_SESSION_SETTINGS* Settings)
     hcs::Scsi scsiController{};
     if (!FeatureEnabled(WslaFeatureFlagsPmemVhds))
     {
-        auto attachScsiDisk = [&](PCWSTR path, ULONG& lun) {
-            lun = m_nextLun++;
+        auto attachScsiDisk = [&](PCWSTR path) {
+            const ULONG lun = AllocateLun();
             hcs::Attachment disk{};
             disk.Type = hcs::AttachmentType::VirtualDisk;
             disk.Path = path;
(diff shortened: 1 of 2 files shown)


src/windows/wslaservice/exe/HcsVirtualMachine.cpp
```diff
@@ -478,21 +478,25 @@ try
...
     }
     else
     {
         THROW_IF_FAILED(result);
     }
 
-    m_attachedDisks.emplace(*Lun, std::move(disk));
+    m_attachedDisks.emplace(lun, std::move(disk));
+
+    // Only increment after AddVhd succeeds to avoid leaking LUN numbers on failure.
```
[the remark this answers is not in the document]
benhillis: I'm going to adjust this a bit.

src/windows/wslaservice/exe/HcsVirtualMachine.cpp
```diff
@@ -466,6 +464,16 @@ try
     std::lock_guard lock(m_lock);
 
     DiskInfo disk{Path};
+    const ULONG lun = AllocateLun();
```
OneBlue: nit: Having the same variable name with different casing can be a bit confusing. Could we rename to something like `allocatedLun` ?

src/windows/wslaservice/exe/HcsVirtualMachine.cpp
```diff
@@ -513,6 +522,8 @@ try
         hcs::RevokeVmAccess(m_vmIdString.c_str(), it->second.Path.c_str());
     }
 
+    FreeLun(Lun);
```
OneBlue: nit: We might want to release the Lun() as soon as `RemoveScsiDisk()`  returns so the lun isn't leaked if `RevokeVmAccess()` fails

OneBlue (approved):

src/windows/wslaservice/exe/HcsVirtualMachine.cpp
benhillis: good call

PR #14255, opened 2026-02-23
Fix print-fatal-signals being immediately disabled after enabling

benhillis:
The code wrote '1' to enable fatal signal logging, then immediately wrote '0' to the same sysctl, disabling it. Remove the second write so fatal signal logging actually stays enabled as intended.

### DIFF
--- a/src/linux/init/WSLAInit.cpp
+++ b/src/linux/init/WSLAInit.cpp
@@ -914,6 +914,7 @@ int WSLAEntryPoint(int Argc, char* Argv[])
     //
     // Enable dump collection when processes crash.
     //
+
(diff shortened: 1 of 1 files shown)


src/linux/init/WSLAInit.cpp
```diff
@@ -925,11 +926,6 @@ int WSLAEntryPoint(int Argc, char* Argv[])
         return -1;
     }
 
-    if (WriteToFile("/proc/sys/kernel/print-fatal-signals", "0\n") < 0)
```
OneBlue, line 928: Ah ! That looks a bad copy-paste from me. Great catch

OneBlue (approved):

PR #14257, opened 2026-02-23
Fix ContainerTrackingReference move constructor double-unregister

benhillis:
The defaulted move constructor copies m_tracker and m_id but doesn't null the source's m_tracker. Both the moved-from and moved-to objects then call Reset() in their destructors, causing a double-unregister from the event tracker. This can corrupt the callback vector.

The move assignment operator already handled this correctly by nulling the source. Apply the same pattern to the move constructor.

### DIFF
--- a/src/windows/wslasession/ContainerEventTracker.cpp
+++ b/src/windows/wslasession/ContainerEventTracker.cpp
@@ -48,6 +48,13 @@ void ContainerEventTracker::ContainerTrackingReference::Reset()
     }
 }
 
+ContainerEventTracker::ContainerTrackingReference::ContainerTrackingReference(ContainerTrackingReference&& other) noexcept :
+    m_id(other.m_id), m_tracker(other.m_tracker)
+{
+    other.m_tracker = nullptr;
+    other.m_id = {};
+}
+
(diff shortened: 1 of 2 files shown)


src/windows/wslasession/ContainerEventTracker.cpp
```diff
@@ -48,6 +48,13 @@ void ContainerEventTracker::ContainerTrackingReference::Reset()
     }
 }
 
+ContainerEventTracker::ContainerTrackingReference::ContainerTrackingReference(ContainerTrackingReference&& other) noexcept :
+    m_id(other.m_id), m_tracker(other.m_tracker)
+{
+    other.m_tracker = nullptr;
```
OneBlue, line 54: That's a great catch ! Did you see a repro of this causing corruption ? I'd except that the double free() would have hit on the asserts() that we have in the event tracker

OneBlue (approved):

src/windows/wslasession/ContainerEventTracker.cpp
benhillis: I think you are right. Nope found this by reviewing.

PR #14258, opened 2026-02-23
Switch GetExitEvent COM interface to use handle marshalling and fix potential handle truncation in ClientRunningWSLAProcess.

benhillis:
This change updates the GetExitEvent API to use system handle marshalling since the handle type is always known (an event handle). It also resolves a minor potential issue with casting HANDLE values to a ULONG in the GetStdHandle method.

### DIFF
--- a/src/windows/WslcSDK/wslcsdk.cpp
+++ b/src/windows/WslcSDK/wslcsdk.cpp
@@ -645,18 +645,7 @@ try
     RETURN_HR_IF_NULL(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), internalType->process);
     RETURN_HR_IF_NULL(E_POINTER, exitEvent);
 
-    *exitEvent = nullptr;
-
-    ULONG ulongHandle = 0;
-
-    HRESULT hr = internalType->process->GetExitEvent(&ulongHandle);
-
-    if (SUCCEEDED_LOG(hr))
-    {
-        *exitEvent = ULongToHandle(ulongHandle);
-    }
-
-    return hr;
+    return internalType->process->GetExitEvent(exitEvent);
 }
 CATCH_RETURN();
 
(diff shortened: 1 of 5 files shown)


src/windows/wslasession/WSLAProcess.cpp
```diff
@@ -31,10 +31,10 @@ try
 }
 CATCH_RETURN();
 
-HRESULT WSLAProcess::GetExitEvent(ULONG* Event)
+HRESULT WSLAProcess::GetExitEvent(HANDLE* Event)
 try
 {
-    *Event = HandleToUlong(common::wslutil::DuplicateHandleToCallingProcess(m_control->GetExitEvent().get(), SYNCHRONIZE));
```
OneBlue, line 37: Can we control the access rights of the event handle we return ? I like that the handle that we return to the caller can only be waited for (an evil caller won't be able to set the event for instance)

benhillis: Good call. I think the best way to do that is going to be to duplicate the handle.

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -645,18 +645,7 @@ try
...
-
-    HRESULT hr = internalType->process->GetExitEvent(&ulongHandle);
-
-    if (SUCCEEDED_LOG(hr))
-    {
-        *exitEvent = ULongToHandle(ulongHandle);
-    }
-
-    return hr;
+    return internalType->process->GetExitEvent(exitEvent);
```
OneBlue, line 648: nit: I did like the zeroing of the `exitEvent` by default

OneBlue (approved):

src/windows/WslcSDK/wslcsdk.cpp
benhillis: I tend to think that output parameters should not be modified unless the API returns success. We can always tweak this in a follow-up.

PR #14260, opened 2026-02-24
Use LoadImage for test image setup

yao-msft:
Use load image for test image setup. Skip pull image test if rate limited.

## PR Checklist

### DIFF
--- a/test/windows/WSLATests.cpp
+++ b/test/windows/WSLATests.cpp
@@ -2161,6 +2161,30 @@ class WSLATests
             ValidateProcessOutput(process, {}, 0);
         }
 
+        // Validate that custom DNS options are correctly wired.
+        {
+            WSLAContainerLauncher launcher(
+                "debian:latest", "test-dns-options", {"/bin/grep", "-iF", "timeout:1", "/etc/resolv.conf"});
+
+            launcher.SetDnsOptions({"timeout:1"});
+
+            auto container = launcher.Launch(*m_defaultSession);
+            auto process = container.GetInitProcess();
+            ValidateProcessOutput(process, {}, 0);
+        }
+
+        // Validate that multiple DNS options are correctly wired.
+        {
+            WSLAContainerLauncher launcher(
+                "debian:latest", "test-dns-options-multiple", {"/bin/grep", "-iF", "timeout:2", "/etc/resolv.conf"});
+
+            launcher.SetDnsOptions({"timeout:1", "timeout:2"});
+
+            auto container = launcher.Launch(*m_defaultSession);
+            auto process = container.GetInitProcess();
+            ValidateProcessOutput(process, {}, 0);
+        }
+
         // Validate that the username is correctly wired.
         {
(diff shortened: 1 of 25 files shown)


test/windows/WSLATests.cpp
```diff
@@ -329,7 +342,24 @@ class WSLATests
...
-            VERIFY_SUCCEEDED(m_defaultSession->PullImage("hello-world:linux", nullptr, nullptr));
+            HRESULT pullResult = m_defaultSession->PullImage("hello-world:linux", nullptr, nullptr);
+
+            // Skip test if error is due to rate limit.
+            if (pullResult == E_FAIL)
+            {
+                auto comError = wsl::windows::common::wslutil::GetCOMErrorInfo();
+                if (comError.has_value())
+                {
+                    auto tooManyRequests = wil::make_cotaskmem_string(L"toomanyrequests");
```
benhillis: why not just do a std::wstring here so you can just use ==?

benhillis: if (std::wstring_view(L"toomanyrequests") == comError->Message.get())

test/windows/WSLATests.cpp
```diff
@@ -41,6 +41,19 @@ class WSLATests
     wil::com_ptr<IWSLASession> m_defaultSession;
     static inline auto c_testSessionName = L"wsla-test";
 
+    void LoadTestImage(PCWSTR imageName)
+    {
+        std::filesystem::path imageTar = std::filesystem::path{g_testDataPath} / imageName;
+        wil::unique_handle imageTarFileHandle{
```
benhillis: wil::unique_hfile

Then you can just use THROW_LAST_ERROR_IF(!imageTarFileHandle)

test/windows/WSLATests.cpp
```diff
@@ -41,6 +41,19 @@ class WSLATests
     wil::com_ptr<IWSLASession> m_defaultSession;
     static inline auto c_testSessionName = L"wsla-test";
 
+    void LoadTestImage(PCWSTR imageName)
+    {
+        std::filesystem::path imageTar = std::filesystem::path{g_testDataPath} / imageName;
```
benhillis: maybe imagePath and the handle can be called imageFile

benhillis (changes requested):
a few minor comments, we need to get this in since it's currently blocking everything.

test/windows/WSLATests.cpp
yao-msft: The error string is very long, "you've reached limit of unauthenticated pulls .... " and the core substing I want t match is "toomanyrequests", so I'm using substring search, not exact == match

benhillis: ok that's fair, still we don't need to allocate a cotask string here.

benhillis: Just change to
```
if (wcsstr(comError->Message.get(), L"toomanyrequests") != nullptr)
```

benhillis (approved):

PR #14264, opened 2026-02-24
CLI - Update Signal argument validation to accept string or integer, map to enum

dkbennett:
Updates the Signal argument validation to validate against the WSLASignal enum:
* Accepts case-insensitive string representation ("SIGKILL" or "sigkill" or any combination)
* Accepts with or without the prefix "SIG", so "KILL" or "kill" also works.
* Accepts integer representation "9" as long as it is within the bounds of the enum and maps to an actual signal.
* Value extraction returns the actual WSLASignal enum, not an integer.
* Updated Stop and Kill inputs in the services wrapper to take the WSLASignal enum directly.
* Added tests for all of the above and the Stop command line parsing examples.

## PR Checklist

Does what the summary says. Also refactored the validation of integer template to the header, which is where it should have been.

## Validation Steps Performed
Added tests to ensure argument validation is correct as well as the command line parsing of the stop command tests.


src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -43,28 +44,62 @@ void Argument::Validate(const ArgMap& execArgs) const
...
     }
 }
 
-template <typename T>
-T GetIntegerFromString(const std::wstring& value, const std::wstring& argName)
+// Convert string to WSLASignal enum - accepts either signal name (e.g., "SIGKILL") or number (e.g., "9")
+WSLASignal GetWSLASignalFromString(const std::wstring& input, const std::wstring& argName)
 {
-    std::string narrowValue = string::WideToMultiByte(value);
+    constexpr int MIN_SIGNAL = 1;  // WSLASignalSIGHUP
```
AmelBawa-msft: super nit: Can we assign it directly to the min enum? Same for max.
```
    constexpr int MIN_SIGNAL = WSLASignalSIGHUP;
```

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -43,28 +44,62 @@ void Argument::Validate(const ArgMap& execArgs) const
...
+        {
+            return signalValue;
+        }
+    }
 
+    // User may have input an integer representation instead.
+    int signalValue = GetIntegerFromString<int>(input, argName);
+    if (signalValue < MIN_SIGNAL || signalValue > MAX_SIGNAL)
+    {
+        throw ArgumentException(L"Invalid " + argName + L" value: " + input + L" is out of valid range (1-31)");
```
AmelBawa-msft: nit: 

        throw ArgumentException(L"Invalid " + argName + L" value: " + input + L" is out of valid range (" MIN_SIGNAL + L"-" + MAX_SIGNAL + L")");

AmelBawa-msft (approved):

src/windows/wslc/arguments/ArgumentValidation.cpp
dkbennett: Another good one, ty!

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -43,28 +44,62 @@ void Argument::Validate(const ArgMap& execArgs) const
...
+    }
+
+    return static_cast<WSLASignal>(signalValue);
+}
```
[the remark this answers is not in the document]
dkbennett: Good suggestion copilot, I've updated it with a more specific message if it fails both the map and integer conversion

src/windows/wslc/arguments/ArgumentValidation.cpp
dkbennett: Nothing nit about it, it's a fantastic suggestion that future proofs us from enum expansion, thank you!  Though the text map will need to be updated if the enum expands, at least the integer will just work.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -43,28 +44,77 @@ void Argument::Validate(const ArgMap& execArgs) const
...
+    for (const auto& [signalName, signalValue] : SignalMap)
+    {
+        if (IsEqual(normalizedInput, signalName, true))
+        {
+            return signalValue;
+        }
+    }
+
+    // User may have input an integer representation instead.
+    int signalValue;
```
OneBlue: Even though that's not causing any issue here, I recommend always initializing stack variables that don't have default constructor (`int signalValue{};` ). 

This has been one of our top 3 biggest sources of bugs in the past

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -43,28 +44,77 @@ void Argument::Validate(const ArgMap& execArgs) const
...
+    try
+    {
+        signalValue = GetIntegerFromString<int>(input, argName);
+    }
+    // If it fails to be converted give a better user message than just the integer conversion
+    // failure since we also know it failed to be found in the map.
+    catch (ArgumentException)
+    {
+        throw ArgumentException(
+            L"Invalid " + argName + L" value: " + input +
```
OneBlue: I recommend using std::format for this kind of string formatting: 

```
throw ArgumentException(std::format(L"Invalid {} value {}", argName , input));
```

This will also make it easier to move to localized strings in the future 

(Same below)

OneBlue (approved):
LGTM, minor comments

src/windows/wslc/arguments/ArgumentValidation.cpp
dkbennett: Adjusted all validation messages to use format

src/windows/wslc/arguments/ArgumentValidation.cpp
dkbennett: Fixed

PR #14267, opened 2026-02-25
Use latest version of Visual Studio if multiple versions are present

benhillis:
#14160 broke cmake files generation if multiple versions of Visual Studio are installed. This change adds the "-latest" flag so the most recent version is used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,7 +262,7 @@ endif()
 # N.B. The version is set to VS2022 to ensure local runs match pipeline behavior
 # Require that Clang be installed in the product to potentially de-deduplicate
 execute_process(
-    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -version "[17.0,19.0)" -products * -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath -prerelease
+    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -prerelease -latest -version "[17.0,19.0)" -products * -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath
     OUTPUT_VARIABLE VS_INSTALL_DIR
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -262,7 +262,7 @@ endif()
 # N.B. The version is set to VS2022 to ensure local runs match pipeline behavior
 # Require that Clang be installed in the product to potentially de-deduplicate
 execute_process(
-    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -version "[17.0,19.0)" -products * -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath -prerelease
+    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -prerelease -latest -version "[17.0,19.0)" -products * -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath
```
OneBlue, line 265: I'd rather do that right now, because doing this will cause formatSource.ps1 to use the latest VS's clang-format, which will produce a result that's incompatible with our current pipeline. 

I think the right long term solution is to upgrade our pipeline to vs 2026, but I'd prefer to do that after feature/wsl-for-apps is merged so we don't end up in merge hell

benhillis: ok, we need to revert #14160 then, do you want to do that?

benhillis: there are not onebranch runners that have vs2026 yet, so upgrading the pipeline isn't an option.

benhillis: https://github.com/microsoft/WSL/pull/14268

OneBlue: How about we cherry-pick #14222 on master for now ?

OneBlue: That should give us the best of both worlds until the pipeline can run vs 2026

PR #14268, opened 2026-02-25
Revert "Enable detection cmake script for Visual Studio 2026 and pre-release versions of Visual Studio"

benhillis:
Reverts microsoft/WSL#14160

This causes issues when multiple versions of Visual Studio are installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,7 +262,7 @@ endif()
 # N.B. The version is set to VS2022 to ensure local runs match pipeline behavior
 # Require that Clang be installed in the product to potentially de-deduplicate
 execute_process(
-    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -version "[17.0,19.0)" -products * -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath -prerelease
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -262,7 +262,7 @@ endif()
 # N.B. The version is set to VS2022 to ensure local runs match pipeline behavior
 # Require that Clang be installed in the product to potentially de-deduplicate
 execute_process(
-    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -version "[17.0,19.0)" -products * -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath -prerelease
+    COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -version "[17.0,18.0)" -products * -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath
```
benhillis, line 265: I could be convinced to add -prerelease back, but I'm not sure there even are prerelease versions of VS2022 anymore...

OneBlue (approved):
Sad that we have to revert but given that there's no support for VS 2026 in our pipeline at the moment I think this is the right thing to do for now

PR #14269, opened 2026-02-25
Add support for parrallel operations in WSLASession

OneBlue:

This change switches WSLASession to use a read-write lock so that operations that don't change the state of the session can run at the same time. 

To do this, this change introduces a new `SharedMutex` class, which allows shared, exclusive, and atomic shared -> exclusive upgrades.

Considered alternatives: 

- `std::shared_mutex` doesn't support atomic exclusive upgrade
- `wil::srw_lock` doesn't support atomic exclusive upgrade
- `boost shared_mutex` would work well but requires linking with boost libraries 

## PR Checklist


benhillis:
This looks reasonable to me - @Brian-Perkins do you have any concern with an approach like this?

src/windows/common/SharedMutex.h
```diff
@@ -0,0 +1,76 @@
...
+        void ReleaseExclusive()
+        {
+            WI_ASSERT(m_exclusive.owns_lock());
+
+            m_exclusive.unlock();
+        }
+
+    private:
+        SharedMutex<TMutex>& m_mutex;
+        std::unique_lock<TMutex> m_shared;
```
OneBlue: will change to a std::shared_mutex so multiple readers can actually lock at the same time

src/windows/common/SharedMutex.h
```diff
@@ -0,0 +1,77 @@
...
+
+        LockInstance(SharedMutex<TMutex>& mutex, std::shared_lock<std::shared_mutex>&& shared) :
+            m_mutex(mutex), m_shared(std::move(shared))
+        {
+        }
+
+        void UpgradeToExclusive()
+        {
+            m_exclusive = std::unique_lock<TMutex>{m_mutex.m_exclusive};
+        }
```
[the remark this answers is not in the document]
OneBlue: That's a good point. I need to rethink this

benhillis:
Maybe we should just switch to a wil::srwlock for now and replace with a different lock in a future change? That would have some benefit from the current approach already.

OneBlue:
> Maybe we should just switch to a wil::srwlock for now and replace with a different lock in a future change? That would have some benefit from the current approach already.

Yeah doing a bit more research it's looking like implementing SharedMutex will be quite a bit more complex than that. I'll just go with wil::srwlock for now and we can bring back the SharedMutex approach later if needed

OneBlue:
https://github.com/microsoft/WSL/pull/14280/ is required to make this change stable

src/windows/wslasession/WSLASession.h
```diff
@@ -86,7 +86,6 @@ class DECLSPEC_UUID("4877FEFC-4977-4929-A958-9F36AA1892A4") WSLASession
     IFACEMETHOD(UnmapVmPort)(_In_ int Family, _In_ short WindowsPort, _In_ short LinuxPort) override;
 
     common::relay::MultiHandleWait CreateIOContext();
-    WSLAVirtualMachine& GetVirtualMachine();
```
OneBlue, line 89: I decided to remove this because this method required the lock held, which doesn't work now that the session lock isn't recursive. Passing in the virtualmachine like we do for other fields is the simplest way to solve this

benhillis (approved):

PR #14274, opened 2026-02-25
Implement more SDK APIs

JohnMcPMS:
Implements a few more WLSC SDK APIs targeted at upcoming demos.

## PR Checklist

Implements `WslcContainerSettingsSetNetworkingMode`, `WslcContainerSettingsSetPortMapping`, `WslcContainerSettingsSetVolumes` (was `WslcContainerSettingsAddVolume`), `WslcProcessSettingsSetEnvVariables`, and `WslcSessionImageLoad`. Adds tests for same.

### SDK API Updates / Open Questions
I changed the volume setting function from `WslcContainerSettingsAddVolume` to `WslcContainerSettingsSetVolumes` to reflect the fact that one can only set the entire volume list, not add volume(s) repeatedly.

The port mapping SDK provides the user with UDP/TCP control and a `socket_address` that have no apparent mapping to the runtime. ***Should they be removed?*** Suggested change: Remove unless runtime wants to support them.

The port mapping in the runtime provides a choice between v4 and v6; ***should the SDK just create two mapping at all times (one for each IP version)?  Or should we provide the user this control?***

## Validation Steps Performed
Added tests that cover these new APIs.


test/windows/WslcSdkTests.cpp
```diff
@@ -145,6 +149,44 @@ ContainerOutput RunContainerAndCapture(
...
+//
+// Reads from Handle until Content appears in the accumulated output.
+// Fails the test if the handle closes before Content is found.
+//
+void WaitForOutput(wil::unique_handle handle, std::string_view targetValue, std::chrono::milliseconds timeout = 60s)
+{
+    wsl::windows::common::relay::MultiHandleWait io;
+    io.AddHandle(std::make_unique<ReadHandleWithTargetValue>(std::move(handle), targetValue));
+    io.Run(timeout);
+}
```
JohnMcPMS: @OneBlue , if you like this setup I can move it somewhere shared so that the WSLATest TODO can use it as well.
https://github.com/microsoft/WSL/blob/7955489ad24b41d6500ef8af835d0a0845b8622d/test/windows/WSLATests.cpp#L1423-L1427

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -90,6 +90,19 @@ WSLASignal ConvertSignal(WslcSignal signal)
...
+WSLA_CONTAINER_NETWORK_TYPE Convert(WslcContainerNetworkingMode mode)
+{
+    switch (mode)
+    {
+    case WSLC_CONTAINER_NETWORKING_MODE_NONE:
+        return WSLA_CONTAINER_NETWORK_NONE;
+    case WSLC_CONTAINER_NETWORKING_MODE_BRIDGED:
+        return WSLA_CONTAINER_NETWORK_BRIDGE;
+    default:
+        THROW_HR(E_INVALIDARG);
```
OneBlue: nit: we might want to log the value to make debugging easier if this ever hits: 

```
THROW_HR_MSG(E_INVALIDARG, "Invalid networking mode: %i",mode);
```

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -360,6 +364,8 @@ try
     *internalType = {};
 
     internalType->image = imageName;
+    // Default network configuration to WSLC SDK `0`, which is NONE.
+    internalType->networking = WSLA_CONTAINER_NETWORK_NONE;
```
OneBlue, line 386: We might want to default to NAT since this is often the default when creating containers on the command line

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -400,12 +406,45 @@ try
...
+        {
+            const WslcContainerPortMapping& internalPort = internalContainerSettings->ports[i];
+            WSLA_PORT_MAPPING& convertedPort = convertedPorts[i];
+
+            convertedPort.HostPort = internalPort.windowsPort;
+            convertedPort.ContainerPort = internalPort.containerPort;
+            // TODO: Only other supported value right now is AF_INET6; no user access.
+            convertedPort.Family = AF_INET;
+
+            // TODO: Unused protocol?
```
OneBlue, line 458: That's right, those aren't implemented in the service right now

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -400,12 +406,45 @@ try
         // containerOptions.InitProcessOptions.User;
     }
 
-    // TODO: Implement
-    // containerOptions.Volumes;
-    // containerOptions.VolumesCount;
-    // containerOptions.Ports;
-    // containerOptions.PortsCount;
-    // containerOptions.ContainerNetwork;
+    std::unique_ptr<WSLA_VOLUME[]> convertedVolumes;
```
OneBlue, line 427: nit: We could use an std::vector<WSLA_VOLUME> to simplify things a bit (the internal pointer can be accessed via `.data()`

test/windows/WslcSdkTests.cpp
OneBlue: Unfortunately that's not guaranteed to work because Collect() might not be called at all if all the IO completes synchronously.

To achieve this we could instead do something like this (untested): 

```
{
    wsl::windows::common::relay::MultiHandleWait io;

    std::string content;
    auto onRead = [&](const gsl::span<char>& buffer) {
       content.append(buffer.data(), buffer.size());
       if (content.find(m_targetValue) != std::string::npos)
       {
             m_io.Cancel();
       }
    };
    io.AddHandle(std::make_unique<ReadHandle>(std::move(handle), std::move(onRead)));
    io.Run(timeout);

    // Fail if the IO completed before the expected content was found
    VERIFY_IS_TRUE(content.find(m_targetValue) != std::string::npos);
}

```

I do think that having a timeout is better than the current `WaitForOutput()` in WSLATests.cpp though, and if we do implement the above it would be worth factoring it in Common.cpp

test/windows/WslcSdkTests.cpp
```diff
@@ -157,6 +199,43 @@ class WslcSdkTests
...
+        }
+        else
+        {
+            THROW_HR_MSG(E_INVALIDARG, "Unknown test image: %hs", imageName.data());
+        }
+
+        return result;
+    }
+
+    void LoadTestImage(std::string_view imageName)
```
OneBlue, line 164: Nice ! This is probably also worth factoring

test/windows/WslcSdkTests.cpp
```diff
@@ -486,6 +570,396 @@ class WslcSdkTests
...
+            opts.ContentLength = 1;
+            VERIFY_ARE_EQUAL(WslcSessionImageLoad(m_defaultSession, &opts), E_INVALIDARG);
+        }
+
+        // Negative: zero ContentLength must fail.
+        {
+            WslcLoadImageOptions opts{};
+            opts.ImageHandle = GetCurrentProcess();
+            opts.ContentLength = 0;
+            VERIFY_ARE_EQUAL(WslcSessionImageLoad(m_defaultSession, &opts), E_INVALIDARG);
```
OneBlue, line 564: nit: There's a TODO for this in WSLATests, but we should also have test coverage for trying to load an invalid image file. We could for instance open a handle to our current executable and verify that this fails

test/windows/WslcSdkTests.cpp
```diff
@@ -486,6 +570,396 @@ class WslcSdkTests
...
+            wil::unique_socket clientSocket{socket(AF_INET, SOCK_STREAM, IPPROTO_TCP)};
+            VERIFY_IS_TRUE(!!clientSocket);
+
+            sockaddr_in addr{};
+            addr.sin_family = AF_INET;
+            addr.sin_port = htons(12341);
+            addr.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+            VERIFY_ARE_NOT_EQUAL(connect(clientSocket.get(), reinterpret_cast<sockaddr*>(&addr), sizeof(addr)), SOCKET_ERROR);
+
+            const char* request = "GET / HTTP/1.0\r\nHost: localhost\r\n\r\n";
```
OneBlue: Nit; We could factor and reuse `ExpectHttpResponse()` for this

test/windows/WslcSdkTests.cpp
```diff
@@ -486,6 +570,396 @@ class WslcSdkTests
...
+
+            ContainerOutput output;
+            wsl::windows::common::relay::MultiHandleWait io;
+            io.AddHandle(std::make_unique<wsl::windows::common::relay::ReadHandle>(
+                std::move(ownedStdout), [&](const auto& buf) { output.stdoutOutput.append(buf.data(), buf.size()); }));
+            io.AddHandle(std::make_unique<wsl::windows::common::relay::ReadHandle>(
+                std::move(ownedStderr), [&](const auto& buf) { output.stderrOutput.append(buf.data(), buf.size()); }));
+            io.Run(60s);
+
+            VERIFY_ARE_EQUAL(WaitForSingleObject(exitEvent, 10 * 1000), static_cast<DWORD>(WAIT_OBJECT_0));
```
OneBlue: I recommend a longer timeout (probably 60 seconds) since the CI can be slow sometimes

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: My only argument is that we should default to the 0 value of the SDK enum.  NAT isn't a runtime value though, not sure if that is another name for HOST or BRIDGE, or a reference to the session networking mode.

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: If they are both planned I can at least `E_NOTIMPL` what I can.

The port mapping in the runtime provides a choice between v4 and v6; ***should the SDK just create two mapping at all times (one for each IP version)?  Or should we provide the user this control?***

src/windows/WslcSDK/wslcsdk.cpp
OneBlue: Sorry yeah by NAT I actually meant "bridged". We could make that the '0' value yeah

src/windows/WslcSDK/wslcsdk.cpp
OneBlue: I think we should let the user control that

test/windows/WslcSdkTests.cpp
JohnMcPMS: That method doesn't end in the rare event that there is always data to read synchronously.  I've updated my derived type to better cover all reads and the completion before retrieving the target value.  I've also moved it to common.

test/windows/WslcSdkTests.cpp
JohnMcPMS: I added a test for it and the load is successful... need to debug into the docker interaction.

test/windows/WslcSdkTests.cpp
JohnMcPMS: The actual load uses the runtime/SDK API to do the load between the two.

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -473,24 +512,54 @@ try
...
+    auto internalType = CheckAndGetInternalType(containerSettings);
+    RETURN_HR_IF(E_INVALIDARG, (volumes == nullptr && volumeCount != 0) || (volumes != nullptr && volumeCount == 0));
+
+    for (uint32_t i = 0; i < volumeCount; ++i)
+    {
+        RETURN_HR_IF_NULL(E_INVALIDARG, volumes[i].windowsPath);
+        RETURN_HR_IF_NULL(E_INVALIDARG, volumes[i].containerPath);
+    }
+
+    internalType->volumes = volumes;
```
OneBlue, line 579: Not blocking in this PR, but the service will refuse paths that are not absolute, but that will only happen when we call CreateSession().

To make this a path nicer for the user, we could fail if any of the volumes have relative paths here

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -748,9 +823,23 @@ CATCH_RETURN();
 STDAPI WslcSessionImageLoad(_In_ WslcSession session, _In_ const WslcLoadImageOptions* options)
```
OneBlue, line 843: nit For a future PR: We might want to also expose something like: 

```
WslcSessionImageLoadFromFile(WslcSession session, LPCWSTR Path);
```

So that the caller doesn't have to manually open the handle and get the file size

test/windows/WslcSdkTests.cpp
OneBlue: non-blocking: I don't love the subclassing of `ReadHandle` for this, because we only run the check in the IO callbacks, and while this works now because there's only one handle (and so we can't fail / time out before scheduling IO on it at least once), it's feeling a bit brittle.

It's a good catch that we should exit if we keep getting synchronous reads though. I'll fix this in a separate PR. 

For this one I'll suggest using the ReadHandle callback to exit early. My original suggestion can actually be simplified to: 


```
{
    wsl::windows::common::relay::MultiHandleWait io;

    std::string content;
    auto onRead = [&](const gsl::span<char>& buffer) {
       content.append(buffer.data(), buffer.size());
       if (content.size() >= targetValue.size())
       {
             m_io.Cancel();
       }
    };
    io.AddHandle(std::make_unique<ReadHandle>(std::move(handle), std::move(onRead)));
    io.Run(timeout);

    VERIFY_IS_TRUE(content.find(m_targetValue) != std::string::npos);
}
```

OneBlue (approved):

test/windows/WslcSdkTests.cpp
OneBlue: Here's the PR for the cancellation fix: https://github.com/microsoft/WSL/pull/14296

PR #14277, opened 2026-02-25
CLI - Add logs command, refactor session/shell service

dkbennett:
* Adds the container 'logs' command, which is also aliased to the root:  `wslc container logs` or `wslc logs` both work and execute the same code.
* Refactors the Session service to absorb the Shell service and have all session-related code in the SessionService.
* Refactors the SessionTasks to contain all session-related functions consumed by commands, namely the CreateSession task and the AttachToSession task were moved to this task file.

## PR Checklist

* Added logs command to container and aliased to root.
* Added logs execution task to do the attach.
* Added logs to the container service. This code was mostly moved directly from wsladiag.exe with no intentional or obvious functional deviation.
* Refactored SessionService to contain all session related interactions (it essentially just absorbed the ShellService). The attach and list functions that were moved to the SessionService are not new, they are moved from the now-deleted ShellService.
* Refactored SessionTasks to have session list, session attach, and session create functionality. Create was moved from the container tasks since it is really a session task.

## Validation Steps Performed
Added tests for logs, test pass for CLI.
Manual verification of success, though I did not see any log output. The actual log functionality may have bugs or possibly user error. If there are bugs to fix with this I'll fix address in subsequent PR if it isn't working correctly.


src/windows/wslc/commands/ContainerCommand.cpp
```diff
@@ -21,12 +21,13 @@ namespace wsl::windows::wslc {
 std::vector<std::unique_ptr<Command>> ContainerCommand::GetCommands() const
 {
     std::vector<std::unique_ptr<Command>> commands;
```
benhillis, line 23: minor: could just define a std::vector or std::list inline and not have to .reserve

src/windows/wslc/commands/ContainerLogsCommand.cpp
```diff
@@ -0,0 +1,53 @@
...
+{
+    return {L"View container logs"};
+}
+
+std::wstring ContainerLogsCommand::LongDescription() const
+{
+    return {L"View logs for a container."};
+}
+
+// clang-format off
```
benhillis: don't override the clang formatter please.

src/windows/wslc/services/SessionService.cpp
```diff
@@ -14,12 +14,99 @@ Module Name:
...
+    // Launch with terminal fds (PTY).
+    wsl::windows::common::WSLAProcessLauncher launcher{shell, {shell, "--login"}, {"TERM=xterm-256color"}, WSLAProcessFlagsTty | WSLAProcessFlagsStdin};
+    launcher.SetTtySize(rows, cols);
+
+    auto process = launcher.Launch(*session);
+
+    auto tty = process.GetStdHandle(WSLAFDTty);
+
+    // Configure console for interactive usage.
+    wsl::windows::common::ConsoleState console;
```
benhillis: ConsoleState has a method to get the window size, you declare this above and you don't need to do the other call above.

src/windows/wslc/services/SessionService.cpp
```diff
@@ -14,12 +14,99 @@ Module Name:
...
+    // Start input relay thread to forward console input to TTY
+    // Runs in parallel with output relay (main thread)
+    auto exitEvent = wil::unique_event(wil::EventOptions::ManualReset);
+    std::thread inputThread([&] {
+        try
+        {
+            wsl::windows::common::relay::StandardInputRelay(
+                GetStdHandle(STD_INPUT_HANDLE), tty.get(), updateTerminalSize, exitEvent.get());
+        }
+        catch (...)
```
benhillis, line 76: This doesn't seem necessary. If an exception is thrown, the thread will already be exiting.

src/windows/wslc/services/SessionService.cpp
```diff
@@ -14,12 +14,99 @@ Module Name:
...
+    });
+
+    // Relay tty output -> console (blocks until output ends).
+    wsl::windows::common::relay::InterruptableRelay(tty.get(), GetStdHandle(STD_OUTPUT_HANDLE), exitEvent.get());
+
+    process.GetExitEvent().wait();
+
+    auto exitCode = process.GetExitCode();
+
+    std::wstring shellWide(shell.begin(), shell.end());
```
benhillis: Don't do narrow to wide strings like this, use wsl::shared::string::MultiByteToWide

src/windows/wslc/services/SessionService.cpp
```diff
@@ -33,4 +120,26 @@ Session SessionService::CreateSession(const std::optional<SessionOptions>& optio
...
+    wil::com_ptr<IWSLASessionManager> sessionManager;
+    THROW_IF_FAILED(CoCreateInstance(__uuidof(WSLASessionManager), nullptr, CLSCTX_LOCAL_SERVER, IID_PPV_ARGS(&sessionManager)));
+    wsl::windows::common::security::ConfigureForCOMImpersonation(sessionManager.get());
+
+    wil::unique_cotaskmem_array_ptr<WSLA_SESSION_INFORMATION> sessions;
+    THROW_IF_FAILED(sessionManager->ListSessions(&sessions, sessions.size_address<ULONG>()));
+    for (size_t i = 0; i < sessions.size(); ++i)
+    {
+        const auto& current = sessions[i];
+        SessionInformation info;
```
benhillis: SessionInformation info{};

benhillis (changes requested):
mostly good, some minor comments.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -250,4 +244,13 @@ void StopContainers(CLIExecutionContext& context)
...
     }
 }
+
+void ViewContainerLogs(CLIExecutionContext& context)
+{
+    WI_ASSERT(context.Data.Contains(Data::Session));
+    auto& session = context.Data.Get<Data::Session>();
+    auto containerId = context.Args.Get<ArgType::ContainerId>();
+    bool follow = context.Args.Contains(ArgType::Follow);
+    ContainerService::Logs(session, string::WideToMultiByte(containerId), follow);
```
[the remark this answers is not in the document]
dkbennett: False positive here Copilot; it obviously builds because it passed the checks which verify the build. (but I'll change it anyway for consistency with the rest of the file which does not do the prefix)

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -224,4 +227,28 @@ InspectContainer ContainerService::Inspect(Session& session, const std::string&
...
+    wil::com_ptr<IWSLAContainer> container;
+    THROW_IF_FAILED(session.Get()->OpenContainer(id.c_str(), &container));
+    wil::unique_handle stdoutLogs;
+    wil::unique_handle stderrLogs;
+    WSLALogsFlags flags = WSLALogsFlagsNone;
+    WI_SetFlagIf(flags, WSLALogsFlagsFollow, follow);
+
+    THROW_IF_FAILED(container->Logs(flags, reinterpret_cast<ULONG*>(&stdoutLogs), reinterpret_cast<ULONG*>(&stderrLogs), 0, 0, 0));
+
+    wsl::windows::common::relay::MultiHandleWait io;
```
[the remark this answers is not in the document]
dkbennett: Fixed

src/windows/wslc/commands/ContainerLogsCommand.cpp
dkbennett: I did it for readability of the function chains. I can achieve that with some comment breaks, which is a common practice to do this sort of thing, so removed it and fixed the formatting with the comments.

src/windows/wslc/services/SessionService.cpp
dkbennett: I agree it probably isn't necessary but this was a code move and not a new addition so I'd prefer to leave it alone so as to not risk regression.

As I looked into it, I saw the implementation in wsladiag is significantly different (and cleaner) than this code, and updating it would be a significant change to Attach and would also address the issues you point out. I would like to fix attach all-up to better align with wsladiag implementation, but would prefer to do that in its own PR. So I am leaving this alone but will work on some improvements to Attach specifically.

src/windows/wslc/commands/ContainerCommand.cpp
dkbennett: Defining the vector inline doesn't compile because unique_ptr cannot be copied the initializer list isn't supported for move-only types. The current implementation is efficient and only moves, and uses the copy elision, but the reserve needing updating is a maintenance issue.

Copilot advised against using std::list in this instance due to memory overhead, slower iteration, and no performance benefits. 

I resolved this by simply removing the reserve call to improve maintainability. The compiler should handle this correctly (or so copilot tells me) and not need the micro-optimization of the reserve size. So going to trust the compiler is good at its job and will do the right thing in this situation.

dkbennett:
I will make similar optimizations in other code (removing reserve call, use comments to force formatting breaks) next time they are touched or do an PR specifically for that.  Also plan to do a separate PR for fixing attach to be more like wsladiag. It looks like OneBlue did a change of it late last month and this code was based on a prior version.

src/windows/wslc/services/SessionService.cpp
benhillis: that's fine, let's file something to track that so we don't forget about it.

src/windows/wslc/commands/ContainerLogsCommand.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+std::wstring ContainerLogsCommand::LongDescription() const
+{
+    return {L"View logs for a container."};
+}
+
+void ContainerLogsCommand::ExecuteInternal(CLIExecutionContext& context) const
+{
+    context              //
```
OneBlue, line 47: nit: looks like empty comments

src/windows/wslc/services/SessionService.cpp
```diff
@@ -14,12 +14,91 @@ Module Name:
...
-
+namespace wslutil = wsl::windows::common::wslutil;
 DEFINE_ENUM_FLAG_OPERATORS(WSLASessionFlags);
 
+int SessionService::Attach(const std::wstring& sessionName)
+{
+    THROW_HR_IF(E_INVALIDARG, sessionName.empty());
+
+    wil::com_ptr<IWSLASessionManager> manager;
+    THROW_IF_FAILED(CoCreateInstance(__uuidof(WSLASessionManager), nullptr, CLSCTX_LOCAL_SERVER, IID_PPV_ARGS(&manager)));
```
OneBlue, line 30: nit: We might want to factor the opening + configuration of a WSLASession (including error messages), but we can do that in a future change

OneBlue (approved):

src/windows/wslc/commands/ContainerLogsCommand.cpp
dkbennett: Its for formatting since Ben did not want me to disable the clang formatter for this function, and it is apparently a common practice in popular repos, or so Copilot tells me.  Agree it doesn't look as pretty as I'd like but it more readable than them all being on one line.

OneBlue: Ok thank you. I don't have a strong opinion on that

benhillis: I don't love this either, but it's fine.

benhillis (approved):

PR #14279, opened 2026-02-25
WSLA: Fix container destructor iterating cleared member instead of local copy

benhillis:
In ~WSLAContainerImpl(), m_processes is copied to a local variable 'processes' and then m_processes is cleared under the lock. However, the for loop that calls OnContainerReleased() iterated m_processes (now empty) instead of the local copy 'processes'.

This meant OnContainerReleased() was never called on exec'd processes when a container was destroyed, causing clients waiting on those processes to hang indefinitely.


OneBlue (approved):
Nice catch

PR #14281, opened 2026-02-25
wsla: do not return sessions via ListSessions that the caller does not have access to

benhillis:
ListSessions returned all sessions to any caller without checking access, allowing non-elevated users to enumerate other users' sessions. Fixed by calling CheckTokenAccess to filter results, matching the pattern already used by OpenSession and OpenSessionByName.

### DIFF
--- a/src/windows/wslaservice/exe/WSLASessionManager.cpp
+++ b/src/windows/wslaservice/exe/WSLASessionManager.cpp
@@ -149,9 +149,15 @@ void WSLASessionManagerImpl::OpenSessionByName(LPCWSTR DisplayName, IWSLASession
 
 void WSLASessionManagerImpl::ListSessions(_Out_ WSLA_SESSION_INFORMATION** Sessions, _Out_ ULONG* SessionsCount)
 {
+    auto tokenInfo = GetCallingProcessTokenInfo();
     std::vector<WSLA_SESSION_INFORMATION> sessionInfo;
 
     ForEachSession<void>([&](auto& entry, const auto&) {
+        if (FAILED(CheckTokenAccess(entry, tokenInfo)))
+        {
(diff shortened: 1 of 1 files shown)


src/windows/wslaservice/exe/WSLASessionManager.cpp
```diff
@@ -149,9 +149,15 @@ void WSLASessionManagerImpl::OpenSessionByName(LPCWSTR DisplayName, IWSLASession
 
 void WSLASessionManagerImpl::ListSessions(_Out_ WSLA_SESSION_INFORMATION** Sessions, _Out_ ULONG* SessionsCount)
 {
+    auto tokenInfo = GetCallingProcessTokenInfo();
     std::vector<WSLA_SESSION_INFORMATION> sessionInfo;
 
     ForEachSession<void>([&](auto& entry, const auto&) {
+        if (FAILED(CheckTokenAccess(entry, tokenInfo)))
```
OneBlue, line 156: Hmm this is tricky because that would mean that running wslc diag list will show different values elevated and non-elevated, which could lead to confusion. 

Is there any downside in having higher privileged sessions visible to users ? I'm thinking about this the same way we do process visibilty: Anyone can see all processes, but they can only open the process that they actually have access to

benhillis: yeah I considered this - if they can't interact with the session and try to they'd get an error.

I guess either way is fine, it's just odd to return something that isn't usable.

OneBlue: Given that the session listing is a essentially a debug tool, I'd be of the opinion to keep them in the list, just to avoid potential confusion

PR #14283, opened 2026-02-25
Enable VS 2026 solution generation and improve ARM64 developer experience

benhillis:
- Auto-detect host architecture when TARGET_PLATFORM and CMAKE_GENERATOR_PLATFORM are unset, defaulting to ARM64 on ARM64 hosts
- Fix trailing '|' in platform regex that made the else branch unreachable
- Refactor Visual Studio detection to prefer VS2022 and fall back to VS2026 with a clang-format warning
- Document x64 .NET 6.0 runtime setup required for WiX on ARM64

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,10 +4,20 @@ set(CMAKE_SYSTEM_VERSION 10.0.26100.0)
 project(wsl)
 
 # Rationalize TARGET_PLATFORM
+# When neither CMAKE_GENERATOR_PLATFORM nor TARGET_PLATFORM is set, default to the host architecture.
+if("${CMAKE_GENERATOR_PLATFORM}" STREQUAL "" AND "${TARGET_PLATFORM}" STREQUAL "")
+    if("${CMAKE_HOST_SYSTEM_PROCESSOR}" STREQUAL "ARM64")
+        set(TARGET_PLATFORM "arm64")
+    else()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -258,18 +268,31 @@ else()
...
+        COMMAND "${VSWHERE_SOURCE_DIR}/vswhere.exe" -version "${VERSION_RANGE}" -products * -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath -prerelease -latest
+        OUTPUT_VARIABLE _vs_install_dir
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+        COMMAND_ERROR_IS_FATAL ANY
+    )
+
+    set(${OUTPUT_VAR} "${_vs_install_dir}" PARENT_SCOPE)
+endfunction()
+
+find_vs_install_dir("[17.0,18.0)" VS_INSTALL_DIR)
```
OneBlue, line 285: Nice, that's probably the best we can do before the pipeline supports VS 2026

OneBlue (approved):

CMakeLists.txt
benhillis: Yeah, I happened to try to set up the dev environment on my arm64 machine and couldn't find vs2022 so I think this is probably ok for now.

PR #14284, opened 2026-02-26
Retrieve newly created container names from docker instead of using an empty string

OneBlue:

This change updates the container creation logic to retrieve generated names from docker 

## PR Checklist


benhillis (approved):

dkbennett:
Can confirm this also fixes the CLI's empty string create as well.

PR #14288, opened 2026-02-26
Pass nullptr to WSLAContainerLauncher if name is empty string

dkbennett:
The container CLI was passing emptystring into container names if no name was provided, which is a valid name to the ContainerLauncher, resulting in an empty name.

This fix passes nullptr instead if the name is empty string, which signals the launcher to use the generated name.

This is the CLI-side fix for the same problem in #14284; it was broken on both sides.

## PR Checklist

## Validation Steps Performed
Ran the command with no name to confirm the nullptr input is valid and didn't cause problems. The name was still emptystring due to #14284.
Ran the command again on the 14284 integration build and it failed to create anything without a name, but specifying name worked as expected.

Leaving this as a draft until this is debugged.


dkbennett:
The other PR fixed this, and this is unnecessary.

PR #14290, opened 2026-02-26
Update error message text and return some with more text

dabhattimsft:
This PR updates error text returned for certain errors and returns few with error text.

### DIFF
--- a/localization/strings/en-US/Resources.resw
+++ b/localization/strings/en-US/Resources.resw
@@ -248,9 +248,12 @@ Please use '{} --help' to get a list of supported arguments.</value>
     <value>Cannot launch because another instance is running elevated.  Elevated and un-elevated instances are not permitted to run simultaneously.</value>
   </data>
   <data name="MessageNoDefaultDistro" xml:space="preserve">
-    <value>No Linux distribution is installed.
-Run 'wsl.exe --install &lt;Distro&gt;' to install one (for example, 'wsl.exe --install Ubuntu').</value>
-    <comment>{Locked="--install "}Command line arguments, file names and string inserts should not be translated</comment>
+    <value>Windows Subsystem for Linux has no installed distributions.
+You can resolve this by installing a distribution with the instructions below:
+
+Use 'wsl.exe --list --online' to list available distributions
+and 'wsl.exe --install &lt;Distro&gt;' to install.</value>
+    <comment>{Locked="--list "}{Locked="--online'"}{Locked="--install "}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
   <data name="MessageNoRunningDistro" xml:space="preserve">
     <value>There are no running distributions.</value>
--- a/localization/strings/en-US/Resources.resw
+++ b/localization/strings/en-US/Resources.resw
@@ -248,12 +248,9 @@ Please use '{} --help' to get a list of supported arguments.</value>
     <value>Cannot launch because another instance is running elevated.  Elevated and un-elevated instances are not permitted to run simultaneously.</value>
   </data>
   <data name="MessageNoDefaultDistro" xml:space="preserve">
-    <value>Windows Subsystem for Linux has no installed distributions.
-You can resolve this by installing a distribution with the instructions below:
-
-Use 'wsl.exe --list --online' to list available distributions
-and 'wsl.exe --install &lt;Distro&gt;' to install.</value>
-    <comment>{Locked="--list "}{Locked="--online'"}{Locked="--install "}Command line arguments, file names and string inserts should not be translated</comment>
+    <value>No Linux distribution is installed.
+Run 'wsl.exe --install &lt;Distro&gt;' to install one (for example, 'wsl.exe --install Ubuntu'), or install a distribution from the Microsoft Store.</value>
+    <comment>{Locked="--install "}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
   <data name="MessageNoRunningDistro" xml:space="preserve">
     <value>There are no running distributions.</value>
@@ -305,8 +302,9 @@ For information please visit https://aka.ms/enablevirtualization</value>
     <value>Too many virtual hard disks or physical disks are attached.</value>
   </data>
   <data name="MessageUserVhdAlreadyAttached" xml:space="preserve">
-    <value>VHD already mounted via wsl.exe --mount, please unmount the disk before launching.</value>
-    <comment>{Locked="--mount,"}Command line arguments, file names and string inserts should not be translated</comment>
+    <value>The virtual disk is already attached.
+Unmount it with 'wsl.exe --unmount &lt;DiskPath&gt;' before attaching again.</value>
+    <comment>{Locked="--unmount "}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
   <data name="MessageWsl1NotSupported" xml:space="preserve">
     <value>WSL1 is not supported with your current machine configuration.
@@ -1013,7 +1011,30 @@ Falling back to NAT networking.</value>
     <value>Sparse VHD is supported on WSL2 only.</value>
   </data>
   <data name="MessageLocalSystemNotSupported" xml:space="preserve">
-    <value>Running WSL as local system is not supported.</value>
+    <value>WSL cannot run under the LocalSystem account.
+Use a standard user account instead, or launch PowerShell as administrator.</value>
+  </data>
+  <data name="MessageClassNotRegistered" xml:space="preserve">
+    <value>WSL component registration failed ('Class Not Registered').
+Repair or reinstall WSL from Settings &gt; Programs &gt; System Components &gt; WSL &gt; Repair.</value>
+  </data>
+  <data name="MessageUnexpectedError" xml:space="preserve">
+    <value>An unexpected error occurred while starting WSL.
+Restart WSL with 'wsl.exe --shutdown' or update it with 'wsl.exe --update', then try again.</value>
+    <comment>{Locked="--shutdown'"}{Locked="--update'"}Command line arguments, file names and string inserts should not be translated</comment>
+  </data>
+  <data name="MessageServiceConnectionFailed" xml:space="preserve">
+    <value>WSL couldn't connect to its service (network or service error 0x8007274c).
+Reset your network and restart the Hyper-V/WSL service, then try again.</value>
+  </data>
+  <data name="MessageDetachDiskNotFound" xml:space="preserve">
+    <value>WSL could not find the specified disk to detach.
+Check that the file path is correct and that the virtual disk exists.</value>
+  </data>
+  <data name="MessageServiceDisabled" xml:space="preserve">
+    <value>WSL service 'WSLService' is disabled.
+Start it with 'sc.exe config WSLService start= auto' and 'net.exe start WSLService', then try again.</value>
+    <comment>{Locked="sc.exe config WSLService start= auto"}{Locked="net.exe start WSLService"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
   <data name="MessagePerformanceTip" xml:space="preserve">
     <value>Performance Tip:</value>
--- a/localization/strings/en-US/Resources.resw
+++ b/localization/strings/en-US/Resources.resw
@@ -249,7 +249,7 @@ Please use '{} --help' to get a list of supported arguments.</value>
   </data>
   <data name="MessageNoDefaultDistro" xml:space="preserve">
     <value>No Linux distribution is installed.
-Run 'wsl.exe --install &lt;Distro&gt;' to install one (for example, 'wsl.exe --install Ubuntu'), or install a distribution from the Microsoft Store.</value>
+Run 'wsl.exe --install &lt;Distro&gt;' to install one (for example, 'wsl.exe --install Ubuntu').</value>
     <comment>{Locked="--install "}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
   <data name="MessageNoRunningDistro" xml:space="preserve">
@@ -1014,27 +1014,14 @@ Falling back to NAT networking.</value>
     <value>WSL cannot run under the LocalSystem account.
 Use a standard user account instead, or launch PowerShell as administrator.</value>
   </data>
-  <data name="MessageClassNotRegistered" xml:space="preserve">
-    <value>WSL component registration failed ('Class Not Registered').
-Repair or reinstall WSL from Settings &gt; Programs &gt; System Components &gt; WSL &gt; Repair.</value>
-  </data>
   <data name="MessageUnexpectedError" xml:space="preserve">
-    <value>An unexpected error occurred while starting WSL.
+    <value>An unexpected error occurred.
 Restart WSL with 'wsl.exe --shutdown' or update it with 'wsl.exe --update', then try again.</value>
     <comment>{Locked="--shutdown'"}{Locked="--update'"}Command line arguments, file names and string inserts should not be translated</comment>
-  </data>
-  <data name="MessageServiceConnectionFailed" xml:space="preserve">
-    <value>WSL couldn't connect to its service (network or service error 0x8007274c).
-Reset your network and restart the Hyper-V/WSL service, then try again.</value>
   </data>
   <data name="MessageDetachDiskNotFound" xml:space="preserve">
     <value>WSL could not find the specified disk to detach.
 Check that the file path is correct and that the virtual disk exists.</value>
-  </data>
-  <data name="MessageServiceDisabled" xml:space="preserve">
-    <value>WSL service 'WSLService' is disabled.
-Start it with 'sc.exe config WSLService start= auto' and 'net.exe start WSLService', then try again.</value>
-    <comment>{Locked="sc.exe config WSLService start= auto"}{Locked="net.exe start WSLService"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
   <data name="MessagePerformanceTip" xml:space="preserve">
     <value>Performance Tip:</value>
--- a/localization/strings/en-US/Resources.resw
+++ b/localization/strings/en-US/Resources.resw
@@ -304,7 +304,7 @@ For information please visit https://aka.ms/enablevirtualization</value>
   <data name="MessageTooManyDisks" xml:space="preserve">
     <value>Too many virtual hard disks or physical disks are attached.</value>
   </data>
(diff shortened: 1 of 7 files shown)


localization/strings/en-US/Resources.resw
```diff
@@ -1013,7 +1011,30 @@ Falling back to NAT networking.</value>
     <value>Sparse VHD is supported on WSL2 only.</value>
   </data>
   <data name="MessageLocalSystemNotSupported" xml:space="preserve">
-    <value>Running WSL as local system is not supported.</value>
+    <value>WSL cannot run under the LocalSystem account.
+Use a standard user account instead, or launch PowerShell as administrator.</value>
+  </data>
+  <data name="MessageClassNotRegistered" xml:space="preserve">
+    <value>WSL component registration failed ('Class Not Registered').
+Repair or reinstall WSL from Settings &gt; Programs &gt; System Components &gt; WSL &gt; Repair.</value>
```
benhillis: Does this actually do anything? I'm not sure we have repair implemented in our MSI.

src/windows/common/wslutil.cpp
```diff
@@ -745,6 +745,18 @@ std::wstring wsl::windows::common::wslutil::GetErrorString(HRESULT result)
     case WSL_E_HIGHER_INTEGRITY:
         return Localization::MessageHigherIntegrity();
 
+    case REGDB_E_CLASSNOTREG:
+        return Localization::MessageClassNotRegistered();
+
+    case E_UNEXPECTED:
+        return Localization::MessageUnexpectedError();
+
```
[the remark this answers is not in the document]
benhillis: Agreed, might be mapping too much to this specific error code.

src/windows/common/wslutil.cpp
```diff
@@ -766,6 +745,18 @@ std::wstring wsl::windows::common::wslutil::GetErrorString(HRESULT result)
     case WSL_E_HIGHER_INTEGRITY:
         return Localization::MessageHigherIntegrity();
 
+    case REGDB_E_CLASSNOTREG:
+        return Localization::MessageClassNotRegistered();
```
OneBlue: This is not necessarily true. `REGDB_E_CLASSNOTREG` can be thrown by other components than WSL. 

If we do want to specialize an error message for WSL's COM class being missing, we should have a try / catch for this when we instantiate WSL COM classes

src/windows/common/wslutil.cpp
```diff
@@ -766,6 +745,18 @@ std::wstring wsl::windows::common::wslutil::GetErrorString(HRESULT result)
...
+
+    case E_UNEXPECTED:
+        return Localization::MessageUnexpectedError();
```
OneBlue: The error message for this starts with:

```
An unexpected error occurred while starting WSL.
```

If we want to have a specialized error message for startup we should only display it when we actually fail to start WSL

src/windows/common/wslutil.cpp
```diff
@@ -766,6 +745,18 @@ std::wstring wsl::windows::common::wslutil::GetErrorString(HRESULT result)
...
+
+    case HRESULT_FROM_WIN32(WSAETIMEDOUT):
+        return Localization::MessageServiceConnectionFailed();
+
+    case HRESULT_FROM_WIN32(ERROR_SERVICE_DISABLED):
+        return Localization::MessageServiceDisabled();
```
OneBlue: This is not necessarily true. In the past we've been hitting this error when dependent services are disabled (like winnat for instance)

src/windows/common/wslutil.cpp
```diff
@@ -766,6 +745,18 @@ std::wstring wsl::windows::common::wslutil::GetErrorString(HRESULT result)
...
         return Localization::MessageHigherIntegrity();
 
+    case REGDB_E_CLASSNOTREG:
+        return Localization::MessageClassNotRegistered();
+
+    case E_UNEXPECTED:
+        return Localization::MessageUnexpectedError();
+
+    case HRESULT_FROM_WIN32(WSAETIMEDOUT):
+        return Localization::MessageServiceConnectionFailed();
```
OneBlue: That's not necessarily true. This could be a network error caused by a download failure for instance

localization/strings/en-US/Resources.resw
```diff
@@ -248,12 +248,9 @@ Please use '{} --help' to get a list of supported arguments.</value>
...
   </data>
   <data name="MessageNoDefaultDistro" xml:space="preserve">
-    <value>Windows Subsystem for Linux has no installed distributions.
-You can resolve this by installing a distribution with the instructions below:
-
-Use 'wsl.exe --list --online' to list available distributions
-and 'wsl.exe --install &lt;Distro&gt;' to install.</value>
-    <comment>{Locked="--list "}{Locked="--online'"}{Locked="--install "}Command line arguments, file names and string inserts should not be translated</comment>
+    <value>No Linux distribution is installed.
+Run 'wsl.exe --install &lt;Distro&gt;' to install one (for example, 'wsl.exe --install Ubuntu'), or install a distribution from the Microsoft Store.</value>
```
OneBlue: We should not direct users towards the Store for distributions anymore, wsl --install is the preferred way to install distributions

OneBlue (changes requested):

localization/strings/en-US/Resources.resw
OneBlue: The MSI package should actually do some repair, but that error message is confusing because: 

- WSL isn't in the list of programs (users might actually do 'repair' on the MSIX package, which won't do anything)

- This message is currently displayed for any ClassNotRegistered error, which can be misleading since we can receive it from other components

src/windows/common/wslutil.cpp
dabhattimsft: @benhillis, @craigloewen-msft thoughts on this?

craigloewen-msft: I agree, @OneBlue 's comments are valid.

benhillis: yep, this is wrong.

benhillis (changes requested):

localization/strings/en-US/Resources.resw
dabhattimsft: Removed the MessageClassNotRegistered string and the REGDB_E_CLASSNOTREG mapping entirely. We can revisit adding a targeted catch at the COM instantiation site in a follow-up if needed.

src/windows/common/wslutil.cpp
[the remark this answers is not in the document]
dabhattimsft: Removed the global E_UNEXPECTED mapping from GetErrorString(). The specialized message is now only attached via THROW_HR_WITH_USER_ERROR in the LaunchProcess() → CreateInstanceNoThrow() path in WslClient.cpp, so it only shows during actual startup failures.

src/windows/common/wslutil.cpp
dabhattimsft: Removed the global mapping and the resource string.

src/windows/common/ExecutionContext.h
```diff
@@ -11,6 +11,15 @@ namespace wsl::windows::common {
...
     THROW_HR(Result)
 
+#define THROW_HR_WITH_USER_ERROR_IF(Result, Message, Condition) \
+    do \
+    { \
+        if (Condition) \
+        { \
+            THROW_HR_WITH_USER_ERROR(Result, Message); \
+        } \
+    } while (0, 0)
```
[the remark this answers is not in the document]
dabhattimsft: Fixed

localization/strings/en-US/Resources.resw
dabhattimsft: Fixed

localization/strings/en-US/Resources.resw
```diff
@@ -1013,7 +1011,17 @@ Falling back to NAT networking.</value>
     <value>Sparse VHD is supported on WSL2 only.</value>
   </data>
   <data name="MessageLocalSystemNotSupported" xml:space="preserve">
-    <value>Running WSL as local system is not supported.</value>
+    <value>WSL cannot run under the LocalSystem account.
+Use a standard user account instead, or launch PowerShell as administrator.</value>
```
OneBlue: I don't recommend mentioning powershell here since powershell doesn't have anything to do with this issue. 

I also wouldn't mention "Administrator" since that could be confusing given that the local system account has the highest amount of privileges possible

src/windows/common/wslutil.cpp
dabhattimsft: Removed WSAETIMEDOUT mapping.

src/windows/common/wslutil.cpp
dabhattimsft: Removed the MessageServiceDisabled mapping

src/windows/common/wslutil.cpp
dabhattimsft: made the message text generic ("An unexpected error occurred.") and scoped it to the CreateInstanceNoThrow() startup path only.

localization/strings/en-US/Resources.resw
```diff
@@ -1013,7 +1011,17 @@ Falling back to NAT networking.</value>
     <value>Sparse VHD is supported on WSL2 only.</value>
   </data>
   <data name="MessageLocalSystemNotSupported" xml:space="preserve">
-    <value>Running WSL as local system is not supported.</value>
+    <value>WSL cannot run under the LocalSystem account.
+Use a standard user account instead, or launch PowerShell as administrator.</value>
+  </data>
+  <data name="MessageUnexpectedError" xml:space="preserve">
+    <value>An unexpected error occurred.
+Restart WSL with 'wsl.exe --shutdown' or update it with 'wsl.exe --update', then try again.</value>
```
OneBlue: We should be careful about recommending `wsl --shutdown` and `wsl --update`, since those will kill everything. 

I'm not sure that this error message is worth adding, it's a pretty generic error and recommending wsl --shutdown might cause more harm than good

localization/strings/en-US/Resources.resw
```diff
@@ -1013,7 +1011,17 @@ Falling back to NAT networking.</value>
...
+Use a standard user account instead, or launch PowerShell as administrator.</value>
+  </data>
+  <data name="MessageUnexpectedError" xml:space="preserve">
+    <value>An unexpected error occurred.
+Restart WSL with 'wsl.exe --shutdown' or update it with 'wsl.exe --update', then try again.</value>
+    <comment>{Locked="--shutdown'"}{Locked="--update'"}Command line arguments, file names and string inserts should not be translated</comment>
+  </data>
+  <data name="MessageDetachDiskNotFound" xml:space="preserve">
+    <value>WSL could not find the specified disk to detach.
+Check that the file path is correct and that the virtual disk exists.</value>
```
OneBlue: This error isn't necessarily about a virtual disk, it can also be about a physical disk. 

I would recommend rewriting as `Disk not found: <disk>` to make it easy for users to diagnose

localization/strings/en-US/Resources.resw
```diff
@@ -305,8 +302,9 @@ For information please visit https://aka.ms/enablevirtualization</value>
     <value>Too many virtual hard disks or physical disks are attached.</value>
   </data>
   <data name="MessageUserVhdAlreadyAttached" xml:space="preserve">
-    <value>VHD already mounted via wsl.exe --mount, please unmount the disk before launching.</value>
-    <comment>{Locked="--mount,"}Command line arguments, file names and string inserts should not be translated</comment>
+    <value>The virtual disk is already attached.
+Unmount it with 'wsl.exe --unmount &lt;DiskPath&gt;' before attaching again.</value>
```
OneBlue, line 331: I recommend using an insert to give the actual disk path here (you can use "{}" to generate string inserts, there are examples of that in this file)

OneBlue (changes requested):

localization/strings/en-US/Resources.resw
dabhattimsft: Fixed

localization/strings/en-US/Resources.resw
dabhattimsft: Removed this mapping entirely

localization/strings/en-US/Resources.resw
dabhattimsft: Fixed

localization/strings/en-US/Resources.resw
dabhattimsft: Fixed

localization/strings/en-US/Resources.resw
```diff
@@ -248,12 +248,9 @@ Please use '{} --help' to get a list of supported arguments.</value>
...
   </data>
   <data name="MessageNoDefaultDistro" xml:space="preserve">
-    <value>Windows Subsystem for Linux has no installed distributions.
-You can resolve this by installing a distribution with the instructions below:
-
-Use 'wsl.exe --list --online' to list available distributions
-and 'wsl.exe --install &lt;Distro&gt;' to install.</value>
-    <comment>{Locked="--list "}{Locked="--online'"}{Locked="--install "}Command line arguments, file names and string inserts should not be translated</comment>
+    <value>No Linux distribution is installed.
+Run 'wsl.exe --install &lt;Distro&gt;' to install one (for example, 'wsl.exe --install Ubuntu').</value>
```
benhillis: I'm not convinced this is better than before

localization/strings/en-US/Resources.resw
```diff
@@ -305,8 +302,9 @@ For information please visit https://aka.ms/enablevirtualization</value>
     <value>Too many virtual hard disks or physical disks are attached.</value>
   </data>
   <data name="MessageUserVhdAlreadyAttached" xml:space="preserve">
-    <value>VHD already mounted via wsl.exe --mount, please unmount the disk before launching.</value>
-    <comment>{Locked="--mount,"}Command line arguments, file names and string inserts should not be translated</comment>
+    <value>The virtual disk '{}' is already attached.
+Unmount it with 'wsl.exe --unmount &lt;DiskPath&gt;' before attaching again.</value>
+    <comment>{FixedPlaceholder="{}"}{Locked="--unmount "}Command line arguments, file names and string inserts should not be translated</comment>
```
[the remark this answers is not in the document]
dabhattimsft: @craigloewen-msft/ @benhillis/ @OneBlue does this get handled by loc pipeline?

localization/strings/en-US/Resources.resw
craigloewen-msft: I agree. I prefer the original over this change.

craigloewen-msft (changes requested):

localization/strings/en-US/Resources.resw
dabhattimsft: reverted to original

localization/strings/en-US/Resources.resw
```diff
@@ -1013,7 +1014,12 @@ Falling back to NAT networking.</value>
...
   </data>
   <data name="MessageLocalSystemNotSupported" xml:space="preserve">
-    <value>Running WSL as local system is not supported.</value>
+    <value>WSL cannot run under the LocalSystem account.
+Use a standard user account instead.</value>
+  </data>
+  <data name="MessageDetachDiskNotFound" xml:space="preserve">
+    <value>Disk not found: {}</value>
+    <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
```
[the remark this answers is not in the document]
dabhattimsft: Will be fixed by nighty loc pipeline

src/windows/common/wslutil.cpp
```diff
@@ -768,7 +769,7 @@ std::wstring wsl::windows::common::wslutil::GetErrorString(HRESULT result)
         return Localization::MessageTooManyDisks();
 
     case WSL_E_USER_VHD_ALREADY_ATTACHED:
-        return Localization::MessageUserVhdAlreadyAttached();
+        return Localization::MessageUserVhdAlreadyAttached2(c_unknownDiskName);
```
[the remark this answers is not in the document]
benhillis: I think this is accurate, probably makes sense to have both especially cince c_unknownDiskName is not localized.

benhillis (changes requested):

src/windows/common/wslutil.cpp
```diff
@@ -33,6 +33,7 @@ constexpr auto c_releaseListUrl = L"https://api.github.com/repos/Microsoft/WSL/r
 constexpr auto c_specificReleaseListUrl = L"https://api.github.com/repos/Microsoft/WSL/releases/tags/";
 constexpr auto c_userAgent = L"wsl-install"; // required to use the GitHub API
 constexpr auto c_pipePrefix = L"\\\\.\\pipe\\";
+constexpr auto c_unknownDiskName = L"<unknown>";
```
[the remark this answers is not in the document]
benhillis: agreed, let's have two separate strings, one without a disk name, one with.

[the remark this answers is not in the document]
dabhattimsft: fixed

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -1333,7 +1334,7 @@ std::pair<int, LX_MINI_MOUNT_STEP> WslCoreVm::DetachDisk(_In_opt_ PCWSTR Disk)
         }
     }
 
-    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND), !deleted);
+    THROW_HR_WITH_USER_ERROR_IF(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND), Localization::MessageDetachDiskNotFound(Disk ? Disk : L""), !deleted);
```
benhillis: Is printing an empty string for Disk really what we want here?

dabhattimsft: Should we have separate error string for that scenario?

dabhattimsft: Should we have separate error string for that scenario?

benhillis: yes, I think an empty disk string is not very helpful.

dabhattimsft: fixed

benhillis:
@dabhattimsft Hey! This PR has been quiet for a while — is it still something you're planning to land? Let us know, otherwise we'll close it out. Thanks!

dabhattimsft:
Discussed offline and closing this one for now.

PR #14292, opened 2026-02-26
wsla: move CollectCrashDumps from HcsVirtualMachine to WSLAVirtualMachine

benhillis:
This change moves Linux user space crash collect out of the system service and into the wslasession process which runs as the user. This is preferrable because these crashdumps will be written as the user and do not require any impersonation.

### DIFF
--- a/src/windows/wslasession/WSLAVirtualMachine.cpp
+++ b/src/windows/wslasession/WSLAVirtualMachine.cpp
@@ -46,6 +46,12 @@ void WSLAVirtualMachine::Initialize()
 {
     THROW_IF_FAILED(m_vm->GetId(&m_vmId));
 
+    // Start crash dump collection thread.
+    auto crashDumpSocket = hvsocket::Listen(m_vmId, LX_INIT_UTILITY_VM_CRASH_DUMP_PORT);
+    THROW_LAST_ERROR_IF(!crashDumpSocket);
+
+    m_crashDumpThread = std::thread{[this, socket = std::move(crashDumpSocket)]() mutable { CollectCrashDumps(std::move(socket)); }};
+
     // Establish a socket channel with mini_init in the VM.
     wil::unique_socket socket;
     THROW_IF_FAILED(m_vm->AcceptConnection(reinterpret_cast<HANDLE*>(&socket)));
@@ -99,6 +105,11 @@ WSLAVirtualMachine::~WSLAVirtualMachine()
     if (m_processExitThread.joinable())
     {
         m_processExitThread.join();
+    }
+
+    if (m_crashDumpThread.joinable())
+    {
+        m_crashDumpThread.join();
     }
 
     // Clear the state of all remaining processes now that the VM has exited.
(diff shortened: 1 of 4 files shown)


OneBlue (approved):
I think this is a good short term fix. The long term fix would be to move the crash dump collection thread to WSLASession.exe and use the IO thread to collect crash dumps

benhillis:
> I think this is a good short term fix. The long term fix would be to move the crash dump collection thread to WSLASession.exe and use the IO thread to collect crash dumps



> I think this is a good short term fix. The long term fix would be to move the crash dump collection thread to WSLASession.exe and use the IO thread to collect crash dumps

Agreed, let me take a quick look at that, I think I might have been confused between this and the vmgs tool crashes.

src/windows/wslasession/WSLAVirtualMachine.cpp
```diff
@@ -46,6 +46,12 @@ void WSLAVirtualMachine::Initialize()
 {
     THROW_IF_FAILED(m_vm->GetId(&m_vmId));
 
+    // Start crash dump collection thread.
+    auto crashDumpSocket = hvsocket::Listen(m_vmId, LX_INIT_UTILITY_VM_CRASH_DUMP_PORT);
+    THROW_LAST_ERROR_IF(!crashDumpSocket);
+
+    m_crashDumpThread = std::thread{[this, socket = std::move(crashDumpSocket)]() mutable { CollectCrashDumps(std::move(socket)); }};
```
benhillis, line 53: I am creating this thread early to catch any early crashes.

src/windows/wslasession/WSLAVirtualMachine.cpp
```diff
@@ -894,3 +905,71 @@ wil::unique_socket WSLAVirtualMachine::ConnectUnixSocket(const char* Path)
...
+                       e.path().filename().string().find(dumpPrefix) == 0;
+            };
+
+            wslutil::EnforceFileLimit(crashDumpFolder.c_str(), 10, pred);
+
+            wil::unique_hfile file{CreateFileW(fullPath.c_str(), GENERIC_WRITE, 0, nullptr, CREATE_NEW, FILE_ATTRIBUTE_TEMPORARY, nullptr)};
+            THROW_LAST_ERROR_IF(!file);
+
+            channel.SendResultMessage<std::int32_t>(0);
+            relay::InterruptableRelay(reinterpret_cast<HANDLE>(channel.Socket()), file.get(), nullptr);
```
OneBlue, line 971: not part of the change, but we should pass `m_vmTerminatingEvent` to avoid getting blocked the guest side ever gets stuck

OneBlue (approved):
LGTM

src/windows/wslasession/WSLAVirtualMachine.cpp
benhillis: Ok I can make a pass in a future change to make sure we're passing the terminating event everywhere we should be.

PR #14293, opened 2026-02-26
Disable the process creation cursor animation by default

OneBlue:

This sets the `STARTF_FORCEOFFFEEDBACK` by default whenever we create new Windows processes. 

This particularly useful during testing, since it prevents the cursor from jumping between "normal" and "process creation pending" state back and forth. 

The flag is left untouched when creating interop processes, since this path can create interactive GUI processes on behalf of the user

## PR Checklist


benhillis (approved):
Awesome, I did not know about this flag.

PR #14294, opened 2026-02-26
WSLASession::Terminate: stop IORelay before clearing containers

benhillis:
Prevents an ABBA deadlock between Terminate and container event callbacks. Terminate holds session m_lock then container destruction acquires the event tracker lock. Concurrently, an IORelay event callback holds the event tracker lock then ReleaseResources calls GetVirtualMachine which tries to acquire session m_lock.

By stopping the IORelay first, no new docker event callbacks can fire during container teardown, breaking the deadlock cycle.

### DIFF
--- a/src/windows/wslasession/WSLASession.cpp
+++ b/src/windows/wslasession/WSLASession.cpp
@@ -968,16 +968,14 @@ try
 
     std::lock_guard lock{m_lock};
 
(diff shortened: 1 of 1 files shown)


OneBlue (commented):
I'd prefer to fix this via: https://github.com/microsoft/WSL/pull/14269 

Which removes the need for `GetVirtualMachine()`

PR #14299, opened 2026-02-27
Initial support for wslc E2E testing

AmelBawa-msft:
### Code summary
- Added the foundation for E2E testing for WSLC
- Added E2E tests

### Tests added
- WSLCE2E_Container_Create_HelpCommand
- WSLCE2E_Container_Create_MissingImage
- WSLCE2E_Container_Create_InvalidImage
- WSLCE2E_Container_Create_Valid
- WSLCE2E_Container_Create_DuplicateContainerName
- WSLCE2E_Container_HelpCommand
- WSLCE2E_Container_InvalidCommand_DisplaysErrorMessage
- WSLCE2E_HelpCommand
- WSLCE2E_InvalidCommand_DisplaysErrorMessage

### 

## PR Checklist


test/windows/wslc/WSLCE2ETests.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    WSLCExecutionResult ExecuteWSLC(const std::wstring& cmd)
+    {
+        auto fullCmd = L"C:\\src\\wsl\\bin\\x64\\Debug\\wslc.exe " + cmd;
+        wsl::windows::common::SubProcess process(nullptr, fullCmd.c_str());
+        const auto output = process.RunAndCaptureOutput();
+        return {.CommandLine = cmd, .Stdout = output.Stdout, .Stderr = output.Stderr, .ExitCode = HRESULT_FROM_WIN32(output.ExitCode)};
+    }
+
+    void VerifyOutput(const WSLCExecutionResult& result, const WSLCExecutionResult& expected) const
```
OneBlue: nit: Since this test class is specifically for wslc, we could simplify this into: 

```
void VerifyOutput(const std::wstring& Cmd, const std::wstring& Stdout, const std::wstring& Stderr = ", int ExitCode = 0);
```

That way we can write tests as:

```
    TEST_METHOD(WSLCE2E_HelpCommand_DisplaysCopyrightInfo)
    {
        VerifyOutput(L"--help", L"Copyright (c) Microsoft Corporation. All rights reserved.");
    }

```

OneBlue: Also I recommend testing output strings for an exact match (including CRLF) so we can be certain that the strings are exactly what we expect

OneBlue: Here's a good example of a similar construct that we use in other tests: https://github.com/microsoft/WSL/blob/80c100adb739d386bf8a5f1511561a6a0df71e67/test/windows/WSLATests.cpp#L186

test/windows/wslc/e2e/WSLCCommand.h
```diff
@@ -0,0 +1,45 @@
...
+Abstract:
+
+    This file contains helper functions for executing WSLC commands in end-to-end tests.
+--*/
+
+#pragma once
+
+#include "WSLCExecutorHelpers.h"
+
+#define WSLC_COMMAND(Name, Command) \
```
OneBlue: Having a macro to generate a variadic template function with variable name feels a bit overkill for factoring the command line arguments. 

Looking through this, it think it would be preferrable to specific the full command line when we run tests, this would make debugging quite a bit easier: 

```
auto result = RunWlsc("container create X");`
```

Will be easier to debug from the test logs

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -0,0 +1,144 @@
...
+        auto result = WSLCCommand::ContainerCreate("--name", WslcContainerName);
+        result.Verify({.Stdout = GetOutput(), .Stderr = L"Required argument not provided: 'image'\r\n", .ExitCode = E_INVALIDARG});
+    }
+
+    TEST_METHOD(WSLCE2E_Container_Create_InvalidImage)
+    {
+        auto result = WSLCCommand::ContainerCreate("--name", WslcContainerName, WslcInvalidImageName);
+
+        auto expectedError = L"Image '" + WslcInvalidImageName + L"' not found, pulling";
+        VERIFY_IS_TRUE(result.Stderr->find(expectedError) != std::wstring::npos);
```
OneBlue: I recommend validating for the exact stderr that we expect here to make sure that the format matches what we want

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -0,0 +1,116 @@
...
+void WSLCExecutor::ExecuteAndVerify(const std::wstring& cmd, const WSLCExecutionResult& expected)
+{
+    Execute(cmd).Verify(expected);
+}
+
+std::wstring GetWslcHeader()
+{
+    std::wstringstream header;
+    header << L"Windows Subsystem for Linux Container CLI (Preview) v1.0.0\r\n"
+           << L"Copyright (c) Microsoft Corporation. All rights reserved.\r\n\r\n";
```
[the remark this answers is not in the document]
AmelBawa-msft: We should address CLI versioning, the test aligns with what is hardcoded in code message. This test will start failing once that has been changed (which is good).

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -0,0 +1,116 @@
...
+        VERIFY_ARE_EQUAL(*expected.Stderr, *Stderr);
+    }
+
+    if (expected.ExitCode)
+    {
+        VERIFY_ARE_EQUAL(*expected.ExitCode, *ExitCode);
+    }
+}
+
+void WSLCExecutionResult::VerifyNoErrors(std::optional<std::wstring> expectedOutput) const
```
OneBlue: `VerifyNoErrors` is a bit misleading because having something written on stderr doesn't necessarily mean that errors occurred. 

We could probably have a single method for both this and the above, maybe: 

```
void WSLCExecutionResult::Verify(const std::wstring>& ExpectedStdout,
                                                      const std::wstring& ExpectedStdErr ="",
                                                      HRESULT = S_OK);
```

That way by default this would validate that there's nothing on stderr, but callers can override the second & third argument when needed

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+void WSLCExecutor::ExecuteAndVerify(const std::wstring& cmd, const WSLCExecutionResult& expected)
+{
+    Execute(cmd).Verify(expected);
+}
+
+std::wstring GetWslcHeader()
+{
+    std::wstringstream header;
+    header << L"Windows Subsystem for Linux Container CLI (Preview) v1.0.0\r\n"
```
OneBlue: I'm not sure where the version number is coming from in wslc, but we should use `WSL_PACKAGE_VERSION` for this

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -0,0 +1,167 @@
...
+        EnsureContainerDoesNotExist(WslcContainerName);
+
+        // Create the container with a valid image
+        auto result = WSLCCommand::ContainerCreate("--name", WslcContainerName, WslcUbuntuImageName);
+        result.VerifyNoErrors();
+        auto containerId = GetStdoutOneLine(result);
+
+        // Attempt to create another container with the same name
+        result = WSLCCommand::ContainerCreate("--name", WslcContainerName, WslcUbuntuImageName);
+        result.Dump();
```
OneBlue: I recommend avoiding to dump too much output when everything behaves as expected otherwise the test output can become different to work with. 

Maybe `Verify()` could dump on failure ?

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -0,0 +1,167 @@
...
+        result.Verify(
+            {.Stderr = L"Conflict. The container name \"/" + WslcContainerName + L"\" is already in use by container \"" +
```
OneBlue: nit: I recommend std::format to make this a bit easier to read: 

```
.Stderr = std::format(L"Conflict. The container name \"{}\" is already in use by container \"{}\"..., WslcContainerName , containerId)

File: test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp [hunk 9]
```diff
@@ -0,0 +1,167 @@
...
+                       containerId + L"\". You have to remove (or rename) that container to be able to reuse that name.\r\nError code: ERROR_ALREADY_EXISTS\r\n",
+             .ExitCode = 1});
+
+        // Clean up by deleting the container
+        result = WSLCCommand::ContainerDelete(WslcContainerName, "--force");
```
OneBlue: I recommend doing this in a scope_exit_log so that the container is always deleted even if the test fails

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -0,0 +1,167 @@
...
+        VERIFY_ARE_EQUAL(1, *result.ExitCode);
+    }
+
+    TEST_METHOD(WSLCE2E_Container_Create_Valid)
+    {
+        // Ensure the container does not already exist
+        EnsureContainerDoesNotExist(WslcContainerName);
+
+        // Create the container with a valid image
+        auto result = WSLCCommand::ContainerCreate("--name", WslcContainerName, WslcUbuntuImageName);
```
OneBlue: This will cause an image pull on ubuntu, which will likely cause failures in the tests because of the rate limit. Instead I recommend using debian:latest, which can be imported like this: https://github.com/microsoft/WSL/blob/079b6dbc56710b10cc53317c3301504530ed0c99/test/windows/WSLATests.cpp#L74 

(same for other images)

test/windows/wslc/e2e/WSLCE2EContainerTests.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+#include "precomp.h"
+#include "windows/Common.h"
+#include "WSLCCLITestHelpers.h"
+#include "WSLCExecutor.h"
+#include "WSLCCommand.h"
+
+namespace WSLCE2ETests {
+
+class WSLCE2EContainerTests
```
OneBlue, line 21: For now I would recommend writing the CLI tests under the same class, so that way we can have only one TestClassSetup() method where we import all the test images that the CLI tests need

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+private:
+    const std::wstring WslcContainerName = L"wslc-test-container";
+    const std::wstring WslcInvalidImageName = L"mcr.microsoft.com/invalid-image:latest";
+    const std::wstring WslcUbuntuImageName = L"ubuntu:latest";
+
+    std::wstring GetOutput() const
+    {
+        std::wstringstream output;
+        output << GetWslcHeader()        //
```
OneBlue, line 108: nit: Besides the wslc header, all other methods are only called in one place. 

We could simplify this a bit by using something like: 

```
static std::wstring GetHelpMessage()
{
    return GetWslcHeader() + LR("
Creates a container.
Usage: wslc container create [<options>] <image> [<command>] [<arguments>...]
The following arguments are available:
[...]
)";
```

I also recommend renaming to something like `GetHelpMessage()` since GetOutput() could be a bit misleading here

test/windows/wslc/e2e/WSLCExecutor.cpp
dkbennett: I have a PR to address this. The Version is not going to be in the header and the header is going to be a in the Resources. So can use the exact localization string here Localization::WSLCCLI_CopyrightHeader() once that change is merged in.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
dkbennett: The TestSetup and TestCleanup for anything that creates containers should also delete any containers the tests may create if they exist. If we have any leftover containers from a previous test this will taint the execution of subsequent tests and it would be better to find that issue in the TestSetup failing if it finds existing containers and was not able to delete them than to have strange behavior/failures in the test due to a corrupt environment that may be difficult to quickly identify.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
dkbennett: Given the entire help is generated and is easily testable without an end-to-end test, this might be noisy if we were to add/change arguments to a test and difficult to fix the test without waiting for tests to execute, which can take a long time.

I would recommend moving all the help output tests to unit tests which can be quickly run and validated locally if we are expecting certain formats for help and want to avoid regression here. Using an E2E test for this is overkill and would add some unnecessary execution time to the tests and make it take longer to find a problem we could find in a few seconds in a unit test.

test/windows/wslc/e2e/WSLCExecutor.cpp
AmelBawa-msft: @OneBlue, I started a thread about versioning and @dkbennett was quick with making a PR 😊 if the fix PR is completed soon, I will update this test to reflect the changes.

test/windows/wslc/e2e/WSLCExecutor.cpp
AmelBawa-msft: That's a fair suggestion. I do think having optional arguments is going to be helpful here since sometimes we cannot verify the output for a stream. For example, `wslc container create` outputs the hash for the created container. In that case, I am only interested in verifying that the stderr is empty and that the exit code is 0.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
AmelBawa-msft: Good catch! I added Dump to speed up test writing and for debugging purposes. That was an oversight on my part, I forgot to remove it from here.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
AmelBawa-msft: I added the `EnsureContainerDoesNotExist` in the test class/method hook methods.

test/windows/wslc/e2e/WSLCE2EContainerTests.cpp
AmelBawa-msft: I initially considered this approach, but I think splitting the tests into multiple classes will be better in the long run. It improves readability and scalability, and it reduces merge conflicts when multiple developers are adding tests. If the TestClassSetup ends up being duplicated across several classes, I’ll refactor it into a shared helper and invoke it from each class.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
AmelBawa-msft: Thanks to both of you for the feedback. I don’t have a strong preference here, so I’ll keep this as-is for now. We can revisit and adjust later if maintaining this code starts to become cumbersome.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
AmelBawa-msft: That approach is more reliable thank you for the suggestion! In WSLC we could use `wslc image load`, but it hasn’t been implemented yet, so I opened a PR to help unblock this.

- #14347

AmelBawa-msft: Updated

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -0,0 +1,160 @@
...
+        std::wstringstream expectedError;
+        expectedError << L"Image '" << InvalidImage.NameAndTag() << L"' not found, pulling\r\n"
+                      << L"pull access denied for library/"
+                      << InvalidImage.Name << L", repository does not exist or may require 'docker login': denied: requested access to the resource is denied\r\n"
+                      << L"Error code: WSLA_E_IMAGE_NOT_FOUND\r\n";
+        result.Verify({.Stderr = expectedError.str(), .ExitCode = 1});
+    }
+
+    TEST_METHOD(WSLCE2E_Container_Create_Valid)
+    {
```
benhillis, line 70: similar to the wslatests class I think we should only run these for the WSL2 test pass.

A while back I tried to split all the WSLA tests into a separate pass, which I think might also be worth doing, but let's just do my suggestion above for now otherwise the tests will effectively run twice.

test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
```diff
@@ -0,0 +1,103 @@
...
+    {
+        RunWslcAndVerify(L"--help", {.Stdout = GetHelpMessage(), .Stderr = L"", .ExitCode = 0});
+    }
+
+    TEST_METHOD(WSLCE2E_InvalidCommand_DisplaysErrorMessage)
+    {
+        RunWslcAndVerify(L"INVALID_CMD", {.Stdout = GetHelpMessage(), .Stderr = L"Unrecognized command: 'INVALID_CMD'\r\n", .ExitCode = 1});
+    }
+
+private:
```
benhillis, line 47: a bunch of this is duplicated, is it not?

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
```diff
@@ -0,0 +1,110 @@
...
+    result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+    auto outputLines = result.GetStdoutLines();
+    for (const auto& line : outputLines)
+    {
+        if (line.find(containerNameOrId) != std::wstring::npos)
+        {
+            const std::wstring message =
+                L"Container '" + containerNameOrId + L"' found in container list output but it should not be listed";
+            VERIFY_FAIL(message.c_str());
```
benhillis, line 49: huh? VERIFY_FAIL is supposed to check an hresult I think, you're passing a string.

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
```diff
@@ -0,0 +1,110 @@
...
+    result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+    auto outputLines = result.GetStdoutLines();
+    for (const auto& line : outputLines)
+    {
+        if (line.find(containerNameOrId) != std::wstring::npos)
+        {
+            const std::wstring message = L"Container '" + containerNameOrId + L"' found in container list output but status '" +
+                                         status + L"' was not found in the same line";
+            VERIFY_IS_TRUE(line.find(status) != std::wstring::npos, message.c_str());
```
benhillis: verify_is_true with != is an odd choice, maybe verify_not_equal is available?

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
```diff
@@ -0,0 +1,110 @@
...
+            return;
+        }
+    }
+
+    const std::wstring message = L"Container '" + containerNameOrId + L"' not found in container list output";
+    VERIFY_FAIL(message.c_str());
```
benhillis, line 72: Again, not sure what you're trying to do here.

benhillis (changes requested):
few questions.

test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
AmelBawa-msft: Assuming you’re referring to GetHelpMessage(), each workflow in wslc has its own distinct help message. While the overall structure or view may look similar, the actual content differs for each workflow.

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
AmelBawa-msft: From copilot, `VERIFY_FAIL(...)` accepts an optional message, which seems consistent with its usage elsehwere in the unit test code.

E.g. InstallerTests.cpp, MountTests.cpp, PluginTests.cpp, etc...

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
AmelBawa-msft: Same as comment above ☝️

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    TEST_METHOD(WSLCE2E_Container_Create_DuplicateContainerName)
+    {
+        WSL2_TEST_ONLY();
+        VerifyContainerIsNotListed(WslcContainerName);
+
+        // Create the container with a valid image
+        auto result = RunWslc(std::format(L"container create --name {} {}", WslcContainerName, DebianImage.NameAndTag()));
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
```
benhillis, line 90: S_OK is not an exit code, this should just be "0"

benhillis: here and elsewhere in the PR.

AmelBawa-msft: Good catch! I plan on having another follow up PR with additional tests, I can apply this change in the other PR 😊

benhillis (approved):

PR #14305, opened 2026-02-27
Fix AcceptEx stack buffer use-after-free in PortRelay

benhillis:
Move AcceptBuffer from stack-local variable in ScheduleAccept() to a member of the PortRelay struct. When AcceptEx returns WSA_IO_PENDING, the function returns and the stack frame is destroyed, but the OS still writes address data into AcceptBuffer on completion - causing a write to freed stack memory.

This matches the pattern used by SingleAcceptHandle in relay.hpp, which correctly stores AcceptBuffer as a class member.


OneBlue (approved):
Great catch !

PR #14306, opened 2026-02-27
Fix GetLabels freeing uninitialized pointers on error path

benhillis:
This change resolves a potential free of uninitialized pointers in the WSLAContainerImpl::GetLabels function. Previously the output array was not zero-initialized so if allocation of a string failed, the cleanup lambda would attempt to free random values.

Fix this by using RAII types to hold strings and transfer ownership to the caller after all allocations have succeeded.

### DIFF
--- a/src/windows/wslasession/WSLAContainer.cpp
+++ b/src/windows/wslasession/WSLAContainer.cpp
@@ -1320,27 +1320,27 @@ void WSLAContainerImpl::GetLabels(WSLA_LABEL_INFORMATION** Labels, ULONG* Count)
         return;
     }
 
-    auto count = m_labels.size();
-    auto labelsArray = wil::make_unique_cotaskmem<WSLA_LABEL_INFORMATION[]>(count);
+    // Build labels locally using RAII strings. If an allocation throws mid-loop,
+    // the vector destructor frees everything already built.
+    std::vector<std::pair<wil::unique_cotaskmem_ansistring, wil::unique_cotaskmem_ansistring>> localLabels;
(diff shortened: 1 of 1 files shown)


src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -1320,27 +1320,27 @@ void WSLAContainerImpl::GetLabels(WSLA_LABEL_INFORMATION** Labels, ULONG* Count)
         return;
     }
 
-    auto count = m_labels.size();
-    auto labelsArray = wil::make_unique_cotaskmem<WSLA_LABEL_INFORMATION[]>(count);
+    // Build labels locally using RAII strings. If an allocation throws mid-loop,
+    // the vector destructor frees everything already built.
+    std::vector<std::pair<wil::unique_cotaskmem_ansistring, wil::unique_cotaskmem_ansistring>> localLabels;
```
OneBlue, line 1325: I think that works, another approach we could take would be to do: 

```
    ZeroMemory(labelsArray.get(), sizeof(WSLA_LABEL_INFORMATION) * count);
```

Right after `make_unique_cotaskmem` returns, that way everything is zero initialized and we don't need to transfer ownership on failure

OneBlue (approved):
Good catch ! I was under the assumptions

src/windows/wslasession/WSLAContainer.cpp
benhillis: I prefer this, it lets us get rid of the ugly cleanup loop and RAII just handles it for us.

PR #14310, opened 2026-02-27
Make ForEachSession enforce noexcept lambdas

benhillis:
ForEachSession uses std::ranges::remove_if which leaves the container in an unspecified (partially-moved) state if the predicate throws. Previously, callers used THROW_IF_FAILED inside lambdas which could corrupt m_sessions.

Add static_assert(is_nothrow_invocable_v) to enforce noexcept at compile time. Update all 4 callers to return errors via the result optional instead of throwing:

- CreateSession: return CheckTokenAccess HRESULT directly
- OpenSession: return CheckTokenAccess HRESULT directly
- OpenSessionByName: return CheckTokenAccess HRESULT directly
- ListSessions: LOG_IF_WIN32_BOOL_FALSE + skip entry on SID conversion failure

### DIFF
--- a/src/windows/wslaservice/exe/WSLASessionManager.cpp
+++ b/src/windows/wslaservice/exe/WSLASessionManager.cpp
@@ -56,18 +56,18 @@ void WSLASessionManagerImpl::CreateSession(const WSLA_SESSION_SETTINGS* Settings
     std::lock_guard lock(m_wslaSessionsLock);
 
     // Check for an existing session first.
-    auto result = ForEachSession<HRESULT>([&](auto& entry, const wil::com_ptr<IWSLASession>& session) -> std::optional<HRESULT> {
-        if (wsl::shared::string::IsEqual(entry.DisplayName.c_str(), Settings->DisplayName))
+    auto result = ForEachSession<HRESULT>([&](auto& entry, const wil::com_ptr<IWSLASession>& session) noexcept -> std::optional<HRESULT> {
+        if (!wsl::shared::string::IsEqual(entry.DisplayName.c_str(), Settings->DisplayName))
         {
-            RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_ALREADY_EXISTS), WI_IsFlagClear(Flags, WSLASessionFlagsOpenExisting));
-
-            THROW_IF_FAILED(CheckTokenAccess(entry, tokenInfo));
-
-            session.copy_to(WslaSession);
-            return S_OK;
+            return {};
         }
 
-        return std::optional<HRESULT>{};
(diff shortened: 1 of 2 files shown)


src/windows/wslaservice/exe/WSLASessionManager.cpp
```diff
@@ -151,9 +149,12 @@ void WSLASessionManagerImpl::ListSessions(_Out_ WSLA_SESSION_INFORMATION** Sessi
...
-    ForEachSession<void>([&](auto& entry, const auto&) {
+    ForEachSession<void>([&](auto& entry, const auto&) noexcept {
         wil::unique_hlocal_string sidString;
-        THROW_IF_WIN32_BOOL_FALSE(ConvertSidToStringSidW(entry.Owner.TokenInfo->User.Sid, &sidString));
+        if (!LOG_IF_WIN32_BOOL_FALSE(ConvertSidToStringSidW(entry.Owner.TokenInfo->User.Sid, &sidString)))
+        {
+            return;
+        }
 
         auto& it = sessionInfo.emplace_back(WSLA_SESSION_INFORMATION{.SessionId = entry.SessionId, .CreatorPid = entry.CreatorPid});
```
benhillis: I think this can throw.

src/windows/wslaservice/exe/WSLASessionManager.cpp
```diff
@@ -56,18 +56,18 @@ void WSLASessionManagerImpl::CreateSession(const WSLA_SESSION_SETTINGS* Settings
...
-            THROW_IF_FAILED(CheckTokenAccess(entry, tokenInfo));
+        RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_ALREADY_EXISTS), WI_IsFlagClear(Flags, WSLASessionFlagsOpenExisting));
 
-            session.copy_to(WslaSession);
-            return S_OK;
-        }
+        RETURN_IF_FAILED(CheckTokenAccess(entry, tokenInfo));
 
-        return std::optional<HRESULT>{};
+        session.copy_to(WslaSession);
```
OneBlue: Not part of the change, but we should RETURN_IF_FAILED() here (it surprised me that the lambda was noexcept because I thought this could throw, but it doesn't)

OneBlue: (Same for bellow calls to copy_to)

OneBlue (approved):

src/windows/wslaservice/exe/WSLASessionManager.cpp
benhillis: I'll go ahead and fix these now.

PR #14314, opened 2026-02-27
Track `bind` syscall when port is 0

damanm24:
With mirrored and virtio networking, we observed an issue when trying to connect to VSCode remote instance inside a WSL environment. The issue is that VSCode will call bind with the localhost IP address with port 0 specified. This indicates to the kernel that it should assign any available ephemeral port to the requesting process. 

In WSL's port tracker, we ignore `bind` syscalls where the port is 0 because the kernel will assign one when the syscall completes. When the port is not 0, we send a PortMappingRequest to the host, which will create a new host-side socket bound to the specified local address and port where incoming connections can be listened to. 

To address the issue, we should still track the `bind` syscall when the port is 0, but follow-up after the syscall has been completed. We do this by duplicating the socket that was created and calling `getsockname` to check was port has been assigned. Then we can send the same PortMappingRequest to the host.

### DIFF
--- a/src/linux/init/GnsPortTracker.cpp
+++ b/src/linux/init/GnsPortTracker.cpp
@@ -14,6 +14,14 @@
 #include "GnsPortTracker.h"
 #include "lxinitshared.h"
 
+// pidfd_open (Linux 5.3) and pidfd_getfd (Linux 5.6) may not be defined in older headers.
+#ifndef SYS_pidfd_open
+#define SYS_pidfd_open 434
+#endif
+#ifndef SYS_pidfd_getfd
+#define SYS_pidfd_getfd 438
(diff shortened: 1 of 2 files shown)


src/linux/init/GnsPortTracker.cpp
```diff
@@ -14,6 +14,14 @@
 #include "GnsPortTracker.h"
 #include "lxinitshared.h"
 
+// pidfd_open (Linux 5.3) and pidfd_getfd (Linux 5.6) may not be defined in older headers.
+#ifndef SYS_pidfd_open
```
damanm24, line 18: Defined in <sys/syscall.h> remove.

PR #14316, opened 2026-02-27
wsla: narrow Stop() timeout from LONGLONG to LONG

benhillis:
LONGLONG allows values > ULONG_MAX which silently truncate to 0 when cast to ULONG for the Docker API, causing an immediate kill with no grace period. LONG (max ~68 years) is more than sufficient and eliminates the truncation risk. Negative values still serve as the 'no timeout / use default' sentinel.


benhillis:
the copilot messages about COM interface are correct, but we are still free to change this interface since it's internal only.

OneBlue (approved):

PR #14318, opened 2026-02-27
wslc: change default networking mode to virtio proxy

benhillis:
We are expecting to ship with virtio proxy networking mode (not NAT), this change switches wslc.exe to use the right networking mode.

### DIFF
--- a/src/windows/wslc/services/SessionModel.cpp
+++ b/src/windows/wslc/services/SessionModel.cpp
@@ -26,7 +26,7 @@ SessionOptions SessionOptions::Default()
     options.m_sessionSettings.BootTimeoutMs = 30 * 1000;
     options.m_sessionSettings.StoragePath = m_defaultPath.c_str();
     options.m_sessionSettings.MaximumStorageSizeMb = 10000; // 10GB.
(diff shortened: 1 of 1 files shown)


src/windows/wslc/services/SessionModel.cpp
```diff
@@ -26,7 +26,7 @@ SessionOptions SessionOptions::Default()
     options.m_sessionSettings.BootTimeoutMs = 30 * 1000;
     options.m_sessionSettings.StoragePath = m_defaultPath.c_str();
     options.m_sessionSettings.MaximumStorageSizeMb = 10000; // 10GB.
-    options.m_sessionSettings.NetworkingMode = WSLANetworkingModeNAT;
+    options.m_sessionSettings.NetworkingMode = WSLANetworkingModeVirtioProxy;
```
ptrivedi, line 29: @benhillis: The idea is to start getting some test coverage as the team works with the new networking mode? We will do any additional sanity and functional tests in parallel followed by performance testing?

benhillis: Correct, I want to identify any issues up front. We've already switched to virtio proxy in the tests and other entrypoints.

OneBlue (approved):

PR #14320, opened 2026-02-28
Deferred mounting volumes and mapping ports to Start

kvega005:

This PR defers port reservation and volume mounting to after the Docker container is started. Previously these resources were set up during `Create()`, but they only need to be active while the container is running. This also enables bridge mode support, where Docker assigns the VM port dynamically during container start.

### Changes

Deferred resource setup 
 - Moved volume mounting and host port mapping from `Create()` to `Start()`
 - Added `ReleaseRuntimeResources()` to release port relays and unmount volumes when the container stops (via `OnStopped()` or `Delete()`)

VM port allocation at Create time (WSLAContainer / WSLAVirtualMachine)
 - For bridge network mode, VM ports are allocated from the VM port pool during `Create()` so they can be passed to Docker's `PortBindings` in the container create request
 - For host mode, `VmPort == ContainerPort`, no allocation needed
 - VM port allocations are released in `ReleaseResources()` (on container deletion)
 
 Port relay activation at Start time
 - MapPorts() binds the host port and activates the relay to the pre-allocated VM port at Start() time
 - UnmapPorts() tears down the relay when the container stops

Volume mount lifecycle
 - MountVolumes() mounts host folders into the parent VM at Start() time
 - UnmountVolumes() unmounts them when the container stops.

## PR Checklist


src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -1004,17 +1031,15 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Open(
     auto metadata = ParseContainerMetadata(metadataIt->second.c_str());
     labels.erase(metadataIt);
 
-    // TODO: Offload volume mounting and port mapping to the Start() method so that its still possible
-    // to open containers that are not running.
-    auto volumeErrorCleanup = MountVolumes(metadata.Volumes, virtualMachine);
-    auto [vmPorts, errorCleanup] = MapPorts(metadata.Ports, virtualMachine);
+    WI_ASSERT(dockerContainer.State != ContainerState::Running);
```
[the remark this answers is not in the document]
kvega005: I think we can still assume containers are not running because RecoverExistingContainers() should be called immediately after the session starts, so the there can only be containers created from previous sessions. None should be running since the VM would have been shut down and container would have been stopped when the previous session exited.

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -466,7 +466,7 @@ interface IWSLASession : IUnknown
     HRESULT MountWindowsFolder([in] LPCWSTR WindowsPath, [in] LPCSTR LinuxPath, [in] BOOL ReadOnly);
     HRESULT UnmountWindowsFolder([in] LPCSTR LinuxPath);
     HRESULT MapVmPort([in] int Family, [in] short WindowsPort, [in] short LinuxPort);
-    HRESULT UnmapVmPort([in] int Family, [in] short WindowsPort, [in] short LinuxPort);
+    HRESULT UnmapVmPort([in] int Family, [in] short WindowsPort);
```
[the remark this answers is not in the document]
kvega005: Fair enough. We don't need to change the com interface. It just doesn't make a  lot of sense to pass around a parameter that is completely ignored.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -81,74 +79,76 @@ auto ValidatePortMappings(const WSLA_CONTAINER_OPTIONS& options)
...
+}
+
+// Resolve VM ports from Docker's inspect data for bridge mode ports (where VmPort == 0).
+void ResolveVmPorts(std::vector<WSLAPortMapping>& ports, const DockerInspectContainer& inspect)
+{
+    for (auto& e : ports)
+    {
+        if (e.VmPort != 0)
+        {
+            continue; // Host mode port — VM port is already known.
```
ptrivedi: What does it mean by "host mode port" here?

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -466,16 +475,44 @@ void WSLAContainerImpl::Start(WSLAContainerStartFlags Flags)
...
         m_dockerClient.StartContainer(m_id);
     }
     CATCH_AND_THROW_DOCKER_USER_ERROR("Failed to start container '%hs'", m_id.c_str());
 
+    // For bridge mode, inspect the container to discover the VM ports Docker assigned.
+    if (m_networkMode == "bridge")
+    {
+        auto dockerInspect = m_dockerClient.InspectContainer(m_id);
+        ResolveVmPorts(m_mappedPorts, dockerInspect);
+    }
```
ptrivedi: If the networking mode is not "bridge", then we should not call ActivatePortRelays. Is the idea that the CLI/API will sanity check that if the port mappings are provided, then the networking mode would have to be 'bridge'? In the standard case, for host network containers, the port mapping is not passed by the user.

ptrivedi: If any of these above calls fail, do we want to stop/delete the started container? Otherwise we will have a started container without the VM ports resolved

src/windows/wslasession/WSLAVirtualMachine.cpp
```diff
@@ -681,14 +681,19 @@ void WSLAVirtualMachine::MapPortImpl(_In_ int Family, _In_ short WindowsPort, _I
     THROW_IF_FAILED_MSG(result, "Failed to map port: WindowsPort=%d, LinuxPort=%d, Family=%d, Remove=%d", WindowsPort, LinuxPort, Family, Remove);
 }
 
+void WSLAVirtualMachine::ReserveHostPort(_In_ int Family, _In_ short WindowsPort)
+{
+    MapPortImpl(Family, WindowsPort, 0, false);
```
ptrivedi: nit: this last bool parameter is very confusing. Ideally we should use a meaningful enum here to indicate that false means "we are NOT unmapping ports" and true means "we are unmapping ports"

At the very least, add comments next to these.

src/windows/wslasession/WSLAVirtualMachine.cpp
```diff
@@ -681,14 +681,19 @@ void WSLAVirtualMachine::MapPortImpl(_In_ int Family, _In_ short WindowsPort, _I
...
 void WSLAVirtualMachine::MapPort(_In_ int Family, _In_ short WindowsPort, _In_ short LinuxPort)
 {
     MapPortImpl(Family, WindowsPort, LinuxPort, false);
 }
 
-void WSLAVirtualMachine::UnmapPort(_In_ int Family, _In_ short WindowsPort, _In_ short LinuxPort)
+void WSLAVirtualMachine::UnmapPort(_In_ int Family, _In_ short WindowsPort)
 {
-    MapPortImpl(Family, WindowsPort, LinuxPort, true);
+    MapPortImpl(Family, WindowsPort, 0, true);
```
ptrivedi: nit: again, a comment saying unmap does not require LinuxPort to be provided could be very helpful for readability

src/windows/wslrelay/localhost.cpp
```diff
@@ -468,7 +473,7 @@ void AcceptThread(std::vector<std::shared_ptr<PortRelay>>& ports, const GUID& Vm
         std::vector<HANDLE> events{ExitEvent};
         for (auto& e : ports)
         {
-            if (!e->Pending)
+            if (!e->Pending && e->IsActive())
```
ptrivedi: The pending and active checks are not happening atomically, could this cause a race? Similar questions for the port mapping related functionality happening from within this thread

test/windows/WSLATests.cpp
```diff
@@ -4428,6 +4428,64 @@ class WSLATests
...
+            launcher.SetName("deferred-port-2");
+            auto container2 = launcher.Create(*m_defaultSession);
+            VERIFY_ARE_EQUAL(container2.State(), WslaContainerStateCreated);
+
+            // Start container — should succeed.
+            VERIFY_SUCCEEDED(container.Get().Start(WSLAContainerStartFlagsNone));
+            VERIFY_ARE_EQUAL(container.State(), WslaContainerStateRunning);
+
+            // Start container 2 — should fail because the host port is already reserved by container 1.
+            VERIFY_ARE_EQUAL(container2.Get().Start(WSLAContainerStartFlagsNone), HRESULT_FROM_WIN32(ERROR_ALREADY_EXISTS));
```
ptrivedi: A more meaningful error message around port mapping failure due to conflict would be useful

src/windows/wslasession/WSLAContainer.h
```diff
@@ -128,6 +130,7 @@ class WSLAContainerImpl
     WSLA_CONTAINER_STATE m_state = WslaContainerStateInvalid;
     WSLASession& m_wslaSession;
     WSLAVirtualMachine& m_virtualMachine;
+    std::string m_networkMode;
```
OneBlue: I recommend storing the networking mode as `WSLA_CONTAINER_NETWORK_TYPE` instead of a string here

src/windows/wslasession/WSLAVirtualMachine.cpp
OneBlue: If we want to be able to map host ports without actually starting to relay, we should flags to signal the difference between: 

1) I have previously reserved this port, and now I want to start relaying to guest port X
2) I want to map this host port for the first time 

Otherwise this can create a race where if let's say container A is starting and does 1), container B could come and `steal` the port by doing 2) before container A is done allocating the port fully

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -81,74 +79,76 @@ auto ValidatePortMappings(const WSLA_CONTAINER_OPTIONS& options)
...
-        e.MappedToHost = true;
     }
+}
+
+// Resolve VM ports from Docker's inspect data for bridge mode ports (where VmPort == 0).
+void ResolveVmPorts(std::vector<WSLAPortMapping>& ports, const DockerInspectContainer& inspect)
+{
+    for (auto& e : ports)
+    {
+        if (e.VmPort != 0)
```
OneBlue: This shouldn't be possible right ? ResolveVmPorts() is only called in bridge mode, and in this mode we always set VmPort to 0

OneBlue: If that's the case, I'd recommend switching to a WI_ASSERT here

src/windows/wslrelay/localhost.cpp
```diff
@@ -404,6 +404,11 @@ struct PortRelay
         }
     }
 
+    bool IsActive() const
+    {
+        return LinuxPort != 0;
```
OneBlue: `LinuxPort` should be changed to an std::atomic here since we're now reading & writing it from multiple threads. 

In truth we should probably rewrite this relay logic, but for now let's just make it an std::atomic

src/windows/wslasession/WSLAContainer.cpp
OneBlue: The port relays need to activated in host mode as well

kvega005: That's actually a good point. I need to rethink this a bit. I think if we wait to start the container then fail to activate the port relays, we would inadvertently have allowed the init process to run for a brief period of time. This means there may be a side effect. We probably do not want to fail the start container call after the init process has ran for some time.

src/windows/inc/docker_schema.h
```diff
@@ -144,6 +145,13 @@ struct InspectMount
     NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(InspectMount, Type, Source, Destination, RW);
 };
 
+struct NetworkSettings
```
OneBlue: Looks like we don't need this anymore (same for Binds)

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -1107,19 +1146,42 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Open(
...
     labels.erase(metadataIt);
 
-    // TODO: Offload volume mounting and port mapping to the Start() method so that its still possible
-    // to open containers that are not running.
-    auto volumeErrorCleanup = MountVolumes(metadata.Volumes, virtualMachine);
-    auto [vmPorts, errorCleanup] = MapPorts(metadata.Ports, virtualMachine);
+    // Re-register recovered VM ports in the allocation pool to prevent conflicts.
+    // Only needed for containers that haven't stopped yet — stopped containers release their
+    // VM port allocations, so those ports are available for reuse.
+    std::set<uint16_t> allocatedVmPorts;
```
OneBlue: We should have a scope_exit() to deallocate those if an exception is thrown below.

test/windows/WSLATests.cpp
```diff
@@ -4930,6 +4930,64 @@ class WSLATests
...
+        // Two containers created with the same host port, only the first Start() succeeds.
+        {
+            WSLAContainerLauncher launcher("debian:latest", "deferred-port", {"sleep", "99999"}, {}, WSLA_CONTAINER_NETWORK_BRIDGE);
+            launcher.AddPort(1240, 8000, AF_INET);
+
+            // Both Create() calls should succeed because ports are not reserved until Start().
+            auto container = launcher.Create(*m_defaultSession);
+            VERIFY_ARE_EQUAL(container.State(), WslaContainerStateCreated);
+
+            launcher.SetName("deferred-port-2");
```
OneBlue, line 5135: Ah ! That's a clever way to reuse the same launcher

OneBlue (commented):
LGTM, minor comments

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -492,12 +530,18 @@ void WSLAContainerImpl::Start(WSLAContainerStartFlags Flags, LPCSTR DetachKeys)
...
     m_stoppedNotifiedEvent.ResetEvent();
 
+    auto volumeCleanup = MountVolumes(m_mountedVolumes, m_virtualMachine);
+    auto portCleanup = MapPorts(m_mappedPorts, m_virtualMachine);
+
     try
     {
         m_dockerClient.StartContainer(m_id, DetachKeys == nullptr ? std::nullopt : std::optional<std::string>(DetachKeys));
     }
     CATCH_AND_THROW_DOCKER_USER_ERROR("Failed to start container '%hs'", m_id.c_str());
```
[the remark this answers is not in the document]
kvega005: Updated PR description.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -89,70 +89,81 @@ auto ValidatePortMappings(const WSLAContainerOptions& options)
...
+
+auto AllocateVmPorts(std::vector<WSLAPortMapping>& ports, WSLAVirtualMachine& vm)
+{
+    auto allocatedPorts = std::make_shared<std::set<uint16_t>>();
+    auto errorCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [allocatedPorts, &vm]() { ReleaseVmPorts(*allocatedPorts, vm); });
 
-    // Check if we need to allocate VM ports for bridge mode (VmPort == 0).
-    size_t portsToAllocate = std::count_if(ports.begin(), ports.end(), [](const auto& p) { return p.VmPort == 0; });
+    // Count ports that need VM port allocation (bridge mode ports have VmPort == 0).
+    auto portsToAllocate = static_cast<uint16_t>(std::ranges::count_if(ports, [](const auto& p) { return p.VmPort == 0; }));
```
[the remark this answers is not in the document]
kvega005: This is by design. We are ok with Host mode containers failing if a Bridge mode container is using its port.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -249,6 +268,24 @@ WSLAContainerState DockerStateToWSLAState(ContainerState state)
...
+    {
+        return WSLAContainerNetworkTypeHost;
+    }
+    else if (mode == "none")
+    {
+        return WSLAContainerNetworkTypeNone;
+    }
+
+    THROW_HR_MSG(E_INVALIDARG, "Invalid networking mode: %hs", mode.c_str());
+}
```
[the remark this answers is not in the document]
kvega005: I don't think this is accurate. According to Docker Engine API (1.53), supported values for network mode are "bridge", "host", "none", and "container:<name|id>". Any other value is taken as a custom network's name to which this container should connect to.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -1131,22 +1190,36 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Open(
...
+
+    // Re-register recovered VM ports in the allocation pool to prevent conflicts.
+    auto vmPorts = std::make_shared<std::set<uint16_t>>();
+    if (networkMode == WSLAContainerNetworkTypeBridged)
+    {
+        for (const auto& port : metadata.Ports)
+        {
+            WI_ASSERT(port.VmPort != 0);
+            vmPorts->insert(port.VmPort);
+        }
```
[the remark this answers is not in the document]
kvega005: This is by design.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -89,70 +89,81 @@ auto ValidatePortMappings(const WSLAContainerOptions& options)
...
+    auto allocatedPorts = std::make_shared<std::set<uint16_t>>();
+    auto errorCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [allocatedPorts, &vm]() { ReleaseVmPorts(*allocatedPorts, vm); });
 
-    // Check if we need to allocate VM ports for bridge mode (VmPort == 0).
-    size_t portsToAllocate = std::count_if(ports.begin(), ports.end(), [](const auto& p) { return p.VmPort == 0; });
+    // Count ports that need VM port allocation (bridge mode ports have VmPort == 0).
+    auto portsToAllocate = static_cast<uint16_t>(std::ranges::count_if(ports, [](const auto& p) { return p.VmPort == 0; }));
 
     if (portsToAllocate > 0)
     {
```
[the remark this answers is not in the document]
kvega005: This is by design.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -89,70 +89,81 @@ auto ValidatePortMappings(const WSLAContainerOptions& options)
     }
 }
 
-auto MapPorts(std::vector<WSLAPortMapping>& ports, WSLAVirtualMachine& vm)
+void UnmapPorts(std::vector<WSLAPortMapping>& ports, WSLAVirtualMachine& vm)
```
OneBlue, line 92: nit: Now we don't do these operations during Create() anymore, we could move `UnmapPorts()`, `MapPorts()`, `ReleaseVmPorts()` and `AllocateVmPorts()` inside WSLAContainerImpl

OneBlue (approved):
LGTM

PR #14321, opened 2026-02-28
Add tmpfs mount support for WSLA containers

beena352:
Adds tmpfs mount support for WSLA containers. Tmpfs mounts are memory-backed filesystems commonly used for temporary data (e.g., /run, /tmp). This change enables callers to specify tmpfs mounts when creating containers via the WSLA API, passes them through to Docker's HostConfig.Tmpfs, and surfaces them in container inspect output.

## PR Checklist

**IDL changes (wslaservice.idl)**
- Added WSLA_TMPFS_MOUNT struct with Destination (required) and [unique] Options (nullable) fields
- Added [unique, size_is(TmpfsCount)] Tmpfs and TmpfsCount to WSLA_CONTAINER_OPTIONS, following the existing Volumes/Labels pattern

 **Docker schema (docker_schema.h)**
- Added std::map<std::string, std::string> Tmpfs to HostConfig and registered it with the JSON serialization macro

**Container creation (WSLAContainer.cpp)**
- In Create(): validates Tmpfs array pointer when TmpfsCount > 0, validates each Destination is non-null, handles nullable Options with fallback to "", and populates request.HostConfig.Tmpfs
- In BuildInspectContainer(): iterates dockerInspect.HostConfig.Tmpfs and populates InspectMount entries with Type = "tmpfs", ReadWrite = true

**Client launcher (WSLAContainerLauncher.h/.cpp)**
- Added AddTmpfs(const std::string& ContainerPath, const std::string& Options) following the AddLabel/AddVolume pattern with std::deque for stable LPCSTR pointer storage
- Wired TmpfsCount/Tmpfs into CreateNoThrow() using the existing size() > 0 ? .data() : nullptr pattern

### DIFF
--- a/src/windows/wslaservice/inc/wslaservice.idl
+++ b/src/windows/wslaservice/inc/wslaservice.idl
@@ -169,6 +169,12 @@ struct WSLA_LABEL
     [string] LPCSTR Value;
 };
 
+struct WSLA_TMPFS_MOUNT
+{
+    LPCSTR Destination;
+    [unique] LPCSTR Options;
+};
+
 struct WSLA_LABEL_INFORMATION
 {
     [string] LPSTR Key;
@@ -232,6 +238,8 @@ struct WSLA_CONTAINER_OPTIONS
 
     ULONGLONG ShmSize;
     struct WSLA_CONTAINER_NETWORK ContainerNetwork;
+    [unique, size_is(TmpfsCount)] const struct WSLA_TMPFS_MOUNT* Tmpfs;
+    ULONG TmpfsCount;
 };
 
 enum WSLA_CONTAINER_STATE
--- a/src/windows/wslasession/WSLAContainer.cpp
+++ b/src/windows/wslasession/WSLAContainer.cpp
@@ -794,11 +794,11 @@ WslaInspectContainer WSLAContainerImpl::BuildInspectContainer(const DockerInspec
     }
 
     // Map tmpfs mounts from Docker inspect data.
-    for (const auto& [destination, options] : dockerInspect.HostConfig.Tmpfs)
+    for (const auto& entry : dockerInspect.HostConfig.Tmpfs)
     {
         wsla_schema::InspectMount mountInfo{};
         mountInfo.Type = "tmpfs";
-        mountInfo.Destination = destination;
+        mountInfo.Destination = entry.first;
         // Tmpfs mounts are read-write by default. We currently do not parse tmpfs options
         // (e.g. "ro") for inspect output; Docker enforces actual mount behavior.
         mountInfo.ReadWrite = true;
@@ -928,6 +928,12 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
             const auto& tmpfs = containerOptions.Tmpfs[i];
 
             THROW_HR_IF_NULL_MSG(E_INVALIDARG, tmpfs.Destination, "Tmpfs mount at index %lu has null destination", i);
+
+            THROW_HR_IF_MSG(
+                HRESULT_FROM_WIN32(ERROR_ALREADY_EXISTS),
+                request.HostConfig.Tmpfs.contains(tmpfs.Destination),
+                "Duplicate tmpfs destination: '%hs'",
+                tmpfs.Destination);
 
             request.HostConfig.Tmpfs[tmpfs.Destination] = tmpfs.Options != nullptr ? tmpfs.Options : "";
         }
--- a/src/windows/wslasession/WSLAContainer.cpp
+++ b/src/windows/wslasession/WSLAContainer.cpp
@@ -781,7 +781,7 @@ WslaInspectContainer WSLAContainerImpl::BuildInspectContainer(const DockerInspec
     }
 
     // Map volume mounts using WSLA's host-side data.
-    wslaInspect.Mounts.reserve(m_mountedVolumes.size());
+    wslaInspect.Mounts.reserve(m_mountedVolumes.size() + dockerInspect.HostConfig.Tmpfs.size());
     for (const auto& volume : m_mountedVolumes)
     {
         wsla_schema::InspectMount mountInfo{};
--- a/src/windows/wslasession/WSLAContainer.cpp
+++ b/src/windows/wslasession/WSLAContainer.cpp
@@ -793,6 +793,18 @@ WslaInspectContainer WSLAContainerImpl::BuildInspectContainer(const DockerInspec
         wslaInspect.Mounts.push_back(std::move(mountInfo));
     }
 
+    // Map tmpfs mounts from Docker inspect data.
+    for (const auto& [destination, options] : dockerInspect.HostConfig.Tmpfs)
+    {
+        wsla_schema::InspectMount mountInfo{};
+        mountInfo.Type = "tmpfs";
+        mountInfo.Destination = destination;
+        // Tmpfs mounts are read-write by default. We currently do not parse tmpfs options
+        // (e.g. "ro") for inspect output; Docker enforces actual mount behavior.
+        mountInfo.ReadWrite = true;
+        wslaInspect.Mounts.push_back(std::move(mountInfo));
+    }
+
     return wslaInspect;
 }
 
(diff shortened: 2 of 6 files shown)


test/windows/WSLATests.cpp
```diff
@@ -2273,6 +2273,21 @@ class WSLATests
...
 
+        // Validate that tmpfs mounts are correctly wired.
+        {
+            WSLAContainerLauncher launcher(
+                "debian:latest",
+                "test-tmpfs",
+                {"/bin/sh", "-c", "mount | grep 'tmpfs on /mnt/wsla-tmpfs1' && mount | grep 'tmpfs on /mnt/wsla-tmpfs2'"});
+
+            launcher.AddTmpfs("/mnt/wsla-tmpfs1", "rw,noexec,nosuid,size=65536k");
+            launcher.AddTmpfs("/mnt/wsla-tmpfs2", "");
```
kvega005, line 2284: You should also test the behavior when you add a tmpfs that already exists. Something like:

            launcher.AddTmpfs("/mnt/wsla-tmpfs", "");
            launcher.AddTmpfs("/mnt/wsla-tmpfs", "");

And verify that CreateContainer() fails

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -906,6 +918,21 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
...
+    if (containerOptions.TmpfsCount > 0)
+    {
+        THROW_HR_IF_NULL_MSG(E_INVALIDARG, containerOptions.Tmpfs, "Tmpfs is null with TmpfsCount=%lu", containerOptions.TmpfsCount);
+
+        for (ULONG i = 0; i < containerOptions.TmpfsCount; i++)
+        {
+            const auto& tmpfs = containerOptions.Tmpfs[i];
+
+            THROW_HR_IF_NULL_MSG(E_INVALIDARG, tmpfs.Destination, "Tmpfs mount at index %lu has null destination", i);
+
```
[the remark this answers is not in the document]
beena352: Yes, HostConfig.Tmpfs is a map, so using operator[] would silently overwrite duplicate destinations.
I’ve added a duplicate check before insertion and return an error if the same destination is specified more than once. This makes duplicate tmpfs entries explicit and easier to diagnose.

test/windows/WSLATests.cpp
beena352: added a test that verifies duplicate tmpfs destinations are rejected with ERROR_ALREADY_EXISTS.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -793,6 +793,18 @@ WslaInspectContainer WSLAContainerImpl::BuildInspectContainer(const DockerInspec
...
+    // Map tmpfs mounts from Docker inspect data.
+    for (const auto& [destination, options] : dockerInspect.HostConfig.Tmpfs)
+    {
+        wsla_schema::InspectMount mountInfo{};
+        mountInfo.Type = "tmpfs";
+        mountInfo.Destination = destination;
+        // Tmpfs mounts are read-write by default. We currently do not parse tmpfs options
+        // (e.g. "ro") for inspect output; Docker enforces actual mount behavior.
+        mountInfo.ReadWrite = true;
+        wslaInspect.Mounts.push_back(std::move(mountInfo));
```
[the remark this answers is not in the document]
beena352: updated the existing reserve() to include both volumes and tmpfs: reserve(m_mountedVolumes.size() + dockerInspect.HostConfig.Tmpfs.size()).

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -906,6 +918,27 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
...
+
+        for (ULONG i = 0; i < containerOptions.TmpfsCount; i++)
+        {
+            const auto& tmpfs = containerOptions.Tmpfs[i];
+
+            THROW_HR_IF_NULL_MSG(E_INVALIDARG, tmpfs.Destination, "Tmpfs mount at index %lu has null destination", i);
+
+            THROW_HR_IF_MSG(
+                HRESULT_FROM_WIN32(ERROR_ALREADY_EXISTS),
+                request.HostConfig.Tmpfs.contains(tmpfs.Destination),
```
OneBlue: Given that we can't easily compare linux paths here, I'd recommend just letting this through in case there are duplicated tmpfs entries. 

The last tmpfs mount seems to override the previous in: 


```
docker run --tmpfs /tmp --tmpfs /tmp -it debian:latest
```

test/windows/WSLATests.cpp
```diff
@@ -2273,6 +2273,31 @@ class WSLATests
...
+
+            auto container = launcher.Launch(*m_defaultSession);
+            auto process = container.GetInitProcess();
+            ValidateProcessOutput(process, {}, 0);
+        }
+
+        // Validate that duplicate tmpfs destinations are rejected.
```
OneBlue: We should add test coverage for: 

- tmpfs paths that are not absolute (like 'foo' for instance). Those should fail with an appropriate error message from docker 

- Invalid tmpfs options (we should also fail with an appropriate error message)

beena352: Added test for non-absolute tmpfs paths and invalid tmpfs options. Both cases now assert failure and validate that the surfaced error message contains the relevant Docker error text.

src/windows/wslasession/WSLAContainer.cpp
beena352: removed the duplicate tmpfs destination check

test/windows/WSLATests.cpp
```diff
@@ -537,6 +537,19 @@ class WSLATests
         }
     }
 
+    void ValidateCOMErrorContains(const std::wstring& Substring)
+    {
+        auto comError = wsl::windows::common::wslutil::GetCOMErrorInfo();
+        VERIFY_IS_TRUE(comError.has_value());
+        VERIFY_IS_NOT_NULL(comError->Message.get());
+
+        if (wcsstr(comError->Message.get(), Substring.c_str()) == nullptr)
```
OneBlue: Unless the error contains things that we can't predict like the container ID, I recommend doing an exact string compare instead of a subset here so we can lock the exact behavior that we want

test/windows/WSLATests.cpp
```diff
@@ -2273,6 +2286,43 @@ class WSLATests
...
+            ValidateProcessOutput(process, {}, 0);
+        }
+
+        // Validate that relative tmpfs paths are rejected by Docker.
+        {
+            WSLAContainerLauncher launcher("debian:latest", "test-tmpfs-relative", {"/bin/cat"});
+            launcher.AddTmpfs("relative-path", "");
+
+            auto [hresult, container] = launcher.LaunchNoThrow(*m_defaultSession);
+            VERIFY_IS_TRUE(FAILED(hresult));
```
OneBlue: We should validate the exact HRESULT here (and below) that we expect. Probably E_INVALIDARG here

beena352: the returned HRESULT is E_FAIL (0x80004005) for these Docker validation failures, so I updated the tests to assert E_FAIL explicitly.

test/windows/WSLATests.cpp
beena352: Makes sense, switched it to exact string comparison.

test/windows/WSLATests.cpp
OneBlue: Hmm that's interesting. We might need to change our error handling logic for CreateContainer() then, but that can be done in a followup change

src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -232,6 +238,8 @@ struct WSLA_CONTAINER_OPTIONS
 
     ULONGLONG ShmSize;
     struct WSLA_CONTAINER_NETWORK ContainerNetwork;
+    [unique, size_is(TmpfsCount)] const struct WSLA_TMPFS_MOUNT* Tmpfs;
+    ULONG TmpfsCount;
 };
```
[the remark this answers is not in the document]
beena352: This is already how the rest of the struct works. Volumes, Ports, and Labels all use the same [unique, size_is(XCount)] pointer + XCount pattern, so Tmpfs/TmpfsCount follows the same approach. The client and service ship together from the same build, so version compatibility isn't an issue here.

OneBlue (approved):
LGTM

Issue #14312, opened 2026-02-27
WSL Settings crash after putting too large number in VM Idle Timeout


krzysztof-barden-hexagon:
### Windows Version

Microsoft Windows [Version 10.0.26100.7781]

### WSL Version

2.6.3.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.87.2-1

### Distro Version

irrelevant

### Other Software

WSLg version: 1.0.71
MSRDC version: 1.2.6353
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release

### Repro Steps

1. open WSL Settings
2. Optional features
3. VM Idle Timeout
4. type very large number (for example 999999999999999999)
5. wsl settings crash

### Expected Behavior

some kind of error or fallback instead of crashing the window/app

### Actual Behavior

crash

krzysztof-barden-hexagon:
attaching logs from event viewer 
[wsl-settings.log](https://github.com/user-attachments/files/25675408/wsl-settings.log)

OneBlue:
Fixed by https://github.com/microsoft/WSL/commit/26bf49db1e7e4a96577bda435ad17b26365a8b8c. This will be included in the next release

PR #14323, opened 2026-02-28
Refactor: Use TryParse in view models for safer numeric input parsing

dhtzs:
Replaces parsing logic with TryParse across view models to prevent overflow/format crashes from invalid user input. Specifically fixes the VM Idle Timeout crash reported in #14312, while improving stability for other numeric settings. Also removes an orphaned reference to SystemdSettingsExpander in OptionalFeaturesPage.xaml.cs which caused a Debug.Assert failure and crashed the application when the Optional Features page was opened. This expander is no longer present in the XAML markup.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/14312

### DIFF
--- a/src/windows/wslsettings/ViewModels/Settings/FileSystemViewModel.cs
+++ b/src/windows/wslsettings/ViewModels/Settings/FileSystemViewModel.cs
@@ -38,7 +38,14 @@ public partial class FileSystemViewModel : WslConfigSettingViewModel
         {
             if (ValidateInput(value, Constants.WholeNumberRegex))
             {
-                Set(ref _defaultVHDSize!, Convert.ToUInt64(value));
+                if (UInt64.TryParse(value, out ulong parsedValue))
+                {
+                    Set(ref _defaultVHDSize!, parsedValue);
+                }
+                else
+                {
+                    OnPropertyChanged();
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


src/windows/wslsettings/ViewModels/Settings/FileSystemViewModel.cs
```diff
@@ -38,7 +38,14 @@ public string DefaultVHDSize
...
             if (ValidateInput(value, Constants.WholeNumberRegex))
             {
-                Set(ref _defaultVHDSize!, Convert.ToUInt64(value));
+                if (UInt64.TryParse(value, out ulong parsedValue))
+                {
+                    Set(ref _defaultVHDSize!, parsedValue);
+                }
+                else
+                {
+                    OnPropertyChanged();
```
benhillis, line 47: Out of curiosity, why do we need to call OnPropertyChanged if parsing failed but not if it succeeded?

dhtzs: When parsing succeeds, the base `Set()` method automatically handles calling `OnPropertyChanged` to update the UI. When parsing fails, we skip calling `Set()`. We have to call `OnPropertyChanged()` manually to force the text box to discard the invalid input and revert to the last good value.

benhillis (approved):

PR #14326, opened 2026-03-01
Fix data race: acquire m_containersLock in CreateContainer

benhillis:
## PR Checklist

### DIFF
--- a/src/windows/wslasession/WSLASession.cpp
+++ b/src/windows/wslasession/WSLASession.cpp
@@ -805,6 +805,7 @@ try
 
     try
     {
+        std::lock_guard containersLock{m_containersLock};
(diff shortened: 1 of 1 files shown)


src/windows/wslasession/WSLASession.cpp
```diff
@@ -805,6 +805,7 @@ try
 
     try
     {
+        std::lock_guard containersLock{m_containersLock};
         auto& it = m_containers.emplace_back(WSLAContainerImpl::Create(
```
OneBlue, line 809: Great catch ! 

We should probably first call `WSLAContainerImpl::Create()` outside the lock and only acquire it to move the created container to `m_containers`, since Create() can take a while

OneBlue (approved):

PR #14327, opened 2026-03-01
init: fix variable shadowing, uninitialized bool, symlink check, open mode, and signal mask

benhillis:
- Fix variable shadowing: inner 'Stop' declaration in main.cpp shadowed the outer variable, preventing idle detection from working correctly.
- Fix uninitialized bool: ReclaimIdling was used before being assigned.
- Fix CreateCaptureCrashSymlink: use THROW_LAST_ERROR_IF/CATCH_RETURN_ERRNO to match idiomatic error handling in the codebase.
- Fix missing sigprocmask call: child process after fork did not reset the signal mask, inheriting the parent's blocked signals.
- Remove WSLA_OPEN message type and handler: dead code with no callers.

### DIFF
--- a/src/linux/init/WSLAInit.cpp
+++ b/src/linux/init/WSLAInit.cpp
@@ -64,15 +64,13 @@ struct WSLAState
 static WSLAState g_state;
 
 int CreateCaptureCrashSymlink()
+try
 {
-    if (symlink("/wsl-init", "/" LX_INIT_WSL_CAPTURE_CRASH) < 0)
-    {
-        LOG_ERROR("symlink({}, {}) failed {}", "/wsl-init", "/" LX_INIT_WSL_CAPTURE_CRASH, errno);
-        return errno;
-    }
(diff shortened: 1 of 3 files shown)


src/linux/init/WSLAInit.cpp
```diff
@@ -503,6 +503,7 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLA_FORK& Mes
             {
                 sigset_t SignalMask;
                 sigemptyset(&SignalMask);
+                sigprocmask(SIG_SETMASK, &SignalMask, nullptr);
```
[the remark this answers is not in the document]
benhillis: I'll need to think about this one.

src/linux/init/WSLAInit.cpp
benhillis: @OneBlue - I'm not sure this method is ever called.

OneBlue: Indeed we can probably get rid of it, it's not used anymore since we don't support opening files for process fds now

benhillis: done.

OneBlue (approved):

PR #14329, opened 2026-03-02
Initial inspect image

yao-msft:
## PR Checklist

### DIFF
--- a/src/windows/inc/docker_schema.h
+++ b/src/windows/inc/docker_schema.h
@@ -188,37 +188,35 @@ struct ImportStatus
 
 struct ImageConfig
 {
-    std::string Hostname;
-    std::string Domainname;
     std::string User;
-    std::map<std::string, EmptyObject> ExposedPorts;
-    std::vector<std::string> Env;
-    std::vector<std::string> Cmd;
-    std::vector<std::string> Entrypoint;
-    std::string Image;
-    std::map<std::string, EmptyObject> Volumes;
+    std::optional<std::map<std::string, EmptyObject>> ExposedPorts;
+    std::optional<std::vector<std::string>> Env;
+    std::optional<std::vector<std::string>> Cmd;
+    std::optional<std::vector<std::string>> Entrypoint;
+    std::optional<std::map<std::string, EmptyObject>> Volumes;
     std::string WorkingDir;
-    std::map<std::string, std::string> Labels;
+    std::optional<std::map<std::string, std::string>> Labels;
+    std::string StopSignal;
+    std::optional<std::vector<std::string>> Shell;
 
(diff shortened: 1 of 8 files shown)


src/windows/inc/docker_schema.h
```diff
@@ -181,6 +189,61 @@ struct ImportStatus
...
+    std::string Comment;
+    std::string Created;
+    std::optional<ImageConfig> Config;
+    std::string Author;
+    std::string Architecture;
+    std::string Variant;
+    std::string Os;
+    std::string OsVersion;
+    uint64_t Size{};
+    std::optional<GraphDriverData> GraphDriver;
```
OneBlue, line 242: I'm not sure that we want to expose this in WSLA yet since that's a docker implementation detail

src/windows/wslasession/WSLASession.cpp
```diff
@@ -779,6 +815,48 @@ try
...
+
+    *Output = nullptr;
+
+    auto lock = m_lock.lock_shared();
+    RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_dockerClient.has_value());
+
+    docker_schema::InspectImage dockerInspect;
+    try
+    {
+        dockerInspect = m_dockerClient->InspectImage(ImageNameOrId);
```
OneBlue, line 969: nit: We might want to add some validation on the image name so that the caller can't 'break' our HTTP request with something like: 

```/another/api/call`\\\ 

We should also check for `WSLA_MAX_IMAGE_NAME_LENGTH`

File: test/windows/WSLATests.cpp [hunk 3]
```diff
@@ -1101,6 +1101,67 @@ class WSLATests
...
+            VERIFY_IS_TRUE(config.Env->size() > 0);
+            VERIFY_IS_FALSE(config.Labels.has_value());
+        }
+
+        // Negative test: Image not found
+        {
+            wil::unique_cotaskmem_ansistring output;
+            VERIFY_ARE_EQUAL(WSLA_E_IMAGE_NOT_FOUND, m_defaultSession->InspectImage("nonexistent:image", &output));
+            ValidateCOMErrorMessage(L"No such image: nonexistent:image");
+        }
```
OneBlue, line 1454: We should add test coverage for invalid image names as well

src/windows/inc/docker_schema.h
yao-msft: correct, the docker_schema is just following the json response from the docker api spec. The ones we want to expose is defined in wsla_schema.

src/windows/wslasession/WSLASession.cpp
yao-msft: Checked for max length. Since we already have url escape, added some negative case tests instead.

OneBlue (approved):

PR #14330, opened 2026-03-02
wsla: wrap SIGKILL fallback in try/catch in Terminate()

benhillis:
Inside the catch(...) handler for the SIGTERM wait, Signal(SIGKILL) and Wait(10s) can both throw. If they do, the exception exits Terminate() via CATCH_RETURN(), skipping VM unmount/reset and leaving m_terminated = false.

Wrap the fallback kill path in its own try/CATCH_LOG() so VM cleanup always proceeds.

### DIFF
--- a/src/windows/wslasession/WSLASession.cpp
+++ b/src/windows/wslasession/WSLASession.cpp
@@ -1003,8 +1003,12 @@ try
         catch (...)
         {
             LOG_CAUGHT_EXCEPTION();
-            m_dockerdProcess->Get().Signal(WSLASignalSIGKILL);
-            exitCode = m_dockerdProcess->Wait(10 * 1000);
+            try
+            {
+                m_dockerdProcess->Get().Signal(WSLASignalSIGKILL);
+                exitCode = m_dockerdProcess->Wait(10 * 1000);
+            }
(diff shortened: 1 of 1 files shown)


src/windows/wslasession/WSLASession.cpp
```diff
@@ -1003,8 +1003,12 @@ try
...
         {
             LOG_CAUGHT_EXCEPTION();
-            m_dockerdProcess->Get().Signal(WSLASignalSIGKILL);
-            exitCode = m_dockerdProcess->Wait(10 * 1000);
+            try
+            {
+                m_dockerdProcess->Get().Signal(WSLASignalSIGKILL);
+                exitCode = m_dockerdProcess->Wait(10 * 1000);
+            }
+            CATCH_LOG();
```
dkbennett, line 1011: If we have an exception here the exitCode may not be set at all. Should set it to something indicating there was an issue with the shutdown.

dkbennett (approved):

src/windows/wslasession/WSLASession.cpp
benhillis: exitCode is set to -1 above.

PR #14332, opened 2026-03-02
wslc: Propagate process exit code correctly

benhillis:
Currently the exit codes of the wslc.exe process are not correctly plumbed, or are returning HRESULTS as process exit codes which is non-standard. I'm open to suggestions if this is not the intended way to fix this, let me know!

ContainerService::Run, ContainerService::Exec, and SessionService::Attach all return an int exit code, but the callers in ContainerTasks and SessionTasks discarded the return value. This meant 'wslc run <image> -- /bin/false && echo success' would print 'success' because CoreMain always returned the HRESULT (S_OK) instead of the process exit code.

Add std::optional<int> ExitCode to CLIExecutionContext. Tasks that launch processes store the exit code there, and CoreMain returns it when set.

### DIFF
--- a/src/windows/wslc/core/Main.cpp
+++ b/src/windows/wslc/core/Main.cpp
@@ -79,7 +79,7 @@ try
         // A command exception means there was an input failure. Display the help
         // along with the error message to help the user correct their input.
         command->OutputHelp(&ce);
-        return E_INVALIDARG;
+        return 1;
     }
     // Any other type of error unrelated to the command parsing.
     catch (...)
@@ -105,11 +105,16 @@ try
         }
     }
 
-    return result;
+    if (context.ExitCode.has_value())
+    {
+        return context.ExitCode.value();
+    }
+
+    return FAILED(result) ? 1 : 0;
 }
 catch (...)
 {
-    return E_UNEXPECTED;
+    return 1;
 }
 } // namespace wsl::windows::wslc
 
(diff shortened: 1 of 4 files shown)


src/windows/wslc/core/Main.cpp
```diff
@@ -79,7 +79,7 @@ try
         // A command exception means there was an input failure. Display the help
         // along with the error message to help the user correct their input.
         command->OutputHelp(&ce);
-        return E_INVALIDARG;
+        return 1;
```
dkbennett, line 82: We should return well known exit codes for the program or at the least a constant.

src/windows/wslc/core/Main.cpp
```diff
@@ -105,11 +105,16 @@ try
         }
     }
 
-    return result;
+    if (context.ExitCode.has_value())
+    {
+        return context.ExitCode.value();
+    }
+
+    return FAILED(result) ? 1 : 0;
```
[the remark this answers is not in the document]
dkbennett: This will happen with E2E tests Amir is cooking up.

dkbennett (approved):
I plan to do more error/exit flow plumbing this week, but this is good start and consistent with the model.

Exit code comment is more of a broader discussion I think about what our return codes should be and how we document them, it's fine for now.

src/windows/wslc/core/Main.cpp
benhillis: Yes I think that would be a good improvement - I'll leave that to you guys.

PR #14333, opened 2026-03-02
Track `bind` syscall when port is 0

damanm24:
With mirrored and virtio networking, we observed an issue when trying to connect to VSCode remote instance inside a WSL environment. The issue is that VSCode will call bind with the localhost IP address with port 0 specified. This indicates to the kernel that it should assign any available ephemeral port to the requesting process.

In WSL's port tracker, we ignore bind syscalls where the port is 0 because the kernel will assign one when the syscall completes. When the port is not 0, we send a PortMappingRequest to the host, which will create a new host-side socket bound to the specified local address and port where incoming connections can be listened to.

To address the issue, we should still track the bind syscall when the port is 0, but follow-up after the syscall has been completed. We do this by duplicating the socket that was created and calling getsockname to check was port has been assigned. Then we can send the same PortMappingRequest to the host.

### DIFF
--- a/src/linux/init/GnsPortTracker.cpp
+++ b/src/linux/init/GnsPortTracker.cpp
@@ -119,9 +119,16 @@ void GnsPortTracker::Run()
 
             if (bindCall->PortZeroBind.has_value())
             {
-                std::lock_guard lock(m_deferredMutex);
-                m_deferredQueue.push_back(std::move(bindCall->PortZeroBind.value()));
-                m_deferredCv.notify_one();
+                try
+                {
+                    std::lock_guard lock(m_deferredMutex);
+                    m_deferredQueue.push_back(std::move(bindCall->PortZeroBind.value()));
+                    m_deferredCv.notify_one();
+                }
+                catch (const std::exception& e)
+                {
+                    GNS_LOG_ERROR("Failed to queue port-0 bind for deferred resolution, {}", e.what());
+                }
             }
         }
 
@@ -521,10 +528,15 @@ wil::unique_fd GnsPortTracker::DuplicateSocketFd(pid_t Pid, int SocketFd)
 }
 
 void GnsPortTracker::TrackPort(PortAllocation allocation)
+try
 {
     // Use insert_or_assign so the deallocation timeout is refreshed if the same
     // port key is already present (emplace would silently keep the old entry).
     m_allocatedPorts.insert_or_assign(std::move(allocation), std::make_optional(time(nullptr) + c_bind_timeout_seconds));
+}
+catch (const std::exception& e)
+{
+    GNS_LOG_ERROR("Failed to track port allocation, {}", e.what());
 }
 
 void GnsPortTracker::RunDeferredResolve()
--- a/src/linux/init/GnsPortTracker.cpp
+++ b/src/linux/init/GnsPortTracker.cpp
@@ -600,6 +600,7 @@ void GnsPortTracker::ResolvePortZeroBind(DeferredPortLookup lookup)
         }
         else
         {
+            GNS_LOG_ERROR("Port-0 bind: unexpected address family ({}) for pid {}", resolvedFamily, lookup.Pid);
             return;
         }
 
--- a/src/linux/init/GnsPortTracker.cpp
+++ b/src/linux/init/GnsPortTracker.cpp
@@ -14,6 +14,14 @@
 #include "GnsPortTracker.h"
 #include "lxinitshared.h"
 
+// pidfd_open (Linux 5.3) and pidfd_getfd (Linux 5.6) may not be defined in older headers.
+#ifndef SYS_pidfd_open
+#define SYS_pidfd_open 434
+#endif
+#ifndef SYS_pidfd_getfd
+#define SYS_pidfd_getfd 438
+#endif
+
 constexpr size_t c_bind_timeout_seconds = 60;
 constexpr auto c_sock_diag_refresh_delay = std::chrono::milliseconds(500);
 constexpr auto c_sock_diag_poll_timeout = std::chrono::milliseconds(10);
@@ -114,6 +122,21 @@ void GnsPortTracker::Run()
             catch (const std::exception& e)
             {
                 GNS_LOG_ERROR("Failed to complete bind request, {}", e.what());
+            }
+
+            // If this was a port-0 bind, the kernel should have assigned an ephemeral port.
+            // This check might be susceptible to races, but it's a best effort to resolve the 
+            // actual port in the case of port-0 binds so that it can be tracked properly.
+            if (bindCall->PortZeroBind.has_value())
+            {
+                try
+                {
+                    ResolvePortZeroBind(bindCall->PortZeroBind.value());
+                }
+                catch (const std::exception& e)
+                {
+                    GNS_LOG_ERROR("Failed to resolve port-0 bind for pid {}, {}", bindCall->PortZeroBind->Pid, e.what());
+                }
             }
         }
 
@@ -300,7 +323,7 @@ std::optional<GnsPortTracker::BindCall> GnsPortTracker::ReadNextRequest()
     catch (const std::exception& e)
     {
         GNS_LOG_ERROR("Fetch to read bind() call info with ID {}lu for pid {}, {}", callInfo.id, callInfo.pid, e.what());
-        return {{{}, callInfo.id}};
+        return {{{}, {}, callInfo.id}};
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/linux/init/GnsPortTracker.cpp
```diff
@@ -442,6 +500,131 @@ int GnsPortTracker::GetSocketProtocol(int pid, int fd)
...
+    // thread after CompleteRequest() unblocks it. Poll getsockname() briefly until
+    // the kernel assigns a port.
+    constexpr int maxRetries = 10;
+    constexpr auto retryDelay = std::chrono::milliseconds(100);
+
+    in_port_t port = 0;
+    in6_addr address = {};
+    int resolvedFamily = 0;
+
+    for (int attempt = 0; attempt < maxRetries; ++attempt)
```
benhillis, line 584: This should probably use our retry helper.

benhillis: I've decided this is fine as-is.

test/windows/NetworkTests.cpp
```diff
@@ -2023,6 +2023,129 @@ class NetworkTests
...
+                wil::unique_socket sock(socket(AF_INET, SOCK_STREAM, IPPROTO_TCP));
+                THROW_LAST_ERROR_IF(!sock);
+
+                SOCKADDR_IN addr{};
+                addr.sin_family = AF_INET;
+                addr.sin_port = htons(assignedPort);
+                THROW_HR_IF(E_FAIL, bind(sock.get(), reinterpret_cast<SOCKADDR*>(&addr), sizeof(addr)) == SOCKET_ERROR);
+            },
+            std::chrono::seconds(1),
+            std::chrono::minutes(2)));
```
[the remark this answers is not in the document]
benhillis: IPV6 is not supported for virtio proxy yet, we can add this variation later.

benhillis (dismissed):

src/linux/init/GnsPortTracker.cpp
```diff
@@ -115,6 +116,13 @@ void GnsPortTracker::Run()
             {
                 GNS_LOG_ERROR("Failed to complete bind request, {}", e.what());
             }
+
+            if (bindCall->PortZeroBind.has_value())
+            {
+                std::lock_guard lock(m_deferredMutex);
```
keith-horton: can you put this in try/catch blocks like the earlier calls? we don't want a failure to add this to our queue to allow a C++ exception to terminate this Run() loop.

src/linux/init/GnsPortTracker.cpp
```diff
@@ -442,6 +499,132 @@ int GnsPortTracker::GetSocketProtocol(int pid, int fd)
...
+    wil::unique_fd dupFd(static_cast<int>(syscall(SYS_pidfd_getfd, pidFd.get(), SocketFd, 0u)));
+    if (!dupFd)
+    {
+        GNS_LOG_INFO("Port-0 bind: pidfd_getfd failed for pid {} fd {} (errno {})", Pid, SocketFd, errno);
+    }
+
+    return dupFd;
+}
+
+void GnsPortTracker::TrackPort(PortAllocation allocation)
```
keith-horton, line 530: please add a function scoped try/catch block around this function. we don't want this failure to cause an exception to be emitted at either of the 2 callsites.

or the 2 callsites need to have try catch blocks around the TrackPort call

src/linux/init/GnsPortTracker.cpp
```diff
@@ -424,7 +481,7 @@ int GnsPortTracker::GetSocketProtocol(int pid, int fd)
 
     if (result < 0)
     {
-        RuntimeErrorWithSourceLocation(std::format("Failed to read protocol for socket: {}, {}", path, errno));
+        throw RuntimeErrorWithSourceLocation(std::format("Failed to read protocol for socket: {}, {}", path, errno));
```
keith-horton, line 491: oh boy, thank you for finding and fixing this!

src/linux/init/GnsPortTracker.cpp
```diff
@@ -442,6 +499,132 @@ int GnsPortTracker::GetSocketProtocol(int pid, int fd)
...
+            std::unique_lock lock(m_deferredMutex);
+            m_deferredCv.wait(lock, [&] { return !m_deferredQueue.empty(); });
+            lookup = std::move(m_deferredQueue.front());
+            m_deferredQueue.pop_front();
+        }
+
+        const auto pid = lookup.Pid;
+        try
+        {
+            ResolvePortZeroBind(std::move(lookup));
```
keith-horton, line 559: it would be preferrable to spin up a std::thread to call ResolvePortZeroBind so that this thread is not blocked, since ResolvePortZeroBind can spin for 2500 ms. which means this could be a backlog of requests for port 0 and we might end up losing some of these requests (some sockets are very short lived).

we might just preallocate a vector<thread> of size 10 or so, so we never need to grow the vector, and if the vector is full because there are N # of port 0 bind requests pended, we can just block this main thread.

if this change needs to go in sooner-than-later, we can have a more scalable solution be addressed later.

keith-horton (changes requested):
just a couple of simple fixes, and this change is good!

@benhillis fyi.

src/linux/init/GnsPortTracker.cpp
benhillis: good call, let's address this as a follow-up, I can probably get to this today.

keith-horton (approved):
Thanks!

src/linux/init/GnsPortTracker.cpp
```diff
@@ -442,6 +506,137 @@ int GnsPortTracker::GetSocketProtocol(int pid, int fd)
...
+    // We cannot use open("/proc/pid/fd/N") for sockets because the symlink target
+    // (socket:[inode]) is not a valid filesystem path. Use pidfd_getfd() instead.
+    wil::unique_fd pidFd(static_cast<int>(syscall(SYS_pidfd_open, Pid, 0u)));
+    if (!pidFd)
+    {
+        GNS_LOG_INFO("Port-0 bind: pidfd_open failed for pid {} (errno {})", Pid, errno);
+        return {};
+    }
+
+    wil::unique_fd dupFd(static_cast<int>(syscall(SYS_pidfd_getfd, pidFd.get(), SocketFd, 0u)));
```
OneBlue, line 521: It looks like we're duplicating a bound socket in our process here. Is there a risk of this creating invalid states ? For instance, if a process does: 

```
socket = bind(..., port = <ephemeral>);

auto boundPort = getPortFromBoundSocket(socket);
close(socket);

reopenedSocket = bind(..., port = boundPort);
```

Is there a risk of us having a copy of `socket` opened after the process closes the original socket ? If so, that could lead to weird behaviors

OneBlue: I guess maybe one way we could solve this would be to make sure that whenever we get a bind from a process, we should try to resolve all the pending ephemeral binds that we have for that process before continuing

src/linux/init/GnsPortTracker.cpp
```diff
@@ -442,6 +506,137 @@ int GnsPortTracker::GetSocketProtocol(int pid, int fd)
...
+        catch (const std::exception& e)
+        {
+            GNS_LOG_ERROR("Failed to resolve port-0 bind for pid {}, {}", pid, e.what());
+        }
+    }
+}
+
+void GnsPortTracker::ResolvePortZeroBind(DeferredPortLookup lookup)
+{
+    // The socket fd was already duplicated (via pidfd_getfd) while the target process
```
OneBlue, line 570: Given that we're retrying for up to 2.5 seconds, there's a risk that we run "behind" if a bunch of sockets are in the queue here. 

Maybe one we could solve this would be by having a list of "in-progress" resolves, and keep track of how many attempts we made to resolve, and just fail them once it goes above 25 (and loop through all pending sockets at once). That would prevent a possible buildup of sockets there

OneBlue: Probably not blocking for this PR, could be done in a followup though

OneBlue: Actually, how about we trace to return of the bind() method via ebpf ? We can have one hook to decide whether to allow / reject the bind, keep track of the pending bind() and examine it when bind() returns

OneBlue (approved):
Approving since the current change is an improvement over the current behavior. Let's look into using a post bind() hook in a followup change. Thank you for doing this !

benhillis:
Agreed with your comments, @OneBlue - I'll look at doing some kind of post bind hook as a follow up, but merging this now to light this up for mirrored / virtioproxy.

PR #14334, opened 2026-03-02
Fix AddShare resource leak: add scope guard for cleanup on failure

benhillis:
If m_shares.emplace throws (e.g., OOM), the Plan9 share or VirtioFS device was already created but not tracked, causing a resource leak. Fix this by inserting into the list first and remove the entry if adding the share fails.

### DIFF
--- a/src/windows/wslaservice/exe/HcsVirtualMachine.cpp
+++ b/src/windows/wslaservice/exe/HcsVirtualMachine.cpp
@@ -524,10 +524,14 @@ try
 
     std::lock_guard lock(m_lock);
 
-    THROW_IF_FAILED(CoCreateGuid(ShareId));
-    auto shareName = wsl::shared::string::GuidToString<wchar_t>(*ShareId, wsl::shared::string::None);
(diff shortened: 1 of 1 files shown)


src/windows/wslaservice/exe/HcsVirtualMachine.cpp
```diff
@@ -543,12 +547,13 @@ try
...
     {
-        deviceInstanceId = m_guestDeviceManager->AddGuestDevice(
+        it->second = m_guestDeviceManager->AddGuestDevice(
             VIRTIO_FS_DEVICE_ID, m_virtioFsClassId, shareName.c_str(), L"", WindowsPath, VIRTIO_FS_FLAGS_TYPE_FILES, m_userToken.get());
     }
 
-    m_shares.emplace(*ShareId, deviceInstanceId);
+    cleanup.release();
 
+    *ShareId = shareIdLocal;
```
benhillis, line 556: also updated the function to only set the ShareId on success.

OneBlue (approved):

PR #14338, opened 2026-03-03
CLI: Update copyright header and add --version option

dkbennett:
Fixes copyright header for WSL and updates the version information. Copyright header uses same text as WSL.exe.

Adds basic version information via --version (-v) option off the root:
`wslc --version`

This is very bare bones version information that is using the WSL version information and inspired by WSL --version. More can be added in the future, or possibly turned into a command if we want more functionality, but this should be good for now to at least know basic version info for troubleshooting.

## PR Checklist

Changed the "Info" argument to "Version" and updated it to output the WSL package version.
Updated the header for help to show the same copyright as WSL.exe.

## Validation Steps Performed
Added test, manual validation.


src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -71,6 +70,7 @@ _(TMPFS,          "tmpfs",               NO_ALIAS,          Kind::Value,       L
 _(TTY,            "tty",                 L"t",              Kind::Flag,        L"Open a TTY with the container process.") \
 _(User,           "user",                L"u",              Kind::Value,       L"User ID for the process (name|uid|uid:gid)") \
 _(Verbose,        "verbose",             L"v",              Kind::Flag,        L"Output verbose details") \
+_(Version,        "version",             L"v",              Kind::Flag,        Localization::WSLCCLI_InfoArgDescription()) \
```
yao-msft: nit: do we need to update the WSLCCLI_InfoArgDescription to be WSLCCLI_VersionArgDescription

src/windows/wslc/core/Command.h
```diff
@@ -29,6 +29,10 @@ using namespace wsl::windows::wslc::execution;
 using namespace wsl::windows::wslc::argument;
 
 namespace wsl::windows::wslc {
+
+constexpr std::wstring_view s_ExecutableName = L"wslc";
+constexpr std::wstring_view s_PackageVersion = L"" WSL_PACKAGE_VERSION;
```
yao-msft: Should we use a version with 0.* instead before we are public preview/GA?

yao-msft (approved):

src/windows/wslc/arguments/ArgumentDefinitions.h
dkbennett: I think it's close enough that I didn't want to give localizers any extra work in case we decide to change it later or move it off of an argument into an actual command like Docker.  (e.g. docker version). Should have probably not put it in the localize strings yet until we finalize the argument descriptions, which is why most of them don't have localized strings yet.

src/windows/wslc/core/Command.h
dkbennett: For now reporting the same version as the WSL package, which is how the CLI builds and will ship. I expect this option (or possibly command) to change a bit and we may have a different CLI version or SDK version from the WSL package version, and we'd report all of them here.  Until we finalize what we want the versions to be this seems like the best version to report at this time. There's a few discussions going on about this. I figure something (the WSL package version) is better than nothing.

localization/strings/en-US/Resources.resw
```diff
@@ -2048,6 +2048,11 @@ Usage:
      <value>Alias arguments with a value must be last in the alias chain: '{}'</value>
      <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
+  <data name="WSLCCLI_CopyrightHeader" xml:space="preserve">
+     <value>Copyright (c) Microsoft Corporation. All rights reserved.
+For privacy information about this product please visit https://aka.ms/privacy.</value>
+     <comment>Copyright notice and privacy link</comment>
+  </data>
   <data name = "MessageWslaInvalidImage" xml:space = "preserve" >
```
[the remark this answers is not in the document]
dkbennett: How I have it is how WSL.exe does it already in this loc file, so I'm inclined to leave it alone. There also is no leading whitespace in the output so I think this is a false positive.

src/windows/wslc/arguments/ArgumentDefinitions.h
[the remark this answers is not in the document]
dkbennett: This isn't a problem. Version only apears on the root and we dont expect it to be anywhere else. If this is ever a problem we'll remove the alias for version most likely.  Edit: copilot has nagged me sufficiently to motivate me to just remove the old string and put in a placeholder for version.

src/windows/wslc/core/Command.h
```diff
@@ -29,6 +29,10 @@ using namespace wsl::windows::wslc::execution;
...
+
```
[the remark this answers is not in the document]
dkbennett: The code as-is passes all checks and the compiler correctly does the widening, so I prefer this stay simple than introduce another macro.

src/windows/wslc/arguments/ArgumentDefinitions.h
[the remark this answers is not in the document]
dkbennett: Ok fine I'll remove it and remove the loc string from the file.

test/windows/wslc/CommandLineTestCases.h
```diff
@@ -21,6 +21,7 @@ Module Name:
 // Root command tests
 COMMAND_LINE_TEST_CASE(L"", L"root", true)
 COMMAND_LINE_TEST_CASE(L"--help", L"root", true)
+COMMAND_LINE_TEST_CASE(L"--version", L"root", true)
```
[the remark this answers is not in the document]
dkbennett: Its probably redundant, but strictly speaking this is a correct test to add so I've added it.

src/windows/wslc/commands/RootCommand.cpp
```diff
@@ -63,6 +59,14 @@ std::wstring RootCommand::LongDescription() const
 
 void RootCommand::ExecuteInternal(CLIExecutionContext& context) const
 {
+    if (context.Args.Contains(ArgType::Version))
+    {
+        std::wostringstream versionOut;
+        versionOut << s_ExecutableName << L" v" << s_PackageVersion;
+        wsl::windows::common::wslutil::PrintMessage(versionOut.str(), stdout);
```
OneBlue: FYI we have custom std::formatters that handle string <-> wstring automatically. So here we could do: 

```
wsl::windows::common::wslutil::PrintMessage(std::format(L"{} v{}", s_ExecutableName, WSL_PACKAGE_VERSION));
```

src/windows/wslc/core/Command.h
OneBlue: Agreed. I'd rather use the same version number that other executables use unless we have a strong reason not to, since doing otherwise will lead to user confusion

OneBlue (approved):

src/windows/wslc/commands/RootCommand.cpp
dkbennett: Nice, updated!

PR #14358, opened 2026-03-04
DO NOT MERGE: test virtioproxy without setting gateway_ip

benhillis:
I am debugging an issue with virtio proxy that I am not able to reproduce locally. As part of my ipv6 change I assumed that I did not need to pass the gateway_ip to the wsldevicehost via the options string. This works locally, but in CI I am getting dns resolution failures.

This will help me validate my theory.


benhillis:
This confirmed my theory, yay!

PR #14362, opened 2026-03-05
Add VHD mounts as named volumes

kvega005:
This PR adds named volume support to WSLA containers and wires it through session, container create/open, metadata persistence, and tests.

## PR Checklist

- Added session-level named volume support:
   - `CreateVolume` and `DeleteVolume` in session API/implementation.
   - Volumes are tracked by name in session state and mounted into the guest at `/mnt/wsla-volumes/{name}`.
   - `DeleteVolume` blocks when any container still references the named volume.
- Added container create support for named volumes:
   - New container option field for named volume mounts (`Name`, `ContainerPath`, `ReadOnly`).
   - Named volume mounts resolve via session volume map and are injected as bind mounts using the guest mount path.
- Added duplicate mount-target validation:
   - Duplicate container target paths are rejected with `ERROR_ALREADY_EXISTS`.
   - Validation applies across both regular bind volumes and named volumes.
   - Reusing the same named volume multiple times is allowed when container target paths are different.
- Added persistence/recovery metadata for named volumes:
   - Named volume mount metadata is serialized with container metadata and reconstructed on Open.
 

## Validation Steps Performed
- Added `NamedVolumesTest`:
 - Covers duplicate named volume creation failure
 - Named volumes mount in guest
 - Same named volume mounted multiple times to different container paths is allowed
 - Duplicate container target path is rejected (named+named and bind+named)
 - In-use delete rejection and successful delete after container removal


src/windows/wslaservice/inc/wslaservice.idl
```diff
@@ -236,6 +236,22 @@ typedef enum _WSLAContainerStartFlags
 
 cpp_quote("DEFINE_ENUM_FLAG_OPERATORS(WSLAContainerStartFlags);")
 
+// Options for creating a named volume.
+struct WSLA_VOLUME_OPTIONS
+{
+    LPCWSTR Name;
+    LPCWSTR Type;
+    [unique] LPCWSTR Options;
+};
```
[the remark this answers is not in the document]
kvega005: We will be implementing volume persistence in follow-up PR.

src/windows/wslasession/WSLAContainerMetadata.h
```diff
@@ -46,14 +46,24 @@ struct WSLAVolumeMount
     NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(WSLAVolumeMount, HostPath, ParentVMPath, ContainerPath, ReadOnly);
 };
 
+struct WSLANamedVolumeMount
+{
+    std::wstring VolumeName;
```
OneBlue: I would recommend using utf8 strings (std::string) over wide strings here, since everything we do inside the guest is utf8 anyway

src/windows/wslasession/WSLAVhdVolume.cpp
```diff
@@ -0,0 +1,208 @@
...
+    m_virtualMachine(VirtualMachine),
+    m_dockerClient(DockerClient)
+{
+}
+
+WSLAVhdVolumeImpl::~WSLAVhdVolumeImpl()
+{
+    try
+    {
+        Delete();
```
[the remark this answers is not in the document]
kvega005: The destructor is called when the service is terminating so at this point, the important part is deleting the docker volume since the VM is shutting down anyway,

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -1036,6 +1084,8 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
         }
     }
 
+    ProcessNamedVolumes(containerOptions, sessionVolumes, containerMountTargets, request);
+
```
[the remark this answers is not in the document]
kvega005: This is coming up in follow up PR.

src/windows/wslasession/WSLASession.cpp
```diff
@@ -1261,17 +1270,68 @@ try
...
+    RETURN_HR_IF_NULL(E_POINTER, Options);
+    RETURN_HR_IF_NULL(E_POINTER, Options->Name);
+    RETURN_HR_IF_NULL(E_POINTER, Options->Type);
+
+    std::string name = Options->Name;
+    std::string type = Options->Type;
+
+    ValidateName(name.c_str());
+
+    auto lock = m_lock.lock_exclusive();
```
OneBlue: I don't recommend acquiring an exclusive lock because this will block until all operations are complete. For instance if there's a SaveImage(), the volume creation will have to wait for it to complete. 

Instead I would recommend doing the same thing that we do with `m_containersLock` so we can have a lock that's dedicated to `m_volumes` (and make sure to always acquire m_lock first)

src/windows/wslasession/WSLAVhdVolume.cpp
```diff
@@ -0,0 +1,208 @@
...
+        }
+        catch (...)
+        {
+            THROW_HR_WITH_USER_ERROR(E_INVALIDARG, Localization::MessageWslaInvalidVolumeOptions(options.Options));
+        }
+
+        const auto it = parsedOptions.find("SizeBytes");
+        if (it == parsedOptions.end())
+        {
+            return c_defaultVolumeSizeBytes;
```
OneBlue: I would recommend requiring the caller to pass in the size and failing if it's absent

src/windows/wslasession/WSLAVhdVolume.cpp
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessageInvalidSize(sizeBytesString.c_str()), sizeBytesString[0] == '-');
+
+        ULONGLONG sizeBytes = 0;
+        size_t parsedCount = 0;
+
+        try
+        {
+            sizeBytes = std::stoull(sizeBytesString, &parsedCount);
```
OneBlue: Given that we're using json, could we directly use an integer here ?

src/windows/wslasession/WSLAVhdVolume.cpp
```diff
@@ -0,0 +1,208 @@
...
+{
+    THROW_HR_IF_NULL(E_POINTER, Options.Name);
+    THROW_HR_IF_NULL(E_POINTER, Options.Type);
+
+    std::string name = Options.Name;
+    std::string type = Options.Type;
+
+    ULONGLONG sizeBytes = ParseSizeBytes(Options);
+
+    const auto hostPath = StoragePath / "volumes" / (name + ".vhdx");
```
OneBlue: I forgot what was the conclusion there. @craigloewen-msft : Do we want to allow users to set custom storage paths for VHD volumes ? (for instance on other physical disks)

src/windows/wslasession/WSLAVhdVolume.cpp
```diff
@@ -0,0 +1,208 @@
...
+    auto virtualMachinePath = std::format("/mnt/wsla-volumes/{}", name);
+    VirtualMachine.Mount(device.c_str(), virtualMachinePath.c_str(), "ext4", "", 0);
+
+    auto mountCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { VirtualMachine.Unmount(virtualMachinePath.c_str()); });
+
+    try
+    {
+        docker_schema::CreateVolume request{};
+        request.Name = name;
+        request.Driver = "local";
```
OneBlue: Should we also set labels so that we can recover the volume type / VHD path when the VM reboots ?

src/windows/wslasession/WSLAVhdVolume.cpp
```diff
@@ -0,0 +1,208 @@
...
+    m_virtualMachine(VirtualMachine),
+    m_dockerClient(DockerClient)
+{
+}
+
+WSLAVhdVolumeImpl::~WSLAVhdVolumeImpl()
+{
+    try
+    {
+        Delete();
```
OneBlue: I don't think that we want to always delete the volume when the class is destroyed, we should only do it if the user explicitly asks for a delete

kvega005: Well, in the follow-up PR, we will change this to just Detach(), but for now I am deleting because otherwise it leaves the docker named volume entry and leaks the VHD since we cannot recover it in a future session.

src/windows/wslasession/WSLAVhdVolume.cpp
kvega005: Yeah, that was coming in a future PR, but I can just add it to this one, since that code is pretty much done anyway.

OneBlue: Ah I see. I'm ok with either then

src/windows/wslasession/WSLAVhdVolume.cpp
kvega005: Do you mean use a JSON schema that is 
```
{
    int SizeInBytes;
    // Future etc.
}
```

Or do you mean deserialize the `Options` string to a map and then deserialize the size value to an int using `FromJson<ULONGLONG>`?

The reason we didn't do the former is that we wanted to keep the options as a generic list of key value pairs. This makes it easier to add options later without having to update the CLI. Its also in keeping with how docker handles `DriverOpts` which is essentially what we are mirroring here.

We certainly can do the latter.

src/windows/wslasession/WSLAVhdVolume.cpp
kvega005: Fair enough.

src/windows/wslasession/WSLAVhdVolume.cpp
kvega005: I think in our design discussion we decided to start with just a default path, then later add the option to specify a path. Regardless, it should be a very small change to add this option, so I am happy to include it here.

OneBlue: Ok in that case I think it would be OK to merge as-is for now and add the option later

craigloewen-msft: Agreed, let's start default for now and can add this later.

This is juts a file operation anyway, and a Windows program has the ability to create a file anywhere it has access so I don't see any issue with adding this feature in the future.

src/windows/wslasession/WSLAVhdVolume.cpp
```diff
@@ -0,0 +1,207 @@
...
+    std::string type = Options.Type;
+
+    auto options = wsl::shared::FromJson<std::map<std::string, std::string>>(Options.Options);
+
+    const auto it = options.find("SizeBytes");
+    THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessageWslaInvalidVolumeOptions(Options.Options), it == options.end());
+
+    auto sizeBytes = wsl::shared::FromJson<ULONGLONG>(it->second.c_str());
+    THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessageInvalidSize(it->second.c_str()), sizeBytes == 0);
+
```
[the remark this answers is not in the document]
kvega005: This is wrong. The "-1" input does return E_INVALIDARG.

src/windows/service/inc/wslaservice.idl
```diff
@@ -546,6 +563,9 @@ interface IWSLASession : IUnknown
         [in] const WSLASessionInitSettings* Settings,
         [in] IWSLAVirtualMachine* Vm);
 
+    // Volume management.
+    HRESULT CreateVolume([in] const WSLAVolumeOptions* Options);
+    HRESULT DeleteVolume([in] LPCSTR Name);
 }
```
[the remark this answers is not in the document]
kvega005: @OneBlue Is this right? Should we create a second IID, or is it fine since we haven't shipped any version of the COM interface yet?

src/windows/wslasession/WSLASession.cpp
```diff
@@ -1473,4 +1531,35 @@ void WSLASession::RecoverExistingContainers()
...
+    }
+    CATCH_AND_THROW_DOCKER_USER_ERROR("Failed to enumerate docker volumes");
+
+    for (const auto& volume : volumes)
+    {
+        try
+        {
+            WI_ASSERT(!m_volumes.contains(volume.Name));
+
+            auto vhdVolume = WSLAVhdVolumeImpl::Open(volume, m_virtualMachine.value(), m_dockerClient.value());
```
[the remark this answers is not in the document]
kvega005: This is not a valid comment. Docker should only be used by WSLA in the WSLA VM.

src/windows/inc/docker_schema.h
```diff
@@ -40,6 +40,36 @@ struct EmptyRequest
...
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(Volume, Name, Driver, Mountpoint, Options, Labels);
+};
+
+struct ListVolumesResponse
+{
+    std::vector<Volume> Volumes;
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(ListVolumesResponse, Volumes);
+};
```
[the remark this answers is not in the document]
kvega005: It doesn't return null. It returns an empty list.

src/windows/wslasession/WSLAVhdVolume.cpp
```diff
@@ -0,0 +1,201 @@
...
+    catch (const DockerHTTPException& e)
+    {
+        THROW_HR_WITH_USER_ERROR_IF(
+            HRESULT_FROM_WIN32(ERROR_SHARING_VIOLATION), Localization::MessageWslaVolumeInUse(m_name.c_str()), e.StatusCode() == 409);
+        THROW_HR_WITH_USER_ERROR_IF(WSLA_E_VOLUME_NOT_FOUND, Localization::MessageWslaVolumeNotFound(m_name.c_str()), e.StatusCode() == 404);
+        THROW_DOCKER_USER_ERROR_MSG(e, "Failed to delete volume '%hs'", m_name.c_str());
+    }
+
+    Detach();
+    LOG_IF_WIN32_BOOL_FALSE(DeleteFileW(m_hostPath.c_str()));
```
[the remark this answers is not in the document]
kvega005: Returning success and removing the volume from the session map ensures the caller doesn't mistakenly continue using a volume that has already been logically deleted (detached from Docker, unmounted, etc.).

src/windows/service/inc/wslaservice.idl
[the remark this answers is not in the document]
OneBlue: Yeah this not a problem until we actually ship the first release

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -1018,6 +1078,9 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
         THROW_HR_IF_NULL_MSG(E_INVALIDARG, volume.HostPath, "Volumes[%lu].HostPath is null", i);
         THROW_HR_IF_NULL_MSG(E_INVALIDARG, volume.ContainerPath, "Volumes[%lu].ContainerPath is null", i);
 
+        const auto [_, inserted] = containerMountTargets.insert(volume.ContainerPath);
```
OneBlue: Do we actually need to validate the uniqueness of the mount targets ? If not I'd be of the opinion to just let those through, and docker can fail if there are duplicated mounts

OneBlue: (That would also give the user a nicer error most likely) -

src/windows/wslasession/WSLASession.cpp
```diff
@@ -1268,17 +1265,72 @@ try
...
+    THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessageWslaInvalidVolumeType(type), type != "vhd");
+
+    auto volume = WSLAVhdVolumeImpl::Create(*Options, m_storageVhdPath.parent_path(), m_virtualMachine.value(), m_dockerClient.value());
+    auto [it, inserted] = m_volumes.insert({name, std::move(volume)});
+    WI_VERIFY(inserted);
+
+    WSL_LOG(
+        "VolumeCreated",
+        TraceLoggingValue(name.c_str(), "VolumeName"),
+        TraceLoggingValue(std::format("/mnt/wsla-volumes/{}", name).c_str(), "VirtualMachinePath"));
```
OneBlue: I'd recommend logging just the volume name here (so we don't have a hardcoded path here that might be out of sync if we ever change that implemented details later)

src/windows/wslasession/WSLAVhdVolume.cpp
```diff
@@ -0,0 +1,201 @@
...
+    THROW_HR_IF_NULL(E_POINTER, Options.Options);
+
+    std::string name = Options.Name;
+
+    auto options = wsl::shared::FromJson<std::map<std::string, std::string>>(Options.Options);
+
+    const auto it = options.find("SizeBytes");
+    THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessageWslaInvalidVolumeOptions(Options.Options), it == options.end());
+
+    auto sizeBytes = wsl::shared::FromJson<ULONGLONG>(it->second.c_str());
```
OneBlue, line 79: Ah I didn't know we could even do that !

OneBlue (approved):
LGTM, minor comments

src/windows/wslasession/WSLASession.cpp
```diff
@@ -1498,4 +1549,35 @@ void WSLASession::RecoverExistingContainers()
...
+        }
+
+        try
+        {
+            WI_ASSERT(!m_volumes.contains(volume.Name));
+
+            auto vhdVolume = WSLAVhdVolumeImpl::Open(volume, m_virtualMachine.value(), m_dockerClient.value());
+            auto [_, inserted] = m_volumes.insert({volume.Name, std::move(vhdVolume)});
+            WI_VERIFY(inserted);
+        }
```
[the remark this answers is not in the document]
kvega005: Lock isnt necessary here as this is only ever called as part of the init function

PR #14364, opened 2026-03-06
Track container state change time and display time since state change in CLI

beena352:
Added state change timestamp tracking to containers. wslc.exe container list now shows how long ago the last state change happened instead of just the state name. Renamed the STATE column to STATUS.

## PR Checklist

- Added m_stateChangedAt to WSLAContainerImpl, set to system_clock::now() in Transition()
- The timestamp is returned by GetStateChangedAt() as Unix epoch seconds and stored as ULONGLONG in the IDL
- The timestamp is passed from ListContainers through ContainerModel to ContainerService::List() where it's used for display
- Updated ContainerStateToString to format relative time
This is the first step , created timestamp will be added in a follow-up PR.

## Validation Steps Performed

- Built and deployed to host
- Ran wslc.exe container list --all and verified STATUS column with relative time
- Stopped and started containers, verified timestamps update correctly
- Ran WSLATests::ContainerState


src/windows/wslasession/WSLAContainer.h
```diff
@@ -132,6 +133,7 @@ class WSLAContainerImpl
     wil::unique_event m_stoppedNotifiedEvent{wil::EventOptions::ManualReset};
     DockerHTTPClient& m_dockerClient;
     WSLA_CONTAINER_STATE m_state = WslaContainerStateInvalid;
+    std::chrono::system_clock::time_point m_stateChangedAt{std::chrono::system_clock::now()};
```
OneBlue: Given that we're only returning the time as a unix timestamp, I would recommend directly store it as a std::uint64_t. That will make things easier when we load an existing container from storage

src/windows/wslc/services/ContainerModel.h
```diff
@@ -61,6 +61,7 @@ struct ContainerInformation
     std::string Name;
     std::string Image;
     WSLA_CONTAINER_STATE State;
+    ULONGLONG StateChangedAt;
```
OneBlue: nit: I recommend default initializing here (`ULONGLONG StateChangeAt{}`)

src/windows/wslasession/WSLAContainer.h
OneBlue: (We can also use std::time() directly here to make things easier)

OneBlue (commented):
This approach looks good. We should also load the latest state change when we look at the container state in OpenContainer(), so we can correctly set it when we restore a container from storage 

Adding @dkbennett for review of the CLI side

src/windows/wslasession/WSLAContainer.h
beena352: That makes sense. I’ll switch stateChangedAt to std::uint64_t and use std::time(nullptr) directly so it’s easier to persist and restore from storage.

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -100,23 +100,54 @@ static void StopInternal(IWSLAContainer& container, WSLASignal signal = WSLASign
...
+
+    if (elapsed < 60)
+    {
+        return std::format(L"{} {} seconds ago", stateString, elapsed);
+    }
+    else if (elapsed < 3600)
+    {
+        return std::format(L"{} {} minutes ago", stateString, elapsed / 60);
+    }
+    else if (elapsed < 86400)
```
dkbennett: The time values are used multiple times, so would prefer constexpr definitions with some easy to read time math, such as using std::chrono for clarity and simplicity.

example: 

```
constexpr LONGLONG SecondsPerMinute= std::chrono::duration_cast<std::chrono::seconds>(1min).count();
constexpr LONGLONG SecondsPerHour = std::chrono::duration_cast<std::chrono::seconds>(1h).count();
...
else if (elapsed < SecondsPerHour)
... elapsed / SecondPerMinute ...
```

Just for readability and ensuring the numbers are correct, especially since they are used multiple times.

beena352: Thanks! I’ll switch those to constexpr values using std::chrono for clarity.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -1131,6 +1163,30 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Open(
...
+        if (!timestamp.empty())
+        {
+            auto parsed = ParseDockerTimestamp(timestamp);
+            if (parsed > 0)
+            {
+                restoredTimestamp = parsed;
+            }
+        }
+    }
+    CATCH_LOG();
```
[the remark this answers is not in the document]
beena352: I think we still need to prefer InspectContainer() here for correctness. metadata.StateChangedAt is only the persisted fallback value and can be stale if the container state changed while the session wasn’t active. Using metadata first would avoid the extra inspect call, but it could restore an older timestamp instead of the latest state change.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -249,6 +249,31 @@ WSLAContainerState DockerStateToWSLAState(ContainerState state)
     }
 }
 
+std::uint64_t ParseDockerTimestamp(const std::string& timestamp)
+{
+    // Docker timestamps are UTC ISO 8601, e.g. "2026-03-05T10:30:00.123456789Z".
+    // Strip fractional seconds and trailing 'Z' since we only need second precision.
+    std::string input(timestamp);
+
+    auto dot = input.find('.');
```
OneBlue: Could we use `%S` instead of manually having to remove the decimals after the second ?

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -249,6 +249,31 @@ WSLAContainerState DockerStateToWSLAState(ContainerState state)
...
+
+    if (!input.empty() && input.back() == 'Z')
+    {
+        input.pop_back();
+    }
+
+    std::chrono::sys_seconds utcSeconds;
+    std::istringstream stream(input);
+    stream >> std::chrono::parse("%FT%T", utcSeconds);
+    THROW_HR_IF(E_INVALIDARG, stream.fail());
```
OneBlue: I recommend also logging the input string for easier debugging: 

```
THROW_HR_IF_MSG(E_INVALIDARG, stream.fail(), "Failed to parse '%hs'", timestamp.c_str());

File: src/windows/wslasession/WSLAContainer.cpp [hunk 7]
```diff
@@ -1131,6 +1163,30 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Open(
...
+    try
+    {
+        auto inspectData = DockerClient.InspectContainer(dockerContainer.Id);
+        auto state = DockerStateToWSLAState(dockerContainer.State);
+        const auto& timestamp = (state == WslaContainerStateRunning) ? inspectData.State.StartedAt : inspectData.State.FinishedAt;
+
+        if (!timestamp.empty())
+        {
+            auto parsed = ParseDockerTimestamp(timestamp);
+            if (parsed > 0)
```
OneBlue: Are there situations where ParseDockerTimestamp() could return 0 here ?

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -1047,6 +1078,7 @@ std::unique_ptr<WSLAContainerImpl> WSLAContainerImpl::Create(
     metadata.InitProcessFlags = containerOptions.InitProcessOptions.Flags;
     metadata.Volumes = volumes;
     metadata.Ports = ports;
+    metadata.StateChangedAt = static_cast<std::uint64_t>(std::time(nullptr));
```
OneBlue: Since we're using Inspect(), I recommend removing this since this could be misleading (this will always represent the creation time)

src/windows/wslasession/WSLAContainer.cpp
beena352: No, Since ParseDockerTimestamp() already throws on parse failure, I’ll remove the parsed > 0 check and use the parsed value directly.

OneBlue (approved):

PR #14372, opened 2026-03-06
Only run the DNS tunneling tests on Windows 11

OneBlue:
## PR Checklist

This test disables the DNS tunnelling tests on Windows < 11, since that feature isn't supported before Windows 11 

### DIFF
--- a/test/windows/WSLATests.cpp
+++ b/test/windows/WSLATests.cpp
@@ -1831,6 +1831,7 @@ class WSLATests
 
     TEST_METHOD(NATNetworkingWithDnsTunneling)
     {
+        WINDOWS_11_TEST_ONLY();
         ValidateNetworking(WSLANetworkingModeNAT, true);
     }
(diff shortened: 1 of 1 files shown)


test/windows/WSLATests.cpp
```diff
@@ -1831,6 +1831,7 @@ class WSLATests
 
     TEST_METHOD(NATNetworkingWithDnsTunneling)
     {
```
benhillis, line 1833: This is probably fine, but less exact than the check in NetworkTests.cpp.

I suspect we'll be removing this feature flag soon anyway, so this is fine.

benhillis (approved):

test/windows/WSLATests.cpp
OneBlue: Yeah that's exactly my thinking. For now I just want to unblock the release pipeline

PR #14374, opened 2026-03-06
build: ensure dependency graph of Linux binaries includes shared headers

benhillis:
I have noticed that sometimes when modifying the shared header files that sometimes Linux binaries do not get rebuilt. I noticed this most recently when I removed a message type, the Linux binary was still built with the old value.

This seems to resolve the issue.


CMakeLists.txt
```diff
@@ -383,6 +383,7 @@ set(LINUX_LDFLAGS -target ${LLVM_TARGET}
                   -static)
 
 set(COMMON_LINUX_LINK_LIBRARIES configfile)
+file(GLOB COMMON_LINUX_HEADERS CONFIGURE_DEPENDS "${CMAKE_CURRENT_LIST_DIR}/src/shared/inc/*.h")
```
OneBlue, line 386: nit: I would have a preference for adding these headers to the `headers` argument of `add_linux_executable` when we call `add_linux_executable` in src\linux\init\CMakeLists.txt, that way it's explicitly added when we create the target

OneBlue (dismissed):

OneBlue (approved):

PR #14365, opened 2026-03-06
Add more WSLC E2E tests

AmelBawa-msft:
### 😵 WSLCE2EContainerKillTests.cpp‎
- WSLCE2E_Container_Kill_HelpCommand
- WSLCE2E_Container_Kill_KillsRunningContainer
- WSLCE2E_Container_Kill_ByName
- WSLCE2E_Container_Kill_NotFound
- WSLCE2E_Container_Kill_InvalidSignal
- WSLCE2E_Container_Kill_TargetedContainerOnly

### 📋 WSLCE2EContainerListTests.cpp‎
-  WSLCE2E_Container_List_HelpCommand
- WSLCE2E_Container_List_AllOption
- WSLCE2E_Container_List_NoOptions_RunningContainers
- WSLCE2E_Container_List_NoOptions_ExcludesCreatedContainers
- WSLCE2E_Container_List_QuietOption_OutputsIdsOnly
- WSLCE2E_Container_List_InvalidFormatOption
- WSLCE2E_Container_List_JsonFormat

### 🗑️ WSLCE2EContainerRemoveTests.cpp‎
- WSLCE2E_Container_Remove_HelpCommand
- WSLCE2E_Container_Remove_NotFound
- WSLCE2E_Container_Remove_Valid
- WSLCE2E_Container_Remove_ById_Valid
- WSLCE2E_Container_Remove_Force_RunningContainer
- WSLCE2E_Container_Remove_Multiple_Valid

### ⏹️ WSLCE2EContainerStopTests.cpp‎
- WSLCE2E_Container_Stop_HelpCommand
- WSLCE2E_Container_Stop_InvalidSignal
- WSLCE2E_Container_Stop_KillsRunningContainer
- WSLCE2E_Container_Stop_ByName
- WSLCE2E_Container_Stop_NotFound
- WSLCE2E_Container_Stop_TargetedContainerOnly
- WSLCE2E_Container_Stop_SignalByName
- WSLCE2E_Container_Stop_InvalidSignalName
- WSLCE2E_Container_Stop_InvalidTimeout
- WSLCE2E_Container_Stop_ValidTimeoutNegativeOne

### 🖼️ WSLCE2EImageListTests.cpp‎
- WSLCE2E_Image_List_HelpCommand
- WSLCE2E_Image_List_DisplayLoadedImage
- WSLCE2E_Image_List_QuietOption_OutputsNamesOnly
- WSLCE2E_Image_List_InvalidFormatOption
[...]

PR #14389, opened 2026-03-10
CLI: Add Interactive test class and E2E tests for interactive scenarios

dkbennett:
Adds Interactive test class for verifying interactive code paths and scenarios for the CLI.

Adds tests for core interactive scenarios with TTY and Non-TTY:
* container exec -i
* container run -i
* container attach
* container start -a
* session shell

Also fixed a bug in Start -a that was not passing the attach flag as it should.

## PR Checklist

* Fixed the Container Start code not passing the attach flag.
* Fixed test cleanup that was not force deleting a container (added --force)
* Created new test class for validating interactive sessions.
* Created tests for all the commands listed in the summary, TTY and non-TTY versions, except for session shell, which is only TTY.

There is a test issue with non-tty stdout not being relayed correctly, so we do not verify non-tty stdout and instead just verify stdin is working and that closing stdin closes the cat process as expected. This is an issue that does not manifest in console and appears to be related to Non-TTY handling of redirected output and will be followed up on separately. Comments were added to the relevant tests explaining this issue and why output is not verified on those specific tests.

TTY tests all verify output is as we expect with echo and whoami commands verifying output matches expected input.

All tests were added to the existing ContainerCreate E2E tests in an effort to try not to bloat too many test files. It is expected these may get refactored later.

## Validation Steps Performed
* Manual tests for the start change along with these tests.
* Ran the tests many times to identify potential deadlocks and ensure idempotent execution.
* Specific care was taken to inspect and guard against deadlocks in the stdin and stdout pipes, and verify none were observed.


dkbennett:
@copilot review please

dkbennett:
> @dkbennett I've opened a new pull request, #14390, to work on those changes. Once the pull request is ready, I'll request review from you.

No, dont do that. I want you to review this pull request.

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -93,6 +94,61 @@ WSLCExecutionResult RunWslc(const std::wstring& commandLine)
...
+    wil::unique_handle processHandle = process->Start();
+
+    // Close child's ends immediately to prevent deadlocks
+    stdinRead.reset();
+    stdoutWrite.reset();
+    stderrWrite.reset();
+
+    // Verify process started
+    DWORD exitCode = 0;
+    if (GetExitCodeProcess(processHandle.get(), &exitCode) && exitCode != STILL_ACTIVE)
```
OneBlue: This creates a race condition: Depending on the timing of how quickly the child exits, we'll behave differently. 

Overall I don't recommend looking at the process exit code until all the pipes have been read until EOF

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -93,6 +94,61 @@ WSLCExecutionResult RunWslc(const std::wstring& commandLine)
...
+    {
+        // Process exited immediately - read any error output
+        std::string errorOutput;
+        char buffer[4096];
+        DWORD available = 0;
+        if (PeekNamedPipe(stderrRead.get(), nullptr, 0, nullptr, &available, nullptr) && available > 0)
```
OneBlue: Doing this will essentially burn CPU and significantly slow down the tests. 

Instead of trying to use synchronous IO, I would recommend: 

- Using wsl::windows::common::wslutil::OpenAnonymousPipe() to create anonymous pipes all pipes (unlike regular pipes, these support overlapped IO)

- Use a MultiHandleWait to consume IO's from all pipes at once. Here's [a good example](https://github.com/microsoft/WSL/blob/3f7e4d5722c162a50553a0b25c9613e8affa01ce/test/windows/WSLATests.cpp#L4056-L4069) of how to do that

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -194,11 +194,12 @@ CreateContainerResult ContainerService::Create(Session& session, const std::stri
     return {.Id = id};
 }
 
-void ContainerService::Start(Session& session, const std::string& id)
+void ContainerService::Start(Session& session, const std::string& id, bool withAttach)
```
AmelBawa-msft: Nit: Consider using `attach` instead to align with other boolean flags we pass (e.g. `force`, `follow`, etc..)

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,228 @@ class WSLCE2EContainerCreateTests
              .ExitCode = 1});
     }
 
+    TEST_METHOD(WSLCCLI_E2E_Container_RunInteractive_TTY)
```
AmelBawa-msft: Consider using the same prefix as other E2E tests "WSLCE2E_"

AmelBawa-msft: Same comment for all added tests ✨

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,228 @@ class WSLCE2EContainerCreateTests
...
+    {
+        WSL2_TEST_ONLY();
+        EnsureContainerDoesNotExist(WslcContainerName);
```
AmelBawa-msft: `EnsureContainerDoesNotExist` is called in the `TestSetup`, so it is redundant here. You can call `VerifyContainerIsNotListed` instead for sanity check.

AmelBawa-msft: Same comment for all added tests ✨

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,228 @@ class WSLCE2EContainerCreateTests
...
+        // Stdin relay is confirmed working. Stdout verification is skipped due to a known
+        // limitation where we are not getting stdout data correctly from non-TTY proccess.
+        // Calling session.ReadUntil() or WriteAndVerifyOutput(session, "test", "test")
+        // fails due to not receiving any output in the pipe.
+
+        // Close stdin to signal EOF to cat
+        session.CloseStdin();
+
+        // Wait for cat to exit with code 0
+        auto exitCode = session.Wait(10000);
```
AmelBawa-msft, line 434: Does this generally take that long to exit? I wonder if this will cause the test to fail randomly because of the timeout.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,228 @@ class WSLCE2EContainerCreateTests
...
+        EnsureContainerDoesNotExist(WslcContainerName);
+
+        auto session = RunWslcInteractive(std::format(L"container run -i --name {} {} cat", WslcContainerName, DebianImage.NameAndTag()));
+        VERIFY_IS_TRUE(session.IsRunning(), L"Container session should be running");
+
+        // Write test data to stdin
+        session.WriteLine("test line 1");
+        session.WriteLine("test line 2");
+
+        // Stdin relay is confirmed working. Stdout verification is skipped due to a known
```
AmelBawa-msft, line 426: 💡 To ensure stdin works reliably, we could write the input to a file first and then run `wslc container exec test-container cat file.txt` to verify that it was passed through correctly

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
```diff
@@ -82,7 +82,7 @@ void EnsureContainerDoesNotExist(const std::wstring& containerName)
     {
         if (line.find(containerName) != std::wstring::npos)
         {
-            auto result = RunWslc(std::format(L"container delete {}", containerName));
+            auto result = RunWslc(std::format(L"container delete --force {}", containerName));
```
AmelBawa-msft: I could be wrong, but I recall `--force` does not delete containers in some tests (I forgot which ones). So, instead in my [other PR](https://github.com/microsoft/WSL/pull/14365/changes), I updated this method to be:

```cpp
 if (line.find(L"running") != std::wstring::npos)
 {
     auto result = RunWslc(std::format(L"container kill {}", containerName));
     result.Verify({.Stdout = L"", .Stderr = L"", .ExitCode = S_OK});
}
```

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -93,6 +94,61 @@ WSLCExecutionResult RunWslc(const std::wstring& commandLine)
     return {.CommandLine = commandLine, .Stdout = output.Stdout, .Stderr = output.Stderr, .ExitCode = output.ExitCode};
 }
 
+WSLCInteractiveSession RunWslcInteractive(const std::wstring& commandLine)
+{
+    auto wslcPath = std::filesystem::path(wslutil::GetMsiPackagePath().value()) / L"wslc.exe";
```
AmelBawa-msft: Nit: Consider moving the wslc.exe path computation to a shared method (e.g. `GetFullWslcCommand`)

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
dkbennett: Yes I think we should change the prefix given WSLC is now ambiguous with the container API, but that can be a separate PR so I'll re-align.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
dkbennett: no when I run it, it's instantaneous, but I wanted to avoid random timing issues if it happens to go longer than expected.

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
dkbennett: We should still try to force delete anyway for test resilience right?

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
dkbennett: Going to leave this alone for now, we verify the write works because it advances and then closes. This is a temporary thing until we get the non-TTY functional with stdout being readable.

test/windows/wslc/e2e/WSLCExecutor.cpp
dkbennett: Fixed it all to use that, was tricky but tests appear reliable with the new model.

test/windows/wslc/e2e/WSLCExecutor.cpp
dkbennett: Fixed by removing it, we dont need it there anyway, we can check IsRunning() in the tests after the drain thread has started and is setup where checking is safe.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,227 @@ class WSLCE2EContainerCreateTests
...
+        // limitation where we are not getting stdout data correctly from non-TTY proccess.
```
OneBlue: What's the limitation here ? I can have a look if it's easy to repro

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -86,13 +87,36 @@ std::wstring WSLCExecutionResult::GetStdoutOneLine() const
...
+
+    auto [childStdinRead, parentStdinWrite] = wsl::windows::common::wslutil::OpenAnonymousPipe(65536, false, true);
+    auto [parentStdoutRead, childStdoutWrite] = wsl::windows::common::wslutil::OpenAnonymousPipe(65536, true, false);
+    auto [parentStderrRead, childStderrWrite] = wsl::windows::common::wslutil::OpenAnonymousPipe(65536, true, false);
+
+    THROW_IF_WIN32_BOOL_FALSE(SetHandleInformation(childStdinRead.get(), HANDLE_FLAG_INHERIT, HANDLE_FLAG_INHERIT));
+    THROW_IF_WIN32_BOOL_FALSE(SetHandleInformation(childStdoutWrite.get(), HANDLE_FLAG_INHERIT, HANDLE_FLAG_INHERIT));
+    THROW_IF_WIN32_BOOL_FALSE(SetHandleInformation(childStderrWrite.get(), HANDLE_FLAG_INHERIT, HANDLE_FLAG_INHERIT));
+
+    auto process = std::make_unique<wsl::windows::common::SubProcess>(nullptr, cmd.c_str());
```
OneBlue: nit: `process` doesn't need to be a pointer here

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -106,4 +130,313 @@ std::wstring GetWslcHeader()
...
+
+    io.AddHandle(std::make_unique<ReadHandle>(HandleWrapper{std::move(m_stderrRead)}, std::move(stderrCallback)));
+
+    ready.set_value();
+
+    while (!m_stopDraining)
+    {
+        try
+        {
+            io.Run(std::chrono::milliseconds(100));
```
OneBlue: Instead of doing a loop, I would recommend having a `wil::unique_event` class field. 

We can then do something like this: 

```
        io.AddHandle(std::make_unique<EventHandle>(m_stopEvent.get()), MultiHandleWait::CancelOnCompleted);

```

That way we can just do: 

```
WSLCInteractiveSession::StopDraining()
{
    m_stopEvent.SetEvent()
    if (m_drainThread.joinable())
    {
        m_drainThread.join();
    }

```

(This will also have the benefit of avoiding logging a timeout error every 100ms in the trace)

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -106,4 +130,313 @@ std::wstring GetWslcHeader()
...
+        m_drainThread.join();
+    }
+}
+
+std::string WSLCInteractiveSession::ReadStdout(DWORD timeoutMs)
+{
+    auto startTime = GetTickCount64();
+
+    // Only use background buffer - no race conditions
+    while (GetTickCount64() - startTime < timeoutMs)
```
OneBlue: This design can lead to races because it will return as soon as anything has been read, but there's no guarantee that the entire pending output out of the process's stdout has actually been flushed. 

If the objective is to validate the content of a handle that won't receive an EOF, I recommend using  `PartialHandleRead`  and passing in the output that we're expecting to see

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -106,4 +130,313 @@ std::wstring GetWslcHeader()
...
+}
+
+void WSLCInteractiveSession::Write(const std::string& data)
+{
+    OVERLAPPED overlapped{};
+    wil::unique_event event(wil::EventOptions::ManualReset);
+    overlapped.hEvent = event.get();
+
+    DWORD written = 0;
+    if (!WriteFile(m_stdinWrite.get(), data.c_str(), static_cast<DWORD>(data.size()), &written, &overlapped))
```
OneBlue, line 190: nit: We could simplify this by doing a regular (synchronous) write

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -106,4 +130,313 @@ std::wstring GetWslcHeader()
...
+{
+    auto ensureCleanup = wil::scope_exit([this] { StopDraining(); });
+
+    auto result = WaitForSingleObject(m_processHandle.get(), timeoutMs);
+    if (result == WAIT_TIMEOUT)
+    {
+        DWORD processId = GetProcessId(m_processHandle.get());
+
+        // Collect diagnostics
+        std::string remainingStdout, remainingStderr;
```
OneBlue: I recommend killing the process first, and then reading its entire output. This should make the diagnostic more complete and won't require polling

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -106,4 +130,313 @@ std::wstring GetWslcHeader()
...
+}
+
+bool WSLCInteractiveSession::Terminate(UINT exitCode)
+{
+    return TerminateProcess(m_processHandle.get(), exitCode) != FALSE;
+}
+
+void WSLCInteractiveSession::VerifyNoErrors()
+{
+    auto errors = ReadStderr(500);
```
OneBlue: This 0.5 seconds timeout creates a race condition depending on how quickly the child process writes on stdout

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -226,7 +226,7 @@ void StartContainer(CLIExecutionContext& context)
     WI_ASSERT(context.Data.Contains(Data::Session));
     WI_ASSERT(context.Args.Contains(ArgType::ContainerId));
     const auto& id = WideToMultiByte(context.Args.Get<ArgType::ContainerId>());
-    ContainerService::Start(context.Data.Get<Data::Session>(), id);
+    ContainerService::Start(context.Data.Get<Data::Session>(), id, context.Args.Contains(ArgType::Attach));
```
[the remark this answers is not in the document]
dkbennett: This is interesting feedback but it does not appear to affect functionality of the code, and I'm hesitant to change it without careful consideration. I'm going to take this as a bug and investigate/fix separately as needed.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -127,6 +120,217 @@ class WSLCE2EContainerCreateTests
         VerifyContainerIsNotListed(WslcContainerName);
```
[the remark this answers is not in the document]
dkbennett: Ty copilot, this was a bad merge resolution.

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -194,11 +194,12 @@ CreateContainerResult ContainerService::Create(Session& session, const std::stri
     return {.Id = id};
 }
 
-void ContainerService::Start(Session& session, const std::string& id)
+void ContainerService::Start(Session& session, const std::string& id, bool attach)
 {
     wil::com_ptr<IWSLAContainer> container;
     THROW_IF_FAILED(session.Get()->OpenContainer(id.c_str(), &container));
-    THROW_IF_FAILED(container->Start(WSLAContainerStartFlags::WSLAContainerStartFlagsNone, nullptr)); // TODO: Error message, detach keys
+    WSLAContainerStartFlags flags = attach ? WSLAContainerStartFlagsAttach : WSLAContainerStartFlagsNone;
```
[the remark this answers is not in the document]
dkbennett: This is good feedback, and I will followup separately on this, as it is more involved than the tests are aiming to fulfill here and this passes tests.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
dkbennett: I have lost at least a day, maybe two trying to get this functional. I would like to address it separately and leave these tests as-is for now.

The issue is that only in the non-tty tests the stdout isn't being output to the pipe. In a real console it works correctly and stdin works correctly. It repros consistently if yo set this test to be like the tty ones with:
WriteAndVerifyOutput(session, "hello", "hello");

The test fails because the output is not detected, there's nothing at all in the stdout (and stderr) pipes.

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -240,11 +240,12 @@ CreateContainerResult ContainerService::Create(Session& session, const std::stri
...
+    THROW_IF_FAILED(container->Start(flags, nullptr));
```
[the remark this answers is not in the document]
dkbennett: The tests verify that this does in fact work as expected, so leaving it alone.

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -106,4 +105,181 @@ std::wstring GetWslcHeader()
...
+        if (error == ERROR_IO_PENDING)
+        {
+            THROW_LAST_ERROR_IF(WaitForSingleObject(event.get(), 5000) != WAIT_OBJECT_0);
+            THROW_IF_WIN32_BOOL_FALSE(GetOverlappedResult(m_stdinWrite.get(), &overlapped, &written, FALSE));
+        }
+        else
+        {
+            THROW_WIN32(error);
+        }
+    }
```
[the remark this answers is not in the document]
dkbennett: We aren't concerned about partial success, and the buffers we are writing are small in the tests in practice, so this seems like adding unnecessary complexity.

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -240,11 +240,12 @@ CreateContainerResult ContainerService::Create(Session& session, const std::stri
     return {.Id = id};
 }
 
-void ContainerService::Start(Session& session, const std::string& id)
+void ContainerService::Start(Session& session, const std::string& id, bool attach)
 {
     wil::com_ptr<IWSLAContainer> container;
```
OneBlue, line 286: nit: I'd like to keep the TODO for `detachKeys` here since those aren't implemented yet

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -106,4 +106,206 @@ std::wstring GetWslcHeader()
...
+
+WSLCInteractiveSession::WSLCInteractiveSession(
+    std::wstring commandLine, wil::unique_hfile stdinWrite, wil::unique_hfile stdoutRead, wil::unique_hfile stderrRead, wil::unique_handle processHandle) :
+    CommandLine(std::move(commandLine)),
+    m_stdinWrite(std::move(stdinWrite)),
+    m_stdoutRead(std::move(stdoutRead)),
+    m_stderrRead(std::move(stderrRead)),
+    m_processHandle(std::move(processHandle))
+{
+    m_stdoutReader = std::make_unique<PartialHandleRead>(m_stdoutRead.get());
```
OneBlue, line 144: nit: `m_stdoutReader` and `m_stderrReader` don't need to pointers here (they could just be regular fields as long as they're initialized after `m_stdoutRead` / `m_stderrRead`

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -106,4 +106,206 @@ std::wstring GetWslcHeader()
...
+    childStderrWrite.reset();
+
+    return WSLCInteractiveSession(
+        commandLine, std::move(parentStdinWrite), std::move(parentStdoutRead), std::move(parentStderrRead), std::move(processHandle));
+}
+
+// WSLCInteractiveSession implementation
+
+WSLCInteractiveSession::WSLCInteractiveSession(
+    std::wstring commandLine, wil::unique_hfile stdinWrite, wil::unique_hfile stdoutRead, wil::unique_hfile stderrRead, wil::unique_handle processHandle) :
```
OneBlue, line 137: nit: const string for the command line to avoid copies

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -106,4 +106,206 @@ std::wstring GetWslcHeader()
...
+    m_stderrRead(std::move(stderrRead)),
+    m_processHandle(std::move(processHandle))
+{
+    m_stdoutReader = std::make_unique<PartialHandleRead>(m_stdoutRead.get());
+    m_stderrReader = std::make_unique<PartialHandleRead>(m_stderrRead.get());
+}
+
+WSLCInteractiveSession::~WSLCInteractiveSession()
+{
+    // Best-effort cleanup to avoid orphaned wslc process if Exit()/Wait() were not called.
```
OneBlue, line 150: I think it's a good idea to do that. Instead of manually killing it though, I would recommend using a job object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`. 

This will guarantee that the child process is killed when we close the job handle, and will also cover the case where the tests crash. 

You can directly use the job object from this [helper method](https://github.com/microsoft/WSL/blob/e057d89f687781bb2d3b52ff57291419c722e46c/src/windows/common/wslutil.cpp#L353-L354) 

And then just call `AssignProcessToJobObject()` after creating the process

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -106,4 +106,206 @@ std::wstring GetWslcHeader()
...
+{
+    OVERLAPPED overlapped{};
+    wil::unique_event event(wil::EventOptions::ManualReset);
+    overlapped.hEvent = event.get();
+
+    DWORD written = 0;
+    if (!WriteFile(m_stdinWrite.get(), data.c_str(), static_cast<DWORD>(data.size()), &written, &overlapped))
+    {
+        DWORD error = GetLastError();
+        if (error == ERROR_IO_PENDING)
```
OneBlue, line 193: nit: This shouldn't be possible since we're using a synchronous handle on the write side of the pipe, so we can remove it

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -127,6 +128,283 @@ class WSLCE2EContainerCreateTests
...
+
+        auto session = RunWslcInteractive(std::format(L"container run -i --name {} {} cat", WslcContainerName, DebianImage.NameAndTag()));
+        VERIFY_IS_TRUE(session.IsRunning(), L"Container session should be running");
+
+        // Write test data to stdin
+        session.WriteLine("test line 1");
+        session.WriteLine("test line 2");
+
+        // Stdin relay is confirmed working. Stdout verification is skipped due to a known
+        // limitation where we are not getting stdout data correctly from non-TTY process.
```
OneBlue, line 427: This is definitely not expected, let's chat about this and see if we can figure this out, but not blocking for this PR

OneBlue (approved):
LGTM, this approach feels pretty clean and easy to follow, I like it ! 

Minor comments

dkbennett:
Discussed with OneBlue, going to complete this PR and have a followup that uses the fix for the overlapped IO for non-tty, and also address the other comments above.

PR #14394, opened 2026-03-10
Implement most of remaining SDK APIs

JohnMcPMS:
Implements most of the remaining SDK APIs.

## PR Checklist

Implements all but 6 of the remaining SDK APIs.  Makes some changes to the header for alignment:
1. Enum naming / values for consistency with runtime
2. Use of `CONST` macro instead of `const` keyword directly
3. Make `WslcImageInfo` directly freeable by using a char array instead of pointer, reducing complexity for callers. Also removes values that the runtime does not provide to us.


src/windows/common/string.cpp
```diff
@@ -190,6 +190,73 @@ std::wstring wsl::windows::common::string::BytesToHex(const std::vector<BYTE>& b
...
+    return static_cast<BYTE>(strtoul(hex, endPtr, 16));
+}
+
+BYTE ConvertHexByte(const wchar_t* hex, wchar_t** endPtr)
+{
+    return static_cast<BYTE>(wcstoul(hex, endPtr, 16));
+}
+
+template <typename T>
+std::vector<BYTE> HexToBytesT(std::basic_string_view<T> input)
```
OneBlue, line 215: nit: Even though part of this method is just being moved, it would be a good idea to add test coverage for it, since we're changing its implementation

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -96,7 +96,7 @@ STDAPI WslcSessionSettingsSetCpuCount(_In_ WslcSessionSettings* sessionSettings,
 STDAPI WslcSessionSettingsSetMemory(_In_ WslcSessionSettings* sessionSettings, _In_ uint32_t memoryMb);
 STDAPI WslcSessionSettingsSetTimeout(_In_ WslcSessionSettings* sessionSettings, _In_ uint32_t timeoutMS);
 
-STDAPI WslcSessionSettingsSetVHD(_In_ WslcSessionSettings* sessionSettings, _In_ const WslcVhdRequirements* vhdRequirements);
+STDAPI WslcSessionSettingsSetVHD(_In_ WslcSessionSettings* sessionSettings, _In_ CONST WslcVhdRequirements* vhdRequirements);
```
OneBlue: What's the reason for switching `CONST` ?

test/windows/WslcSdkTests.cpp
```diff
@@ -438,6 +600,73 @@ class WslcSdkTests
...
+        VERIFY_SUCCEEDED(WslcContainerStart(container.get(), WSLC_CONTAINER_START_FLAG_NONE));
+
+        // State while running: RUNNING.
+        {
+            WslcContainerState state{};
+            VERIFY_SUCCEEDED(WslcContainerGetState(container.get(), &state));
+            VERIFY_ARE_EQUAL(state, WSLC_CONTAINER_STATE_RUNNING);
+        }
+
+        VERIFY_SUCCEEDED(WslcContainerStop(container.get(), WSLC_SIGNAL_SIGKILL, 10));
```
OneBlue: I recommend setting the timeout to 0 so the test doesn't wait for 10 seconds here

tools/test/test-setup.ps1
```diff
@@ -59,7 +59,8 @@ if ($Package) {
             )
 
         $exitCode = (Start-Process -Wait "msiexec.exe" -ArgumentList $MSIArguments -NoNewWindow -PassThru).ExitCode
-        if ($exitCode -Ne 0)
+        # 1605 means that ProductCode was not present on the system
```
OneBlue, line 62: Hmm I'm surprised that we're hitting this, that would essentially mean that the registry keys were leaked. 

Is there a repro to end up in this state ?

src/windows/WslcSDK/wslcsdk.h
JohnMcPMS: Was a mistake based on my misunderstanding of C keyword support. Reverting.

tools/test/test-setup.ps1
JohnMcPMS: Changed it to fail by default but allow an override in case one is in this state.  I can only guess that I was in this state after the other issue I had where the MSI was uninstallable due to the MSI being missing (in the MSIX that was also gone).

OneBlue: For now I'd recommend moving this to separate PR. I think this is a symptom of the MSI package being in a bad state, and I'm worried that this could lead to confusion where developers might think that the package is correctly updated but it actually failed in background

src/windows/WslcSDK/WslcsdkPrivate.cpp
```diff
@@ -30,6 +30,11 @@ WSLC_CONTAINER_OPTIONS_INTERNAL* GetInternalType(WslcContainerSettings* settings
     return reinterpret_cast<WSLC_CONTAINER_OPTIONS_INTERNAL*>(settings);
 }
 
+const WSLC_CONTAINER_OPTIONS_INTERNAL* GetInternalType(const WslcContainerSettings* settings)
```
OneBlue: nit: Probably for a future change, but we should probably use the same casing for all our type. This was probably based on the old service types, but we moved everything to the same casing in https://github.com/microsoft/WSL/pull/14366 

We should probably do the same in the SDK

test/windows/WslcSdkTests.cpp
```diff
@@ -384,6 +394,158 @@ class WslcSdkTests
...
+                CreateFileW(pathToSelf.c_str(), GENERIC_READ, FILE_SHARE_READ, nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr)};
+            VERIFY_IS_FALSE(INVALID_HANDLE_VALUE == selfFileHandle.get());
+
+            LARGE_INTEGER fileSize{};
+            VERIFY_IS_TRUE(GetFileSizeEx(selfFileHandle.get(), &fileSize));
+
+            WslcLoadImageOptions opts{};
+            opts.ImageHandle = selfFileHandle.get();
+            opts.ContentLength = static_cast<uint64_t>(fileSize.QuadPart);
+            VERIFY_FAILED(WslcSessionImageLoad(m_defaultSession, &opts));
```
OneBlue: nit: I recommend validating the exact HRESULT that we expect (probably E_FAIL here) to make sure that we lock down the exact behavior that we want

test/windows/WslcSdkTests.cpp
```diff
@@ -871,6 +1000,113 @@ class WslcSdkTests
...
+            WslcProcessSettings execProcSettings;
+            VERIFY_SUCCEEDED(WslcProcessInitSettings(&execProcSettings));
+            const char* execArgv[] = {"/bin/echo", "exec-hello"};
+            VERIFY_SUCCEEDED(WslcProcessSettingsSetCmdLineArgs(&execProcSettings, execArgv, ARRAYSIZE(execArgv)));
+
+            UniqueProcess execProcess;
+            VERIFY_SUCCEEDED(WslcContainerExec(container.get(), &execProcSettings, &execProcess));
+
+            auto output = WaitForProcessOutput(execProcess.get(), 30s);
+            VERIFY_IS_TRUE(output.stdoutOutput.find("exec-hello") != std::string::npos);
```
OneBlue: nit: I recommend validating the exact output in test cases like this one (including the LF) so we fail if we ever get additional output written to stdout / stderr here

OneBlue: (Same for below test cases)

test/windows/WslcSdkTests.cpp
```diff
@@ -871,6 +1000,113 @@ class WslcSdkTests
...
+        {
+            WslcProcessSettings execProcSettings;
+            VERIFY_SUCCEEDED(WslcProcessInitSettings(&execProcSettings));
+            const char* execArgv[] = {"/bin/echo", "exec-hello"};
+            VERIFY_SUCCEEDED(WslcProcessSettingsSetCmdLineArgs(&execProcSettings, execArgv, ARRAYSIZE(execArgv)));
+
+            UniqueProcess execProcess;
+            VERIFY_SUCCEEDED(WslcContainerExec(container.get(), &execProcSettings, &execProcess));
+
+            auto output = WaitForProcessOutput(execProcess.get(), 30s);
```
OneBlue: I'd recommend increasing the timeouts to up to 2 minutes by default. Our test machines can be pretty slow sometimes (sadly)

OneBlue: (Same for other timeouts below)

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -762,27 +858,47 @@ CATCH_RETURN();
...
-    UNREFERENCED_PARAMETER(exitCode);
-    return E_NOTIMPL;
+    auto internalType = CheckAndGetInternalType(process);
+    RETURN_HR_IF_NULL(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), internalType->process);
+    RETURN_HR_IF_NULL(E_POINTER, exitCode);
+
+    *exitCode = -1;
+
+    WSLAProcessState runtimeState{};
+    RETURN_HR(internalType->process->GetState(&runtimeState, exitCode));
```
OneBlue: We should fail if the state is not `Exited` here, otherwise `exitCode` is undefined. 

I actually wonder if we should return the exit code as part of `WslcProcessGetState` to make this easier for the caller (but we can figure that out in a followup change)

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -879,19 +995,66 @@ CATCH_RETURN();
...
-    UNREFERENCED_PARAMETER(session);
-    UNREFERENCED_PARAMETER(NameOrId);
-    return E_NOTIMPL;
+    auto internalType = CheckAndGetInternalType(session);
+    RETURN_HR_IF_NULL(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), internalType->session);
+    RETURN_HR_IF_NULL(E_POINTER, NameOrId);
+
+    WSLADeleteImageOptions options{};
+    options.Image = NameOrId;
+    // TODO: Flags? (Force and NoPrune)
```
OneBlue, line 1006: Agreed we should probably have flags for that

OneBlue (commented):
Change looks great ! Minor comments

tools/test/test-setup.ps1
JohnMcPMS: Changed it to just have a more detailed error message but no ability to bypass.

OneBlue (approved):

PR #14396, opened 2026-03-10
First step in updating build-stage.yml to generate a CAB file for wsl.msi

1wizkid:
## PR Checklist


1wizkid:
Note, this creates the CAB file but does not sign.

.pipelines/build-stage.yml
```diff
@@ -266,6 +266,14 @@ stages:
...
                       $arch = '${{ platform }}'
                       Copy-Item -Path "bin\$arch\release\wsl.msi" -Destination "$(ob_outputDirectory)\bundle\wsl.$(version.WSL_PACKAGE_VERSION).$arch.msi"
+              
+              - task: PowerShell@2
+                displayName: "Create CAB from ${{ platform }} installer msi"
+                inputs:
+                  targetType: inline
+                  script: |
+                      $arch = '${{ platform }}'
+                      MakeCab "$(ob_outputDirectory)\bundle\wsl.$(version.WSL_PACKAGE_VERSION).$arch.msi" "$(ob_outputDirectory)\bundle\wsl.$(version.WSL_PACKAGE_VERSION).$arch.cab"
```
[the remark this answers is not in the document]
benhillis: I think this is legit.

.pipelines/build-stage.yml
```diff
@@ -266,6 +266,73 @@ stages:
...
+                displayName: "Create CAB from ${{ platform }} installer msi"
+                inputs:
+                  targetType: inline
+                  script: |
+                      $arch = '${{ platform }}'
+                      $bundleDir = "$(ob_outputDirectory)\bundle"
+                      $msiPath = Join-Path $bundleDir "wsl.$(version.WSL_PACKAGE_VERSION).$arch.msi"
+                      $cabPath = Join-Path $bundleDir "wsl.$(version.WSL_PACKAGE_VERSION).$arch.cab"
+
+                      if (-not (Test-Path -Path $msiPath)) {
```
OneBlue, line 280: I don't think we need to manually handle the msi being not found, or the cab not being created. If any of those happen, makecab should return non-zero, which will fail the task

.pipelines/build-stage.yml
```diff
@@ -266,6 +266,73 @@ stages:
                       New-Item -ItemType Directory -Path "$(ob_outputDirectory)\bundle" -Force
                       $arch = '${{ platform }}'
                       Copy-Item -Path "bin\$arch\release\wsl.msi" -Destination "$(ob_outputDirectory)\bundle\wsl.$(version.WSL_PACKAGE_VERSION).$arch.msi"
+              
+              - task: PowerShell@2
```
florelis, line 270: What if instead of having the script here you added a cmake target for creating the cab and then added that to the `targets` template parameter?

That way you wouldn't need to duplicate the signing task, and it would be easier to build the cab locally if we need to.

OneBlue: I think that would work as well. I have a small preference for doing this at the pipeline level (even though it's a bit less elegant) because the CAB has no usage on the dev loop, so I'd rather not to build it every time I build locally

PR #14400, opened 2026-03-10
CLI: Implement --rm argument in create and run

dkbennett:
Implements the --rm option for create and run.
Adds E2E tests to ensure create and run with the --rm option actually removes the container.

## PR Checklist

## Validation Steps Performed
Added E2E tests for this, and manually verified it works as well.


test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,38 @@ class WSLCE2EContainerCreateTests
...
+    {
+        WSL2_TEST_ONLY();
+        VerifyContainerIsNotListed(WslcContainerName);
+
+        // Create the container with a valid image
+        auto result = RunWslc(std::format(L"container create --rm --name {} {} echo hello", WslcContainerName, DebianImage.NameAndTag()));
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+        // Start the container.
+        result = RunWslc(std::format(L"container start {}", WslcContainerName, DebianImage.NameAndTag()));
```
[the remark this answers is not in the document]
OneBlue: Agreed with this

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,38 @@ class WSLCE2EContainerCreateTests
...
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+        // Wait to allow the delete to happen.
+        Sleep(1000);
```
OneBlue: We should not rely on sleeps on the tests (they will break depending on the machine load, especially in CI). 

Instead recommend using `wsl::shared::retry::RetryWithTimeout` and retry for up to a minute for the container to be deleted

OneBlue: (Same for below)

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,38 @@ class WSLCE2EContainerCreateTests
...
+        WSL2_TEST_ONLY();
+        VerifyContainerIsNotListed(WslcContainerName);
+
+        // Run the container with a valid image
+        auto result = RunWslc(std::format(L"container run --rm --name {} {} echo hello", WslcContainerName, DebianImage.NameAndTag()));
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+        // Wait to allow the delete to happen.
+        Sleep(1000);
+        VerifyContainerIsNotListed(WslcContainerName);
```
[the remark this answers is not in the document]
benhillis: agreed, we have a retry helper that would make this straightforward.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,36 @@ class WSLCE2EContainerCreateTests
...
     }
 
+    TEST_METHOD(WSLCE2E_Container_Create_Remove)
+    {
+        WSL2_TEST_ONLY();
+        VerifyContainerIsNotListed(WslcContainerName);
+
+        // Create the container with a valid image
+        auto result = RunWslc(std::format(L"container create --rm --name {} {} echo hello", WslcContainerName, DebianImage.NameAndTag()));
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
```
AmelBawa-msft, line 107:
```suggestion
        result.Verify({.Stderr = L"", .ExitCode = 0});
```

dkbennett: These are equivalent so there is no harm at present and I want to be consistent with the rest of the file for now, if we want to do a pass to change to something else I'm OK with that, but I don't want my tests to be inconsistent with the rest in the file.

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,36 @@ class WSLCE2EContainerCreateTests
...
+    {
+        WSL2_TEST_ONLY();
+        VerifyContainerIsNotListed(WslcContainerName);
+
+        // Run the container with a valid image
+        auto result = RunWslc(std::format(L"container run --rm --name {} {} echo hello", WslcContainerName, DebianImage.NameAndTag()));
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+        // Verify with retry timeout of 1 minute.
+        VerifyContainerIsNotListed(WslcContainerName, std::chrono::seconds(2), std::chrono::minutes(1));
```
OneBlue: nit: The deletion should be completed after `run` returns, so we shouldn't need to retry with a timeout here

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
```diff
@@ -36,23 +37,44 @@ const TestImage& InvalidTestImage()
...
+    {
+        const std::wstring message =
+            L"Container '" + containerNameOrId + L"' found in container list output" +
+            (std::chrono::duration_cast<std::chrono::milliseconds>(timeout).count() > 0 ? L" after timeout" : L" but it should not be listed");
+        VERIFY_FAIL(message.c_str());
     }
 }
 
+// Explicit template instantiations for common duration types
+template void VerifyContainerIsNotListed(const std::wstring&, std::chrono::milliseconds, std::chrono::milliseconds);
```
OneBlue: nit: I'd move the method to a header to avoid having to explicitly instantiate it for each type we need

OneBlue (approved):

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -97,6 +97,36 @@ class WSLCE2EContainerCreateTests
...
+    {
+        WSL2_TEST_ONLY();
+        VerifyContainerIsNotListed(WslcContainerName);
+
+        // Run the container with a valid image
+        auto result = RunWslc(std::format(L"container run --rm --name {} {} echo hello", WslcContainerName, DebianImage.NameAndTag()));
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+        // Run should be deleted on return so no retry.
+        VerifyContainerIsNotListed(WslcContainerName);
```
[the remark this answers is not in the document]
dkbennett: While I generally agree, the comment indicates why there is no retry. If the container should be deleted when run completes then it would be a bug if verify requires a retry.

test/windows/wslc/e2e/WSLCE2EHelpers.h
```diff
@@ -14,6 +14,8 @@ Module Name:
 #pragma once
 
 #include "WSLCExecutor.h"
+#include <docker_schema.h>
```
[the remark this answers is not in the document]
dkbennett: this was from a merge i'm leaving it alone

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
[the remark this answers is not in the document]
dkbennett: Already commented on this Copilot.

PR #14412, opened 2026-03-11
Fix use-after-free: restore shared_from_this() in Dmesg thread

benhillis:
The worker thread captures raw 'this' but can outlive the DmesgCollector object. Restore shared_from_this() to extend the object lifetime.


OneBlue (approved):
Ideally I think that we should avoid runaway threads, but definitely better to fix this for now

PR #14413, opened 2026-03-11
Add bounds check for WaitForMultipleObjects result in port relay

benhillis:
WaitForMultipleObjects can return WAIT_ABANDONED_0+n or other unexpected values. Validate the index is within the ports vector before access.

### DIFF
--- a/src/windows/wslrelay/localhost.cpp
+++ b/src/windows/wslrelay/localhost.cpp
@@ -488,11 +488,15 @@ void AcceptThread(std::vector<std::shared_ptr<PortRelay>>& ports, const GUID& Vm
             break;
         }
 
+        // Validate result is within expected range (accounts for exit event at index 0)
+        const DWORD portIndex = result - 1;
+        THROW_HR_IF(E_UNEXPECTED, portIndex >= static_cast<DWORD>(ports.size()));
+
         // Otherwise complete the accept and start a relay
         try
         {
-            ports[result - 1]->CompleteAccept();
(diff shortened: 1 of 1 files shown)


src/windows/wslrelay/localhost.cpp
```diff
@@ -488,11 +488,15 @@ void AcceptThread(std::vector<std::shared_ptr<PortRelay>>& ports, const GUID& Vm
             break;
         }
 
+        // Validate result is within expected range (accounts for exit event at index 0)
+        const DWORD portIndex = result - 1;
+        THROW_HR_IF(E_UNEXPECTED, portIndex >= static_cast<DWORD>(ports.size()));
```
OneBlue, line 493: That could probably be a WI_ASSERT(), WAIT_ABANDONED is only returned in the case of mutexes in a bad state, which should never happen on this code path

OneBlue (approved):

src/windows/wslrelay/localhost.cpp
```diff
@@ -488,11 +488,15 @@ void AcceptThread(std::vector<std::shared_ptr<PortRelay>>& ports, const GUID& Vm
...
+
```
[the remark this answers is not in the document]
benhillis: thinking about this...

src/windows/wslrelay/localhost.cpp
benhillis: yeah agreed. this one isn't a clear win yet.

PR #14414, opened 2026-03-11
wslc: fix size_t underflow in TablePrinter::PrintRow

benhillis:
If row[i].size() > m_widths[i], the unsigned subtraction wraps to a huge value, causing std::wstring to allocate enormous memory or crash. Clamp the padding to zero when the row value exceeds the column width.


dkbennett (approved):
This is going to be replaced soon but no reason not to fix this now.

PR #14416, opened 2026-03-11
CLI: Add initial support for image tag command

AmelBawa-msft:
- Added support for `wslc image tag`

## PR Checklist


test/windows/wslc/WSLCCLIRepoTagUnitTests.cpp
```diff
@@ -0,0 +1,65 @@
...
+            VERIFY_ARE_EQUAL(expectedTag, result.Tag);
+        }
+    }
+
+    TEST_METHOD(RepoTag_Parse_EdgeCases_ReturnExpectedRepoAndTag)
+    {
+        std::vector<std::tuple<std::string, std::string, std::string>> testCases = {
+            { "", "", "" },
+            { "repo:", "repo", "" },
+            { ":latest", "", "latest" },
```
AmelBawa-msft: @OneBlue, should the CLI be responsible for handling invalid target input, or should the service validate that the target options are present and valid? The [REST API](https://docs.docker.com/reference/api/engine/version/v1.53/#tag/Image/operation/ImageTag) marks the Repo and Tag options as optional, so we might not see the same error behavior as the Docker CLI in WSL

### Docker
```bash
PS C:\> docker images
IMAGE           ID             DISK USAGE   CONTENT SIZE   EXTRA
debian:latest   24fb189462d7        258MB          124MB

PS C:\src\wsl> docker tag debian:latest :mock
error parsing reference: ":mock" is not a valid repository/tag: invalid reference format
```

### WSLC
```bash
PS C:\> wslc images
+---------------+-----------+
| NAME          | SIZE (MB) |
+---------------+-----------+
| debian:latest | 114.30    |
+---------------+-----------+
PS C:\> wslc tag debian:latest :mock
# Exit code: 0
```

OneBlue: @AmelBawa-msft: The service validates all the inputs it receives. If the CLI doesn't need to parse an input, it should let the service do it and handle errors

src/windows/wslc/services/ImageService.cpp
```diff
@@ -199,6 +199,24 @@ void ImageService::Pull(wsl::windows::wslc::models::Session& session, const std:
     THROW_IF_FAILED(session.Get()->PullImage(image.c_str(), nullptr, callback));
 }
 
+void ImageService::Tag(wsl::windows::wslc::models::Session& session, const std::string& sourceImage, const std::string& targetImage)
+{
+    EnumReferenceFormat format;
+    auto parsed = ParseImage(targetImage, &format);
```
OneBlue: nit: We can make this a bit easier to read with: 

```
auto [repo, tag] = ParseImage(targetImage, &format);
```

src/windows/wslc/services/ImageService.cpp
```diff
@@ -199,6 +199,24 @@ void ImageService::Pull(wsl::windows::wslc::models::Session& session, const std:
...
+    if (format == EnumReferenceFormat::Digest)
+    {
+        THROW_HR_WITH_USER_ERROR(E_INVALIDARG, Localization::MessageWslcTagImageInvalidFormat(targetImage.c_str()));
+    }
+
+    WSLCTagImageOptions options{};
+    options.Image = sourceImage.c_str();
+    options.Repo = parsed.first.c_str();
+    const std::string tag = parsed.second ? *parsed.second : "";
+    options.Tag = tag.c_str();
```
OneBlue: nit: We can directly assign do: 

```
options.Tag = parsed.second ? parsed.second->c_str() : "";
```

To avoid the copy here

src/windows/wslc/services/ImageService.cpp
```diff
@@ -199,6 +199,24 @@ void ImageService::Pull(wsl::windows::wslc::models::Session& session, const std:
     THROW_IF_FAILED(session.Get()->PullImage(image.c_str(), nullptr, callback));
 }
 
+void ImageService::Tag(wsl::windows::wslc::models::Session& session, const std::string& sourceImage, const std::string& targetImage)
+{
+    EnumReferenceFormat format;
+    auto parsed = ParseImage(targetImage, &format);
+    if (format == EnumReferenceFormat::Digest)
+    {
+        THROW_HR_WITH_USER_ERROR(E_INVALIDARG, Localization::MessageWslcTagImageInvalidFormat(targetImage.c_str()));
```
OneBlue, line 208: I'm not sure that this is the behavior we want. For instance this is valid with docker: 

```
docker tag debian@sha256:2c91e484d93f0830a7e05a2b9d92a7b102be7cab562198b984a84fdbc7806d91 foo
```
What's the reason for this check ?

AmelBawa-msft: Your example should work in wslc too. The check here is for the target and not the source. We don't allow digest for the target. For example, if I do:
```
docker tag debian:latest debian:example@sha256:2c91e484d93f0830a7e05a2b9d92a7b102be7cab562198b984a84fdbc7806d91
refusing to create a tag with a digest reference
```

Similar in wslc:
```
wslc tag debian:latest debian:example@sha256:2c91e484d93f0830a7e05a2b9d92a7b102be7cab562198b984a84fdbc7806d91
Invalid image tag format: 'debian:example@sha256:2c91e484d93f0830a7e05a2b9d92a7b102be7cab562198b984a84fdbc7806d91'. Expected format is 'name:tag'
Error code: E_INVALIDARG
```

OneBlue: Oh that's right, I misread this. Thank you for clarifying !

OneBlue (approved):

PR #14426, opened 2026-03-13
Refactor: trim unnecessary DLL deps from COMMON_LINK_LIBRARIES

benhillis:
This change limits executables to only link against libraries that they actually need to. This reduces the memory overhead and launch time of each process that was previously loading unneeded dlls. To accomplish this some utilities were moved to separate .cpp files so separate .obj files are created, then the linker can do that right things.

- Split MSI/Wintrust install functions from wslutil.cpp into install.cpp
- Remove MI.lib, wsldeps.lib, msi.lib, Wintrust.lib from COMMON_LINK_LIBRARIES
- Add per-target INSTALL_LINK_LIBRARIES and SERVICE_LINK_LIBRARIES
- Result: wslhost, wslrelay, wslcsdk no longer import msi/wintrust; wslhost also loses MI/wsldeps; wslservice loses msi/wintrust in debug builds


src/windows/common/install.cpp
```diff
@@ -0,0 +1,453 @@
...
+            INSTALLLOGMODE_RESOLVESOURCE | INSTALLLOGMODE_OUTOFDISKSPACE | INSTALLLOGMODE_ACTIONSTART | INSTALLLOGMODE_ACTIONDATA |
+            INSTALLLOGMODE_COMMONDATA | INSTALLLOGMODE_INITIALIZE | INSTALLLOGMODE_TERMINATE | INSTALLLOGMODE_SHOWDIALOG,
+        (void*)&Callback);
+
+    MsiSetInternalUI(INSTALLUILEVEL(INSTALLUILEVEL_NONE | INSTALLUILEVEL_UACONLY | INSTALLUILEVEL_SOURCERESONLY), nullptr);
+}
+
+} // namespace
+
+int wsl::windows::common::wslutil::CallMsiPackage()
```
OneBlue: This looks like a mistake (this method is in the wslutil namespace, but in `install.cpp`)

benhillis: oops, yep I think you're right.

OneBlue (approved):

PR #14430, opened 2026-03-13
SDK verb-noun and error message complete change

JohnMcPMS:
Finish up the change started with #14395 by changing to verb-noun for the SDK function names and adding error strings where appropriate.

## PR Checklist

Swaps the remaining functions to verb-noun and adds error strings to the functions that seem likely to need them (open to more/less if desired).

Improves the error string handling so that we initialize the output pointer to null and can capture the error string when appropriate in a more formal way.

## Validation Steps Performed
Regression tests pass.


src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -137,18 +137,48 @@ void ConvertSHA256Hash(const char* hashString, uint8_t sha256[32])
...
+    }
+
+    operator HRESULT() const
+    {
+        return m_hr;
+    }
+
+private:
+    PWSTR* m_errorMessage = nullptr;
+    HRESULT m_hr = S_OK;
```
OneBlue: nit: it might be a good idea to make `m_hr` an std::optional so we can assert if the caller tries to get the hresult when it's not been written to

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -137,18 +137,48 @@ void ConvertSHA256Hash(const char* hashString, uint8_t sha256[32])
     memcpy(sha256, &hashBytes[0], s_sha256ByteCount);
 }
 
-void GetErrorInfoFromCOM(PWSTR* errorMessage)
+struct ErrorInfoWrapper
```
OneBlue, line 142: This works, but I'd like to recommend a different design (this is the pattern that we use in other places in the codebase). We could do something like this: 

```
class SDKExecutionContext : public wsl::windows::common::ExecutionContext
{
    ClientExecutionContext(PWSTR* errorMessage);

   // Automatically called when a WIL error is thrown (via THROW_HR* for instance) 
    void CollectErrorImpl(HRESULT result) override
    {
        if (m_errorMessage)
        {
            auto errorInfo = wsl::windows::common::wslutil::GetCOMErrorInfo();
            if (errorInfo)
            {
                *m_errorMessage = wil::make_unique_string<wil::unique_cotaskmem_string>(errorInfo->Message.get()).release();
            }
        }
    }

private:
    PWSTR* m_errorMessage{};
}
```

Then a caller could just do: 

```
STDAPI WslcCreateSession([...]_Outptr_opt_result_z_ PWSTR* errorMessage)
try
{
    SDKExecutionContext(errorMessage);

    THROW_IF_FAILED(sessionManager->CreateSession([...])

}
CATCH_RETURN();


```

The main benefit of that approach is that we don't have to remember to call `CaptureResult` or store the hresult inside the helper struct

OneBlue: Approving now to unblock the API change. Feel free to do this in a followup change

OneBlue (approved):
LGTM, one minor suggestion. Feel free to merge as-is (with the copilot-suggested changes) and address in a followup change

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: Added a TODO to make this change when telemetry is integrated.

PR #14438, opened 2026-03-16
Add port mappings to container list output

beena352:
Add port mappings to container list output. When containers are created with `-p`, the PORTS column in `container list` now displays the mapped ports.

## PR Checklist

- Added WSLAContainerPortMapping struct and flat ports array to ListContainers
- Changed WSLAPortMapping.BindingAddress to fixed-size char array 
- Populated port data in `ListContainers` on the service side
- Added `PortInformation` struct and `FormatPorts()` on the CLI side
- Added PORTS column to `container list` table output (only shown for running containers)
- Updated SDK to pass Protocol field

## Validation Steps Performed
Ran `PortMappingsNat` test — verified `ListContainers` returns correct HostPort, ContainerPort, and Family for both AF_INET and AF_INET6 mappings, across Bridged and Host network modes


src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -225,6 +226,29 @@ void SetContainerOptionsFromArgs(CLIExecutionContext& context)
...
+        for (const auto& publishArg : context.Args.GetAll<ArgType::Publish>())
+        {
+            auto portStr = WideToMultiByte(publishArg);
+
+            uint16_t hostPort = 0;
+            uint16_t containerPort = 0;
+            if (auto colon = portStr.find(':'); colon != std::string::npos)
+            {
+                hostPort = static_cast<uint16_t>(std::stoi(portStr.substr(0, colon)));
+                containerPort = static_cast<uint16_t>(std::stoi(portStr.substr(colon + 1)));
```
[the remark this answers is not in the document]
beena352: Good catch, updated to use validation::GetIntegerFromString<uint16_t>() which handles range validation and error messages.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -225,6 +226,29 @@ void SetContainerOptionsFromArgs(CLIExecutionContext& context)
         }
     }
 
+    if (context.Args.Contains(ArgType::Publish))
+    {
```
AmelBawa-msft, line 264: I think some of the code in this PR is duplicated here:
- https://github.com/microsoft/WSL/pull/14225

We may still need the port information in the list command table from this PR 😊

test/windows/WSLATests.cpp
```diff
@@ -3739,6 +3741,34 @@ class WSLATests
             ExpectHttpResponse(L"http://127.0.0.1:1234", 200);
             ExpectHttpResponse(L"http://[::1]:1234", {});
 
+            // Verify that ListContainers returns the port data.
+            {
+                wil::unique_cotaskmem_array_ptr<WSLAContainerEntry> containers;
```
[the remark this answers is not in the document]
beena352: Added scope_exit to free nested Ports arrays.

src/windows/wslc/tasks/ContainerTasks.cpp
beena352: Oh, I wasn’t aware of this PR earlier, so I went ahead and implemented this on my side. Makes sense, I’ll wait for it to  merge and then rebase this PR to remove the duplicate code.

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -323,6 +350,25 @@ std::vector<ContainerInformation> ContainerService::List(Session& session)
         entry.Id = current.Id;
         entry.StateChangedAt = current.StateChangedAt;
         entry.CreatedAt = current.CreatedAt;
+
+        // Extract ports
```
dkbennett: Recommend putting port extraction/processing into a common method, we probably want to use the same for argument validation.

beena352: Good call, I’ll pull out a PortInformationFromWSLAPortMapping() helper and use it in List().

src/windows/wslc/services/ContainerModel.h
```diff
@@ -57,6 +57,17 @@ struct KillContainerOptions
     int Signal = WSLASignalSIGKILL;
 };
 
+struct PortInformation
```
AmelBawa-msft, line 62: We may be able to leverage the existing `PublishPort` struct which seem to have the same information

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -246,6 +257,35 @@ std::wstring ContainerService::ContainerStateToString(WSLAContainerState state,
     return std::format(L"{} {}", stateString, FormatRelativeTime(stateChangedAt));
 }
 
+std::wstring ContainerService::FormatPorts(const std::vector<PortInformation>& ports)
```
AmelBawa-msft: Few observations compared to docker:
- If the container is not in running state, the port mapping is empty
- Docker displays mapping for port ranges: `0.0.0.0:8085-8090->8085-8090/tcp, [::]:8085-8090->8085-8090/tcp`. We can consider a logic that clusters consecutive ports in the CLI, but IIRC the service API may enable port range mapping in the future but not sure if the CLI logic for grouping ports would change (CC: @OneBlue)

beena352: Good points, Will add a state check so ports only display for running containers.
For port ranges, yeah that would be nice, but since the service might support range mappings later, I’d rather not add grouping logic in the CLI right now if it’s going to change. We can handle that in a follow-up.

src/windows/wslc/services/ContainerModel.h
beena352: PublishPort is used on the -p parsing path, while this struct is just for formatting the mappings returned from ListContainers(). Since they’re used in different contexts, I kept a small separate helper here, but happy to revisit if you think we should align them.

src/windows/wslasession/WSLAContainer.h
```diff
@@ -92,6 +92,7 @@ class WSLAContainerImpl
     const std::string& Image() const noexcept;
     const std::string& Name() const noexcept;
     WSLAContainerState State() const noexcept;
+    const std::vector<ContainerPortMapping>& GetPorts() const noexcept;
```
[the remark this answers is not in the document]
beena352: ListContainers only reads fields set at construction (HostPort, BindAddress, Protocol, ContainerPort). MapPorts/UnmapPorts only update VmPort and the VM pointer, and the vector is not modified.
ContainerPortMapping is non-copyable, so returning it by value is not an option

src/windows/wslasession/WSLASession.cpp
```diff
@@ -1161,6 +1161,17 @@ try
     auto output = wil::make_unique_cotaskmem<WSLAContainerEntry[]>(m_containers.size());
 
     size_t index = 0;
+    auto errorCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
+        for (size_t i = 0; i < index; ++i)
+        {
+            for (ULONG j = 0; j < output[i].PortsCount; ++j)
+            {
+                CoTaskMemFree(const_cast<LPSTR>(output[i].Ports[j].BindingAddress));
```
OneBlue: Hmm I wonder if we should switch `BindingAddress` to a fixed size string, (since we know it won't be bigger than an ipv6 address), this would simplify this quite a bit

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -250,6 +260,40 @@ std::wstring ContainerService::ContainerStateToString(WSLAContainerState state,
...
+        return L"";
+    }
+
+    std::wstring result;
+    for (size_t i = 0; i < ports.size(); ++i)
+    {
+        const auto& port = ports[i];
+
+        std::wstring hostIp = wsl::shared::string::MultiByteToWide(port.BindingAddress);
+        if (hostIp.empty())
```
OneBlue: Are there known situations where `hostIp` is empty ?

src/windows/service/inc/wslaservice.idl
```diff
@@ -294,6 +294,8 @@ typedef struct _WSLAContainerEntry
     ULONGLONG StateChangedAt;
     ULONGLONG CreatedAt;
     WSLAContainerState State;
+    [unique, size_is(PortsCount)] WSLAPortMapping* Ports;
```
OneBlue: Unfortunately I think we're getting bitten by how COM memory allocation works here (having to go delete inner fields inside of `WSLAContainerEntry` is pretty heavy, and easy to forget), so how about instead of adding to `WSLAContainerEntry`, we did something like this: 

```

struct WSLAContainerPortMapping
{
    WSLAContainerId Id;
    WSLAPortMapping PortMapping;
}
    HRESULT ListContainers([out, size_is(, *Count)] WSLAContainerEntry** Images, 
                                           [out] ULONG* Count,
                                           [out, size_is(, *PortsCounts)] WSLAContainerPortMapping **Ports,
                                           [out] ULONG* PortsCount);
```

If we turn `WSLAPortMapping::BindingAddress` into a fixed size string, that would give us two trivial arrays, and so we could get rid of all the manual deallocation logic

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1144,13 +1144,15 @@ try
...
-HRESULT WSLASession::ListContainers(WSLAContainerEntry** Containers, ULONG* Count)
+HRESULT WSLASession::ListContainers(WSLAContainerEntry** Containers, ULONG* Count, WSLAContainerPortMapping** Ports, ULONG* PortsCount)
 try
 {
     COMServiceExecutionContext context;
 
     *Count = 0;
     *Containers = nullptr;
+    *Ports = nullptr;
+    *PortsCount = 0;
```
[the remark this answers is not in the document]
beena352: Since ListContainers previously didn’t validate Containers or Count, adding validation just for the new parameters would be inconsistent.

src/windows/service/inc/wslaservice.idl
```diff
@@ -296,6 +298,12 @@ typedef struct _WSLAContainerEntry
     WSLAContainerState State;
 } WSLAContainerEntry;
 
+typedef struct _WSLAContainerPortMapping
+{
+    WSLAContainerId Id;
+    WSLAPortMapping PortMapping;
+} WSLAContainerPortMapping;
```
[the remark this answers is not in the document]
beena352: Updated.

src/windows/common/WSLAContainerLauncher.cpp
```diff
@@ -94,13 +94,15 @@ void WSLAContainerLauncher::AddPort(uint16_t WindowsPort, uint16_t ContainerPort
 {
     THROW_HR_IF(E_INVALIDARG, Family != AF_INET && Family != AF_INET6);
 
-    auto& inserted = m_bindingAddressStorage.emplace_back(BindingAddress.value_or(Family == AF_INET ? "127.0.0.1" : "::1"));
-    m_ports.emplace_back(WSLAPortMapping{
+    const auto& address = BindingAddress.value_or(Family == AF_INET ? "127.0.0.1" : "::1");
```
OneBlue: Here's taking a reference on a string, that will be destroyed right after value_or() returns if `BindingAddress` was empty. 

Now that we just copying the string inside the structure, I recommend doing something like: 

```
if (BindingAddress.empty())
{
    // Copy default based on Family
}
else
{
    // Copy BindingAddress.c_str()
}

File: src/windows/common/WSLAContainerLauncher.cpp [hunk 14]
```diff
@@ -94,13 +94,15 @@ void WSLAContainerLauncher::AddPort(uint16_t WindowsPort, uint16_t ContainerPort
...
+    WSLAPortMapping port{
         .HostPort = WindowsPort,
         .ContainerPort = ContainerPort,
         .Family = Family,
         .Protocol = Protocol,
-        .BindingAddress = inserted.c_str()});
+    };
+    THROW_HR_IF(E_INVALIDARG, strcpy_s(port.BindingAddress, address.c_str()) != 0);
```
OneBlue: nit: I recommend printing the value on failure: 

```
    THROW_HR_IF_MSG(E_INVALIDARG, strcpy_s(port.BindingAddress, address.c_str()) != 0, "Invalid address: %hs", address.c_str());
```

This will make debugger easier if this ever fails

src/windows/wslasession/WSLASession.cpp
```diff
@@ -1169,9 +1180,45 @@ try
         e->GetState(&output[index].State);
         e->GetStateChangedAt(&output[index].StateChangedAt);
         e->GetCreatedAt(&output[index].CreatedAt);
+
+        const auto& ports = e->GetPorts();
```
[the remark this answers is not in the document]
OneBlue: Agreed with this comment.  We should not return a reference to `m_port` here since another thread could modify those while we're reading them. 

Instead of doing this, we could have this method just return a std::vector<WSLAContainerPortMapping>(), that way we just add those to our list directly

test/windows/WSLCTests.cpp
```diff
@@ -4132,6 +4140,45 @@ class WSLATests
...
+
+                VERIFY_ARE_EQUAL(2, containerPorts.size());
+                VERIFY_ARE_EQUAL(1234, containerPorts[0].HostPort);
+                VERIFY_ARE_EQUAL(8000, containerPorts[0].ContainerPort);
+                VERIFY_ARE_EQUAL(AF_INET, containerPorts[0].Family);
+                VERIFY_ARE_EQUAL(1234, containerPorts[1].HostPort);
+                VERIFY_ARE_EQUAL(8000, containerPorts[1].ContainerPort);
+                VERIFY_ARE_EQUAL(AF_INET6, containerPorts[1].Family);
+                VERIFY_ARE_EQUAL(IPPROTO_TCP, containerPorts[0].Protocol);
+                VERIFY_ARE_EQUAL(IPPROTO_TCP, containerPorts[1].Protocol);
```
OneBlue, line 4457: nit: I recommend adding test coverage for containers that have port mappings, that created, but not actually started yet

OneBlue: Same for stopped containers (we should only return ports if the container is actually running)

OneBlue (changes requested):

src/windows/common/WSLAContainerLauncher.cpp
beena352: Thanks for the suggestion! Using has_value() here since BindingAddress is an std::optional<std::string>, and with the if/else we now copy directly into the fixed-size field.

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -332,6 +374,15 @@ std::vector<ContainerInformation> ContainerService::List(Session& session)
...
         entry.StateChangedAt = current.StateChangedAt;
         entry.CreatedAt = current.CreatedAt;
+
+        for (const auto& port : ports)
+        {
+            if (strcmp(port.Id, current.Id) == 0)
+            {
+                entry.Ports.push_back(PortInformationFromWSLAPortMapping(port.PortMapping));
+            }
+        }
```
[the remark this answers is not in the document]
beena352: The number of containers and port mappings is small right now, so a linear scan is fine. We can optimize it later if needed when more features are added.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1182,11 +1185,40 @@ try
         e->GetState(&output[index].State);
         e->GetStateChangedAt(&output[index].StateChangedAt);
         e->GetCreatedAt(&output[index].CreatedAt);
+
+        if (output[index].State != WslcContainerStateRunning)
```
OneBlue: I recommend moving this check inside WSLAContainerImpl::GetPorts(), since the state could change between the state check and the time we call `e->GetPorts()`. 

GetPorts() could simply return an empty array if the container is not running

test/windows/WSLCTests.cpp
```diff
@@ -4206,6 +4214,66 @@ class WSLCTests
...
```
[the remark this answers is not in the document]
beena352: Ports are stored in a vector and returned in insertion order.

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -349,6 +391,15 @@ std::vector<ContainerInformation> ContainerService::List(Session& session)
...
+        for (const auto& port : ports)
+        {
+            if (strcmp(port.Id, current.Id) == 0)
+            {
+                entry.Ports.push_back(PortInformationFromWSLCPortMapping(port.PortMapping));
+            }
+        }
+
         result.emplace_back(std::move(entry));
     }
```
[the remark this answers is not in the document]
OneBlue: That's fine, this list isn't expected to be big

OneBlue (approved):
LGTM, the test failures in the pipeline are unrelated to your change. Feel free to restart the pipeline. One of them should go away once #40062 is merged

PR #14443, opened 2026-03-16
fix(init): prevent binfmt_misc handler loss on distro termination

### LINKED ISSUE #13885: binfmt config cleared on all distro when distro with systemd is terminated
### Windows Version

Microsoft Windows [Version 10.0.26200.7462]

### WSL Version

WSL version: 2.6.2.0

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.87.2-microsoft-standard-WSL2

### Distro Version

Ubuntu 24.04

### Other Software

_No response_

### Repro Steps

1. Install two distros with systemd. For example Ubuntu 24.04 and Ubuntu 22.04
2. Run both of them in separate terminals
3. terminate one of those in a third terminal (`wsl.exe -t "Ubuntu 22.04"`)
4. Observe `ls /proc/sys/fs/binfmt_misc` in the one left open

### Expected Behavior

In the distro that is left open, the registered binfmt handlers aren't cleared

### Actual Behavior

All binfmt handlers in the non-terminated distro are gone. Including `WSLInterop`. As if the binfmt handlers were shared somehow.

### Diagnostic Logs

_No response_

yeelam-gordon:
## Summary

**Problem:** Terminating a WSL2 distro with systemd causes `cmd.exe` and other Windows executables to stop working in all other running distros (#13885).

**Root cause:** When systemd shuts down, it unmounts `/proc/sys/fs/binfmt_misc` and unregisters all binfmt handlers  including WSL's `WSLInterop` handler that enables running Windows executables from Linux. Since binfmt_misc is shared across all distros in the same VM, this breaks interop everywhere.

**Fix:** Three layered defenses to protect the WSL binfmt handler:

1. **Keep binfmt_misc mounted**  Override `proc-sys-fs-binfmt_misc.mount` to prevent systemd from unmounting it during shutdown
2. **Re-register after restart**  Add `ExecStartPost` to `systemd-binfmt.service` that re-registers WSL's handler after the service runs, ensuring priority over conflicting distro configs
3. **Declarative config**  Write `/run/binfmt.d/WSLInterop.conf` so the handler is automatically restored on `systemd-binfmt` restart

Also splits the binfmt registration macro into WSL1 (`:P`) and WSL2 (`:FP`) variants, since WSL1 does not support the `F` (fix-binary) flag.

## PR Checklist

- [x] Fixes #13885
- [x] Unit tests + multi-distro integration test added
- [x] WSL1 compatibility preserved (separate macro, CI-validated)
- [x] Manual validation performed (see below)

## Test Coverage

**Automated tests:**

| Test | What it validates |
|------|-------------------|
| `SystemdBinfmtIsRestored` | Generated config files, handler survives distro restart, `systemctl restart/stop systemd-binfmt`, `daemon-reload`, and `protectBinfmt=false` cleanup |
| `BinfmtSurvivesDistroTermination` | Core bug scenario: two distros running, terminate one, verify interop still works in the other |
| `Interop` | End-to-end binfmt registration on both WSL1 and WSL2 CI pipelines |

**Manual validation (WSL2, systemd enabled):**

| Scenario | Result |
|----------|--------|
| Interop works before and after distro restart | PASS |
| Interop survives `systemctl restart systemd-binfmt` | PASS |
| Interop survives `daemon-reload` + restart | PASS |
| Generated files have correct content (config, service override, mount override) | PASS |
| Handler registered with `flags: PF`, `interpreter /init` | PASS |

**WSL1 compatibility:** The `:P`/`:FP` macro split is validated by `UnitTests::Interop` which runs on both WSL1 and WSL2 CI pipelines (WSL1 passed on build #285326).

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


src/linux/init/binfmt.h
```diff
@@ -21,6 +21,6 @@ Module Name:
 #define LX_INIT_BINFMT_NAME "WSLInterop"
 #define BINFMT_MISC_MOUNT_TARGET "/proc/sys/fs/binfmt_misc"
 #define BINFMT_MISC_REGISTER_FILE BINFMT_MISC_MOUNT_TARGET "/register"
-#define BINFMT_INTEROP_REGISTRATION_STRING(Name) ":" Name ":M::MZ::" LX_INIT_PATH ":P"
+#define BINFMT_INTEROP_REGISTRATION_STRING(Name) ":" Name ":M::MZ::" LX_INIT_PATH ":FP"
```
benhillis: This will break WSL1 which does not support the 'F' flag.

yeelam-gordon: Fixed: Split into two macros  BINFMT_INTEROP_REGISTRATION_STRING (:P) for WSL1 compatibility and BINFMT_INTEROP_REGISTRATION_STRING_VM (:FP) for WSL2-only paths (mini_init, systemd generator). config.cpp (WSL1 path, gated by !UtilIsUtilityVm()) continues using :P.

test/windows/UnitTests.cpp
```diff
@@ -372,28 +372,51 @@ class UnitTests
...
+            auto validateGeneratedBinfmtFiles = []() {
+                auto [binfmtConfig, _] = LxsstuLaunchWslAndCaptureOutput(L"cat /run/binfmt.d/WSLInterop.conf");
+                VERIFY_ARE_EQUAL(binfmtConfig, L":WSLInterop:M::MZ::/init:FP\n");
+
+                auto [serviceOverride, __] =
+                    LxsstuLaunchWslAndCaptureOutput(L"cat /run/systemd/generator/systemd-binfmt.service.d/override.conf");
+                VERIFY_IS_TRUE(serviceOverride.find(L"ExecStop=") != std::wstring::npos);
+                VERIFY_IS_TRUE(serviceOverride.find(L"ExecStart=") == std::wstring::npos);
+
+                auto [mountOverride, ___] =
```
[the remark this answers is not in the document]
yeelam-gordon: This is the established convention across the WSL test codebase. `_`, `__`, and `___` are used extensively as structured binding placeholders in MountTests.cpp, NetworkTests.cpp, SimpleTests.cpp, UnitTests.cpp, and more (50+ instances). While technically reserved in C++, MSVC compiles them without warnings and changing this pattern would be inconsistent with the rest of the test suite.

src/linux/init/init.cpp
```diff
@@ -365,33 +365,52 @@ ExecStart=/bin/mount -o bind,ro,X-mount.mkdir -t none /mnt/wslg/.X11-unix /tmp/.
...
-                THROW_LAST_ERROR_IF(WriteToFile(std::format("{}/override.conf", overrideFolder).c_str(), systemdBinfmtContent.c_str()) < 0);
+                THROW_LAST_ERROR_IF(WriteToFile(std::format("{}/override.conf", overrideFolder).c_str(), serviceOverrideContent) < 0);
+            }
+
+            // Install the override for proc-sys-fs-binfmt_misc.mount.
+            {
+                auto overrideFolder = std::format("{}/proc-sys-fs-binfmt_misc.mount.d", installPath);
+                THROW_LAST_ERROR_IF(UtilMkdirPath(overrideFolder.c_str(), 0755) < 0);
+                THROW_LAST_ERROR_IF(WriteToFile(std::format("{}/override.conf", overrideFolder).c_str(), mountOverrideContent) < 0);
             }
```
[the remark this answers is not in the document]
yeelam-gordon: Valid suggestion, but the test infrastructure only supports a single test distro (`test_distro` defined in Common.h). Adding multi-distro support would require significant test harness changes (installing/managing a second distro, coordinating lifecycle). The existing test validates the mechanism: generated files persist across restart, service stop/restart, and daemon-reload. The mount override with `Before=umount.target` is what actually prevents the cross-distro issue  and that's validated indirectly by `systemctl stop systemd-binfmt.service` not breaking interop. A dedicated multi-distro integration test could be added as a follow-up.

src/linux/init/init.cpp
```diff
@@ -365,33 +365,58 @@ ExecStart=/bin/mount -o bind,ro,X-mount.mkdir -t none /mnt/wslg/.X11-unix /tmp/.
...
+[Unit]
+DefaultDependencies=no
+Before=umount.target
+
+[Mount]
+Options=nosuid,nodev,noexec
+)";
+
+            THROW_LAST_ERROR_IF(UtilMkdirPath(binfmtConfigDirectory, 0755) < 0);
+            THROW_LAST_ERROR_IF(WriteToFile(binfmtConfigPath, binfmtConfigContent.c_str()) < 0);
```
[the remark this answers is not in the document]
yeelam-gordon: Fixed in 5d6c831: added an else branch that calls unlink(/run/binfmt.d/WSLInterop.conf) when the feature is disabled. Note that /run is tmpfs so this only matters within a single VM lifetime (distro restart already clears it), but this makes cleanup explicit. The generator overrides in /run/systemd/generator/ are already handled correctly since the generator simply does not create them when disabled.

yeelam-gordon:
@benhillis , please help to retake a look if you have any concern. Thanks in advanced.

src/linux/init/binfmt.h
```diff
@@ -21,6 +21,14 @@ Module Name:
...
 #define BINFMT_MISC_REGISTER_FILE BINFMT_MISC_MOUNT_TARGET "/register"
+
+//
+// N.B. The 'P' flag preserves Argv[0]. The 'F' flag (fix-binary) opens the interpreter at
+//      registration time, making it available across mount namespaces and chroot environments.
+//      WSL1 (lxcore) does not support the 'F' flag, so it must only be used in WSL2 (VM) paths.
+//
+
 #define BINFMT_INTEROP_REGISTRATION_STRING(Name) ":" Name ":M::MZ::" LX_INIT_PATH ":P"
+#define BINFMT_INTEROP_REGISTRATION_STRING_VM(Name) ":" Name ":M::MZ::" LX_INIT_PATH ":FP"
```
OneBlue, line 32: @benhillis: This sounds like something that we want. Any concerns with this flag ?

src/linux/init/init.cpp
```diff
@@ -365,34 +365,64 @@ ExecStart=/bin/mount -o bind,ro,X-mount.mkdir -t none /mnt/wslg/.X11-unix /tmp/.
...
 ExecStop=
-ExecStart=/bin/sh -c '(echo -1 > {}/{}) ; (echo "{}" > {})' )",
+ExecStartPost=/bin/sh -c '(echo -1 > {}/{} 2>/dev/null || true) ; echo "{}" > {}'
+)",
                 BINFMT_MISC_MOUNT_TARGET,
                 LX_INIT_BINFMT_NAME,
-                BINFMT_INTEROP_REGISTRATION_STRING(LX_INIT_BINFMT_NAME),
+                BINFMT_INTEROP_REGISTRATION_STRING_VM(LX_INIT_BINFMT_NAME),
                 BINFMT_MISC_REGISTER_FILE);
+            constexpr auto* mountOverrideContent = R"(# Note: This file is generated by WSL to keep binfmt_misc mounted during shutdown.
```
OneBlue: nit: I recommend moving this to the block where this is written. It's a bit confusing to read in its current state. For a moment I thought we were writting this to `binfmtConfigContent`

src/linux/init/init.cpp
```diff
@@ -365,34 +365,64 @@ ExecStart=/bin/mount -o bind,ro,X-mount.mkdir -t none /mnt/wslg/.X11-unix /tmp/.
...
+            {
+                auto overrideFolder = std::format("{}/proc-sys-fs-binfmt_misc.mount.d", installPath);
+                THROW_LAST_ERROR_IF(UtilMkdirPath(overrideFolder.c_str(), 0755) < 0);
+                THROW_LAST_ERROR_IF(WriteToFile(std::format("{}/override.conf", overrideFolder).c_str(), mountOverrideContent) < 0);
+            }
+        }
+        else
+        {
+            // Clean up stale binfmt config if the feature was previously enabled.
+            unlink("/run/binfmt.d/WSLInterop.conf");
```
OneBlue: `/run` is a tmpfs, so we don't need to delete files in future invocations

src/linux/init/init.cpp
```diff
@@ -365,34 +365,64 @@ ExecStart=/bin/mount -o bind,ro,X-mount.mkdir -t none /mnt/wslg/.X11-unix /tmp/.
...
 
             // Install the override for binfmt-support.service.
             {
                 auto overrideFolder = std::format("{}/binfmt-support.service.d", installPath);
                 THROW_LAST_ERROR_IF(UtilMkdirPath(overrideFolder.c_str(), 0755) < 0);
-                THROW_LAST_ERROR_IF(WriteToFile(std::format("{}/override.conf", overrideFolder).c_str(), systemdBinfmtContent.c_str()) < 0);
+                THROW_LAST_ERROR_IF(WriteToFile(std::format("{}/override.conf", overrideFolder).c_str(), serviceOverrideContent.c_str()) < 0);
             }
+
+            // Install the override for proc-sys-fs-binfmt_misc.mount.
```
OneBlue: What's the impact of systemd unmounting `binfmt_misc` on shutdown ? I would expect that this wouldn't affect other distributions, because the unmount would only apply to the current mount namespace

OneBlue (commented):
Added a couple comments. 

Also would it be possible to add a new test case that reproduces the binfmt issue that this change solves, so we can validate the fix and protect if from future regression ?

src/linux/init/init.cpp
yeelam-gordon: Fixed in a5a9aa3: removed the \mountOverrideContent\ variable entirely (along with the mount override block), so this nit is resolved  no more confusing variable ordering.

src/linux/init/init.cpp
yeelam-gordon: Fixed in a5a9aa3: removed the `unlink` and the entire `else` branch. As you noted, `/run` is tmpfs  it's recreated each boot, and `protectBinfmt` changes only take effect on distro restart anyway.

src/linux/init/init.cpp
yeelam-gordon: You're right  the unmount only affects the current mount namespace and wouldn't impact other distros.

The cross-distro handler loss is actually caused by **handler unregistration** (services writing `-1` to `/proc/sys/fs/binfmt_misc/WSLInterop` during shutdown), which is a kernel-global operation. The `ExecStop=` (empty) in the service override is what prevents this.

Since the mount override doesn't add cross-distro protection beyond what `ExecStop=` already provides, I've removed it in a5a9aa3. The PR now has two defense layers:

1. **`ExecStop=`**  prevents services from unregistering the WSL handler on shutdown
2. **`ExecStartPost`**  re-registers WSL's handler after `systemd-binfmt` starts, ensuring priority over conflicting distro configs

Tests updated to match.

benhillis:
@yeelam-gordon FYI — I rebased your branch onto current master (it was 221 commits behind) and added some hardening to your test, then opened #40617 to land it. Your 5 fix commits are unchanged (full credit preserved).

Test changes:
- `WSL2_TEST_METHOD` (TAEF can skip on WSL1)
- Uses `EnableSystemd()` helper for the primary test distro
- Adds `^flags:.*F` check on `/proc/sys/fs/binfmt_misc/WSLInterop` to catch a regression from `:FP` back to `:P`

Happy to close #40617 if you'd prefer to update this PR yourself — let me know which you prefer and I'll defer.

Context for the new test: I also tried an MS_PRIVATE mount-propagation fix in #40612 / #40614 before realizing binfmt_misc is keyed by user-ns, not mount-ns. CI failures on those PRs confirmed your approach is correct.

chemwolf6922:
Same fix merged in https://github.com/microsoft/WSL/pull/40621

Issue #14427, opened 2026-03-13
collect-wsl-logs.ps1 should warn users when a required tool is not installed


andremueiot:
**Is your feature request related to a problem? Please describe.**
WSL developers need complete logs in order to investigate issues. Certain logs can only be collected when proper tools are installed on Linux.
It's common for people to miss these details when reading the documentation. Others don't read docs at all. This can result in incomplete logs and prevent issues from being root caused.

**Describe the solution you'd like**
It would be very useful for the script that collects the logs to check if the required tools are installed and write a clear warning in case some are missing, making it clear to the user that they should seriously consider installing the tool.

**Describe alternatives you've considered**
An alternative considered would be failing the script's execution when a tool is missing, but that is probably too drastic.
There might be a legitimate reason for the tool to not be installed, and blocking the collection of the other logs would arguably be worse.

**Additional context**
Suggested warnings:
WARNING: tcpdump not found on WSL. For a more complete log collection install tcpdump.
WARNING: iptables not found on WSL. For a more complete log collection install iptables.

andremueiot:
Working on a fix.

PR #14447, opened 2026-03-16
Add warning to collect-wsl-logs.ps1 to be displayed when tool is missing

andremueiot:
On executing the log collection script, it will first confirm the required tools `tcpdump` and `iptables` are installed. For each tool
missing a warning is displayed, reminding the user that the tool should be installed prior to executing the script to get a more
complete log collection.

## PR Checklist

- [ ] **Closes:** #14427 
- [ ] **Tests:** No automated test is deemed necessary to cover the scenarios
- [ ] **Localization:** Wrote warning in plain English, like the messages already present in the log collection script
- [ ] **Dev docs:** No doc update needed
- [ ] **Documentation updated:** No doc update needed

## Validation Steps Performed
Ran the command below (from log collection instructions) manually in both scenarios where tools are installed and not installed.
Visually confirmed that both warning messages show up when the respective tool is not installed.
`.\collect-wsl-logs.ps1 -LogProfile networking`

### DIFF
--- a/diagnostics/collect-wsl-logs.ps1
+++ b/diagnostics/collect-wsl-logs.ps1
@@ -11,24 +11,45 @@ Set-StrictMode -Version Latest
 
 function Test-WslApplication {
     param (
-        $Name,
-        $Command
+        $Name
     )
 
     # Log warning when tool is not present in WSL
     try
     {
-        & wsl.exe -e sh -lc "command -v $Command >/dev/null 2>&1"
-        $available = ($LASTEXITCODE -eq 0)
+        # Capture any output/error from wsl.exe so we can distinguish
+        # between "command not found" and "WSL invocation failed".
+        $wslOutput = & wsl.exe -e sh -lc "command -v $Name >/dev/null 2>&1" 2>&1
+        $exitCode = $LASTEXITCODE
 
-        if (-not $available)
+        if ($exitCode -eq 0)
         {
-            Write-Warning "$Name not found in WSL. For a more complete log collection install $Name."
+            return $true
         }
 
-        return $available
+        # POSIX shells typically return 1 when command -v does not find the command, but we have seen sh returning 127.
+        if ($exitCode -eq 1 -or $exitCode -eq 127)
+        {
+            Write-Warning "$Name not found in WSL. For a more complete log collection install $Name."
+            return $false
+        }
+
+        # Any other exit code is assumed to indicate WSL itself failed
+        # (e.g., no distro installed, WSL disabled, or other startup error).
+        if (-not [string]::IsNullOrWhiteSpace($wslOutput))
+        {
+            Write-Warning "Unable to check for $Name in WSL. wsl.exe exited with code $exitCode. Output: $wslOutput"
+        }
+        else
+        {
+            Write-Warning "Unable to check for $Name in WSL. wsl.exe exited with code $exitCode."
+        }
+        return $false
     }
-    catch {}
+    catch
+    {
+        Write-Warning "Unexpected error while checking for $Name in WSL."
+    }
 
     return $false
 }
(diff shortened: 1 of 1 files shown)


diagnostics/collect-wsl-logs.ps1
```diff
@@ -9,6 +9,51 @@ Param (
...
+        $wslOutput = & wsl.exe -e sh -lc "command -v $Name >/dev/null 2>&1" 2>&1
+        $exitCode = $LASTEXITCODE
+
+        if ($exitCode -eq 0)
+        {
+            return $true
+        }
+
+        # POSIX shells typically return 1 when command -v does not find the command, but we have seen sh returning 127.
+        if ($exitCode -eq 1 -or $exitCode -eq 127)
```
OneBlue: I would recommend just considering non-zero as failures. Trying to guess if wsl failed based on output is pretty tricky and could cause confusion

diagnostics/collect-wsl-logs.ps1
```diff
@@ -9,6 +9,51 @@ Param (
...
+        }
+        else
+        {
+            Write-Warning "Unable to check for $Name in WSL. wsl.exe exited with code $exitCode."
+        }
+        return $false
+    }
+    catch
+    {
+        Write-Warning "Unexpected error while checking for $Name in WSL."
```
OneBlue: If we hit an exception we should probably just let it fail the entire script

diagnostics/collect-wsl-logs.ps1
```diff
@@ -9,6 +9,51 @@ Param (
...
+    param (
+        $Name
+    )
+
+    # Log warning when tool is not present in WSL
+    try
+    {
+        # Capture any output/error from wsl.exe so we can distinguish
+        # between "command not found" and "WSL invocation failed".
+        $wslOutput = & wsl.exe -e sh -lc "command -v $Name >/dev/null 2>&1" 2>&1
```
OneBlue: nit: We don't need to redirect stderr to stdout if we also do that at the powershell level

OneBlue (dismissed):

diagnostics/collect-wsl-logs.ps1
andremueiot: All non-zero results are being handled as failures. It just happens that we are providing more appropriate warning messages depending on the error code. After this first "if" block all other error codes are handled altogether, since the success case was already handled earlier.

Cases:
**0**: success, return $true
**1 or 127**: command not found, return $false
**anything else**: unable to check, return $false

I agree that it can be tricky to handle specific error codes, but these seem to work well. We can make further changes in the future if needed.

diagnostics/collect-wsl-logs.ps1
andremueiot: The inner ">/dev/null 2>&1" is there to get rid of stderr and stdout from the linux side, and the outer stderr redirection is there so that any errors logged by WSL end up stored in $wslOutput. So, there's reason to keep both. I agree it looks strange/wrong at first sight.

diagnostics/collect-wsl-logs.ps1
andremueiot: I'll update the PR accordingly.

diagnostics/collect-wsl-logs.ps1
```diff
@@ -9,6 +9,44 @@ Param (
...
+function Test-WslApplication {
+    param (
+        $Name
+    )
+
+    # Log warning when tool is not present in WSL
+
+    # Capture any output/error from wsl.exe so we can distinguish
+    # between "command not found" and "WSL invocation failed".
+    $wslOutput = & wsl.exe -e sh -lc "command -v $Name >/dev/null 2>&1" 2>&1
```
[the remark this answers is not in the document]
andremueiot: The function is only used by the script itself, so shell-injection is not a concern.

diagnostics/collect-wsl-logs.ps1
```diff
@@ -9,6 +9,44 @@ Param (
...
+    # Any other exit code is assumed to indicate WSL itself failed
+    # (e.g., no distro installed, WSL disabled, or other startup error).
+    if (-not [string]::IsNullOrWhiteSpace($wslOutput))
+    {
+        Write-Warning "Unable to check for $Name in WSL. wsl.exe exited with code $exitCode. Output: $wslOutput"
+    }
+    else
+    {
+        Write-Warning "Unable to check for $Name in WSL. wsl.exe exited with code $exitCode."
+    }
```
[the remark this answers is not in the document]
andremueiot: I prefer to rely on error codes, unless problems are observed.

keith-horton (approved):
@benhillis , what would be the process to ingest this change?

OneBlue (approved):

PR #14450, opened 2026-03-16
Add DCAT registration remediation

florelis:

To remediate the case where the registration key gets deleted, we re-set the key on every launch.

## PR Checklist

I added the remediation to `wslservice`, near the end of the `OnServiceStarting()`. The registration function writes to the registry without checking, as we would want to update the version if it is wrong for some reason.
I added an option to set the registration version to 0.0.0.0, which can be used for the initial installation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -239,7 +239,8 @@ add_compile_definitions(UNICODE
                         WSL_PACKAGE_VERSION_MAJOR=${PACKAGE_VERSION_MAJOR}
                         WSL_PACKAGE_VERSION_MINOR=${PACKAGE_VERSION_MINOR}
                         WSL_PACKAGE_VERSION_REVISION=${PACKAGE_VERSION_REVISION}
-                        WSL_BUILD_WSL_SETTINGS=${WSL_BUILD_WSL_SETTINGS})
+                        WSL_BUILD_WSL_SETTINGS=${WSL_BUILD_WSL_SETTINGS}
+                        DCAT_REGISTRATION_KEY=R"\(${DCAT_REGISTRATION_KEY}\)")
 
 if (${OFFICIAL_BUILD})
     add_compile_definitions(WSL_OFFICIAL_BUILD)
--- a/src/windows/common/helpers.cpp
+++ b/src/windows/common/helpers.cpp
@@ -713,9 +713,6 @@ try
     }
 
     wil::unique_hkey dcatKey = wsl::windows::common::registry::CreateKey(HKEY_LOCAL_MACHINE, TEXT(DCAT_REGISTRATION_KEY), KEY_SET_VALUE);
-    if (dcatKey)
-    {
-        wsl::windows::common::registry::WriteString(dcatKey.get(), nullptr, L"Version", registeredVersion.c_str());
-    }
+    wsl::windows::common::registry::WriteString(dcatKey.get(), nullptr, L"Version", registeredVersion.c_str());
 }
 CATCH_LOG()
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -239,7 +239,8 @@ add_compile_definitions(UNICODE
                         WSL_PACKAGE_VERSION_MAJOR=${PACKAGE_VERSION_MAJOR}
                         WSL_PACKAGE_VERSION_MINOR=${PACKAGE_VERSION_MINOR}
                         WSL_PACKAGE_VERSION_REVISION=${PACKAGE_VERSION_REVISION}
-                        WSL_BUILD_WSL_SETTINGS=${WSL_BUILD_WSL_SETTINGS})
+                        WSL_BUILD_WSL_SETTINGS=${WSL_BUILD_WSL_SETTINGS}
+                        DCAT_REGISTRATION_KEY=R"\(${DCAT_REGISTRATION_KEY}\)")
```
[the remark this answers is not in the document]
florelis: The `\` is for escaping `(` in CMake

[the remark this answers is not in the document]
florelis: The `\` is for escaping in CMake, it is not seen by C++

src/windows/common/helpers.cpp
```diff
@@ -689,3 +698,21 @@ bool wsl::windows::common::helpers::TryAttachConsole()
...
+try
+{
+    std::wstring registeredVersion;
+    if (IncludeVersionNumber)
+    {
+        registeredVersion.assign(TEXT(WSL_PACKAGE_VERSION));
+    }
+    else
+    {
+        registeredVersion.assign(L"0.0.0.0");
```
OneBlue, line 712: Should we remove this ? It's looking like this is always set to true

OneBlue (approved):
LGTM. One small comment

src/windows/common/helpers.cpp
JohnMcPMS: I expect to use this when doing an install via the SDK API.

florelis: To expand on what @JohnMcPMS said

My understanding is that the way to do an initial install via DCAT is to register a fake version 0.0 and then scan for updates, which would find an "update" to the latest version.

OneBlue: Ok I see. Do we know what the "scan" API is going to look like ? 

This is probably OK for now, but I don't love the idea of the SDK writing that key, since that creates a race condition if an install is running in parallel

JohnMcPMS: I'm not sure that we want to go to the extents required to prevent a race between (un)installs and an SDK API install.  I don't see a race here resulting in a negative outcome other than the expected arbitrary state at the end of the race which is unavoidable.

If you mean the existing scan API in the OS, it is https://learn.microsoft.com/en-us/windows/win32/api/wuapi/nf-wuapi-iupdatesearcher-search

If you mean the SDK "scan" API, that exists (CanRun), but I'm not sure of the implementation details.  And it won't be performing the OS scan.  It may end up reading from this value though.

OneBlue: > I'm not sure that we want to go to the extents required to prevent a race between (un)installs and an SDK API install. I don't see a race here resulting in a negative outcome other than the expected arbitrary state at the end of the race which is unavoidable.

Well MSI installs happen under a global lock, so no packages can be installed / uninstalled at the same time, but if this code runs after an MSI installation, it could put us into a broken state where the package is installed, but its version number is incorrect. 

Let's check back once we have a better view of what the DCAT acquisition API looks like

PR #14451, opened 2026-03-16
Implement SDK Import and shared file spec

JohnMcPMS:
Allow both Import and Load to take in file path or HANDLE and implement Import.

## PR Checklist

Move both Import and Load options to use a shared file specification structure that can accept a path or HANDLE+length. Add image name to import options. Implement Import and Load using a shared path to normalize to HANDLE+length for runtime usage.

## Validation Steps Performed
New tests.

### DIFF
--- a/src/windows/WslcSDK/wslcsdk.cpp
+++ b/src/windows/WslcSDK/wslcsdk.cpp
@@ -223,6 +223,54 @@ bool CopyProcessSettingsToRuntime(WSLAProcessOptions& runtimeOptions, const Wslc
         return false;
     }
 }
+
+// Normalizes file inputs to HANDLE+length.
+struct ImageFileResolver
+{
+    ImageFileResolver(const WslcImageFileSpecifier& imageFile)
+    {
+        bool pathSpecified = imageFile.path != nullptr;
+        bool handleSpecified = imageFile.content.handle != nullptr && imageFile.content.handle != INVALID_HANDLE_VALUE;
+
+        // Throw if both or neither options are given
+        THROW_HR_IF(E_INVALIDARG, (pathSpecified && handleSpecified) || (!pathSpecified && !handleSpecified));
+
+        if (pathSpecified)
+        {
+            wil::unique_handle imageFileHandle{
+                CreateFileW(imageFile.path, GENERIC_READ, FILE_SHARE_READ, nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr)};
+            THROW_LAST_ERROR_IF(!imageFileHandle);
+
+            LARGE_INTEGER fileSize{};
+            THROW_IF_WIN32_BOOL_FALSE(GetFileSizeEx(imageFileHandle.get(), &fileSize));
+
+            m_fileHandle = std::move(imageFileHandle);
+            m_rawHandle = m_fileHandle.get();
+            m_length = static_cast<ULONGLONG>(fileSize.QuadPart);
+        }
+        else
+        {
+            THROW_HR_IF(E_INVALIDARG, imageFile.content.length == 0);
+            m_rawHandle = imageFile.content.handle;
+            m_length = imageFile.content.length;
+        }
+    }
+
+    HANDLE Handle() const
+    {
+        return m_rawHandle;
+    }
+
+    ULONGLONG Length() const
+    {
+        return m_length;
+    }
+
+private:
+    HANDLE m_rawHandle;
+    ULONGLONG m_length;
+    wil::unique_handle m_fileHandle;
+};
 } // namespace
 
 // SESSION DEFINITIONS
@@ -969,10 +1017,17 @@ CATCH_RETURN();
 STDAPI WslcImportSessionImage(_In_ WslcSession session, _In_ const WslcImportImageOptions* options, _Outptr_opt_result_z_ PWSTR* errorMessage)
 try
 {
-    UNREFERENCED_PARAMETER(session);
-    UNREFERENCED_PARAMETER(options);
-    UNREFERENCED_PARAMETER(errorMessage);
-    return E_NOTIMPL;
+    ErrorInfoWrapper errorInfoWrapper{errorMessage};
+    auto internalType = CheckAndGetInternalType(session);
+    RETURN_HR_IF_NULL(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), internalType->session);
+    RETURN_HR_IF_NULL(E_POINTER, options);
+    RETURN_HR_IF_NULL(E_INVALIDARG, options->imageName);
+    ImageFileResolver imageFile{options->imageFile};
+
+    auto progressCallback = ProgressCallback::CreateIf(options);
+
+    return errorInfoWrapper.CaptureResult(
+        internalType->session->ImportImage(HandleToULong(imageFile.Handle()), options->imageName, progressCallback.get(), imageFile.Length()));
 }
 CATCH_RETURN();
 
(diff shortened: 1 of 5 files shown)


src/windows/WslcSDK/wslcsdk.h
```diff
@@ -368,9 +368,23 @@ typedef struct WslcPullImageOptions
...
+typedef struct WslcImageFileContent
+{
+    _In_opt_ HANDLE handle;
+    _In_ uint64_t length;
+} WslcImageFileContent;
+
+// Specifying both path and content will result in an error.
+typedef struct WslcImageFileSpecifier
+{
+    _In_opt_z_ PCWSTR path;
```
OneBlue: If we want to offer a way for the caller to just give us a path, I think we should make it into two API's, with one wrapper the other. Something like: 

```
WslcLoadSessionImage({Handle, ContentLength, ...});

WslcLoadSessionImageFromFile({Path, ...)
````

That way it's impossible for a user to "misuse" and set both handle and path for instance

Comment (JohnMcPMS) [reply src/windows/WslcSDK/wslcsdk.h]: It isn't just two functions, it is twice the number of functions that take in files.

I think I would be more inclined to agree if we separate out the required inputs from the optional ones:
```
WslcLoadSessionImage(session, handle, length, options, error);
WslcLoadSessionImageFromFile(session, path, options, error);
```
This way we have one "options" that contains all the optional parts.  Any new inputs (ex. cancellation handle) will most likely be optional as they aren't needed right now and thus can go into the options.

I mostly don't like having a different options struct for each of the file offering variants.

Comment (OneBlue) [reply src/windows/WslcSDK/wslcsdk.h]: Agreed, I think that's a good idea. We can have a shared struct for all the optional arguments and use function arguments for required arguments

File: src/windows/WslcSDK/wslcsdk.cpp [hunk 2]
```diff
@@ -223,6 +223,50 @@ bool CopyProcessSettingsToRuntime(WSLAProcessOptions& runtimeOptions, const Wslc
...
+    ImageFileResolver(PCWSTR path)
+    {
+        THROW_HR_IF_NULL(E_POINTER, path);
+
+        wil::unique_handle imageFileHandle{
+            CreateFileW(path, GENERIC_READ, FILE_SHARE_READ, nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr)};
+        THROW_LAST_ERROR_IF(!imageFileHandle);
+
+        LARGE_INTEGER fileSize{};
+        THROW_IF_WIN32_BOOL_FALSE(GetFileSizeEx(imageFileHandle.get(), &fileSize));
```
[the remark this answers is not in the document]
JohnMcPMS: Check out `handle_null_resource_policy` (used by `unique_handle`) and `handle_invalid_resource_policy` (used by `unique_hfile`) and maybe you will stop giving this feedback.

https://github.com/microsoft/wil/blob/3ecba603d9789435b0661be7b7154d45c195ef98/include/wil/resource.h#L2700-L2717

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -223,6 +223,50 @@ bool CopyProcessSettingsToRuntime(WSLAProcessOptions& runtimeOptions, const Wslc
...
+        return m_rawHandle;
+    }
+
+    ULONGLONG Length() const
+    {
+        return m_length;
+    }
+
+private:
+    HANDLE m_rawHandle;
```
OneBlue: nit: I recommend always default initializing here (even though it's assigned in the constructor). 

Uninitialized variables is one of our top bug creator :)

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -223,6 +223,50 @@ bool CopyProcessSettingsToRuntime(WSLAProcessOptions& runtimeOptions, const Wslc
...
+    ULONGLONG m_length;
+    wil::unique_handle m_fileHandle;
```
OneBlue: nit: We can use a `wsl::windows::common::relay::HandleWrapper` here. It can represent either an owning (wil::unique_handle) or non-owning (HANDLE) reference

test/windows/WslcSdkTests.cpp
```diff
@@ -450,41 +439,89 @@ class WslcSdkTests
...
+
+        // Negative: null image name must fail.
+        VERIFY_ARE_EQUAL(WslcImportSessionImageFromFile(m_defaultSession, nullptr, exportedImageTar.c_str(), &opts, nullptr), E_POINTER);
+
+        // Negative: missing file input must fail.
+        VERIFY_ARE_EQUAL(WslcImportSessionImageFromFile(m_defaultSession, "missing-file-input:test", nullptr, &opts, nullptr), E_POINTER);
+
+        // Negative: zero ContentLength must fail.
+        VERIFY_ARE_EQUAL(WslcImportSessionImage(m_defaultSession, "zero-length:test", GetCurrentThreadEffectiveToken(), 0, &opts, nullptr), E_INVALIDARG);
     }
```
OneBlue, line 525: nit: We should validate that fail with E_FAIL or similar if the caller tries to important an invalid file

OneBlue (approved):
LGTM. Minor comments

test/windows/WslcSdkTests.cpp
JohnMcPMS: Added duplicate of the load version... and it successfully fails even though load still "succeeds" despite their very common code paths...

OneBlue: Hmm what do you mean by 'load still "succeeds" ' ? Are you saying that there are code paths where we don't correctly detect errors ?

JohnMcPMS: The test `LoadImageNonTar` and new `ImportImageNonTar` pass the current executable/module to construct the image.  Import fails as expected.  Load has always not failed, but not had any effect on the image list either.

If I shell into the session and run docker load with bad files it does return error strings.  But when we call the engine it is just returning a 200 while ignoring the input (when I last debugged it).

OneBlue: Ok this is a runtime bug. I'll look into it

OneBlue: Here's the fix for that broken behavior: https://github.com/microsoft/WSL/pull/14483/changes

Feel free to merge now and do this as a followup once the runtime fix is merged

test/windows/WslcSdkTests.cpp
```diff
@@ -450,41 +439,89 @@ class WslcSdkTests
...
+        WslcLoadImageOptions opts{};
+
+        // Negative: null ImageHandle must fail.
+        VERIFY_ARE_EQUAL(WslcLoadSessionImage(m_defaultSession, nullptr, 1, &opts, nullptr), E_INVALIDARG);
+
         // Negative: INVALID_HANDLE_VALUE must fail.
+        VERIFY_ARE_EQUAL(WslcLoadSessionImage(m_defaultSession, INVALID_HANDLE_VALUE, 1, &opts, nullptr), E_INVALIDARG);
+
+        // Negative: zero ContentLength must fail.
+        VERIFY_ARE_EQUAL(WslcLoadSessionImage(m_defaultSession, GetCurrentThreadEffectiveToken(), 0, &opts, nullptr), E_INVALIDARG);
```
OneBlue, line 473: What's the reason for passing `GetCurrentThreadEffectiveToken()` ? Are we just trying to get any valid handle here ?

OneBlue (approved):
LGTM, one nit question in the tests

test/windows/WslcSdkTests.cpp
JohnMcPMS: Yes, because `GetCurrentProcess` uses `INVALID_HANDLE_VALUE` as its pseudo-handle and I wanted something that I knew would be guaranteed to be a real HANDLE.

PR #14453, opened 2026-03-17
Add cancellation to image build

johnstep:


src/windows/wslasession/WSLASession.cpp
```diff
@@ -536,7 +536,28 @@ try
...
+        // torn down. If it doesn't exit within 10 seconds, SIGKILL forces termination.
+        LOG_IF_FAILED(buildProcess.Get().Signal(WSLASignalSIGTERM));
+        try
+        {
+            buildProcess.Wait(10 * 1000);
+        }
+        catch (...)
+        {
+            LOG_IF_FAILED(buildProcess.Get().Signal(WSLASignalSIGKILL));
+            buildProcess.Wait();
```
OneBlue: nit: We should probably have a hard timeout here as well (just in case the VM gets into a bad state so we don't potentially get stuck here)

src/windows/wslc/core/Main.cpp
```diff
@@ -89,6 +104,13 @@ try
         // Using WSL shared utility to get the HRESULT from the caught exception.
         // CLIExecutionContext is a derived class of wsl::windows::common::ExecutionContext.
         result = wil::ResultFromCaughtException();
+
+        // If the user pressed Ctrl-C, exit silently instead of printing an error.
+        if (context.CancelEvent.is_signaled())
+        {
+            return 1;
```
dkbennett, line 117: Should print a message to the user acknowledging the cancel.

dkbennett (approved):

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -620,7 +620,35 @@ try
...
+        // cancel the BuildKit build through the daemon's gRPC API, ensuring running build steps are
+        // torn down. If it doesn't exit within 10 seconds, SIGKILL forces termination.
+        LOG_IF_FAILED(buildProcess.Get().Signal(WSLCSignalSIGTERM));
+        try
+        {
+            buildProcess.Wait(10 * 1000);
+        }
+        catch (...)
+        {
+            LOG_IF_FAILED(buildProcess.Get().Signal(WSLCSignalSIGKILL));
```
OneBlue: nit: I recommend running this only if the error code was ERROR_TIMEOUT

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -620,7 +620,35 @@ try
 
     io.AddHandle(std::make_unique<relay::LineBasedReadHandle>(buildProcess.GetStdHandle(2), captureOutput, false));
 
-    io.Run({});
+    try
+    {
+        io.Run({});
```
OneBlue, line 650: There's unfortunately a risk that we could end up in pipe deadlock here because if `CancelEvent` is set, we exit the IO loop, so if the build process has pending output, we could get stuck (or drop it) since we're not consuming it. 

To work around it, we could do something like this (instead of passing the handle to CreateIOContext())

(untested) 
```

auto killProcess = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { 

    LOG_IF_FAILED(buildProcess.Get().Signal(WSLCSignalSIGKILL));
     try
     {
          buildProcess.Wait(10 * 1000);
     }
     catch (...)
     {
         LOG_CAUGHT_EXCEPTION_MSG("Build process did not exit after SIGKILL");
      }
};

auto onCancel = [&](){ 

    // Create a timer so we have a hard stop in the case process doesn't exit in time
    auto killTimer = wil::unique_handle{CreateWaitableTimerW(nullptr, true, nullptr)};
    THROW_LAST_ERROR_IF(!killTimer);

    LARGE_INTEGER timeout{};
    timeout.QuadPart = -10 * 1000 * * 1000 * 10; // 10 seconds
    THROW_LAST_ERROR_IF_WIN32_BOOL_FAILED(SetWaitableTimer(killTimer.get(), &timeout, 0, nullptr, nullptr, 0);

    auto onTimeout = [&](){
         THROW_HR(E_ABORT);
     };

    io.AddHandle(std::make_unique<relay::EventHandle>(std::move(killTimer), std::move(onTimeout)), Flags::NeedNotComplete);


    // Cancel event signaled, try to gracefully terminate the build process
    buildProcess.Get().Signal(WSLCSignalSIGTERM);

};

io.AddHandle(std::make_unique<relay::EventHandle>(CancelEvent, std::move(onCancel)), Flags::NeedNotComplete);

io.Run({});

killProcess.release();  // Cancel the kill scope_exit if the IO is complete

OneBlue (approved):
LGTM. I think for now we should merge this change with the `ERROR_TIMEOUT` check for the kill logic, and we can do the deeper timeout change in a followup change

OneBlue (approved):

OneBlue (approved):

Issue #14193, opened 2026-02-11
WSL2 crashes on waking up from sleep


alexpf:
### Windows Version

Microsoft Windows [Version 10.0.26200.7623]

### WSL Version

2.6.3.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.87.2-1

### Distro Version

Ubuntu 24.04

### Other Software

Version: 1.109.2 (user setup)
Commit: 591199df409fbf59b4b52d5ad4ee0470152a9b31
Date: 2026-02-10T20:18:23.520Z
Electron: 39.3.0
ElectronBuildId: 13168319
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

### Repro Steps

Reproduces once in maybe 4 days.
I close the laptop at the end of the working day.
Once opening it back the next day, VSCode can't connect and asks to reload the window, which doesn't help.
On connecting to WSL through Windows Terminal, it returns Catastrophic failure Error code: Wsl/Service/E_UNEXPECTED.
wsl --shutdown is the workaround that helps.

### Expected Behavior

WSL never crashes.
This issue is really annoying, seems like there's no sense in developing under Windows - the environment is too unstable.

### Actual Behavior

VSCode can't connect and asks to reload the window, which doesn't help.
On connecting to WSL through Windows Terminal, it returns Catastrophic failure Error code: Wsl/Service/E_UNEXPECTED.
wsl --shutdown is the workaround that helps.

alexpf:
It's impossible: this script is asking me to reproduce the problem.
But as I said, it happens on waking from sleep, ~ once in 4 times, so I can't "replay" this situation.

l0tkaa:
Have you been able to see the memory usage of Node and of WSL to check whether or not that is causing the crash? I would open another terminal and run `watch -n 1 free -h` to see if it could be a memory issue with WSL and look for `vmmemWSL` in the details pane in the task manager to see any host memory being taken up by WSL. There's a possibility that Nodejs is leaking and causing the subsystem to use up all your memory.

alexpf:
@l0tkaa But once the laptop is back from sleep, WSL is already inaccessible. So another terminal won't help.

alexpf:
I'd really like to pinpoint the issue.
When I run the log collector, it asks me to reproduce the issue after the
[...]

PR #14455, opened 2026-03-17
Fix broken WSLCorePort channel after receive timeout

chemwolf6922:

This pattern shows up in multiple sleep -> wake -> wsl stuck reports:
image

In the current sequence number logic, the receive sequence will ++ without receiving any message. If timeout is allowed on the channel and it's not destroyed, the next receive will always get the N-1 message due to:
1. The stale message arrived after the timeout. Or,
2. The reply end never received the request, so it's send counter is N-1.

This will lock the channel in an unusable state.

This PR makes these changes to keep the WSLCorePort channels working after a receive timeout.
1. Replace the sequence number with transaction id. And add transaction step in the message header.
2. The transaction and non-transaction messages are differentiated by the transaction step. 
3. The old sequence number logic stays the same for non-transaction messages.
4. For transaction messages. The reply side will reply with the same id in the incoming transaction request.
5. For transaction messages. The receive loop will discard stale messages based on the transaction id.

I tried my best to update all the applicable receive and sender ends to use the transaction logic. But please help double check I'm not missing anything. Since a mismatch will cause the transaction to always fail.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/14193  https://github.com/microsoft/WSL/issues/14055 https://github.com/microsoft/WSL/issues/14014
    All but 8 unit test fails. Where 6 of them failed because of GPO settings or powershell issues. 2 of them (CGroupv1 and CaseSensitivity) failed but seems unrelated to the changes. I have appended logs for those failed tests in the end. 
    I'm also dog fooding this build right now. 

### Failed unit tests

```

UnitTests::UnitTests::Warnings [Failed]
Error: Verify: AreEqual(expectedWarnings, warnings) - Values (, wsl: Due to a current compatibility issue with Global Secure Access Client, DNS Tunneling is disabled.
) [File: D:\workspace\WSL\test\windows\UnitTests.cpp, Function: UnitTests::UnitTests::Warnings::<lambda_1>::operator (), Line: 2022]

UnitTests::UnitTests::KernelModules [Failed]
Error: Caught std::exception: D:\workspace\WSL\test\windows\Common.cpp(261)\wsltests.dll!00007FFD0D986FF8: (caller: 00007FFD0D9868B9) Exception(57) tid(8178) 8000FFFF Catastrophic failure
    Msg:[Command "C:\WINDOWS\system32\wsl.exe echo ok"returned unexpected exit code (0 != -1). Stdout: 'ok
'Stderr: 'wsl: The .wslconfig setting 'wsl2.kernel' is disabled by the computer policy.
'] [LxsstuLaunchCommandAndCaptureOutput(E_UNEXPECTED)]

UnitTests::UnitTests::VersionFlavorParsing [Failed]
WSL1 is disabled by the computer policy.
Please run 'wsl.exe --set-version tmpdistro 2' to upgrade to WSL2.
Error code: Wsl/Service/CreateInstance/WSL_E_WSL1_DISABLED
Error: Verify: AreEqual(LxsstuLaunchWsl(std::format(L"-d {} cat /etc/os-release || true", Distro).c_str()), 0L) - Values (4294967295, 0) [File: D:\workspace\WSL\test\windows\UnitTests.cpp, Function: UnitTests::UnitTests::VersionFlavorParsing::<lambda_1>::operator (), Line: 3924]

UnitTests::UnitTests::CaseSensitivity [Failed]
Verify: IsFalse(getCaseSensitivity(std::format(L"{}/l1/l2/l3-other", testDir)))
Error: Caught std::exception: D:\workspace\WSL\src\windows\common\filesystem.cpp(339)\wsltests.dll!00007FFD0D2E0169: (caller: 00007FFD0D07A634) Exception(11) tid(1a18) C0000101     [`anonymous-namespace'::EnsureCaseSensitiveDirectoryRecursive::<lambda_1>::operator ()(NtSetInformationFile(Directory, &IoStatus, &CaseInfo, sizeof(CaseInfo), FileCaseSensitiveInformation))]

UnitTests::UnitTests::CustomModulesVhd [Failed]
Error: Caught std::exception: D:\workspace\WSL\test\windows\Common.cpp(261)\wsltests.dll!00007FFD0D986FF8: (caller: 00007FFD0D987DAF) Exception(1) tid(3f90) 8000FFFF Catastrophic failure
    Msg:[Command "Powershell -NoProfile -Command "$acl = Get-Acl 'D:\workspace\WSL\test-modules.vhd' ; $acl.RemoveAccessRuleAll((New-Object System.Security.AccessControl.FileSystemAccessRule(\"Everyone\", \"Read\", \"None\", \"None\", \"Allow\"))); Set-Acl -Path 'D:\workspace\WSL\test-modules.vhd' -AclObject $acl""returned unexpected exit code (1 != 0). Stdout: ''Stderr: 'Get-Acl : The 'Get-Acl' command was found in the module 'Microsoft.PowerShell.Security', but the module could not be loaded. For more informa
tion, run 'Import-Module Microsoft.PowerShell.Security'.
At line:1 char:8
+ $acl = Get-Acl 'D:\workspace\WSL\test-modules.vhd' ; $acl.RemoveAcces ...
+        ~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Get-Acl:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CouldNotAutoloadMatchingModule

You cannot call a method on a null-valued expression.
At line:1 char:54
+ ... ules.vhd' ; $acl.RemoveAccessRuleAll((New-Object System.Security.Acce ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull

Set-Acl : The 'Set-Acl' command was found in the module 'Microsoft.PowerShell.Security', but the module could not be loaded. For more informa
tion, run 'Import-Module Microsoft.PowerShell.Security'.
At line:1 char:190
+ ... sRule("Everyone", "Read", "None", "None", "Allow"))); Set-Acl -Path ' ...
+                                                           ~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Set-Acl:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CouldNotAutoloadMatchingModule

'] [LxsstuLaunchCommandAndCaptureOutput(E_UNEXPECTED)]

UnitTests::UnitTests::BrokenDistroImport [Failed]
Error: Caught std::exception: D:\workspace\WSL\test\windows\Common.cpp(261)\wsltests.dll!00007FFD0D986FF8: (caller: 00007FFD0D987DAF) Exception(1) tid(13dc8) 8000FFFF Catastrophic failure
    Msg:[Command "Powershell -NoProfile -Command "New-Vhd EmptyVhd.vhdx  -SizeBytes 20MB""returned unexpected exit code (1 != 0). Stdout: ''Stderr: 'New-Vhd : Failed to create the virtual hard disk.
The system failed to create 'D:\workspace\WSL\EmptyVhd.vhdx'.
Failed to create the virtual hard disk.
The system failed to create 'D:\workspace\WSL\EmptyVhd.vhdx': The file exists. (0x80070050).
At line:1 char:1
+ New-Vhd EmptyVhd.vhdx  -SizeBytes 20MB
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [New-VHD], VirtualizationException
    + FullyQualifiedErrorId : OperationFailed,Microsoft.Vhd.PowerShell.Cmdlets.NewVhd

'] [LxsstuLaunchCommandAndCaptureOutput(E_UNEXPECTED)]

UnitTests::UnitTests::WslDebug [Failed]
wsl: The .wslconfig setting 'wsl2.kernelCommandLine' is disabled by the computer policy.
Error: Verify: AreEqual(LxsstuLaunchWsl(L"dmesg | grep -iF 'vmbus_send_tl_connect_request'"), 0L) - Values (1, 0) [File: D:\workspace\WSL\test\windows\UnitTests.cpp, Function: UnitTests::UnitTests::WslDebug, Line: 6425]

UnitTests::UnitTests::CGroupv1 [Failed]
Error: Verify: AreEqual(out, expected) - Values (/sys/fs/cgroup/unified cgroup2 cgroup2 rw,nosuid,nodev,noexec,relatime,nsdelegate
, ) [File: D:\workspace\WSL\test\windows\UnitTests.cpp, Function: UnitTests::UnitTests::CGroupv1::<lambda_1>::operator (), Line: 6435]

```


src/shared/inc/SocketChannel.h
```diff
@@ -199,47 +212,104 @@ class SocketChannel
...
+                }
+
+                // Reply matched (or ignore_sequence) — clear expected reply.
+                m_expected_reply_sequence.reset();
+            }
+            else
+            {
+                // We are the responder receiving a request — cache for echo-back.
+                m_pending_reply_sequence = receivedHeader->SequenceNumber;
+            }
```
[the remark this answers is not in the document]
chemwolf6922: This is not how the channel is designed to work.

src/shared/inc/SocketChannel.h
```diff
@@ -199,47 +213,118 @@ class SocketChannel
...
+                            THROW_ERRNO(EINVAL);
+#endif
+                        }
+                    }
+
+                    m_expected_reply_sequence.reset();
+                }
+                else
+                {
+                    m_pending_reply_sequence = receivedHeader->SequenceNumber;
```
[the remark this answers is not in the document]
chemwolf6922: There is no such "multi message response"

src/shared/inc/SocketChannel.h
```diff
@@ -199,47 +208,117 @@ class SocketChannel
...
-                THROW_HR_MSG(HCS_E_CONNECTION_TIMEOUT, "Timeout: %d, expected type: %hs, channel: %hs", timeout, ToString(TMessage::Type), m_name);
+
+#ifdef WIN32
+                if (errno == HCS_E_CONNECTION_TIMEOUT)
+                {
+                    THROW_HR_MSG(HCS_E_CONNECTION_TIMEOUT, "Timeout: %d, expected type: %hs, channel: %hs", timeout, ToString(TMessage::Type), m_name);
+                }
+#endif
+
+                return {nullptr, {}};
```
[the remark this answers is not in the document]
chemwolf6922: No, I don't want to touch the original logic. This system is too chaotic.

src/shared/inc/SocketChannel.h
```diff
@@ -199,47 +208,117 @@ class SocketChannel
...
+                        LOG_WARNING("Discard stale response on channel: {}. StaleSeq: {}, ExpectedSeq: {}", m_name, receivedHeader->SequenceNumber, m_sent_messages);
+#endif
+                        continue;
+                    }
+
+                    if (diff != 0)
+                    {
+#ifdef WIN32
+                        THROW_HR_MSG(E_UNEXPECTED, "Unexpected response sequence: %u, expected: %u, channel: %hs", receivedHeader->SequenceNumber, m_sent_messages, m_name);
+#else
```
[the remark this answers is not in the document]
chemwolf6922: You are reading outdated code.

src/shared/inc/SocketChannel.h
```diff
@@ -103,12 +108,30 @@ class SocketChannel
...
         THROW_INVALID_ARG_IF(m_name == nullptr || span.size() < sizeof(TMessage));
 
-        m_sent_messages++;
+        uint32_t sequenceNumber = 0;
+        if (m_strict_request_end)
+        {
+            std::lock_guard<std::mutex> sequenceLock{m_sequenceMutex};
+            m_sent_messages += m_next_send_increment;
+            m_next_send_increment = nextSendIncrement;
+            sequenceNumber = m_sent_messages;
```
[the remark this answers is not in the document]
chemwolf6922: ?, it also allows any other number.

src/shared/inc/SocketChannel.h
```diff
@@ -199,53 +222,124 @@ class SocketChannel
...
+                }
 
-            return {nullptr, {}};
-        }
+                if (m_strict_request_end)
+                {
+                    // Skip stale message for strict request end.
+                    std::lock_guard<std::mutex> sequenceLock{m_sequenceMutex};
+                    uint32_t expectedSequence = m_sent_messages + expectedOffset;
+                    auto diff = static_cast<int32_t>(receivedHeader->SequenceNumber - expectedSequence);
```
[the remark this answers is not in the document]
chemwolf6922: This is intentional to cover the overflow case.

src/shared/inc/SocketChannel.h
```diff
@@ -199,47 +208,117 @@ class SocketChannel
...
+#else
+                        LOG_ERROR("Unexpected response sequence: {}, expected: {}, channel: {}", receivedHeader->SequenceNumber, m_sent_messages, m_name);
+                        THROW_ERRNO(EINVAL);
+#endif
+                    }
+                }
+                else if (m_strict_reply_end)
+                {
+                    // The send id must catch up to the latest received id.
+                    std::lock_guard<std::mutex> sequenceLock{m_sequenceMutex};
```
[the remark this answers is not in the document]
chemwolf6922: overflow is handled when comparing.

src/windows/service/exe/LxssCreateProcess.h
```diff
@@ -85,16 +85,15 @@ class LxssCreateProcess
         wsl::shared::MessageWriter<CREATE_PROCESS_MESSAGE> message(LxInitCreateProcess);
         message.WriteString(message->PathIndex, Path);
         gsl::copy(as_bytes(gsl::span(ArgumentsData)), message.InsertBuffer(message->CommandLineIndex, ArgumentsData.size()));
-        channel.SendMessage<CREATE_PROCESS_MESSAGE>(message.Span());
+        channel.SendMessage<CREATE_PROCESS_MESSAGE>(message.Span(), 2);
```
ptrivedi: why is the 'next send increment' set to 2 here?

src/shared/inc/SocketChannel.h
```diff
@@ -103,12 +108,30 @@ class SocketChannel
...
+            sequenceNumber = m_sent_messages;
+        }
+        else if (m_strict_reply_end)
+        {
+            std::lock_guard<std::mutex> sequenceLock{m_sequenceMutex};
+            m_sent_messages++;
+            sequenceNumber = m_sent_messages;
+        }
+        else
+        {
```
ptrivedi, line 158: Looks like m_sequenceMutex is protecting m_sent_messages, as a receiver thread may try to use it.

In the 'else' block it is unprotected because we think that the receiver thread will only try to access it when m_strict_request_end or m_strict_reply_end are true?

src/shared/inc/SocketChannel.h
```diff
@@ -199,53 +222,124 @@ class SocketChannel
...
-        m_received_messages++;
-
-        auto receivedSpan = ReceiveImpl(TMessage::Type, timeout);
-        if (receivedSpan.empty())
+        for (;;)
         {
 
-#ifdef WIN32
-            if (errno == HCS_E_CONNECTION_TIMEOUT)
+            m_received_messages++;
```
ptrivedi: Looks like m_received_messages++ will fire on every iteration including stale message discards. For non-strict channels, could these increments cause false sequence errors, especially if a channel were ever switched from
  strict to non-strict mode. Maybe it is not an issue today, but something to think about in case this will cause an inconsistency

src/windows/service/exe/WslCoreInstance.cpp
```diff
@@ -378,6 +378,9 @@ void WslCoreInstance::Initialize()
     // Create a console manager that will be used to manage session leaders.
     m_consoleManager = ConsoleManager::CreateConsoleManager(m_initChannel);
 
+    // Enable strict request-end sequencing.
+    m_initChannel->GetChannel().SetStrictRequestEnd();
```
ptrivedi: Maybe this needs to happen before m_initChannel is passed to CreateConsoleManager. Is there a possiblity that CreateConsoleManager ends up sending messages synchronously before the SetStrictRequestEnd() gets set? If m_initChannel traffic can somehow flow before SetStrictRequestEnd() is called/completed, there's a narrow window where the sequence can diverge

src/shared/inc/SocketChannel.h
```diff
@@ -199,53 +222,124 @@ class SocketChannel
...
 
-            return {nullptr, {}};
-        }
+                if (m_strict_request_end)
+                {
+                    // Skip stale message for strict request end.
+                    std::lock_guard<std::mutex> sequenceLock{m_sequenceMutex};
+                    uint32_t expectedSequence = m_sent_messages + expectedOffset;
+                    auto diff = static_cast<int32_t>(receivedHeader->SequenceNumber - expectedSequence);
+                    if (diff < 0)
```
ptrivedi: Should there be a limit on how many stale messages can be discarded before treating the channel as broken? What if there's a corrupted channel and a runaway loop draining messages from it... although there is trace logging here, but from caller's perspective it would be silent

src/shared/inc/SocketChannel.h
```diff
@@ -321,33 +427,29 @@ class SocketChannel
...
-            (!m_ignore_sequence && header.SequenceNumber != expectedSequence))
+
+        if (header.MessageSize < sizeof(header) || (expected != LxMiniInitMessageAny && header.MessageType != expected))
         {
 #ifdef WIN32
 
             THROW_HR_MSG(
                 E_UNEXPECTED,
-                "Protocol error: Received message size: %u, type: %u, sequence: %u. Expected type: %u, expected sequence: %u, "
+                "Protocol error: Received message size: %u, type: %u. Expected type: %u, "
```
ptrivedi: I see that the validation involving sequence number and expected sequence moved upstream, so expectedSequence is not strictly needed here... but wonder if we should leave the sequence parameter in ValidateMessageHeader just for logging/debugging purposes.

src/shared/inc/SocketChannel.h
```diff
@@ -199,53 +222,124 @@ class SocketChannel
 #endif
         }
 
-        m_received_messages++;
-
-        auto receivedSpan = ReceiveImpl(TMessage::Type, timeout);
-        if (receivedSpan.empty())
+        for (;;)
         {
```
ptrivedi, line 263: nit: extra blank line

ptrivedi:
> ## Summary of the Pull Request
> This pattern shows up in multiple sleep -> wake -> wsl stuck reports: image
> 
> In the current sequence number logic, the receive sequence will ++ without receiving any message. If timeout is allowed on the channel and it's not destroyed, the next receive will always get the N-1 message due to:
> 
> 1. The stale message arrived after the timeout. Or,
> 2. The reply end never received the request, so it's send counter is N-1.
> 
> This will lock the channel in an unusable state.
> 
> This PR makes these changes to keep the WSLCorePort channels working after a receive timeout.
> 
> 1. Add options to enable an alternative sequence number sync & check logic that:
> 2. The reply side always sync up to the request side's latest sequence number.
> 3. The request side skips any stale message.
> 
> These are only applied to the WSLCorePort channels to reduce risk. Though other channels may face the same problem.
> 
> ## PR Checklist
> * [x]  **Closes:** [WSL2 crashes on waking up from sleep #14193](https://github.com/microsoft/WSL/issues/14193)  [WSL 2.6.3.0: Terminal crash after hibernation/sleep with [process exited with code 1] #14055](https://github.com/microsoft/WSL/issues/14055) [Error code: Wsl/Service/E_UNEXPECTED #14014](https://github.com/microsoft/WSL/issues/14014)
> * [ ]  **Communication:** I've discussed this with core contributors already. If work hasn't been agreed, this work might be rejected
> * [x]  **Tests:** Added/updated if needed and all pass
>   All but 8 unit test fails. Where 6 of them failed because of GPO settings or powershell issues. 2 of them (CGroupv1 and CaseSensitivity) failed but seems unrelated to the changes. I have appended logs for those failed tests in the end.
>   I'm also dog fooding this build right now.
> * [ ]  **Localization:** All end user facing strings can be localized
> * [ ]  **Dev docs:** Added/updated if needed
> * [ ]  **Documentation updated:** If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/wsl/) and link it here: #xxx
> 
> ## Detailed Description of the Pull Request / Additional comments
> ## Validation Steps Performed
> ### Failed unit tests
> ```
> 
> UnitTests::UnitTests::Warnings [Failed]
> Error: Verify: AreEqual(expectedWarnings, warnings) - Values (, wsl: Due to a current compatibility issue with Global Secure Access Client, DNS Tunneling is disabled.
> ) [File: D:\workspace\WSL\test\windows\UnitTests.cpp, Function: UnitTests::UnitTests::Warnings::<lambda_1>::operator (), Line: 2022]
> 
> UnitTests::UnitTests::KernelModules [Failed]
> Error: Caught std::exception: D:\workspace\WSL\test\windows\Common.cpp(261)\wsltests.dll!00007FFD0D986FF8: (caller: 00007FFD0D9868B9) Exception(57) tid(8178) 8000FFFF Catastrophic failure
>     Msg:[Command "C:\WINDOWS\system32\wsl.exe echo ok"returned unexpected exit code (0 != -1). Stdout: 'ok
> 'Stderr: 'wsl: The .wslconfig setting 'wsl2.kernel' is disabled by the computer policy.
> '] [LxsstuLaunchCommandAndCaptureOutput(E_UNEXPECTED)]
> 
> UnitTests::UnitTests::VersionFlavorParsing [Failed]
> WSL1 is disabled by the computer policy.
> Please run 'wsl.exe --set-version tmpdistro 2' to upgrade to WSL2.
> Error code: Wsl/Service/CreateInstance/WSL_E_WSL1_DISABLED
> Error: Verify: AreEqual(LxsstuLaunchWsl(std::format(L"-d {} cat /etc/os-release || true", Distro).c_str()), 0L) - Values (4294967295, 0) [File: D:\workspace\WSL\test\windows\UnitTests.cpp, Function: UnitTests::UnitTests::VersionFlavorParsing::<lambda_1>::operator (), Line: 3924]
> 
> UnitTests::UnitTests::CaseSensitivity [Failed]
> Verify: IsFalse(getCaseSensitivity(std::format(L"{}/l1/l2/l3-other", testDir)))
> Error: Caught std::exception: D:\workspace\WSL\src\windows\common\filesystem.cpp(339)\wsltests.dll!00007FFD0D2E0169: (caller: 00007FFD0D07A634) Exception(11) tid(1a18) C0000101     [`anonymous-namespace'::EnsureCaseSensitiveDirectoryRecursive::<lambda_1>::operator ()(NtSetInformationFile(Directory, &IoStatus, &CaseInfo, sizeof(CaseInfo), FileCaseSensitiveInformation))]
> 
> UnitTests::UnitTests::CustomModulesVhd [Failed]
> Error: Caught std::exception: D:\workspace\WSL\test\windows\Common.cpp(261)\wsltests.dll!00007FFD0D986FF8: (caller: 00007FFD0D987DAF) Exception(1) tid(3f90) 8000FFFF Catastrophic failure
>     Msg:[Command "Powershell -NoProfile -Command "$acl = Get-Acl 'D:\workspace\WSL\test-modules.vhd' ; $acl.RemoveAccessRuleAll((New-Object System.Security.AccessControl.FileSystemAccessRule(\"Everyone\", \"Read\", \"None\", \"None\", \"Allow\"))); Set-Acl -Path 'D:\workspace\WSL\test-modules.vhd' -AclObject $acl""returned unexpected exit code (1 != 0). Stdout: ''Stderr: 'Get-Acl : The 'Get-Acl' command was found in the module 'Microsoft.PowerShell.Security', but the module could not be loaded. For more informa
> tion, run 'Import-Module Microsoft.PowerShell.Security'.
> At line:1 char:8
> + $acl = Get-Acl 'D:\workspace\WSL\test-modules.vhd' ; $acl.RemoveAcces ...
> +        ~~~~~~~
>     + CategoryInfo          : ObjectNotFound: (Get-Acl:String) [], CommandNotFoundException
>     + FullyQualifiedErrorId : CouldNotAutoloadMatchingModule
> 
> You cannot call a method on a null-valued expression.
> At line:1 char:54
> + ... ules.vhd' ; $acl.RemoveAccessRuleAll((New-Object System.Security.Acce ...
> +                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>     + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
>     + FullyQualifiedErrorId : InvokeMethodOnNull
> 
> Set-Acl : The 'Set-Acl' command was found in the module 'Microsoft.PowerShell.Security', but the module could not be loaded. For more informa
> tion, run 'Import-Module Microsoft.PowerShell.Security'.
> At line:1 char:190
> + ... sRule("Everyone", "Read", "None", "None", "Allow"))); Set-Acl -Path ' ...
> +                                                           ~~~~~~~
>     + CategoryInfo          : ObjectNotFound: (Set-Acl:String) [], CommandNotFoundException
>     + FullyQualifiedErrorId : CouldNotAutoloadMatchingModule
> 
> '] [LxsstuLaunchCommandAndCaptureOutput(E_UNEXPECTED)]
> 
> UnitTests::UnitTests::BrokenDistroImport [Failed]
> Error: Caught std::exception: D:\workspace\WSL\test\windows\Common.cpp(261)\wsltests.dll!00007FFD0D986FF8: (caller: 00007FFD0D987DAF) Exception(1) tid(13dc8) 8000FFFF Catastrophic failure
>     Msg:[Command "Powershell -NoProfile -Command "New-Vhd EmptyVhd.vhdx  -SizeBytes 20MB""returned unexpected exit code (1 != 0). Stdout: ''Stderr: 'New-Vhd : Failed to create the virtual hard disk.
> The system failed to create 'D:\workspace\WSL\EmptyVhd.vhdx'.
> Failed to create the virtual hard disk.
> The system failed to create 'D:\workspace\WSL\EmptyVhd.vhdx': The file exists. (0x80070050).
> At line:1 char:1
> + New-Vhd EmptyVhd.vhdx  -SizeBytes 20MB
> + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>     + CategoryInfo          : NotSpecified: (:) [New-VHD], VirtualizationException
>     + FullyQualifiedErrorId : OperationFailed,Microsoft.Vhd.PowerShell.Cmdlets.NewVhd
> 
> '] [LxsstuLaunchCommandAndCaptureOutput(E_UNEXPECTED)]
> 
> UnitTests::UnitTests::WslDebug [Failed]
> wsl: The .wslconfig setting 'wsl2.kernelCommandLine' is disabled by the computer policy.
> Error: Verify: AreEqual(LxsstuLaunchWsl(L"dmesg | grep -iF 'vmbus_send_tl_connect_request'"), 0L) - Values (1, 0) [File: D:\workspace\WSL\test\windows\UnitTests.cpp, Function: UnitTests::UnitTests::WslDebug, Line: 6425]
> 
> UnitTests::UnitTests::CGroupv1 [Failed]
> Error: Verify: AreEqual(out, expected) - Values (/sys/fs/cgroup/unified cgroup2 cgroup2 rw,nosuid,nodev,noexec,relatime,nsdelegate
> , ) [File: D:\workspace\WSL\test\windows\UnitTests.cpp, Function: UnitTests::UnitTests::CGroupv1::<lambda_1>::operator (), Line: 6435]
> ```

Is there a possibility to have a unit test that specifically covers timeout-then-recover scenario -- basically simulate a timed-out receive follwed by a successful send/receive?

src/windows/service/exe/LxssCreateProcess.h
chemwolf6922: This message expects 2 replies. Which will use the message id N and N + 1. Which means to avoid overlap, the next outbound message should have the message id N + 2 instead of the normal N + 1. So the next increment is set to 2 here.

src/shared/inc/SocketChannel.h
chemwolf6922: Yes, they could face the same problem. And this fix is designed so that all other channels remain the same behavior. As blindly changing those may introduce bigger problems...

For the future, I'll suggest improve the channel design so it can handle stuffs like concurrent requests / bi-directional requests and stream replies (one request, multiple reply) natively. But that will take major changes to the code base.

src/shared/inc/SocketChannel.h
chemwolf6922: The current logic is not setting any arbitrary limit on this and allows half of the range of int32 to be stale messages.
Choosing that arbitrary value is difficult. And it's likely not going to make the situation better.
This loop won't be forever though. It will end up in a timeout or an invalid message exception if the expected message is not received.

src/windows/service/exe/WslCoreInstance.cpp
chemwolf6922: The CreateConsoleManager call doesn't do anything. But yes, moving it ahead of that call is more clear.

src/shared/inc/SocketChannel.h
chemwolf6922: Add the sequence number back. But since it's already checked. Only adding back one copy.

src/shared/inc/SocketChannel.h
chemwolf6922: Yes, this is to avoid unnecessary lock usage.

chemwolf6922:
> Is there a possibility to have a unit test that specifically covers timeout-then-recover scenario -- basically simulate a timed-out receive followed by a successful send/receive?

@ptrivedi  The WSL project is lacking test infrastructures for internal components like this. The Windows tests are e2e tests. And the Linux tests are syscall tests. There might need to be some major changes to make this happen.

src/shared/inc/SocketChannel.h
```diff
@@ -103,12 +108,30 @@ class SocketChannel
 
         THROW_INVALID_ARG_IF(m_name == nullptr || span.size() < sizeof(TMessage));
 
-        m_sent_messages++;
+        uint32_t sequenceNumber = 0;
+        if (m_strict_request_end)
+        {
+            std::lock_guard<std::mutex> sequenceLock{m_sequenceMutex};
+            m_sent_messages += m_next_send_increment;
```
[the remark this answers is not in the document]
chemwolf6922: If the caller insists to destroy their code logic. Then let it be.

src/shared/inc/SocketChannel.h
```diff
@@ -199,53 +222,123 @@ class SocketChannel
...
+#else
+                        LOG_ERROR("Unexpected response sequence: {}, expected: {}, channel: {}", receivedHeader->SequenceNumber, expectedSequence, m_name);
+                        THROW_ERRNO(EINVAL);
+#endif
+                    }
+                }
+                else if (m_strict_reply_end)
+                {
+                    // The send id must catch up to the latest received id.
+                    std::lock_guard<std::mutex> sequenceLock{m_sequenceMutex};
```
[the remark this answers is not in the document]
chemwolf6922: overflow is allowed

OneBlue (commented):
Thank you for doing this. 

For context we added those "sequence numbers" to detect when things go out sync, so most of the time when they don't match it's because something went wrong. 

If we want to be able "recover" from these sequences getting out of sync, I would recommend moving into a "transaction-id" model instead, where we can safely discard outdated messages without having to "guess" from SocketChannel if the caller meant to reply to the current transaction / start a new one. 

To do something like that, I think the cleanest model would to be to have transaction RAII object to keep track of the current transaction. Something like: 


```
struct MESSAGE_HEADER
{
    [...]
    unsigned int TransactionId; // Replaces SequenceNumber
};

class Transaction
{
    Transaction(SocketChannel& Channel, unsigned int TransactionId);

    // Send a message and set the correct transaction ID
    template <typename TMessage>
    void Send(TMessage& Message); 

    // Receive a message and drops message that don't match the transaction ID
    template <typename TMessage>
    TMessage& ReceiveMessage(gsl::span<gsl::byte>* responseSpan = nullptr, TTimeout timeout = DefaultSocketTimeout)
};
```

We can then update `SocketChannel` to handle transactions like this: 

```
 class SocketChannel
{
    Transaction StartTransaction();
}  
```

On the receiving side, we could do something like this: 

```
 class SocketChannel
{
    std::tuple<Transaction, Message, Span> ReceiveTransaction();
}
```

So the receiver gets a transaction with the correctly set transaction number. 

Obviously this makes for a bigger change, but I think that would be the best way to solve this issue for the long term

chemwolf6922:
Hi @OneBlue . I updated the design to use transaction id + transaction step to better identify different messages. Please help review. Please also help verify if I updated all the correct pairs. Thanks.

src/linux/init/GnsEngine.cpp
```diff
@@ -723,22 +724,23 @@ void GnsEngine::run()
...
 
             GNS_LOG_INFO("Processing Next Message Successful ({:#x})", return_value);
-            statusRoutine(return_value, "");
+            statusRoutine(return_value, "", transaction);
         }
         catch (const std::exception& e)
         {
             GNS_LOG_ERROR("Error while processing message: {}", e.what());
-            statusRoutine(-1, e.what());
+            statusRoutine(-1, e.what(), transaction);
```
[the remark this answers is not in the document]
chemwolf6922: This PR does not change the original behavior here.
The statusRoutine will return early with should_continue set to false if the channel breaks. Later exceptions all have the receive step completed. And since the receive has no timeout and it's waiting for the next transaction request, it's less likely to throw inside the receive compare to before. As wrong messages will be ignored now instead of creating an exception.

chemwolf6922:
Hi @benhillis, I resolved the copilot comments. Could you please help take another look? Thanks.

src/linux/init/config.cpp
```diff
@@ -618,7 +618,7 @@ try
 }
 CATCH_LOG()
 
-int ConfigInitializeInstance(wsl::shared::SocketChannel& Channel, gsl::span<gsl::byte> Buffer, wsl::linux::WslDistributionConfig& Config)
+int ConfigInitializeInstance(const std::function<void(const gsl::span<gsl::byte>&)>& SendResponse, gsl::span<gsl::byte> Buffer, wsl::linux::WslDistributionConfig& Config)
```
OneBlue, line 621: What's the reason for taking an std::function instead of a reference to the transaction ?

src/linux/init/GnsEngine.cpp
```diff
@@ -723,22 +724,23 @@ void GnsEngine::run()
 
     while (true)
     {
+        auto transaction = channel.ReceiveTransaction();
```
OneBlue, line 727: if `gns` is running is test mode, `channel` would be an empty channel here. I'm not sure that we should allow transaction on empty channels since that could lead to some weird states, maybe the easiest way here would be to make the transaction an std::optional, and plumb the optional through `NotificationRoutine` and `StatusRoutine` so we don't have to deal with it in this class

src/linux/init/init.cpp
```diff
@@ -2479,15 +2498,18 @@ Return Value:
...
-                ConfigRemountDrvFs(Span, channel, Config);
+                ConfigRemountDrvFs(Span, transaction, Config);
                 break;
 
             case LxInitMessageTerminateInstance:
-                InitTerminateInstance(Span, channel, Config);
-                break;
+            {
+                auto SendResult = [&](bool result) { transaction.SendResultMessage<bool>(result); };
+                InitTerminateInstance(Span, SendResult, Config);
```
OneBlue, line 2507: Same question here, could we pass a reference to the transaction instead ?

OneBlue: Same for below calls

src/shared/inc/SocketChannel.h
```diff
@@ -199,20 +252,167 @@ class SocketChannel
...
+                break;
+            }
 
+            // Handle transaction messages
+            if (header->TransactionStep == static_cast<uint32_t>(TRANSACTION_STEP::NONE))
+            {
+                // Skip stale non-transaction messages
 #ifdef WIN32
-            if (errno == HCS_E_CONNECTION_TIMEOUT)
+                WSL_LOG(
```
OneBlue, line 333: nit: If we receive unexpected messages, or out of order message, I recommend logging the message type, to make debugging easier

OneBlue (commented):
This looks great, very clean and should help us catch issues & recover from socket timeouts. 

Couple minor comments, but I think we can merge this soon. Thank you for doing this !

src/linux/init/config.cpp
chemwolf6922: This function is also used for WSL1, where the transaction is initiated with UtilReadMessageLxBus. Which does not work with the new transaction model.
So a function is used here so WSL1 can keep the original behavior while WSL2 can use the transaction.

src/linux/init/init.cpp
chemwolf6922: Please see the comment above. This is also for compatibility with WSL1.

src/linux/init/GnsEngine.cpp
chemwolf6922: Creating a transaction on an empty channel is benign. It will not disrupt that channel's state. And using that transaction will behave the same as calling receive or send on that empty channel. The test path in StartGns already correctly ignores the input Transaction. 
I feel that making the created transaction optional is too complicated for normal callers. Maybe we should keep it as is for this.

src/shared/inc/SocketChannel.h
```diff
@@ -199,20 +252,180 @@ class SocketChannel
...
+                    }
+                }
+                break;
+            }
 
+            // Handle transaction messages
+            if (header->TransactionStep == static_cast<uint32_t>(TRANSACTION_STEP::NONE))
+            {
+                // Skip stale non-transaction messages
 #ifdef WIN32
```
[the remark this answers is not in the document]
chemwolf6922: The non transaction message's counter is incremented before the previous non transaction message receive timeouts. So it should not be incremented here again for the stale message.

src/linux/init/config.cpp
OneBlue: Ok that makes sense, thank you !

OneBlue (approved):

PR #14462, opened 2026-03-17
Implement SDK IO callbacks

JohnMcPMS:
Implements IO callbacks for processes created through the SDK.

## PR Checklist

Implements `WslcSetProcessSettingsIOCallback` and the use of those inputs for both initial and subsequent processes.  A new thread is started to run the `MultiHandleWait` and the container/process objects hold a `shared_ptr` reference to it.  The caller must keep one of the objects alive for the callback thread to keep working.

**Open question:** I made it an error to specify callbacks and not ATTACH on the container start. ***Should we just force ATTACH if callbacks are provided?***

## Validation Steps Performed
Tests added and run.

### DIFF
--- a/src/windows/WslcSDK/IOCallback.cpp
+++ b/src/windows/WslcSDK/IOCallback.cpp
@@ -14,7 +14,8 @@ Abstract:
 #include "precomp.h"
 #include "WslcsdkPrivate.h"
 
-IOCallback::IOCallback(IWSLAProcess* process, const WslcContainerProcessIOCallbackOptions& options)
+IOCallback::IOCallback(IWSLAProcess* process, const WslcContainerProcessIOCallbackOptions& options) :
+    m_process(process), m_callbackOptions(std::make_unique<WslcContainerProcessIOCallbackOptions>(options))
 {
     using namespace wsl::windows::common::relay;
 
@@ -22,8 +23,8 @@ IOCallback::IOCallback(IWSLAProcess* process, const WslcContainerProcessIOCallba
         std::function<void(const gsl::span<char>& Buffer)> function;
         if (callback)
         {
-            function = [callback, context](const gsl::span<char>& buffer) {
-                callback(reinterpret_cast<const BYTE*>(buffer.data()), static_cast<uint32_t>(buffer.size()), context);
+            function = [ioHandle, callback, context](const gsl::span<char>& buffer) {
+                callback(ioHandle, reinterpret_cast<const BYTE*>(buffer.data()), static_cast<uint32_t>(buffer.size()), context);
             };
         }
         else
@@ -34,15 +35,34 @@ IOCallback::IOCallback(IWSLAProcess* process, const WslcContainerProcessIOCallba
         m_io.AddHandle(std::make_unique<ReadHandle>(GetIOHandle(process, ioHandle), std::move(function)));
     };
 
-    addIOCallback(WSLC_PROCESS_IO_HANDLE_STDOUT, options.stdOutCallback, options.stdOutCallbackContext);
-    addIOCallback(WSLC_PROCESS_IO_HANDLE_STDERR, options.stdErrCallback, options.stdErrCallbackContext);
+    addIOCallback(WSLC_PROCESS_IO_HANDLE_STDOUT, options.onStdOut, options.callbackContext);
+    addIOCallback(WSLC_PROCESS_IO_HANDLE_STDERR, options.onStdErr, options.callbackContext);
 
-    m_io.AddHandle(std::make_unique<EventHandle>(m_cancelEvent.get()), MultiHandleWait::CancelOnCompleted);
+    if (options.onExit)
+    {
+        wil::unique_handle processExitEvent;
+        THROW_IF_FAILED(process->GetExitEvent(&processExitEvent));
+        m_io.AddHandle(std::make_unique<EventHandle>(std::move(processExitEvent)));
+    }
+
+    m_io.AddHandle(std::make_unique<EventHandle>(m_cancelEvent.get()), MultiHandleWait::CancelOnCompleted | MultiHandleWait::NeedNotComplete);
 
     m_thread = std::thread([this]() {
         try
         {
-            m_io.Run({});
+            // Will be false when cancelled.
+            bool runResult = m_io.Run({});
+
+            if (runResult && m_process && m_callbackOptions && m_callbackOptions->onExit)
+            {
+                WSLAProcessState state{};
+                int exitCode = -2;
+
+                // Prefer to make the callback even if we don't properly retrieve the exit code.
+                LOG_IF_FAILED(m_process->GetState(&state, &exitCode));
+
+                m_callbackOptions->onExit(exitCode, m_callbackOptions->callbackContext);
+            }
         }
         CATCH_LOG();
     });
@@ -69,7 +89,7 @@ bool IOCallback::HasIOCallback(const WslcContainerProcessOptionsInternal* option
 
 bool IOCallback::HasIOCallback(const WslcContainerProcessIOCallbackOptions& options)
 {
-    return options.stdOutCallback || options.stdErrCallback;
+    return options.onStdOut || options.onStdErr || options.onExit;
 }
 
 wil::unique_handle IOCallback::GetIOHandle(IWSLAProcess* process, WslcProcessIOHandle ioHandle)
--- a/src/windows/WslcSDK/IOCallback.cpp
+++ b/src/windows/WslcSDK/IOCallback.cpp
@@ -56,11 +56,21 @@ IOCallback::IOCallback(IWSLAProcess* process, const WslcContainerProcessIOCallba
             if (runResult && m_process && m_callbackOptions && m_callbackOptions->onExit)
             {
                 WSLAProcessState state{};
-                int exitCode = -2;
+                int exitCode = -1;
 
                 // Prefer to make the callback even if we don't properly retrieve the exit code.
-                LOG_IF_FAILED(m_process->GetState(&state, &exitCode));
+                if (FAILED_LOG(m_process->GetState(&state, &exitCode)))
+                {
+                    // Reset to our known value in case GetState stomped it while failing.
+                    exitCode = -1;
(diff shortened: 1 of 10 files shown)


src/windows/WslcSDK/IOCallback.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    m_thread = std::thread([this]() {
+        try
+        {
+            m_io.Run({});
+        }
+        catch (...)
+        {
+        }
+    });
```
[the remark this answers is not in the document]
JohnMcPMS: Leaving open for comments from WSL(A) experts on what mechanisms might already exist here.  It feels like the actual answer is going to be WIL error logging interactions from the `Run` internals are the best bet, at which point there is nothing to do here.

Yes, I could smuggle the exception pointer out and make a `Release` of some kind fail, but that feels like a burden on users.  Better would probably be some other mechanism for external observation, like another callback.  But ultimately I don't see this as a caller problem, so I don't feel like we should expose them to it directly like that.

src/windows/WslcSDK/IOCallback.cpp
```diff
@@ -0,0 +1,81 @@
...
+    Cancel();
+    if (m_thread.joinable())
+    {
+        m_thread.join();
+    }
+}
+
+void IOCallback::Cancel()
+{
+    m_io.Cancel();
```
OneBlue: I don't recommend calling `Cancel` from another thread. Even if that boolean is moved into a std::atomic, this can still lead to some pretty weird behaviors. 

Instead I would recommend having a wil::unique_event and signaling it from this thread when we want to cancel. 

Here's [an example on how to do that](https://github.com/microsoft/WSL/blob/e91443592d6320d5de9a5f8db4e86b7b2f8ffe38/src/windows/wslasession/WSLAContainer.cpp#L685) 

Then we could just do: 

```
void IOCallback::Cancel()
{
    m_cancelEvent.SetEvent();
    if (m_thread.joinable())
    {
        m_thread.join();
    }
```

src/windows/WslcSDK/IOCallback.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    Holds IO callback objects.
+
+--*/
+#include "precomp.h"
+#include "WslcsdkPrivate.h"
+
+IOCallback::IOCallback(IWSLAProcess* process, const WslcContainerProcessIOCallbackOptions& options)
+{
+    if (options.stdOutCallback)
```
OneBlue: Even if `stdOutCallback` is unset, we should drain its output (otherwise the pipe queue is going to fill up) (same for stderr)

src/windows/WslcSDK/IOCallback.cpp
```diff
@@ -0,0 +1,81 @@
...
+                localCallback(reinterpret_cast<const BYTE*>(buffer.data()), static_cast<uint32_t>(buffer.size()), localContext);
+            }));
+    }
+
+    m_thread = std::thread([this]() {
+        try
+        {
+            m_io.Run({});
+        }
+        catch (...)
```
OneBlue: Probably a good idea to have a CATCH_LOG() here

src/windows/WslcSDK/IOCallback.cpp
```diff
@@ -0,0 +1,82 @@
...
+    Cancel();
+    if (m_thread.joinable())
+    {
+        m_thread.join();
+    }
+}
+
+void IOCallback::Cancel()
+{
+    m_cancelEvent.SetEvent();
```
OneBlue, line 92: nit: Looks like this is never called separately, so might as well have everything in the destructor

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -914,19 +945,30 @@ try
 }
 CATCH_RETURN();
 
-STDAPI WslcSetProcessSettingsIoCallback(
-    _In_ WslcProcessSettings* processSettings, _In_ WslcProcessIoHandle ioHandle, _In_opt_ WslcStdIOCallback stdIOCallback, _In_opt_ PVOID context)
+STDAPI WslcSetProcessSettingsIOCallback(
```
OneBlue: I'm just realizing this now, but we need a way to unregister the IO callbacks and synchronize them. 

Something like: 

```

HRESULT WslcClearProcessIOCallbacks( WslcProcessSettings* processSettings, _In_ WslcProcessIOHandle ioHandle);
```

(Once that method returns, the caller is guaranteed that all pending callbacks are completed and that no future callback can be received).

JohnMcPMS: The current design is that releasing all of the process/container objects related to the process does that.  Container+all retrieved InitProcess instances for init processes, just the one process for any created afterward.

We could additionally have a way to explicitly do it, but I don't think that it is strictly necessary.  It would have to take in a `WslcProcess` handle, and we would need to consider if we wanted a second function to take in a container, removing the need to get the init process just to stop the IO callback.

A reason that this might be needed is to ensure that **ALL** output has been received (the buffers flushed) before teardown as the current release design cancels any future data flow.  If the process is still running that isn't really an issue.  If it has stopped, we know that there is an end to the pipe and we should get there in reasonably short order.

I think the caller flow would be like:
```
process = CreateContainerProcess(withIOCallbacks);
WaitForSingle(GetProcessExitEvent(process));
DisableIOCallbacksAndFlushBuffers(process, timeout); // without this, output could be lost in transit when the release is called
ReleaseProcess(process);
```
the implementation for (pseudoname) `DisableIOCallbacksAndFlushBuffers` would be like:
```
if (process is still running right now) // if you don't want this race, wait for the process exit event
{
  IOCancel();
}
if (!WaitForIOWithTimeout(timeout))
{
  IOCancel();
}
JoinThread();
```

OneBlue: Unfortunately this won't be sufficient because the service doesn't guarantee that all the pending IO has been written the pipe after the exit event is signaled. 

For context, two things happen when a process terminates

1) A zero byte read is signaled on the pipes
2) The process exit event is signaled

But those can happen in either order. 

Thinking through this, I think this might be something that we've overlooked when we designed this. The only way I can think of fixing this in a way that's not potentially confusing for the other would be something like: 

```
using WslcOnProcessExitedCallback = Function<void(int, Context)>;
using WslcStdIOCallback = Function<void(WslcProcessIOHandle, Context)>;

struct ProcessIOCallback
{
    WslcStdIOCallback OnStdout;
    WslcStdIOCallback OnStderr;
    WslcOnProcessExitedCallback OnExited;
}

STDAPI WslcSetProcessSettingsIOCallback(
_In_ WslcProcessSettings* processSettings,
 _In_ const ProcessIOCallback* Callbacks,
 _In_opt_ WslcStdIOCallback stdIOCallback,
 _In_opt_ PVOID context);

```

We could make this work by having `IOCallback` watch all of: 

1) The stdout handle
2) The stderr handle
3) The process exit event

And only call `OnExited()` once all of those are completed, which would guarantee that all IO has been flushed. 

Forcing the user to register all callbacks at once will push them to implement this correctly and only release the process after OnExited is called.

test/windows/WslcSdkTests.cpp
```diff
@@ -1386,6 +1387,363 @@ class WslcSdkTests
...
+
+        HANDLE exitEvent = nullptr;
+        VERIFY_SUCCEEDED(WslcGetProcessExitEvent(process.get(), &exitEvent));
+        VERIFY_ARE_EQUAL(WaitForSingleObject(exitEvent, 60 * 1000), static_cast<DWORD>(WAIT_OBJECT_0));
+
+        // Join the IOCallback thread before inspecting the accumulator.
+        process.reset();
+        container.reset();
+
+        VERIFY_IS_TRUE(stdoutData.size() >= c_expectedMinBytes);
```
OneBlue: What's the reason for `>=` here ? If it's the newlines, we can add `-w 0` to get rid of them

src/windows/WslcSDK/IOCallback.cpp
```diff
@@ -0,0 +1,102 @@
...
+            // Will be false when cancelled.
+            bool runResult = m_io.Run({});
+
+            if (runResult && m_process && m_callbackOptions && m_callbackOptions->onExit)
+            {
+                WSLAProcessState state{};
+                int exitCode = -2;
+
+                // Prefer to make the callback even if we don't properly retrieve the exit code.
+                LOG_IF_FAILED(m_process->GetState(&state, &exitCode));
```
OneBlue: if `option.onExit` is not set on the constructor, we can call `GetState` before the process actually terminates. 

To fix this, I recommend always waiting for the process exit event before completing the IO, and just skipping this callback if not set. 

We should probably fail if we get anything else than `WslaProcessStateExited` here

OneBlue: I recommend explicitly setting `exitCode` to -1 if GetState() fails here, since it can write `exitCode`

test/windows/WslcSdkTests.cpp
```diff
@@ -1386,6 +1387,363 @@ class WslcSdkTests
...
+    }
+
     // -----------------------------------------------------------------------
     // Stub tests for unimplemented (E_NOTIMPL) functions.
```
OneBlue, line 1819: nit: I recommend adding test coverage for the case where the caller releases the process / container reference while the process is still running and writing output

OneBlue (approved):
LGTM, minor comments

src/windows/WslcSDK/IOCallback.cpp
JohnMcPMS: But we cannot get to this code if `onExit` is not set (line 56 conditions).

Added assert for state.

JohnMcPMS: I chose a different number for the exit code than the value that it returns if the process is still running to disambiguate.  But I agree that if we start trying to encode data into that it becomes hard to keep it from stepping on the actual number space.

OneBlue: Good point ! I missed that

PR #14463, opened 2026-03-17
Moves all Ubuntu distros to the tar-based format

cnihelton:
This PR aims to move all Ubuntu LTSes to the new tar-based format, removing support for the versioned appx based distros.
The existing instances will continue working as expected, but new installations shouldn't fetch appx's from Microsoft Store. Instead, new installations are going to leverage the new tar based format, bringing consistency to all Ubuntu LTSes on WSL.
The default unversioned Ubuntu appx is preserved as is, for backwards compatibility, ensuring that the "Default" field works on both modern and legacy WSL setups.

## PR Checklist

## Validation Steps Performed
Used the resulting `DistributionInfo.json` as a replacement manifest by writing its path in the registry key `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Lxss` `DIstributionListURL` value, then installed `Ubuntu-22.04` and `Ubuntu-20.04`.

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -270,36 +270,6 @@
             "PackageFamilyName": "KaliLinux.54290C8133FEE_ey8k8hqnwqnmg"
         },
         {
-            "Name": "Ubuntu-20.04",
-            "FriendlyName": "Ubuntu 20.04 LTS",
-            "StoreAppId": "9MTTCL66CPXJ",
-            "Amd64": true,
-            "Arm64": true,
-            "Amd64PackageUrl": "https://publicwsldistros.blob.core.windows.net/wsldistrostorage/Ubuntu2004-230608_x64.appx",
-            "Arm64PackageUrl": "https://publicwsldistros.blob.core.windows.net/wsldistrostorage/Ubuntu2004-230608_ARM64.appx",
-            "PackageFamilyName": "CanonicalGroupLimited.Ubuntu20.04LTS_79rhkp1fndgsc"
-        },
-        {
-            "Name": "Ubuntu-22.04",
-            "FriendlyName": "Ubuntu 22.04 LTS",
-            "StoreAppId": "9PN20MSR04DW",
-            "Amd64": true,
-            "Arm64": true,
-            "Amd64PackageUrl": "https://publicwsldistros.blob.core.windows.net/wsldistrostorage/Ubuntu2204LTS-230518_x64.appx",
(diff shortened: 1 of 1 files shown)


OneBlue:
@CarlosNihelton: One side effect of doing this is that old Windows builds that don't have tar based distribution support will fail if the user runs: 

```
wsl --install
```

Since 'ubuntu' is the default. 
This will happen on builds that don't have the WSL package and still use the inbox version. 

These builds are technically out of support, but I would still like to avoid hard-breaking them. I can think of two solutions: 

1) Keep one 'ubuntu' appx image for backwards compatibility

2) Change the default to another distribution (note that 'Default' applies to both tar based and appx distros)



@craigloewen-msft  FYI

cnihelton:
> @CarlosNihelton: One side effect of doing this is that old Windows builds that don't have tar based distribution support will fail if the user runs:
> 
> ```
> wsl --install
> ```
> 
> Since 'ubuntu' is the default. This will happen on builds that don't have the WSL package and still use the inbox version.
> 
> These builds are technically out of support, but I would still like to avoid hard-breaking them. I can think of two solutions:
> 
>     1. Keep one 'ubuntu' appx image for backwards compatibility
> 
>     2. Change the default to another distribution (note that 'Default' applies to both tar based and appx distros)
> 
> 

Hi @OneBlue ! Thanks for the quick notice!

If I got you correctly, we should be fine as the unversioned Ubuntu appx is preserved as is (no breaking changes for `wsl --install` on inbox WSL), right? https://github.com/microsoft/WSL/pull/14463/changes#diff-37fca1278e3439911b7860efe087b392e165e11b29fbeeb21035a9402cf134c4L217

Let me rephrase my PR description, as I'm only touching the versioned apps.

Yet, it's worth raising the question: until when do we believe the appx will still be necessary?

cnihelton:
> Yet, it's worth raising the question: until when do we believe the appx will still be necessary?

We can discuss this topic later if there is no clear answer as of now.

OneBlue:
> Let me rephrase my PR description, as I'm only touching the versioned apps.

Ah indeed I misread the diff and assumed that all the appx ubuntu distros were being moved. If we're keeping the default unversionned appx, then I'm not worried about this. 


> Yet, it's worth raising the question: until when do we believe the appx will still be necessary?

That's a great question ! I don't have a clear answer to it right now. If this ever become a problem to maintain on the Ubuntu side, we should probably discuss this with @craigloewen-msft

OneBlue (approved):

benhillis (approved):

PR #14470, opened 2026-03-18
cleanup: remove wslaservice and combine with wslservice

benhillis:
This change combines wslaservice functonality into wslservice.exe. Previously these were isolated because we were planning on shipping these things separately, but now the plan is to ship wsla functionality as a WSL feature so they can be combined.

This is primarily a move, in a future change we can do a wsla -> wslc rename.


benhillis:
All three Copilot comments are false positives

src/windows/service/exe/LxssUserSessionFactory.cpp
```diff
@@ -35,43 +36,43 @@ std::optional<wsl::windows::service::PluginManager> g_pluginManager;
...
-        // This is required because the moved-from std::optional<T> isn't made empty, so this needs to be done explicitly.
-        g_sessions.reset();
+            // Unload plugins
+            g_pluginManager.reset();
+        }
     }
 
-    ClearSessionsAndBlockNewInstancesLockHeld(sessions);
+    // Also tear down WSLA sessions.
+    wsl::windows::service::wsla::ClearWslaSessionsAndBlockNewInstances();
```
OneBlue: I don't think we always want to do that for WSLA, since ClearSessionsAndBlockNewInstances() is also called when the WSL policy is changed. 

For now I think we should just keep these into two separate methods, and update the logic once we have the registry key for the WSLA policies checked in

OneBlue (approved):
LGTM. One minor comment

src/windows/service/exe/LxssUserSessionFactory.cpp
benhillis: good call I forgot this gets used by the session policy.

tools/test/setup-vm-for-tests.ps1
```diff
@@ -139,6 +139,7 @@ Invoke-Command -Session $Session -ArgumentList $RemoteFolder -ScriptBlock {
 }
 Copy-Item -ToSession $Session -Path "$Bin/installer.msix"  -Destination $RemoteFolder -Force
 Copy-Item -ToSession $Session -Path "$Bin/wsltests.dll"  -Destination $RemoteFolder -Force
+Copy-Item -ToSession $Session -Path "$Bin/wslcsdk.dll"  -Destination $RemoteFolder -Force
```
benhillis, line 142: unrelated but needed to run the WSLC tests in a VM.

OneBlue (approved):

PR #14477, opened 2026-03-19
test: disable LoopbackExplicit due to OS build 29555 regression

benhillis:
This test started failing as soon as our CloudTest images were updated to the 3/17 re_prerelease build.


OneBlue (approved):

keith-horton:
@benhillis , yep, we see the cause and are working to fix it.

Issue #13985, opened 2025-12-25
UID user collision between different WSL distributions


rdbisme:
### Windows Version

Microsoft Windows [Version 10.0.26220.7344]

### WSL Version

2.6.2.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.87.2-1

### Distro Version

Archlinux

### Other Software

systemd 259-1

### Repro Steps

- Enable systemd in my Arch WSL
-  Start `podman machine start`
- Start my work WSL
- My Arch work WSL can't start systemd because of UID collision with the user UID of the podman VM

### Expected Behavior

I can start the two VMs independently. I'd argue that two VMs should not interfere between each other.

### Actual Behavior

The `user@1000.service` can't start because it's already running in the other VM. Why are they conflicting? 

```
wsl: Failed to start the systemd user session for 'rdb'. See journalctl for more details.
```

Changing the uid manually in one of the distributions solves the problem, but I'd assume the distributions would be isolated? 

### Diagnostic Logs

```
Dec 25 15:51:47 CiD unknown: WSL (2 - Interop) ERROR: operator():2755: systemctl is-active user@1000.service returned: failed
Dec 25 15:51:47 CiD kernel: misc dxg: dxgk: dxgkio_is_feature_enabled: Ioctl failed: -22
Dec 25 15:51:47 CiD kernel: misc dxg: dxgk: dxgkio_query_adapter_info: Ioctl failed: -22
Dec 25 15:51:47 CiD kernel: misc dxg: dxgk: dxgkio_query_adapter_info: Ioctl failed: -22
Dec 25 15:51:47 CiD kernel: misc dxg: dxgk: dxgkio_query_adapter_info: Ioctl failed: -22
Dec 25 15:51:47 CiD kernel: misc dxg: dxgk: dxgkio_query_adapter_info: Ioctl failed: -2
Dec 25 15:51:47 CiD kernel: misc dxg: dxgk: dxgkio_is_feature_enabled: Ioctl failed: -22
Dec 25 15:51:47 CiD kernel: misc dxg: dxgk: dxgkio_query_adapter_info: Ioctl failed: -22
Dec 25 15:51:47 CiD kernel: misc dxg: dxgk: dxgkio_query_adapter_info: Ioctl failed: -22
Dec 25 15:51:47 CiD kernel: misc dxg: dxgk: dxgkio_query_adapter_info: Ioctl failed: -22
[...]

PR #14479, opened 2026-03-19
Fix UID collision between WSL distributions with systemd enabled

shuaiyuanxx:

Fix UID collision between WSL distributions when multiple distros with systemd enabled share the same UID (e.g. UID 1000). When a podman machine and a user distro run simultaneously, the second distro's systemd fails with `Failed to attach to cgroup /user.slice/user-1000.slice/user@1000.service: Device or resource busy` because all distros share the same global cgroup hierarchy.

This PR isolates each distro's cgroup hierarchy by creating a per-distro cgroup under `/sys/fs/cgroup/wsl/<distro-name>`, moving the init process into it, and unsharing the cgroup namespace before chroot. This allows each distro's systemd to independently manage `user@<UID>.service` without conflicts.

## PR Checklist

- [x] **Closes:** #13985

### Root Cause

All WSL2 distros run inside a single lightweight VM sharing one Linux kernel. Each distro gets its own PID, mount, and UTS namespaces via `clone()`, but **not** a cgroup namespace. When two distros with the same UID (default 1000) both enable systemd, the second distro's systemd tries to attach to the same cgroup path (`/user.slice/user-1000.slice/user@1000.service`) already occupied by the first, causing exit code 219/CGROUP.

### Fix

A new function `IsolateCgroupNamespace()` is added and called from `LaunchInit()` before `Chroot()`:

1. **cgroup v2 detection**: Checks for `/sys/fs/cgroup/cgroup.controllers` to skip cgroup v1 (where this issue does not manifest).
2. **Per-distro cgroup creation**: Creates `/sys/fs/cgroup/wsl/<DistributionName>/`.
3. **Controller delegation**: Reads available controllers from `cgroup.controllers` and enables them via `cgroup.subtree_control` on intermediate cgroups, so systemd can manage resources (cpu, memory, io, pids, etc.).
4. **Process migration**: Writes current PID to the per-distro `cgroup.procs`.
5. **Namespace unshare**: Calls `unshare(CLONE_NEWCGROUP)` so the per-distro cgroup becomes the namespace root.

After this, systemd inside each distro sees an isolated cgroup tree. For example, Ubuntu-22.04's `user@1000.service` cgroup resolves to `/sys/fs/cgroup/wsl/Ubuntu-22.04/user.slice/user-1000.slice/user@1000.service` in the global hierarchy, while podman-machine-default's resolves under `/sys/fs/cgroup/wsl/podman-machine-default/...`.

### Design Decisions

- **Placement in `LaunchInit()`**: This is the single function called by all code paths (GUI, non-GUI, system distro), ensuring complete coverage. It runs before `Chroot()` when `/sys/fs/cgroup` is still accessible.
- **`unshare()` instead of `CLONE_NEWCGROUP` in clone flags**: Avoids modifying the existing `clone()` call. The isolation is done in the child process that will become the distro init, so the parent process is unaffected.
- **Graceful degradation**: Every step logs warnings on failure but does not prevent the distro from starting. This preserves backwards compatibility on kernels that may not support `CLONE_NEWCGROUP` or on cgroup v1 configurations.
- **No changes to clone flags**: The original `CLONE(CLONE_NEWNS | CLONE_NEWPID | CLONE_NEWUTS | SIGCHLD)` is untouched.

## Validation Steps Performed

1. **Reproduction confirmed** (before fix):
   - `wsl --shutdown`
   - `podman machine start` (creates a WSL distro with UID 1000 and systemd)
   - `wsl -d Ubuntu-22.04` -> **Failed**: `wsl: Failed to start the systemd user session for 'shuaiyuan'`
   - journalctl showed: `Failed to attach to cgroup /user.slice/user-1000.slice/user@1000.service: Device or resource busy`

2. **Fix validated** (after deploying patched initrd.img):
   - `wsl --shutdown`
   - `podman machine start`
   - `wsl -d Ubuntu-22.04` -> **Success**: no error
   - `systemctl is-active user@1000.service` -> `active`
   - `systemctl --user status` -> `State: running`
   - Cgroup namespace inode differs between distros (confirmed isolation)
   - Two Ubuntu distros (20.04 + 22.04) with podman machine all running simultaneously with UID 1000, no conflicts


shuaiyuanxx:
Closing this PR as the current approach, while functional, is a workaround rather than a proper fix. 

The root cause is an architectural gap in WSL: cgroup isolation between distributions should be guaranteed by the WSL infrastructure layer before a distribution's init is ever exec'd, not negotiated by the init process itself. 
Getting this right requires more careful design at the infrastructure level.

PR #14482, opened 2026-03-19
Fix StateChangedAt mismatch during container recovery from storage

beena352:
Fixed a flaky test failure in ContainerRecoveryFromStorage where StateChangedAt differed by 1 second between the live and recovery paths. The root cause was that Transition() used std::time(nullptr) (host clock) while Open() restored from Docker's FinishedAt (VM clock), producing different values.

## PR Checklist

When a container is stopped, Transition() recorded m_stateChangedAt using std::time(nullptr) (the Windows host clock). When the container was later recovered via Open(), the timestamp was restored from Docker's FinishedAt field (the VM clock). These two clocks could differ by ~1 second, causing the ContainerRecoveryFromStorage test assertion AreEqual(containers[0].StateChangedAt, originalStateChangedAt) to fail intermittently.
- Added an optional stateChangedAt parameter to Transition() with a default of std::nullopt (falls back to std::time(nullptr), preserving existing behavior for all other callers).
- In Stop(), use a std::promise/std::future to receive the stop timestamp from the Docker event stream, avoiding an InspectContainer() HTTP call that could fail during session teardown. OnEvent(Stop) fulfills the promise with the event timestamp when a Stop() call is in progress; otherwise it handles the transition directly.
- Both the live path (event timestamp) and recovery path (InspectContainer/FinishedAt) now use timestamps from Docker's clock, eliminating the host-vs-VM clock mismatch.
-Added process cleanup (OnContainerReleased + clear) in Stop() to match OnEvent() behavior, preventing exec'd processes from remaining stuck after a Stop()-initiated shutdown.

## Validation Steps Performed
Ran ContainerRecoveryFromStorage test repeatedly, no longer fails.


src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -597,7 +597,20 @@ void WSLAContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCod
...
+            // Use Docker's FinishedAt so the event path matches the
+            // value restored during recovery in Open().
+            std::optional<std::uint64_t> finishedAt;
+            try
+            {
+                auto inspectData = m_dockerClient.InspectContainer(m_id);
+                if (!inspectData.State.FinishedAt.empty())
+                {
+                    finishedAt = ParseDockerTimestamp(inspectData.State.FinishedAt);
+                }
```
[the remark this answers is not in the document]
beena352: ParseDockerTimestamp() was already there and is used in Open() for the recovery path. If it couldn’t handle Docker’s FinishedAt, recovery would already be failing. Happy to harden the parser in a follow-up if needed, but that’s out of scope for this change.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -667,7 +680,20 @@ void WSLAContainerImpl::Stop(WSLASignal Signal, LONG TimeoutSeconds)
...
+    {
+        auto inspectData = m_dockerClient.InspectContainer(m_id);
+        if (!inspectData.State.FinishedAt.empty())
+        {
+            finishedAt = ParseDockerTimestamp(inspectData.State.FinishedAt);
+        }
+    }
+    CATCH_LOG();
+
+    Transition(WslaContainerStateExited, finishedAt);
```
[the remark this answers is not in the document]
beena352: That makes sense. I looked into using the event timestamp, but it only gives when the event was emitted, not when the container actually exited. We use Docker’s FinishedAt in Open(), so to keep things consistent we need to use the same value here as well.
Since the event doesn’t include FinishedAt, we have to call InspectContainer(). If we skip it and use the host time instead, we can hit the same mismatch again.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -597,7 +597,20 @@ void WSLAContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCod
         // This can happen if Delete() is called by the user.
         if (previousState == WslaContainerStateRunning)
         {
-            Transition(WslaContainerStateExited);
+            // Use Docker's FinishedAt so the event path matches the
+            // value restored during recovery in Open().
+            std::optional<std::uint64_t> finishedAt;
+            try
+            {
+                auto inspectData = m_dockerClient.InspectContainer(m_id);
```
OneBlue: Calling Inspect() here might be a bit risky because we might be in situations where the session is terminating and the call will fail. 

Instead of using Inspect(), could we update the `ContainerEventTracker` to pass in timestamps ? For instance in the case the container stopping, we could use the `time` field. Sample event: 

```
{""""status"""":""""die"""",""""id"""":""""c7731122e63a54b7b91be45f7307f1bc5ea951bcf54fca5d443abf6b2fa06773"""",""""from"""":""""nginx:latest"""",""""Type"""":""""container"""",""""Action"""":""""die"""",""""Actor"""":{""""ID"""":""""c7731122e63a54b7b91be45f7307f1bc5ea951bcf54fca5d443abf6b2fa06773"""",""""Attributes"""":{""""com.microsoft.wsl.container.metadata"""":""""{\""""V1\"""":{\""""Flags\"""":0,\""""InitProcessFlags\"""":3,\""""Ports\"""":[],\""""Volumes\"""":[]}}"""",""""execDuration"""":""""13"""",""""exitCode"""":""""0"""",""""image"""":""""nginx:latest"""",""""maintainer"""":""""NGINX Docker Maintainers \u003cdocker-maint@nginx.com\u003e"""",""""name"""":""""happy_bardeen""""}},""""scope"""":""""local"""",""""time"""":1773945390,""""timeNano"""":1773945390741964813} ""
``` 

We could provide this timestamp to the caller by adding a uint64_t argument to ContainerStateChangeCallback

beena352: Yes, I can update ContainerEventTracker to parse the time field from the Docker event JSON and pass it through the callback, so OnEvent() can use that timestamp directly and avoid calling InspectContainer().

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -597,7 +597,7 @@ void WSLAContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCod
         // This can happen if Delete() is called by the user.
         if (previousState == WslaContainerStateRunning)
         {
-            Transition(WslaContainerStateExited);
+            Transition(WslaContainerStateExited, eventTime);
```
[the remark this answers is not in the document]
beena352: We intentionally use the event timestamp in OnEvent() to avoid calling Inspect() in that path, per earlier feedback. Using GetDockerFinishedAt() here would reintroduce the inspect call.

src/windows/wslasession/ContainerEventTracker.cpp
```diff
@@ -147,11 +147,14 @@ void ContainerEventTracker::OnEvent(const std::string_view& event)
 
     std::lock_guard lock{m_lock};
 
+    auto timeEntry = parsed.find("time");
+    std::uint64_t eventTime = timeEntry != parsed.end() ? timeEntry->get<std::uint64_t>() : static_cast<std::uint64_t>(std::time(nullptr));
```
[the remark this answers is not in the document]
beena352: We treat time the same way as exitCode and execID, optional and checked with find. Falling back to std::time(nullptr) is better than skipping the event, since that could miss state transitions. In practice, Docker always includes time, so this is just a defensive fallback

OneBlue: Are there situations were `time` is not set ? If so i'd always assign it here and fail if it's absent

beena352: Good point, Docker events always include time, so updated this to treat it as required and fail if it’s missing.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -1399,7 +1399,22 @@ __requires_exclusive_lock_held(m_lock) void WSLAContainerImpl::DisconnectComWrap
...
+    try
+    {
+        auto inspectData = m_dockerClient.InspectContainer(m_id);
+        if (!inspectData.State.FinishedAt.empty())
+        {
+            return ParseDockerTimestamp(inspectData.State.FinishedAt);
+        }
+    }
+    CATCH_LOG();
+
```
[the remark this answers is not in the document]
beena352: CATCH_LOG() already covers the InspectContainer() failure case. This helper is only used from Stop(), after StopContainer() has already succeeded, so we’d expect FinishedAt to be available here. The nullopt path is just a defensive fallback, and I’d prefer not to add extra logging for that in this change.

src/windows/wslasession/WSLAContainer.cpp
```diff
@@ -612,7 +612,7 @@ void WSLAContainerImpl::Stop(WSLASignal Signal, LONG TimeoutSeconds)
         }
     }
 
-    Transition(WslaContainerStateExited);
+    Transition(WslaContainerStateExited, GetDockerFinishedAt());
```
OneBlue: Hmm doing this might be a bit risky, because if the session is terminating, this call could fail and then we wouldn't call `ReleaseRuntimeResources()`.

Here's an alternative idea: We could rethink Stop() to always wait for the stop notification to be triggered, that way we always get the correct stop timestamp. We might need to rethink the locking story a bit though, because `OnEvent` currently acquires `m_lock` as well

The easiest way I can think of would be have a waitable "stop state". Something like this: 

```
std::mutex m_stopStateLock;
std::optional<std::promise<uint64_t>> m_stopState;
```

Then in Stop() we could do something like this: 

```

    {
        std::mutex m_stopStateLock;
        m_stopState.emplace();
    }

    // Perform the stop
    [..]

    auto stopTimestamp = m_stopState.get_future().wait_for(...)
    Transition(WslaContainerStateExited, topTimestamp);

```

With this, `OnEvent()` can check if `m_stopState` is set and just set the value if so (since there's a Stop call() in progress) without having to acquire `m_lock` 

Let me know if that makes sense

beena352: Yes, that makes sense. The promise/future approach is definitely cleaner, it avoids calling InspectContainer() during teardown and keeps the timestamp handling consistent.
I’ll also make sure that OnEvent() doesn’t call set_value more than once, likely by resetting the optional after it’s consumed.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -560,13 +560,25 @@ void WSLAContainerImpl::Start(WSLAContainerStartFlags Flags, LPCSTR DetachKeys)
...
+
+        // If a Stop() call is in progress, provide the timestamp via the promise
+        // and let Stop() handle the state transition.
+        {
+            std::lock_guard stopLock{m_stopStateLock};
+            if (m_stopState.has_value())
+            {
+                m_stopState->set_value(eventTime);
+                return;
+            }
```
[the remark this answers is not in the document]
beena352: Docker sends one stop event per stop, so set_value() is called at most once. If it throws, we should let it fail.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -560,13 +560,25 @@ void WSLAContainerImpl::Start(WSLAContainerStartFlags Flags, LPCSTR DetachKeys)
...
 
         THROW_HR_IF(E_UNEXPECTED, !exitCode.has_value());
+
+        // If a Stop() call is in progress, provide the timestamp via the promise
+        // and let Stop() handle the state transition.
+        {
+            std::lock_guard stopLock{m_stopStateLock};
+            if (m_stopState.has_value())
+            {
+                m_stopState->set_value(eventTime);
```
OneBlue, line 573: Since we're holding `m_stopStateLock` I would recommend clearing `m_stopState` right after we set the value, that way if another callback arrived while Stop() is still running (and hasn't cleared `m_stopState` yet), we don't risk signaling this twice

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -657,7 +685,26 @@ void WSLAContainerImpl::Stop(WSLASignal Signal, LONG TimeoutSeconds)
...
+    std::optional<std::uint64_t> stopTimestamp;
+    if (stopFuture.wait_for(60s) == std::future_status::ready)
+    {
+        stopTimestamp = stopFuture.get();
+    }
+
+    Transition(WslaContainerStateExited, stopTimestamp);
+
+    {
+        std::lock_guard processesLock{m_processesLock};
```
OneBlue: nit: We could extract this logic into a method and factor it with that OnEvent() does

OneBlue (approved):
LGTM, minor comments

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -560,13 +560,25 @@ void WSLAContainerImpl::Start(WSLAContainerStartFlags Flags, LPCSTR DetachKeys)
     cleanup.release();
 }
 
-void WSLAContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCode)
+void WSLAContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCode, std::uint64_t eventTime)
 {
     if (event == ContainerEvent::Stop)
     {
         m_stoppedNotifiedEvent.SetEvent();
```
OneBlue: We can remove this event now since we don't need it anymore

src/windows/wslcsession/ContainerEventTracker.cpp
```diff
@@ -145,13 +145,18 @@ void ContainerEventTracker::OnEvent(const std::string_view& event)
...
+    std::uint64_t eventTime = timeEntry->get<std::uint64_t>();
+
     std::lock_guard lock{m_lock};
 
     for (const auto& e : m_callbacks)
     {
         if (e.ContainerId == containerId && (!e.ExecId.has_value() || e.ExecId == execId))
         {
-            e.Callback(it->second, exitCode);
+            e.Callback(it->second, exitCode, eventTime);
```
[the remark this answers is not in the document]
beena352: This is pre-existing behavior and not introduced by this PR. My change only adds the eventTime parameter to the callback, the iteration and callback dispatch logic remain unchanged.

PR #14484, opened 2026-03-19
Split image into repo and tag arguments when relevant

OneBlue:

This change splits the single "image" string into two "repo" and "tags" arguments when pulling or importing images. 

This allows us to easily default to "latest" in these operations, and have the correct behavior when doing something like: 

```
$ wslc pull debian # Now behaves as 'wslc pull debian:latest'
```

## PR Checklist


src/windows/service/inc/wslaservice.idl
```diff
@@ -512,10 +512,10 @@ interface IWSLASession : IUnknown
     HRESULT GetState([out] WSLASessionState* State);
 
     // Image management.
-    HRESULT PullImage([in] LPCSTR ImageUri, [in, unique] const  WslaRegistryAuthInformation* RegistryAuthenticationInformation, [in, unique] IProgressCallback* ProgressCallback);
+    HRESULT PullImage([in] LPCSTR Repo, [in] LPCSTR Tag, [in, unique] const  WslaRegistryAuthInformation* RegistryAuthenticationInformation, [in, unique] IProgressCallback* ProgressCallback);
```
kevpar, line 515: I think we should keep it so `PullImage` accepts just an image name, which is then parsed into repo and tag. Partly this is to make it so clients don't need to do the parsing themselves, but also because we need to support pulling an image by digest (e.g. `docker pull ubuntu@sha256:66460d557b25769b102175144d538d88219c077c678a49af4afca6fbfc1b5252`). You could make `PullImage` takes a repo, tag, and digest, as an alternative.

kevpar: [This comment](https://github.com/microsoft/WSL/pull/14484#discussion_r2963577980) from Copilot also touches on some of the parsing particularities. I don't know if there is a spec that defines exactly what form an image string can take, unfortunately.

OneBlue: Hmm that's a good point. The main reason for this change was that today if you run `wslc pull debian`, we pull every single image in the debian repo, which is obviously not the behavior that we want.

I'll rethink this. Maybe we can make this work by splitting the image into repo & tag on the service side and defaulting to `latest` if no tag is passed. 

Sadly I don't think there's a way around us having to parse repo & tags

OneBlue: This looks like a good reference for an image parser: https://deepwiki.com/openshift/docker-distribution/9-image-references

OneBlue:
Superseeded by https://github.com/microsoft/WSL/pull/14496

PR #14272, opened 2026-02-25
Mask NetworkManager-wait-online.service during boot

dhtzs:
## Summary of the Pull Request
Fixes #13772. Similar to PR #13611, this masks NetworkManager-wait-online.service to prevent 60-second timeouts during boot since WSL interfaces are unmanaged by NetworkManager. Also added the service to the discouraged units list in validate-modern.py and added a unit test.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/13772
- [x] **Communication:** I've discussed this with core contributors already. If work hasn't been agreed, this work might be rejected
- [x] **Tests:** Added/updated if needed and all pass
- [ ] **Localization:** All end user facing strings can be localized
- [ ] **Dev docs:** Added/updated if needed
- [ ] **Documentation updated:** If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/wsl/) and link it here: #xxx

PR #14490, opened 2026-03-20
Mask console-getty.service to prevent multi-distro failures (#13595)

benhillis:
When multiple WSL distros run concurrently, /dev/tty devices are shared at the VM level. The second distro's console-getty.service fails because the tty is already held by the first, causing systemd to report failed units and triggering user@UID.service failures.

Mask console-getty.service during WSL systemd unit generation, similar to the existing masking of networkd-wait-online. This service provides no value in WSL since users don't connect to the underlying tty.

Fixes #13595

### DIFF
--- a/src/linux/init/init.cpp
+++ b/src/linux/init/init.cpp
@@ -336,6 +336,10 @@ int GenerateSystemdUnits(int Argc, char** Argv)
         // Mask NetworkManager-wait-online.service for the same reason, as it causes timeouts on distros using NetworkManager.
         THROW_LAST_ERROR_IF(symlink("/dev/null", std::format("{}/NetworkManager-wait-online.service", installPath).c_str()) < 0);
 
+        // Mask console-getty.service since /dev/tty devices are shared at the VM level across all distros.
+        // When multiple distros are running, the second distro's getty fails because the tty is already held.
+        THROW_LAST_ERROR_IF(symlink("/dev/null", std::format("{}/console-getty.service", installPath).c_str()) < 0);
+
         // Only create the wslg unit if both enabled in wsl.conf, and if the wslg folder actually exists.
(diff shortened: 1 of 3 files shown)


dhtzs:
Should `console-getty.service` be added to the discouraged units list in `validate-modern.py` as well? (Similar to what was done in #14272)

benhillis:
> Should `console-getty.service` be added to the discouraged units list in `validate-modern.py` as well? (Similar to what was done in #14272)

That's a really good call.

src/linux/init/init.cpp
```diff
@@ -336,6 +336,10 @@ int GenerateSystemdUnits(int Argc, char** Argv)
         // Mask NetworkManager-wait-online.service for the same reason, as it causes timeouts on distros using NetworkManager.
         THROW_LAST_ERROR_IF(symlink("/dev/null", std::format("{}/NetworkManager-wait-online.service", installPath).c_str()) < 0);
 
+        // Mask console-getty.service since /dev/tty devices are shared at the VM level across all distros.
+        // When multiple distros are running, the second distro's getty fails because the tty is already held.
+        THROW_LAST_ERROR_IF(symlink("/dev/null", std::format("{}/console-getty.service", installPath).c_str()) < 0);
```
cnihelton: Did you consider masking the `systemd-getty-generator` instead? The `console-getty.service` is a dynamic unit created at startup time by that process. By masking the generator we save a few extra CPU cycles.

[systemd docs about that](https://www.freedesktop.org/software/systemd/man/latest/systemd-getty-generator.html)

OneBlue (approved):

mhegreberg:
thanks! we mitigate this in Arch Linux, but it'll be nice to have the patch across all distros now.

PR #14493, opened 2026-03-20
Fix potential pipe deadlock when wslc's stdin doesn't support overlapped IO

OneBlue:

This change fixes a potential pipe deadlock that can happen where stdin is a pipe that doesn't support overlapped IO. 

This can be reproduced by: 

`cat | wslc run -i debian:latest cat`

## PR Checklist


dkbennett (approved):
I cherry picked and confirmed this fix also fixes the issue I had in E2E interactive tests for non-tty :shipit:

PR #14496, opened 2026-03-21
Rethink image parsing logic

OneBlue:

This change fixes various issues with PullImage() not correctly handling the reference format. Most notably: 

` wslc pull debian` would pull every single image in the `debian` repo. Now the tag is defaulted to `latest` if not specified. 

## PR Checklist


kevpar:
Does this supersede #14484 ?

OneBlue:
> Does this supersede #14484 ?

It does indeed !

src/windows/wslasession/WSLASession.cpp
```diff
@@ -574,7 +566,7 @@ try
     RETURN_HR_IF_NULL(E_POINTER, ImageName);
     RETURN_HR_IF(E_INVALIDARG, strlen(ImageName) > WSLA_MAX_IMAGE_NAME_LENGTH);
 
-    auto [repo, tag] = ParseImage(ImageName);
+    auto [repo, tag] = wslutil::ParseImage(ImageName);
 
     THROW_HR_IF_MSG(E_INVALIDARG, !tag.has_value(), "Expected tag for image import: %hs", ImageName);
```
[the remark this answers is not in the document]
OneBlue: Is is. See: https://docs.docker.com/reference/api/engine/version/v1.52/#tag/Image/operation/ImageCreate

src/windows/wslasession/WSLASession.cpp
```diff
@@ -314,7 +301,12 @@ try
 
     RETURN_HR_IF_NULL(E_POINTER, ImageUri);
 
-    auto [repo, tag] = ParseImage(ImageUri);
+    auto [repo, tag] = wslutil::ParseImage(ImageUri);
```
kevpar: Suggest renaming `tag` to `tagOrDigest`, and do the same in the Docker client `PullImage` signature.

src/windows/wslasession/WSLASession.cpp
```diff
@@ -575,7 +567,7 @@ try
     RETURN_HR_IF_NULL(E_POINTER, ImageName);
     RETURN_HR_IF(E_INVALIDARG, strlen(ImageName) > WSLA_MAX_IMAGE_NAME_LENGTH);
 
-    auto [repo, tag] = ParseImage(ImageName);
```
kevpar, line 578: Same here

kevpar:
As an aside, I noticed we are implicitly prepending `library` here: https://github.com/microsoft/WSL/blob/feature/wsl-for-apps/src/windows/wslasession/DockerHTTPClient.cpp#L124

I don't think this is correct to do other than for Docker Hub images. Has image pull been tested with other registries?

OneBlue:
> As an aside, I noticed we are implicitly prepending `library` here: https://github.com/microsoft/WSL/blob/feature/wsl-for-apps/src/windows/wslasession/DockerHTTPClient.cpp#L124
> 
> I don't think this is correct to do other than for Docker Hub images. Has image pull been tested with other registries?

We don't support custom registries yet, but added a TODO to track that.

kevpar (approved):

benhillis (approved):

Issue #13979, opened 2025-12-24
wsl --uninstall should warn users it's **not** for removing distribution.


MikuAuahDark:
**Is your feature request related to a problem? Please describe.**
`wsl --install Distro` is the command-line option to install distribution. So that means user expects `wsl --uninstall Distro` is the command-line to uninstall distribution.

Unfortunately, it's not. `wsl --uninstall` regardless if you add the distribution name or not, removes the **whole** WSL installation. Removing distribution is `wsl --unregister Distro` instead. I think this is bit confusing.

**Describe the solution you'd like**
I can think of 2 solutions:
1. Make `wsl --uninstall` warn user of removing WSL entirely instead of removing distribution. Add `-y` to force-confirm.
2. Make `wsl --uninstall Distro`  an alias to `wsl --unregister Distro`.

Both solution is not mutually exclusive, so either one of them or both can be implemented.

**Describe alternatives you've considered**
The alternative is to read the command-line options.

**Additional context**
From user perspective, it makes sense to think like this:
1. Installing distribution in WSL is `wsl --install Distro`
2. So, the opposite of `--install` is `--uninstall`.
3. That means `wsl --uninstall Distro` should be the command to uninstall a distribution.

Fotunately no data loss is observed when running `wsl --uninstall` followed by `wsl --install Distro`. At least, the existing distribution is still there.

PR #14507, opened 2026-03-23
Fix: Fail and warn the user when --uninstall is given parameters

chemwolf6922:

Fail and warn the user when --uninstall is given parameters. The user is likely misusing the command to uninstall the distro instead of uninstall WSL.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/13979

## Validation Steps Performed

* Tested on my local system with `wsl --uninstall Ubuntu`
* Tested with the newly added unit test.


localization/strings/cs-CZ/Resources.resw
```diff
@@ -960,6 +960,10 @@ Návrat k sítím NAT.</value>
     <value>Aktualizace se nezdařila (ukončovací kód: {}).</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
+  <data name="MessageUninstallNoArguments" xml:space="preserve">
```
benhillis, line 963: How did you generate these localization files? You only need to modify en-US and the rest are auto-generated.

benhillis (changes requested):

localization/strings/cs-CZ/Resources.resw
chemwolf6922: They were AI generated. Reverted non-en-US resource files.

chemwolf6922:
Github seems to stuck at this state forever...
image
If I'm still seeing this tomorrow. I'll create a new PR instead.

chemwolf6922:
Closing this one as github is stuck.

PR #14523, opened 2026-03-24
Fix singular/plural grammar in FormatRelativeTime

beena352:
Fixes grammatically incorrect output in FormatRelativeTime where singular time units were always displayed as plural (e.g., "1 seconds ago", "1 minutes ago"). Now correctly outputs "1 second ago", "1 minute ago", "1 hour ago", "1 day ago".

## PR Checklist

ContainerService::FormatRelativeTime was using hardcoded plural unit strings for all time values. This change introduces a ternary check for each time unit seconds, minutes, hours, and days to produce grammatically correct output.

### DIFF
--- a/src/windows/wslc/services/ContainerService.cpp
+++ b/src/windows/wslc/services/ContainerService.cpp
@@ -165,18 +165,22 @@ std::wstring ContainerService::FormatRelativeTime(ULONGLONG timestamp)
 
     if (elapsed < SecondsPerMinute)
     {
-        return std::format(L"{} seconds ago", elapsed);
+        const auto seconds = elapsed;
+        return std::format(L"{} {} ago", seconds, (seconds == 1 ? L"second" : L"seconds"));
     }
     else if (elapsed < SecondsPerHour)
     {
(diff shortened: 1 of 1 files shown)


OneBlue (approved):
LGTM. We should probably add test coverage to cover these, but we can do that in a followup PR

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -165,18 +165,22 @@ std::wstring ContainerService::FormatRelativeTime(ULONGLONG timestamp)
 
     if (elapsed < SecondsPerMinute)
     {
-        return std::format(L"{} seconds ago", elapsed);
+        const auto seconds = elapsed;
+        return std::format(L"{} {} ago", seconds, (seconds == 1 ? L"second" : L"seconds"));
```
benhillis, line 169: minor, can we do this instead?
```
std::format(L"{} {}{} ago", seconds, L"second", (seconds == 1 ? L"" : L"s"));
```

benhillis: This is fine for now since this needs to be replaced by localized strings anyway...

PR #14539, opened 2026-03-25
Implement SDK CanRun

JohnMcPMS:
Implements `WslcCanRun` and changes the observed error when the `WSLASessionManager` class is not registered.

## PR Checklist

Implements `WslcCanRun` using existing primitives.  Updates the component flag names to help better describe their meaning and adds comments to further do so.

Also changes the error returned when we cannot create a `WSLASessionManager` and WSL is installed at a version below the minimum for WSLC support.

## Validation Steps Performed
Test success.


src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -264,6 +264,71 @@ struct ImageFileResolver
...
+// TODO: Implement Server SKU specific checks
+bool NeedsVirtualMachineServicesInstalled()
+{
+    return !wsl::windows::common::wslutil::IsVirtualMachinePlatformInstalled();
+}
+
+bool NeedsWslOptionalFeatureInstalled()
+{
+    return !wsl::windows::common::helpers::IsWindows11OrAbove() &&
+           !wsl::windows::common::helpers::IsServicePresent(L"lxssmanager");
```
benhillis: this is incorrect, the service you should be looking for is "lxss" it's the WSL1 driver. Unless I'm missing something...

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -264,6 +264,71 @@ struct ImageFileResolver
...
+}
+
+bool DoesWslRuntimeVersionSupportWslc(const std::optional<std::tuple<uint32_t, uint32_t, uint32_t>>& version)
+{
+    constexpr auto minimalPackageVersion = std::tuple<uint32_t, uint32_t, uint32_t>{2, 7, 0};
```
benhillis: Needs to be 2.8.0

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -264,6 +264,71 @@ struct ImageFileResolver
...
+    NotInstalled,
+    InstalledWithoutWslcSupport,
+    InstalledWithWslcSupport,
+};
+
+WslRuntimeState CheckWslRuntimeState()
+{
+    auto version = wsl::windows::common::wslutil::GetInstalledPackageVersion();
+
+    return version.has_value() ? (DoesWslRuntimeVersionSupportWslc(version) ? WslRuntimeState::InstalledWithWslcSupport
```
benhillis: double ternary is hard to read imo.

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -264,6 +264,71 @@ struct ImageFileResolver
...
+        WslRuntimeState currentState = CheckWslRuntimeState();
+        THROW_WIN32_IF_MSG(
+            ERROR_NOT_SUPPORTED,
+            currentState == WslRuntimeState::InstalledWithoutWslcSupport,
+            "The currently installed WSL version does not support WSLC; upgrade to 2.7 or later.");
+        THROW_HR_IF_MSG(
+            hr,
+            currentState == WslRuntimeState::InstalledWithWslcSupport,
+            "The WSL install appears to be corrupted; session manager class was not registered.");
+        THROW_HR(hr);
```
benhillis: not needed, line 326 will throw.

benhillis (changes requested):

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: That will complicate tests and usage of this function until we move the minor version forward.  I would rather leave a TODO until that happens.

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: I based it on this code from earlier by @OneBlue :
https://github.com/microsoft/WSL/blob/cb2595c3f18ff6cf9f5fb29c594aa44de4a1c7b7/src/windows/wslaclient/DllMain.cpp#L43-L66

benhillis: Ok, it's still wrong.

src/windows/WslcSDK/wslcsdk.cpp
benhillis: Ok I'm fine with this.

src/windows/WslcSDK/wslcsdk.cpp
benhillis: @OneBlue - is the WSL1 optional component required at all for WSLC? All that's in there is p9rdr, lxcore, lxutil, none of which we use.

OneBlue: We might not "require" it in wslc, but wslservice.exe will exit if the component is not enabled for now. 

We should probably remove that requirement in ServiceMain() and then test that wslc works correctly on Windows 10

benhillis: https://github.com/microsoft/WSL/pull/14546

benhillis: It works just fine.

benhillis: Remove this whole method after https://github.com/microsoft/WSL/pull/14546 is merged.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -477,8 +449,13 @@ STDAPI WslcCreateSessionVhd(_In_ WslcSession session, _In_ const WslcVhdRequirem
 typedef enum WslcComponentFlags
 {
     WSLC_COMPONENT_FLAG_NONE = 0,
-    WSLC_COMPONENT_FLAG_VMPOC = 1,
-    WSLC_COMPONENT_FLAG_WSL_OC = 2,
+    // Services provided by the Virtual Machine Platform optional feature (other optional features may provide these services as
+    // well). Installing this component will require a reboot.
+    WSLC_COMPONENT_FLAG_VIRTUAL_MACHINE_SERVICES = 1,
```
benhillis: Virtual Machine Platform

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -477,8 +449,13 @@ STDAPI WslcCreateSessionVhd(_In_ WslcSession session, _In_ const WslcVhdRequirem
...
+    // The Windows Subsystem for Linux optional feature (required before Windows 11).
+    // Installing this component will require a reboot.
+    WSLC_COMPONENT_FLAG_WSL_OPTIONAL_FEATURE = 2,
```
benhillis: remove this flag

src/windows/WslcSDK/wslcsdk.h
JohnMcPMS: That seems likely to cause confusion (as it did to me) if we state that you need this optional feature but what you really need are specific services that are also provided via other optional features (Hyper-V).  My dev machine has all of the things that are checked for (`IsVirtualMachinePlatformInstalled` returns `true`) despite listing the `VirtualMachinePlatform` optional feature as disabled in the OS.

But if we think the comment is sufficient to remove that confusion I will change it.

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -264,6 +264,71 @@ struct ImageFileResolver
...
+WslRuntimeState CheckWslRuntimeState()
+{
+    auto version = wsl::windows::common::wslutil::GetInstalledPackageVersion();
+
+    if (!version.has_value())
+    {
+        return WslRuntimeState::NotInstalled;
+    }
+
+    return DoesWslRuntimeVersionSupportWslc(version) ? WslRuntimeState::InstalledWithWslcSupport : WslRuntimeState::InstalledWithoutWslcSupport;
```
OneBlue, line 298: I just realized that instead of checking the package version, we can just try to instantiate `WSLCSessionManager`, and if we get `REGDB_E_CLASSNOTREG`, then we know that either the package isn't installed, or it's not recent enough. 

Doing this would have the benefit of getting rid of the chicken-egg problem of adding test coverage for this

OneBlue (approved):

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -264,6 +264,87 @@ struct ImageFileResolver
...
+{
+    auto [result, hr] = CreateSessionManagerRaw();
+
+    if (hr == REGDB_E_CLASSNOTREG)
+    {
+        WslRuntimeState currentState = CheckWslRuntimeState();
+        THROW_WIN32_IF_MSG(
+            ERROR_NOT_SUPPORTED,
+            currentState == WslRuntimeState::InstalledWithoutWslcSupport,
+            "The currently installed WSL version does not support WSLC; upgrade to " WSLC_API_MIN_VERSION_SUPPORTED_STRING
```
benhillis: This string is not user visible, it will only go to traces, as such adding the version string here (and having a separate define for it seems not very useful).

benhillis (approved):
minor comment on strings, this is fine though.

PR #14543, opened 2026-03-26
[WSLC] add WSLC CMake file in NuGet

chemwolf6922:

Add the CMake file in the WSLC NuGet package so developers can consume the sdk in CMake easier.
Example usage:
1. Install the NuGet package:
```sh
 nuget install Microsoft.WSL.Containers -OutputDirectory .\packages -Prerelease
```
2. Reference in CMakeLists.txt
```cmake
file(GLOB WSLC_PACKAGE_DIR "${CMAKE_SOURCE_DIR}/packages/Microsoft.WSL.Containers.*")
list(APPEND CMAKE_PREFIX_PATH "${WSLC_PACKAGE_DIR}")
find_package(Microsoft.WSL.Containers REQUIRED)

target_link_libraries(test_app PRIVATE Microsoft.WSL.Containers::SDK)

```

## PR Checklist

## Validation Steps Performed

Tested locally with this minimal test app:
```c++
// main.cpp
#include <wslcsdk.h>
#include <cstdio>

int main()
{
    BOOL canRun{};
    WslcComponentFlags missingComponents{};
    HRESULT hr = WslcCanRun(&canRun, &missingComponents);

    if (FAILED(hr))
    {
        printf("WslcCanRun failed: 0x%08lx\n", hr);
        return 1;
    }

    printf("canRun: %d, missingComponents: %d\n", canRun, missingComponents);
    return 0;
}
```
```cmake
# CMakeLists.txt
cmake_minimum_required(VERSION 3.25)
project(wslc-nuget-test)

file(GLOB WSLC_PACKAGE_DIR "${CMAKE_SOURCE_DIR}/packages/Microsoft.WSL.Containers.*")
list(APPEND CMAKE_PREFIX_PATH "${WSLC_PACKAGE_DIR}")
find_package(Microsoft.WSL.Containers REQUIRED)

add_executable(test_app main.cpp)
target_link_libraries(test_app PRIVATE Microsoft.WSL.Containers::SDK)

add_custom_command(TARGET test_app POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E copy_if_different
        $<TARGET_RUNTIME_DLLS:test_app>
        $<TARGET_FILE_DIR:test_app>
    COMMAND_EXPAND_LISTS
)
```
The test app compiles fine. And prints the not implemented error code as expected.

### DIFF
--- a/nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
+++ b/nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
@@ -1,0 +1,57 @@
+if(TARGET Microsoft.WSL.Containers::sdk)
+    return()
+endif()
+
+# Determine target architecture
+if(CMAKE_GENERATOR_PLATFORM)
+    string(TOLOWER "${CMAKE_GENERATOR_PLATFORM}" _wslcsdk_platform)
+    if(_wslcsdk_platform STREQUAL "x64")
+        set(_wslcsdk_arch "x64")
+    elseif(_wslcsdk_platform STREQUAL "arm64")
+        set(_wslcsdk_arch "arm64")
+    else()
+        message(FATAL_ERROR "Microsoft.WSL.Containers: Unsupported platform '${CMAKE_GENERATOR_PLATFORM}'. Supported: x64, ARM64.")
+    endif()
+    unset(_wslcsdk_platform)
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "[Aa][Mm][Dd]64|[Xx]86_64|[Xx]64")
+    set(_wslcsdk_arch "x64")
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "[Aa][Rr][Mm]64|[Aa]arch64")
+    set(_wslcsdk_arch "arm64")
+else()
+    message(FATAL_ERROR
+        "Microsoft.WSL.Containers: Unsupported architecture '${CMAKE_SYSTEM_PROCESSOR}'. "
+        "Supported: x64, ARM64.")
+endif()
+
+# Compute paths relative to package root (<root>/build/cmake/ -> <root>/)
+get_filename_component(_wslcsdk_root "${CMAKE_CURRENT_LIST_DIR}/../.." ABSOLUTE)
+set(_wslcsdk_include_dir "${_wslcsdk_root}/include")
+set(_wslcsdk_lib_dir "${_wslcsdk_root}/runtimes/win-${_wslcsdk_arch}")
+
+# Validate that expected files exist
+if(NOT EXISTS "${_wslcsdk_include_dir}/wslcsdk.h")
+    message(FATAL_ERROR "Microsoft.WSL.Containers: Cannot find wslcsdk.h in '${_wslcsdk_include_dir}'")
+endif()
+
+if(NOT EXISTS "${_wslcsdk_lib_dir}/wslcsdk.lib")
+    message(FATAL_ERROR "Microsoft.WSL.Containers: Cannot find wslcsdk.lib in '${_wslcsdk_lib_dir}'")
+endif()
+
+# Create imported target
+add_library(Microsoft.WSL.Containers::sdk SHARED IMPORTED GLOBAL)
+set_target_properties(Microsoft.WSL.Containers::sdk PROPERTIES
+    INTERFACE_INCLUDE_DIRECTORIES "${_wslcsdk_include_dir}"
+    IMPORTED_IMPLIB "${_wslcsdk_lib_dir}/wslcsdk.lib"
+    IMPORTED_LOCATION "${_wslcsdk_lib_dir}/wslcsdk.dll"
+)
+
+# Standard find_package output variables
+set(Microsoft.WSL.Containers_FOUND TRUE)
+set(Microsoft.WSL.Containers_INCLUDE_DIRS "${_wslcsdk_include_dir}")
+set(Microsoft.WSL.Containers_LIBRARIES Microsoft.WSL.Containers::sdk)
+
+# Clean up temporary variables
+unset(_wslcsdk_arch)
+unset(_wslcsdk_root)
+unset(_wslcsdk_include_dir)
+unset(_wslcsdk_lib_dir)
--- a/nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
+++ b/nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
@@ -1,4 +1,4 @@
-if(TARGET Microsoft.WSL.Containers::sdk)
+if(TARGET Microsoft.WSL.Containers::SDK)
     return()
 endif()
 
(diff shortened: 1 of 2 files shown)


nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
```diff
@@ -0,0 +1,57 @@
...
+        set(_wslcsdk_arch "x64")
+    elseif(_wslcsdk_platform STREQUAL "arm64")
+        set(_wslcsdk_arch "arm64")
+    else()
+        message(FATAL_ERROR "Microsoft.WSL.Containers: Unsupported platform '${CMAKE_GENERATOR_PLATFORM}'. Supported: x64, ARM64.")
+    endif()
+    unset(_wslcsdk_platform)
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "[Aa][Mm][Dd]64|[Xx]86_64|[Xx]64")
+    set(_wslcsdk_arch "x64")
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "[Aa][Rr][Mm]64|[Aa]arch64")
```
ptrivedi: minor nit: could be simplified to just arm64|aarch64 with a case-insensitive compare
string(TOLOWER "${CMAKE_SYSTEM_PROCESSOR}" _proc_lower)

nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
```diff
@@ -0,0 +1,57 @@
...
+    set(_wslcsdk_arch "arm64")
+else()
+    message(FATAL_ERROR
+        "Microsoft.WSL.Containers: Unsupported architecture '${CMAKE_SYSTEM_PROCESSOR}'. "
+        "Supported: x64, ARM64.")
```
ptrivedi: This 'else' block will be reached if CMAKE_GENERATOR_PLATFORM is not set

nuget/Microsoft.WSL.Containers/cmake/Microsoft.WSL.ContainersConfig.cmake
```diff
@@ -0,0 +1,57 @@
+if(TARGET Microsoft.WSL.Containers::sdk)
+    return()
+endif()
+
```
ptrivedi, line 4: Should we set the Microsoft.WSL.Containers_VERSION to 1.0? That way if somone does a find_package with that version, it would show up. 
@florelis @1wizkid

ptrivedi (approved):

nuget/Microsoft.WSL.Containers/cmake/Microsoft.WSL.ContainersConfig.cmake
florelis: I would expect the version to match the package version for everything else

nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
florelis: Isn't it if `CMAKE_GENERATOR_PLATFORM` isn't set and `CMAKE_SYSTEM_PROCESSOR` doesn't match?

nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
```diff
@@ -0,0 +1,57 @@
...
+set(_wslcsdk_lib_dir "${_wslcsdk_root}/runtimes/win-${_wslcsdk_arch}")
+
+# Validate that expected files exist
+if(NOT EXISTS "${_wslcsdk_include_dir}/wslcsdk.h")
+    message(FATAL_ERROR "Microsoft.WSL.Containers: Cannot find wslcsdk.h in '${_wslcsdk_include_dir}'")
+endif()
+
+if(NOT EXISTS "${_wslcsdk_lib_dir}/wslcsdk.lib")
+    message(FATAL_ERROR "Microsoft.WSL.Containers: Cannot find wslcsdk.lib in '${_wslcsdk_lib_dir}'")
+endif()
```
florelis: Do we need this? If we ever publish a version of the package so broken so as to hit this, it would be pretty obvious in other ways.

nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
```diff
@@ -0,0 +1,57 @@
...
+# Create imported target
+add_library(Microsoft.WSL.Containers::sdk SHARED IMPORTED GLOBAL)
+set_target_properties(Microsoft.WSL.Containers::sdk PROPERTIES
+    INTERFACE_INCLUDE_DIRECTORIES "${_wslcsdk_include_dir}"
+    IMPORTED_IMPLIB "${_wslcsdk_lib_dir}/wslcsdk.lib"
+    IMPORTED_LOCATION "${_wslcsdk_lib_dir}/wslcsdk.dll"
+)
+
+# Standard find_package output variables
+set(Microsoft.WSL.Containers_FOUND TRUE)
```
florelis: Do we need to set this? From my reading of the docs I understood that it was set by `find_packages()`

nuget/Microsoft.WSL.Containers/cmake/Microsoft.WSL.ContainersConfig.cmake
florelis, line 1: I don't fully understand the search rules `find_package()` uses, but I got the impression that if we put this under `lib/cmake` instead of `build/cmake`, it would be able to find it with just having to provide the root.

So, doing
`set(Microsoft.WSL.Containers_DIR "${WSLC_PACKAGE_DIR}")`
instead of
`set(Microsoft.WSL.Containers_DIR "${WSLC_PACKAGE_DIR}/build/cmake")`
in the consumer.

nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
```diff
@@ -0,0 +1,57 @@
...
+set(Microsoft.WSL.Containers_INCLUDE_DIRS "${_wslcsdk_include_dir}")
+set(Microsoft.WSL.Containers_LIBRARIES Microsoft.WSL.Containers::sdk)
```
florelis: Do we need these? What I understood from docs is that these are only for Find-module Packages, but this is a Config-file Package

nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
```diff
@@ -0,0 +1,57 @@
+if(TARGET Microsoft.WSL.Containers::sdk)
```
florelis: Nit: It feels weird to use both Capitalized and lowercase words here. Why not `Microsoft.WSL.Containers::SDK`?

florelis: Is there a reason to call it `Microsoft.WSL.Containers::sdk` and not just `Microsoft.WSL.Containers`?

nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
```diff
@@ -0,0 +1,57 @@
...
+    message(FATAL_ERROR "Microsoft.WSL.Containers: Cannot find wslcsdk.lib in '${_wslcsdk_lib_dir}'")
+endif()
+
+# Create imported target
+add_library(Microsoft.WSL.Containers::sdk SHARED IMPORTED GLOBAL)
+set_target_properties(Microsoft.WSL.Containers::sdk PROPERTIES
+    INTERFACE_INCLUDE_DIRECTORIES "${_wslcsdk_include_dir}"
+    IMPORTED_IMPLIB "${_wslcsdk_lib_dir}/wslcsdk.lib"
+    IMPORTED_LOCATION "${_wslcsdk_lib_dir}/wslcsdk.dll"
+)
```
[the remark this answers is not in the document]
chemwolf6922: Removed all file checks. Trusting the NuGet package.

nuget/Microsoft.WSL.Containers/build/cmake/Microsoft.WSL.ContainersConfig.cmake
chemwolf6922: We are planning to add container image build targets here. Which will use other ::xxx names.

nuget/Microsoft.WSL.Containers/cmake/Microsoft.WSL.ContainersConfig.cmake
chemwolf6922: The lib folder for a NuGet package is for .net binaries. It may not be a good idea to put build files there.

nuget/Microsoft.WSL.Containers/cmake/Microsoft.WSL.ContainersConfig.cmake
chemwolf6922: The package version control is already done when the user specified the NuGet package. The cmake package version seems redundant here.

nuget/Microsoft.WSL.Containers/cmake/Microsoft.WSL.ContainersConfig.cmake
florelis: Looking at the [folder conventions](https://learn.microsoft.com/nuget/create-packages/creating-a-package#from-a-convention-based-working-directory), my understanding is that it only matters in that it helps MSBuild take certain actions automatically. Putting it in `lib/` would just be doing the same thing for CMake.

I don't think it would be a problem to have the `.cmake` file there. The only possible issue I can imagine is if it was automatically copied to the output and ended up being included in the consumer's app. But according to the docs about `lib/`, MSBuild only does that for specific file types:
> Assemblies are added as references for compile as well as runtime; .xml and .pdb copied into project folders.

florelis (approved):
Looks good to me, aside from my comment about the file location.

nuget/Microsoft.WSL.Containers/cmake/Microsoft.WSL.ContainersConfig.cmake
chemwolf6922: Moved it to nuget_root/cmake/ so users can use it by appending the nuget folder to the CMAKE_PREFIX_PATH.

PR #14545, opened 2026-03-26
Implement SDK inspect and VHD management

JohnMcPMS:
Implements `WslcInspectContainer` and `WslcCreateSessionVhd`; adds `WslcSetContainerSettingsNamedVolumes` so that the VHDs are useful.

## PR Checklist

`WslcInspectContainer` is a straightforward pass-through.

~~`WslcCreateSessionVhd`~~ `WslcCreateSessionVhdVolume` was also relatively straightforward, except it required adding a name to the existing VHD struct.  Fixed VHDs cannot be requested via the runtime API currently, so attempting to use them is an error.  If we never plan to allow them I can simply remove the option.

In order to actually use the created VHD, I had to add `WslcSetContainerSettingsNamedVolumes` so that they could be mounted in a container.  This is basically a copy of the Windows <-> container volume setting with the Windows path replaced by the VHD name.  And to enable removal, `WslcDeleteSessionVhdVolume` was added.

## Validation Steps Performed
Tests added for inspect and VHD creation/use.


src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -405,10 +394,25 @@ CATCH_RETURN();
 STDAPI WslcCreateSessionVhd(_In_ WslcSession session, _In_ const WslcVhdRequirements* options, _Outptr_opt_result_z_ PWSTR* errorMessage)
```
OneBlue: nit: Given that our design has changed a bit, I wonder if we should rename as `WslcCreateVhdVolume()` to make this a bit more clear ? The name gives the impression that it's creating the "main" storage VHD for the sesssions

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -405,10 +394,25 @@ CATCH_RETURN();
...
-    return E_NOTIMPL;
+    ErrorInfoWrapper errorInfoWrapper{errorMessage};
+
+    auto internalType = CheckAndGetInternalType(session);
+    RETURN_HR_IF_NULL(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), internalType->session);
+    RETURN_HR_IF_NULL(E_POINTER, options);
+
+    RETURN_HR_IF_NULL(E_INVALIDARG, options->name);
+    RETURN_HR_IF(E_INVALIDARG, options->sizeInBytes == 0);
+    RETURN_HR_IF(E_NOTIMPL, options->type == WSLC_VHD_TYPE_FIXED);
```
OneBlue: nit: We should fail if `options->type` is not `WSLC_VHD_TYPE_DYNAMIC` here

src/windows/WslcSDK/wslcsdk.cpp
OneBlue: And either way we should have a method to delete the created volume, otherwise those will be unremovable

OneBlue (approved):

PR #14548, opened 2026-03-26
Initial wslc settings support

yao-msft:
Initial settings support. Still need localization as current wslc does not have localization framework hooked up yet.

## PR Checklist

## Validation Steps Performed
Added tests. Also manually verified settings commands work and settings values do get populated to session objects defaults.


src/windows/wslc/commands/SettingsCommand.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    const auto path = settings::User().SettingsFilePath();
+
+    // Some versions of windows will fail if no file extension association exists, other will pop up the dialog
+    // to make the user pick their default.
+    HINSTANCE res = ShellExecuteW(nullptr, nullptr, path.c_str(), nullptr, nullptr, SW_SHOW);
+    if (static_cast<int>(reinterpret_cast<uintptr_t>(res)) <= 32)
+    {
+        // User doesn't have file type association. Default to notepad
+        ShellExecuteW(nullptr, nullptr, L"notepad", path.c_str(), nullptr, SW_SHOW);
```
OneBlue: nit: I recommend passing the full path to notepad here (we can use `wil::GetSystemDirectoryW` for that)

src/windows/wslc/services/SessionModel.cpp
```diff
@@ -14,21 +14,24 @@ Module Name:
...
+#include "UserSettings.h"
 
 namespace wsl::windows::wslc::models {
 SessionOptions SessionOptions::Default()
 {
     // Use a function-local static to defer path initialization until first use.
-    static const std::filesystem::path defaultPath = {wsl::windows::common::filesystem::GetLocalAppDataPath(nullptr) / "wsla"};
+    static const std::filesystem::path storagePath =
+        settings::User().Get<settings::Setting::SessionStoragePath>().empty()
+            ? std::filesystem::path{wsl::windows::common::filesystem::GetLocalAppDataPath(nullptr) / L"wslc\\defaultstorage"}
```
OneBlue: nit: I'd recommend using something like this as the default: 

```
common::filesystem::GetLocalAppDataPath(nullptr) / "wslc" / "storage"
```

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -0,0 +1,268 @@
...
+    "# https://aka.ms/wslc-settings\n"
+    "\n"
+    "session:\n"
+    "  # Number of virtual CPUs allocated to the session (default: 4)\n"
+    "  # cpuCount: 4\n"
+    "\n"
+    "  # Memory limit for the session in megabytes (default: 2048)\n"
+    "  # memorySizeMb: 2048\n"
+    "\n"
+    "  # Maximum disk image size in megabytes (default: 10000)\n"
```
OneBlue: 10 GB is the current default, but I should we should increase since some images are already bigger than that. Maybe 100GB as the default would be a good value ?

src/windows/wslc/settings/UserSettings.h
```diff
@@ -0,0 +1,172 @@
...
+        return details::SettingMapping<S>::DefaultValue;
+    }
+};
+
+// Indicates which source the settings were loaded from.
+enum class UserSettingsType
+{
+    Default,  // Neither settings file existed (first run); built-in defaults are used.
+    Standard, // Primary file (UserSettings.yaml) loaded successfully.
+    Backup,   // Primary file failed to parse; backup file (UserSettings.yaml.bak) was used.
```
OneBlue: Could we remove the logic to load the backup file ? I don't think we need it, so that would simplify this a bit

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -0,0 +1,268 @@
...
+    }
+}
+
+void UserSettings::Reset() const
+{
+    std::filesystem::create_directories(m_primaryPath.parent_path());
+    std::ofstream file(m_primaryPath);
+    THROW_HR_IF_MSG(E_UNEXPECTED, !file.is_open(), "Failed to create settings file");
+    file << s_DefaultSettingsTemplate;
+    file.flush();
```
OneBlue: nit: flush() and close() aren't needed here

src/windows/wslc/core/Main.cpp
```diff
@@ -55,6 +56,13 @@ try
 
     try
     {
+        // Emit any settings load warnings.
+        for (const auto& warning : settings::User().GetWarnings())
```
OneBlue: I recommend only loading the settings when we need them so we don't delay executions for commands like `--version` for instance. 

We could have a method to load the settings and print warnings if needed. Given that we're in a CLI, I think it would be OK for the method that loads the settings to print the warnings directly

OneBlue: Also we should print the warnings on stderr, not stdout

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -0,0 +1,252 @@
...
+    // Attempts to parse a YAML document from the given file path.
+    // Returns an empty optional and pushes a warning if the file exists but fails to parse.
+    std::optional<YAML::Node> TryLoadYaml(const std::filesystem::path& path, std::vector<Warning>& warnings)
+    {
+        std::ifstream stream(path);
+        if (!stream.is_open())
+        {
+            // If the file exists but cannot be opened (permissions, sharing violation, etc.),
+            // emit a warning so the user understands why settings were ignored.
+            if (std::filesystem::exists(path))
```
OneBlue: nit: instead of checking if the file exists, I recommend checking `errno` and logging an error like: 

```
Failed to open <file>, errno: <errno>, using default settings
```

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -0,0 +1,252 @@
...
+            }
+
+            return std::nullopt;
+        }
+
+        try
+        {
+            return YAML::Load(stream);
+        }
+        catch (const YAML::Exception& e)
```
OneBlue: nit: We could catch for `std::exception`  and factor both catch blocks into one

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -0,0 +1,252 @@
...
+        catch (...)
+        {
+            warnings.push_back({std::format(L"Warning: '{}' could not be parsed.", path.filename().wstring()), {}});
+            return std::nullopt;
+        }
+    }
+
+    const std::filesystem::path& SettingsDir()
+    {
+        static const std::filesystem::path dir =
```
OneBlue: nit: I would have a preference to write the settings directly under `wslc` (and skipping the `settings` folder)

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -0,0 +1,252 @@
...
+    return instance;
+}
+
+UserSettings::UserSettings() : UserSettings(SettingsDir())
+{
+}
+
+UserSettings::UserSettings(const std::filesystem::path& settingsDir)
+{
+    m_settingsPath = settingsDir / L"UserSettings.yaml";
```
OneBlue: nit: Could we call this `Settings.yml` (given that this file is already inside a user profile)

CMakeLists.txt
```diff
@@ -48,6 +48,18 @@ FetchContent_Declare(nlohmannjson
 FetchContent_MakeAvailable(nlohmannjson)
 FetchContent_GetProperties(nlohmannjson SOURCE_DIR NLOHMAN_JSON_SOURCE_DIR)
 
+FetchContent_Declare(yaml-cpp
+                     URL https://github.com/jbeder/yaml-cpp/releases/download/yaml-cpp-0.9.0/yaml-cpp-yaml-cpp-0.9.0.tar.gz
```
OneBlue, line 52: We should register this new dependency in `cgmanifest.json` 

As I'm writing this I just noticed that we don't have our json library recorded there. I'll fix that in a separate PR

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -0,0 +1,252 @@
...
+static constexpr std::string_view s_DefaultSettingsTemplate =
+    "# wslc user settings\n"
+    "# https://aka.ms/wslc-settings\n"
+    "\n"
+    "session:\n"
+    "  # Number of virtual CPUs allocated to the session (default: 4)\n"
+    "  # cpuCount: 4\n"
+    "\n"
+    "  # Memory limit for the session in megabytes (default: 2048)\n"
+    "  # memorySizeMb: 2048\n"
```
OneBlue: Could we use storage string for these values ? 

Something like: 

```
memory: 10GB | 10000MB
```

I would have a preference for that format since that's what docker-compose does and it's a bit more user friendly IMO. 


We already have a method to parse those strings: https://github.com/microsoft/WSL/blob/bf759a092be8ed481dce70bf4f01f16c1b9b54a8/src/shared/inc/stringshared.h#L296

OneBlue (commented):
LGTM. Minor comments

src/windows/wslc/settings/UserSettings.cpp
yao-msft: I think we want to consider other settings in the future, like AdminSettings, etc.

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -36,18 +36,25 @@ static constexpr std::string_view s_DefaultSettingsTemplate =
...
     "  # Default path for container storage (default: %LocalAppData%\\wslc\\storage)\n"
     "  # defaultStoragePath: \"\"\n";
 
 // Validate individual setting specializations
 namespace details {
 
+    std::optional<uint32_t> ParseSettingsMemoryValue(const std::string& value)
+    {
+        auto parsed = wsl::shared::string::ParseMemorySize(value.c_str());
+        auto converted = parsed.has_value() ? *parsed / 1048576 : 0; // To Mb, and anything less than 1Mb is considered invalid.
```
OneBlue, line 54: nit: We can use the `_1MB` macro for this (in `helpers.hpp`)

OneBlue (approved):
LGTM. 

I think in the longer term we might want to add more advanced defaults for CPU & memory like we do in WSL (by default we use all the CPU cores, and half of the available memory), but I think this is something that we should spec and definitely not required for selfhost. 

Thank you for doing this !

src/windows/wslc/settings/UserSettings.cpp
yao-msft: I'll update in the followup pr for networkMode setting

PR #14557, opened 2026-03-27
[CLI] Add initial support for image saving feature

AmelBawa-msft:
- Add initial support for image saving feature
- 🐞http://task.ms/61609188

## PR Checklist


src/windows/wslc/services/ImageService.cpp
```diff
@@ -142,8 +142,14 @@ void ImageService::Push()
 {
 }
 
-void ImageService::Save()
+void ImageService::Save(wsl::windows::wslc::models::Session& session, const std::string& image, const std::wstring& output)
 {
+    wil::unique_hfile outputFile{
+        CreateFileW(output.c_str(), GENERIC_WRITE, FILE_SHARE_READ, nullptr, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr)};
```
ptrivedi, line 150: Should use a 0 (no sharing) instead of FILE_SHARE_READ here. Don't want to allow a concurrent read

src/windows/wslc/commands/ImageSaveCommand.cpp
```diff
@@ -0,0 +1,51 @@
...
+    return {
+        Argument::Create(ArgType::ImageId, true),
+        Argument::Create(ArgType::Output, true),
+        Argument::Create(ArgType::Session),
+    };
+}
+
+std::wstring ImageSaveCommand::ShortDescription() const
+{
+    return {L"Save images."};
```
ptrivedi, line 37: nit: okay for now, but short and long descriptions are returning the same things

ptrivedi (commented):
Couple of minor comments, looks good otherwise

ptrivedi (approved):

OneBlue (approved):

src/windows/wslc/services/ImageService.cpp
OneBlue: Concurrent read isn't especially a problem from wslc's perspective. Usually as long as it doesn't create an issue for ourselves, we allow FILE_SHARE_READ on the handles that we open (otherwise it can make debugging / the UX a bit annoying if the files are completely locked)

src/windows/wslc/services/ImageService.cpp
```diff
@@ -142,8 +142,14 @@ void ImageService::Push()
...
+    THROW_LAST_ERROR_IF(!outputFile);
+
+    // TODO Handle Ctrl-C and progress callback
```
OneBlue: For progress bar I recommend looking into `HandleConsoleProgressBar`. It should fix this scenario pretty well

src/windows/wslc/services/ImageService.cpp
```diff
@@ -142,8 +142,14 @@ void ImageService::Push()
 {
 }
 
-void ImageService::Save()
+void ImageService::Save(wsl::windows::wslc::models::Session& session, const std::string& image, const std::wstring& output)
 {
+    wil::unique_hfile outputFile{
```
OneBlue, line 149: nit: We should output to stdout if `-o` isn't specified (and probably fail if stdin isn't a terminal in that case). 

This will allow callers to easily export to a pipe

AmelBawa-msft: http://task.ms/61660160

PR #14560, opened 2026-03-27
Implement WSLAContainer::Kill()

OneBlue:
## PR Checklist

### DIFF
--- a/src/windows/wslc/services/ContainerService.cpp
+++ b/src/windows/wslc/services/ContainerService.cpp
@@ -309,7 +309,8 @@ void ContainerService::Kill(Session& session, const std::string& id, WSLCSignal 
 {
     wil::com_ptr<IWSLCContainer> container;
     THROW_IF_FAILED(session.Get()->OpenContainer(id.c_str(), &container));
-    StopInternal(*container, signal);
+
+    THROW_IF_FAILED(container->Kill(signal));
 }
 
 void ContainerService::Delete(Session& session, const std::string& id, bool force)
--- a/src/windows/wslc/services/ContainerService.cpp
+++ b/src/windows/wslc/services/ContainerService.cpp
@@ -153,11 +153,6 @@ static wsl::windows::common::RunningWSLCContainer CreateInternal(
     return std::move(*runningContainer);
 }
 
-static void StopInternal(IWSLCContainer& container, WSLCSignal signal = WSLCSignalNone, LONG timeout = -1)
-{
-    THROW_IF_FAILED(container.Stop(signal, timeout)); // TODO: Error message
-}
-
 std::wstring ContainerService::FormatRelativeTime(ULONGLONG timestamp)
 {
     constexpr LONGLONG SecondsPerMinute = std::chrono::duration_cast<std::chrono::seconds>(1min).count();
@@ -308,14 +303,13 @@ void ContainerService::Stop(Session& session, const std::string& id, StopContain
 {
     wil::com_ptr<IWSLCContainer> container;
     THROW_IF_FAILED(session.Get()->OpenContainer(id.c_str(), &container));
-    StopInternal(*container, options.Signal, options.Timeout);
+    THROW_IF_FAILED(container->Stop(options.Signal, options.Timeout));
 }
 
 void ContainerService::Kill(Session& session, const std::string& id, WSLCSignal signal)
 {
     wil::com_ptr<IWSLCContainer> container;
     THROW_IF_FAILED(session.Get()->OpenContainer(id.c_str(), &container));
-
     THROW_IF_FAILED(container->Kill(signal));
 }
 
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -613,12 +613,12 @@ void WSLCContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCod
         TraceLoggingValue((int)event, "Event"));
 }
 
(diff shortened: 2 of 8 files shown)


src/windows/wslc/services/ContainerService.cpp
```diff
@@ -315,7 +315,8 @@ void ContainerService::Kill(Session& session, const std::string& id, WSLCSignal
 {
     wil::com_ptr<IWSLCContainer> container;
     THROW_IF_FAILED(session.Get()->OpenContainer(id.c_str(), &container));
-    StopInternal(*container, signal);
```
AmelBawa-msft, line 318: Super nit: `StopInternal()` was shared between `Stop()` and `Kill()`, after this change we can remove `StopInternal()` and inline the code in `Stop()` method

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -620,12 +620,12 @@ void WSLCContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCod
...
 }
 
-void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds)
+void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill)
 {
     // Acquire an exclusive lock since this method modifies m_state.
     auto lock = m_lock.lock_exclusive();
 
-    if (m_state == WslcContainerStateExited)
+    if (m_state == WslcContainerStateExited && !Kill)
```
AmelBawa-msft, line 628: For kill, do we always signal?

Is there a semantic difference between `wslc stop -s SIGKill -t 0` and `wslc kill` ?

OneBlue: The main difference is that kill should fail when the container isn't running, but if it is, those two commands should behave the same way

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -639,9 +639,20 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds)
             m_state);
```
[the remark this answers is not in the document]
OneBlue: This error message is OK for now. We'll revisit when we look into error localization

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -639,9 +639,20 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds)
...
 
+    std::optional<WSLCSignal> SignalArg;
+    if (Signal != WSLCSignalNone)
+    {
+        SignalArg = Signal;
+    }
+
+    // Don't wait for the container to stop if we're not sending SIGKILL, since it may not stop the container.
+    // N.B. If the signal was SIGTERM for instance, we'll receive the stop notification via OnEvent().
+    bool waitForStop = !Kill || (SignalArg.value_or(WSLCSignalSIGKILL) == WSLCSignalSIGKILL);
```
kvega005, line 650: Why didn't we need this logic before for `Stop`? Or is this just a bug we have always had?

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -639,9 +639,20 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds)
             m_state);
     }
```
kvega005, line 640: Nit: With the `Kill` parameter, this method now has a lot of interleaved branching (`if (Kill)`, `if (waitForStop)`, etc.) that makes it hard to follow. Would it be cleaner to split `Stop` and `Kill` into separate methods on `WSLCContainerImpl` and extract the shared post-stop teardown into a helper?

Something like:

```
void CompleteStopTransition(std::future<std::uint64_t>& stopFuture)
{
    std::optional<std::uint64_t> stopTimestamp;

    if (stopFuture.wait_for(60s) == std::future_status::ready)
        stopTimestamp = stopFuture.get();

    Transition(WslcContainerStateExited, stopTimestamp);
    ReleaseProcesses();
    ReleaseRuntimeResources();

    if (WI_IsFlagSet(m_containerFlags, WSLCContainerFlagsRm))
        DeleteExclusiveLockHeld(WSLCDeleteFlagsForce);
}
```

Then `Stop()` and `Kill()` each handle their own state validation, future setup, and Docker call, but both call `CompleteStopTransition()` for the shared tail. The future still gets created before the Docker call in each method (which is required so `OnEvent()` can resolve the promise), but each path reads linearly without needing to track the Kill/waitForStop flags across 60+ lines.

We can maybe even create a helper to get the future and cleanup: `auto [stopFuture,  stateReset] = GetStopFuture()`

Not blocking — the current code is correct, just suggesting for readability.

kvega005 (approved):

kvega005 (approved):

kvega005 (approved):

src/windows/wslcsession/WSLCContainer.cpp
OneBlue: Well we did. The reason for the wait is that we want to make sure that the notification that we get from docker gets consumed before we return, otherwise we could end up with a race where if the caller did: 

```
Container->Stop();
Container->Start();
```

The notification could arrive after `Start()`, which would put into a bad state

src/windows/wslcsession/WSLCContainer.cpp
OneBlue: Agreed, I considered doing this, but because we need logic before, and after the API call, it felt more complex to fork the paths. 

I'll complete as-is for now, but will look into refactoring this in the future

PR #14564, opened 2026-03-28
Add networking mode setting as fallback

yao-msft:
Added session networking setting as temp fallback if virtionet has issues. As it's only temporary, we did not add the setting to default template. The setting may be removed once we fixed the virtionet issue.

## PR Checklist

## Validation Steps Performed
Manually tested the changes. As this change is only temporary, test was not updated.

### DIFF
--- a/src/windows/wslc/settings/UserSettings.cpp
+++ b/src/windows/wslc/settings/UserSettings.cpp
@@ -51,7 +51,7 @@ namespace details {
     std::optional<uint32_t> ParseSettingsMemoryValue(const std::string& value)
     {
         auto parsed = wsl::shared::string::ParseMemorySize(value.c_str());
-        auto converted = parsed.has_value() ? *parsed / 1048576 : 0; // To Mb, and anything less than 1Mb is considered invalid.
+        auto converted = parsed.has_value() ? *parsed / _1MB : 0; // To Mb, and anything less than 1Mb is considered invalid.
         return converted > 0 ? std::optional{static_cast<uint32_t>(converted)} : std::nullopt;
     }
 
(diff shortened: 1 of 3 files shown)


src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -80,6 +80,24 @@ namespace details {
...
+    {
+        if (value == "none")
+        {
+            return WSLCNetworkingModeNone;
+        }
+        else if (value == "nat")
+        {
+            return WSLCNetworkingModeNAT;
+        }
+        else if (value == "virtionet")
```
benhillis: virtioproxy would be better.

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -80,6 +80,24 @@ namespace details {
         return MultiByteToWide(value);
     }
 
+    WSLC_VALIDATE_SETTING(SessionNetworkingMode)
+    {
+        if (value == "none")
+        {
+            return WSLCNetworkingModeNone;
+        }
+        else if (value == "nat")
```
ptrivedi: nit: don't need 'else' if the 'if' block contains an unconditional return

ptrivedi:
lgtm, once the comments are resolved

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -80,6 +80,24 @@ namespace details {
...
+    {
+        if (value == "none")
+        {
+            return WSLCNetworkingModeNone;
+        }
+        if (value == "nat")
+        {
+            return WSLCNetworkingModeNAT;
+        }
+        if (value == "virtioproxy")
```
OneBlue, line 93: nit: I would have a small preference for `virtionet` here, since that's how we've been describing it so far. What do you think @benhillis  ?

OneBlue (approved):

PR #14567, opened 2026-03-28
Fix deadlock between WatchForExitedProcesses and VMProcessControl destructor

benhillis:
Use shared_ptr for VMProcessControl in m_trackedProcesses so that WatchForExitedProcesses can safely hold a reference while calling OnExited() outside m_trackedProcessesLock. This eliminates both the deadlock and any use-after-free risk from the raw pointer approach.

### DIFF
--- a/src/windows/wslcsession/WSLCVirtualMachine.cpp
+++ b/src/windows/wslcsession/WSLCVirtualMachine.cpp
@@ -328,7 +328,10 @@ WSLCVirtualMachine::~WSLCVirtualMachine()
     // Clear the state of all remaining processes now that the VM has exited.
     for (auto& e : m_trackedProcesses)
     {
-        e->OnVmTerminated();
+        if (auto locked = e.lock())
+        {
+            locked->OnVmTerminated();
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


anshika755:
> Use shared_ptr for VMProcessControl in m_trackedProcesses so that WatchForExitedProcesses can safely hold a reference while calling OnExited() outside m_trackedProcessesLock. This eliminates both the deadlock and any use-after-free risk from the raw pointer approach.

Using shared_ptr here seems like a solid way to handle the lifetime and avoid the deadlock and use-after-free issues.                                                                                                                                                                                                                                   I have one question: in a concurrent scenario, if OnExited() is called while there are still references in m_trackedProcesses, could the delayed destruction from shared_ptr affect the cleanup order or lead to any unexpected behavior?

👍

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -419,26 +422,31 @@ try
             TraceLoggingValue(message->Signaled, "Signaled"));
 
         // Signal the exited process, if it's been monitored.
+        // N.B. Lock weak_ptr under lock, then call OnExited outside it to avoid
+        // deadlock with the destructor's m_lock -> m_trackedProcessesLock ordering.
```
OneBlue, line 426: I'm not following the potential deadlock here. The destructor doesn't hold `m_lock`. It only waits for `m_processExitThread` after killing the VM.

OneBlue: Discussed offline. The deadlock is between this method and VMProcessControl's destructor

OneBlue (approved):

Issue #14568, opened 2026-03-28
WSL2 DNS: NODATA AAAA responses not cached by systemd-resolved, causing ~50 to 200ms latency per lookup


syed-dawood:
## Summary

When using WSL2 with `dnsTunneling` (the default in mirrored networking mode), every DNS lookup for an IPv4-only hostname incurs an unnecessary full-round-trip penalty (typically 50–200ms depending on environment) because Ubuntu's `systemd-resolved` does not cache NODATA responses for AAAA queries. This is due to Ubuntu's compile-time patch that changes the default `Cache=` setting from upstream's `yes` to `no-negative`, which explicitly skips caching negative/NODATA DNS responses.

Since the DNS tunnel routes queries through the Windows host's DNS infrastructure — adding significant per-query latency — and modern applications perform parallel A + AAAA lookups, every resolution of an IPv4-only hostname pays the full tunnel round-trip cost **on every lookup** because the NODATA AAAA result is never cached.

## Environment

- **Windows**: 11 Pro, Build 26100
- **WSL Version**: 2.4.13.0 (Kernel: 5.15.167.4-1)
- **Distro**: Ubuntu 24.04 LTS (Noble)
- **systemd-resolved**: v255.4-1ubuntu8.14
- **Networking mode**: Mirrored (`networkingMode=mirrored` in `.wslconfig`)
- **DNS Tunneling**: Enabled (default for mirrored mode)

### The DNS Tunnel Architecture

When `dnsTunneling` is enabled, WSL configures:
- A tunnel endpoint at `10.255.255.254/32` on the loopback interface
- `/etc/resolv.conf` → `nameserver 127.0.0.53` (systemd-resolved stub)
- systemd-resolved forwards queries to `10.255.255.254` (the WSL DNS tunnel)

The tunnel routes DNS queries from Linux through the Windows host's DNS infrastructure. This adds significant latency per round-trip (measured at 90–160ms in our environment; varies by DNS server distance, VPN, and network topology).

### The NODATA Problem

When an application resolves a hostname (e.g., via `getaddrinfo()`), the resolver typically sends **both** an A (IPv4) and AAAA (IPv6) query. For IPv4-only hostnames (which are still very common — especially on corporate/enterprise networks), the AAAA query returns a **NODATA** response: RCODE=NOERROR but with zero answer records.

### Ubuntu's `Cache=no-negative` Default

[...]

Issue #13415, opened 2025-08-21
WSL DNS tunnel nameserver (eg. 10.255.255.254) appends search suffixes to IPv6 queries for domains without AAAA records, causing timeouts and mismatches


eyk:
### Windows Version

Microsoft Windows 11 Pro [10.0.26100 Build 26100]

### WSL Version

2.3.26.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

5.15.167.4-1

### Distro Version

Ubuntu 24.04 and Debian 13

### Other Software

GitHub Copilot Chat runs in a timeout, likely because api.github.com hat no IPv6 and this bug causes slow DNS responses. 

1. Other issues may be caused by this one, eg. https://github.com/microsoft/WSL/issues/13391

### Repro Steps

Be aware that the behavior may differ, if you have no DNS search suffix configured on your host. I could only test it with search suffixes. The behavior occurred on newly installed distros. 

When running DNS queries in WSL (tested with Ubuntu and Debian), the virtual nameserver (eg. 10.255.255.254, DNS tunnel mode) appends Windows search suffixes to IPv6 (AAAA) queries for domains that only have IPv4 addresses (e.g., github.com, api.github.com, nbsp.de). This results in repeated, slow DNS requests and "question section mismatch" errors, as shown in the attached screenshots. The suffixes themselves are correct and match our company configuration.

<img width="924" height="344" alt="Image" src="https://github.com/user-attachments/assets/fb4f58e0-0736-4f66-bdcd-a111736f7d16" />

On native Windows (PowerShell), DNS queries for the same domains are fast and do not show mismatches. I tested ist with different Nameservers (public, company internal), result were always correct and quick.

<img width="576" height="189" alt="Image" src="https://github.com/user-attachments/assets/4871809b-ad9c-4cea-87bb-52bdf7e7b7e8" />

[...]

Issue #9423, opened 2023-01-02
DNS resolver sets TTL for all answers to `0`


oliversalzburg:
### Version

Microsoft Windows [Version 10.0.19044.2075]

### WSL Version

- [X] WSL 2
- [ ] WSL 1

### Kernel Version

5.10.102.1-microsoft-standard-WSL2

### Distro Version

Ubuntu 20.04

### Repro Steps

```
dig -x 1.1.1.1 +noall +answer
```

### Expected Behavior

Sub-second response from the DNS resolver on my LAN. On second execution, near-instant response below ~50ms from local DNS cache.

### Actual Behavior

The WSL DNS resolver is very slow and it replaces the TTL with `0`, causing all future requests to go through the same, slow process.

### Diagnostic Logs

```
$ # Local Resolver
$ dig -x 1.1.1.1 @172.22.176.1 +noall +answer
1.1.1.1.in-addr.arpa.   0       IN      PTR     one.one.one.one.

$ # Google DNS
$ dig -x 1.1.1.1 @8.8.8.8 +noall +answer
1.1.1.1.in-addr.arpa.   1006    IN      PTR     one.one.one.one.
```

CatalinFetoiu:
Hi. Please try enabling the experimental "dnsTunneling" feature and let us know if it fixes the issue.

You can find more details at https://devblogs.microsoft.com/commandline/windows-subsystem-for-linux-september-2023-update/

oliversalzburg:
@CatalinFetoiu

> Please note this feature is currently only available to Windows Insiders canary and Release Preview Channel with the latest Windows 11

I don't have that. I feel like my example should be enough for others to verify if the behavior changed at all.

PR #14572, opened 2026-03-28
dns: preserve TTL values and enable systemd-resolved caching

syed-dawood:

`DnsQueryRaw` internally injects `DNS_QUERY_BYPASS_CACHE`, causing all DNS responses tunneled through WSL to bypass the Windows DNS cache. This results in:
1. TTL values reset to their original values, preventing downstream caches from working correctly
2. Combined with systemd-resolved's default `Cache=no-negative`, negative responses (NODATA/NXDOMAIN) are never cached, causing repeated wire queries for unsupported record types (e.g. AAAA on IPv4-only networks)

This PR adds `DNS_QUERY_DONT_RESET_TTL_VALUES` to preserve real TTL values, and enables full DNS caching in systemd-resolved via a drop-in config.

## PR Checklist

- [x] **Closes:** #14568

### Root Cause

`DnsQueryRaw` (dnsapi.dll) always adds `DNS_QUERY_BYPASS_CACHE` to `queryOptions` internally. This was confirmed empirically by calling `DnsQueryRaw` via P/Invoke and inspecting the returned `queryOptions` in `DNS_QUERY_RAW_RESULT` — the `0x8` flag is always present regardless of input flags.

This means:
- Every DNS query goes to wire (no Windows DNS cache hit)
- TTL values in responses are reset to their original values (not decremented)
- `DNS_PROTOCOL_NO_WIRE` is never returned

### Changes

**1. `src/windows/common/DnsResolver.cpp` — Add `DNS_QUERY_DONT_RESET_TTL_VALUES`**

Sets the `DNS_QUERY_DONT_RESET_TTL_VALUES` (`0x00100000`) flag alongside the existing `DNS_QUERY_NO_MULTICAST`. This tells the Windows DNS client to preserve real decremented TTL values from upstream resolvers, enabling systemd-resolved to cache responses with meaningful lifetimes.

This is safe because `HandleDnsQueryCompletion` performs a raw `CopyMemory` of response bytes — no WSL code inspects or modifies TTL values in the DNS response.

**2. `src/linux/init/init.cpp` — Enable `Cache=yes` in systemd-resolved**

Writes a drop-in config to `/run/systemd/resolved.conf.d/wsl-dns-cache.conf` changing systemd-resolved from `Cache=no-negative` (default) to `Cache=yes`. This enables caching of both positive and negative DNS responses.

- Written to `/run/` (tmpfs) so it's regenerated each boot and user overrides in `/etc/` take precedence
- Harmless no-op on distros without systemd-resolved
- Uses existing `UtilMkdirPath()`/`WriteToFile()` pattern from the same function

### Related Issues

- Related: #9423 — TTL=0 in DNS responses
- Related: #13415 — Slow DNS resolution with tunneling enabled
- Related: #14532 — DNS over TCP debugging (virtio net DNS revert)

### Empirical DnsQueryRaw testing (Windows)

A [C# test program](https://github.com/syed-dawood/WSL/blob/fix/dns-cache-ttl-preservation/tools/test/DnsRawCacheTest.cs) calling `DnsQueryRaw` via P/Invoke confirmed:

| Test | Flags | Result |
|------|-------|--------|
| `NO_MULTICAST` only | returned `0x80000800` | TTL=60 (reset to original) |
| `+ DONT_RESET_TTL` | returned `0x80100800` | TTL=58→55 (decremented = real) |
| `NO_WIRE_QUERY` | `ERROR_INVALID_PARAMETER` | Not supported by DnsQueryRaw |
| AAAA NODATA (ipv4only.arpa) | every query hits wire | 31 bytes, 0 answers, proto=UDP |

### DNS benchmarking (WSL)

Dual-stack DNS benchmark (`dig A` + `dig AAAA` for 7 domains):

| Metric | Before (Cache=no-negative) | After (Cache=yes) |
|--------|---------------------------|-------------------|
| Average per-query latency | ~108ms | ~13ms |
| AAAA NODATA queries | 51-219ms (wire every time) | 0ms (cached) |
| Cache misses | 14/14 | 2/14 |


syed-dawood:
cc @benhillis @OneBlue @craigloewen-msft @jstarks @damanm24 — requesting review on this DNS performance fix.

This addresses the `DnsQueryRaw` cache bypass behavior that causes all DNS responses tunneled through WSL to hit the wire on every query. The fix is two complementary changes:

1. **`DNS_QUERY_DONT_RESET_TTL_VALUES`** in `DnsResolver.cpp` — preserves real TTL values so downstream caching works correctly
2. **`Cache=yes` drop-in** in `init.cpp` — enables full DNS response caching in systemd-resolved (default `no-negative` discards NODATA/NXDOMAIN responses)

Both changes were empirically validated. See the PR description for detailed test results and safety analysis.

src/linux/init/init.cpp
```diff
@@ -336,6 +336,13 @@ int GenerateSystemdUnits(int Argc, char** Argv)
         // Mask NetworkManager-wait-online.service for the same reason, as it causes timeouts on distros using NetworkManager.
         THROW_LAST_ERROR_IF(symlink("/dev/null", std::format("{}/NetworkManager-wait-online.service", installPath).c_str()) < 0);
 
+        // Enable DNS response caching in systemd-resolved. The default Cache=no-negative discards negative (NODATA/NXDOMAIN)
+        // responses, which causes repeated wire queries for unsupported record types (e.g. AAAA on IPv4-only networks).
+        // With DnsQueryRaw bypassing the Windows DNS cache, systemd-resolved is the only caching layer available.
```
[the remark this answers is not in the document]
syed-dawood: Good point — updated the comment in the latest push to clarify this is distro-specific rather than a universal default. The wording now reads:

```cpp
// Enable DNS response caching in systemd-resolved. On some distros (notably Ubuntu) the default
// Cache=no-negative discards negative (NODATA/NXDOMAIN) responses, which causes repeated wire
// queries for unsupported record types (e.g. AAAA on IPv4-only networks). This override aligns
// all distros with the upstream systemd default of Cache=yes, ensuring systemd-resolved provides
// effective DNS caching inside the VM.
```

src/linux/init/init.cpp
```diff
@@ -336,6 +336,13 @@ int GenerateSystemdUnits(int Argc, char** Argv)
...
+        THROW_LAST_ERROR_IF(UtilMkdirPath("/run/systemd/resolved.conf.d", 0755) < 0);
+        constexpr auto* resolvedCacheConfig = "[Resolve]\nCache=yes\n";
+        THROW_LAST_ERROR_IF(WriteToFile("/run/systemd/resolved.conf.d/wsl-dns-cache.conf", resolvedCacheConfig) < 0);
```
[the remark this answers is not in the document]
syed-dawood: Good catch on the `O_TRUNC` observation. However, this is safe in practice for this specific case:

1. The file is in `/run/` (tmpfs), which is cleared every boot
2. This code runs in a **systemd generator** (`wsl-generator`) that executes once during early boot — the file doesn't exist yet at that point
3. The new content (`"[Resolve]\nCache=yes\n"`) is always the same fixed string, so even in a hypothetical re-run scenario, the same bytes would be written
4. This is the **same pattern** used by every other `WriteToFile()` call in this function (e.g., `x11.conf`, `override.conf`) and throughout the codebase — none use `O_TRUNC`

Changing `WriteToFile` to add `O_TRUNC` would be a separate concern affecting 20+ call sites and is outside the scope of this fix.

syed-dawood:

benhillis:
Hey @syed-dawood 👋 — Following up on this PR. All review threads are resolved and this looks like a clean, well-documented DNS perf fix. However, the **full CI pipeline hasn't run** yet (only CLA check passed). A maintainer will need to trigger the build pipeline. Is this change still needed on your end?

syed-dawood:
Hey @benhillis 👋 — Yes, this change is still needed! Just merged master to resolve the conflict (the `console-getty.service` mask landed in the same region as the DNS cache config block — both are now cleanly integrated).

The PR is ready for CI whenever a maintainer can trigger it. Happy to address any additional feedback.

src/linux/init/init.cpp
```diff
@@ -340,6 +340,15 @@ int GenerateSystemdUnits(int Argc, char** Argv)
         // When multiple distros are running, the second distro's getty fails because the tty is already held.
         THROW_LAST_ERROR_IF(symlink("/dev/null", std::format("{}/console-getty.service", installPath).c_str()) < 0);
 
+        // Enable DNS response caching in systemd-resolved. On some distros (notably Ubuntu) the default
+        // Cache=no-negative discards negative (NODATA/NXDOMAIN) responses, which causes repeated wire
+        // queries for unsupported record types (e.g. AAAA on IPv4-only networks). This override aligns
+        // all distros with the upstream systemd default of Cache=yes, ensuring systemd-resolved provides
+        // effective DNS caching inside the VM.
+        THROW_LAST_ERROR_IF(UtilMkdirPath("/run/systemd/resolved.conf.d", 0755) < 0);
+        constexpr auto* resolvedCacheConfig = "[Resolve]\nCache=yes\n";
```
OneBlue, line 364: I don't think we should try to change the resolved configuration here. When DNS tunneling is enabled, WSL writes its own resolver IP address to `/etc/resolv.conf`, so even if resolved is enabled, it will be bypassed

syed-dawood: @OneBlue — You're right. With the default `generateResolvConf=true`, WSL writes `nameserver 10.255.255.254` directly to `/etc/resolv.conf` (via `ConfigCreateResolvConfSymlink` + `ConfigUpdateNetworkInformation`), so systemd-resolved is never in the query path even if it's running. The `Cache=yes` drop-in would be a no-op for most users.

My testing was done with `generateResolvConf=false` and `/etc/resolv.conf` manually symlinked to the systemd-resolved stub (`127.0.0.53`), which puts resolved in the path and makes the drop-in effective — but that's a non-default setup.

I'll drop the `init.cpp` change from this PR. The TTL fix in `DnsResolver.cpp` (`DNS_QUERY_DONT_RESET_TTL_VALUES`) is independent and still valuable — it ensures any downstream cache (whether resolved, nscd, or a future WSL-side mechanism) receives real decremented TTL values instead of reset-to-original ones.

benhillis:

benhillis:
@syed-dawood Hey! This PR still has merge conflicts from the last master merge. Could you rebase/resolve when you get a chance? The change looks good and we'd like to get it in. Thanks!

src/linux/init/init.cpp
keith-horton: The DnsResolver.cpp looks good - I'm going to confirm with the dev owner of the DNS API. But I agree with the response that we don't want to make these systemd related changes in init

PR #40031, opened 2026-03-30
Validate empty host path in VolumeMount::Parse

benhillis:
When parsing volume mount specs without a mode suffix (e.g. ':container'), the host path validation was skipped, producing a VolumeMount with an empty host path. This would propagate to MountWindowsFolder('') and produce a confusing error. Add validation after computing m_hostPath to reject empty host paths with a clear error message.


OneBlue (approved):
LGTM. We just need to update the tests to cover this case

test/windows/wslc/WSLCVolumeMountUnitTests.cpp
```diff
@@ -33,9 +33,7 @@ class WSLCVolumeMountUnitTests
             {LR"(C:\hostPath:/containerPath:rw)", LR"(C:\hostPath)", R"(/containerPath)", false},
             {LR"(C:\host Path:/container Path:ro)", LR"(C:\host Path)", R"(/container Path)", true},
             {LR"(C:\host Path:/container Path:rw)", LR"(C:\host Path)", R"(/container Path)", false},
-            {LR"(:/containerPath)", LR"()", R"(/containerPath)", false},
             {LR"(C:\hostPath::ro)", LR"(C:\hostPath)", R"()", true},
```
AmelBawa-msft, line 36: Should we also consider throwing when container path is empty?

PR #40035, opened 2026-03-30
Fix WaitForMultipleObjects exceeding 64-handle limit in port relay

benhillis:
AcceptThread passes all port accept events plus the exit event to WaitForMultipleObjects, which has a hard limit of 64 handles. With 64+ mapped ports, the call fails and crashes the relay thread, making ALL port mappings inaccessible. When the handle count exceeds MAXIMUM_WAIT_OBJECTS, poll in batches with a 100ms timeout.

### DIFF
--- a/src/windows/wslrelay/localhost.cpp
+++ b/src/windows/wslrelay/localhost.cpp
@@ -487,7 +487,7 @@ void AcceptThread(std::vector<std::shared_ptr<PortRelay>>& ports, const GUID& Vm
             events.push_back(e->AcceptEvent.get());
         }
 
-        // Then wait for IO, or exit event.
+        // WaitForMultipleObjects supports at most MAXIMUM_WAIT_OBJECTS (64) handles.
         auto result = WaitForMultipleObjects(static_cast<DWORD>(events.size()), events.data(), false, INFINITE);
         THROW_LAST_ERROR_IF(result == WAIT_FAILED);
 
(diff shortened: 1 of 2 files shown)


benhillis:
I don’t love this approach but this is a real issue. Maybe we can just fail with over 64 for now?

ptrivedi:
> I don’t love this approach but this is a real issue. Maybe we can just fail with over 64 for now?



> I don’t love this approach but this is a real issue. Maybe we can just fail with over 64 for now?

Ah, this new approach is much better than the throw inside AcceptThread. This will proactively never allow ports to grow beyond the limit and AcceptThread will stay healthy. 

I was looking up the epoll equivalent in Windows, and it seems to be the IOCP way -- this could be a heavier lift. And while limiting to 63 ports is a safe stopgap for now, I wonder if using RegisterWaitForSingleObject and then having a separate threadpool for each port's accept handling might be a lower lift than IOCP in the shorter term... and would likely address the issues.

benhillis:
> > I don’t love this approach but this is a real issue. Maybe we can just fail with over 64 for now?
> 
> > I don’t love this approach but this is a real issue. Maybe we can just fail with over 64 for now?
> 
> Ah, this new approach is much better than the throw inside AcceptThread. This will proactively never allow ports to grow beyond the limit and AcceptThread will stay healthy.
> 
> I was looking up the epoll equivalent in Windows, and it seems to be the IOCP way -- this could be a heavier lift. And while limiting to 63 ports is a safe stopgap for now, I wonder if using RegisterWaitForSingleObject and then having a separate threadpool for each port's accept handling might be a lower lift than IOCP in the shorter term... and would likely address the issues.

Yeah I think we should switch to IOCP, and I'm working on doing that but it is definitely more complicated.

Also we are planning on moving the port relay logic into the virtio networking implementation so maybe this is good enough for now. Will chat with @OneBlue once he's online.

OneBlue (commented):
Hmm I wasn't aware that the handle limit was that low. 

The best way to fix this would probably be to move to IO completion ports to do our IO. 

Approving so we can solve this crash for now

OneBlue (approved):

PR #40039, opened 2026-03-30
wslc: fix TOCTOU race in cached COM wrapper getters

benhillis:
Fixes a race in the cached COM wrapper getters (`GetState`, `GetInitProcess`, `GetId`, `GetName`) where `DisconnectComWrapper()` could populate the cache and disconnect the impl between the initial cache check and the forwarded call, causing the getter to fail with `RPC_E_DISCONNECTED` even though the cached value was available.

Changes:

- Removes the upfront cache check from all four getters.
- Calls through to the impl first; if the call fails with `RPC_E_DISCONNECTED`, falls back to the cache which `DisconnectComWrapper()` guarantees is populated.
- `GetId` and `GetName` use `wil::ResultFromException` to convert the `LockImpl` exception into an HRESULT before checking for `RPC_E_DISCONNECTED`.

### DIFF
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -1531,17 +1531,26 @@ HRESULT WSLCContainer::GetState(WSLCContainerState* Result)
     COMServiceExecutionContext context;
     RETURN_HR_IF_NULL(E_POINTER, Result);
 
+    *Result = WslcContainerStateInvalid;
+    HRESULT hr = CallImpl(&WSLCContainerImpl::GetState, Result);
+    if (SUCCEEDED(hr))
+    {
+        return S_OK;
+    }
+
+    // DisconnectComWrapper() populates the cache before setting m_impl to null,
+    // so if CallImpl failed with RPC_E_DISCONNECTED, the cache must be populated.
+    if (hr == RPC_E_DISCONNECTED)
     {
         auto cacheLock = m_cacheLock.lock_shared();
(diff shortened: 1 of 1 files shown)


src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1531,17 +1531,36 @@ HRESULT WSLCContainer::GetState(WSLCContainerState* Result)
...
+    auto state = getCachedState();
+    if (state)
+    {
+        *Result = state.value();
+        return S_OK;
     }
 
     *Result = WslcContainerStateInvalid;
-    return CallImpl(&WSLCContainerImpl::GetState, Result);
+    HRESULT hr = CallImpl(&WSLCContainerImpl::GetState, Result);
```
OneBlue, line 1535: nit: I would have a preference for just calling CallImpl() and then looking at the cache if the result was `RPC_E_DISCONNECTED` since that would simplify things a bit. 

Especially given that we should implement the same logic in GetInitProcess(), GetId() and GetName()

benhillis: ok yeah I'm good with that. I assumed you were using the cache as a perf improvement but I like that design more anyway.

OneBlue: Yeah "Cache" is a bit misleading here. Maybe we should rename to "FinalState" or "TerminalState" ?

kvega005 (approved):

PR #40041, opened 2026-03-30
[WSLC] Add --workdir / -w option to 'wslc exec'

ptrivedi:

- Adds `--workdir` / `-w` option to `wslc exec` to set the working directory inside the container
- Passes the value through `SetContainerOptionsFromArgs` into `ContainerOptions.WorkingDirectory`
- Forwards the working directory to the container service via `ContainerService::Exec`
- Validates that the `--workdir` value is non-empty or all-whitespace using `std::iswspace` (covers all Unicode whitespace characters)

## PR Checklist

**Changes:**

- `ArgumentDefinitions.h`: register `WorkDir` argument with long form `--workdir` and short alias `-w` (appended at end to preserve sorted order of X-macro list)
- `ContainerExecCommand.cpp`: include `WorkDir` in the exec command's argument list
- `ContainerModel.h`: add `WorkingDirectory` field to `ContainerOptions`
- `ContainerTasks.cpp`: read `WorkDir` arg and populate `options.WorkingDirectory`
- `ContainerService.cpp`: pass `WorkingDirectory` when calling exec on the container
- `ArgumentValidation.cpp`: validate `--workdir` is non-empty/non-whitespace using `std::all_of` with `std::iswspace` (with explicit `wint_t` cast) to correctly handle all Unicode whitespace, not just space and tab

**Tests:**

- `CommandLineTestCases.h`: parsing cases for `--workdir` and `-w` (valid with option before positional, missing value, invalid specifier after positional)
- `WSLCCLIExecutionUnitTests.cpp`: unit tests covering:
  - `SetContainerOptionsFromArgs` sets `WorkingDirectory` when `--workdir` is provided
  - `SetContainerOptionsFromArgs` leaves `WorkingDirectory` empty when not provided
  - End-to-end parse of `--workdir` long option sets `WorkingDirectory`
  - End-to-end parse of `-w` short alias sets `WorkingDirectory`
  - Empty `--workdir` value throws `ArgumentException`
- `WSLCE2EContainerExecTests.cpp`: E2E tests covering:
  - Help message includes `-w,--workdir`
  - `exec --workdir /tmp <container> pwd` prints `/tmp`
  - `exec -w /tmp <container> pwd` prints `/tmp`

## Validation Steps Performed

- [x] Build and run unit tests: `bin\<platform>\<target>\test.bat /name:*WSLCCLIExecutionUnitTests*`
- [x] Verify `wslc exec --workdir /app <container> pwd` prints `/app`
- [x] Verify `wslc exec -w /tmp <container> pwd` prints `/tmp`
- [x] Verify `wslc exec <container> pwd` uses the container's default working directory


src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -78,6 +78,7 @@ _(TTY,            "tty",                 L"t",              Kind::Flag,        L
 _(User,           "user",                L"u",              Kind::Value,       L"User ID for the process (name|uid|uid:gid)") \
 _(Verbose,        "verbose",             L"v",              Kind::Flag,        L"Output verbose details") \
 _(Version,        "version",             L"v",              Kind::Flag,        L"Show version information for this tool") \
+_(WorkDir,        "workdir",             L"w",              Kind::Value,       L"Working directory inside the container") \
```
AmelBawa-msft: Nit: Consider appending WorkDir at the end to preserve the sorted characteristics of the X-macro list

AmelBawa-msft (commented):
LGTM! I suspect E2E tests will fail because we need to update the expected E2E test help messages. We should consider adding E2E tests as well but possibly in a separate PR since we still have a lot more E2E tests to write 😊

test/windows/wslc/WSLCCLIExecutionUnitTests.cpp
```diff
@@ -116,6 +118,62 @@ class WSLCCLIExecutionUnitTests
         // This one will just verify all the data types in the Data Map work as expected.
     }
 
+    // Test: SetContainerOptionsFromArgs sets WorkingDirectory when --workdir is provided
+    TEST_METHOD(SetContainerOptionsFromArgs_WithWorkDir_SetsWorkingDirectory)
```
dkbennett, line 122: Clever unit tests here to verify the settings, we should add more like these later.

src/windows/wslc/arguments/ArgumentDefinitions.h
dkbennett: +1 would prefer to keep this list sorted

dkbennett (approved):
Looks good, like the setting unit tests.  An E2E with working directory would be nice but this looks like we are correctly setting the value in the CLI and passing it into the container runtime. Assuming that has tests this should be good, though it doesn't hurt to have a CLI E2E or two for this.

nit: sort order of _WorkDir as Amir mentioned.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -46,6 +46,16 @@ void Argument::Validate(const ArgMap& execArgs) const
         validation::ValidateVolumeMount(execArgs.GetAll<ArgType::Volume>());
         break;
 
+    case ArgType::WorkDir:
+    {
+        const auto& value = execArgs.Get<ArgType::WorkDir>();
+        if (value.empty() || value.find_first_not_of(L" \t") == std::wstring::npos)
+        {
+            throw ArgumentException(std::format(L"Invalid {} argument value: working directory cannot be empty or whitespace", m_name));
```
OneBlue, line 56: nit: I don't recommend validating the directory on the CLI path. If we want to add checks for this, we should do it in the service to it applies to the SDK as well

ptrivedi: Yeah, thought about this. It's mainly the empty check so we could short-circuit faster, service should do it anyway. Don't have a strong opinion either way.

ptrivedi:
> LGTM! I suspect E2E tests will fail because we need to update the expected E2E test help messages. We should consider adding E2E tests as well but possibly in a separate PR since we still have a lot more E2E tests to write 😊

Added E2E tests. @dkbennett @AmelBawa-msft PTAL and please hit merge if good. thanks!

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -47,6 +47,16 @@ void Argument::Validate(const ArgMap& execArgs) const
...
+        }
```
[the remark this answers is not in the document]
ptrivedi: @copilot apply changes based on [this feedback](https://github.com/microsoft/WSL/pull/40041#discussion_r3034425791)

PR #40042, opened 2026-03-30
wslc: Switch port relay AcceptThread from WaitForMultipleObjects to IO completion ports

benhillis:
Replace the WaitForMultipleObjects-based wait loop in `relay::MultiHandleWait` with an IO completion port (IOCP), removing the `MAXIMUM_WAIT_OBJECTS` (64) handle limit for all callers (IORelay, container IO, socket accept, etc.).

Switch the port relay `AcceptThread` from `WaitForMultipleObjects` to a direct IOCP, removing the 63-port mapping limit.

### relay::MultiHandleWait changes (relay.hpp/cpp)
- Add `unique_registered_wait` RAII type for `RegisterWaitForSingleObject` handles
- Add `Register()` pure virtual to `OverlappedIOHandle`, implemented by all subclasses
- Handles that support IOCP (pipes, sockets) register directly; handles that don't (console handles, events) use a `RegisterWaitForSingleObject` bridge
- Convert `Run()` to use `GetQueuedCompletionStatus` instead of `WaitForMultipleObjects`
- `Cancel()` posts a key=0 completion to wake `Run()`; key=0 sets `m_cancel` on dequeue
- Fix member order: `m_iocp` declared before `m_handles` for correct destruction

### Port relay changes (localhost.cpp)
- Associate listen sockets directly with an IOCP via `CreateIoCompletionPort`
- `AcceptEx` completions go straight to the IOCP — no events or thread pool waits
- Stop signal via `PostQueuedCompletionStatus` key=0
- Remove `AcceptEvent` from `PortRelay` (no longer needed)
- Remove the `MAXIMUM_WAIT_OBJECTS` port limit check


OneBlue (commented):
I think this is a good idea, but given that the relay code is not our long term plan, I would recommend making this change at the `relay::MultiHandleWait` level so it will also solve the 64 handles issue in other places in the code (specifically the container IO thread, which could go over the limit if many Logs() / Exec() calls are made)

If we need to, we can then move the relay to use `MultiHandleWait` and match the rest of the code

benhillis:
> I think this is a good idea, but given that the relay code is not our long term plan, I would recommend making this change at the `relay::MultiHandleWait` level so it will also solve the 64 handles issue in other places in the code (specifically the container IO thread, which could go over the limit if many Logs() / Exec() calls are made)
> 
> If we need to, we can then move the relay to use `MultiHandleWait` and match the rest of the cod

That's a good call, let me look at that.

benhillis:
Hey @benhillis 👋 — Following up on this PR. The \\wsl-github-pr\\ check shows **action_required** status. There are also **2 unresolved review threads** remaining (out of 32 total):
- \\AddHandle\\/\\Register\\ ordering concern
- \\Cancel()\\ no longer setting the \\m_cancel\\ flag

This is a large PR (3170+/1359−, 90 files) — great progress resolving 30 of 32 threads! Just those last 2 items and the CI status to sort out.

PR #40043, opened 2026-03-30
Add repository, tag, ID and created columns to image list output

beena352:
Enhance `wslc image ls` output to display REPOSITORY, TAG, IMAGE ID, CREATED, SIZE.
image
image

## PR Checklist

- [ x] **Communication:** I've discussed this with core contributors already. If work hasn't been agreed, this work might be rejected

- Extended ImageInformation with new fields: Repository, Tag, Id, and Created
- Updated ImageService::List() to use the existing ParseImage() helper and populate fields that were previously ignored
- Truncated image ID to 12 characters and reused ContainerService::FormatRelativeTime() for consistency
- Adjusted E2E tests and helpers to handle separate repository and tag columns


src/windows/wslc/services/ImageService.cpp
```diff
@@ -89,7 +89,11 @@ std::vector<ImageInformation> ImageService::List(wsl::windows::wslc::models::Ses
     {
         const WSLCImageInformation& image = *ptr;
         ImageInformation info{};
-        info.Name = image.Image;
+        auto parsed = wsl::windows::common::wslutil::ParseImage(image.Image);
+        info.Repository = parsed.first;
+        info.Tag = parsed.second.value_or("latest");
```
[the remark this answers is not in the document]
yao-msft: For dangling image, the service api will return `<none>:<none>`, so it's not a problem here

yao-msft: I'm not sure if we want to append latest as default for list output

beena352: Yeah, I'll change the fallback to <none> to match Docker's behavior. Images without an explicit tag should
  show <none> in the TAG column, not latest.

src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -72,7 +73,7 @@ void ListImages(CLIExecutionContext& context)
         // Print only the image names.
         for (const auto& image : images)
         {
-            PrintMessage(MultiByteToWide(image.Name));
+            PrintMessage(MultiByteToWide(image.Repository + ":" + image.Tag));
         }
 
         return;
```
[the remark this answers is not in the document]
dkbennett: We have a description override for arguments in the Create method. The description in the ArgumentDefinitions.h is just a default. Quiet is one that makes sense might have different variations depending on context nad would use the description overload. It is used in a few places and addresses this copilot feedback.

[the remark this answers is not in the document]
beena352: Sounds good.
I’m also waiting for your PR to merge since there’s some overlap, I’ll updraft mine after that.

dkbennett (approved):

src/windows/wslc/services/ImageService.cpp
```diff
@@ -145,7 +145,23 @@ std::vector<ImageInformation> ImageService::List(wsl::windows::wslc::models::Ses
...
         const WSLCImageInformation& image = *ptr;
         ImageInformation info{};
-        info.Name = image.Image;
+
+        // Parse the image reference, but handle dangling images like "<none>:<none>"
+        std::string imageRef = image.Image;
+        if (imageRef == "<none>:<none>")
+        {
+            info.Repository = "<none>";
+            info.Tag = "<none>";
```
OneBlue: Instead of having `<none>`, could we make the repository and tag into std::optional, and when they're missing we could show `<untagged>`, so that that our format would be compatible with docker's ?

src/windows/wslc/services/ImageService.cpp
```diff
@@ -145,7 +145,23 @@ std::vector<ImageInformation> ImageService::List(wsl::windows::wslc::models::Ses
...
+        }
+        else
+        {
+            auto parsed = wsl::windows::common::wslutil::ParseImage(imageRef);
+            info.Repository = parsed.first;
+            info.Tag = parsed.second.value_or("<none>");
```
OneBlue: Is it possible for an image to have a repo and not a tag ?

OneBlue (commented):
Change looks good, two minor comments

test/windows/wslc/e2e/WSLCE2EImageListTests.cpp
```diff
@@ -50,7 +50,7 @@ class WSLCE2EImageListTests
     {
         WSL2_TEST_ONLY();
 
-        const auto result = RunWslc(L"image list");
+        const auto result = RunWslc(L"image list -q");
```
AmelBawa-msft: This method tests the list subcommand without options. Does the test fail without -q?

The next method tests the quiet option: WSLCE2E_Image_List_QuietOption_OutputsNamesOnly

beena352: Good catch, reverted to image list. The test now searches for Name and Tag separately since they're in different columns with the new table format.

src/windows/wslc/services/ImageService.cpp
beena352: Done, Repository and Tag are now std::optional<std::string>.

src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -97,14 +98,25 @@ void ListImages(CLIExecutionContext& context)
...
+                           : wsl::windows::wslc::TableOutput<5>({L"REPOSITORY", L"TAG", L"IMAGE ID", L"CREATED", L"SIZE"});
 
         for (const auto& image : images)
         {
-            table.OutputLine({MultiByteToWide(image.Name), std::format(L"{:.2f} MB", static_cast<double>(image.Size) / (1024 * 1024))});
+            table.OutputLine({
+                MultiByteToWide(image.Repository.value_or("<untagged>")),
+                MultiByteToWide(image.Tag.value_or("<untagged>")),
+                MultiByteToWide(TruncateId(image.Id, trunc)),
+                ContainerService::FormatRelativeTime(image.Created > 0 ? static_cast<ULONGLONG>(image.Created) : 0),
```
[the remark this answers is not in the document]
OneBlue: I guess this isn't supposed to happen ? Probably not a huge deal

OneBlue (approved):

PR #40047, opened 2026-03-31
CLI: Create elevated and non-elevated sessions

dkbennett:

The CLI now has two default sessions: 'wslc-cli' and 'wslc-cli-admin'. Which session is created depends on the elevation of the process creating the session. 

The following is a non-elevated command prompt executing commands and session targeting. Prior to this change if the session was created elevated this command prompt was completely non-functional due to permissions. Now with the change it can access its own session as-expected but cannot access the admin session (as expected).
image

If an elevated process creates a session, it will be the wslc-cli-admin session. If a non-elevated process creates the session, it will be the wslc-cli session. By default, elevation determines which session the CLI targets. If a session that is not the default for the process elevation is specified, it will use OpenSession, which falls back on token permissions for access. 

So for example, an admin cli process can open the wslc-cli session if it exists, but it cannot create it. A non-elevated cli process can create and open the wslc-cli session but it will get access denied if it tries to target the admin session. 

## PR Checklist

* Session root path changed slightly to avoid collision since it is nested deeper. "wslc\sessions" by default. Under the root path we will store the sessions by name, either as wslc-cli or wslc-cli-admin depending on whichever is selected.

* All logic of selecting the session and elevation information is encapsulated in the SessionOptions class. Two public methods were exposed to replace the direct static string checking - GetDefaultSessionName() and IsDefaultSessionName(). These are elevation-aware methods that return the appropriate result.

* The SessionTasks were updated to check if the provided session is the default session name, and if so, it calls create with default settings. Otherwise it calls Open session. 

* Tests were updated to enable testing of elevation scenarios. RunWslc and RunWslcInteractive now have an ElevationType parameter (defaults to Elevated) to enable specific testing of the elevated commands.

* Added tests for verifying the session for elevated and non-elevated, and verifying non-elevated cannot access an elevated-created session.

* Moved the Session Shell test to be in the Global Tests with the other session tests and updated it to account for now having two types of sessions.  This test was updated to reflect the new session names and elevation behavior.

## Validation Steps Performed
* Added E2E tests to verify elevation behavior.
* Updated the E2E test code to allow for more robust elevation testing in the future as-needed.
* Manual verification of the results (screenshot at the top).


src/windows/wslc/services/SessionModel.cpp
```diff
@@ -39,4 +41,25 @@ SessionOptions::SessionOptions()
...
+{
+    static const std::filesystem::path basePath = []() {
+        return settings::User().Get<settings::Setting::SessionStoragePath>().empty()
+                   ? std::filesystem::path{wsl::windows::common::filesystem::GetLocalAppDataPath(nullptr) / SessionOptions::s_defaultStorageSubPath}
+                   : settings::User().Get<settings::Setting::SessionStoragePath>().c_str();
+    }();
+
+    static const std::filesystem::path storagePathNonAdmin = basePath / std::wstring{s_defaultSessionName};
+    static const std::filesystem::path storagePathAdmin = basePath / std::wstring{s_defaultAdminSessionName};
+
```
kvega005, line 65: Why not just return `return basePath / std::wstring{GetDefaultSessionName()};`

src/windows/wslc/tasks/SessionTasks.cpp
```diff
@@ -46,17 +46,22 @@ void CreateSession(CLIExecutionContext& context)
...
+        // a session and not create it. Example: Admin process attempting to open
+        // a non-admin session will fail to create but succeed to open, preventing
+        // accidental creation of a non-admin session with admin permissions.
         const auto& sessionName = context.Args.Get<ArgType::Session>();
-        if (!wsl::shared::string::IsEqual(sessionName, SessionOptions::s_defaultSessionName))
+        if (!SessionOptions::IsDefaultSessionName(sessionName))
         {
             context.Data.Add<Data::Session>(SessionService::OpenSession(sessionName));
             return;
         }
```
kvega005, line 59: Does this mean that if a non-elevated user targets wslc-cli-admin before it exists, they'd get ERROR_NOT_FOUND thrown as an exception rather than the ERROR_ELEVATION_REQUIRED? We should test what happens if the wslc-cli-admin session has not been created yet.

test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
```diff
@@ -49,6 +49,70 @@ class WSLCE2EGlobalTests
...
+        // First ensure admin session is created by running container list.
+        auto result = RunWslc(L"container list", ElevationType::Elevated);
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+        // Try to explicitly target the admin session from non-elevated process
+        result = RunWslc(L"container list --session wslc-cli-admin", ElevationType::NonElevated);
+
+        // Should fail with access denied.
+        result.Verify({.Stderr = L"The requested operation requires elevation. \r\nError code: ERROR_ELEVATION_REQUIRED\r\n", .ExitCode = 1});
+    }
```
kvega005, line 99: I mentioned in another comment, but I would like to see a version of this test that verifies running `container list --session wslc-cli-admin` as a `NonElevated` user still fails even if the admin session doesn't exist. Whatever the expected behavior is there, we should make sure its deterministic.

kvega005 (commented):
LGTM, left a few minor comments.

src/windows/wslc/services/SessionModel.cpp
dkbennett: These are static to avoid maintaining the strings since they are not going to change for the lifetime of the sessions, and the WSLCSessionSettings structure has a pointer to the path, so we have to maintain the lifetime of the data to the lifetime of the WSLCSessionSettings object. IMO it would be better for the WSLCSessionSettings to manage that lifetime itself but this solution works well and is efficient.

src/windows/wslc/tasks/SessionTasks.cpp
dkbennett: Yes that is correct, it should get ERROR_NOT_FOUND if it has not yet been created.

I plan to expand upon the tests in this PR once we have Session::Terminate() that should remove and clean up a session, and then we can more fully test this and ensure that is the case. I thought it best to wait and make that part of the Session::Terminate() testing, which is also a priority for either myself or blue in the next day or so.

test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
dkbennett: Totally agree and I wanted to add those initially, but we do not have Session Terminate functionallity yet but it is next on my list so I will expand upon this as part of the terminate tests.

src/windows/wslc/services/SessionModel.cpp
```diff
@@ -39,4 +41,25 @@ SessionOptions::SessionOptions()
     m_sessionSettings.FeatureFlags = WslcFeatureFlagsVirtioFs;
 }
 
+bool SessionOptions::IsElevated()
+{
+    wil::unique_handle token;
+    THROW_IF_WIN32_BOOL_FALSE(::OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &token));
```
OneBlue: nit: We could use `wil::open_current_access_token();` here

OneBlue (approved):
LGTM, minor comment

src/windows/wslc/services/SessionModel.cpp
dkbennett: I'll update it when I add in Terminate to save the turnaround time of the churn.
edit: nevermind, I have to churn anyway due to merge conflicts, will fix it with the merge conflicts.

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -42,7 +42,7 @@ static constexpr std::string_view s_DefaultSettingsTemplate =
     "  # Maximum disk image size in megabytes (default: 100GB)\n"
     "  # maxStorageSize: 100GB\n"
     "\n"
-    "  # Default path for container storage (default: %LocalAppData%\\wslc\\storage)\n"
+    "  # Default path for container storage (default: %LocalAppData%\\wslc\\sessions)\n"
```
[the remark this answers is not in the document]
benhillis: I think this is worth updating.

src/windows/wslc/services/SessionModel.cpp
```diff
@@ -39,4 +41,24 @@ SessionOptions::SessionOptions()
     m_sessionSettings.FeatureFlags = WslcFeatureFlagsVirtioFs;
 }
 
+bool SessionOptions::IsElevated()
+{
+    auto token = wil::open_current_access_token(TOKEN_QUERY);
+    return wsl::windows::common::security::IsTokenElevated(token.get());
+}
```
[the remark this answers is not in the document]
dkbennett: Good catch copilot, fixed this after consulting with blue.

src/windows/wslc/settings/UserSettings.cpp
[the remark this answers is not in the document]
dkbennett: Fixed.

AmelBawa-msft (approved):

src/windows/wslc/settings/UserSettings.cpp
```diff
@@ -42,11 +42,13 @@ static constexpr std::string_view s_DefaultSettingsTemplate =
     "  # Maximum disk image size in megabytes (default: 100GB)\n"
     "  # maxStorageSize: 100GB\n"
     "\n"
-    "  # Default path for container storage (default: %LocalAppData%\\wslc\\storage)\n"
-    "  # defaultStoragePath: \"\"\n";
+    "  # Default path for session storage. By default, storage is per-session under:\n"
+    "  #   %LocalAppData%\\wslc\\sessions\\wslc-cli        (standard sessions)\n"
+    "  #   %LocalAppData%\\wslc\\sessions\\wslc-cli-admin (elevated sessions)\n"
```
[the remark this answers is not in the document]
dkbennett: fixed

test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
```diff
@@ -49,6 +49,70 @@ class WSLCE2EGlobalTests
...
+
+        // First ensure admin session is created by running container list.
+        auto result = RunWslc(L"container list", ElevationType::Elevated);
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+        // Try to explicitly target the admin session from non-elevated process
+        result = RunWslc(L"container list --session wslc-cli-admin", ElevationType::NonElevated);
+
+        // Should fail with access denied.
+        result.Verify({.Stderr = L"The requested operation requires elevation. \r\nError code: ERROR_ELEVATION_REQUIRED\r\n", .ExitCode = 1});
```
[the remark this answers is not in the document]
dkbennett: I agree but this is how we have been doing most of our tests with as exact matching as possible. It does make them more brittle but it also catches any changes.

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -23,6 +23,23 @@ namespace WSLCE2ETests {
...
+    wil::unique_handle GetNonElevatedPrimaryToken()
+    {
+        // This method is necessary because GetNonElevatedToken(TokenPrimary) does
+        // not actually give a de-elevated token when called from an elevated process.
+        // By getting impersonation token first this de-elevates the token, and then
+        // converts it to a primary token.
+        auto impersonationToken = GetNonElevatedToken(TokenImpersonation);
+        wil::unique_handle primaryToken;
+        THROW_IF_WIN32_BOOL_FALSE(
+            DuplicateTokenEx(impersonationToken.get(), TOKEN_ALL_ACCESS, nullptr, SecurityImpersonation, TokenPrimary, &primaryToken));
```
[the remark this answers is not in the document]
dkbennett: We aren't testing that here, and the reason GetNonElevatedPrimaryToken is not in Common is so it doesn't break other things unexpectedly. If we create a test later where this is a problem we can address it then. Dont want to fix something that isn't broken.

test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
```diff
@@ -49,6 +49,70 @@ class WSLCE2EGlobalTests
...
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+        // Verify session list shows the non-admin session name
+        result = RunWslc(L"session list", ElevationType::NonElevated);
+        result.Verify({.Stderr = L"", .ExitCode = S_OK});
+
+        VERIFY_IS_TRUE(result.Stdout.has_value());
+
+        // The space after cli is important to differentiate from the admin session which is wslc-cli-admin.
+        VERIFY_IS_TRUE(result.Stdout->find(L"wslc-cli ") != std::wstring::npos);
```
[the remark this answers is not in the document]
dkbennett: Admin may be present because we aren't removing sessions. It works at present, if it stops working I'll fix it.

PR #40048, opened 2026-03-31
Set Distro Env HOSTTYPE to aarch64 for ARM64 build

wangxin12:
Set Distro Env HOSTTYPE to aarch64 for ARM64 build

## PR Checklist

- [x] **Closes:** Link to issue #40040 


benhillis (changes requested):
Instead of adding more #defines, can you do something similar to what we do for WSL_PACKAGE_VERSION?

benhillis (approved):

PR #40050, opened 2026-03-31
socketshared: add maximum message size to avoid very large allocations

benhillis:
This change adds a maximum single message size of 4MiB to the socket helpers. This avoids a potentially huge (4GB) allocation from a malicious or buggy sender.

### DIFF
--- a/src/shared/inc/socketshared.h
+++ b/src/shared/inc/socketshared.h
@@ -64,6 +64,15 @@ try
 #endif
     }
 
+    if (MessageSize > 4 * 1024 * 1024) // 4 MiB
+    {
+#if defined(_MSC_VER)
+        THROW_HR_MSG(E_UNEXPECTED, "Message size too large: %llu", MessageSize);
+#elif defined(__GNUC__)
+        THROW_UNEXCEPTED();
+#endif
+    }
+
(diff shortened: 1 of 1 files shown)


src/shared/inc/socketshared.h
```diff
@@ -64,6 +64,15 @@ try
 #endif
     }
 
+    if (MessageSize > 4 * 1024 * 1024) // 4 MiB
```
OneBlue, line 67: Went and double checked, the biggest message I could think of is LX_MINI_INIT_IMPORT_RESULT, and we cap the icon size to 1MB, so I think 4 should be OK

src/shared/inc/socketshared.h
```diff
@@ -64,6 +64,15 @@ try
...
+    {
+#if defined(_MSC_VER)
+        THROW_HR_MSG(E_UNEXPECTED, "Message size too large: %llu", MessageSize);
+#elif defined(__GNUC__)
+        THROW_UNEXCEPTED();
```
OneBlue, line 72: nit: might be a good idea to LOG_ERROR() here, to make root causing this easier if it ever hits on the linux side

OneBlue (approved):

PR #40054, opened 2026-03-31
Add support for build logs

johnstep:


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -29,6 +30,110 @@ constexpr auto c_containerdStorage = "/var/lib/docker";
...
+    THROW_IF_WIN32_BOOL_FALSE(CryptStringToBinaryA(
+        encoded.c_str(), static_cast<DWORD>(encoded.size()), CRYPT_STRING_BASE64, reinterpret_cast<BYTE*>(result.data()), &size, nullptr, nullptr));
+
+    result.resize(size);
+    return result;
+}
+
+// Resolve \r overwrites: for each \n-delimited line, keep only the content after the last \r.
+// This collapses terminal progress updates (e.g. "50%\r75%\r100%") to their final state.
+std::string ResolveCarriageReturns(const std::string& input)
```
OneBlue, line 49: nit: I would recommend moving to common so we can add unit test for this method, TailLines() and IndentLines()

OneBlue (approved):
LGTM. Feel free to do the test suggestion in a followup change

src/windows/wslcsession/WSLCSession.cpp
johnstep: That sounds like a good idea, will defer to a follow-up change. Thanks!

PR #40062, opened 2026-04-01
Keep track of handle types in the service

OneBlue:

I've seen a couple test failures where we fail to correctly determine the type of an output handle. For now let's just keep track of handle types in the service, this should avoid the issue entirely

## PR Checklist

### DIFF
--- a/src/windows/common/wslutil.cpp
+++ b/src/windows/common/wslutil.cpp
@@ -1348,11 +1348,11 @@ WSLCHandle wsl::windows::common::wslutil::ToCOMOutputHandle(HANDLE Handle, DWORD
     switch (Type)
     {
     case WSLCHandleTypeFile:
-        return WSLCHandle{.Type = WSLCHandleTypeFile, .Handle = {.File =  duplicatedHandle.release()}};
+        return WSLCHandle{.Type = WSLCHandleTypeFile, .Handle = {.File = duplicatedHandle.release()}};
     case WSLCHandleTypePipe:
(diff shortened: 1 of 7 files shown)


src/windows/common/wslutil.cpp
```diff
@@ -1339,6 +1339,26 @@ WSLCHandle wsl::windows::common::wslutil::ToCOMOutputHandle(HANDLE Handle, DWORD
     return comHandle;
 }
 
+WSLCHandle wsl::windows::common::wslutil::ToCOMOutputHandle(HANDLE Handle, DWORD Access, WSLCHandleType Type)
+{
+    wil::unique_handle duplicatedHandle{DuplicateHandle(Handle, Access)};
+
+    // N.B. COM closes the handle when returning an out parameter.
+    // The return value of this method should always be passed to a COM out parameter.
+    HANDLE raw = duplicatedHandle.release();
```
benhillis: minor: won't we leak the handle if we throw below?

OneBlue: Hmm that's true. Fixed by only releasing on return

src/windows/common/wslutil.cpp
JohnMcPMS, line 1331: It feels like this one shouldn't exist to force explicitly providing the type, but I don't have enough context to know that for sure.

OneBlue: In the case of the CLI, we unfortunately don't always have this information (for instance standard handles can be files, pipes, sockets, etc).

benhillis (approved):

PR #40074, opened 2026-04-02
Suppress MSI-initiated reboots during Store updates

benhillis:
When the WSL MSIX package is updated via the Microsoft Store, the WslInstaller service automatically upgrades the MSI package by calling MsiInstallProduct. This call was made with INSTALLUILEVEL_NONE (silent install) but without setting the REBOOT=ReallySuppress property.

Per Windows Installer documentation, when a silent install encounters files in use and REBOOT is not suppressed, the system reboots automatically without any user prompt. This could cause unexpected machine restarts after a Store update when WSL binaries (e.g. wslservice.exe) were in use during the upgrade.

Every deployment script in the repo already passes /norestart to msiexec (deploy-to-host.ps1, deploy-to-vm.ps1, install-latest-wsl.ps1, test-setup.ps1), but the programmatic MsiInstallProduct path used by the WslInstaller service lacked the equivalent property.

This change:
- Always appends REBOOT=ReallySuppress to MsiInstallProduct arguments, preventing Windows Installer from ever initiating a system restart.
- Treats ERROR_SUCCESS_REBOOT_REQUIRED (3010) as success. This return code means the install succeeded but some files will be replaced on the next natural reboot. Previously it was treated as a failure (exitCode != 0), which would falsely report update failed to users.
 
Fixes #14536

### DIFF
--- a/src/windows/common/install.cpp
+++ b/src/windows/common/install.cpp
@@ -104,7 +104,11 @@ int UpdatePackageImpl(bool preRelease, bool repair)
 
         const auto exitCode = UpgradeViaMsi(downloadPath.c_str(), L"", logFile.c_str(), &MsiMessageCallback);
 
-        if (exitCode != 0)
+        if (exitCode == ERROR_SUCCESS_REBOOT_REQUIRED)
+        {
+            PrintSystemError(ERROR_SUCCESS_REBOOT_REQUIRED);
+        }
+        else if (exitCode != 0)
         {
             clearLogs.release();
             THROW_HR_WITH_USER_ERROR(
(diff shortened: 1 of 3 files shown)


src/windows/common/install.cpp
```diff
@@ -358,18 +358,31 @@ int wsl::windows::common::install::UpdatePackage(bool PreRelease, bool Repair)
...
+    auto result = MsiInstallProduct(PackageLocation, args.c_str());
+    WSL_LOG("MsiInstallResult", TraceLoggingValue(result, "result"), TraceLoggingValue(args.c_str(), "ExtraArgs"));
 
     WriteInstallLog(std::format("MSI upgrade result: {}", result));
 
+    // ERROR_SUCCESS_REBOOT_REQUIRED (3010) means the installation succeeded but some
+    // files will be replaced on the next reboot. This is not a failure.
+    if (result == ERROR_SUCCESS_REBOOT_REQUIRED)
+    {
+        WriteInstallLog("MSI upgrade completed successfully but a reboot is needed to replace files that were in use");
```
OneBlue: nit: We might want to propagate the error here and handle this at the wslclient level so we can display an stderr message if a reboot is required

OneBlue (approved):
Great catch, thank you for doing this

src/windows/common/install.cpp
```diff
@@ -358,18 +358,31 @@ int wsl::windows::common::install::UpdatePackage(bool PreRelease, bool Repair)
...
-        TraceLoggingValue(result, "result"),
-        TraceLoggingValue(ExtraArgs != nullptr ? ExtraArgs : L"", "ExtraArgs"));
+    auto result = MsiInstallProduct(PackageLocation, args.c_str());
+    WSL_LOG("MsiInstallResult", TraceLoggingValue(result, "result"), TraceLoggingValue(args.c_str(), "ExtraArgs"));
 
     WriteInstallLog(std::format("MSI upgrade result: {}", result));
 
+    // ERROR_SUCCESS_REBOOT_REQUIRED (3010) means the installation succeeded but some
+    // files will be replaced on the next reboot. This is not a failure.
+    if (result == ERROR_SUCCESS_REBOOT_REQUIRED)
```
benhillis: Still thinking about this part.

JohnMcPMS (approved):

kmj251:
Does this fix #14536?

benhillis:
> Does this fix #14536?

yes it should...

OneBlue (approved):
LGTM. We could potentially had some logic to keep track of the "reboot required" state, but since we're trying to get rid of the MSIX installer path, I think this is good for now

PR #40080, opened 2026-04-02
devicehost: stop re-signing and rely on MSIRMSHUTDOWN for file-in-use handling

benhillis:
Cherry-pick of #40075 onto feature/wsl-for-apps.

This change:
- Sources wsldevicehost.dll directly from the DeviceHost package output (already signed), removing redundant re-signing
- Removes wsldevicehost.dll from the build-stage publish pattern
- Removes the CMake staging link for wsldevicehost.dll
- Bumps Microsoft.WSL.DeviceHost to 1.2.10-0
- Relies on the existing MSIRMSHUTDOWN=1 MSI property to handle file-in-use scenarios during install, instead of a custom action


OneBlue (approved):

OneBlue:
Overriding to get a release build sooner. The wslc test passed a full testpass was done on master

PR #40081, opened 2026-04-02
Allow small timestamp difference in ContainerRecoveryFromStorage test

beena352:
The ContainerRecoveryFromStorage test was failing because the live and recovery paths use different Docker timestamp sources (event.time vs InspectContainer().State.FinishedAt), which can differ by about a second. Updated the assertion to allow StateChangedAt to be within 60 seconds instead of requiring an exact match.

## PR Checklist

- [ x] **Tests:** Added/updated if needed and all pass

### DIFF
--- a/test/windows/WSLCTests.cpp
+++ b/test/windows/WSLCTests.cpp
@@ -5182,7 +5182,8 @@ class WSLCTests
             wil::unique_cotaskmem_array_ptr<WSLCContainerPortMapping> ports;
             VERIFY_SUCCEEDED(session->ListContainers(&containers, containers.size_address<ULONG>(), &ports, ports.size_address<ULONG>()));
             VERIFY_ARE_EQUAL(containers.size(), 1);
-            VERIFY_ARE_EQUAL(containers[0].StateChangedAt, originalStateChangedAt);
+            VERIFY_IS_TRUE(containers[0].StateChangedAt <= originalStateChangedAt + 60);
+            VERIFY_IS_TRUE(containers[0].StateChangedAt + 60 >= originalStateChangedAt);
             VERIFY_ARE_EQUAL(containers[0].CreatedAt, originalCreatedAt);
 
             VERIFY_SUCCEEDED(container.Get().Delete(WSLCDeleteFlagsNone));
--- a/test/windows/WSLCTests.cpp
+++ b/test/windows/WSLCTests.cpp
@@ -5182,8 +5182,12 @@ class WSLCTests
             wil::unique_cotaskmem_array_ptr<WSLCContainerPortMapping> ports;
             VERIFY_SUCCEEDED(session->ListContainers(&containers, containers.size_address<ULONG>(), &ports, ports.size_address<ULONG>()));
             VERIFY_ARE_EQUAL(containers.size(), 1);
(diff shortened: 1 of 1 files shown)


benhillis:
I think this is just working around the issue, could we replace the eventTime In OnEvent(Stop) from the Docker event stream with the FinishedAt timestamp from InspectContainer()?

beena352:
> I think this is just working around the issue, could we replace the eventTime In OnEvent(Stop) from the Docker event stream with the FinishedAt timestamp from InspectContainer()?

Good catch, that was the original approach. Based on Pierre’s feedback, though, calling Inspect() here could be risky since the session might be terminating and the call could fail. He suggested updating the test to accept timestamps within ~60 seconds instead.

OneBlue:
> Good catch, that was the original approach. Based on Pierre’s feedback, though, calling Inspect() here could be risky since the session might be terminating and the call could fail. He suggested updating the test to accept timestamps within ~60 seconds instead.


Agreed, I think a small (~ 1 second difference) is acceptable if that allows us to avoid additional Inspect() calls. We can always revisit if this turns out to be an issue later

benhillis (approved):

benhillis (changes requested):
is this + or - one minute? I think we just want one second, right? maybe use the absolute value of the values substracted is <= 1?

OneBlue:
> is this + or - one minute? I think we just want one second, right? maybe use the absolute value of the values substracted is <= 1?

The thinking was to put one minute to make sure that we don't have any test instability based on timing

benhillis:
> > is this + or - one minute? I think we just want one second, right? maybe use the absolute value of the values substracted is <= 1?
> 
> The thinking was to put one minute to make sure that we don't have any test instability based on timing

fair enough, let's do abs though.

test/windows/WSLCTests.cpp
```diff
@@ -5238,7 +5182,12 @@ class WSLCTests
             wil::unique_cotaskmem_array_ptr<WSLCContainerPortMapping> ports;
             VERIFY_SUCCEEDED(session->ListContainers(&containers, containers.size_address<ULONG>(), &ports, ports.size_address<ULONG>()));
             VERIFY_ARE_EQUAL(containers.size(), 1);
-            VERIFY_ARE_EQUAL(containers[0].StateChangedAt, originalStateChangedAt);
+
+            // StateChangedAt may differ by ~1s between live (event time) and recovery (FinishedAt).
+            auto stateChangedAtDiff = (containers[0].StateChangedAt > originalStateChangedAt)
```
OneBlue, line 5187: nit: could be simplified to `std::abs(containers[0].StateChangedAt, originalStateChangedAt)`

OneBlue (approved):

test/windows/WSLCTests.cpp
beena352: Good point, though since these are ULONGLONG, std::abs on the subtraction could wrap if recovery is smaller. Happy to change it if there's a preferred helper I'm missing though!

OneBlue: Ah that's true, thank you !

OneBlue (approved):

benhillis (approved):

benhillis:
Merging since random ci runs keep hitting this.

PR #40083, opened 2026-04-02
Add week/month/year support to FormatRelativeTime

benhillis:
Previously FormatRelativeTime only handled seconds, minutes, hours, and days. Containers older than a week showed values like '720 days ago' instead of '2 years ago'.

Now matches Docker-style output with weeks, months, and years. Also refactors repeated pluralization logic into a lambda.


JohnMcPMS (approved):
I know this is probably beyond the scope of what you were doing with this PR, but it would probably be nice to have the last two units displayed rather than just one as the timescales get longer.  There is a lot of difference between `1 year 1 month` and `1 year 11 months` that will both show as `1 year` with this.

benhillis:
> I know this is probably beyond the scope of what you were doing with this PR, but it would probably be nice to have the last two units displayed rather than just one as the timescales get longer. There is a lot of difference between `1 year 1 month` and `1 year 11 months` that will both show as `1 year` with this.

yep, I think this can be improved further, but this is better than it was.

PR #40085, opened 2026-04-02
Validate container path is absolute in VolumeMount::Parse

benhillis:
Add validation that non-empty container paths must start with '/' since they are Linux paths inside the container. This catches cases where Windows drive letter colons (e.g. C:\path) get misinterpreted as the host:container separator, producing invalid container paths like '\hostPath' instead of '/containerPath'.

Previously, 'C:\hostPath:ro' would silently parse as host='C', container='\hostPath', mode=ro - now it throws a clear error.

Updated tests to reflect the new validation and moved previously 'valid' but semantically incorrect cases to a new VolumeMount_Parse_InvalidContainerPath test.

### DIFF
--- a/src/windows/wslc/services/ContainerModel.cpp
+++ b/src/windows/wslc/services/ContainerModel.cpp
@@ -195,6 +195,13 @@ VolumeMount VolumeMount::Parse(const std::wstring& value)
             std::format(L"Invalid volume specifications: '{}'. Host path cannot be empty. Expected format: <host path>:<container path>[:mode]", value));
     }
 
+    if (!vm.m_containerPath.empty() && vm.m_containerPath[0] != '/')
+    {
+        THROW_HR_WITH_USER_ERROR(
+            E_INVALIDARG,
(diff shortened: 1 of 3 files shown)


src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -195,6 +195,20 @@ VolumeMount VolumeMount::Parse(const std::wstring& value)
             std::format(L"Invalid volume specifications: '{}'. Host path cannot be empty. Expected format: <host path>:<container path>[:mode]", value));
     }
 
+    if (!std::filesystem::path(vm.m_hostPath).is_absolute())
```
AmelBawa-msft: Although this isn’t implemented in the CLI yet, once we add support for named volumes, this condition may need to be removed so identifiers can be passed as the host path.

```
wslc volume create test_vol
wslc run -v test_vol:/usr/share/nginx/html nginx
```

JohnMcPMS (approved):

PR #40086, opened 2026-04-02
Fix callback parameter shadow in CreateInternal

benhillis:
The CreateInternal function takes an IProgressCallback* parameter but never uses it. When auto-pulling a missing image, a local PullImageCallback was created that shadowed the parameter name.

Now uses the caller-provided callback if non-null, falling back to the default PullImageCallback. This also removes the unnecessary inner scope braces.

### DIFF
--- a/src/windows/wslc/services/ContainerService.cpp
+++ b/src/windows/wslc/services/ContainerService.cpp
@@ -77,8 +77,7 @@ static void SetContainerArguments(WSLCProcessOptions& options, std::vector<const
     options.CommandLine = {.Values = argsStorage.data(), .Count = static_cast<ULONG>(argsStorage.size())};
 }
 
-static wsl::windows::common::RunningWSLCContainer CreateInternal(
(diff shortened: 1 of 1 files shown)


src/windows/wslc/services/ContainerService.cpp
```diff
@@ -140,9 +140,16 @@ static wsl::windows::common::RunningWSLCContainer CreateInternal(
     if (result == WSLC_E_IMAGE_NOT_FOUND)
     {
         // Attempt to pull the image if not found
-        PullImageCallback defaultCallback;
-        IProgressCallback* pullCallback = callback ? callback : &defaultCallback;
         PrintMessage(L"Image '%hs' not found, pulling", stderr, image.c_str());
+
+        IProgressCallback* pullCallback = callback;
```
OneBlue: nit: it's looking like `callback` is always null so I'd say let's remove this argument. This is an artifact of a previous design

OneBlue (approved):

PR #40088, opened 2026-04-02
Implement wslc session enter

OneBlue:

This change creates a new wslc command: `wslc sesssion enter <storage-path> [--name <session-name>].

This command is design to be a long term replacement to `wsl --wslc` to be used for ephemeral debugging sessions (for instance when debugging the tests) 

## PR Checklist


benhillis (approved):

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -54,6 +54,11 @@ void WSLCSessionManagerImpl::CreateSession(const WSLCSessionSettings* Settings,
     // Ensure that the session display name is non-null and not too long.
     THROW_HR_IF(E_INVALIDARG, Settings->DisplayName == nullptr);
     THROW_HR_IF(E_INVALIDARG, wcslen(Settings->DisplayName) >= std::size(WSLCSessionInformation{}.DisplayName));
+    THROW_HR_IF_MSG(
+        E_INVALIDARG,
+        WI_IsAnyFlagSet(Settings->StorageFlags, ~WSLCSessionStorageFlagsValid),
+        "Invalid storage flags flags: %i",
```
beena352: nit: duplicate "flags" in error message

test/windows/wslc/CommandLineTestCases.h
```diff
@@ -35,6 +35,11 @@ COMMAND_LINE_TEST_CASE(L"session shell session1", L"shell", true)
 COMMAND_LINE_TEST_CASE(L"session shell", L"shell", true)
 COMMAND_LINE_TEST_CASE(L"session terminate session1", L"terminate", true)
 COMMAND_LINE_TEST_CASE(L"session terminate", L"terminate", true)
+COMMAND_LINE_TEST_CASE(L"session enter C:\\storage", L"enter", true)
+COMMAND_LINE_TEST_CASE(L"session enter C:\\storage --name my-session", L"enter", true)
```
dkbennett, line 39: "session enter --name foo c:\\storage" should also be valid, but not necessary, we have lots of coverage of this pattern for commands.

test/windows/wslc/e2e/WSLCE2ESessionEnterTests.cpp
```diff
@@ -0,0 +1,97 @@
...
+        listResult.Verify({.Stderr = L"", .ExitCode = S_OK});
+        VERIFY_IS_TRUE(listResult.Stdout.has_value());
+        VERIFY_IS_FALSE(listResult.Stdout->find(sessionName) != std::wstring::npos);
+    }
+
+    TEST_METHOD(WSLCE2E_SessionEnter_WithoutName_GeneratesGuid)
+    {
+        WSL2_TEST_ONLY();
+
+        auto session = RunWslcInteractive(std::format(L"session enter \"{}\"", SessionOptions::GetStoragePath()));
```
dkbennett, line 80: Is this pattern of using the default storage location for session enter a common way of doing it, or just for convenient testing?  If it's common then perhaps we can make the storage path not required and use GetStoragePath internally if no path is provided to use the default session's storage.

dkbennett (approved):
Looks like an exemplary CLI command implementation! Few comments that are not blocking.

test/windows/wslc/e2e/WSLCE2ESessionEnterTests.cpp
OneBlue: I just used that to make testing convenient (and faster), since we're in the CLI tests we need that the CLI session exists. 

Right now I don't think we'd want `wslc session enter` to use the CLI session storage by default, because that would create a non persistent session that will use the CLI's storage but a different name, which would lead to confusing errors if the users tried to use the CLI session in the "regular" way in parallel

test/windows/wslc/CommandLineTestCases.h
OneBlue: I'm going to push another iteration to fix something else, will add

test/windows/wslc/e2e/WSLCE2ESessionEnterTests.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    WSLC_TEST_CLASS(WSLCE2ESessionEnterTests)
+
+    TEST_CLASS_SETUP(TestClassSetup)
+    {
+        // Ensure that the wslc cli session storage is created.
+        RunWslc(L"image ls").Verify({.ExitCode = 0});
+
+        // Terminate the wslc session since we use its storage path in this test class.
+        RunWslc(L"session terminate");
```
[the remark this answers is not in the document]
OneBlue: That's because that session isn't necessarily running

AmelBawa-msft (approved):

PR #40089, opened 2026-04-02
Localize all user-facing wslc.exe CLI strings

benhillis:
Localize all user-facing strings in the wslc.exe CLI tool.

Previously, most command descriptions and argument help strings were hardcoded English L"..." literals. Only 5 of 35 argument descriptions and none of the ~50 command descriptions used the localization system.

### Changes

**Resources.resw** (90 new entries):
- 55 command description entries (Short + Long for all 26 command classes)
- 31 argument description entries
- 4 settings-related entries
- All entries include {Locked=...} comments for CLI flags, filenames, and technical terms that shouldn't be translated

**ArgumentDefinitions.h**: All 30 previously-hardcoded argument descriptions now use Localization::WSLCCLI_*ArgDescription() calls

**26 command .cpp files**: All ShortDescription() and LongDescription() methods now use Localization::WSLCCLI_* calls

**E2E tests**: Updated baselines to match new description text

Builds and compiles clean (wslc.exe + wsltests.dll). Formatted with FormatSource.ps1.

### DIFF
--- a/localization/strings/en-US/Resources.resw
+++ b/localization/strings/en-US/Resources.resw
@@ -2168,7 +2168,7 @@ For privacy information about this product please visit https://aka.ms/privacy.<
   </data>
   <data name="WSLCCLI_ContainerListLongDesc" xml:space="preserve">
     <value>Lists containers. By default, only running containers are shown; use --all to include all containers.</value>
-    <comment>{Locked="--all"}Command line arguments should not be translated</comment>
+    <comment>{Locked="--all "}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
   <data name="WSLCCLI_ContainerLogsDesc" xml:space="preserve">
     <value>View container logs.</value>
@@ -2187,7 +2187,7 @@ For privacy information about this product please visit https://aka.ms/privacy.<
   </data>
   <data name="WSLCCLI_ContainerRunLongDesc" xml:space="preserve">
     <value>Runs a container. By default, the container is started in the background; use --detach to run in the foreground.</value>
-    <comment>{Locked="--detach"}Command line arguments should not be translated</comment>
+    <comment>{Locked="--detach "}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
   <data name="WSLCCLI_ContainerStartDesc" xml:space="preserve">
     <value>Start a container.</value>
@@ -2329,7 +2329,7 @@ On first run, creates the file with all settings commented out at their defaults
     <value>Follow log output</value>
   </data>
   <data name="WSLCCLI_FormatArgDescription" xml:space="preserve">
-    <value>Output formatting (json or table) (Default:table)</value>
+    <value>Output formatting (json or table) (Default: table)</value>
     <comment>{Locked="json"}{Locked="table"}Command line arguments should not be translated</comment>
   </data>
   <data name="WSLCCLI_ForwardArgsDescription" xml:space="preserve">
@@ -2373,7 +2373,7 @@ On first run, creates the file with all settings commented out at their defaults
   </data>
   <data name="WSLCCLI_SignalArgDescription" xml:space="preserve">
     <value>Signal to send (default: {})</value>
-    <comment>{FixedPlaceholder="{}"}Command line arguments should not be translated</comment>
+    <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
   <data name="WSLCCLI_TagArgDescription" xml:space="preserve">
     <value>Tag for the built image</value>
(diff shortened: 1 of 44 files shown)


localization/strings/en-US/Resources.resw
```diff
@@ -2118,4 +2118,360 @@ For privacy information about this product please visit https://aka.ms/privacy.<
...
+    <value>Invalid format type specified. Supported format types are: json, table</value>
+    <comment>{Locked="json"}{Locked="table"}Command line arguments should not be translated</comment>
+  </data>
+  <data name="WSLCCLI_InvalidSignalError" xml:space="preserve">
+    <value>Invalid {} value: {} is not a recognized signal name or number (Example: SIGKILL, kill, or 9).</value>
+    <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
+  </data>
+  <data name="WSLCCLI_SignalOutOfRangeError" xml:space="preserve">
+    <value>Invalid {} value: {} is out of valid range ({}-{}).</value>
+    <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
```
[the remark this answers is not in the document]
benhillis: nice...

localization/strings/en-US/Resources.resw
```diff
@@ -2118,4 +2118,360 @@ For privacy information about this product please visit https://aka.ms/privacy.<
...
+  <data name="WSLCCLI_SessionListVerboseArgDescription" xml:space="preserve">
+    <value>Show detailed information about the listed sessions.</value>
+  </data>
+  <data name="WSLCCLI_InvalidFormatError" xml:space="preserve">
+    <value>Invalid format type specified. Supported format types are: json, table</value>
+    <comment>{Locked="json"}{Locked="table"}Command line arguments should not be translated</comment>
+  </data>
+  <data name="WSLCCLI_InvalidSignalError" xml:space="preserve">
+    <value>Invalid {} value: {} is not a recognized signal name or number (Example: SIGKILL, kill, or 9).</value>
+    <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
```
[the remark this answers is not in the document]
benhillis: wrong

localization/strings/en-US/Resources.resw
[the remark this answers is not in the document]
benhillis: this is wrong.

AmelBawa-msft (approved):

test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
```diff
@@ -411,31 +413,43 @@ class WSLCE2EGlobalTests
...
-                 << L"  start      Start a container.\r\n"
-                 << L"  stop       Stop containers.\r\n"
-                 << L"\r\n"
-                 << L"For more details on a specific command, pass it the help argument. [-h]\r\n\r\n";
+        commands << Localization::WSLCCLI_AvailableCommands() << L"\r\n";
+        for (const auto& [name, desc] : entries)
+        {
+            commands << L"  " << name << std::wstring(maxLen - name.size() + 2, L' ') << desc << L"\r\n";
+        }
+        commands << L"\r\n" << Localization::WSLCCLI_HelpForDetails() << L" [" << WSLC_CLI_HELP_ARG_STRING << L"]\r\n\r\n";
```
dkbennett, line 452: We probably can get rid of this help output validation, since this test code is just re-implementing the help logic in the Command help generator. I would think this test may be difficult to get right because it would have to match the help generation code of the command exactly. IMO this test should be removed and we can instead have each command verify that expected subcommands are available instead of matching the help text exactly. If we want to test the help generation I think we can better unit the command help generation specifically so it is outputting what is expected. That, combined verifying each command supports the expected options and command line validation should give very high confidence all of this is correct.

test/windows/wslc/e2e/WSLCE2EImageInspectTests.cpp
```diff
@@ -92,7 +93,7 @@ class WSLCE2EImageInspectTests
 
     std::wstring GetDescription() const
     {
-        return L"Inspect images.\r\n\r\n";
+        return Localization::WSLCCLI_ImageInspectLongDesc() + L"\r\n\r\n";
```
dkbennett, line 96: The two returns at the end of this seems a bit off, and I'm not sure what the value of verifying that the description out put matches the same string it is defined using.

dkbennett (approved):
This looks good to me for a starting point.

I think we want to probably remove all the help output validation tests and replace it with:
1) help output generation unit tests
2) command validation that all expected commands are defined, and all expected arguments are defined

However, the second item is sort of already validated with most of the commandline tests where we validate that the command line against what command it executes and whether it is valid/not-valid. 

So I think we can replace the help validation with help output unit tests and expand on the command lines a bit and get better and more robust testing than the help output comparisons, which are fragile when modifying any command.

test/windows/wslc/e2e/WSLCE2EImageInspectTests.cpp
benhillis: agreed, let's look at this in a follow-up

test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
benhillis: ok, let's do that as a follow-up

PR #40092, opened 2026-04-03
Fix random "ERROR_FILE_NOT_FOUND" when unmounting with absolute path

chemwolf6922:

Fix this random issue where sometimes unmount with the absolute path results in ERROR_FILE_NOT_FOUND
```
PS D:\tmp> wsl --mount "D:\tmp\test.vhdx" --vhd --bare
wsl: The .wslconfig setting 'wsl2.kernelCommandLine' is disabled by the computer policy.
The operation completed successfully.
PS D:\tmp> wsl --unmount  "D:\tmp\test.vhdx"
The system cannot find the file specified.
Error code: Wsl/Service/DetachDisk/ERROR_FILE_NOT_FOUND
PS D:\tmp> wsl --unmount  "\\?\D:\tmp\test.vhdx"
The operation completed successfully.
```

## PR Checklist

## Validation Steps Performed

This PR adds the MountTests::MountTests::AbsolutePathVhdUnmountAfterVMTimeout test case. This test case waits for the VM to timeout before unmounting the vhd. Which will trigger the original issue w/o this patch.

### DIFF
--- a/src/windows/common/WslClient.cpp
+++ b/src/windows/common/WslClient.cpp
@@ -1218,9 +1218,8 @@ int Unmount(_In_ const std::wstring& arg)
     wsl::windows::common::SvcComm service;
     const HRESULT result = wil::ResultFromException([&] { value = service.DetachDisk(disk); });
 
-    // support relative paths in unmount
-    // check is the result is the error code for "file not found" and the path is relative
-    if (result == HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND) && PathIsRelative(disk))
+    // Retry with the normalized path to handle relative paths and \\?\ prefix mismatches.
+    if (result == HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND))
     {
         // retry dismounting with the absolute path
         const auto absoluteDisk = wsl::windows::common::filesystem::GetFullPath(filesystem::UnquotePath(disk).c_str());
--- a/test/windows/MountTests.cpp
+++ b/test/windows/MountTests.cpp
@@ -450,6 +450,29 @@ class MountTests
         VERIFY_ARE_EQUAL(LxsstuLaunchWsl(L"--unmount " + absolutePath.wstring()), (DWORD)0);
     }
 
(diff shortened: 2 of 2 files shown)


src/windows/common/WslClient.cpp
```diff
@@ -1218,9 +1218,8 @@ int Unmount(_In_ const std::wstring& arg)
     wsl::windows::common::SvcComm service;
     const HRESULT result = wil::ResultFromException([&] { value = service.DetachDisk(disk); });
 
-    // support relative paths in unmount
-    // check is the result is the error code for "file not found" and the path is relative
-    if (result == HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND) && PathIsRelative(disk))
+    // Retry with the normalized path to handle relative paths and \\?\ prefix mismatches.
+    if (result == HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND))
```
benhillis, line 1222: Did you figure out why the result is not deterministic? That’s very odd.

benhillis:
Hey @chemwolf6922 👋 — Following up on this PR. CI is all green and it's a nice targeted fix. There's **1 unresolved review question** from @benhillis asking why the bug is non-deterministic — could you provide a response when you get a chance? That should help move this toward approval.

src/windows/common/WslClient.cpp
chemwolf6922: After more testing, I can repo it deterministically now. This always happens with mount -> vm timeout -> unmount. When the VM is shutdown, this code gets called instead of the filesystem path comparison. Which does not handle the `\\?\` prefix correctly:
https://github.com/microsoft/WSL/blob/7ee07db18d8d831ec7a2e402fb15ef90cd2c92ca/src/windows/service/exe/LxssUserSession.cpp#L854

chemwolf6922:
Hi, @benhillis , the problem was root caused and I added the test for it. Could you please take another look? Thanks.

benhillis (approved):
Ok this makes sense now, thanks for fixing!

PR #40097, opened 2026-04-03
Add response size limit to Docker HTTP client

benhillis:
SendRequestAndReadResponse accumulates the entire Docker API response body into a std::string with no size limit. While all current callers expect small JSON metadata (list, inspect, create), a pathological or malformed response could cause unbounded memory growth.

Adds a 64 MB limit. This is generous for JSON metadata (typically <1 MB) while catching runaway responses early. Large-payload operations (export, save, attach, logs, pull, import, load) already use streaming via SendRequest and are unaffected.

### DIFF
--- a/src/windows/wslcsession/DockerHTTPClient.cpp
+++ b/src/windows/wslcsession/DockerHTTPClient.cpp
@@ -482,9 +482,22 @@ std::pair<DockerHTTPClient::HTTPResponse, std::string> DockerHTTPClient::SendReq
     auto context = SendRequestImpl(Method, Url, Body, {});
 
     // Read the response header and body.
+    // Limit response size to prevent unbounded memory growth from pathological responses.
+    // All callers expect JSON metadata (list, inspect, create, etc.), not large binary payloads.
+    constexpr size_t MaxResponseSize = 64 * _1MB;
+
     std::optional<HTTPResponse> responseHeader;
     std::string responseBody;
-    auto OnResponse = [&responseBody](const gsl::span<char>& span) { responseBody.append(span.data(), span.size()); };
+    const auto& url = Url;
+    auto OnResponse = [&responseBody, &url](const gsl::span<char>& span) {
+        THROW_HR_IF_MSG(
+            HRESULT_FROM_WIN32(ERROR_FILE_TOO_LARGE),
+            span.size() > MaxResponseSize - responseBody.size(),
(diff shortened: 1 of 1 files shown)


benhillis:
sorry this should have been a draft

src/windows/wslcsession/DockerHTTPClient.cpp
```diff
@@ -482,9 +482,17 @@ std::pair<DockerHTTPClient::HTTPResponse, std::string> DockerHTTPClient::SendReq
...
+    // Limit response size to prevent unbounded memory growth from pathological responses.
+    // All callers expect JSON metadata (list, inspect, create, etc.), not large binary payloads.
+    constexpr size_t MaxResponseSize = 64 * 1024 * 1024; // 64 MB
+
     std::optional<HTTPResponse> responseHeader;
     std::string responseBody;
-    auto OnResponse = [&responseBody](const gsl::span<char>& span) { responseBody.append(span.data(), span.size()); };
+    auto OnResponse = [&responseBody](const gsl::span<char>& span) {
+        THROW_HR_IF_MSG(
+            E_OUTOFMEMORY, responseBody.size() + span.size() > MaxResponseSize, "Docker API response exceeds maximum size");
```
OneBlue: nit: if you push another iteration, I think we should log the URL in this error to make root causing easier if it ever hits

OneBlue (approved):

src/windows/wslcsession/DockerHTTPClient.cpp
```diff
@@ -482,9 +482,17 @@ std::pair<DockerHTTPClient::HTTPResponse, std::string> DockerHTTPClient::SendReq
     auto context = SendRequestImpl(Method, Url, Body, {});
 
     // Read the response header and body.
+    // Limit response size to prevent unbounded memory growth from pathological responses.
+    // All callers expect JSON metadata (list, inspect, create, etc.), not large binary payloads.
+    constexpr size_t MaxResponseSize = 64 * 1024 * 1024; // 64 MB
```
JohnMcPMS:
```suggestion
    constexpr size_t MaxResponseSize = 64 * _1MB;
```

JohnMcPMS (commented):
Even the AI has learned that you have that you need to do many more comments per line changed for smaller PRs.

OneBlue (approved):

PR #40099, opened 2026-04-03
Fixes to route mirroring

FetoiuCatalin:
Testing mirrored mode with Proton VPN exposed some issues with route mirroring:
1) Adding routes using netlink messages did not account for default onlink routes (destination everything, no next hop), leading to default onlink routes for IPv6 to fail to be added. Proton VPN uses such a route on the host, which failed to be mirrored and the IPv6 traffic was not going over the VPN as expected
2) When mirroring routes from windows, the metric used in Linux is determined by adding windows interface metric and route metric. It's possible those are both 0 (high priority), leading to a Linux metric with route 0, but 0 is treated as unspecified and defaulted to 1024 instead. Fix this to change the metric 0 to 1 (which is highest priority metric in Linux)

## Validation Steps Performed
Testing with ProtonVPN and confirming v6 traffic is correctly routed over the mirrored interface that corresponds to the VPN
Running all NetworkTests
Added new tests that verify adding default onlink routes works as expected for both IPv4 and IPv6


test/windows/NetworkTests.cpp
```diff
@@ -352,6 +352,31 @@ class NetworkTests
...
+        defaultRouteV4.DestinationPrefix = LX_INIT_DEFAULT_ROUTE_PREFIX;
+        defaultRouteV4.Family = AF_INET;
+        defaultRouteV4.Metric = 1;
+        SendDeviceSettingsRequest(L"eth0", defaultRouteV4, ModifyRequestType::Add, GuestEndpointResourceType::Route);
+
+        wsl::shared::hns::Route defaultRouteV6;
+        defaultRouteV6.NextHop = L"::";
+        defaultRouteV6.DestinationPrefix = LX_INIT_DEFAULT_ROUTE_V6_PREFIX;
+        defaultRouteV6.Family = AF_INET6;
+        defaultRouteV6.Metric = 1;
```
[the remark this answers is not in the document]
FetoiuCatalin: this is consistent with other tests, that do not explicitly removes the routes added

test/windows/NetworkTests.cpp
```diff
@@ -352,6 +352,31 @@ class NetworkTests
...
     }
 
+    TEST_METHOD(DefaultOnlinkRoutes)
+    {
+        wsl::shared::hns::Route defaultRouteV4;
+        defaultRouteV4.NextHop = L"0.0.0.0";
+        defaultRouteV4.DestinationPrefix = LX_INIT_DEFAULT_ROUTE_PREFIX;
+        defaultRouteV4.Family = AF_INET;
+        defaultRouteV4.Metric = 1;
+        SendDeviceSettingsRequest(L"eth0", defaultRouteV4, ModifyRequestType::Add, GuestEndpointResourceType::Route);
```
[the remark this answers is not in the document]
FetoiuCatalin: fixed in next push

src/linux/netlinkutil/RoutingTable.cpp
```diff
@@ -230,6 +234,27 @@ void RoutingTable::ModifyLoopbackRouteImpl(const Route& route, int operation, in
...
+
+    GNS_LOG_INFO(
+        "SendMessage Route (default), operation ({}), netLinkflags ({})",
+        RouteOperationToString(operation),
+        NetLinkFormatFlagsToString(flags).c_str());
+
+    SendMessage<Message>(route, operation, flags, [&](Message& message) {
+        GNS_LOG_INFO(
+            "InitializeAddressAttribute RTA_DST ([not set]) RTA_GATEWAY ([not set]), RTA_PRIORITY ({})",
+            route.metric);
```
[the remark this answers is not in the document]
FetoiuCatalin: fixed in latest push

src/linux/netlinkutil/RoutingTable.cpp
```diff
@@ -253,8 +278,8 @@ void RoutingTable::ModifyDefaultRouteImpl(const Route& route, int operation, int
 
     SendMessage<Message>(route, operation, flags, [&](Message& message) {
         GNS_LOG_INFO(
-            "InitializeAddressAttribute RTA_DST ([not set]) RTA_GATEWAY ({}), RTA_PRIORITY ({})",
-            route.to.has_value() ? route.to.value().Addr().c_str() : "[empty]",
+            "Netlink message configuration: RTA_DST ([not set]) RTA_GATEWAY ({}), RTA_PRIORITY ({})",
+            route.via.value().Addr().c_str(),
             route.metric);
         utils::InitializeAddressAttribute<TAddr>(message.via, route.via.value(), RTA_GATEWAY);
```
keith-horton, line 295: the difference was not setting the GATEWAY attribute on this route? is this why Linux failed it?

test/windows/NetworkTests.cpp
```diff
@@ -352,6 +352,33 @@ class NetworkTests
...
 
+    TEST_METHOD(DefaultOnlinkRoutes)
+    {
+        WSL2_TEST_ONLY();
+
+        wsl::shared::hns::Route defaultRouteV4;
+        defaultRouteV4.NextHop = L"0.0.0.0";
+        defaultRouteV4.DestinationPrefix = LX_INIT_DEFAULT_ROUTE_PREFIX;
+        defaultRouteV4.Family = AF_INET;
+        defaultRouteV4.Metric = 1;
```
keith-horton, line 354: since we have code now for metric == 0, can we test for that too?

FetoiuCatalin: I agree that would be good, but since the logic that changes the metric from 0 to 1 is in the wsl service, we would need to add such a route on the host and verify it's mirrored, but I think changing the host routes could break connectivity of the machine that runs the tests.

src/linux/netlinkutil/RoutingTable.cpp
FetoiuCatalin: the previous code was going through this function for default onlink routes - which did set the gateway
while the new code is going through ModifyDefaultLinkLocalRouteImpl, which does not set the gateway attribute

so Linux failed to add the IPv6 route because the gateway was set (to "::"). I did not look much into why Linux/netlink rejected such a route, but not setting the gateway fixed it

keith-horton: ok. Linux does have its own special rules for what it allows and doesn't allow. so this makes sense that it would fail an on-link "gateway".

test/windows/NetworkTests.cpp
keith-horton: ok

keith-horton (dismissed):
Looks good. Thanks!

test/windows/NetworkTests.cpp
```diff
@@ -352,6 +352,31 @@ class NetworkTests
...
+        SendDeviceSettingsRequest(L"eth0", defaultRouteV4, ModifyRequestType::Add, GuestEndpointResourceType::Route);
+
+        wsl::shared::hns::Route defaultRouteV6;
+        defaultRouteV6.NextHop = L"::";
+        defaultRouteV6.DestinationPrefix = LX_INIT_DEFAULT_ROUTE_V6_PREFIX;
+        defaultRouteV6.Family = AF_INET6;
+        defaultRouteV6.Metric = 1;
+        SendDeviceSettingsRequest(L"eth0", defaultRouteV6, ModifyRequestType::Add, GuestEndpointResourceType::Route);
```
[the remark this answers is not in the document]
keith-horton: @FetoiuCatalin , is this a real issueto be addressed?

test/windows/NetworkTests.cpp
```diff
@@ -352,6 +352,31 @@ class NetworkTests
...
+
+        const bool defaultV4RouteExists = LxsstuLaunchWsl(L"ip -4 route show | grep \"default dev eth0\" | grep \"metric 1\"") == (DWORD)0;
+        const bool defaultV6RouteExists = LxsstuLaunchWsl(L"ip -6 route show | grep \"default dev eth0\" | grep \"metric 1\"") == (DWORD)0;
```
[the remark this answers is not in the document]
keith-horton: @FetoiuCatalin , this seems accurate?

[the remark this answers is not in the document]
FetoiuCatalin: oh good catch, indeed this should be tightened to match just 1, not 10 or 100 or others

src/linux/netlinkutil/RoutingTable.cpp
```diff
@@ -222,14 +226,33 @@ void RoutingTable::ModifyLoopbackRouteImpl(const Route& route, int operation, in
...
+            "Netlink message configuration: RTA_DST ({}) RTA_GATEWAY ({}) RTA_PRIORITY ([not set])",
             route.to.value().Addr().c_str(),
             route.via.value().Addr().c_str());
         utils::InitializeAddressAttribute<TAddr>(message.to, route.to.value(), RTA_DST);
         utils::InitializeAddressAttribute<TAddr>(message.via, route.via.value(), RTA_GATEWAY);
     });
 }
 
+template <typename TAddr>
+void RoutingTable::ModifyDefaultLinkLocalRouteImpl(const Route& route, int operation, int flags)
```
keith-horton, line 238: did we verify that ModifyLinkLocalRouteImpl() wouldn't work correctly in this case?

test/windows/NetworkTests.cpp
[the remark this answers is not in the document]
FetoiuCatalin: not all testcases are removing routes at the end of the testcase, but it's best to remove them here. will fix

src/linux/netlinkutil/RoutingTable.cpp
FetoiuCatalin: good question - I did not but I expect it won't work because that function assume route.to (which is optional) has a value, but for default routes we don't set that field of the Route object
(see the code in GnsEngine::ProcessRouteChange, that determines if a route is default)

FetoiuCatalin: looks like this function and others are missing checks to confirm if those optional fields are set or not - adding those as well

e.g. ModifyLinkLocalRouteImpl should throw if Route.to does not have a value

FetoiuCatalin: added checks

test/windows/NetworkTests.cpp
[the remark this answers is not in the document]
FetoiuCatalin: fixed to use grep -w

test/windows/NetworkTests.cpp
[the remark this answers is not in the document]
FetoiuCatalin: updated to remove routes

src/linux/netlinkutil/RoutingTable.cpp
```diff
@@ -317,6 +322,10 @@ void RoutingTable::ModifyOfflinkRouteImpl(const Route& route, int operation, int
     {
         throw RuntimeErrorWithSourceLocation("Offlink route is missing its next hop");
     }
+    if (!route.to.has_value())
```
keith-horton, line 342: the code correctly handles when route.to.has_value() and when it doesn't. it should not fail it here.
why was this added? was something breaking? i.e., the code below did not handle it?

FetoiuCatalin: it handles only in the traces (GNS_LOG_INFO)
but it calls
utils::InitializeAddressAttribute<TAddr>(message.to, route.to.value(), RTA_DST);

which assumes route.to always has a value

FetoiuCatalin: I did not see something breaking but realized those checks were not consistent - e.g. we throw if route.via does not have a value, but we don't do the same when route.to does not have a value

keith-horton: If we are requiring route.via and route.to to have a value, please remove the check below that checks if they have a value or not.

keith-horton (changes requested):

src/linux/netlinkutil/RoutingTable.cpp
FetoiuCatalin: fixed

test/windows/NetworkTests.cpp
```diff
@@ -352,6 +352,43 @@ class NetworkTests
...
+        defaultRouteV6.Metric = 1;
+        SendDeviceSettingsRequest(L"eth0", defaultRouteV6, ModifyRequestType::Add, GuestEndpointResourceType::Route);
+
+        const bool defaultV4RouteExists =
+            LxsstuLaunchWsl(L"ip -4 route show | grep \"default dev eth0\" | grep -w \"metric 1\"") == (DWORD)0;
+        const bool defaultV6RouteExists =
+            LxsstuLaunchWsl(L"ip -6 route show | grep \"default dev eth0\" | grep -w \"metric 1\"") == (DWORD)0;
+
+        SendDeviceSettingsRequest(L"eth0", defaultRouteV4, ModifyRequestType::Remove, GuestEndpointResourceType::Route);
+        SendDeviceSettingsRequest(L"eth0", defaultRouteV6, ModifyRequestType::Remove, GuestEndpointResourceType::Route);
```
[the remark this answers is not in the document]
FetoiuCatalin: this is consistent with how other tests are removing routes (e.g. AddRouteWithMetrics)

src/linux/netlinkutil/RoutingTable.cpp
```diff
@@ -300,15 +300,15 @@ void RoutingTable::ModifyLinkLocalRouteImpl(const Route& route, int operation, i
 
     GNS_LOG_INFO(
         "SendMessage Route (to {}, via {}), operation ({}), netLinkflags ({})",
-        route.to.has_value() ? route.to.value().Addr().c_str() : "[empty]",
-        route.via.has_value() ? route.via.value().Addr().c_str() : "[empty]",
+        route.to.value().Addr().c_str(),
+        "[empty]",
```
keith-horton, line 321: route.via might have a value, right?

keith-horton: it's guaranteed in all possible paths that calling ModifyLinkLocalRouteImpl can **never** have a route.via?

src/linux/netlinkutil/RoutingTable.cpp
```diff
@@ -269,8 +269,8 @@ void RoutingTable::ModifyDefaultRouteImpl(const Route& route, int operation, int
 
     GNS_LOG_INFO(
         "SendMessage Route (to {}, via {}), operation ({}), netLinkflags ({})",
-        route.to.has_value() ? route.to.value().Addr().c_str() : "[empty]",
-        route.via.has_value() ? route.via.value().Addr().c_str() : "[empty]",
+        "[empty]",
```
keith-horton, line 285: is it guaranteed that route.to is always empty?
if so, we should fail the request if route.to.has_value().
this would be silently not doing what the caller expected if route.to.has_value()

src/linux/netlinkutil/RoutingTable.cpp
keith-horton: if so, we should fail the request if route.via.has_value().
this would be silently not doing what the caller expected if route.via.has_value()

src/linux/netlinkutil/RoutingTable.cpp
FetoiuCatalin: I added checks to ensure we throw if route.to is not empty while we expect it to be empty
we enter those ModifyDefault* functions when the defaultRoute field of the Route object is set and looking at the code, when we set that, the Route.to is empty

src/linux/netlinkutil/RoutingTable.cpp
FetoiuCatalin: I added checks to ensure we throw if route.via is not empty while we expect it to be empty
we enter the ModifyLinkLocalRouteImpl function when route.IsOnlink() returns true. I made a change to the Route constructor object to ensure that route.via is empty when route.IsOnlink() returns true

src/linux/netlinkutil/Route.cpp
```diff
@@ -6,6 +6,13 @@
 Route::Route(int family, const std::optional<Address>& via, int dev, bool defaultRoute, const std::optional<Address>& to, int metric) :
     family(family), via(via), dev(dev), defaultRoute(defaultRoute), to(to), metric(metric)
 {
+    // When the route is created, the via field may be "0.0.0.0" or "::", which is checked by
+    // IsOnlink(). Clear the via field in this case, which still preserves the return value of future
+    // IsOnlink() calls but makes the Route object cleaner.
+    if (IsOnlink())
```
keith-horton: the original code was wrong to allow via.has_value() when the address was 0.0.0.0 or ::0?
we are now changing how we track Route object properties. the current thinking is this should be correct in all cases?

keith-horton: i.e., IsOnlink()  allows for 2 things:
1. via.has_value() == false
OR
2. via.has_value() == true *and* family == AF_INET && via->Addr == 0.0.0.0 (same with ipv6)

so if we are defining Onlink requires via.has_value == false, then
1. we need to change IsOnlink == via.has_value == false
2. update the c'tor so make the family + via->addr() string check --> and if that matches, then via.reset()

src/linux/netlinkutil/Route.cpp
```diff
@@ -6,6 +6,13 @@
...
+    {
+        this->via.reset();
```
keith-horton: please remove 'this->'

FetoiuCatalin: this was needed because the via field of the struct has the same name as the via param of the constructor - I can change the constructor param to have a different name instead

keith-horton: please never have variable name collision, where the only difference is one has this->.
always give them unique and obvious names.

src/linux/netlinkutil/Route.cpp
FetoiuCatalin: good point, I updated IsOnlink to only check if !via.has_value()
and moved the family + via->addr() check in the constructor

src/linux/netlinkutil/Route.cpp
FetoiuCatalin: updated constructor parameter names

src/linux/netlinkutil/Route.cpp
```diff
@@ -4,8 +4,13 @@
...
 
 Route::Route(int family, const std::optional<Address>& via, int dev, bool defaultRoute, const std::optional<Address>& to, int metric) :
-    family(family), via(via), dev(dev), defaultRoute(defaultRoute), to(to), metric(metric)
+    family(family), via(nextHop), dev(dev), defaultRoute(defaultRoute), to(to), metric(metric)
 {
+    // For onlink routes, ensure the via field is empty
+    if (via.has_value() && ((family == AF_INET && via->Addr() == "0.0.0.0") || (family == AF_INET6 && via->Addr() == "::")))
+    {
+        via.reset();
+    }
```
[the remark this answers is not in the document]
FetoiuCatalin: fixed in latest iteration

keith-horton (dismissed):

keith-horton (dismissed):

test/windows/MountTests.cpp
```diff
@@ -438,10 +438,9 @@ class MountTests
         VERIFY_ARE_EQUAL(LxsstuLaunchWsl(L"--unmount " + absolutePath.wstring()), (DWORD)0);
     }
 
-    TEST_METHOD(AbsolutePathVhdUnmountAfterVMTimeout)
```
FetoiuCatalin: @OneBlue , @benhillis fyi this was causing the build to break for me so I fixed it here
seems to be related to https://github.com/microsoft/WSL/pull/40140

keith-horton (approved):

FetoiuCatalin:
@microsoft-github-policy-service agree company="Microsoft"

OneBlue (approved):

PR #40100, opened 2026-04-03
Add logic to reject invalid flags

OneBlue:

This change adds logic to ensure that we reject flags that we don't understand.

## PR Checklist

### DIFF
--- a/src/windows/service/inc/wslc.idl
+++ b/src/windows/service/inc/wslc.idl
@@ -427,6 +427,15 @@ interface IWSLCVirtualMachine : IUnknown
     HRESULT RemoveShare([in] REFGUID ShareId);
 }
 
+typedef enum _WSLCSessionStorageFlags
+{
+    WSLCSessionStorageFlagsNone = 0,
+    WSLCSessionStorageFlagsNoCreate = 1, // Open an existing storage path, but don't create a new one.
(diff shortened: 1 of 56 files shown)


src/windows/service/inc/wslc.idl
```diff
@@ -238,6 +238,7 @@ typedef enum _WSLCContainerStartFlags
 {
     WSLCContainerStartFlagsNone = 0,
     WSLCContainerStartFlagsAttach = 1, // Attach stdio handles on start.
+    WSLCContainerStartFlagsValid = WSLCContainerStartFlagsAttach
```
yao-msft: nit: do we want to expose them in idl?

WSLC*FlagsAll seems more appropriate

yao-msft (commented):
By the way, I saw some other flags are not checked, like the ones in ListImages. Are all flags to be updated in the future?

src/windows/service/inc/wslc.idl
benhillis: I'm not sure I like having the valid flag as part of the enum.

OneBlue: I originally went that way because it makes it easier to update the `Valid` entry when new flags are added to the enum, but I don't feel strongly about it

benhillis (approved):

PR #40102, opened 2026-04-03
Fix port conflict error message and cleanup on failure

beena352:
Fix port conflict error message in wslc container run. When starting a container on a port that is already in use, users previously saw the misleading message “A distribution with the supplied name already exists.” With this fix, the correct error is shown (for example, “Port 8080 is already in use, cannot start container <id>”). Also clean up the container left behind when Start() fails.

## PR Checklist

- [ x] **Communication:** I've discussed this with core contributors already. If work hasn't been agreed, this work might be rejected


benhillis:
Hey @beena352 👋 — Following up on this draft PR. There are **2 unresolved review threads**:
- Hardcoded strings that need localization
- Port reporting inconsistency in host networking mode

Is this change still being worked on? Addressing those review items would help move this forward.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1443,20 +1441,31 @@ void WSLCContainerImpl::MapPorts()
...
                     m_virtualMachine.TryAllocatePort(e.ContainerPort, e.VmMapping.BindAddress.si_family, e.VmMapping.Protocol);
 
-                THROW_HR_IF_MSG(
+                THROW_HR_WITH_USER_ERROR_IF(
                     HRESULT_FROM_WIN32(ERROR_ALREADY_EXISTS),
-                    !allocatedPort,
-                    "Port %hu is in use, cannot start container %hs",
-                    e.ContainerPort,
-                    m_id.c_str());
+                    std::format(L"Port {} is already in use, cannot start container {}", e.VmMapping.HostPort(), m_id),
```
OneBlue: nit: I recommend logging the full port specific here (binding address, protocol and port number)

OneBlue: nit: This error should be localized (see localization\strings\en-US\Resources.resw)

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1274,12 +1274,10 @@ std::unique_ptr<WSLCContainerImpl> WSLCContainerImpl::Open(
         {
             auto allocation = virtualMachine.TryAllocatePort(e.VmPort, e.Family, e.Protocol);
 
-            THROW_HR_IF_MSG(
+            THROW_HR_WITH_USER_ERROR_IF(
                 HRESULT_FROM_WIN32(ERROR_ALREADY_EXISTS),
```
OneBlue, line 1342: I don't think we should throw a user error here, this code path is only used when we're opening containers from storage, so this error message would be pretty confusing for a user to see

OneBlue: (it should also, in theory, be impossible to hit)

src/windows/wslcsession/WSLCContainer.cpp
beena352: Agreed, will do.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1451,20 +1451,33 @@ void WSLCContainerImpl::MapPorts()
...
-                THROW_HR_IF_MSG(
+                THROW_HR_WITH_USER_ERROR_IF(
                     HRESULT_FROM_WIN32(ERROR_ALREADY_EXISTS),
-                    !allocatedPort,
-                    "Port %hu is in use, cannot start container %hs",
-                    e.ContainerPort,
-                    m_id.c_str());
+                    wsl::shared::Localization::MessageWslcPortInUse(
+                        std::format("{}:{}/{}", e.VmMapping.BindingAddressString(), e.VmMapping.HostPort(), e.ProtocolString()), m_id),
+                    !allocatedPort);
```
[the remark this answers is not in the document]
beena352: IPv6 port mappings aren't supported yet, but good catch, I'll wrap the address.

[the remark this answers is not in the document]
OneBlue: ipv6 port mappings are currently supported (at least at the service level). Here's a sample test case that uses an ipv6 port mapping: https://github.com/microsoft/WSL/blob/ed62b7c8cbfa66bee2b48bee2076dad93394d721/test/windows/WSLCTests.cpp#L4490-L4491

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -231,6 +231,16 @@ std::string ExtractContainerName(const std::vector<std::string>& names, const st
     return CleanContainerName(names[0]);
 }
 
+std::string FormatPortEndpoint(const ContainerPortMapping& portMapping)
+{
+    auto addr = portMapping.VmMapping.BindingAddressString();
+    return std::format(
+        "{}:{}/{}",
+        (addr.find(':') != std::string::npos) ? std::format("[{}]", addr) : addr,
```
OneBlue: nit: I would recommend adding an `IsIpv6()` method to `VMPortMapping` instead

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1509,20 +1519,30 @@ void WSLCContainerImpl::MapPorts()
...
                 allocatedPorts.emplace(e.ContainerPort, allocatedPort);
             }
         }
 
-        m_virtualMachine.MapPort(e.VmMapping);
+        try
+        {
+            m_virtualMachine.MapPort(e.VmMapping);
+        }
+        catch (const wil::ResultException& ex)
```
OneBlue: nit: I recommend using this method: 

```
try
{
    [...]
}
catch (...)
{
    auto result = wil::ResultFromCaughtException();
}
```

That way we'll also be able to handle errors not directly reported through wil (like STL and cppwinrt errors)

localization/strings/en-US/Resources.resw
```diff
@@ -2114,6 +2114,10 @@ For privacy information about this product please visit https://aka.ms/privacy.<
     <value>Failed to create volume '{}': {}</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
+  <data name = "MessageWslcPortInUse" xml:space = "preserve" >
+    <value>Port {} is already in use, cannot start container {}</value>
```
OneBlue, line 2130: nit: I'd recommend also including the container id in the error, that way if the user tries to start multiple containers in a `wslc start` command, it will be easier to know which one failed to start

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -945,9 +944,16 @@ class WSLCE2EContainerCreateTests
             GetPythonHttpServerScript(ContainerTestPort)));
         result1.Verify({.Stderr = L"", .ExitCode = 0});
 
+        auto containersBefore = ListAllContainers().size();
+
         // Attempt to start another container mapping the same host port
         auto result2 = RunWslc(std::format(L"container run -p {}:{} {}", HostTestPort1, ContainerTestPort, DebianImage.NameAndTag()));
         result2.Verify({.ExitCode = 1});
+        VERIFY_IS_TRUE(result2.Stderr.has_value() && result2.Stderr.value().find(L"is already in use") != std::wstring::npos);
```
OneBlue: nit: I recommend validating the complete error message

localization/strings/en-US/Resources.resw
beena352: The container ID is already included as the second {} placeholder:
PS Q:\source\WSL> bin\x64\Debug\wslc.exe container run -p 8080:8080 -d debian:latest
Port 127.0.0.1:8080/tcp is already in use, cannot start container d83c84b2809c8a47cef69dd17907982715fdfce2f665046d78027f95313d0789
Error code: ERROR_ALREADY_EXISTS

OneBlue (approved):

PR #40104, opened 2026-04-04
wslc: implement dns tunneling for virtio proxy networking mode

benhillis:
This change implements dns tunneling for the virtioproxy networking mode. This will go through the same path as dnstunneling for NAT until the built in dns support is fixed in the device host dll.

### DIFF
--- a/src/windows/common/VirtioNetworking.cpp
+++ b/src/windows/common/VirtioNetworking.cpp
@@ -15,13 +15,29 @@ static constexpr auto c_eth0DeviceName = L"eth0";
 static constexpr auto c_loopbackDeviceName = TEXT(LX_INIT_LOOPBACK_DEVICE_NAME);
 
 VirtioNetworking::VirtioNetworking(
-    GnsChannel&& gnsChannel, VirtioNetworkingFlags flags, LPCWSTR dnsOptions, std::shared_ptr<GuestDeviceManager> guestDeviceManager, wil::shared_handle userToken) :
+    GnsChannel&& gnsChannel,
+    VirtioNetworkingFlags flags,
+    LPCWSTR dnsOptions,
+    std::shared_ptr<GuestDeviceManager> guestDeviceManager,
+    wil::shared_handle userToken,
+    wil::unique_socket&& dnsHvsocket) :
     m_guestDeviceManager(std::move(guestDeviceManager)),
     m_userToken(std::move(userToken)),
     m_gnsChannel(std::move(gnsChannel)),
     m_flags(flags),
     m_dnsOptions(dnsOptions)
 {
+    THROW_HR_IF_MSG(
+        E_INVALIDARG,
+        ((!!dnsHvsocket != WI_IsFlagSet(m_flags, VirtioNetworkingFlags::DnsTunnelingSocket)) ||
+         (WI_IsFlagSet(m_flags, VirtioNetworkingFlags::DnsTunnelingSocket) && WI_IsFlagSet(m_flags, VirtioNetworkingFlags::DnsTunneling))),
+        "Incompatible DNS settings");
+
+    if (dnsHvsocket)
+    {
+        networking::DnsResolverFlags resolverFlags{};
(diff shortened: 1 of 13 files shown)


src/windows/wslc/settings/UserSettings.h
```diff
@@ -81,6 +82,7 @@ namespace details {
     DEFINE_SETTING_MAPPING(SessionStoragePath,       std::string, std::wstring,       {},                            "session.defaultStoragePath")
     DEFINE_SETTING_MAPPING(SessionNetworkingMode,    std::string, WSLCNetworkingMode, WSLCNetworkingModeVirtioProxy, "session.networkingMode")
     DEFINE_SETTING_MAPPING(SessionHostFileShareMode, std::string, HostFileShareMode,  HostFileShareMode::VirtioFs,   "session.hostFileShareMode")
+    DEFINE_SETTING_MAPPING(SessionDnsTunneling,      bool,        bool,               false,                         "session.dnsTunneling")
```
OneBlue: Should we just enable DNS tunneling by default ?

test/windows/WSLCTests.cpp
```diff
@@ -2331,6 +2350,77 @@ class WSLCTests
...
+        if (includeTcp)
+        {
+            VerifyDigDnsResolution(session, "dig +tcp +short +time=5 A bing.com");
+            VerifyDigDnsResolution(session, "dig +tcp +short +time=5 -x 8.8.8.8");
+        }
+    }
+
+    void VerifyDnsResolutionRecordTypes(IWSLCSession* session)
+    {
+        VerifyDigDnsResolution(session, "dig +short +time=5 MX bing.com");
```
OneBlue: Not blocking for this change, but historically DNS checks like this have been flaky. 

I wonder if we could run a container with named and use fully local DNS resolution instead

OneBlue (approved):

src/windows/wslc/settings/UserSettings.h
benhillis: good call, let's flip the default.

src/windows/common/VirtioNetworking.cpp
```diff
@@ -15,13 +15,29 @@ static constexpr auto c_eth0DeviceName = L"eth0";
...
 {
+    THROW_HR_IF_MSG(
+        E_INVALIDARG,
+        ((!!dnsHvsocket != WI_IsFlagSet(m_flags, VirtioNetworkingFlags::DnsTunnelingSocket)) ||
+         (WI_IsFlagSet(m_flags, VirtioNetworkingFlags::DnsTunnelingSocket) && WI_IsFlagSet(m_flags, VirtioNetworkingFlags::DnsTunneling))),
+        "Incompatible DNS settings");
+
+    if (dnsHvsocket)
+    {
+        networking::DnsResolverFlags resolverFlags{};
```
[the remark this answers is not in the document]
benhillis: that flag is experimental.

src/windows/service/exe/HcsVirtualMachine.cpp
```diff
@@ -369,11 +369,17 @@ try
...
         THROW_HR_IF(E_INVALIDARG, DnsSocket == nullptr);
-        THROW_HR_IF_MSG(
-            E_NOTIMPL, m_networkingMode == WSLCNetworkingModeVirtioProxy, "DNS tunneling not supported for VirtioProxy");
 
-        THROW_IF_FAILED(wsl::core::networking::DnsResolver::LoadDnsResolverMethods());
-        dnsSocketHandle.reset(reinterpret_cast<SOCKET>(wslutil::DuplicateHandle(*DnsSocket)));
+        const auto result = wsl::core::networking::DnsResolver::LoadDnsResolverMethods();
+        if (FAILED(result))
+        {
+            LOG_HR_MSG(result, "Failed to load DNS resolver methods, DNS tunneling will be disabled");
```
benhillis, line 376: Decided to make asking for dns tunneling non-fatal.

PR #40116, opened 2026-04-06
Add /attachdebugger option to automatically launch WinDbgX for test debugging

benhillis:
## Summary

Adds a new `/attachdebugger` option to `test.bat` that automatically launches WinDbgX and attaches it to the test host process, replacing the manual `/waitfordebugger` workflow.

## Usage

```
bin\x64\debug\test.bat /attachdebugger /name:*UnitTest*
bin\x64\debug\test.bat /attachdebugger /inproc /name:*MyTest*
```

## How it works

When `/attachdebugger` is passed:
1. Strips it from the args and adds `/waitfordebugger` (so TAEF pauses the test host)
2. Starts `te.exe` in the background with output in the current console
3. Polls WMI for the `TE.ProcessHost.exe` child process
4. Launches `WinDbgX.exe -p <pid>` to attach directly — no child-debugging noise
5. With `/inproc`, attaches to `TE.exe` itself instead

Falls back gracefully if WinDbgX isn't installed or the child process can't be found.

## Changes
- `tools/test/run-tests.ps1` — Core implementation
- `doc/docs/dev-loop.md` — Updated debugging docs
- `.github/copilot-instructions.md` — Updated developer instructions
- `test/README.md` — Added `/attachdebugger` section

### DIFF
--- a/tools/test/run-tests.ps1
+++ b/tools/test/run-tests.ps1
@@ -54,6 +54,12 @@ if ($TeArgs -and ($TeArgs -icontains '/attachdebugger'))
     {
         $AttachDebugger = $true
         $TeArgs += '/waitfordebugger'
+        # Run in-process so WinDbgX can attach directly to TE.exe without
+        # polling for a TE.ProcessHost.exe child process.
+        if (-not ($TeArgs -icontains '/inproc'))
+        {
+            $TeArgs += '/inproc'
+        }
     }
     else
     {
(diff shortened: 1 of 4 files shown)


tools/test/run-tests.ps1
```diff
@@ -45,9 +45,52 @@ if ($Fast)
...
+        Start-Process "WinDbgX.exe" -ArgumentList "-p $targetPid"
+    }
+    else
+    {
+        Write-Warning "Could not find TE.ProcessHost.exe within 60 seconds."
+        Write-Host "Attach a debugger manually to TE.exe (PID: $($teProcess.Id))."
+    }
+}
+
+$teProcess | Wait-Process
```
OneBlue: nit: We could do `exit ($teProcess | Wait-Process -PassThru).ExitCode` here

tools/test/run-tests.ps1
```diff
@@ -45,9 +45,52 @@ if ($Fast)
...
+$teProcess = Start-Process -FilePath "te.exe" -ArgumentList $teArgList -PassThru -NoNewWindow
+
+if ($AttachDebugger)
 {
-    exit 1
-}
\ No newline at end of file
+    $targetPid = if ($TeArgs -icontains '/inproc') { $teProcess.Id } else { $null }
+
+    if (-not $targetPid)
```
OneBlue: nit: I think it would be OK to add `/inproc` if `/attachdebugger` is set so we can avoid having to poll for the right process to attach to

OneBlue (dismissed):

OneBlue (approved):

dkbennett (approved):

PR #40121, opened 2026-04-06
Definition for SDK projection / object model

florelis:
This is a proposal for a projection / object model for the SDK API, specifically intended for the C# projection.

It is written in MIDL 3, which means it represents a WinRT interface, but everything maps directly to C# so it should still allow us to talk about the design of the interface.

This is more or less a direct translation of the C header.
* Mapped all the `struct`s to classes.
* Mapped the settings/options types that had required fields to classes with a single constructor taking the required fields.
* Mapped the creation of the main objects (like `Session`) to a static `Create()` factory method to avoid having constructors that do too much work (and in case we ever want to have an async variant).
* Collapsed the functions for loading an image to a single overload taking a file path; additional overloads can be added later.
* Removed the `Release()` functions.
* Split `WslcGetProcessIOHandle()` into two functions to be able to return properly typed input/output streams.
* Removed the error message out parameters, as the language supports exceptions.

This interface could be implemented in C++ as WinRT, and then using CsWinRT we can get a pretty natural C# projection with no or minimal changes. That would allow us to not have to deal with raw pointers and having to pin data from C#.


src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -0,0 +1,348 @@
...
+    };
+
+    enum SessionTerminationReason
+    {
+        Unknown = 0,
+        Shutdown = 1,
+        Crashed = 2,
+    };
+
+    delegate void SessionTerminationHandler(Session session, SessionTerminationReason reason);
```
florelis, line 31: I changed "Callback" to "Handler" because I think that's more idiomatic in C#

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+    runtimeclass Session
+    {
+        static Session Create(SessionSettings settings);
+
+        void Terminate();
+
+        void PullImage(PullImageOptions options);
+
+        // C# equivalent type: System.IO.StreamReader
+        void ImportImage(String imageName, Windows.Storage.Streams.IInputStream imageStream, ImportImageOptions options);
```
florelis: The C API uses a HANDLE for the image content. I changed it to a stream based on what I understood from how it's used, but I may be wrong.

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+
+    runtimeclass ContainerPortMapping
+    {
+        ContainerPortMapping(UInt16 windowsPort, UInt16 containerPort, PortProtocol protocol);
+
+        UInt16 WindowsPort { get; };
+        UInt16 ContainerPort { get; };
+        PortProtocol Protocol { get; };
+
+        // TODO sockaddr_storage* WindowsAddress { get; set; };
```
florelis: This one I didn't know how to map, so I left it as a TODO for now.

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+    runtimeclass ProcessSettings
+    {
+        ProcessSettings();
+
+        String CurrentDirectory { get; set; };
+        // C projected type: System.Collections.Generic.IList<String>
+        IVector<String> CmdLine { get; set; };
+        // C# equivalent type: System.Collections.Specialized.StringDictionary
+        // C# projected type: System.Collections.Generic.IDictionary<string,string?>
+        IMap<String, String> EnvironmentVariables { get; set; };
```
florelis: This is an array of strings in the C API. I changed it to match the C# Process type.

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+    {
+        UInt32 Pid { get; };
+        ProcessState State { get; };
+        Int32 ExitCode { get; };
+
+        void Signal(Signal signal);
+        // C# equivalent Windowstype: System.IO.StreamReader
+        Windows.Storage.Streams.IInputStream GetOutputStream(ProcessOutputHandle ioHandle);
+        // C# type: System.IO.StreamWriter
+        Windows.Storage.Streams.IOutputStream GetInputStream();
```
florelis: These are a single function in the C API. For WinRT there is no generic Stream type that can do both AFAICT, but if we do direct C# implementation it can be a `System.IO.Stream` and a single function to match.

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+
+    runtimeclass ProcessSettings
+    {
+        ProcessSettings();
+
+        String CurrentDirectory { get; set; };
+        // C projected type: System.Collections.Generic.IList<String>
+        IVector<String> CmdLine { get; set; };
+        // C# equivalent type: System.Collections.Specialized.StringDictionary
+        // C# projected type: System.Collections.Generic.IDictionary<string,string?>
```
florelis: C# equivalent type: What we could use if we define it in C# directly.
C# projected type: What we would get if we use CsWinRT to get a projection from the .idl.

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -0,0 +1,348 @@
...
+        void LoadImage(Windows.Storage.Streams.IInputStream imageStream, LoadImageOptions options);
+        void LoadImageFromFile(String path, LoadImageOptions options);
+
+        void DeleteSessionImage(String nameOrId);
+
+        void CreateSessionVhdVolume(VhdRequirements options);
+        void DeleteSessionVhdVolume(String name);
+
+        // C# projected type: System.Collections.Generic.IReadOnlyList<ImageInfo>
+        IVectorView<ImageInfo> Images { get; };
```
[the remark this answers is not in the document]
florelis: This is allowed [shorthand](https://learn.microsoft.com/en-us/uwp/midl-3/advanced#shorthand) for collections.

src/windows/WslcSDK/wslcsdk-winrt.idl
OneBlue: I don't think we can easily do that unfortunately, since the actual implementation of this API will have no way to "transform" a stream back to a handle to give it to the service

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -0,0 +1,348 @@
...
+
+    runtimeclass Process
+    {
+        UInt32 Pid { get; };
+        ProcessState State { get; };
+        Int32 ExitCode { get; };
+
+        void Signal(Signal signal);
+        // C# type: System.IO.Stream
+        Windows.Storage.Streams.IInputStream GetOutputStream(ProcessOutputHandle ioHandle);
```
OneBlue: In this direction I think returning a stream is the right thing to do for C#, but doing so will prevent the caller from doing asynchronous IO on stdout & stderr. 

I'm not sure that there's an easy way to represent this in C#, but maybe what we could do could be to return a subclass of IInputStream that has a "GetHandle()" method to get the best of both worlds

OneBlue (commented):
LGTM overall, couple minor comments

src/windows/WslcSDK/wslcsdk-winrt.idl
florelis: I see. For now I'll assume we're okay if the method looks like `void ImportImage(T image, ImportImageOptions options)`, possibly with multiple overloads, and try to find types that better map to each type of handle once I'm looking at the actual implementation.

src/windows/WslcSDK/winrt/wslcsdk.idl
florelis: Why would it prevent the caller from doing asynchronous IO? Whether this ends up being a C# Stream or a WinRT IInputStream, it will have a ReadyAsync() method. Are you thinking of IStream?

OneBlue: I did a bit of research on this, and it's looking like ReadyAsync() specifically handles overlapped IO, which is what we use, so I think this **should** work, although this is definitely something that we should test

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+        Crashed = 2,
+    };
+
+    delegate void SessionTerminationHandler(Session session, SessionTerminationReason reason);
+
+    runtimeclass SessionSettings
+    {
+        SessionSettings(String name, String storagePath);
+
+        String Name { get; };
```
JohnMcPMS: These could easily be settable as well as long as they perform the same validation that the constructor does when setting.

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+        String StoragePath { get; };
+
+        UInt32 CpuCount { get; set; };
+        UInt32 MemoryMb { get; set; };
+        UInt32 TimeoutMS { get; set; };
```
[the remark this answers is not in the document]
JohnMcPMS: The API review feedback was that 2 letter "words" are all caps, 3 or more is Pascal.  That changes the feedback on how to make things consistent in that this one is correct by `MemoryMb` should be `MemoryMB` (and also the convention of megaBYTES, not megabits).

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+        VhdRequirements VhdRequirements { get; set; };
+        SessionFeatureFlags FeatureFlags { get; set; };
+        SessionTerminationHandler TerminationHandler { get; set; };
```
JohnMcPMS: `event`?  Although for idiomatic usage that should probably go onto the Session itself.  The interop `Session` object would then always provide the callback and invoke the event in turn.  That would also need handling to invoke the specific delegate if the delegate was added after the session were terminated.

src/windows/WslcSDK/wslcsdk-winrt.idl
JohnMcPMS: Feels like unless we can find a built-in function to create FILE semantics around a stream-like, we may have to just rely on the file path function.

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+    {
+        static Session Create(SessionSettings settings);
+
+        void Terminate();
+
+        void PullImage(PullImageOptions options);
+
+        // C# type: System.IO.Stream
+        void ImportImage(String imageName, Windows.Storage.Streams.IInputStream imageStream, ImportImageOptions options);
+        void ImportImageFromFile(String imageName, String path, ImportImageOptions options);
```
JohnMcPMS: Are we avoiding overloads intentionally?

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -0,0 +1,348 @@
...
+        UInt32 MemoryMb { get; set; };
+        UInt32 TimeoutMS { get; set; };
+        VhdRequirements VhdRequirements { get; set; };
+        SessionFeatureFlags FeatureFlags { get; set; };
+        SessionTerminationHandler TerminationHandler { get; set; };
+    };
+
+    runtimeclass Session
+    {
+        static Session Create(SessionSettings settings);
```
JohnMcPMS: Why not a constructor?

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+
+        // C# type: System.IO.Stream
+        void LoadImage(Windows.Storage.Streams.IInputStream imageStream, LoadImageOptions options);
+        void LoadImageFromFile(String path, LoadImageOptions options);
+
+        void DeleteSessionImage(String nameOrId);
```
JohnMcPMS: Drop "Session" from all function names (here and 2 below).

src/windows/WslcSDK/wslcsdk-winrt.idl
JohnMcPMS: I don't know the full use case of it, but currently we just read the IP version and IP address out of it (it is just big enough to hold either type).  Feels like this could just be replaced with those to values, which may already exist as types if we want to leverage extant WinRT.

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -0,0 +1,348 @@
...
+
+    enum ProcessOutputHandle
+    {
+        StandardOutput = 1,
+        StandardError = 2,
+    };
+
+    delegate void ProcessOutputHandler(UInt8[] data);
+
+    delegate void ProcessExitHandler(Process process, Int32 exitCode);
```
JohnMcPMS: The output handler excludes all context information, presumably expecting the caller to store it in their delegate object.

This handler includes the process object.

Why the difference?

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+    {
+        StandardOutput = 1,
+        StandardError = 2,
+    };
+
+    delegate void ProcessOutputHandler(UInt8[] data);
+
+    delegate void ProcessExitHandler(Process process, Int32 exitCode);
+
+    runtimeclass ProcessEventHandlers
```
JohnMcPMS: This was only a separate struct as an argument (and internal usage) mechanism.  It should just be combined into `ProcessSettings`.

src/windows/WslcSDK/wslcsdk-winrt.idl
```diff
@@ -0,0 +1,348 @@
...
+        UInt32 Revision;
+    };
+
+    delegate void InstallProgressHandler(ComponentFlags component, UInt32 progress, UInt32 total);
+
+    runtimeclass WslcService
+    {
+        static Boolean CanRun(out ComponentFlags missingComponents);
+        static Version GetVersion();
+        static void InstallWithDependencies(InstallProgressHandler progressCallback);
```
JohnMcPMS: It seems odd to me that we would limit ourselves to WinRT containers and streams and not use the async paradigms.

src/windows/WslcSDK/winrt/wslcsdk.idl
florelis: Because the creation can fail and I wouldn't expect a constructor to fail.

src/windows/WslcSDK/winrt/wslcsdk.idl
florelis: No reason; the output handler should also have it. I didn't include it because I didn't immediately think of a use for the process object, but I think in both cases it needs to be included as the sender.

src/windows/WslcSDK/wslcsdk-winrt.idl
florelis: No, I just didn't think of it.

src/windows/WslcSDK/wslcsdk-winrt.idl
florelis: I considered using an event in the Session, but then there would be a potential gap if the handler is added after it is terminated. I don't think it should be an event in the settings because the session would have to somehow take ownership of the event (if not the whole settings object) and that feels weird. But this still doesn't feel right, now that I'm thinking about the lifetime of the listener

src/windows/WslcSDK/wslcsdk-winrt.idl
JohnMcPMS: Given the implementation in the other PR, not so easy to set.  Isn't a high priority.

src/windows/WslcSDK/wslcsdk-winrt.idl
JohnMcPMS: > then there would be a potential gap if the handler is added after it is terminated

That is what I was mentioning, the Session would need to keep track that it was already terminated and invoke the incoming delegate immediately.  That is itself a more complicated thing to handle on the client side, but I think it would always be the case for something like this.

src/windows/WslcSDK/winrt/wslcsdk.idl
JohnMcPMS: I don't feel like either C++ or C# hold that line, and I don't think that the COM underlying this does either (there is still an HRESULT returned).  I'm not hard line either way really, just seems more natural to me to use the constructor than needing to invoke a static.

src/windows/WslcSDK/wslcsdk-winrt.idl
florelis: I imagine it would be easier if I used the service API directly, but using the public SDK was quicker since that's what this API is based on. Maybe could change later, especially once we have an API review

benhillis:

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -0,0 +1,327 @@
...
+        Crashed = 2,
+    };
+
+    delegate void SessionTerminationHandler(Session session, SessionTerminationReason reason);
+
+    runtimeclass SessionSettings
+    {
+        SessionSettings(String name, String storagePath);
+
+        String Name;
```
JohnMcPMS, line 37: We kept the return codes on the individual set functions so that these properties could fail as they are invoked.  This change and your comments suggest that you moved the implementation to not call the set functions per-property set.  I prefer it that the properties fail as you set them rather than trying to determine what failed from messages or logs later on.

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -0,0 +1,327 @@
...
+    {
+        static Session Create(SessionSettings settings);
+        void Terminate();
+
+        Container CreateContainer(ContainerSettings containerSettings);
+
+        Windows.Foundation.IAsyncActionWithProgress<ImageProgress> PullImageAsync(PullImageOptions options);
+        // TODO: Explore additional overloads for other types of input streams
+        Windows.Foundation.IAsyncActionWithProgress<ImageProgress> ImportImageAsync(String path, String imageName);
+        Windows.Foundation.IAsyncActionWithProgress<ImageProgress> LoadImageAsync(String path);
```
JohnMcPMS, line 59: Should we use options for these as well so that we don't need to add overloads if we can ever take in handles?

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -0,0 +1,327 @@
...
+        ContainerState State { get; };
+    };
+
+    enum ProcessOutputHandle
+    {
+        StandardOutput = 1,
+        StandardError = 2,
+    };
+
+    delegate void ProcessOutputHandler(Process process, UInt8[] data);
```
JohnMcPMS: `ProcessOutputHandle` parameter as well?

florelis: There's no need since each handle gets its own event.

src/windows/WslcSDK/winrt/wslcsdk.idl
florelis: In the implementation I'll have validation in the setters, before creating the raw object and setting its properties.

src/windows/WslcSDK/winrt/wslcsdk.idl
florelis: If we ever support using handles instead of paths, I'd prefer the overloads than having an options object that could hold both. I think the path/handle would be the first parameter directly, and everything else would be in an options object as the second parameter.

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -0,0 +1,328 @@
...
+    {
+        SessionSettings(String name, String storagePath);
+
+        String Name;
+        String StoragePath;
+
+        Windows.Foundation.IReference<UInt32> CpuCount;
+        Windows.Foundation.IReference<UInt32> MemoryMB;
+        Windows.Foundation.IReference<UInt32> TimeoutMS;
+        VhdRequirements VhdRequirements;
```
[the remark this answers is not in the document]
florelis: I'll change the name, but it makes more sense to me to name the type 'options' and the property 'requirements'

src/windows/WslcSDK/winrt/Session.cpp
```diff
@@ -13,44 +13,74 @@ Module Name:
...
+}
+winrt::event_token Session::Terminated(winrt::Microsoft::WSL::Containers::SessionTerminationHandler const& handler)
+{
+    throw hresult_not_implemented();
+}
+void Session::Terminated(winrt::event_token const& token) noexcept
 {
-    return &m_session;
+    assert(false); // TODO: not implemented, but this can't throw
 }
```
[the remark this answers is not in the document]
florelis: This is intentionally a stub with no implementation for now.

src/windows/WslcSDK/winrt/WslcService.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace winrt::Microsoft::WSL::Containers::implementation {
+struct WslcService
+{
+    WslcService() = default;
+
+    static winrt::Microsoft::WSL::Containers::ComponentFlags GetMissingComponents();
+    static winrt::Microsoft::WSL::Containers::ServiceVersion GetVersion();
+    static winrt::Windows::Foundation::IAsyncActionWithProgress<winrt::Microsoft::WSL::Containers::InstallProgress> InstallWithDependenciesAsync();
+};
```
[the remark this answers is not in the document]
florelis: This compiles fine. I think it being a static class makes a difference.

src/windows/WslcSDK/winrt/Session.cpp
```diff
@@ -13,44 +13,74 @@ Module Name:
...
-WSLC::Session implementation::Session::Create(WSLC::SessionSettings const& settings)
+winrt::Microsoft::WSL::Containers::Session Session::Start()
 {
-    auto session = winrt::make_self<implementation::Session>();
-    wil::unique_cotaskmem_string errorMessage;
-    auto hr = WslcCreateSession(implementation::GetStructPointer(settings), implementation::GetHandlePointer(session), &errorMessage);
-    THROW_MSG_IF_FAILED(hr, errorMessage);
-    return *session;
+    throw hresult_not_implemented();
 }
```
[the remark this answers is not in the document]
florelis: I just overwrote everything with the new stubs from the updated IDL. This isn't used yet, so there's no harm in not re-adding an implementation yet.

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -14,32 +14,250 @@ Abstract:
...
+        Windows.Foundation.IReference<UInt32> TimeoutMS;
+        VhdOptions VhdRequirements;
+        SessionFeatureFlags FeatureFlags;
     };
 
     runtimeclass Session
     {
         static Session Create(SessionSettings settings);
 
+        Session Start();
```
JohnMcPMS: I know I mentioned this model, but seeing it here without any other use makes it seem out of place.  And it makes me think that callers might be confused about whether it is a new Session or the existing Session.

[Applies to Process as well]

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -14,32 +14,250 @@ Abstract:
...
+        Windows.Foundation.IReference<UInt32> CpuCount;
+        Windows.Foundation.IReference<UInt32> MemoryMB;
+        Windows.Foundation.IReference<UInt32> TimeoutMS;
+        VhdOptions VhdRequirements;
+        SessionFeatureFlags FeatureFlags;
     };
 
     runtimeclass Session
     {
         static Session Create(SessionSettings settings);
```
JohnMcPMS: Now that this effectively cannot fail, would you feel better about it being a constructor?

JohnMcPMS (commented):
Additionally, it might be nice to have a read-only copy of the settings object hanging off the object that they created.  `Session`, `Container`, `Process` having a `CreationSettings` get-property that is a clone of the settings with an internal bool that prevents all attempts to set.  Could be added later easily enough though.

src/windows/WslcSDK/winrt/wslcsdk.idl
florelis: Changed to `void Start()`

src/windows/WslcSDK/winrt/wslcsdk.idl
florelis: Changed to a constructor. For Process/Container, I still left it as a method in Container/Session

JohnMcPMS (approved):

PR #40125, opened 2026-04-07
Add E2E test for --version flag

ptrivedi:
Ensures both entry points (the 'version' subcommand and the '--version' flag) are covered by E2E tests, preventing the two paths from drifting.

### DIFF
--- a/test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
+++ b/test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
@@ -56,6 +56,12 @@ class WSLCE2EGlobalTests
         WSL2_TEST_ONLY();
         RunWslcAndVerify(L"version", {.Stdout = GetVersionMessage(), .Stderr = L"", .ExitCode = 0});
     }
+
+    TEST_METHOD(WSLCE2E_VersionFlag)
+    {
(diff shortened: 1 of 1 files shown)


test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
```diff
@@ -56,6 +56,12 @@ class WSLCE2EGlobalTests
         WSL2_TEST_ONLY();
         RunWslcAndVerify(L"version", {.Stdout = GetVersionMessage(), .Stderr = L"", .ExitCode = 0});
     }
+
+    TEST_METHOD(WSLCE2E_VersionFlag)
+    {
+        WSL2_TEST_ONLY();
+        RunWslcAndVerify(L"--version", {.Stdout = GetVersionMessage(), .Stderr = L"", .ExitCode = 0});
+    }
```
benhillis, line 64: nit: newline

benhillis (approved):

AmelBawa-msft (approved):
LGTM! In the future, I suspect the information for `version` and `--version` might possibly communicate different information (?). Version command will have `--format` option which prints a verbose text but also allows json format as well. Whereas `--version` will likely not be configurable so we the information printed may be minimal.

PR #40128, opened 2026-04-07
[CLI] Initialize tmpfs support

AmelBawa-msft:
- wslc create/run `--tmpfs` option

## PR Checklist

### DIFF
--- a/src/windows/wslc/services/ContainerModel.cpp
+++ b/src/windows/wslc/services/ContainerModel.cpp
@@ -283,4 +283,19 @@ std::vector<std::wstring> EnvironmentVariable::ParseFile(const std::wstring& fil
 
     return envVars;
 }
+
+TmpfsMount TmpfsMount::Parse(const std::string& value)
+{
+    TmpfsMount result{};
+    auto colonPos = value.find(':');
+    if (colonPos == std::string::npos)
+    {
+        result.m_containerPath = value;
+        return result;
+    }
+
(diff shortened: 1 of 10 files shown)


src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -283,4 +283,19 @@ std::vector<std::wstring> EnvironmentVariable::ParseFile(const std::wstring& fil
 
     return envVars;
 }
+
+TmpfsMount TmpfsMount::Parse(const std::string& value)
```
dkbennett, line 287: This parsing should follow the argument parsing example found in ArgumentValidation.cpp similar to how we do the WSLCSignal processing. We should not have arg validation parsing scattered or executing during the task flow if we can catch the error on the command line before we make any potential alterations to the system.

dkbennett (changes requested):
Looks good except the validation should be refactored.

src/windows/wslc/services/ContainerModel.cpp
dkbennett: Updating that this is OK as-is because we do not actually have any validation here. All input strings are valid so there is nothing for the CLI to catch and any bad format errors will get caught by the service.

dkbennett (approved):
Is OK, I was a little overzealous on the validation and had incorrect understanding here.

PR #40130, opened 2026-04-07
Fix race condition when building multiple IDL files

OneBlue:

This change solves a build failure that happens because multiple processes write do `dlldata*.c` as the same time. This change solves the issue by making sure that they don't run at the same time 

## PR Checklist


yao-msft (approved):
Recently every time I rebuild I hit this

OneBlue:
> Recently every time I rebuild I hit this

Same !

benhillis (approved):

PR #40131, opened 2026-04-07
CLI: Fix forwarded args beginning with '-' from being a parser error

dkbennett:
This fixes a parser bug where positional arguments in commands with forwarded arguments after the first were considered errors if they began with '-'. Now for any command with forwarded args, once the first positional has been found we will accept anything as an argument for subsequent positionals as they may be part of a forwarded arg string. We simply assume nothing about arguments after the anchor positional has been found when forwarded args are present. 

This fix is constrained to situations with forward args are present in the command and so we must be more permissible in the parsing. Existing behavior for commands without forward args is unchanged.

The test cases which discovered this bug have been added to the parser and command tests.

## PR Checklist

It is as the summary states. In addition, the parser tests were adjusted to correctly use the "ImageId" to mimic the run command instead of the ContainerId argument. This is not a functional change in the test, just fixed the argument type and adjusted the tests to correctly reflect that it is the imageid being specified, not a containerid.

Also added tests for multi-line specifiers with backslashes in the command line based on how Windows and the container shell will process them.

## Validation Steps Performed
* Manual test of the deployed fix.
* Added unit tests to ensure the behavior is correct.
* Added test cases that discovered the bug as test cases in the unit tests to ensure no regression.

### DIFF
--- a/test/windows/wslc/CommandLineTestCases.h
+++ b/test/windows/wslc/CommandLineTestCases.h
@@ -55,6 +55,18 @@ COMMAND_LINE_TEST_CASE(L"container list --format json", L"list", true)
 COMMAND_LINE_TEST_CASE(L"container list --format table", L"list", true)
 COMMAND_LINE_TEST_CASE(L"container list --format badformat", L"list", false)
 COMMAND_LINE_TEST_CASE(L"run ubuntu", L"run", true)
+COMMAND_LINE_TEST_CASE(L"run --rm -it --entrypoint bash archlinux:latest -c \"echo 123\"", L"run", true)
+COMMAND_LINE_TEST_CASE(L"run jrottenberg/ffmpeg:4.4-alpine -i http://url/to/media.mp4 -stats", L"run", true)
+COMMAND_LINE_TEST_CASE(
+    L"run -v ${PWD}:/data jrottenberg/ffmpeg:4.4-scratch -stats -i http://www.hevc-10bit.mkv -c:v libx265 -pix_fmt yuv420p10 -t "
+    L"5 -f mp4 test.mp4",
+    L"run",
+    true)
+COMMAND_LINE_TEST_CASE(
+    L"run -v ${PWD}:/data -it jrottenberg/ffmpeg:4.4-scratch -stats -i https://file-examples/file_example_MP4_480_1_5MG.mp4 -c:v "
+    L"libx265 -pix_fmt yuv420p10 -t 5 -f mp4 /dataout.mp4",
+    L"run",
+    true)
 COMMAND_LINE_TEST_CASE(L"container run ubuntu bash -c 'echo Hello World'", L"run", true)
 COMMAND_LINE_TEST_CASE(L"container run ubuntu", L"run", true)
 COMMAND_LINE_TEST_CASE(L"container run -it --name foo ubuntu", L"run", true)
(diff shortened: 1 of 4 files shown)


test/windows/wslc/CommandLineTestCases.h
```diff
@@ -55,6 +55,18 @@ COMMAND_LINE_TEST_CASE(L"container list --format json", L"list", true)
 COMMAND_LINE_TEST_CASE(L"container list --format table", L"list", true)
 COMMAND_LINE_TEST_CASE(L"container list --format badformat", L"list", false)
 COMMAND_LINE_TEST_CASE(L"run ubuntu", L"run", true)
+COMMAND_LINE_TEST_CASE(L"run --rm -it --entrypoint bash archlinux:latest -c \"echo 123\"", L"run", true)
```
AmelBawa-msft, line 58: Can we also add the following E2E test? That's a pretty common scenario that would leverage this fix
```
 wslc run --rm --entrypoint /bin/bash debian:latest -c ls
```

dkbennett: Done and added (it passes of course).

test/windows/wslc/CommandLineTestCases.h
```diff
@@ -77,7 +90,6 @@ COMMAND_LINE_TEST_CASE(L"exec --workdir /app cont1 echo Hello", L"exec", true)
 COMMAND_LINE_TEST_CASE(L"exec -w /app cont1 echo Hello", L"exec", true)
 COMMAND_LINE_TEST_CASE(L"container exec --workdir /app cont1 sh", L"exec", true)
 COMMAND_LINE_TEST_CASE(L"exec --workdir", L"exec", false)                       // Missing value for --workdir
-COMMAND_LINE_TEST_CASE(L"exec cont1 --workdir", L"exec", false)                 // Invalid argument specifier after container id
```
AmelBawa-msft: Was this removed during merge conflict?

AmelBawa-msft (approved):

test/windows/wslc/WSLCCLIParserUnitTests.cpp
```diff
@@ -73,16 +73,23 @@ class WSLCCLIParserUnitTests
                     stateMachine.ThrowIfError();
                 }
```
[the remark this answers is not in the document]
dkbennett: This was a good catch, I made the tests more robust here and ensured failed tests did in fact fail and added in required argument validation.

beena352 (approved):

test/windows/wslc/ParserTestCases.h
```diff
@@ -76,42 +76,50 @@ inline std::vector<wsl::windows::wslc::Argument> GetArgumentsForSet(ArgumentSet
...
+WSLC_PARSER_TEST_CASE(Run, true, LR"(wslc image1 command forward"hello world")") \
+WSLC_PARSER_TEST_CASE(Run, true, LR"(wslc image1 command f="hello world" forward echo)") \
+WSLC_PARSER_TEST_CASE(Run, true, LR"(wslc --verbose image1 command f="hello world" forward echo)") \
+WSLC_PARSER_TEST_CASE(Run, true, LR"(wslc image1 \\command\\?"" --f -z forward hello world)") \
+\
+/* Once the image name is parsed, the next token becomes the optional <command> positional \
+ * and everything after that goes into ForwardArgs. Neither <command> nor ForwardArgs are  \
+ * interpreted as wslc options. The second case uses '\' + newline between tokens, which   \
+ * CommandLineToArgvW passes through as literal '\' tokens that the container shell        \
+ * handles correctly. */ \
```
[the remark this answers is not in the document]
dkbennett: I've already reworded this multiple times from copilot suggestions, so you're just criticizing your own comments and I'm done with this cycle. It's fine.

PR #40132, opened 2026-04-08
Initial prune image support in wslc service

yao-msft:
Initial prune image support in wslc service

## PR Checklist

## Validation Steps Performed
Added tests.


test/windows/WSLCTests.cpp
```diff
@@ -6645,6 +6645,109 @@ class WSLCTests
...
+        // Validate dangling prune: create a dangling image by re-tagging, then prune it.
+        {
+            // Load alpine, then tag debian:latest as alpine:latest to make the original alpine dangling.
+            LoadTestImage("alpine:latest");
+            auto cleanup = wil::scope_exit([&]() {
+                pruneImages(WSLCPruneImagesFlagsDanglingTrue);
+                LOG_IF_FAILED(DeleteImageNoThrow("alpine:latest", WSLCDeleteImageFlagsNone).first);
+            });
+
+            WSLCTagImageOptions tagOptions{.Image = "debian:latest", .Repo = "alpine", .Tag = "latest"};
```
OneBlue: nit: I would recommend using a different tag than debian:latest, because if this test leaves debian:latest as a tag actually pointing to alpine:latest, that could lead to some confusing errors later

test/windows/WSLCTests.cpp
```diff
@@ -6645,6 +6645,109 @@ class WSLCTests
...
+            // Mutually exclusive dangling flags.
+            ImagePruneResult result;
+            WSLCPruneImagesOptions invalidOptions{};
+            invalidOptions.Flags = WSLCPruneImagesFlagsDanglingTrue | WSLCPruneImagesFlagsDanglingFalse;
+            VERIFY_ARE_EQUAL(m_defaultSession->PruneImages(&invalidOptions, &result.result), E_INVALIDARG);
+
+            // Invalid flags.
+            invalidOptions.Flags = 0x4;
+            VERIFY_ARE_EQUAL(m_defaultSession->PruneImages(&invalidOptions, &result.result), E_INVALIDARG);
+        }
```
OneBlue, line 6783: nit: We should also add test coverage for label filters

src/windows/service/inc/wslc.idl
```diff
@@ -615,6 +648,7 @@ interface IWSLCSession : IUnknown
     HRESULT DeleteImage([in] const WSLCDeleteImageOptions* Options, [out, size_is(, *Count)] WSLCDeletedImageInformation** DeletedImages, [out] ULONG* Count);
     HRESULT TagImage([in] const WSLCTagImageOptions* Options);
     HRESULT InspectImage([in] LPCSTR ImageNameOrId, [out] LPSTR* Output);
+    HRESULT PruneImages([in, unique] const WSLCPruneImagesOptions* Options, [out] WSLCPruneImagesResults* Result);
```
OneBlue: One thing we could consider to make this a bit nicer to the caller would be to do something like this: 

```
    HRESULT PruneImages([in, unique] const WSLCPruneImagesOptions* Options, 
[out, size_is(DeletedImagesCount)] WSLCDeletedImageInformation* DeletedImages,
[out] ULONG* DeletedImagesCount,
[out] ULONGLONG* SpaceReclaimed)
```

That way don't need `ImagePruneResult`, we can directly use a `wil::unique_cotaskmem_array_ptr` to delete the returned list for us

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1317,6 +1317,110 @@ try
...
+        {
+            filters.until = Options->Until;
+        }
+
+        if (Options->Labels != nullptr && Options->LabelsCount > 0)
+        {
+            for (ULONG i = 0; i < Options->LabelsCount; ++i)
+            {
+                const auto& filter = Options->Labels[i];
+                if (filter.Label.Key != nullptr)
```
OneBlue: Should we fail if `filter.Label.Key` is a null here ?

src/windows/service/inc/wslc.idl
```diff
@@ -547,6 +547,39 @@ typedef struct _WSLCTagImageOptions
     LPCSTR Tag;   // Target tag name.
 } WSLCTagImageOptions;
 
+typedef struct _WSLCPruneImageLabelFilter
+{
+    WSLCLabel Label;
+    BOOL Present; // If true, filter:{label=key(=value)}, if false filter:{label!=key(=value)}
+} WSLCPruneImageLabelFilter;
```
OneBlue: Could we factor this with `WSLCContainerPruneFilter` ? 

It would be nice to use the same type, and if possible factor the json structure & generation as well

OneBlue (commented):

src/windows/wslcsession/WSLCSession.cpp
yao-msft: We could, I'm just following previous ListImages pattern. I guess I can make both return E_INVALIDARG if key is null

test/windows/WSLCTests.cpp
```diff
@@ -6629,14 +6629,160 @@ class WSLCTests
...
+
+            // Prune with a label filter that no dangling image has - should not prune anything.
+            auto [deletedImages, spaceReclaimed] =
+                pruneImages(WSLCPruneImagesFlagsNone, 0, {{.Key = "nonexistent.label", .Value = nullptr, .Present = true}});
+            VERIFY_ARE_EQUAL(deletedImages.size(), 0u);
+
+            // Prune with absent label filter - dangling image doesn't have the label, so it matches.
+            auto [deletedImages2, spaceReclaimed2] =
+                pruneImages(WSLCPruneImagesFlagsNone, 0, {{.Key = "nonexistent.label", .Value = nullptr, .Present = false}});
+            VERIFY_IS_TRUE(deletedImages2.size() > 0);
```
OneBlue, line 6733: nit: I recommend adding a test case for a label value match so we can test all code paths

OneBlue (approved):
LGTM, one minor test suggestion (feel free to do as a followup)

test/windows/WSLCTests.cpp
yao-msft: Currently our test images do not have labels. I'll add this test later when we touch the testdata packages.

PR #40133, opened 2026-04-08
Design: WslcGetCLISession API for inner-loop development

yeelam-gordon:
## Summary

Design document for the `WslcGetCLISession` public API in the WSLC library.

This API returns the active CLI session for the current process, enabling inner-loop development flows (build, run, debug) where a Windows application needs to interact with the container session established by the WSLC toolchain.

## Key Design Decisions

- **STDAPI/HRESULT** return type  consistent with all existing WSL public APIs
- **Ref-counted owned handle**  caller gets an owned reference; `WslcCloseSession()` only releases the caller's ref, never the CLI session itself
- **Process-scoped publish-once** via CAS  thread-safe singleton, never replaced
- **`#ifdef _DEBUG` pattern**  debug builds reuse CLI session, release builds create standalone session; container lifecycle code is identical

## Files Changed

- `doc/docs/technical-documentation/wslc-get-cli-session-api.md`  design document
- `doc/mkdocs.yml`  added to site navigation under WSLC API section

## Review Notes

This is a design document only  no implementation code. Looking for feedback on:
- API surface and naming
- Ownership/lifetime model
- Error code allocation
- Implementation file locations

### DIFF
--- a/doc/docs/technical-documentation/wslc-get-cli-session-api.md
+++ b/doc/docs/technical-documentation/wslc-get-cli-session-api.md
@@ -1,0 +1,330 @@
+# WslcGetCLISession API Design
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


yeelam-gordon:
This PR #40141 implements already. I will close this one.

PR #40139, opened 2026-04-08
CLI: Initial support for volume commands

AmelBawa-msft:
### volume create
```
C:\> wslc volume create --opt SizeBytes=1073741824 demo
```
#### Options
| Alias | Option| Description                              |
|--------|------------|------------------------------------------|
| -d     | --driver   | Specify volume driver name (default vhd) |
| -o     | --opt      | Set driver specific options              |

### volume rm
```
C:\> wslc volume rm demo
```

### volume list
```
C:\> wslc volume ls

DRIVER   VOLUME NAME
vhd      demo
```
#### Options
| Alias | Option| Description                              |
|--------|------------|------------------------------------------|
|        | --format | Output formatting (json or table) (Default: table) |
| -q     | --quiet      | Outputs the volume names only              |

### volume inspect
```
C:\> wslc volume inspect demo
[
  {
    "Name": "demo",
    "Type": "vhd",
    "VhdVolume": {
      "HostPath": "C:\\Users\\<USER_NAME>\\AppData\\Local\\wslc\\sessions\\wslc-cli-admin\\volumes\\demo.vhdx",
      "SizeBytes": 1073741824
    }
  }
]
```

## PR Checklist


test/windows/wslc/e2e/WSLCE2EVolumeRemoveTests.cpp
```diff
@@ -0,0 +1,187 @@
...
+        EnsureContainerDoesNotExist(WslcContainerName);
+        EnsureVolumeDoesNotExist(TestVolumeName);
+        EnsureVolumeDoesNotExist(TestVolumeName2);
+        return true;
+    }
+
+    TEST_CLASS_CLEANUP(ClassCleanup)
+    {
+        EnsureContainerDoesNotExist(WslcContainerName);
+        EnsureImageIsDeleted(DebianImage);
```
OneBlue, line 43: We should not delete the debian image in the tests cleanup, since that would cause it to be reimported later and make the tests slower. 

This isn't blocking for this change, but I think we should figure out one place to import all of our test images so the tests don't delete & re-import the same tests images during a testpass

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -213,26 +213,28 @@ VolumeMount VolumeMount::Parse(const std::wstring& value)
...
+    else
     {
-        THROW_HR_WITH_USER_ERROR(E_INVALIDARG, Localization::WSLCCLI_VolumeHostPathInvalid(value, rawHostPath));
-    }
+        // Not a named volume, so it must be a path.
+        // Use wil::GetFullPathNameW to resolve relative paths against the CWD.
+        std::wstring resolvedHostPath;
+        if (FAILED(wil::GetFullPathNameW(rawHostPath.c_str(), resolvedHostPath)))
+        {
+            THROW_HR_WITH_USER_ERROR(E_INVALIDARG, Localization::WSLCCLI_VolumeHostPathInvalid(value, rawHostPath));
```
OneBlue: We should throw the error code that `wil::GetFullPathNameW` returned to us here

src/windows/wslc/services/VolumeModel.h
```diff
@@ -0,0 +1,38 @@
...
+--*/
+
+#pragma once
+
+#include "JsonUtils.h"
+#include <string>
+
+namespace wsl::windows::wslc::models {
+
+struct VolumeInformation
```
OneBlue: Could we reuse `WSLCVolumeInformation` directly instead ? We can define a `to_json` from that type to allow easy serialization to make things easier

OneBlue (commented):
LGTM, minor comments

test/windows/wslc/e2e/WSLCE2EVolumeRemoveTests.cpp
AmelBawa-msft: I created a task to think through this more thoroughly:
http://task.ms/61936722

Right now, our E2E tests follow a model where each test group/class assumes a clean starting environment and explicitly loads the assets it needs. This is intentional to avoid cross-test interference and flakiness that can arise from residual state when tests are executed in different orders. In some cases, tests also validate behavior by asserting on the exact set of loaded images, so any leftover artifacts from other tests would cause those validations to fail.

Because of that, each test class currently performs its own cleanup as part of setup/teardown to ensure isolation.
That said, you’re absolutely right, this comes at a performance cost due to repeated import/unload cycles. It was a deliberate trade-off, but there may be better approaches. For example, we could explore more centralized image management or scoped cleanup strategies that still preserve isolation while reducing redundant work.
Definitely worth exploring further, thanks again for raising it

src/windows/wslc/services/VolumeModel.h
AmelBawa-msft: We could reuse WSLCVolumeInformation today, but this layer keeps the runtime decoupled from the CLI and gives us flexibility to shape the data specifically for CLI output (table and JSON) over time.

OneBlue: In this specific usecase, I think we'd want them to be coupled given that the CLI outputs everything that the service returns. 

In case like this I'd say let's reuse existing types when possible, and introduce an additional layer when the need for it arises (less types also makes code navigation easier imo)

AmelBawa-msft: That's fair, I updated to use the existing WSLCVolumeInformation and added the corresponding to_json from_json

src/shared/inc/JsonUtils.h
```diff
@@ -171,4 +172,24 @@ struct adl_serializer<wsl::shared::string::MacAddress>
...
+#ifdef WIN32
+template <>
+struct adl_serializer<WSLCVolumeInformation>
+{
+    static void to_json(json& j, const WSLCVolumeInformation& volume)
+    {
+        j = json{{"Name", std::string(volume.Name)}, {"Driver", std::string(volume.Driver)}};
+    }
+
+    static void from_json(const json& j, WSLCVolumeInformation& volume)
```
OneBlue, line 184: nit: We probably don't need from_json for this structure

OneBlue (approved):
LGTM, minor comment

src/shared/inc/JsonUtils.h
AmelBawa-msft: from_json is used by the end-to-end test to parse the stdout JSON into a WSLCVolumeInformation object

PR #40141, opened 2026-04-09
[WSLC] Add wslc get cli session api

chemwolf6922:

* Add the WslcGetCliSession api.

## PR Checklist

## Validation Steps Performed

Add test: WslcSdkTests::GetCliSession

### DIFF
--- a/test/windows/WslcSdkTests.cpp
+++ b/test/windows/WslcSdkTests.cpp
@@ -18,6 +18,7 @@ Abstract:
 #include "WslcsdkPrivate.h"
 #include "WSLCContainerLauncher.h"
 #include "wslc_schema.h"
+#include "e2e/WSLCExecutor.h"
 #include <optional>
 
 extern std::wstring g_testDataPath;
@@ -41,6 +42,16 @@ void CloseSession(WslcSession session)
 }
 
 using UniqueSession = wil::unique_any<WslcSession, decltype(CloseSession), CloseSession>;
+
+void ReleaseSession(WslcSession session)
+{
+    if (session)
+    {
+        WslcReleaseSession(session);
+    }
+}
+
+using UniqueSessionRef = wil::unique_any<WslcSession, decltype(ReleaseSession), ReleaseSession>;
 
 void CloseContainer(WslcContainer container)
 {
@@ -263,6 +274,29 @@ class WslcSdkTests
         // Null settings pointer must fail.
         UniqueSession session2;
         VERIFY_ARE_EQUAL(WslcCreateSession(nullptr, &session2, nullptr), E_POINTER);
+    }
+
+    WSLC_TEST_METHOD(GetCliSession)
+    {
+        // Null output pointer must fail.
+        VERIFY_ARE_EQUAL(WslcGetCliSession(nullptr, nullptr), E_POINTER);
+
+        // Ensure no CLI session is running.
+        WSLCE2ETests::RunWslc(L"session terminate");
+
+        // WslcGetCliSession must return ERROR_NOT_FOUND when no CLI session exists.
+        UniqueSessionRef notFoundSession;
+        VERIFY_ARE_EQUAL(WslcGetCliSession(&notFoundSession, nullptr), HRESULT_FROM_WIN32(ERROR_NOT_FOUND));
+        VERIFY_IS_NULL(notFoundSession.get());
+
+        // Start the CLI session by running a wslc command.
+        auto result = WSLCE2ETests::RunWslc(L"container list");
+        VERIFY_ARE_EQUAL(result.ExitCode.value(), (DWORD)0);
+
+        // Now WslcGetCliSession should find the running CLI session.
+        UniqueSessionRef foundSession;
+        VERIFY_SUCCEEDED(WslcGetCliSession(&foundSession, nullptr));
+        VERIFY_IS_NOT_NULL(foundSession.get());
     }
 
     WSLC_TEST_METHOD(TerminationCallbackViaTerminate)
--- a/src/windows/WslcSDK/wslcsdk.cpp
+++ b/src/windows/WslcSDK/wslcsdk.cpp
@@ -442,6 +442,26 @@ try
 }
 CATCH_RETURN();
 
+STDAPI WslcGetCliSession(_Out_ WslcSession* session, _Outptr_opt_result_z_ PWSTR* errorMessage)
+try
+{
+    RETURN_HR_IF_NULL(E_POINTER, session);
+    *session = nullptr;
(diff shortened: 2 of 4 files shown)


test/windows/WslcSdkTests.cpp
```diff
@@ -265,6 +276,23 @@ class WslcSdkTests
...
+        VERIFY_ARE_EQUAL(WslcGetCliSession(nullptr, nullptr), E_POINTER);
+
+        // Start the CLI session by running a wslc command.
+        auto result = WSLCE2ETests::RunWslc(L"container list");
+        VERIFY_ARE_EQUAL(result.ExitCode.value(), (DWORD)0);
+
+        // Now WslcGetCliSession should find the running CLI session.
+        UniqueSessionRef session;
+        VERIFY_SUCCEEDED(WslcGetCliSession(&session, nullptr));
+        VERIFY_IS_NOT_NULL(session.get());
```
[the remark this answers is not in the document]
chemwolf6922: Tests are always required to run as admin.

test/windows/WslcSdkTests.cpp
```diff
@@ -263,6 +274,21 @@ class WslcSdkTests
         VERIFY_ARE_EQUAL(WslcCreateSession(nullptr, &session2, nullptr), E_POINTER);
     }
 
+    WSLC_TEST_METHOD(GetCliSession)
```
ptrivedi, line 279: Need to add negative testing -- a test for the "session not found" failure case

src/windows/common/WSLCSessionDefaults.h
```diff
@@ -0,0 +1,44 @@
+
+/*++
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    WSLCSessionDefaults.h
```
ptrivedi, line 7: nit:  existing common/ headers use the Wsl prefix (e.g., WslSecurity.h, WslCoreMessageQueue.h). The all-caps WSLCSessionDefaults.h is inconsistent with that

test/windows/WslcSdkTests.cpp
```diff
@@ -263,6 +274,21 @@ class WslcSdkTests
         VERIFY_ARE_EQUAL(WslcCreateSession(nullptr, &session2, nullptr), E_POINTER);
     }
 
+    WSLC_TEST_METHOD(GetCliSession)
+    {
+        // Null output pointer must fail.
+        VERIFY_ARE_EQUAL(WslcGetCliSession(nullptr, nullptr), E_POINTER);
+
+        // Start the CLI session by running a wslc command.
+        auto result = WSLCE2ETests::RunWslc(L"container list");
```
ptrivedi, line 293: i think this will by default create a persistent session.. but please double-check on this. in case this is a non-persistent session, it could exit and your check below will be inconsistent due to timing/race

src/windows/common/WslSecurity.cpp
```diff
@@ -187,6 +187,17 @@ bool wsl::windows::common::security::IsTokenLocalSystem(_In_opt_ HANDLE token)
...
 }
 
+bool wsl::windows::common::security::IsElevated()
+{
+    auto token = wil::open_current_access_token(TOKEN_QUERY);
+
+    // IsTokenElevated checks if the integrity level is exactly HIGH.
+    // We must also check for local system because it is above HIGH.
+    // However, IsTokenLocalSystem() does not work correctly and fails.
+    // TODO: Add proper handling for system user callers.
```
ptrivedi: I know this is pre-existing and not related to your PR @chemwolf6922, but is the fix in IsTokenElevated() as simple as changing the == to >= for proper handling of system user callers, @benhillis ?
Now that IsElevated is moved to the shared WslSecurity, this bug/limitation could have more exposure

src/windows/common/WslSecurity.cpp
```diff
@@ -187,6 +187,17 @@ bool wsl::windows::common::security::IsTokenLocalSystem(_In_opt_ HANDLE token)
...
+    return IsTokenElevated(token.get());
```
ptrivedi: Also, could the issue referenced above mean that a SYSTEM-context caller of WslcGetCliSession would get the wrong session name (wslc-cli instead of wslc-cli-admin), and so it could likely result in a 'not found'?

yao-msft:
Just curious about the context of the change, what're the scenarios that an app will need to access a cli session?

src/windows/common/WSLCSessionDefaults.h
chemwolf6922: I was following names like WSLCProcessLauncher and WSLCContianerLauncher. It seems that this is to distinguish from files like WslClient or WslCoreConfig.

chemwolf6922:
> Just curious about the context of the change, what're the scenarios that an app will need to access a cli session?

Hi @yao-msft , Please refer to the build experience design doc. I'll try to send you internally. In short, this is for the dev loop where the developer builds their image in the cli session and pick it up directly when debugging.

test/windows/WslcSdkTests.cpp
chemwolf6922: Confirmed, this will create a persistent session.

src/windows/common/WslSecurity.cpp
chemwolf6922: If we assume the user runs cli commands in the same context as the get cli session call. They will always get the same session. No matter if it's the correct one. (local system will get wslc-cli). 
For the dev loop scenario, this assumption is true.

src/windows/common/WslSecurity.cpp
chemwolf6922: I modifed the IsElevated function to use >=. And renamed it to IsElevatedOrAbove to aboid confusion with the IsTokenElevated fucntion. 
I don't think we should modify the IsTokenElevated function though. As that affects other parts of the codebase.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -990,7 +1000,19 @@ WslcInspectContainer WSLCContainerImpl::BuildInspectContainer(const DockerInspec
...
+        if (volume.SourceFilename.empty())
+        {
+            mountInfo.Source = wsl::shared::string::WideToMultiByte(volume.HostPath);
+        }
+        else
+        {
+            std::filesystem::path fullPath(volume.HostPath);
+            fullPath /= volume.SourceFilename;
+            mountInfo.Source = fullPath.string();
+        }
```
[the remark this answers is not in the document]
chemwolf6922: Wrong branch operation.

benhillis:

chemwolf6922:
deprecated.

PR #40144, opened 2026-04-09
wslc: prevent session name squatting for default WSLc sessions

benhillis:
## Summary

Prevents malicious users from squatting reserved WSLc session names (`wslc-cli`, `wslc-cli-admin`) by moving default session name and settings resolution to the server.

## Changes

- **Server-side default sessions**: `CreateSession(nullptr)` and `OpenSessionByName(nullptr)` resolve the default session name from the caller's token and username (e.g. `wslc-cli-alice`, `wslc-cli-admin-bob`). The client no longer knows or sends session names for default sessions.
- **Reserved name rejection**: Explicit use of reserved names via `CreateSession` returns `WSLC_E_SESSION_RESERVED`. Invalid session names return `WSLC_E_INVALID_SESSION_NAME`.
- **User impersonation**: Settings.yaml is loaded under the caller's identity via `wil::impersonate_token`.
- **Dedicated `EnterSession` API**: Custom sessions (`wslc session enter`) use a separate API that reads `settings.yaml` server-side and sets `NoCreate` storage flags.
- **`wslcsettings` shared library**: Extracted `UserSettings` into a shared library used by both `wslc.exe` and `wslservice.exe`. `SettingsCommand` now calls `UserSettings` directly instead of duplicating file creation logic.
- **Cleanup**: Removed `SessionOptions` class, `SessionModel.cpp`, `WslcSettingsTemplate.h`. Moved `EnumVariantMap.h` to common. `SessionSettings` is `NON_MOVABLE` (uses `unique_ptr` to avoid `c_str()` pointer invalidation).

## Testing

- E2E test verifies reserved names are rejected and default session creation still works.
- All E2E tests updated for username-qualified session names.


localization/strings/en-US/Resources.resw
```diff
@@ -1935,6 +1935,15 @@ Usage:
...
   </data>
+  <data name="MessageWslcDefaultSessionNotFound" xml:space="preserve">
+    <value>Default session not found</value>
+  </data>
+  <data name="MessageWslcOpenDefaultSessionFailed" xml:space="preserve">
+    <value>Failed to open default session</value>
+  </data>
+  <data name="MessageWslcTerminateDefaultSessionFailed" xml:space="preserve">
+    <value>Default session termination failed</value>
+  </data>
```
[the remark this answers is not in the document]
benhillis: That is because the other locale entries are auto-generated. This is bad feedback.

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -88,6 +167,14 @@ void WSLCSessionManagerImpl::CreateSession(const WSLCSessionSettings* Settings,
 
     wslutil::StopWatch stopWatch;
 
+    // Initialize settings for the default session.
+    std::unique_ptr<SessionSettings> defaultSettings;
+    if (Settings == nullptr)
+    {
+        defaultSettings = SessionSettings::Default(callerToken.get(), tokenInfo.Elevated);
+        Settings = &defaultSettings->Settings;
+    }
```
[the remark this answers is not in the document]
benhillis: I think this is preferred to dropping the lock and re-acquiring which could introduce a race.

src/windows/common/WSLCUserSettings.h
```diff
@@ -156,9 +156,7 @@ class UserSettings
     // Overwrites the settings file with the commented-out defaults template.
     void Reset() const;
 
-protected:
```
yao-msft, line 159: Is "protected" removed by accident?

test/windows/wslc/WSLCCLISettingsUnitTests.cpp
```diff
@@ -16,7 +16,7 @@ Module Name:
 #include "precomp.h"
 #include "windows/Common.h"
 #include "WSLCCLITestHelpers.h"
-#include "UserSettings.h"
+#include "WSLCUserSettings.h"
```
yao-msft, line 19: nit: since the settings code is moved, test could be moved out of wslc cli tests too

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -51,22 +112,42 @@ WSLCSessionManagerImpl::~WSLCSessionManagerImpl()
...
+        THROW_HR_IF(E_INVALIDARG, wcslen(Settings->DisplayName) >= std::size(WSLCSessionInformation{}.DisplayName));
+        THROW_HR_IF_MSG(
+            E_INVALIDARG,
+            WI_IsAnyFlagSet(Settings->StorageFlags, ~WSLCSessionStorageFlagsValid),
+            "Invalid storage flags: %i",
+            Settings->StorageFlags);
+
+        // Reserved names can only be assigned server-side via null Settings.
+        THROW_HR_IF(
+            E_ACCESSDENIED,
```
dkbennett: It would be nice if we could establish a pattern moving forward to create specific WSLC_E_<ERROR> codes for situations like this for easier diagnosability since E_INVALIDARG and E_ACCESSDENIED can be thrown by many things in the system. 

Recommend something like WSLC_E_SESSION_RESERVED , WSLC_E_INVALID_SESSION_STORAGE, and WSLC_E_INVALID_SESSION_NAME  (this one could also be used by the reserved names)

dkbennett (approved):
All the session changes as they relate to the CLI look good to me. Minor comments about having more specific error codes for better diagnosability. We can always add this later but would be easier to pay as we go.

src/windows/common/WSLCUserSettings.cpp
```diff
@@ -4,18 +4,23 @@ Copyright (c) Microsoft. All rights reserved.
...
-    UserSettings.cpp
+    WSLCUserSettings.cpp
 
 Abstract:
 
     Implementation of UserSettings — YAML loading and validation.
 
 --*/
-#include "UserSettings.h"
+#include "precomp.h"
```
OneBlue, line 14: This file has logic to display warnings if some settings are invalid when constructed. This won't work from the service unfortunately. 

We can probably leave as is for now and maybe look for a way to wire warnings through the EMIT_USER_WARNING macro in a future change though

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -29,12 +29,73 @@ Module Name:
...
+    WSLCSessionSettings Settings{};
+
+    NON_COPYABLE(SessionSettings);
+    NON_MOVABLE(SessionSettings);
+
+    // Default session: name and storage path determined from caller's token.
+    static std::unique_ptr<SessionSettings> Default(HANDLE UserToken, bool Elevated)
+    {
+        auto localAppData = wsl::windows::common::filesystem::GetLocalAppDataPath(UserToken);
+        settings::UserSettings userSettings(localAppData / L"wslc");
```
OneBlue: We should impersonate the user when we load the settings yaml (same for below)

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -51,22 +112,42 @@ WSLCSessionManagerImpl::~WSLCSessionManagerImpl()
...
-
     auto tokenInfo = GetCallingProcessTokenInfo();
+    const auto callerToken = wsl::windows::common::security::GetUserToken(TokenImpersonation);
+
+    // Resolve display name upfront (for both default and custom sessions).
+    std::wstring resolvedDisplayName;
+    if (Settings == nullptr)
+    {
+        // Default session: name determined from token.
+        resolvedDisplayName = tokenInfo.Elevated ? wsl::windows::wslc::DefaultAdminSessionName : wsl::windows::wslc::DefaultSessionName;
```
OneBlue: Since we're resolving the session name inside the service, I think we should append the username to the session name, that way two users can use wslc at the same time without their sessions colliding

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -160,6 +248,14 @@ void WSLCSessionManagerImpl::OpenSessionByName(LPCWSTR DisplayName, IWSLCSession
 {
     auto tokenInfo = GetCallingProcessTokenInfo();
 
+    // Null name = default session, resolved from caller's token.
+    std::wstring resolvedName;
+    if (DisplayName == nullptr)
+    {
+        resolvedName = tokenInfo.Elevated ? wsl::windows::wslc::DefaultAdminSessionName : wsl::windows::wslc::DefaultSessionName;
+        DisplayName = resolvedName.c_str();
```
OneBlue, line 261: nit: This might be a good idea to factor (especially if we append the username to the session name)

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -260,6 +372,40 @@ CallingProcessTokenInfo WSLCSessionManagerImpl::GetCallingProcessTokenInfo()
...
+    wchar_t username[256 + 1] = {};
+    DWORD usernameLen = ARRAYSIZE(username);
+    wchar_t domain[MAX_PATH] = {};
+    DWORD domainLen = ARRAYSIZE(domain);
+    SID_NAME_USE sidType;
+    THROW_IF_WIN32_BOOL_FALSE(LookupAccountSidW(nullptr, TokenInfo.TokenInfo->User.Sid, username, &usernameLen, domain, &domainLen, &sidType));
+
+    auto baseName = TokenInfo.Elevated ? wsl::windows::wslc::DefaultAdminSessionName : wsl::windows::wslc::DefaultSessionName;
+    return std::format(L"{}-{}", baseName, username);
+}
```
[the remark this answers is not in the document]
benhillis: can adjust this in a follow-up.

OneBlue (approved):

PR #40149, opened 2026-04-10
Fix incorrect --detach description in container run help text

benhillis:
The help text for `wslc container run` incorrectly stated that containers start in the background by default and `--detach` runs in the foreground. The actual behavior (confirmed by `ContainerService.cpp` line 299: `WI_SetFlagIf(startFlags, WSLCContainerStartFlagsAttach, !runOptions.Detach)`) is the opposite: containers start attached (foreground) by default, and `--detach` runs them in the background.

This fix updates the en-US locale string. Other locales will pick up the correction through the standard localization pipeline.

---

**Before:** *Runs a container. By default, the container is started in the background; use --detach to run in the foreground.*

**After:** *Runs a container. By default, the container is started in the foreground; use --detach to run in the background.*


AmelBawa-msft (approved):
LGTM. The help message in E2E likely need to be updated as well.

PR #40155, opened 2026-04-10
Add git entries for libarchive and boost to fix NOTICE generation

benhillis:
The OSPO `notice@0` task cannot resolve licenses for components registered as `type: "other"` in cgmanifest.json (source tarballs). This causes "Not a supported type" warnings and missing NOTICE entries for libarchive and boost.

Fix by adding parallel `type: "git"` entries pointing at the GitHub repos with the exact commit SHAs for the release tags:
- libarchive v3.7.7: `b439d586f53911c84be5e380445a8a259e19114c`
- boost 1.90.0: `1bed2b0712b2119f20d66c5053def9173c8462a5`

The existing `type: "other"` entries are kept for accurate provenance tracking of the actual tarball downloads used by CMake FetchContent.

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -13,12 +13,41 @@
         },
         {
             "component": {
+                "type": "git",
+                "git": {
+                    "repositoryUrl": "https://github.com/libarchive/libarchive",
+                    "commitHash": "b439d586f53911c84be5e380445a8a259e19114c"
+                }
+            }
+        },
+        {
+            "component": {
                 "type": "other",
                 "other": {
                     "name": "libarchive",
                     "version": "3.7.7",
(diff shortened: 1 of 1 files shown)


cgmanifest.json
```diff
@@ -22,6 +31,26 @@
                 }
             }
         },
+        {
+            "component": {
```
OneBlue, line 35: Should this stay on the feature branch for now ? We don't use boost in `master`

benhillis: I was just trying to avoid having to submit two PR's.

OneBlue: Ok that's probably OK for now then

OneBlue (approved):

PR #40158, opened 2026-04-10
Detect unexpected VM exit and terminate wslcsession.exe

benhillis:
Detect unexpected VM exits and terminate wslcsession.exe cleanly instead of hanging.

## Changes

- **GetTerminationEvent** (IDL): Added `GetTerminationEvent([out, system_handle(sh_event)] HANDLE* Event)` to `IWSLCVirtualMachine`. The service creates a manual-reset event (`m_vmExitEvent`) in the HcsVirtualMachine constructor and signals it when the HCS VM exits. `GetTerminationEvent` duplicates the handle to the caller so both sides share the same kernel event object.

- **OnVmExited handler**: The session registers `m_vmExitedEvent` with the IO relay during `Initialize()`. When the VM exits unexpectedly, `OnVmExited()` logs and calls `Terminate()`. The IO relay's `Stop()` method detects self-stop (called from its own thread) and skips the join to avoid deadlock.

- **Hardened Terminate()**: An `m_terminating` atomic guard prevents double-execution (`OnVmExited` can race with an explicit `Terminate()` call). When the VM is already dead, `Terminate()` skips `StopProcess`/`Unmount` calls that would hang waiting for a dead VM, but always resets process/VM resources.

## Tests

- **VmKillTerminatesSession**: Creates a session, kills its VM via `hcsdiag kill` (identified by owner name), verifies the session reaches the terminated state and the VM is gone.

- **VmKillFailsInFlightOperations**: Creates a session with a running `sleep` process, kills the VM, verifies both the session terminates and the in-flight process exits cleanly.

### DIFF
--- a/src/windows/service/exe/HcsVirtualMachine.cpp
+++ b/src/windows/service/exe/HcsVirtualMachine.cpp
@@ -581,6 +581,15 @@ try
 }
 CATCH_RETURN()
 
+HRESULT HcsVirtualMachine::GetTerminationEvent(_Out_ HANDLE* Event)
+try
+{
+    *Event = wslutil::DuplicateHandle(m_vmExitEvent.get());
+
+    return S_OK;
+}
+CATCH_RETURN()
+
 void CALLBACK HcsVirtualMachine::OnVmExitCallback(HCS_EVENT* Event, void* Context)
 try
 {
--- a/src/windows/wslcsession/IORelay.cpp
+++ b/src/windows/wslcsession/IORelay.cpp
@@ -61,7 +61,8 @@ void IORelay::Stop()
     m_exit = true;
     m_refreshEvent.SetEvent();
 
-    if (m_thread.joinable())
+    // Skip join if called from the IORelay thread itself (e.g., from a handle callback).
+    if (m_thread.joinable() && m_thread.get_id() != std::this_thread::get_id())
     {
         m_thread.join();
     }
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -2360,7 +2360,9 @@ try
 
     // Check if the VM has already exited (e.g., killed externally).
     // If so, skip operations that require a live VM to avoid unnecessary waits.
-    if (m_vmExitedEvent.is_signaled())
+    // N.B. m_vmExitedEvent may be uninitialized if Terminate() is called from the
+    // Initialize() error path before GetTerminationEvent() succeeds.
+    if (m_vmExitedEvent && m_vmExitedEvent.is_signaled())
     {
         WSL_LOG("SkippingGracefulShutdown_VmDead", TraceLoggingValue(m_id, "SessionId"));
     }
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -2302,8 +2302,10 @@ CATCH_RETURN();
 HRESULT WSLCSession::Terminate()
 try
 {
-    // Ensure only one Terminate() runs to completion. OnVmExited() can race
-    // with an external Terminate() call; the second caller is a no-op.
+    // Ensure only one Terminate() runs. This must be checked before taking m_lock
+    // because OnVmExited() is called from the IORelay thread — if an external Terminate()
+    // holds m_lock and calls m_ioRelay.Stop(), the relay thread must not re-enter
+    // Terminate() and deadlock on m_lock.
     if (m_terminating.exchange(true))
     {
         return S_OK;
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -240,6 +240,9 @@ try
 
     m_virtualMachine->Initialize();
 
+    // Get an event from the service that is signaled when the VM exits.
+    THROW_IF_FAILED(Vm->GetTerminationEvent(&m_vmExitedEvent));
+
     // Configure storage.
     ConfigureStorage(*Settings, tokenInfo->User.Sid);
 
@@ -259,6 +262,10 @@ try
 
     //  Start the event tracker.
     m_eventTracker.emplace(m_dockerClient.value(), m_id, m_ioRelay);
+
+    // Monitor for unexpected VM exit.
+    m_ioRelay.AddHandle(
+        std::make_unique<windows::common::relay::EventHandle>(m_vmExitedEvent.get(), std::bind(&WSLCSession::OnVmExited, this)));
 
     // Recover any existing containers from storage.
     RecoverExistingNetworks();
@@ -374,6 +381,18 @@ void WSLCSession::OnContainerdExited()
     {
         WSL_LOG("UnexpectedContainerdExit", TraceLoggingValue(m_displayName.c_str(), "Name"));
     }
+}
+
+void WSLCSession::OnVmExited()
+{
+    WSL_LOG(
+        "VmExited",
+        TraceLoggingLevel(WINEVENT_LEVEL_WARNING),
+        TraceLoggingValue(m_id, "SessionId"),
+        TraceLoggingValue(m_displayName.c_str(), "Name"),
+        TraceLoggingValue(!m_sessionTerminatingEvent.is_signaled(), "Unexpected"));
+
+    LOG_IF_FAILED(Terminate());
 }
 
 void WSLCSession::OnProcessLog(const gsl::span<char>& Buffer, PCSTR Source)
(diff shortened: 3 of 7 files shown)


test/windows/WSLCTests.cpp
```diff
@@ -6740,4 +6740,165 @@ class WSLCTests
...
+            if (id.size() >= 36)
+            {
+                ids.insert(id);
+            }
+        }
+
+        return ids;
+    }
+
+    // Helper: kill the VM backing a session using hcsdiag.
```
benhillis: I think the VM owner is displayed in the hcsdiag list output, maybe we can use that instead? Maybe we should make the owner WSLC-<session_name> or something? I'm not sure what the limits on the vm owner name are....

src/windows/service/exe/HcsVirtualMachine.cpp
```diff
@@ -581,6 +589,35 @@ try
 }
 CATCH_RETURN()
 
+HRESULT HcsVirtualMachine::RegisterTerminationEvent(_In_ HANDLE Event)
```
benhillis: Not a big fan of this signal delivery stuff, is there a better way to handle it?

test/windows/WSLCTests.cpp
```diff
@@ -438,28 +438,48 @@ class WSLCTests
...
+                    {wsl::shared::string::MultiByteToWide(entry["Id"].get<std::string>()),
+                     wsl::shared::string::MultiByteToWide(entry["Owner"].get<std::string>())});
             }
         }
 
+        return vms;
+    }
+
+    // Returns the set of VM owner names currently reported by hcsdiag.
+    static std::vector<std::wstring> ListVmOwners()
```
JohnMcPMS: Should this be an actual `std::set` so that owners doesn't have repeats?  Or is owner not able to repeat?  Find by owner suggests that at least for the tests you don't expect it to have any within the test context.

test/windows/WSLCTests.cpp
```diff
@@ -8135,4 +8155,90 @@ class WSLCTests
...
+    {
+        for (DWORD i = 0; i < timeoutSeconds; i++)
+        {
+            Sleep(1000);
+
+            WSLCSessionState state{};
+            auto hr = session->GetState(&state);
+
+            // RPC errors indicating the session process exited.
+            if (hr == HRESULT_FROM_WIN32(RPC_S_SERVER_UNAVAILABLE) || hr == HRESULT_FROM_WIN32(RPC_S_CALL_FAILED) ||
```
JohnMcPMS: Is the design that the session server terminates itself in response to the underlying session VM terminating? While the caller still has active objects?

test/windows/WSLCTests.cpp
```diff
@@ -8135,4 +8155,90 @@ class WSLCTests
...
+
+        VERIFY_IS_TRUE(WaitForSessionTermination(session.get()));
+    }
+
+    WSLC_TEST_METHOD(VmKillFailsInFlightOperations)
+    {
+        auto settings = GetDefaultSessionSettings(L"wslc-vm-kill-inflight-test");
+        auto session = CreateSession(settings);
+
+        WSLCProcessLauncher launcher("/bin/sleep", {"/bin/sleep", "99999"});
```
JohnMcPMS:
```suggestion
        WSLCProcessLauncher launcher("/bin/sleep", {"/bin/sleep", "99"});
```
Because 100K seconds seems like a long time to wait if the kill somehow fails.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -240,6 +240,9 @@ try
 
     m_virtualMachine->Initialize();
 
+    // Get an event from the service that is signaled when the VM exits.
+    THROW_IF_FAILED(Vm->GetTerminationEvent(&m_vmExitedEvent));
```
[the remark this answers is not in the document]
benhillis: wrong again

test/windows/WSLCTests.cpp
benhillis: good call, simplified this.

test/windows/WSLCTests.cpp
benhillis: Yes, when the VM exits unexpectedly the session is unrecoverable, so we self-terminate to clean up gracefully. In-flight operations fail with appropriate errors rather than hanging or returning confusing HCS failures. This follows the same Terminate() path as an explicit shutdown.

test/windows/WSLCTests.cpp
benhillis: good call, changed this to one minute

src/windows/wslcsession/IORelay.cpp
```diff
@@ -61,7 +61,8 @@ void IORelay::Stop()
     m_exit = true;
     m_refreshEvent.SetEvent();
 
-    if (m_thread.joinable())
+    // Skip join if called from the IORelay thread itself (e.g., from a handle callback).
+    if (m_thread.joinable() && m_thread.get_id() != std::this_thread::get_id())
     {
         m_thread.join();
     }
```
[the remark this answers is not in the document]
benhillis: Stop() only skips the join when called from its own thread. Object destruction happens on a different thread, so Stop() joins normally.

JohnMcPMS (approved):

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -376,6 +383,17 @@ void WSLCSession::OnContainerdExited()
     }
 }
 
+void WSLCSession::OnVmExited()
+{
+    WSL_LOG(
+        "UnexpectedVmExit",
+        TraceLoggingLevel(WINEVENT_LEVEL_WARNING),
+        TraceLoggingValue(m_id, "SessionId"),
+        TraceLoggingValue(m_displayName.c_str(), "Name"));
```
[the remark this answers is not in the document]
benhillis: Dismissing: the log already includes an `Unexpected` field to distinguish normal vs unexpected exits, so this is filterable in telemetry.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -376,6 +383,17 @@ void WSLCSession::OnContainerdExited()
...
+
+    LOG_IF_FAILED(Terminate());
+}
```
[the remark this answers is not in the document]
benhillis: Dismissing: this is by design. The IORelay::Stop() self-call guard handles this case, and spawning a separate thread adds unnecessary lifetime complexity.

src/windows/wslcsession/IORelay.cpp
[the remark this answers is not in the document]
benhillis: Dismissing: the shutdown ordering ensures the relay thread exits naturally before the IORelay is destroyed, so this is safe in practice.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2283,6 +2302,12 @@ CATCH_RETURN();
 HRESULT WSLCSession::Terminate()
 try
 {
+    // Ensure only one Terminate() runs to completion. OnVmExited() can race
+    // with an external Terminate() call; the second caller is a no-op.
+    if (m_terminating.exchange(true))
```
OneBlue, line 2309: Do we need this ? I'd expect `m_lock` to protect us from a race here

test/windows/WSLCTests.cpp
```diff
@@ -8135,4 +8142,69 @@ class WSLCTests
...
+    static void WaitForSessionTermination(IWSLCSession* session)
+    {
+        wsl::shared::retry::RetryWithTimeout<void>(
+            [&]() {
+                WSLCSessionState state{};
+                THROW_IF_FAILED(session->GetState(&state));
+                THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_RETRY), state != WSLCSessionStateTerminated);
+            },
+            std::chrono::seconds{1},
+            std::chrono::seconds{10});
```
OneBlue: I recommend putting a longer timeout (maybe 1 - 2 minutes) just so this doesn't fail if the test VM is slow

OneBlue (approved):
LGTM. Long term we'll probably want to add logic to terminate the VM if inactive for a while, and have way for wslcsession.exe to restart it on-demand, but this is a great improvement for now

src/windows/wslcsession/WSLCSession.cpp
benhillis: We need it because OnVmExited() is called from the IORelay thread. If an external Terminate() holds m_lock and calls m_ioRelay.Stop(), the relay thread must not re-enter Terminate() and deadlock on m_lock. The atomic exchange gives us a lock-free single-entry guard. Expanded the comment to explain this.

test/windows/WSLCTests.cpp
benhillis: Good call, bumped to 2 minutes.

benhillis:
> LGTM. Long term we'll probably want to add logic to terminate the VM if inactive for a while, and have way for wslcsession.exe to restart it on-demand, but this is a great improvement for now

yep, I was planning on adding that as a follow up to this

PR #40159, opened 2026-04-10
Fix VHD ownership after cross-volume move to prevent E_ACCESSDENIED

benhillis:
When MoveDistribution moves a VHD across volumes, MoveFileEx copies the file and the new file's owner may not be the user's SID. This causes HcsGrantVmAccess to fail with E_ACCESSDENIED when later launching the distro, because the impersonated user lacks WRITE_DAC on the file (only implicitly granted to the owner).

Fix by explicitly setting the VHD owner to the user's SID after the move, matching what CreateVhd already does at creation time. Uses handle-based SetSecurityInfo with FILE_FLAG_OPEN_REPARSE_POINT to avoid TOCTOU races and symlink following.

Also fixes a pre-existing build break in MountTests.cpp from the test refactor (WSL2_TEST_ONLY -> WSL2_TEST_METHOD).

### DIFF
--- a/src/windows/service/exe/LxssUserSession.cpp
+++ b/src/windows/service/exe/LxssUserSession.cpp
@@ -939,13 +939,20 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
         THROW_WIN32(error.value());
     }
 
+    // Read the original VHD owner before the move so we can restore it after.
+    // Cross-volume MoveFileEx may set the owner to BUILTIN\Administrators for
+    // elevated callers, which breaks HcsGrantVmAccess (needs WRITE_DAC via
+    // ownership) from non-elevated contexts.
+    PSID originalOwner = nullptr;
+    wil::unique_hlocal originalDescriptor;
+    THROW_IF_WIN32_ERROR(GetNamedSecurityInfoW(
+        distro.VhdFilePath.c_str(), SE_FILE_OBJECT, OWNER_SECURITY_INFORMATION, &originalOwner, nullptr, nullptr, nullptr, &originalDescriptor));
+
     // Move the VHD to the new location.
     THROW_IF_WIN32_BOOL_FALSE(MoveFileEx(distro.VhdFilePath.c_str(), newVhdPath.c_str(), MOVEFILE_COPY_ALLOWED | MOVEFILE_WRITE_THROUGH));
 
-    // Set the VHD owner to the user's SID. Cross-volume MoveFileEx sets the
-    // owner to BUILTIN\Administrators for elevated callers, which breaks
-    // HcsGrantVmAccess (needs WRITE_DAC via ownership) from non-elevated contexts.
-    auto setVhdOwner = [this](const std::filesystem::path& vhdPath) {
+    // Restore the original VHD owner on the moved file.
+    auto setVhdOwner = [&originalOwner](const std::filesystem::path& vhdPath) {
         wil::unique_hfile vhdHandle(CreateFileW(
             vhdPath.c_str(), WRITE_OWNER, FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, FILE_FLAG_OPEN_REPARSE_POINT, nullptr));
         THROW_LAST_ERROR_IF(!vhdHandle);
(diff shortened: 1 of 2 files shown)


test/windows/UnitTests.cpp
```diff
@@ -2801,6 +2801,80 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
...
+        constexpr auto moveNonElevatedFolder = L"move-owner-non-elevated";
+
+        // Import a WSL2 distro.
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(std::format(L"--import {} . \"{}\" --version 2", name, g_testDistroPath)), 0L);
+
+        auto cleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [name]() {
+            LxsstuLaunchWsl(std::format(L"--unregister {}", name));
+            std::filesystem::remove_all(moveElevatedFolder);
+            std::filesystem::remove_all(moveNonElevatedFolder);
+        });
```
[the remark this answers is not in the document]
benhillis: this is standard practice for the tests

benhillis:
Re: [comment about moving run_as_self() before CreateFileW](https://github.com/microsoft/WSL/pull/40159#discussion_r3066258139)

This suggestion is **incorrect and would introduce a security vulnerability**. The `CreateFileW` is intentionally called while still impersonating the user so that the access check (`WRITE_OWNER`) is performed against the caller's token. This ensures we can only set ownership on files the user already has access to.

If we moved `run_as_self()` before `CreateFileW` as suggested, the file would be opened as SYSTEM -- meaning a malicious caller could potentially direct us to open *any* file on the system and take ownership of it. The impersonated `CreateFileW` is the security gate here.

The user will always have `WRITE_OWNER` on this file because they just created it via `MoveFileEx` on line 943.

test/windows/UnitTests.cpp
```diff
@@ -2801,6 +2801,80 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
         }
     }
 
+    TEST_METHOD(MoveVhdOwnership)
+    {
+        constexpr auto name = L"move-owner-test-distro";
+        constexpr auto moveElevatedFolder = L"move-owner-elevated";
+        constexpr auto moveNonElevatedFolder = L"move-owner-non-elevated";
```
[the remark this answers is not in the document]
benhillis: standard practice for tests

src/windows/service/exe/LxssUserSession.cpp
```diff
@@ -942,10 +942,29 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
...
+    // Set the VHD owner to the user's SID. Cross-volume MoveFileEx sets the
+    // owner to BUILTIN\Administrators for elevated callers, which breaks
+    // HcsGrantVmAccess (needs WRITE_DAC via ownership) from non-elevated contexts.
+    auto setVhdOwner = [this](const std::filesystem::path& vhdPath) {
+        wil::unique_hfile vhdHandle(CreateFileW(
+            vhdPath.c_str(), WRITE_OWNER, FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, FILE_FLAG_OPEN_REPARSE_POINT, nullptr));
+        THROW_LAST_ERROR_IF(!vhdHandle);
+
+        auto runAsSelf = wil::run_as_self();
+        auto privileges = wsl::windows::common::security::AcquirePrivilege(SE_RESTORE_NAME);
```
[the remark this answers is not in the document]
benhillis: VERY VERY BAD WORK COPILOT.

src/windows/service/exe/LxssUserSession.cpp
```diff
@@ -942,10 +942,29 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
...
+        THROW_IF_WIN32_ERROR(
+            ::SetSecurityInfo(vhdHandle.get(), SE_FILE_OBJECT, OWNER_SECURITY_INFORMATION, GetUserSid(), nullptr, nullptr, nullptr));
```
OneBlue: This works, but I'm a bit worried that this could be used to change the owner of a file that the caller doesn't actually own. 

Instead of using GetUserSid(), could we look up the owner of the original VHD, and apply it to the newly created VHD ? At least that way the owner will be the same either way

benhillis: Yeah that's probably possible - but the user would have had permission to delete (move) the file before, so I think this isn't an issue.

test/windows/UnitTests.cpp
```diff
@@ -2801,6 +2801,80 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
         }
     }
 
+    TEST_METHOD(MoveVhdOwnership)
```
OneBlue, line 2804: nit: This should be `WSL2_TEST_METHOD` since test only applies to wsl2

OneBlue (approved):

test/windows/UnitTests.cpp
benhillis: Will fix in follow up.

PR #40162, opened 2026-04-10
Fix mount test build in main

benhillis:
## Summary

`AbsolutePathVhdUnmountAfterVMTimeout` was using the old `TEST_METHOD` + `WSL2_TEST_ONLY()` pattern which broke the build after the TAEF metadata refactor in #40140.

## Changes

- Replace `TEST_METHOD` + `WSL2_TEST_ONLY()` with `WSL2_TEST_METHOD` macro in `MountTests.cpp`, consistent with all other WSL2-only tests.

## Validation

- Clean build succeeds (x64 Debug)
- No other test methods use the old pattern


OneBlue (approved):
Thank you, I was trying to figure out why was my other PR on master was hitting this

PR #40170, opened 2026-04-13
Add admin protection error message for shadow admin scenarios

benhillis:
When Windows Admin Protection is enabled, the elevated process runs as a shadow admin with a different SID, so distributions registered under the real user are not visible.

This adds an informational message explaining Admin Protection in two cases:

1. **Launching a distribution by name** that is not found (WSL_E_DISTRO_NOT_FOUND)
2. **Listing distributions** when none are registered (WSL_E_DEFAULT_DISTRO_NOT_FOUND)

The message links to https://aka.ms/apdevguide for more details.

### DIFF
--- a/src/windows/common/wslutil.cpp
+++ b/src/windows/common/wslutil.cpp
@@ -522,8 +522,7 @@ wsl::windows::common::wslutil::GetDefaultVersion(void)
     return version;
 }
 
-namespace
-{
+namespace {
 
 // Returns true if the current process is running as a shadow admin under
 // Windows Admin Protection. Caches the DLL lookup on first call.
@@ -539,8 +538,7 @@ bool IsAdminProtectionEnabled()
     static std::optional<LxssDynamicFunction<ShadowAdminEnabledFn>> s_fn;
     static std::once_flag s_initFlag;
 
-    std::call_once(s_initFlag, []()
-    {
+    std::call_once(s_initFlag, []() {
         LxssDynamicFunction<ShadowAdminEnabledFn> fn{DynamicFunctionErrorLogs::None};
         if (SUCCEEDED(fn.load(L"SecurityHealthUdk.dll", "Shield_LUAIsShadowAdminEnabled")))
         {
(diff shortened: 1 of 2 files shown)


src/windows/common/WslSecurity.cpp
```diff
@@ -135,6 +135,24 @@ bool wsl::windows::common::security::IsTokenElevated(_In_ HANDLE token)
...
+    }
+
+    using ShadowAdminEnabledFn = BOOL(WINAPI)();
+    LxssDynamicFunction<ShadowAdminEnabledFn> isShadowAdminEnabled{DynamicFunctionErrorLogs::None};
+    if (FAILED(isShadowAdminEnabled.load(L"SecurityHealthUdk.dll", "Shield_LUAIsShadowAdminEnabled")))
+    {
+        return false;
+    }
+
+    return isShadowAdminEnabled();
```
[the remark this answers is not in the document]
benhillis: I don't think this is worth it.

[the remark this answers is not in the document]
benhillis: well, maybe we can use a std::once actually.

OneBlue (dismissed):

src/windows/common/wslutil.cpp
```diff
@@ -522,10 +522,40 @@ wsl::windows::common::wslutil::GetDefaultVersion(void)
...
 }
 
+namespace {
+
+// Returns true if the current process is running as a shadow admin under
+// Windows Admin Protection. Caches the DLL lookup on first call.
+bool IsAdminProtectionEnabled()
+{
+    const auto token = wil::open_current_access_token();
+    if (!wsl::windows::common::security::IsTokenElevated(token.get()))
```
benhillis: @OneBlue - Perhaps we should return this error in both cases? (even non-admin)

OneBlue: This would only apply in the case where the user is elevated right ? If the other was not found and the user was non-elevated, then we know for sure that it's a user error, so probably best not to display this imo

benhillis: I mean if tha shadow admin user has distros but the non admin user does not.

OneBlue: Oh I see. Yeah that's a good point.

OneBlue (approved):

Issue #10688, opened 2023-10-27
Blocking I/O behavior in wslrelay.exe/localhost(/init) hangs with multiplexed protocols on forwarded connections


n1hility:
## Problem Description

The implementation of localhost relaying on the Linux portion is using blocking I/O in a single thread to copy both sides of a connection. While this will work with a half-duplex request-reply style system (e.g. http/1.1), it will lead to connection hangs when there is a multiplexed protocol and bidirectional traffic (e.g. ssh, http/2, etc). 

As an example using the reproducer linked later in this report, the following strace output shows a thread in the localhost forwarding process (init ran under the name localhost in the host CBL OS) hung while writing to its destination VSOCK connection (FD=6). This blocks forever since the other end of the connection (wslrelay.exe + the client win application) has a full read buffer, and the protocol can not advance until the WSL init process reads from the VSOCK fd (opening up more buffer space between the two ends).

```
[pid   778] ppoll([{fd=6, events=POLLIN}, {fd=9, events=POLLIN}], 2, NULL, NULL, 8) = 2 ([{fd=6, revents=POLLIN}, {fd=9, revents=POLLIN}])
[pid   778] read(6, "table language.\n\nThe most import"..., 131072) = 16384
[pid   778] write(9, "table language.\n\nThe most import"..., 16384) = 16384
[pid   778] read(9, "suitable language.\n\nThe most imp"..., 131072) = 131072
[pid   778] write(6, "suitable language.\n\nThe most imp"..., 131072 <unfinished ...>    <----- Hang
```

While the Windows side (wslrelay.exe) is using Overlapped I/O, it simulates a similar write-blocking pattern. Relay.cpp polls reads on both the VSOCK and TCP win app connection using WaitForMultipleObjects; however, the writing portion on both ends calls an Overlapped WriteFile immediately followed by a WaitForMultipleObjects on that descriptor, blocking until the write fully completes.

## Suggested Fix

Oh the Linux side, I would recommend switching the socket to non-blocking, and introducing a buffer for each direction of the connection, which is managed by a simple state machine. The limit for each direction would need to be reduced by the leftover amount of an incomplete write (essentially pre-filling the read buffer) to establish back-pressure to and allow SOCK_STREAM flow-control to throttle an abusive peer, as well as limit resource consumption.

[...]

PR #40171, opened 2026-04-13
Localhost Relay Supports Full Duplex in NAT Networking

wangxin12:
This PR improves the localhost relay performance and reliability on both Linux and Windows, and migrates the Windows `BidirectionalRelay` to reuse the existing `MultiHandleWait` + `RelayHandle` infrastructure.

### Linux
- Non-blocking poll-based relay: Extracted a generic `UtilRelay` helper (`util.cpp`) that uses `poll()` with non-blocking fds, per-direction buffering, back-pressure via available buffer space.
- Localhost relay (`localhost.cpp`) now calls `UtilRelay` instead of the old blocking relay, with caller-specified buffer sizes.
### Windows
- Add `HalfCloseRelayHandle` for socket specifically, supports half-close (`shutdown(SD_SEND)`) on EOF/error.
- Rewrite `BidirectionalRelay` with `MultiHandleWait` + two `RelayHandle<>` instances, for consistency with other relay callers.
- Buffer size parameter: Added `BufferSize` through `ReadHandle` → `RelayHandle` → `BidirectionalRelay` to preserve caller-specified buffer sizes (128KB for localhost relay vs 4KB default).

### Testing

- Verified with full duplex TCP stress test (50,000 x 32KB frames each direction)
- TCP half-close works end-to-end: one direction can finish while the other continues
- No sequence errors, no data corruption
image
image  
  <br>
  <br>
Without the change, the test tool hangs.  
image

## PR Checklist

- [x] **Closes:** Link to issue #10688 


src/linux/init/localhost.cpp
```diff
@@ -103,40 +143,145 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
+                    THROW_LAST_ERROR_IF(flags < 0);
+                    THROW_LAST_ERROR_IF(fcntl(fd, F_SETFL, flags | O_NONBLOCK) < 0);
+                }
+
+                const auto bufferSize = message->BufferSize;
+                buffer.resize(bufferSize);
+                RelayDirection dirs[2] = {
+                    {relaySocket.get(), tcpSocket.get(), std::move(buffer), 0, 0, false, false},
+                    {tcpSocket.get(), relaySocket.get(), std::vector<gsl::byte>(bufferSize), 0, 0, false, false},
+                };
```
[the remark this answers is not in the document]
wangxin12: `message->BufferSize` is from `wslrelay.exe` Windows host process and it set to 128 KB now, seems to be an acceptable overhead.

src/linux/init/localhost.cpp
```diff
@@ -103,40 +143,145 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
+                            if (!(pfds[j].revents & POLLIN))
+                            {
+                                // No data to read; if the source is gone, mark EOF.
+                                if (pfds[j].revents & (POLLERR | POLLHUP))
+                                {
+                                    d.srcEof = true;
+                                }
+
+                                continue;
+                            }
```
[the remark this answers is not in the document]
wangxin12: Both fds are owned by `unique_fd` and are never closed during the relay loop, so POLLNVAL cannot occur.

src/linux/init/localhost.cpp
```diff
@@ -103,40 +143,145 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
+                                d.done = true;
+                                continue;
                             }
-                            else if (bytesRead < 0)
+
+                            if (!(pfds[j].revents & POLLOUT))
                             {
-                                return;
+                                continue;
+                            }
```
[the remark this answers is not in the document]
wangxin12: Both fds are owned by `unique_fd` and are never closed during the relay loop, so POLLNVAL cannot occur.

guanwenbin123:
“Thanks for the notification. Update received.”



获取 Outlook for iOS<https://aka.ms/o0ukef>
________________________________
发件人: wangxin12 ***@***.***>
发送时间: Tuesday, April 14, 2026 5:43:42 AM
收件人: microsoft/WSL ***@***.***>
抄送: Subscribed ***@***.***>
主题: Re: [microsoft/WSL] Localhost Relay: non-blocking IO on Linux side, IOCP with fully async reads and writes on Windows side (PR #40171)


@wangxin12 commented on this pull request.

________________________________

In src/linux/init/localhost.cpp<https://github.com/microsoft/WSL/pull/40171#discussion_r3075982459>:

> +                        if (pfds[j].events & POLLOUT)
                         {
-                            bytesRead = UtilReadBuffer(pollDescriptors[Index].fd, buffer);
-                            if (bytesRead == 0)
+                            // can't write to dstFd any more
+                            if (pfds[j].revents & (POLLERR | POLLHUP))
                             {
-                                pollDescriptors[Index].fd = -1;
-                                shutdown(outFd[Index], SHUT_WR);
+                                d.done = true;
+                                continue;
                             }
-                            else if (bytesRead < 0)
+
+                            if (!(pfds[j].revents & POLLOUT))
                             {
-                                return;
+                                continue;
+                            }


Both fds are owned by unique_fd and are never closed during the relay loop, so POLLNVAL cannot occur.

―

src/windows/common/relay.cpp
```diff
@@ -38,6 +38,134 @@ LARGE_INTEGER InitializeFileOffset(HANDLE File)
...
+    }
+};
+
+void TryIssueRead(RelayDirection& d)
+{
+    if (d.ReadPending || d.SrcEof || d.Done || d.Available() == 0)
+    {
+        return;
+    }
+
```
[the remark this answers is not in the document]
wangxin12: This is a deliberate trade-off. On Windows, the OS holds a direct pointer into the buffer during async writes, so memmove-style compaction is unsafe while `WritePending` is true. A ring buffer would add significant complexity and potential for bugs with concurrent async operations.  
The impact should be limited, Windows overlapped writes don't short-write, the fill-then-drain stall is bounded by a single write completion — acceptable overhead for correctness and simplicity. Even with the fill-then-drain behavior, the new implementation is a significant improvement over the original code, which used blocking `InterruptableWrite` that completely stalled the opposite direction during every write.  
The suggested change won't help in practice.

src/linux/init/localhost.cpp
```diff
@@ -103,40 +143,145 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
+                                // No data to read; if the source is gone, mark EOF.
+                                if (pfds[j].revents & (POLLERR | POLLHUP))
+                                {
+                                    d.srcEof = true;
+                                }
+
+                                continue;
+                            }
+
+                            d.Compact();
```
chemwolf6922: Could simplify the logic to wait for the buffer to be empty before accepting any reads. This shouldn't be a huge performance penalty as there is already a read buffer in the kernel.

src/linux/init/localhost.cpp
```diff
@@ -103,40 +143,145 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
+                                }
+
+                                d.srcEof = true;
+                                continue;
                             }
-                            else if (UtilWriteBuffer(outFd[Index], buffer.data(), bytesRead) < 0)
+                            else
                             {
-                                return;
+                                d.tail += nread;
```
chemwolf6922: Should do a write here. There is no need to set the write handler most of the time.

src/linux/init/localhost.cpp
```diff
@@ -103,40 +143,145 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
+                for (int fd : {tcpSocket.get(), relaySocket.get()})
+                {
+                    int flags = fcntl(fd, F_GETFL, 0);
+                    THROW_LAST_ERROR_IF(flags < 0);
+                    THROW_LAST_ERROR_IF(fcntl(fd, F_SETFL, flags | O_NONBLOCK) < 0);
+                }
+
+                const auto bufferSize = message->BufferSize;
+                buffer.resize(bufferSize);
+                RelayDirection dirs[2] = {
```
chemwolf6922: nit: Consider use directions and direction instead of dir and d?

src/linux/init/localhost.cpp
```diff
@@ -103,40 +143,145 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
                 {
-                    if ((pollDescriptors[0].fd == -1) || (pollDescriptors[1].fd == -1))
+                    // Complete directions where the source hit EOF and all
+                    // pending data has been flushed to the destination.
+                    for (auto& d : dirs)
+                    {
+                        if (!d.done && d.srcEof && d.Pending() == 0)
+                        {
+                            shutdown(d.dstFd, SHUT_WR);
+                            d.done = true;
```
chemwolf6922: Instead of labeling this as done. Consider remove this direction from the array to release resources.

src/linux/init/localhost.cpp
```diff
@@ -103,40 +143,145 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
+
+                            auto written = TEMP_FAILURE_RETRY(write(d.dstFd, d.buf.data() + d.head, d.Pending()));
+                            if (written < 0)
+                            {
+                                if (errno == EAGAIN || errno == EWOULDBLOCK)
+                                {
+                                    continue;
+                                }
+
+                                d.done = true;
```
chemwolf6922: Should half close the read end?

src/linux/init/localhost.cpp
```diff
@@ -103,40 +143,145 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
...
                 for (;;)
                 {
-                    if ((pollDescriptors[0].fd == -1) || (pollDescriptors[1].fd == -1))
+                    // Complete directions where the source hit EOF and all
+                    // pending data has been flushed to the destination.
+                    for (auto& d : dirs)
+                    {
+                        if (!d.done && d.srcEof && d.Pending() == 0)
+                        {
+                            shutdown(d.dstFd, SHUT_WR);
```
chemwolf6922: Could put half close of both the read and write end in the destructor. Then there is no need to do this complex state tracking.

chemwolf6922:
@guanwenbin123 Hi, could you please configure your auto reply so it does not create a message for every wsl update? Thanks.

src/windows/common/relay.cpp
```diff
@@ -257,128 +385,167 @@ wsl::windows::common::relay::InterruptableWrite(
...
-    LARGE_INTEGER leftOffset{};
-
-    std::vector<gsl::byte> rightBuffer(BufferSize);
-    const auto rightReadSpan = gsl::make_span(rightBuffer);
-    OVERLAPPED rightOverlapped = {0};
-    const wil::unique_event rightOverlappedEvent(wil::EventOptions::None);
-    rightOverlapped.hEvent = rightOverlappedEvent.get();
-    LARGE_INTEGER rightOffset{};
+    // Create a completion port and associate both handles.
+    wil::unique_handle iocp(CreateIoCompletionPort(INVALID_HANDLE_VALUE, nullptr, 0, 1));
```
OneBlue: On the windows I think it would be nice to reuse [MultHandleWait](https://github.com/microsoft/WSL/blob/3168f4c7a9f198db0e001b61cf3c5f2e57eb080b/src/windows/common/relay.hpp#L498-L499) to do all of the IO scheduling (especially once https://github.com/microsoft/WSL/pull/40042/changes is merged) 

We could either cherry-pick the changes from features/wsl-for-apps, or we could wait until that feature branch is merged into master, which should happen in the coming month

src/linux/init/localhost.cpp
```diff
@@ -103,40 +143,145 @@ void ListenThread(sockaddr_vm hvSocketAddress, int listenSocket)
                     return;
                 }
 
-                // Resize the buffer to be the requested size.
-                buffer.resize(message->BufferSize);
+                // Switch both sockets to non-blocking for the relay loop.
+                for (int fd : {tcpSocket.get(), relaySocket.get()})
```
OneBlue: Similarly to what we do in `feature/wsl-for-apps`, I think it would be a good idea to have a generic helper class for relays & polls. We do this in multiple places and having one "proper" implementation that we can reuse everywhere and specialize as needed

src/linux/init/localhost.cpp
wangxin12: Done. Each direction is now in one of two states: reading (buffer empty) or writing (buffer has data). removed Available(), Compact(), and reduced max poll entries from 4 to 2

src/linux/init/localhost.cpp
wangxin12: done

src/linux/init/localhost.cpp
wangxin12: done

src/linux/init/localhost.cpp
wangxin12: Keeping explicit shutdown() at the point where we know is clearer than putting it in destructor. It's one line in one place and avoids adding RAII complexity to RelayDirection

src/linux/init/localhost.cpp
wangxin12: since we now only poll for POLLIN or POLLOUT per direction (never both simultaneously), there's no in-flight read to cancel.

src/linux/init/localhost.cpp
wangxin12: Extracted `UtilRelay()` into `util.h/util.cpp` as a generic non-blocking relay between two file descriptors. Callers specify which fds are sockets via flags so the helper can do `shutdown(SHUT_WR)` on source EOF or error. Bidirectional by default, with a unidirectional option.

src/windows/common/relay.cpp
wangxin12: Good to know. Let's wait for [https://github.com/microsoft/WSL/pull/40042/changes](url) to be merged, then I can refactor changes on Windows side

src/linux/init/util.cpp
```diff
@@ -3427,4 +3428,202 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
...
+
+--*/
+
+{
+    // Set both fds to non-blocking for the relay loop.
+    for (int fd : {fd1, fd2})
+    {
+        int fl = fcntl(fd, F_GETFL, 0);
+        THROW_LAST_ERROR_IF(fl < 0);
+        THROW_LAST_ERROR_IF(fcntl(fd, F_SETFL, fl | O_NONBLOCK) < 0);
```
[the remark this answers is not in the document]
wangxin12: When `UtilRelay` returns, the relay is done — the fds have reached EOF or error.  Restoring blocking mode on dead fds is pointless overhead.

src/windows/common/relay.cpp
wangxin12: Hi @OneBlue @benhillis , I cherry pick [Ben's commit](https://github.com/microsoft/WSL/commit/357a29f8157673b1d365b486a13e03e3c7776022#diff-5de36acd90308dc62abf7855a686ee7052ffb6e762c756fd735fb0c9fbd9595d) and make change on top of it ([the PR](https://github.com/microsoft/WSL/pull/40042) seems to be still in working and have not updated for several weeks). Could you help review this when you get a chance? 
I'd like to get this fix into the next 2.7.x release train if possible. Let me know your thought on this. Thanks!

benhillis:

wangxin12:
Hi @OneBlue, I updated this PR with the latest `MultiHandleWait` + `RelayHandle`,  could you please take another look?

PR #40173, opened 2026-04-13
Add `wslc login`, `wslc logout`, and `wslc push` commands for registry authentication

kvega005:

Adds `wslc login`, `wslc logout`, and `wslc image push` commands, enabling authenticated container registry workflows in wslc.

## PR Checklist

### New Commands
- **`wslc login [-u <user>] [-p <pass>] [--password-stdin] [<server>]`** — Authenticates with a container registry and stores credentials for subsequent push/pull operations. Supports interactive prompts, inline flags, and piped password input. Defaults to Docker Hub when no server is specified.
- **`wslc logout [<server>]`** — Removes stored credentials for a registry.
- **`wslc image push <image>`** — Pushes a local image to a registry, automatically using stored credentials when available.
- **`wslc registry`** — Parent command group for login/logout. All three commands are also available as root-level aliases (`wslc login`, `wslc logout`, `wslc push`).

### Credential Storage (`RegistryService`)
- Two backends, configurable via `credentialStore` in wslc user settings:
  - **WinCred** (default): Uses Windows Credential Manager with a `wslc-credential/` target name prefix to isolate wslc entries from other applications.
  - **File**: DPAPI-encrypted JSON file at `%LocalAppData%\wslc\registry-credentials.json` with retry logic for concurrent access (sharing violation handling).
- Stored credentials are automatically passed to `PullImage` and `PushImage` API calls based on the image's registry domain.

### Auth Integration
- `ImageService::Pull` and `ImageService::Push` extract the registry domain from the image reference and look up stored credentials before calling the session API.
- `Authenticate` delegates to the session's Docker engine and supports both identity token and username/password auth header formats.

## Validation Steps Performed

- E2E tests against a local htpasswd-authenticated registry container:
  - Push/pull succeed after login, fail before login and after logout.
  - Interactive login, `-u`/`-p` flag login, and `--password-stdin` login all verified.
  - Mutual exclusivity and missing-username validation tested.
- Push/pull against an unauthenticated local registry verified.
- Help output tests for all new commands and root aliases.
- Negative path tests: push/pull of non-existent images, double logout.


src/windows/wslc/tasks/RegistryTasks.h
```diff
@@ -0,0 +1,22 @@
...
+
+--*/
+#pragma once
+#include "CLIExecutionContext.h"
+
+using wsl::windows::wslc::execution::CLIExecutionContext;
+
+namespace wsl::windows::wslc::task {
+void Login(CLIExecutionContext& context);
+void Logout(CLIExecutionContext& context);
```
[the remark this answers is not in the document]
kvega005: This is the same pattern used in `ImageTasks.h`. I think this is fine since these headers are only used inside the CLI code.

src/windows/wslc/services/ImageService.cpp
```diff
@@ -223,8 +248,13 @@ InspectImage ImageService::Inspect(wsl::windows::wslc::models::Session& session,
...
 }
 
-void ImageService::Push()
+void ImageService::Push(wsl::windows::wslc::models::Session& session, const std::string& image, IProgressCallback* callback)
 {
+    auto server = GetServerFromImage(image);
+    auto storedAuth = RegistryService::Get(server);
+    auto auth = storedAuth.value_or(BuildRegistryAuthHeader("", "", server));
+
+    THROW_IF_FAILED(session.Get()->PushImage(image.c_str(), auth.c_str(), callback));
```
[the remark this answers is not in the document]
kvega005: This is how the docker CLI handles this because the docker daemon requires X-Registry-Auth for push.

src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -63,13 +63,16 @@ _(Name,           "name",                NO_ALIAS,          Kind::Value,       L
 _(NoPrune,        "no-prune",            NO_ALIAS,          Kind::Flag,        Localization::WSLCCLI_NoPruneArgDescription()) \
 _(NoTrunc,        "no-trunc",            NO_ALIAS,          Kind::Flag,        Localization::WSLCCLI_NoTruncArgDescription()) \
 _(Output,         "output",              L"o",              Kind::Value,       Localization::WSLCCLI_OutputArgDescription()) \
+_(Password,       "password",            L"p",              Kind::Value,       Localization::WSLCCLI_LoginPasswordArgDescription()) \
```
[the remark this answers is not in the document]
kvega005: That's fine. This is copying docker args.

src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -81,6 +84,7 @@ _(Time,           "time",                L"t",              Kind::Value,       L
 _(TMPFS,          "tmpfs",               NO_ALIAS,          Kind::Value,       Localization::WSLCCLI_TMPFSArgDescription()) \
 _(TTY,            "tty",                 L"t",              Kind::Flag,        Localization::WSLCCLI_TTYArgDescription()) \
 _(User,           "user",                L"u",              Kind::Value,       Localization::WSLCCLI_UserArgDescription()) \
+_(Username,       "username",            L"u",              Kind::Value,       Localization::WSLCCLI_LoginUsernameArgDescription()) \
```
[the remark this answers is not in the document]
kvega005: This is just copying docker args.

src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -19,7 +19,7 @@ Module Name:
 #include "ImageModel.h"
 #include "ImageService.h"
 #include "ImageTasks.h"
-#include "PullImageCallback.h"
+#include "ImageProgressCallback.h"
```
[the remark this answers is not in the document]
kvega005: PullImageCallback was removed

localization/strings/en-US/Resources.resw
```diff
@@ -2293,6 +2299,57 @@ For privacy information about this product please visit https://aka.ms/privacy.<
...
+    <value>Log in to a registry.</value>
+  </data>
+  <data name="WSLCCLI_LoginLongDesc" xml:space="preserve">
+    <value>Log in to a registry. If no server is specified, the default is defined by the session.</value>
+  </data>
+  <data name="WSLCCLI_LogoutDesc" xml:space="preserve">
+    <value>Log out from a registry.</value>
+  </data>
+  <data name="WSLCCLI_LogoutLongDesc" xml:space="preserve">
+    <value>Log out from a registry. If no server is specified, the default is defined by the session.</value>
```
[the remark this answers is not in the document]
kvega005: This mirorrs docker's help message. I think in the future we might want to actually query the session for default server just like the docker CLI does.

src/windows/wslc/commands/RegistryCommand.cpp
```diff
@@ -0,0 +1,182 @@
...
+using namespace wsl::shared;
+
+namespace {
+
+auto MaskInput()
+{
+    HANDLE input = GetStdHandle(STD_INPUT_HANDLE);
+    DWORD mode = 0;
+
+    if ((input != INVALID_HANDLE_VALUE) && GetConsoleMode(input, &mode))
```
OneBlue, line 33: nit: I recommend using `wsl::windows::common::wslutil::IsConsoleHandle()` here

src/windows/wslc/commands/RegistryCommand.cpp
```diff
@@ -0,0 +1,182 @@
...
+    {
+        SetConsoleMode(input, mode & ~ENABLE_ECHO_INPUT);
```
OneBlue: nit: We should fail if SetConsoleMode fails here

src/windows/wslc/commands/RegistryCommand.cpp
```diff
@@ -0,0 +1,182 @@
...
+        throw CommandException(Localization::WSLCCLI_LoginPasswordStdinRequiresUsername());
+    }
+}
+
+void RegistryLoginCommand::ExecuteInternal(CLIExecutionContext& context) const
+{
+    // Prompt for username if not provided.
+    if (!context.Args.Contains(ArgType::Username))
+    {
+        context.Args.Add(ArgType::Username, Prompt(L"Username: ", false));
```
OneBlue: nit: `Username` should be localized

src/windows/wslc/services/RegistryService.cpp
```diff
@@ -0,0 +1,365 @@
...
+    THROW_IF_WIN32_BOOL_FALSE(ReadFile(handle.get(), buffer.data(), static_cast<DWORD>(buffer.size()), &bytesRead, nullptr));
+    buffer.resize(bytesRead);
+
+    try
+    {
+        return nlohmann::json::parse(buffer);
+    }
+    catch (...)
+    {
+        return nlohmann::json::object();
```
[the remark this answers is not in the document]
kvega005: If the file is corrupted, we probably want to override or reset it anyway. There shouldn't be partial writes because we hold exclusive access to the file when we write.

src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -64,13 +64,16 @@ _(NoCache,        "no-cache",            NO_ALIAS,          Kind::Flag,        L
 _(NoPrune,        "no-prune",            NO_ALIAS,          Kind::Flag,        Localization::WSLCCLI_NoPruneArgDescription()) \
 _(NoTrunc,        "no-trunc",            NO_ALIAS,          Kind::Flag,        Localization::WSLCCLI_NoTruncArgDescription()) \
 _(Output,         "output",              L"o",              Kind::Value,       Localization::WSLCCLI_OutputArgDescription()) \
+_(Password,       "password",            L"p",              Kind::Value,       Localization::WSLCCLI_LoginPasswordArgDescription()) \
+_(PasswordStdin,  "password-stdin",      NO_ALIAS,          Kind::Flag,        Localization::WSLCCLI_LoginPasswordStdinArgDescription()) \
 _(Path,           "path",                NO_ALIAS,          Kind::Positional,  Localization::WSLCCLI_PathArgDescription()) \
 /*_(Progress,       "progress",            NO_ALIAS,          Kind::Value,       Localization::WSLCCLI_ProgressArgDescription())*/ \
 _(Publish,        "publish",             L"p",              Kind::Value,       Localization::WSLCCLI_PublishArgDescription()) \
```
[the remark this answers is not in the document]
kvega005: This fine. We are copying docker.

src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -82,6 +85,7 @@ _(Time,           "time",                L"t",              Kind::Value,       L
 _(TMPFS,          "tmpfs",               NO_ALIAS,          Kind::Value,       Localization::WSLCCLI_TMPFSArgDescription()) \
 _(TTY,            "tty",                 L"t",              Kind::Flag,        Localization::WSLCCLI_TTYArgDescription()) \
 _(User,           "user",                L"u",              Kind::Value,       Localization::WSLCCLI_UserArgDescription()) \
+_(Username,       "username",            L"u",              Kind::Value,       Localization::WSLCCLI_LoginUsernameArgDescription()) \
```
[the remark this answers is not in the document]
kvega005: This fine. We are copying docker.

localization/strings/en-US/Resources.resw
```diff
@@ -2293,6 +2299,57 @@ For privacy information about this product please visit https://aka.ms/privacy.<
   <data name="WSLCCLI_ImageSaveLongDesc" xml:space="preserve">
     <value>Saves images.</value>
   </data>
+  <data name="WSLCCLI_LoginDesc" xml:space="preserve">
+    <value>Log in to a registry.</value>
+  </data>
+  <data name="WSLCCLI_LoginLongDesc" xml:space="preserve">
+    <value>Log in to a registry. If no server is specified, the default is defined by the session.</value>
+  </data>
```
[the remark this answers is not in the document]
kvega005: This fine. We are copying docker.

src/windows/wslc/commands/RegistryCommand.cpp
kvega005: We actually need `mode` here so we can restore it later, so it can't just call `IsConsoleHandle` as a pure boolean gate and throw the mode away.

I could restructure it to call `IsConsoleHandle`, then `GetConsoleMode` but that's two syscalls for no benefit.

src/windows/wslc/commands/RegistryCommand.cpp
```diff
@@ -0,0 +1,182 @@
...
+        }));
+    }
+
+    return wil::scope_exit(std::function<void()>([] {}));
+}
+
+std::wstring Prompt(const std::wstring& label, bool maskInput)
+{
+    // Write without a trailing newline so the cursor stays inline (matching Docker's behavior).
+    std::wcerr << label;
```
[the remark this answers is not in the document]
kvega005: std::wcerr has unitbuf set by the C++ standard, so every operator<< automatically flushes. An explicit std::flush would be a no-op.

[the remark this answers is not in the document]
kvega005: https://en.cppreference.com/w/cpp/io/cerr.html

src/windows/wslc/services/ImageService.cpp
```diff
@@ -63,6 +64,26 @@ wil::unique_hfile ResolveBuildFile(const std::filesystem::path& contextPath)
     THROW_HR_WITH_USER_ERROR(E_INVALIDARG, Localization::MessageWslcBuildFileNotFound(contextPath));
 }
 
+std::string GetServerFromImage(const std::string& image)
+{
+    // Extract the registry domain from an image reference.
+    // Follows the same logic as Docker's splitDockerDomain in distribution/reference:
+    // If the part before the first '/' contains a '.' or ':', or is "localhost",
+    // it's treated as a registry domain. Otherwise it's a Docker Hub repo path.
+    auto slash = image.find('/');
```
OneBlue: I'd recommend using `wsl::windows::common::wslutil::ParseImage()` to extract the repo, and then `wsl::windows::common::wslutil::NormalizeRepo()` to extract the domain name. 

That method would have the benefit of normalizing the legacy domain and handling ip addresses

src/windows/wslc/services/RegistryService.cpp
```diff
@@ -0,0 +1,365 @@
...
+    wil::unique_hfile handle(CreateFileW(filePath.c_str(), GENERIC_READ | GENERIC_WRITE, 0, nullptr, OPEN_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr));
+    THROW_LAST_ERROR_IF(!handle.is_valid());
+
+    return handle;
+}
+
+wil::unique_hfile OpenFileShared()
+{
+    wil::unique_hfile handle(CreateFileW(
+        GetFilePath().c_str(), GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr));
```
OneBlue: I don't recommend sharing write access when we're reading the file. If someone writes while we're reading, we could fail to decode the json

src/windows/wslc/services/RegistryService.cpp
```diff
@@ -0,0 +1,365 @@
...
+    {
+        return nlohmann::json::object();
+    }
+
+    LARGE_INTEGER zero{};
+    THROW_IF_WIN32_BOOL_FALSE(SetFilePointerEx(handle.get(), zero, nullptr, FILE_BEGIN));
+
+    std::string buffer(static_cast<size_t>(size.QuadPart), '\0');
+    DWORD bytesRead = 0;
+    THROW_IF_WIN32_BOOL_FALSE(ReadFile(handle.get(), buffer.data(), static_cast<DWORD>(buffer.size()), &bytesRead, nullptr));
```
OneBlue: Unfortunately we're not guaranteed that we can read the entire file in a single ReadFile call here. 

The canonical way here would be to keep calling ReadFile() until we get `bytesRead == 0` to indicate the end of the file

OneBlue: Actually, there's an easier way: We could use `_fsopen` to get a FILE while setting the share flags, and then pass that file to `nlohmann::json::parse()`

src/windows/wslc/services/RegistryService.cpp
```diff
@@ -0,0 +1,365 @@
...
+void WriteJsonFile(const wil::unique_hfile& handle, const nlohmann::json& data)
+{
+    LARGE_INTEGER zero{};
+    THROW_IF_WIN32_BOOL_FALSE(SetFilePointerEx(handle.get(), zero, nullptr, FILE_BEGIN));
+    THROW_IF_WIN32_BOOL_FALSE(SetEndOfFile(handle.get()));
+
+    auto content = data.dump(2);
+    DWORD written = 0;
+    THROW_IF_WIN32_BOOL_FALSE(WriteFile(handle.get(), content.data(), static_cast<DWORD>(content.size()), &written, nullptr));
+    THROW_IF_WIN32_BOOL_FALSE(FlushFileBuffers(handle.get()));
```
OneBlue: The buffers are automatically flushed when we close the handle, so this shouldn't be needed

src/windows/wslc/services/RegistryService.cpp
OneBlue: I would recommend letting this user fail (maybe with an error message like `failed to read credentials from: /path/to/file`, so the user can decide what to do here (otherwise we might end up deleting all the credentials from storage)

src/windows/wslc/services/RegistryService.cpp
```diff
@@ -0,0 +1,365 @@
...
+    cred.Type = CRED_TYPE_GENERIC;
+    cred.TargetName = const_cast<LPWSTR>(targetName.c_str());
+    cred.CredentialBlobSize = static_cast<DWORD>(credential.size());
+    cred.CredentialBlob = reinterpret_cast<LPBYTE>(const_cast<char*>(credential.data()));
+    cred.Persist = CRED_PERSIST_LOCAL_MACHINE;
+
+    THROW_IF_WIN32_BOOL_FALSE(CredWriteW(&cred, 0));
+}
+
+std::optional<std::string> RegistryService::WinCredGetCredential(const std::string& serverAddress)
```
OneBlue: nit: to make this a bit easier to read, we could split the WinCred & the file based credential stores into two classes with a common interface. 

Then we could have a method like: 

```
ICredentialStorage OpenCredentialsStorage()
```

Which would return the right implementation based on configuration (this would also make it easy to write tests that apply to both storage backends)

src/windows/wslc/services/RegistryService.cpp
```diff
@@ -0,0 +1,365 @@
...
+namespace {
+
+std::filesystem::path GetFilePath()
+{
+    return wsl::windows::common::filesystem::GetLocalAppDataPath(nullptr) / L"wslc" / L"registry-credentials.json";
+}
+
+wil::unique_hfile RetryOpenFileOnSharingViolation(const std::function<wil::unique_hfile()>& openFunc)
+{
+    return wsl::shared::retry::RetryWithTimeout<wil::unique_hfile>(openFunc, std::chrono::milliseconds(100), std::chrono::seconds(1), []() {
```
OneBlue: nit: We might to return a specialized error message if we ever hit this, like `failed to open file /path/to/file after X seconds`

src/windows/wslc/services/FileCredStorage.cpp
```diff
@@ -0,0 +1,238 @@
...
+}
+
+std::vector<std::wstring> FileCredStorage::List()
+{
+    auto file = RetryOpenFileOnSharingViolation(OpenFileShared);
+    auto data = ReadJsonFile(file.get());
+
+    std::vector<std::wstring> result;
+
+    for (const auto& [key, value] : data.items())
```
OneBlue: Sorry I missed this on my preview review. It's looking like we're storing the credentials as: 

```
{
    "domain": "credential"
}
``` 

This definitely works, but I think we should make it a bit more structured, something like: 

```
{
    version: 1,

     "credentials": {
      {
           "domain": {"token": "<protected-token-value>"}
      }
}
}
```

This would have the benefit of allowing us to easily change the format in a future upgrade, and add more fields  to specific credentials (like timestamp / format) if we never need to

OneBlue: (For this, having a schema would probably be the easiest way)

src/windows/wslc/services/RegistryService.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    // Strip path
+    if (auto pos = input.find('/'); pos != std::string::npos)
+    {
+        input = input.substr(0, pos);
+    }
+
+    // Map Docker Hub aliases to canonical key.
+    if (input == "docker.io" || input == "index.docker.io")
```
OneBlue, line 39: Given that `wsl::windows::common::wslutil::NormalizeRepo()` should deal with this for us, do we need to special case for `docker.io`, or could we just store the default credentials with the `docker.io` key ?

src/windows/wslc/services/FileCredStorage.cpp
```diff
@@ -0,0 +1,238 @@
...
+    }
+    catch (const nlohmann::json::parse_error&)
+    {
+        THROW_HR_WITH_USER_ERROR(WSL_E_INVALID_JSON, Localization::WSLCCLI_CredentialFileCorrupt(GetFilePath()));
+    }
+}
+
+void WriteJsonFile(FILE* f, const nlohmann::json& data)
+{
+    fseek(f, 0, SEEK_SET);
```
OneBlue: We should check for errors when we manipulate `f` here

src/windows/wslc/services/FileCredStorage.cpp
```diff
@@ -0,0 +1,238 @@
...
+    _chsize_s(_fileno(f), 0);
+
+    auto content = data.dump(2);
+    fwrite(content.data(), 1, content.size(), f);
+    fflush(f);
```
OneBlue: We shouldn't need to explicitly flush here

test/windows/wslc/e2e/WSLCE2ERegistryTests.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+            // Negative path for logout command: second logout should fail.
+            result = RunWslc(std::format(L"logout {}", registryAddressW));
+            VERIFY_ARE_EQUAL(1u, result.ExitCode.value_or(0));
+            VERIFY_IS_TRUE(result.Stderr.has_value());
+            VERIFY_IS_TRUE(result.Stderr->find(L"Not logged in to") != std::wstring::npos);
+        }
+    }
+
+    WSLC_TEST_METHOD(WSLCE2E_Registry_Login_HelpCommand)
```
OneBlue, line 108: nit: We should also have coverage for when the username / password is invalid

src/windows/wslc/services/RegistryService.cpp
kvega005: We still need this check for the login/logout path.
Without the mapping, these would be stored/looked up under different keys:

- `wslc login docker.io` → stores under "docker.io"
- `wslc logout index.docker.io` → erases under "index.docker.io" 

By mapping both to DefaultServer, all of these resolve to the same credential key ("https://index.docker.io/v1/"):
- `wslc login (no server)`
- `wslc login docker.io`
- `wslc login index.docker.io`
- `wslc login https://index.docker.io/v1/`

Docker CLI has the identical behavior — getAuthConfigKey maps both docker.io and index.docker.io to https://index.docker.io/v1/.
https://github.com/docker/cli/blob/84b357f6b0685d89828adee1dffced6aa072b4e7/cli/command/registry.go#L223

src/windows/wslc/services/RegistryService.cpp
```diff
@@ -0,0 +1,105 @@
...
+    auto key = ResolveCredentialKey(serverAddress);
+    auto [username, secret] = storage->Get(key);
+
+    if (username == TokenUsername)
+    {
+        return BuildRegistryAuthHeader(secret);
+    }
+
+    return BuildRegistryAuthHeader(username, secret);
+}
```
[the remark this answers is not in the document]
kvega005: This matches the docker CLI behavior. It just sends empty auth header when no credentials exist.

OneBlue (approved):

PR #40181, opened 2026-04-14
Update volume APIs to match docker arguments

kvega005:
## PR Checklist


src/windows/wslcsession/WSLCVhdVolume.cpp
```diff
@@ -26,16 +26,35 @@ using wsl::shared::Localization;
...
+        return name;
+    }
+
+    ULONGLONG ParseSizeBytes(std::map<std::string, std::string>& DriverOpts)
+    {
+        const auto it = DriverOpts.find("SizeBytes");
+        THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessageWslcInvalidVolumeOptions("SizeBytes"), it == DriverOpts.end());
+
+        auto& value = it->second;
+        THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessageInvalidSize(value), value.empty() || value[0] == '-');
```
OneBlue: Is this check required ? I'd expect `ToUInt64()` to fail if `value` was empty or negative

test/windows/WSLCTests.cpp
```diff
@@ -3393,18 +3415,21 @@ class WSLCTests
...
         WSLCVolumeOptions volumeOptions{};
         volumeOptions.Name = volumeName1.c_str();
-        volumeOptions.Type = "vhd";
-        volumeOptions.Options = R"({"SizeBytes":"1073741824"})";
-        VERIFY_SUCCEEDED(m_defaultSession->CreateVolume(&volumeOptions));
+
+        WSLCDriverOption driverOpts[] = {{"SizeBytes", "1073741824"}};
+        volumeOptions.DriverOpts = driverOpts;
+        volumeOptions.DriverOptsCount = ARRAYSIZE(driverOpts);
+        WSLCVolumeInformation volInfo{};
```
OneBlue, line 3425: nit: Here we're leaking the pointers returned by CreateVolume(). 

We might want to create a helper similar to [this][https://github.com/microsoft/WSL/blob/f60ec519d5add21a450d9f9d14a0239f7a99de5d/src/windows/common/wslutil.h#L130) to make it easier for callers to properly deallocate the volume info

src/windows/service/inc/wslc.idl
```diff
@@ -560,24 +568,47 @@ typedef struct _WSLCTagImageOptions
 
 typedef struct _WSLCVolumeOptions
 {
-    LPCSTR Name;
-    LPCSTR Type;
-    [unique] LPCSTR Options;
+    [unique] LPCSTR Name;
+    [unique] LPCSTR Driver;
+    [unique, size_is(DriverOptsCount)] const WSLCDriverOption* DriverOpts;
```
OneBlue, line 573: Given that `DriverOpts` and `Labels` require memory allocation & deallocation, I would recommend removing them from this structure to make things easier for the caller (especially since they're not needed for the simple `wslc volume ls` case). 

If the caller needs to see the driver options or labels, they can call Inspect()

src/windows/wslcsession/WSLCVhdVolume.cpp
```diff
@@ -26,16 +26,35 @@ using wsl::shared::Localization;
 namespace wsl::windows::service::wslc {
 
 namespace {
-    std::string SerializeVhdVolumeMetadata(const std::filesystem::path& HostPath, ULONGLONG SizeBytes)
+    std::string GenerateName()
     {
-        WSLCVolumeMetadata metadata{};
-        metadata.Type = WSLCVhdVolumeType;
+        BYTE randomBytes[32];
+        THROW_IF_NTSTATUS_FAILED(BCryptGenRandom(nullptr, randomBytes, sizeof(randomBytes), BCRYPT_USE_SYSTEM_PREFERRED_RNG));
```
OneBlue: nit: Given that we don't need cryptographically secure random bits, I'd recommend using `std::generate` here (example: https://github.com/microsoft/WSL/blob/acd3cd450ecc555aa987436ca52c264fac04bd57/src/windows/common/WslCoreConfig.cpp#L166-L171 )

OneBlue (commented):
Change looks great! Couple minor comments

src/windows/wslcsession/WSLCVhdVolume.cpp
kvega005: The check `value.empty() || value[0] == '-'` is needed. strtoull does not fail on negative inputs — it silently wraps them. Per the C standard, `strtoull("-1", ...)` returns `ULLONG_MAX` with `errno == 0` and `*end` pointing past the string. So without the `value[0] == '-'` guard, `"-1"` would be accepted as a valid (huge) size.

kvega005: The `value.empty()` isn't really needed because `stoul` would return 0, which gets caught in the later `SizeBytes==0`

src/windows/common/wslutil.cpp
```diff
@@ -1460,4 +1460,28 @@ std::string wsl::windows::common::wslutil::BuildRegistryAuthHeader(const std::st
...
+    std::map<std::string, std::string> result;
+
+    for (ULONG i = 0; i < count; i++)
+    {
+        THROW_HR_IF_NULL_MSG(E_INVALIDARG, pairs[i].Key, "Key at index %lu is null", i);
+        THROW_HR_IF_NULL_MSG(E_INVALIDARG, pairs[i].Value, "Value at index %lu is null", i);
+
+        if (reservedKey != nullptr)
+        {
+            THROW_HR_IF_MSG(E_INVALIDARG, strcmp(pairs[i].Key, reservedKey) == 0, "Key '%hs' is reserved", reservedKey);
```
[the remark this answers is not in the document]
kvega005: They match its the same

OneBlue (approved):

PR #40195, opened 2026-04-15
CLI: Initialize template support

AmelBawa-msft:
Enable using Go templates from C++
```
wslc ps -a --format 'Name is {{.Name}} and state is {{.State}}'
Name is demo and state is 3
Name is demo-2 and state is 3
```

## PR Checklist


src/windows/wslc/tools/gotemplate/render.go
```diff
@@ -0,0 +1,80 @@
...
+	Fail_ParseTemplate   = 3
+	Fail_ExecuteTemplate = 4
+)
+
+// TryRenderGoTemplate renders a Go template with the provided JSON data.
+// Returns 1 on success, 0 on failure. On success, *output contains the rendered result.
+// On failure, *output contains the error message. The caller must free *output with FreeGoString.
+//
+//export TryRenderGoTemplate
+func TryRenderGoTemplate(templateStr *C.char, jsonData *C.char, output **C.char) C.int {
```
OneBlue, line 41: This is clever ! If we easily call into a go dll, this makes supporting templates a lot easier. 

One thing I'll recommend though is that given that are most likely not going to make changes to that dll very often, would to build it separately in https://github.com/microsoft/WSL-deps and consume via a nuget package, so that we don't need to rebuild it locally on every build

OneBlue: It will also probably be easier to build the dll in that repo, since we can use msbuild directly and not have to go through cmake

benhillis:
FreeGoString(rawOutput) is called unconditionally without null check, and if TryRenderGoTemplate throws, rawOutput leaks. Use a wil::scope_exit for cleanup.

PR #40197, opened 2026-04-16
Fix batch of minor bugs

chemwolf6922:

This PR fixes a batch of minor bugs / issues found during code review.

## PR Checklist

### DIFF
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -471,7 +471,7 @@ Return Value:
         return {};
     }
 
-    struct sockaddr_nl Address;
+    struct sockaddr_nl Address{};
     Address.nl_family = AF_NETLINK;
     if (bind(Fd.get(), (struct sockaddr*)&Address, sizeof(Address)) < 0)
     {
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -587,7 +587,7 @@ Return Value:
     std::string content = wsl::shared::string::ReadFile<char, char>(std::format("/sys/block/{}/dev", BlockDeviceName).c_str());
     auto separator = content.find(':');
 
-    if (separator == 0 || separator == std::string::npos)
+    if (separator == std::string::npos || separator == 0 ||  separator + 1 == content.size())
     {
         LOG_ERROR("Failed to parse device number '{}' for device '{}'", content.c_str(), BlockDeviceName.c_str());
         THROW_ERRNO(EINVAL);
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -587,7 +587,7 @@ Return Value:
     std::string content = wsl::shared::string::ReadFile<char, char>(std::format("/sys/block/{}/dev", BlockDeviceName).c_str());
     auto separator = content.find(':');
 
-    if (separator == std::string::npos || separator == 0 ||  separator + 1 == content.size())
+    if (separator == std::string::npos || separator == 0 || separator + 1 == content.size())
     {
         LOG_ERROR("Failed to parse device number '{}' for device '{}'", content.c_str(), BlockDeviceName.c_str());
         THROW_ERRNO(EINVAL);
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -324,7 +324,7 @@ try
 
             sched_param Parameter{};
             Parameter.sched_priority = 0;
-            THROW_LAST_ERROR_IF(pthread_setschedparam(pthread_self(), SCHED_IDLE, &Parameter) < 0);
+            THROW_LAST_ERROR_IF(pthread_setschedparam(pthread_self(), SCHED_IDLE, &Parameter) != 0);
 
             //
             // Periodically check if the machine is idle by querying procfs for CPU usage.
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -587,7 +587,7 @@ Return Value:
     std::string content = wsl::shared::string::ReadFile<char, char>(std::format("/sys/block/{}/dev", BlockDeviceName).c_str());
     auto separator = content.find(':');
 
-    if (separator == 0 || separator - 1 >= content.size() || separator == std::string::npos)
+    if (separator == 0 || separator == std::string::npos)
     {
         LOG_ERROR("Failed to parse device number '{}' for device '{}'", content.c_str(), BlockDeviceName.c_str());
         THROW_ERRNO(EINVAL);
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -429,7 +429,7 @@ try
                 if (PageReportingOrder != 0 && (Start - Stop) > IdleThreshold)
                 {
                     std::this_thread::sleep_for(std::chrono::seconds(1));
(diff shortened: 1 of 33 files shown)


src/linux/init/main.cpp
```diff
@@ -587,7 +587,7 @@ Return Value:
     std::string content = wsl::shared::string::ReadFile<char, char>(std::format("/sys/block/{}/dev", BlockDeviceName).c_str());
     auto separator = content.find(':');
 
-    if (separator == 0 || separator - 1 >= content.size() || separator == std::string::npos)
```
OneBlue, line 590: Reading through this, `separator - 1 >= content.size()` is there to catch the case where the separator is the last character in `content`. What's the reason for removing this check ?

src/linux/init/util.cpp
```diff
@@ -172,6 +172,7 @@ Return Value:
     if (!InteropConnection)
     {
         LOG_ERROR("accept4 failed {}", errno);
+        return InteropConnection;
```
OneBlue: nit: we might want to `return {}` to make it clear that we're returning an empty socket here

src/linux/init/util.cpp
```diff
@@ -784,10 +785,6 @@ Return Value:
         if (Output)
         {
             (*Output) += Buffer.data();
-            if (Result < 0)
```
OneBlue, line 787: Wow, this was essentially broken all the time. Great catch !

src/linux/plan9/p9io.cpp
```diff
@@ -24,14 +24,10 @@ CoroutineIoIssuer::CoroutineIoIssuer(int fd) : m_FileDescriptor(fd)
 void CoroutineIoIssuer::Callback(sigval value)
 {
     const auto operation = static_cast<CoroutineIoOperation*>(value.sival_ptr);
+    auto error = aio_error(&operation->ControlBlock);
     auto bytesTransferred = aio_return(&operation->ControlBlock);
```
OneBlue, line 32: Reading through: https://man7.org/linux/man-pages/man3/aio_return.3.html 

It's looking like we should only call it if `aio_error` returns a value different than `EINPROGRESS`

src/shared/inc/message.h
```diff
@@ -167,12 +167,13 @@ class MessageWriter
 
     size_t GetRelativeIndex(unsigned int& Index)
     {
-        const size_t Offset = reinterpret_cast<char*>(&Index) - reinterpret_cast<char*>(m_buffer.data());
+        const auto* indexPtr = reinterpret_cast<char*>(&Index);
+        const auto* bufferStart = reinterpret_cast<char*>(m_buffer.data());
 
         // Validate that 'Index' is actually within the bounds of our buffer
-        assert(Offset >= 0 && Offset < m_buffer.size());
+        assert(indexPtr >= bufferStart && indexPtr + sizeof(Index) <= bufferStart + m_buffer.size());
```
OneBlue: I'm not sure that this is correct: 

```indexPtr + sizeof(Index) <= bufferStart + m_buffer.size()```

This assumes that the caller is trying to read / write `sizeof(Index)` bytes, which isn't necessarily true (the caller could be looking to write only 1 byte for instance)

**Review by OneBlue — COMMENTED** [review]
Lots of great fixes in there, looks great ! Couple minor questions

File: src/linux/init/main.cpp (hunk 1 above)
Comment (chemwolf6922) [reply src/linux/init/main.cpp:590]: The separator - 1 >= content.size() is a null op. To test if the : ends the string. We should test for separator + 1 == content.size() right?

File: src/shared/inc/message.h [hunk 6]
```diff
@@ -167,12 +167,13 @@ class MessageWriter
...
-        const size_t Offset = reinterpret_cast<char*>(&Index) - reinterpret_cast<char*>(m_buffer.data());
+        const auto* indexPtr = reinterpret_cast<char*>(&Index);
+        const auto* bufferStart = reinterpret_cast<char*>(m_buffer.data());
 
         // Validate that 'Index' is actually within the bounds of our buffer
-        assert(Offset >= 0 && Offset < m_buffer.size());
+        assert(indexPtr >= bufferStart && indexPtr < bufferStart + m_buffer.size());
 
-        return Offset;
+        return static_cast<size_t>(indexPtr - bufferStart);
```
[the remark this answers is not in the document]
chemwolf6922: ?

src/shared/inc/message.h
```diff
@@ -171,7 +171,7 @@ class MessageWriter
         const auto* bufferStart = reinterpret_cast<char*>(m_buffer.data());
 
         // Validate that 'Index' is actually within the bounds of our buffer
-        assert(indexPtr >= bufferStart && indexPtr + sizeof(Index) <= bufferStart + m_buffer.size());
+        assert(indexPtr >= bufferStart && indexPtr < bufferStart + m_buffer.size());
```
OneBlue: What's the reason for that change ? This check is there to make sure that we have at least `sizeof(unsigned int)` bytes of buffer available in the index that we return

src/shared/inc/prettyprintshared.h
```diff
@@ -85,7 +85,7 @@ inline void PrettyPrint(std::stringstream& Out, const T (&Value)[Size])
     Out << "[";
     for (auto i = 0; i < Size; i++)
     {
-        if (i > 0 && i < Size - 1)
```
OneBlue, line 88: What's the thinking for that change ? Won't this add an additional coma at the end of the list ?

src/windows/common/filesystem.cpp
```diff
@@ -843,6 +843,7 @@ std::string wsl::windows::common::filesystem::GetLinuxHostName()
 
     WI_ASSERT((size <= LX_HOST_NAME_MAX) && (hostName.size() == size + 1));
 
+    hostName.resize(size);
```
OneBlue: Reading through this, it's looking like we're allocating an extra character in `hostname`. Should we just change its allocation to: `    std::string hostName(size - 1, '\0');` since `size` returns the number of characters required + 1 on failure ?

src/shared/inc/message.h
chemwolf6922: image

src/shared/inc/prettyprintshared.h
chemwolf6922: This comma print is before the value print. The original logic will miss one between the -2 and -1 element.

src/windows/common/filesystem.cpp
chemwolf6922: I tried passing a buffer of size - 1. It will return ERROR_MORE_DATA.

src/shared/inc/message.h
chemwolf6922: I changed it back. Please let me know if I need to change it again.

OneBlue: Ah, that's funny. Sorry about that, I was looking at the "changes since your last review" diff, which makes it harder to see things like this

src/windows/common/filesystem.cpp
OneBlue: We should still pass `size` to `GetComputerName`, since on input it expects the number of chars in the buffer, and std::string always allocated an additional byte for the '\0'

OneBlue (approved):

benhillis (approved):

PR #40137, opened 2026-04-08
Add support for mounting files and non-existing folders as volumes 

OneBlue:
## Summary of the Pull Request

This change adds support for mounting files and non-existing folders as volumes 

## PR Checklist

- [ ] **Closes:** Link to issue #xxx
- [ ] **Communication:** I've discussed this with core contributors already. If work hasn't been agreed, this work might be rejected
- [X] **Tests:** Added/updated if needed and all pass
- [ ] **Localization:** All end user facing strings can be localized
- [ ] **Dev docs:** Added/updated if needed
- [ ] **Documentation updated:** If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/wsl/) and link it here: #xxx

PR #40201, opened 2026-04-16
Fix std::format type mismatch mixing std::wstring and std::string

benhillis:
## Problem

In `WSLCContainer::Create()`, the bind source path for file mounts is constructed using `std::format` with a narrow format string but passing `sourceFilename` (`std::wstring`) alongside `parentVMPath` (`std::string`).

`std::format` with a narrow format string does not accept `std::wstring` arguments in standard C++20. This is a type mismatch that could cause compilation failures or undefined behavior.

## Fix

Convert `sourceFilename` from `std::wstring` to `std::string` via `WideToMultiByte()` before passing it to `std::format`, consistent with how the codebase handles wide-to-narrow conversions elsewhere.

## Testing

Type-correctness fix in the volume bind path construction for file mounts (introduced in #40137).

### DIFF
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -1191,7 +1191,9 @@ std::unique_ptr<WSLCContainerImpl> WSLCContainerImpl::Create(
         volumes.push_back(WSLCVolumeMount{hostPath, parentVMPath, volume.ContainerPath, static_cast<bool>(volume.ReadOnly), sourceFilename});
 
         auto options = volume.ReadOnly ? "ro" : "rw";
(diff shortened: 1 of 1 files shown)


AmelBawa-msft (approved):

JohnMcPMS (approved):

OneBlue (changes requested):
I don't think this is true (although AIs seem to flag this a lot). 

We have a [set of custom formatters](https://github.com/microsoft/WSL/blob/b2321b9fe2c37a3bb9b383ebfce3d3145a388f86/src/shared/inc/stringshared.h#L770)  that handle narrow <-> wide string conversion for us. We use them all across the codebase

benhillis:
Pierre is right on the custom formatters. Copilot was wrong on this one.

PR #40211, opened 2026-04-16
CLI: Initial support for inspect command

AmelBawa-msft:
- Initial support for inspect command

## PR Checklist


src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -63,6 +63,7 @@ _(Name,           "name",                NO_ALIAS,          Kind::Value,       L
 _(NoCache,        "no-cache",            NO_ALIAS,          Kind::Flag,        Localization::WSLCCLI_NoCacheArgDescription()) \
 _(NoPrune,        "no-prune",            NO_ALIAS,          Kind::Flag,        Localization::WSLCCLI_NoPruneArgDescription()) \
 _(NoTrunc,        "no-trunc",            NO_ALIAS,          Kind::Flag,        Localization::WSLCCLI_NoTruncArgDescription()) \
+_(ObjectId,       "object-id",           NO_ALIAS,          Kind::Positional,  L"Name or Id of any object type") \
```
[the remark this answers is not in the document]
dkbennett: We should continue the pattern of using localization functions here.

src/windows/wslc/commands/InspectCommand.cpp
```diff
@@ -0,0 +1,45 @@
...
+}
+
+std::wstring InspectCommand::ShortDescription() const
+{
+    return {L"Inspect objects."};
+}
+
+std::wstring InspectCommand::LongDescription() const
+{
+    return {L"Inspects objects."};
```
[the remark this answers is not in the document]
dkbennett: Copilot's suggestion is not correct but it is right in that these should be localized.

dkbennett (commented):
Minor loc and should have at least an E2E test that verifies the multi-object functionality for inspect with mixed containers & images.

src/windows/wslc/commands/InspectCommand.cpp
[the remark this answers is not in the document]
AmelBawa-msft: Yes, I still have a few minor enhancements before I publish the PR 😊

AmelBawa-msft:
> Minor loc and should have at least an E2E test that verifies the multi-object functionality for inspect with mixed containers & images.

@dkbennett Thank you for the review. Sorry the PR wasn’t quite ready for review at the time (draft), but I’ve addressed your comments in my latest commits. Please feel free to take another look. I’ll do one more pass to make sure nothing’s missing before publishing. You’re welcome to review now, or wait until it’s published, whichever you prefer 😊

src/windows/wslc/tasks/InspectTasks.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+    try
+    {
+        fn();
+        return true;
+    }
+    catch (const wil::ResultException& ex)
+    {
+        auto errorCode = ex.GetErrorCode();
+        if (errorCode == notFoundError || errorCode == HRESULT_FROM_WIN32(ERROR_BAD_ARGUMENTS) || errorCode == E_INVALIDARG)
```
OneBlue, line 41: What's the reason for not throwing for ERROR_BAD_ARGUMENTS and E_INVALIDARG here ?

AmelBawa-msft: Similar to [Docker's behavior](https://github.com/docker/cli/blob/89958b190cca2d6bc6005103accccefafd5f62f5/cli/command/system/inspect.go#L285-L288), some errors are considered skippable

For instance, removing this condition `errorCode == HRESULT_FROM_WIN32(ERROR_BAD_ARGUMENTS)`, causes the following command to fail:
```
PS C:\> wslc inspect A
invalid reference format: repository name (library/A) must be lowercase
Error code: 0x800700a0
```

And removing this condition `errorCode == E_INVALIDARG`, causes the following to fail:
```
PS C:\> wslc inspect 'debian:latest'
Invalid name: 'debian:latest'
Error code: E_INVALIDARG
```

The reason the above fails is inspect command calls container inspect first, and 'debian:latest' is an invalid container name argument

test/windows/wslc/e2e/WSLCE2EInspectTests.cpp
```diff
@@ -0,0 +1,235 @@
...
+        return commands.str();
+    }
+
+    std::wstring GetAvailableOptions() const
+    {
+        std::wstringstream options;
+        options << L"The following options are available:\r\n"                 //
+                << L"  -t,--type    Type of the object to inspect\r\n"         //
+                << L"  --session    Specify the session to use\r\n"            //
+                << L"  -h,--help    Shows help about the selected command\r\n" //
```
dkbennett: Help flag was just changed, should be -? now.

src/windows/wslc/tasks/InspectTasks.cpp
dkbennett: This makes sense to me. Inspect is generic and is attempting to deduce the type of the object, and I would expect most objects would only be valid for one type of inspection, so if we aren't given the type by the user it is expected to fail for all other types. It would only be an invalid arg overall result if it was not found after attempting all types or the specified type.

dkbennett (dismissed):
I approve but auto-merge is enabled, and checking in as-is will have a test failure in the help. Only putting on a block here to force the fix so as not to break CI.

src/windows/wslc/tasks/InspectTasks.cpp
```diff
@@ -0,0 +1,96 @@
...
+        std::optional<wslc_schema::InspectContainer> container;
+        std::optional<wslc_schema::InspectImage> image;
+
+        if (WI_IsFlagSet(type, InspectType::Container) && TryInspectContainer(session, id, container))
+        {
+            array.push_back(std::move(*container));
+        }
+        else if (WI_IsFlagSet(type, InspectType::Image) && TryInspectImage(session, id, image))
+        {
+            array.push_back(std::move(*image));
```
kevpar, line 91: From my testing, it appears docker will only return a single object from inspect, preferring containers over images (not sure about networks and volumes). We should make sure we are following the same semantics where possible.

AmelBawa-msft: Correct, the order and semantic should is available [here](https://github.com/docker/cli/blob/89958b190cca2d6bc6005103accccefafd5f62f5/cli/command/system/inspect.go#L185)

kevpar: Sounds good

OneBlue (approved):

PR #40215, opened 2026-04-17
Add Begin/End telemetry around long-wait create-instance steps

shuaiyuanxx:
This pull request adds detailed telemetry instrumentation to key lifecycle and startup operations in the WSL service, with the goal of making it easier to diagnose and attribute startup hangs or performance issues. The changes wrap potentially long-running or blocking operations with paired "Begin" and "End" telemetry events, allowing the backend to precisely identify where time is spent or where hangs occur, especially in interactions with third-party plugins, the VM, networking, and the init daemon.

The most important changes are:

**Telemetry Instrumentation for Key Operations:**

* Added paired `WSL_LOG_TELEMETRY` "Begin" and "End" events around plugin hooks such as `OnVMStarted` and `OnDistributionStarted`, enabling attribution of hangs to specific plugins. [[1]](diffhunk://#diff-6802192381c013b43d0d165c72e8c77862e8045453b6906cb456bdffda1377d2L181-R198) [[2]](diffhunk://#diff-6802192381c013b43d0d165c72e8c77862e8045453b6906cb456bdffda1377d2L214-R247)
* Instrumented VM lifecycle operations in `WslCoreVm::Initialize`, including VM creation (`HcsCreateSystem`), start (`HcsStartSystem`), guest connection (`WaitForMiniInitConnect`), reading guest capabilities, and networking setup (including NAT network creation and configuration), to provide granular visibility into each step. [[1]](diffhunk://#diff-3babaa9ca95a69863e28dd9a573938c2fa8446fd734e609774769d7c7dbd9c2fR323-R335) [[2]](diffhunk://#diff-3babaa9ca95a69863e28dd9a573938c2fa8446fd734e609774769d7c7dbd9c2fR358-R360) [[3]](diffhunk://#diff-3babaa9ca95a69863e28dd9a573938c2fa8446fd734e609774769d7c7dbd9c2fR372-R374) [[4]](diffhunk://#diff-3babaa9ca95a69863e28dd9a573938c2fa8446fd734e609774769d7c7dbd9c2fR456-R487) [[5]](diffhunk://#diff-3babaa9ca95a69863e28dd9a573938c2fa8446fd734e609774769d7c7dbd9c2fR576-R584) [[6]](diffhunk://#diff-3babaa9ca95a69863e28dd9a573938c2fa8446fd734e609774769d7c7dbd9c2fR600-R616) [[7]](diffhunk://#diff-3babaa9ca95a69863e28dd9a573938c2fa8446fd734e609774769d7c7dbd9c2fR722-R738)
* Added telemetry to VHD attachment and init message sending in `WslCoreVm::CreateInstance`, to distinguish storage-related delays from guest startup issues. [[1]](diffhunk://#diff-3babaa9ca95a69863e28dd9a573938c2fa8446fd734e609774769d7c7dbd9c2fR1244-R1263) [[2]](diffhunk://#diff-3babaa9ca95a69863e28dd9a573938c2fa8446fd734e609774769d7c7dbd9c2fR1302-R1319)
* Wrapped init daemon connection setup in `WslCoreVm::CreateInstanceInternal` with telemetry to surface hangs on guest connection.

**Enhanced Observability for Init Daemon and DrvFs Initialization:**

* Added begin/end telemetry events around waits for the init daemon's startup, configuration response, and DrvFs initialization in `WslCoreInstance`, making it possible to determine if hangs are due to the init daemon, DrvFs, or other factors. [[1]](diffhunk://#diff-0e25701298b7958b1824830072849de999c1df675098b30c04afd1e7bff97c62R47-R67) [[2]](diffhunk://#diff-0e25701298b7958b1824830072849de999c1df675098b30c04afd1e7bff97c62R392-R411) [[3]](diffhunk://#diff-0e25701298b7958b1824830072849de999c1df675098b30c04afd1e7bff97c62R431-R449)

**Documentation and Comments:**

* Added explanatory comments throughout the code to clarify the rationale for each telemetry event and the types of hangs or delays they are intended to diagnose. (Multiple references above)

These changes significantly improve the service’s ability to diagnose, attribute, and resolve startup hangs and performance bottlenecks by providing fine-grained telemetry on all major operations.
## PR Checklist


yeelam-gordon:
**Field-name casing consistency.** The new events mix casing — `"distroName"`, `"instanceId"`, `"vmId"`, `"timeoutMs"` (camelCase) — while existing events in the same provider are PascalCase (`"Plugin"`, `"Sid"`, `"DistributionId"`). Kusto joins across old and new data get painful. Suggest picking one convention (PascalCase matches what's already there) and applying it uniformly across this PR.

yeelam-gordon:
**`PluginOnDistroStartedBegin/End` breaks continuity with existing queries.** The pre-existing event is `PluginOnDistributionStartedCall`. Renaming to `Distro` means any dashboard or alert keyed on `Distribution` stops working when this ships. Suggest `PluginOnDistributionStartedBegin/End` to preserve the naming chain.

yeelam-gordon:
**Unify the outcome field name.** Today: `PluginOnVmStartedEnd` → `"result"` (HRESULT), `WaitForCreateInstanceResultEnd` → `"initResult"`, `CreateNatNetworkEnd` → `"success"` (bool). Three names for effectively one concept. Suggest standardizing on `"hr"` for HRESULTs and `"success"` / `"status"` for enum or bool outcomes — and documenting which is which in the comment above each pair.

yeelam-gordon:
**Exception safety on inner Begin/End pairs.** If anything between a Begin and its End throws, the End is never emitted and the backend sees the same signal as a real hang. The outermost `CreateInstance` already uses `wil::scope_exit_log` — extending that pattern (or a small `TelemetryScope(beginName, endName, fields...)` RAII helper) to the inner pairs eliminates the ambiguity and roughly halves the boilerplate. Worth doing before merge — bad telemetry is worse than no telemetry here because it drives false alerts.

yeelam-gordon:
**`PDT_ProductAndServicePerformance` privacy tag applied inconsistently.** The HCS / NAT / `InitializeGuest` pairs carry it; the plugin, DrvFs, and `InitConfigResponse` pairs don't. Same data class and same intended use case — should be tagged the same. Easy to miss if not checked explicitly since the macro doesn't enforce it.

yeelam-gordon:
**Add a per-pair `pairId` GUID (and `attempt` on the outer scope).** The outer TraceLogging ActivityId groups everything in one CreateInstance attempt, but within a single attempt there can be multiple Begin/End pairs for the same phase — plugin hooks loop over N plugins, NAT creation can retry before falling back to virtio proxy, overlapping phases (e.g. GPU add alongside kernel version read) can interleave. Without a per-invocation ID, the backend state machine cannot unambiguously match Begin to End in those cases.

Suggestion:
- Mint `GUID pairId = wil::create_guid()` at each Begin, echo it on the matching End as `TraceLoggingValue(pairId, "pairId")`.
- Add `TraceLoggingValue(attempt, "attempt")` on the outermost Start/Stop pair only, so operators can filter retry-only traces without a self-join.

Intentionally **not** recommending `durationMs` or a `timedOut` bool on End events — both are trivially reconstructible in Kusto (`End - Begin` and `durationMs >= timeoutMs` respectively) and would just duplicate state.

benhillis:
Are distinct begin / end events preferable to activities?

shuaiyuanxx:
> Are distinct begin / end events preferable to activities?
Hi @benhillis Thanks for the suggestion, I'll rework the PR.

Issue #12759, opened 2025-03-30
Warning 1946 from WSL update to 2.4.13.0


patheikkinen:
### Windows Version

Microsoft Windows [Version 10.0.26100.3476]

### WSL Version

2.4.13.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Repro Steps

`wsl --update`

> Checking for updates.
> Updating Windows Subsystem for Linux to version: 2.4.13.
> Warning 1946.Property 'System.AppUserModel.ID' for shortcut 'WSL.lnk' could not be set.

Unlike a previous issue with similar symptoms, no PowerToys install is involved.

### Expected Behavior

Clearly a successful install shouldn't spit out warning 1946.

### Actual Behavior

See above.

patheikkinen:
[WslLogs-2025-03-30_15-21-20.zip](https://github.com/user-attachments/files/19525596/WslLogs-2025-03-30_15-21-20.zip)

Probably useless but required by policy.

btw it looks like the lnk file was updated and everything works so far, so the bug is probably that it emits a warning for no good reason...

patheikkinen:
How do you get past that log bot with an install time bug?

OneBlue:
Thank you @patheikkinen. This is unfortunately a powertoys issue. Fixed by https://github.com/microsoft/PowerToys/pull/37654

patheikkinen:
@OneBlue How could it possibly be? I don't have PowerToys installed, as I mentioned in the description.

Ryustiel:
Exact same issue here

Issue #11276, opened 2024-03-08
MSI installer "Warning: 1946.Property [...] could not be set."


ColorfulRhino:
### Windows Version

Microsoft Windows [Version 10.0.22631.3155]

### WSL Version

2.2.1.0

### Are you using WSL 1 or WSL 2?

- [X] WSL 2
- [ ] WSL 1

### Distro Version

Ubuntu 22.04

### Repro Steps

1) Download `wsl.2.2.1.0.x64.msi`, `wsl.2.1.4.0.x64.msi` or `wsl.2.1.1.0.x64.msi` from this repository's releases page.
2) Double click the installer
3) Confirm admin prompt
4) Let installer run
5) Warning appears:
![Screenshot](https://github.com/microsoft/WSL/assets/131405023/3406951e-50f4-4eea-9ba8-632679ece38f)
6) Click OK
7) Installation finishes

### Expected Behavior

No warning during install

### Actual Behavior

Warning appears during install:

![Screenshot](https://github.com/microsoft/WSL/assets/131405023/3406951e-50f4-4eea-9ba8-632679ece38f)

### Diagnostic Logs

From collected logs via `msiexec /i <path/to/msi> /L*V C:\wsllogs.txt`:
(See error where the blank lines are)

```
[...]
MSI (s) (98:0C) [23:30:51:044]: File: C:\Program Files\WSL\tr-TR\msrdc.exe.mui;	To be installed;	Won't patch;	No existing file
MSI (s) (98:0C) [23:30:51:044]: Source for file 'msrdc.exe.mui_tr_TR' is compressed
MSI (s) (98:0C) [23:30:51:045]: Executing op: FileCopy(SourceName=60wmt6ni.mui|rdclientax.dll.mui,SourceCabKey=rdclientax.dll.mui_tr_TR,DestName=rdclientax.dll.mui,Attributes=512,FileSize=143408,PerTick=65536,,VerifyMedia=1,,,,,CheckCRC=0,Version=1.2.5105.0,Language=1033,InstallMode=62914560,,,,,,,)
MSI (s) (98:0C) [23:30:51:045]: File: C:\Program Files\WSL\tr-TR\rdclientax.dll.mui;	To be installed;	Won't patch;	No existing file
MSI (s) (98:0C) [23:30:51:045]: Source for file 'rdclientax.dll.mui_tr_TR' is compressed
[...]

PR #40219, opened 2026-04-17
msi: eliminate Warning 1946 dialog by moving shortcut properties to custom action

yeelam-gordon:
## Summary

Eliminates [Warning 1946](https://github.com/microsoft/WSL/issues/11276) ΓÇö the modal dialog popup (or console warning) during WSL installation caused by shortcut property sharing violations.

### Before

During MSI install, the engine sets shortcut properties via the `MsiShortcutProperty` table. If **any** process holds the `.lnk` file open (search indexer, AV, PowerToys Run, shell extensions), `ShortcutPropertyCreate` fails with `ERROR_SHARING_VIOLATION` (HRESULT 32):

- **Double-click `.msi`**: **Modal dialog popup** blocks the install until user clicks OK
- **`wsl --update`**: Warning text printed to console

Warning 1946 dialog

This affects **any** process holding the file ΓÇö not just PowerToys ([#12759](https://github.com/microsoft/WSL/issues/12759) reported it without PowerToys installed).

### After

- **No Warning 1946** ΓÇö `ShortcutProperty` entries removed from WiX, so the MSI engine never attempts `ShortcutPropertyCreate`
- **Properties still set** ΓÇö new `RepairShortcutProperties` deferred custom action sets all properties via COM `IPropertyStore` with non-exclusive access (`STGM_SHARE_DENY_NONE`) and exponential backoff retry (50msΓåÆ100msΓåÆ200ms...ΓåÆ1s cap, 5s total timeout)
- **No modal dialog, no console warning, no user impact**

### Properties handled

| Shortcut | Property | Value |
|----------|----------|-------|
| `WSL.lnk` | `System.AppUserModel.ID` | `Microsoft.WSL` |
| `WSL.lnk` | `System.AppUserModel.ToastActivatorCLSID` | `{2B9C59C3-...}` |
| `WSL Settings.lnk` | `System.AppUserModel.IsSystemComponent` | `true` |

### Why not just suppress the warning?

- `MsiSetExternalUI` callback suppression only works for `wsl --update` (`INSTALLUILEVEL_NONE`)
- Direct `.msi` double-click uses the MSI engine's built-in UI ΓÇö **no way to suppress the modal dialog from the package side**
- `EmbeddedUI` could work but requires a new DLL for filtering one warning
- Removing `ShortcutProperty` entries is the only approach that eliminates the warning across **all install paths**

### Lifecycle coverage

| Scenario | CA runs? | Correct? |
|----------|:---:|:---:|
| Fresh install | Yes | Sets properties on new shortcuts |
| Major upgrade | Yes | Old MSI uninstalls (skips), new MSI installs (runs) |
| Uninstall | No | Shortcuts deleted by `RemoveFile` |
| Repair | Yes | Re-applies properties |
| Same-version reinstall | Yes | Idempotent |

## Testing

**8 real MSI install scenarios** with active `FileSystemWatcher` contention ΓÇö each ran full uninstallΓåÆlockΓåÆinstallΓåÆverify cycle:

```
[PASS] Scenario 1 : No contention          (MSI=0 W1946=0 AppId=Microsoft.WSL)
[PASS] Scenario 2 : Old PowerToys Run      (MSI=0 W1946=0 AppId=Microsoft.WSL)
[PASS] Scenario 3 : New PowerToys Run      (MSI=0 W1946=0 AppId=Microsoft.WSL)
[PASS] Scenario 4 : AV deep scan           (MSI=0 W1946=0 AppId=Microsoft.WSL)
[PASS] Scenario 5 : Search indexer         (MSI=0 W1946=0 AppId=Microsoft.WSL)
[PASS] Scenario 6 : Another installer      (MSI=0 W1946=0 AppId=Microsoft.WSL)
[PASS] Scenario 7 : Slow process 500ms     (MSI=0 W1946=0 AppId=Microsoft.WSL)
[PASS] Scenario 8 : Unknown process #12759 (MSI=0 W1946=0 AppId=Microsoft.WSL)
TOTAL: 8 passed, 0 failed
```

Each scenario verified: no Warning 1946 in MSI log, CA returned success, MSI exit 0, `AppUserModel.ID` and `ToastActivatorCLSID` set correctly via COM `IPropertyStore` read-back.

Test script: `test/shortcut_property_test/Test-AllScenariosMsi.ps1` (run as admin)

## PR Checklist

- [x] **Closes:** #13469
- [x] **Refs:** #11276, #12759
- [x] **Tests:** 8 real MSI install scenarios with file lock contention, all pass

## Changes

| File | Change |
|------|--------|
| `msipackage/package.wix.in` | Remove `ShortcutProperty` entries, add `RepairShortcutProperties` CA declaration + sequence |
| `src/windows/wslinstall/DllMain.cpp` | `RepairShortcutProperties` + `SetShortcutPropertyWithRetry` implementation |
| `src/windows/wslinstall/wslinstall.def` | Export `RepairShortcutProperties` |
| `src/windows/wslinstall/CMakeLists.txt` | Link `propsys.lib` |
| `test/shortcut_property_test/` | Comprehensive MSI contention test |

### DIFF
--- a/msipackage/package.wix.in
+++ b/msipackage/package.wix.in
@@ -21,8 +21,9 @@
                 <RemoveFile Id="CleanUpWSLShortCut" Directory="ProgramMenuFolder" Name="WSL" On="uninstall"/>
                 <File Id="wsl.exe" Name="wsl.exe" Source="${PACKAGE_INPUT_DIR}/wsl.exe" KeyPath="yes">
                     <Shortcut Id="WSLShortcut" Name="WSL" Description="Windows Subsystem for Linux" Arguments="--cd ~" Advertise="yes" Directory="ProgramMenuFolder" Icon="wsl.ico">
-                        <ShortcutProperty Key="System.AppUserModel.ID" Value="Microsoft.WSL"/>
-                        <ShortcutProperty Key="System.AppUserModel.ToastActivatorCLSID" Value="{2B9C59C3-98F1-45C8-B87B-12AE3C7927E8}"/>
+                        <!-- ShortcutProperty entries removed to prevent Warning 1946 dialog popups.
+                             Properties are set by the RepairShortcutProperties custom action with retry logic.
+                             See: microsoft/WSL#11276, microsoft/WSL#13469 -->
                     </Shortcut>
                 </File>
 
@@ -285,7 +286,8 @@
                     <RemoveFile Id="CleanUpWSLSettingsShortCutNonServer" Directory="ProgramMenuFolder" Name="WSLSettings" On="uninstall"/>
                     <File Id="wslsettings.exe_nonserver" Source="${PACKAGE_INPUT_DIR}/wslsettings/wslsettings.exe" KeyPath="yes" ShortName="kyk8fs6a.exe">
                         <Shortcut Id="WSLSettingsShortcutNonServer" Name="WSL Settings" Description="Windows Subsystem for Linux Settings" Advertise="yes" Directory="ProgramMenuFolder" Icon="wsl.ico">
-                            <ShortcutProperty Key="System.AppUserModel.IsSystemComponent" Value="true"/>
+                            <!-- ShortcutProperty removed to prevent Warning 1946 dialog popup.
+                                 Set by RepairShortcutProperties custom action instead. -->
                         </Shortcut>
                     </File>
                     <!-- Protocol registration -->
@@ -454,6 +456,17 @@
             Execute="deferred"
          />
 
+        <!-- Repair shortcut properties that may fail due to sharing violations (Warning 1946).
+             Runs after CreateShortcuts to re-apply properties with retry logic via COM IPropertyStore.
+             See: microsoft/WSL#11276, microsoft/WSL#13469 -->
+        <CustomAction Id="RepairShortcutProperties"
+            Impersonate="no"
+            BinaryRef="wslinstall.dll"
+            DllEntry="RepairShortcutProperties"
+            Return="ignore"
+            Execute="deferred"
+         />
+
         <CustomAction Id="RegisterLspCategories"
             Impersonate="no"
             BinaryRef="wslinstall.dll"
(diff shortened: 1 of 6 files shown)


benhillis (commented):
Please add a test to InstallerTests that exhibits this behavior.

ptrivedi:
@yeelam-gordon: the copilot comment about GetVersionExW is worth looking at. LGTM otherwise, once the InstallerTests related comment is addressed

OneBlue (commented):
I'm a bit conflicted on what we should do here. This warning isn't great, but it's very likely caused by another application locking our shortcut. 

The warning isn't great, but it's also non-fatal, and only requires user action when the MSI is installation manually, which is relatively rare.

Another parameter to this is that installer changes are high-risk. For instance a bug that would crash the installer during a custom action could leave the user's machine in a broken state, which has a very high potential of negative impact. 

Additionally, this new logic doesn't actually solve the issue, but only retries for a given amount of time, and then fails without warning, which would make root causing this harder. 

Taking all of this into account, my opinion would be to leave the warning as-is for now and look into root-causing the issue if we truly want to eliminate the warning.

@benhillis : what are you thoughts on this ?

yeelam-gordon:
Addressed review feedback in latest push:
- **@ptrivedi** / **@copilot-pull-request-reviewer** (GetVersionExW): Replaced with \IsWindowsServer()\ from VersionHelpers.h — already used elsewhere in this file (line 749). Cleaner, no deprecation warning, no pragma suppress needed ✓

ptrivedi:
> I'm a bit conflicted on what we should do here. This warning isn't great, but it's very likely caused by another application locking our shortcut.
> 
> The warning isn't great, but it's also non-fatal, and only requires user action when the MSI is installation manually, which is relatively rare.
> 
> Another parameter to this is that installer changes are high-risk. For instance a bug that would crash the installer during a custom action could leave the user's machine in a broken state, which has a very high potential of negative impact.
> 
> Additionally, this new logic doesn't actually solve the issue, but only retries for a given amount of time, and then fails without warning, which would make root causing this harder.
> 
> Taking all of this into account, my opinion would be to leave the warning as-is for now and look into root-causing the issue if we truly want to eliminate the warning.
> 
> @benhillis : what are you thoughts on this ?

Isn't the root cause a third-party process locking the .lnk? I understand it would be good to find what process, but we've seen that it could be different ones. My thought is, do these properties have to be set in the installer? What if they are applied later during startup or something? Do they need the banner etc. on the very first install before WSL has even run?

benhillis:
This is a ton of code to work around what seems to be an issue with the MSI infrastructure.

benhillis:

benhillis:
@yeelam-gordon Thanks for the thorough work here! We ended up going with a simpler fix in #40272 that addresses the Warning 1946 issue by removing the Advertise=yes from the MSI shortcuts. Given that, we're going to close this one out. We appreciate the contribution!

PR #40231, opened 2026-04-17
Allow creating anonymous VHD volumes

kvega005:
## PR Checklist

### DIFF
--- a/src/windows/wslcsession/WSLCVhdVolume.cpp
+++ b/src/windows/wslcsession/WSLCVhdVolume.cpp
@@ -26,24 +26,6 @@ using wsl::shared::Localization;
 namespace wsl::windows::service::wslc {
 
 namespace {
-    std::string GenerateName()
-    {
-        std::random_device rd;
-        std::independent_bits_engine<std::default_random_engine, CHAR_BIT, unsigned short> random(rd());
-
-        std::array<unsigned short, 32> randomBytes;
-        std::generate(randomBytes.begin(), randomBytes.end(), random);
-
-        std::string name;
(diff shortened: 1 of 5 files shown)


src/windows/wslcsession/WSLCVhdVolume.cpp
```diff
@@ -97,9 +81,13 @@ std::unique_ptr<WSLCVhdVolumeImpl> WSLCVhdVolumeImpl::Create(
...
 {
-    std::string name = (Name != nullptr && Name[0] != '\0') ? std::string(Name) : GenerateName();
     auto sizeBytes = ParseSizeBytes(DriverOpts);
-    auto hostPath = StoragePath / "volumes" / (name + ".vhdx");
+
+    GUID uuidGuid{};
+    THROW_IF_FAILED(CoCreateGuid(&uuidGuid));
+    auto uuid = wsl::shared::string::GuidToString<char>(uuidGuid, wsl::shared::string::GuidToStringFlags::None);
+
+    auto hostPath = StoragePath / "volumes" / (uuid + ".vhdx");
```
kvega005, line 90: I'm anticipating a comment about this. I agree saving the VHD with the volume name is nicer. We could go back to generating the name and just manually setting the docker "anonymous volume" label when we "generate" the name. I think, long term, we may want to use our own anonymous volume label anyway, so we don't have to rely on docker state tracking.

If you have opinions on this, please share.

kvega005: For now, I think the docker anonymous volume label is important. Otherwise prune and `rm <container> -v` won't work as expected.

PR #40237, opened 2026-04-17
Start containerd before dockerd

dabhattimsft:
Instead of letting dockerd manage its own embedded containerd, this starts containerd as a separate process first and points dockerd at its socket (--containerd /run/containerd/containerd.sock). My testing shows ~1s improvement with this.

## PR Checklist

## Validation Steps Performed
Tests
Ran 100 iterations of below for measurements

>  sc stop wslservice
>  wslc container run --rm ubuntu:22.04 echo "hello"


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -383,9 +396,56 @@ try
...
+    {
+        return;
+    }
+
+    std::string entry = {buffer.begin(), buffer.end()};
+    WSL_LOG("ContainerdProcessLog", TraceLoggingValue(entry.c_str(), "Content"), TraceLoggingValue(m_displayName.c_str(), "Name"));
+}
+CATCH_LOG();
+
+void WSLCSession::StartContainerd()
```
OneBlue, line 424: nit: Could we factor this with StartDockerd() ? We could pass the command line as argument and return the process to make this easier

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -383,9 +396,56 @@ try
 }
 CATCH_LOG();
 
+void WSLCSession::OnContainerdLog(const gsl::span<char>& buffer)
```
OneBlue: nit: Could we factor this with `OnDockerdLog` ?  We could add an additional parameter for the prevenance of the logline

dabhattimsft: Fixed - merged into OnProcessLog(Buffer, Source). The Source parameter is passed as a TraceLogging field to distinguish containerd vs dockerd in traces. Call sites pass "containerd" or "dockerd".

src/windows/wslcsession/WSLCSession.cpp
dabhattimsft: Fixed - added StartProcess() helper that takes the executable, args, log source name, and exit callback. It handles ServiceProcessLauncher, stdout/stderr relay via OnProcessLog, and exit monitoring. StartContainerd() and StartDockerd() now just build their args and call the helper.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -360,52 +366,84 @@ void WSLCSession::OnDockerdExited()
...
         {
             m_containerdReadyEvent.SetEvent();
         }
     }
 }
 CATCH_LOG();
 
-void WSLCSession::StartDockerd()
+ServiceRunningProcess WSLCSession::StartProcess(
+    const std::string& Executable, const std::vector<std::string>& Args, PCSTR LogSource, std::function<void()> ExitCallback)
```
OneBlue: nit: ExitCallback should be an rvalue reference since we're moving it in this method

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -383,9 +396,56 @@ try
...
+{
+    if (buffer.empty())
+    {
+        return;
+    }
+
+    std::string entry = {buffer.begin(), buffer.end()};
+    WSL_LOG("ContainerdProcessLog", TraceLoggingValue(entry.c_str(), "Content"), TraceLoggingValue(m_displayName.c_str(), "Name"));
+}
+CATCH_LOG();
```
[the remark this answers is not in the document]
dabhattimsft: Kept "ContainerdLog" as the ETW event name, but added a Source field to distinguish between containred and dockerd logs.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1975,6 +1997,33 @@ HRESULT WSLCSession::PruneVolumes(const WSLCPruneVolumesOptions* /*Options*/, WS
     return E_NOTIMPL;
 }
 
+int WSLCSession::StopProcess(std::optional<ServiceRunningProcess>& Process, DWORD TerminateTimeoutMs, DWORD KillTimeoutMs)
```
OneBlue: nit: We could directly pass `ServiceRunningProcess&` here since we always check if the optional is set before calling this

OneBlue (approved):
LGTM, two minor comments

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -383,9 +396,56 @@ try
...
+    std::vector<std::string> args{"/usr/bin/containerd", "--address", c_containerdSocket};
+
+    if (WI_IsFlagSet(m_featureFlags, WslcFeatureFlagsDebug))
+    {
+        args.emplace_back("--log-level");
+        args.emplace_back("debug");
+    }
+
+    ServiceProcessLauncher launcher{"/usr/bin/containerd", args, {{"PATH=/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/sbin"}}};
+
```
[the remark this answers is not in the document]
dabhattimsft: Fixed - added --root /var/lib/docker/containerd/daemon and --state /run/docker/containerd/daemon to match what moby's embedded containerd supervisor uses.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2008,31 +2097,15 @@ try
...
+    auto dockerdExitCode = StopProcess(m_dockerdProcess, c_processTerminateTimeoutMs, c_processKillTimeoutMs);
+    WSL_LOG("DockerdExit", TraceLoggingValue(dockerdExitCode, "code"));
+    m_dockerdProcess.reset();
 
-        WSL_LOG("DockerdExit", TraceLoggingValue(exitCode, "code"));
-        m_dockerdProcess.reset();
-    }
+    auto containerdExitCode = StopProcess(m_containerdProcess, c_processTerminateTimeoutMs, c_processKillTimeoutMs);
+    WSL_LOG("ContainerdExit", TraceLoggingValue(containerdExitCode, "code"));
+    m_containerdProcess.reset();
```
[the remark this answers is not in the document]
dabhattimsft: Fixed

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1959,6 +2021,33 @@ HRESULT WSLCSession::PruneVolumes(const WSLCPruneVolumesOptions* /*Options*/, WS
...
+    try
+    {
+        return Process->Wait(TerminateTimeoutMs);
+    }
+    catch (...)
+    {
+        LOG_CAUGHT_EXCEPTION();
+        try
+        {
+            Process->Get().Signal(WSLCSignalSIGKILL);
```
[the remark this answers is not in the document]
dabhattimsft: fixed

src/windows/wslcsession/WSLCSession.cpp
dabhattimsft: fixed

src/windows/wslcsession/WSLCSession.cpp
dabhattimsft: fixed

src/windows/wslcsession/WSLCSession.h
```diff
@@ -161,9 +161,14 @@ class DECLSPEC_UUID("4877FEFC-4977-4929-A958-9F36AA1892A4") WSLCSession
     void ConfigureStorage(const WSLCSessionInitSettings& Settings, PSID UserSid);
     void Ext4Format(const std::string& Device);
     void OnContainerDeleted(const WSLCContainerImpl* Container);
-    void OnDockerdLog(const gsl::span<char>& Data);
+    void OnProcessLog(const gsl::span<char>& Data, PCSTR Source);
+    void OnContainerdExited();
     void OnDockerdExited();
+    ServiceRunningProcess StartProcess(
+        const std::string& Executable, const std::vector<std::string>& Args, PCSTR LogSource, std::function<void()>&& ExitCallback);
```
ptrivedi, line 168: Looks like the PCSTR LogSource gets captured into the lambdas for relaying IO. I guess when string literals are passed, it's okay because they ahve static lifetime. But I wonder if a std::string and capture by value is better here to avoid issues when a caller passes std::string::c_str()

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -362,52 +368,87 @@ void WSLCSession::OnDockerdExited()
...
-    std::string entry = {buffer.begin(), buffer.end()};
-    WSL_LOG("ContainerdLog", TraceLoggingValue(entry.c_str(), "Content"), TraceLoggingValue(m_displayName.c_str(), "Name"));
+    std::string entry = {Buffer.begin(), Buffer.end()};
+    WSL_LOG(
+        "ContainerdLog",
+        TraceLoggingValue(Source, "Source"),
+        TraceLoggingValue(entry.c_str(), "Content"),
+        TraceLoggingValue(m_displayName.c_str(), "Name"));
 
     if (!m_containerdReadyEvent.is_signaled())
```
ptrivedi: This event name is misleading now. Should be m_dockerdReadyEvent or similar

src/windows/wslcsession/WSLCSession.h
OneBlue: Yeah string literals are guaranteed to be alive forever. I have a preference for using those since using std::string will cause a memory allocation

PR #40238, opened 2026-04-18
CLI: Initial support for version format option

AmelBawa-msft:
- Initial support for version format option

## PR Checklist


AmelBawa-msft:
@copilot fix the VersionCommand_HasNoArguments test method in WSLCCLICommandUnitTests.cpp, in this PR I added one argument to the version command.

test/windows/wslc/e2e/WSLCE2EVersionTests.cpp
```diff
@@ -0,0 +1,101 @@
...
+    {
+        return L"Usage: wslc version [<options>]\r\n\r\n";
+    }
+
+    std::wstring GetAvailableOptions() const
+    {
+        std::wstringstream options;
+        options << L"The following options are available:\r\n"
+                << L"  --format   Output formatting (json or table) (Default: table)\r\n"
+                << L"  -h,--help  Shows help about the selected command\r\n"
```
dkbennett: This will be a test failure now that help flag has been changed to -?

src/windows/wslc/tasks/VersionTasks.cpp
```diff
@@ -0,0 +1,66 @@
...
+    {
+    case FormatType::Json:
+    {
+        auto json = ToJson(versionInfo, c_jsonPrettyPrintIndent);
+        PrintMessage(MultiByteToWide(json));
+        break;
+    }
+    case FormatType::Table:
+    {
+        PrintMessage(MultiByteToWide(versionInfo.ToString()));
```
dkbennett: Should use the table output format all the other tables use.

src/windows/wslc/services/VersionModel.h
```diff
@@ -0,0 +1,90 @@
...
+    std::string DXCore = DXCORE_VERSION;
+    std::string Windows = wsl::windows::common::helpers::GetWindowsVersionString();
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(ServerVersion, Kernel, WSLg, MSRDC, Direct3D, DXCore, Windows);
+
+    std::string ToString() const
+    {
+        std::stringstream ss;
+        ss << "Server:\n";
+        ss << WSLC_VERSION_ROW("  Linux kernel", Kernel);
```
dkbennett: These strings should be loc'd for table output in the CLI.

dkbennett (changes requested):
Overall structure and approach looks good, I only have issues with the table output, which looks like it needs loc for table format and should use the same table output as other commands. Also a note about the help change that will cause a test failure.

src/windows/wslc/tasks/VersionTasks.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+Abstract:
+
+    Implementation of version command related execution logic.
+
+--*/
+#include "precomp.h"
+#include "Argument.h"
+#include "ArgumentValidation.h"
+#include "ContainerModel.h"
```
ptrivedi: does ContainerModel.h need to be included here?

test/windows/wslc/e2e/WSLCE2EVersionTests.cpp
AmelBawa-msft:
```suggestion
                << L"  -?,--help  Shows help about the selected command\r\n"
```

src/windows/wslc/tasks/VersionTasks.cpp
AmelBawa-msft: Good call! I likely copied this file originally from ContainerTasks.cpp and forgot to update the header.

src/windows/wslc/tasks/VersionTasks.h
```diff
@@ -0,0 +1,22 @@
...
+
+Abstract:
+
+    Declaration of version command execution tasks.
+--*/
+#pragma once
+#include "CLIExecutionContext.h"
+#include "Task.h"
+
+using wsl::windows::wslc::execution::CLIExecutionContext;
```
ptrivedi: should probably move this 'using' statement inside of the namespace wsl::windows::wslc::task {}

src/windows/wslc/tasks/VersionTasks.cpp
AmelBawa-msft: @copilot remove the ContainerModel.h if it's not being used here

src/windows/wslc/tasks/VersionTasks.h
AmelBawa-msft: Oh good catch! I also likely missed that as it seems to be copy pasted from ContainerTasks.cpp, I prefer to avoid `using` statements in header files.

AmelBawa-msft: @copilot update this header file so we don't use `using wsl::windows::wslc::execution::CLIExecutionContext;` statement. Instead inline the namespace directly. Do the same for ContainerTasks.h and ImageTasks.h.

src/windows/wslc/tasks/VersionTasks.cpp
AmelBawa-msft: Good call! That's something we've discussed internally. For now we will just keep it as is until we have an alternative requirement.

src/windows/wslc/services/VersionModel.h
```diff
@@ -0,0 +1,90 @@
...
+Abstract:
+
+    This file contains the VersionModel definitions
+--*/
+
+#pragma once
+
+namespace wsl::windows::wslc::models {
+
+#define WSLC_VERSION_ROW(key, value) std::format("{:<20}{}\n", key ":", value)
```
ptrivedi, line 18: I see this being #undef'ed later. It is ripe for a constexpr lamba instead of a macro

test/windows/wslc/e2e/WSLCE2EVersionTests.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    WSLC_TEST_METHOD(WSLCE2E_Version_JsonFormat)
+    {
+        const auto result = RunWslc(L"version --format json");
+        result.Verify({.Stderr = L"", .ExitCode = 0});
+        VERIFY_IS_TRUE(result.Stdout.has_value());
+
+        const auto versionInfoFromJson = wsl::shared::FromJson<VersionInfo>(result.Stdout->c_str());
+        VERIFY_ARE_EQUAL(VersionInfo{}.Client.Version, versionInfoFromJson.Client.Version);
```
ptrivedi, line 61: Should verify server version here as well? 

What about some of the other fields under the client/server sections?

benhillis:

benhillis:
@AmelBawa-msft Cleaning up stale pull requests, if this is still needed please open a new pull request against the master branch

PR #40244, opened 2026-04-19
added greek translation

trlef19:
added greek strings for wsl


trlef19:

benhillis:
@craigloewen-msft - we should consider adding Greek to our localized languages. I think this will mean tweaking our Touchdown config.

benhillis:
@trlef19 - thanks for taking the time to do this. Our translations are generated by a service called "touchdown". We will look into adding Greek to the list.

Issue #12512, opened 2025-01-28
WSL takes up to 10 minutes to start


tomdebie:
### Windows Version

Microsoft Windows [Version 10.0.19045.5371]

### WSL Version

2.3.26.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

5.15.167.4-1

### Distro Version

Ubuntu 24.04.1 LTS (GNU/Linux 5.15.167.4-microsoft-standard-WSL2 x86_64)

### Repro Steps

In cmd start up wsl for the first time since boot or after wsl --shutdown.
So basically type wsl in cmd and wait.

### Expected Behavior

I expected that WSL would start up in a few seconds.

### Actual Behavior

Takes several minutes to start up and become responsive.

### Diagnostic Logs

I'm running an AMD R5 3600 on a B45 MSI mortar motherboard. I've enabled AMD-V in the bios.
Ran "wsl --install" to install wsl and selected the default Ubuntu distro as my default distro.
Also tried to reinstall everything according to [this gist](https://gist.github.com/4wk-/889b26043f519259ab60386ca13ba91b)

I've also noticed that both LxssManager and LxssManagerUser_2104d3 services remained stopped during the startup of wsl.
![Image](https://github.com/user-attachments/assets/2e561e1a-ee3b-4271-9048-774a7b2474dd)

Collecting the wsl logs gave me the following errors, but I still seem to have some logs.
[WslNetworkingLogs-2025-01-29_00-07-03.zip](https://github.com/user-attachments/files/18580508/WslNetworkingLogs-2025-01-29_00-07-03.zip)
[WslLogs-2025-01-29_00-07-08.zip](https://github.com/user-attachments/files/18580509/WslLogs-2025-01-29_00-07-08.zip)
![Image](https://github.com/user-attachments/assets/cd9a101d-1706-4806-974b-4c04146a257a)

This is my output from dmesg where you can clearly see that it's stuck waiting on something.
> tom@Tom:/mnt/c/Users/tomde$ dmesg
[    0.000000] Linux version 5.15.167.4-microsoft-standard-WSL2 (root@f9c826d3017f) (gcc (GCC) 11.2.0, GNU ld (GNU Binutils) 2.37) #1 SMP Tue Nov 5 00:21:55 UTC 2024
[...]

Issue #11067, opened 2024-01-24
WSL2: wsl.exe hangs for 10s of minutes before working


DanKaplanSES:
### Windows Version

Microsoft Windows [Version 10.0.19045.3930]

### WSL Version

2.0.14.0

### Are you using WSL 1 or WSL 2?

- [X] WSL 2
- [ ] WSL 1

### Kernel Version

I can't run wsl.exe --status right now because wsl is hanging

### Distro Version

I can't run lsb_release -r right now because wsl is hanging

### Other Software

wsltty, updated earlier today while experiencing this problem: https://github.com/mintty/wsltty/releases/tag/3.7.0.2

### Repro Steps

Unfortunately I can't reproduce this consistently. After a fresh restart, wsl.exe / Ubuntu usually behave as expected, but after hours of uptime, wsl.exe hangs.

Normally I don't wait for it to finish: it can take up to 30 minutes before a wsl.exe command completes, and it doesn't always result in resolving the problem when it does complete. Sometimes it errors, sometimes wsl.exe starts an ubuntu prompt.

### Expected Behavior

I would like wsl.exe to immediately respond with correct output. e.g., if I run `wsl.exe --status`, I would like to immediately see that response. 

If wsl.exe can't be run for some reason, I would _like it to fail fast_ instead of 30 minutes later. I've been waiting 20+ minutes for `wsl.exe --status` to complete as I write this up, but I will probably restart my computer before it finishes because it is blocking my ability to work.

### Actual Behavior

Occasionally, any `wsl.exe` command will hang for 10s of minutes before showing any output.

### Diagnostic Logs

1. [WslLogs-2024-01-23_15-16-07.zip](https://github.com/microsoft/WSL/files/14030937/WslLogs-2024-01-23_15-16-07.zip) These first logs prompted me to create this issue. I ran `wsl` in an administrator cmd.exe and it took a very long time to complete. I captured the log while it was hanging and then noticed it completed. This seemed unique enough to share, as I haven't noticed other issues saying it worked _eventually_ after waiting long enough: it opened up a wsl command prompt (2nd line of screenshot).

    ![image](https://github.com/microsoft/WSL/assets/373675/33bb416e-1d8f-45e1-abde-cd5122e02941)

    The first line occurred after I manually killed the wsl service in the task manager (shown in this screenshot) then ran `wsl`:

[...]

Issue #5845, opened 2020-09-02
Starting WSL 2 distribution hangs forever


pizzafox:
# Environment

```none
Windows build number: Microsoft Windows NT 10.0.20201.0
Your Distribution version: Ubuntu 18.04
Whether the issue is on WSL 2 and/or WSL 1: WSL 2
```

# Steps to reproduce

Open `wsl.exe`, `wt.exe`, or `ubuntu.exe`

Possible duplicate of #5836

## `wsl.exe`

Running `wsl` hangs forever, or at least the 10 minutes I waited.
Running `wsl --help` works properly.
Running `wsl -l` hangs for several minutes but then seems to properly output.

## `wt.exe` Windows Terminal

Opening Windows Terminal displays the error since my default profile of Ubuntu isn't working:

> Could not find your default profile in your list of profiles - using the first profile.
> Check to make sure the "defaultProfile" matches the GUID of one of your profiles.

**WSL logs**: Feedback app hangs, can't collect anything.

#  Expected behavior

WSL opens.

# Actual behavior

WSL doesn't open. See 'Steps to reproduce'.

pizzafox:
Turning off Hyper-V like #5836 said fixed the issue.

pizzafox:
Had my PC BSoD and restart a bunch, same issue occurs. Hyper-V is still off

licanhua:
Are you windows running on physical machine or Windows itself is a virtual machine? I hit the very slow response yesterday because I only allocated 1G memory and limited disk space to Windows.

Try disable swap file to see if it improve the time in your case.
https://github.com/microsoft/WSL/issues/5836#issuecomment-685809778

pizzafox:
I am running Windows standard with several GB of RAM free.

MakuZo:
The same behavior on
```
Platform ServicePack Version      VersionString
-------- ----------- -------      -------------
 Win32NT             10.0.20201.0 Microsoft Windows NT 10.0.20201.0
```
Also constant BSoDs on system shutdown.

afromankenobi:
Exactly same issue here!

afromankenobi:
Someone knows how to fix this issue? Neither disabling HyperV nor write `swap=0` fixed the issue for me. I have the BSoD on shutdown too. The last Insider's update seems to create this issue.

afromankenobi:
> 
> 
[...]

PR #40246, opened 2026-04-20
Add per-call deadline around HcnCreateNetwork to prevent startup hangs

shuaiyuanxx:

  `HcnCreateNetwork` has no per-call timeout. When HNS is slow to start (see #14072, #12512), the RPC blocks inside VM creation for minutes, producing user-visible WSL startup hangs and cloud telemetry `CreateInstance → timeout`. The existing `RetryWithTimeout(..., 3s)` wrapper is a **retry budget**, not a call deadline — a single hanging call bypasses it entirely.

  ## PR Checklist

  - [x] **Closes:** related to #5845, #14072, #12512, #11067 (HNS-caused slow WSL startup)
  - [ ] **Communication:** N/A — contained fix, no behavior change on happy path
  - [x] **Tests:** before/after demo under `test/retry-timeout-demo/`
  - [x] **Localization:** no new user-facing strings (reuses existing `MessageNetworkInitializationFailedFallback2`)
  - [ ] **Dev docs:** N/A

  ## Detailed Description

  - Adds `wsl::shared::retry::CallWithDeadline` in `retryshared.h` — uses `packaged_task` + detached worker so `wait_for(deadline)` enforces a true single-call deadline. The existing `RetryWithTimeout` is unchanged (kept for the 14 callers that want retry-budget semantics); a comment there warns about its real semantics.
  - `NatNetworking::CreateNetwork` wraps the `HcnCreateNetwork` RPC with a 30s deadline. Captures are by value so the detached worker is self-contained.
  - On `DeadlineExceededError`, emits new telemetry event `HcnCreateNetworkDeadlineExceeded` and returns an empty network.  `WslCoreVm.cpp:553-560` already handles the empty-network case by falling back to `VirtioProxy` — no changes needed there.
  - End-to-end effect: "HNS slow" sessions go from ~3 min silent hang (bucketed as `timeout` in telemetry) to ~35s with a single warning line and a working `VirtioProxy` network (bucketed as `complete`).

  ## Validation Steps Performed

### DIFF
--- a/src/shared/inc/retryshared.h
+++ b/src/shared/inc/retryshared.h
@@ -70,8 +70,8 @@ public:
 //   capturing inputs by value and returning outputs via the result type.
 // - `routine` must eventually return. This helper gives up waiting; it does
 //   not and cannot cancel a native call that is truly deadlocked.
-template <typename F>
-auto CallWithDeadline(std::chrono::milliseconds deadline, F routine) -> std::invoke_result_t<F>
+template <typename TRep, typename TPeriod, typename F>
+auto CallWithDeadline(std::chrono::duration<TRep, TPeriod> deadline, F routine) -> std::invoke_result_t<F>
 {
     using R = std::invoke_result_t<F>;
     std::packaged_task<R()> task(std::move(routine));
--- a/src/shared/inc/retryshared.h
+++ b/src/shared/inc/retryshared.h
@@ -54,7 +54,9 @@ T RetryWithTimeout(const std::function<T()>& routine, TPeriod retryPeriod, TTime
 class DeadlineExceededError : public std::runtime_error
 {
 public:
-    DeadlineExceededError() : std::runtime_error("call deadline exceeded") {}
+    DeadlineExceededError() : std::runtime_error("call deadline exceeded")
+    {
+    }
 };
 
 // Enforces a true per-call deadline. If `routine` does not complete within
--- a/src/shared/inc/retryshared.h
+++ b/src/shared/inc/retryshared.h
@@ -14,10 +14,21 @@ Abstract:
 
 #pragma once
 
+#include <chrono>
+#include <functional>
+#include <future>
+#include <stdexcept>
+#include <thread>
+#include <type_traits>
+#include <utility>
+
 namespace wsl::shared::retry {
 
 constexpr auto AlwaysRetry = []() { return true; };
 
+// NOTE: despite its name, `timeout` here is a RETRY BUDGET, not a per-call
+// deadline. If `routine` hangs instead of throwing, `timeout` is never
+// checked. For true per-call deadline semantics, use CallWithDeadline below.
 template <typename T, typename TPeriod, typename TTimeout>
 T RetryWithTimeout(const std::function<T()>& routine, TPeriod retryPeriod, TTimeout timeout, const std::function<bool()>& retryPred = AlwaysRetry)
 {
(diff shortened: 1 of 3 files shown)


OneBlue (changes requested):
I agree that it would be nice to have a deadline when we make calls like that one, but unfortunately, I don't think that we can safely detach a thread that's doing an outbound COM call. 

For instance when the service shuts down, we uninitialize COM (see `WslService::ServiceStopped()`), which will lead to undefined behavior or hang if we still have a thread in a COM call. 

This is a real issue though, to fix this, I think we should: 

1) Figure out why some HNS call take that long, and solve that at the HNS level 

2) Look into other ways we could "abort" the call. Under the hood, HcnCreateNetwork calls a COM method, so we could potentially make the COM call ourselves, and use [CoCancelCall](https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-cocancelcall) to safely cancel if we hit the timeout 

Adding @keith-horton for thoughts on this

keith-horton:
@OneBlue , CoCancelCall is reasonable if HNS is blocked on itself or vswitch. What is the proposal to try if the timeout expires?

OneBlue:
> What is the proposal to try if the timeout expires?

I think we should fail if we don't hear back from HNS after the timeout

keith-horton:
That sounds fine to me.

shuaiyuanxx:
Hi @OneBlue Thanks for the review. I did a spike on `CoCancelCall` with HNS suspended (via `NtSuspendProcess`). Results:

```
CoCancelCall(tid, 0)            -> RPC_E_NO_CONTEXT (0x8001011E)
CoCancelCall(tid, 5000ms)       -> RPC_E_NO_CONTEXT
RpcCancelThread(handle)         -> RPC_S_OK, but HcnCreateNetwork kept blocking
RpcCancelThreadEx(handle, 5s)   -> RPC_S_OK, but HcnCreateNetwork kept blocking
```

`HcnCreateNetwork` only returned (with `S_OK`) when HNS was resumed ~24s later. So in-process cancellation isn't reachable: the HCN RPC binding ignores cancel requests even when the RPC runtime accepts them.

Given that, I'd like to rework this PR along these lines instead:

- Revert the current PR entirely
- Start a threadpool timer before calling `HcnCreateNetwork`.
- If the call exceeds a threshold (e.g. 10s), fire a new telemetry event `HcnCreateNetworkSlow` and record a user-visible warning.

This gives us actionable data for triaging HNS slow-start at its source, without touching COM lifetime. Does that direction work for you before I refactor?

OneBlue:
> CoCancelCall(tid, 0)            -> RPC_E_NO_CONTEXT (0x8001011E)

@shuaiyuanxx: Did you also call `CoEnableCallCancellation` on the thread that does the COM call ? That error message seems to indicate that the cancellation didn't actually happen

shuaiyuanxx:
> > CoCancelCall(tid, 0)            -> RPC_E_NO_CONTEXT (0x8001011E)
> 
> @shuaiyuanxx: Did you also call `CoEnableCallCancellation` on the thread that does the COM call ? That error message seems to indicate that the cancellation didn't actually happen

Yes, `CoEnableCallCancellation` was called on the calling thread, verified with `CoGetApartmentType=MTA` and a nested call also returning S_OK.

So HCN is neither visible to OLE call context, nor does its RPC binding honor cancellation. `computenetwork.h` also has no `HcnCreateNetworkAsync` — only sync entry points

PR #40249, opened 2026-04-20
fix: implement IDisposable on WslConfigService to dispose FileSystemWatcher

benhillis:
## Problem

\\WslConfigService\\ creates a \\FileSystemWatcher\\ but never disposes it. The existing finalizer only frees unmanaged WslConfig objects — the managed \\FileSystemWatcher\\ and its OS handles leak until the GC finalizes them (which is non-deterministic and not guaranteed).

## Fix

Implement the standard \\IDisposable\\ pattern:
- Class now implements \\IDisposable\\
- \\Dispose(bool)\\ handles both managed (\\FileSystemWatcher\\) and unmanaged (\\FreeWslConfig\\) cleanup
- Finalizer calls \\Dispose(false)\\ (skips managed objects)
- \\Dispose()\\ calls \\Dispose(true)\\ + \\GC.SuppressFinalize(this)\\

This follows the [standard .NET dispose pattern](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) and ensures deterministic cleanup of the file watcher's OS handles.

### DIFF
--- a/src/windows/wslsettings/Services/WslConfigService.cs
+++ b/src/windows/wslsettings/Services/WslConfigService.cs
@@ -11,7 +11,6 @@ public class WslConfigService : IWslConfigService, IDisposable
     private WslConfig? _wslConfigDefaults { get; init; }
     private readonly object? _wslCoreConfigInterfaceLockObj = null;
     private FileSystemWatcher? _wslConfigFileSystemWatcher = null;
-    private bool _disposed = false;
 
     public WslConfigService()
     {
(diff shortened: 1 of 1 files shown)


benhillis:
Addressed the synchronization feedback — Dispose(bool) now takes _wslCoreConfigInterfaceLockObj before touching the watcher or config objects. Removed the \_wslConfig = null\ / \_wslConfigDefaults = null\ assignments since \_wslConfigDefaults\ has \{ get; init; }\ (can't assign outside constructor). Also added unsubscribe for \Deleted\ and \Renamed\ handlers. Force-pushed the fix — initial build was failing due to the init-only property assignment.

src/windows/wslsettings/Services/WslConfigService.cs
```diff
@@ -1,205 +1,240 @@
...
+    protected virtual void Dispose(bool disposing)
+    {
+        lock (_wslCoreConfigInterfaceLockObj!)
+        {
+            if (_disposed)
+            {
+                return;
+            }
+
+            _disposed = true;
```
OneBlue: If `Dispose(false)` is called, we'll set `_disposed = true`, and `Dispose(true)` is called later, we won't actually delete `_wslConfigFileSystemWatcher `. We could just use `_wslConfigFileSystemWatcher == null` is our test to see if it needs to be disposed

benhillis: Good point — dropped the \_disposed\ flag entirely and now use \_wslConfigFileSystemWatcher == null\ as the guard. Simpler and avoids the ordering issue you described. Force-pushed.

OneBlue (approved):

PR #40250, opened 2026-04-20
wslc: add network inspect

beena352:
Adds the InspectNetwork method to the WSLC session, allowing callers to retrieve detailed information about a managed network (name, ID, driver, scope, IPAM config, labels). 

## PR Checklist

- Added InspectNetwork to the IDL interface, returning a JSON string with network details
- Created dedicated InspectIPAMConfig, InspectIPAM, and InspectNetwork structs in wslc_schema.h (keeps it independent from docker_schema)
- Added InspectNetwork to DockerHTTPClient calling GET /networks/{name}

## Validation Steps Performed

- NetworkInspectTest - creates a network, inspects it, verifies name, driver, ID, and internal flag
- NetworkInspectWithSubnetTest - creates a network with a subnet, inspects it, verifies IPAM config and subnet value
- NetworkInspectNotFoundTest - verifies WSLC_E_NETWORK_NOT_FOUND is returned with correct error message for a non-existent network

### DIFF
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -54,7 +54,7 @@ std::string IndentLines(const std::string& input, const std::string& prefix)
     return result;
 }
 
-void ValidateName(LPCSTR Name, size_t maxLength = WSLC_MAX_CONTAINER_NAME_LENGTH)
+void ValidateName(LPCSTR Name, size_t maxLength)
 {
     const auto& locale = std::locale::classic();
     size_t i = 0;
@@ -1536,7 +1536,7 @@ try
     // Validate that name & images are valid.
     if (containerOptions->Name != nullptr)
     {
-        ValidateName(containerOptions->Name);
+        ValidateName(containerOptions->Name, WSLC_MAX_CONTAINER_NAME_LENGTH);
     }
 
     RETURN_HR_IF(E_INVALIDARG, strlen(containerOptions->Image) > WSLC_MAX_IMAGE_NAME_LENGTH);
@@ -1581,7 +1581,7 @@ try
 {
     COMServiceExecutionContext context;
 
-    ValidateName(Id);
+    ValidateName(Id, WSLC_MAX_CONTAINER_NAME_LENGTH);
 
     // Look for an exact ID match first.
     auto lock = m_lock.lock_shared();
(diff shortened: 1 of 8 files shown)


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -54,7 +54,7 @@ std::string IndentLines(const std::string& input, const std::string& prefix)
     return result;
 }
 
-void ValidateName(LPCSTR Name)
+void ValidateName(LPCSTR Name, size_t maxLength = WSLC_MAX_CONTAINER_NAME_LENGTH)
```
OneBlue: nit: I recommend making this a required argument to avoid potential mismatches in the future

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2122,6 +2122,68 @@ try
...
+    std::string name = Name;
+    ValidateName(name.c_str(), WSLC_MAX_NETWORK_NAME_LENGTH);
+
+    auto lock = m_lock.lock_shared();
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_dockerClient);
+
+    {
+        std::lock_guard networksLock(m_networksLock);
+
+        auto it = m_networks.find(name);
```
OneBlue: Given that we are keeping track of all the networks that manage, I think we should do something similar to what we do with volumes and load the properties of netwoks when we create / list them, that way we can return them without needing to call docker again 

(this also has the benefit of preventing someone from deleting the network between our name lookup and docker API call)

test/windows/WSLCTests.cpp
```diff
@@ -3901,6 +3901,69 @@ class WSLCTests
...
+
+        auto inspect = wsl::shared::FromJson<wsl::windows::common::wslc_schema::InspectNetwork>(output.get());
+        VERIFY_ARE_EQUAL(inspect.Name, networkName);
+        VERIFY_ARE_EQUAL(inspect.Driver, std::string("bridge"));
+        VERIFY_IS_TRUE(inspect.IPAM.Config.has_value());
+        VERIFY_ARE_EQUAL(1u, inspect.IPAM.Config->size());
+        VERIFY_ARE_EQUAL(std::string("172.30.0.0/16"), inspect.IPAM.Config->at(0).Subnet);
+    }
+
+    WSLC_TEST_METHOD(NetworkInspectNotFoundTest)
```
OneBlue, line 3959: nit: I recommend adding test coverage for: 

- Invalid / unsupported drivers
- Invalid gateway Ip addresses
- Invalid subnets

OneBlue (approved):

test/windows/WSLCTests.cpp
beena352: Sounds good, I’ll add this in a follow-up PR.

PR #40266, opened 2026-04-21
SDK Install function implementation

JohnMcPMS:
Implements `WslcInstallWithDependencies`.

## PR Checklist

Implements `WslcInstallWithDependencies` using existing functions to determine which components are required and to install the Virtual Machine component.  Adds implementation of WU API installation flow.

For testing purposes, the WU API implementation takes in an object factory and is broken into the largely independent steps of searching, downloading, and installing the WSL package.  The actual use leverages a convenience method that implements the overall flow.

Progress is reported using a 70/30 split for the download/install phases.  This is largely arbitrary and is driven by singly defined values that can be easily updated if it feels off.

Automated testing is done exclusively through mocks, as we have not yet published a publicly available package and we would need to build out the test infrastructure to launch and control a child VM to prevent interference with WSL under test.

Manual testing was able to successfully install both components, although a reboot is required in order for the Virtual Machine Platform to be available.  Afterward I was able to remove only WSL and successfully reinstall just it with the API.

## Validation Steps Performed
Manual validation of functionality using internal update testing infrastructure.
Automated unit tests leveraging WU API mocks.


src/windows/common/WindowsUpdateIntegration.cpp
```diff
@@ -0,0 +1,329 @@
...
+{
+    LONG result{};
+    if (m_updates)
+    {
+        THROW_IF_FAILED(m_updates->get_Count(&result));
+    }
+    return static_cast<size_t>(result);
+}
+
+void WindowsUpdateContext::DownloadUpdates(std::function<void(uint32_t)> progress) const
```
OneBlue: nit: `progress` could be taken by const reference here. 

We can also do this: 
```
const std::function<void(uint32_t)>& progress = [](auto){}
```

To get a method that does nothing by default so we don't have to test if `progress` is null before calling it

src/windows/common/WindowsUpdateIntegration.cpp
```diff
@@ -0,0 +1,329 @@
...
+    This file contains objects related to invoking the Windows Update Agent API.
+
+--*/
+
+#include "precomp.h"
+#include "WindowsUpdateIntegration.h"
+
+namespace wsl::windows::common {
+
+namespace anon {
```
OneBlue, line 20: nit: We can just do: `namespace {` if we want an anonymous namespace

src/windows/common/WindowsUpdateIntegration.cpp
```diff
@@ -0,0 +1,329 @@
...
+    THROW_IF_FAILED(m_searcher->Search(queryBSTR.get(), &searchResult));
+
+    OperationResultCode resultCode{};
+    THROW_IF_FAILED(searchResult->get_ResultCode(&resultCode));
+
+    THROW_HR_IF(WSLC_E_WU_SEARCH_FAILED, resultCode != OperationResultCode::orcSucceeded && resultCode != OperationResultCode::orcSucceededWithErrors);
+
+    if (resultCode == OperationResultCode::orcSucceededWithErrors)
+    {
+        // TODO: Consider logging from the warnings
```
OneBlue: Agreed, probably a good idea to logs the output of `get_Warnings` here as well

src/windows/common/WindowsUpdateIntegration.cpp
OneBlue: (Same for below)

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -1503,9 +1506,54 @@ CATCH_RETURN();
...
+    bool needsRuntime = NeedsWslRuntimeInstalled();
+
+    if (!needsVirtualMachine && !needsRuntime)
+    {
+        return S_OK;
+    }
+
+    // Installing these components requires elevation.
+    auto token = wil::open_current_access_token();
+    RETURN_HR_IF(E_ACCESSDENIED, !wsl::windows::common::security::IsTokenElevated(token.get()));
```
OneBlue: I recommend `ERROR_ELEVATION_REQUIRED` to make it clear that the lack of elevation is the issue here

OneBlue: We might also want to let the call through if `wsl::windows::common::security::IsTokenLocalSystem(token.get())` is true, since `IsTokenElevated()` can return false for that specific case

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -1503,9 +1506,54 @@ CATCH_RETURN();
...
+            callback = [progressCallback, context](uint32_t progress) {
+                progressCallback(WSLC_COMPONENT_FLAG_WSL_PACKAGE, progress, 100, context);
+            };
+        }
+
+        wsl::windows::common::WindowsUpdateContext wuContext;
+        wuContext.RunUpdateFlow(true, callback);
+    }
+
+    return S_OK;
```
OneBlue: I recommend returning a different error code if a reboot is required (probably ERROR_SUCCESS_REBOOT_REQUIRED) so the caller doesn't try to call create a session right away of a reboot is required for HCS to become available

OneBlue (commented):
LGTM, minor comments

src/windows/common/WindowsUpdateIntegration.cpp
JohnMcPMS: It is still a public method and the caller could provide an empty function.

src/windows/common/WindowsUpdateIntegration.cpp
JohnMcPMS: My experience with debugging anonymous namespace types through interfaces is that it is terrible.  Having a named namespace resolves the issue.  These are probably less bad than I have had to deal with in other instances, but I would rather have the name and not need it.

src/windows/common/WindowsUpdateIntegration.cpp
```diff
@@ -0,0 +1,360 @@
...
+                    {
+                        continue;
+                    }
+
+                    wil::unique_bstr message;
+                    LONG hr{};
+                    UpdateExceptionContext context{};
+                    warning->get_Message(&message);
+                    warning->get_HResult(&hr);
+                    warning->get_Context(&context);
```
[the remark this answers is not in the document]
JohnMcPMS: > This won’t compile

I'm offended at the insinuation of pushing without even building locally.

> Use message.put()

Apparently, you didn't look at the implementation of `wil::unique_any_t::operator&`.

> ignores the HRESULTs

Yes, but I would still want the event if any of them succeed, and I'm confident that `get_HResult` will either work or crash the process.

OneBlue (approved):

PR #40268, opened 2026-04-22
Update settings defaults and warn on unknown keys

yao-msft:
All settings accept "default" string value to indicate using defaults.
Default cpu count is all cpu.
Default memory is half of available.
Default max storage size is 1TB.
Also warn on unknown keys.

## PR Checklist

## Validation Steps Performed
Tests added. Manually tested.


test/windows/wslc/WSLCCLISettingsUnitTests.cpp
```diff
@@ -155,19 +155,34 @@ class WSLCCLISettingsUnitTests
...
         WriteFile(dir / L"settings.yaml", "");
 
         UserSettingsTest s{dir};
 
-        VERIFY_ARE_EQUAL(0u, s.GetWarnings().size());
-        VERIFY_ARE_EQUAL(4u, s.Get<Setting::SessionCpuCount>());
-        VERIFY_ARE_EQUAL(2048u, s.Get<Setting::SessionMemoryMb>());
-        VERIFY_ARE_EQUAL(102400u, s.Get<Setting::SessionStorageSizeMb>());
+        VERIFY_ARE_EQUAL(static_cast<int>(UserSettingsType::Standard), static_cast<int>(s.GetType()));
+        VERIFY_IS_TRUE(s.GetWarnings().size() >= 1u);
```
OneBlue: nit: I recommend validating the warning string

OneBlue: (Same for below)

src/windows/common/WSLCUserSettings.cpp
```diff
@@ -35,21 +35,22 @@ namespace wsl::windows::wslc::settings {
...
-    "  # Maximum disk image size in megabytes (default: 100GB)\n"
-    "  # maxStorageSize: 100GB\n"
+    "  # Maximum disk image size (default: 1TB)\n"
+    "  # maxStorageSize: 1TB\n"
     "\n"
     "  # Default path for session storage. By default, storage is per-session under:\n"
-    "  #   %LocalAppData%\\wslc\\sessions\\wslc-cli        (standard sessions)\n"
-    "  #   %LocalAppData%\\wslc\\sessions\\wslc-cli-admin (elevated sessions)\n"
-    "  # defaultStoragePath: \"\"\n"
+    "  #   %LocalAppData%\\wslc\\sessions\\wslc-cli-<user>        (standard sessions)\n"
```
OneBlue: This is something that we should probably discuss in a followup PR, but I think we should remove the ability for users to override these paths, because if they change add after the storage is created, everything will be gone. 

My opinion would be to always use `%LocalAppData%\\wslc\\sessions`, and if the user really wants the VHD to go somewhere else, they can symlink `%LocalAppData%\\wslc\\sessions` to wherever they want

src/windows/common/WSLCUserSettings.cpp
```diff
@@ -35,21 +35,22 @@ namespace wsl::windows::wslc::settings {
...
     "\n"
-    "  # Memory limit for the session in megabytes (default: 2GB)\n"
-    "  # memorySize: 2GB\n"
+    "  # Memory limit for the session (e.g. 2GB default: half of available memory)\n"
+    "  # memorySize: default\n"
     "\n"
-    "  # Maximum disk image size in megabytes (default: 100GB)\n"
-    "  # maxStorageSize: 100GB\n"
+    "  # Maximum disk image size (default: 1TB)\n"
+    "  # maxStorageSize: 1TB\n"
```
OneBlue: I think we should use `default` here as well, that way we can still change the default later, even if the config file has been generated

src/windows/common/WSLCUserSettings.cpp
```diff
@@ -216,6 +218,82 @@ namespace {
...
+            {
+                std::string key;
+                try
+                {
+                    key = it->first.as<std::string>();
+                }
+                catch (...)
+                {
+                    auto location = prefix.empty() ? std::wstring(L"root") : MultiByteToWide(prefix);
+                    warnings.push_back({std::format(L"Warning: Non-string key in section '{}'.", location), location});
```
OneBlue, line 267: Sorry I'm just noticing this now: We should used localized strings for these warnings

OneBlue: I also recommend logging the line number of the yaml node to make this easier on the user (looks like we can use `GetMark()` for that

OneBlue (approved):
LGTM, one minor comment

src/windows/common/WSLCUserSettings.cpp
yao-msft: I'll do the settings localization as a whole in a followup pr.

PR #40269, opened 2026-04-22
Emit telemetry when create-instance steps exceed 10s

shuaiyuanxx:

Adds `SlowOperationWatcher`, an RAII guard that emits a `SlowOperation` telemetry event when a scoped step exceeds 10 s, so slow `CreateInstance` starts can be attributed to a specific phase (HCS op, kernel boot, init-daemon wait, plugin hook) instead of the opaque timeout bucket.

Fast path emits nothing. Slow path emits one `SlowOperation` event carrying `name`, `thresholdMs`, and the call site (`file`, `function`, `line`) captured via `std::source_location`.

## PR Checklist

## Detailed Description

`src/windows/common/SlowOperationWatcher.{h,cpp}` is a single-shot threadpool timer guard. `wil::unique_threadpool_timer`'s destroyer cancels and drains on scope exit, so the callback cannot touch `*this` after destruction. `Reset()` disarms early for cases where the scope must outlive the watched call (e.g. to keep a `ReceiveMessage` buffer reference alive). `Name` is restricted to `const char (&)[N]` to enforce static storage.

12 call sites wrapped:

- `WslCoreVm.cpp`: `HcsCreateSystem`, `HcsStartSystem`, `WaitForMiniInitConnect`, `ReadGuestCapabilities`, `CreateNatNetwork`, `AttachDistroVhd`, `WaitForInitDaemonConnect`
- `WslCoreInstance.cpp`: `WaitForCreateInstanceResult`, `WaitForDrvFsInit`, `WaitForInitConfigResponse`
- `PluginManager.cpp`: `PluginOnVmStarted`, `PluginOnDistributionStarted`

Call-site business logic is unchanged. No user-facing path.

## Validation Steps Performed

Validated end-to-end against a Debug `wslservice.exe` with ETW capture:

- [x] **Fast path**: `wsl --shutdown` + `wsl -d Ubuntu -e echo hello` — all guarded sites exercised, **zero** `SlowOperation` events.
- [x] **Slow path**: injected `Sleep(11000)` inside `WaitForMiniInitConnect` — one `SlowOperation{name="WaitForMiniInitConnect", thresholdMs=10000, file, function, line}` event emitted.


src/windows/common/WslSlowOperation.h
```diff
@@ -0,0 +1,53 @@
...
+
+    WslSlowOperation(const WslSlowOperation&) = delete;
+    WslSlowOperation& operator=(const WslSlowOperation&) = delete;
+    WslSlowOperation(WslSlowOperation&&) = delete;
+    WslSlowOperation& operator=(WslSlowOperation&&) = delete;
+
+private:
+    static void CALLBACK OnTimerFired(PTP_CALLBACK_INSTANCE, PVOID Context, PTP_TIMER) noexcept;
+
+    const char* const m_name;
```
OneBlue: nit: I recommend always default initializing pointers to nullptr

src/windows/common/WslSlowOperation.cpp
```diff
@@ -0,0 +1,70 @@
...
+    ULARGE_INTEGER due{};
+    due.QuadPart = static_cast<ULONGLONG>(-(Relative.count() * 10'000LL));
+    FILETIME ft{};
+    ft.dwLowDateTime = due.LowPart;
+    ft.dwHighDateTime = due.HighPart;
+    return ft;
+}
+} // namespace
+
+WslSlowOperation::WslSlowOperation(_In_z_ const char* Name, std::chrono::milliseconds SlowThreshold) :
```
OneBlue: I think it would be a good idea to capture the source of the call (via std::source_location) and log it to make debugging easier

src/windows/common/WslSlowOperation.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    FILETIME due = RelativeFileTime(m_slowThreshold);
+    SetThreadpoolTimer(m_timer.get(), &due, 0, 0);
+}
+
+WslSlowOperation::~WslSlowOperation() noexcept
+{
+    // Cancel any pending fire and block until an in-flight callback drains. This is the
+    // guarantee that OnTimerFired will not dereference `this` after destruction.
+    if (m_timer)
```
OneBlue: I don't think this needs to be done explicitely (wil::unique_threadpool_timer does this for us)

src/windows/common/WslSlowOperation.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    auto* self = static_cast<WslSlowOperation*>(Context);
+
+    WSL_LOG_TELEMETRY(
+        "SlowOperation",
+        PDT_ProductAndServicePerformance,
+        TraceLoggingValue(self->m_name, "name"),
+        TraceLoggingInt64(self->m_slowThreshold.count(), "thresholdMs"));
+
+    WSL_LOG(
```
OneBlue: WSL_LOG_TELEMTRY also logs to ETL so we shouldn't need both

src/windows/common/WslSlowOperation.h
```diff
@@ -0,0 +1,53 @@
...
+
+--*/
+
+#pragma once
+
+#include <windows.h>
+#include <wil/resource.h>
+#include <chrono>
+
+class WslSlowOperation
```
OneBlue: nit: I'd recommend removing `Wsl` from the name, maybe something like `SlowOperationWatcher`  ?

src/windows/common/WslSlowOperation.h
```diff
@@ -0,0 +1,53 @@
...
+{
+public:
+    // Name must have static storage duration (string literal). It is emitted verbatim into
+    // telemetry and log, so keep it a short CamelCase phase identifier that the backend
+    // query can switch on (e.g. "WaitForMiniInitConnect").
+    explicit WslSlowOperation(_In_z_ const char* Name, std::chrono::milliseconds SlowThreshold = std::chrono::seconds{10});
+
+    ~WslSlowOperation() noexcept;
+
+    WslSlowOperation(const WslSlowOperation&) = delete;
```
OneBlue: I recommend adding a `Reset()` method to disarms the callback, this will make it easier to deal with scoped variables like this for instance: 

```
const LX_MINI_INIT_CREATE_INSTANCE_RESULT* resultPtr = nullptr;
{
        WslSlowOperation slowOperation{"WaitForCreateInstanceResult"};
        resultPtr = &m_initChannel->GetChannel().ReceiveMessage<LX_MINI_INIT_CREATE_INSTANCE_RESULT>(&span, m_socketTimeout);
 }
```

Can just become: 

```
WslSlowOperation slowOperation{"WaitForCreateInstanceResult"};
auto result = &m_initChannel->GetChannel().ReceiveMessage<LX_MINI_INIT_CREATE_INSTANCE_RESULT>(&span, m_socketTimeout);

slowOperation.Reset();
```

src/windows/common/SlowOperationWatcher.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+    // Negative FILETIME means "relative to now", in 100ns units. Matches the pattern used
+    // elsewhere in the service (see Lifetime.cpp).
+    return wil::filetime::from_int64(-wil::filetime_duration::one_millisecond * Relative.count());
+}
+
+// std::source_location::file_name() returns the path as the compiler saw it, which on
+// MSVC is an absolute build-agent path. Strip to the basename so telemetry groups the
+// same file across different build environments without leaking machine-specific paths.
+const char* Basename(const char* Path) noexcept
```
OneBlue: nit: given that the file names are known compile-time, I recommend using a constexpr function for that. Something like: 

```
constexpr std::string_view Basename(std::string_view path)
{
    size_t pos = path.find_last_of("\\/");
    return pos == std::string_view::npos ? path : path.substr(pos + 1);
}

static_assert(Basename("/foo/test.cpp") == "test.cpp");
```

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -1217,8 +1237,11 @@ std::shared_ptr<LxssRunningInstance> WslCoreVm::CreateInstance(
     message.WriteString(message->SharedMemoryRootOffset, sharedMemoryRoot);
     message.WriteString(message->InstallPathOffset, installPath);
     message.WriteString(message->UserProfileOffset, userProfile);
-    auto transaction = m_miniInitChannel.StartTransaction();
-    transaction.Send<LX_MINI_INIT_MESSAGE>(message.Span());
+    {
+        SlowOperationWatcher slowOperation{"SendLaunchInit"};
```
OneBlue: I'm not sure that a watcher makes sense here. We're only sending an hvsocket message, we're not actually waiting for a reply here

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -660,7 +675,10 @@ void WslCoreVm::Initialize(const GUID& VmId, const wil::shared_handle& UserToken
     }
 
     // Perform additional initialization.
-    InitializeGuest();
+    {
+        SlowOperationWatcher slowOperation{"InitializeGuest"};
```
OneBlue: I don't think we should have watcher here. There's a lot that can happen in InitializeGuest() and we already log the VM boot time. If we wanted to catch long running operations inside that method, I think we should target specific operations instead of a big set of operations like `InitializeGuest`

OneBlue: A lot of operations inside that method will actually be caught with this change already

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -532,6 +543,7 @@ void WslCoreVm::Initialize(const GUID& VmId, const wil::shared_handle& UserToken
 
     {
         ExecutionContext context(Context::ConfigureNetworking);
+        SlowOperationWatcher slowOperation{"ConfigureNetworking"};
```
OneBlue: I also don't recommend a watcher here, since sub-operations are already covered and ConfigureNetworking() is a big set of operations

shuaiyuanxx:
Hi @OneBlue, Thanks for the thorough review! All of your suggestions have been addressed — PTAL when you have a moment.

OneBlue (approved):
LGTM, sorry for the delay in review !

shuaiyuanxx:
@copilot resolve the merge conflicts in this pull request

src/windows/common/SlowOperationWatcher.cpp
```diff
@@ -0,0 +1,76 @@
...
+void CALLBACK SlowOperationWatcher::OnTimerFired(PTP_CALLBACK_INSTANCE, PVOID Context, PTP_TIMER) noexcept
+try
+{
+    auto* self = static_cast<SlowOperationWatcher*>(Context);
+
+    WSL_LOG_TELEMETRY(
+        "SlowOperation",
+        PDT_ProductAndServicePerformance,
+        TraceLoggingValue(self->m_name, "name"),
+        TraceLoggingInt64(self->m_slowThreshold.count(), "thresholdMs"),
```
ptrivedi, line 71: Currently when the timer fires, we know the operation was slow, but not how slow... e.g. it could have finished at 10.1 sec or it could still have been stuck at 60s. 

This is fine for now, but something to consider in a follow-up if we should calculate/log the elapsed time. It could be a bit complex though, given that logging the elapsed time at scope exit could be the most useful but would add complexity to the destructor. Anyway, food for thought

ptrivedi (approved):

PR #40272, opened 2026-04-22
Remove Advertise=yes from MSI shortcuts

benhillis:
Removes `Advertise=yes` from all MSI shortcuts (WSL, WSL Settings non-server, WSL Settings server) and drops the matching `IsSystemComponent` ShortcutProperty.

## Why

Originally posted as a fix for installer Warning 1946 (`System.AppUserModel.ToastActivatorCLSID for shortcut 'WSL.lnk' could not be set`). Per @OneBlue's review, removing `Advertise=yes` does **not** actually clear that warning, so the change is no longer framed as a Warning 1946 fix.

We still want this change because:

- Advertised shortcuts cause confusing UX when the shortcut gets leaked or the MSI ends up in a bad state (the shortcut tries to repair/install-on-demand instead of just running the binary).
- The WSL MSI has a single feature at `Level=1` (always installed) with `ARPNOMODIFY=yes` and `ARPSYSTEMCOMPONENT=yes`, so MSI advertising (install-on-demand / self-repair) provides no value here.
- `System.AppUserModel.IsSystemComponent` was attached to an advertised shortcut whose Darwin descriptor can't store all property types; removing it alongside the advertise flag avoids any ambiguity about which properties survive.

## Changes

- Drop `Advertise=yes` from the three `<Shortcut>` elements in `msipackage/package.wix.in`.
- Drop the now-unnecessary `IsSystemComponent` `<ShortcutProperty>` from `WSLSettingsShortcutNonServer`.

## Verification

- [x] Toast notifications still work (`System.AppUserModel.ToastActivatorCLSID` set on the WSL shortcut).
- [x] `System.AppUserModel.ID` remains set.
- [x] WSL shortcut visible in Start Menu with the correct icon.
- [x] Build succeeds (C++ and WiX MSI).

### DIFF
--- a/msipackage/package.wix.in
+++ b/msipackage/package.wix.in
@@ -20,7 +20,7 @@
             <Component Id="wsl" Guid="F0C8D6BA-1502-41E7-BF72-D93DFA134730" UninstallWhenSuperseded="yes" DisableRegistryReflection="yes" Bitness="always64">
                 <RemoveFile Id="CleanUpWSLShortCut" Directory="ProgramMenuFolder" Name="WSL" On="uninstall"/>
                 <File Id="wsl.exe" Name="wsl.exe" Source="${PACKAGE_INPUT_DIR}/wsl.exe" KeyPath="yes">
-                    <Shortcut Id="WSLShortcut" Name="WSL" Description="Windows Subsystem for Linux" Arguments="--cd ~" Advertise="yes" Directory="ProgramMenuFolder" Icon="wsl.ico">
+                    <Shortcut Id="WSLShortcut" Name="WSL" Description="Windows Subsystem for Linux" Arguments="--cd ~" Directory="ProgramMenuFolder" Icon="wsl.ico">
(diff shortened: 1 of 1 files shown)


OneBlue (commented):
Did a bit of research on this, and I don't think this will fix the MSI warnings, but I think we should merge it anyway because the `advertise` feature tends to cause a confusing UX if the shortcut is leaked / MSI is in a bad state. 

From [the toast notification documentation](https://learn.microsoft.com/en-us/windows/win32/shell/enable-desktop-toast-with-appusermodelid), this shouldn't impact our notification feature

benhillis:
> Did a bit of research on this, and I don't think this will fix the MSI warnings, but I think we should merge it anyway because the `advertise` feature tends to cause a confusing UX if the shortcut is leaked / MSI is in a bad state.
> 
> From [the toast notification documentation](https://learn.microsoft.com/en-us/windows/win32/shell/enable-desktop-toast-with-appusermodelid), this shouldn't impact our notification feature

I validated that notifications continue to work with this change.

OneBlue (approved):

PR #40273, opened 2026-04-22
Add BUFFER_FIELD macro for safe pretty-printing of flexible array members

benhillis:
Message structs with flexible array members (`char Buffer[]`, `char Content[]`) used `FIELD()` in `PRETTY_PRINT`, which streams the member as a C-string until a NUL byte is encountered. If the buffer happens to not be NUL-terminated (e.g. due to a truncated or malformed message), this could read past the intended bounds.

This PR adds:
- A `BUFFER_FIELD(Name)` macro that computes safe bounds from `Header.MessageSize`
- A `PrettyPrintSafeBufferView` helper that caps reads using `strnlen`
- A `std::string_view` branch in the `PrettyPrint` template
- Updates all 10 affected structs in `lxinitshared.h` to use `BUFFER_FIELD`

This matches the existing pattern in `LX_GNS_RESULT` which already excluded its `Buffer` from `PRETTY_PRINT` with the comment: *'Buffer' doesn't always contain a string, so don't pretty print it.*

### DIFF
--- a/src/shared/inc/prettyprintshared.h
+++ b/src/shared/inc/prettyprintshared.h
@@ -17,6 +17,8 @@ Abstract:
 #pragma once
 
 #include <sstream>
+#include <cstring>
+#include <string_view>
 
 #include "defs.h"
 #include "stringshared.h"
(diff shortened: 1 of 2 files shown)


src/shared/inc/prettyprintshared.h
```diff
@@ -31,6 +32,17 @@ Module Name:
...
 #define STRING_FIELD(Name) #Name, (Name <= 0 ? "<empty>" : ((char*)(this)) + Name)
 
+// Safe pretty-print for flexible array members (char Buffer[]). Bounds the read
+// using the struct's Header.MessageSize so it never reads past the received data.
+#define BUFFER_FIELD(Name) #Name, PrettyPrintSafeBufferView(this, Header.MessageSize, Name)
+
+inline std::string_view PrettyPrintSafeBufferView(const void* structBase, unsigned int messageSize, const char* buffer)
+{
+    const auto offset = static_cast<size_t>(buffer - reinterpret_cast<const char*>(structBase));
+    const size_t maxLen = messageSize > offset ? messageSize - offset : 0;
```
OneBlue: nit: If offset goes past the message size, we might want to do something like: 

```
if (offset >= messageSize)
{
    return "<out-of-bounds>";
}
```

This the issue is obvious in ETL

OneBlue (dismissed):

OneBlue (approved):

PR #40275, opened 2026-04-22
CLI: Add label option in wslc run/create

AmelBawa-msft:
-  Add label option in wslc run/create

## PR Checklist


dkbennett (approved):

src/windows/wslc/services/VolumeModel.h
```diff
@@ -19,6 +19,44 @@ Module Name:
 
 namespace wsl::windows::wslc::models {
 
+struct Label
+{
+    std::string Key() const
```
OneBlue: Could we use the same type for both of those ? Something like `KeyValuePair` ?

OneBlue: Or alternatively we could use `std::pair` for this (this would have the benefit of removing the copy that Key() and Value() do on return in the current implementation)

PR #40278, opened 2026-04-22
Add gateway driverOpt support in CreateNetwork and invalid input tests

beena352:
Add gateway support for custom network creation and expand network test coverage per PR feedback.

## PR Checklist

- Parse the Gateway driverOpt when creating a network (only applies when Subnet is also provided
- Extend NetworkCreateInvalidDriverTest to cover case sensitivity and empty inputs
- Add NetworkCreateInvalidSubnetTest and NetworkCreateInvalidGatewayTest to verify invalid inputs are rejected and the cache remains clean after rollback
- Add NetworkCreateWithGatewayTest to validate gateway round-trip through create and inspect

### DIFF
--- a/test/windows/WSLCTests.cpp
+++ b/test/windows/WSLCTests.cpp
@@ -4026,12 +4026,15 @@ class WSLCTests
     {
         WSLCNetworkOptions options{};
         options.Name = "bad-driver-net";
-        options.Driver = "overlay";
         options.DriverOpts = nullptr;
         options.DriverOptsCount = 0;
 
-        VERIFY_ARE_EQUAL(E_INVALIDARG, m_defaultSession->CreateNetwork(&options));
-        ValidateCOMErrorMessageContains(L"Unsupported network driver:");
+        for (const char* driver : {"overlay", "Bridge", ""})
+        {
+            options.Driver = driver;
+            VERIFY_ARE_EQUAL(E_INVALIDARG, m_defaultSession->CreateNetwork(&options));
+            ValidateCOMErrorMessageContains(L"Unsupported network driver:");
+        }
     }
 
     WSLC_TEST_METHOD(NetworkCreateReservedNameTest)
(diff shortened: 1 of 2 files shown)


test/windows/WSLCTests.cpp
```diff
@@ -4066,6 +4069,79 @@ class WSLCTests
...
+        options.Driver = "bridge";
+        options.DriverOpts = opts;
+        options.DriverOptsCount = ARRAYSIZE(opts);
+
+        auto cleanup = wil::scope_exit([&]() { LOG_IF_FAILED(m_defaultSession->DeleteNetwork(networkName.c_str())); });
+
+        VERIFY_ARE_EQUAL(E_INVALIDARG, m_defaultSession->CreateNetwork(&options));
+
+        wil::unique_cotaskmem_ansistring output;
+        VERIFY_ARE_EQUAL(WSLC_E_NETWORK_NOT_FOUND, m_defaultSession->InspectNetwork(networkName.c_str(), &output));
```
OneBlue, line 4090: nit: I recommend validating the error message here in and in below tests

test/windows/WSLCTests.cpp
```diff
@@ -4066,6 +4069,79 @@ class WSLCTests
...
+
+        WSLCDriverOption opts[] = {{"Subnet", "172.27.0.0/16"}, {"Gateway", "999.999.999.999"}};
+
+        WSLCNetworkOptions options{};
+        options.Name = networkName.c_str();
+        options.Driver = "bridge";
+        options.DriverOpts = opts;
+        options.DriverOptsCount = ARRAYSIZE(opts);
+
+        auto cleanup = wil::scope_exit([&]() { LOG_IF_FAILED(m_defaultSession->DeleteNetwork(networkName.c_str())); });
```
OneBlue: I think we can remove this since we don't actually expect the network to be created (same for below). Otherwise this will always log an error

OneBlue (approved):
LGTM, two minor comments

PR #40298, opened 2026-04-24
Reuse virtiofs shares when possible

OneBlue:

This change adds logic to reuse virtiofs shares whenever possible, making it less likely that we hit the device limit

## PR Checklist

### DIFF
--- a/src/windows/wslcsession/WSLCVirtualMachine.cpp
+++ b/src/windows/wslcsession/WSLCVirtualMachine.cpp
@@ -951,7 +951,11 @@ try
     THROW_HR_IF_MSG(
         HRESULT_FROM_WIN32(ERROR_PATH_NOT_FOUND), !std::filesystem::is_directory(path), "Path is not a directory: '%ls'", WindowsPath);
 
+    const bool readOnly = WI_IsFlagSet(Flags, WSLCMountFlagsReadOnly);
+    auto normalizedPath = path.lexically_normal().wstring();
     GUID shareGuid{};
+    bool reusingShare = false;
+
     {
         std::lock_guard lock(m_lock);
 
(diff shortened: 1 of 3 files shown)


src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -951,16 +951,39 @@ try
...
+        {
+            auto shareIt = m_virtioFsShares.find({normalizedPath, readOnly});
+            if (shareIt != m_virtioFsShares.end())
+            {
+                shareGuid = shareIt->second;
+                reusingShare = true;
+            }
+        }
+
+        if (!reusingShare)
```
benhillis, line 977: this will only ever be true for virtiofs so the below condition is moot.

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -971,7 +994,15 @@ try
         if (WI_VERIFY(mountIt != m_mountedWindowsFolders.end()))
         {
             m_mountedWindowsFolders.erase(mountIt);
-            LOG_IF_FAILED(m_vm->RemoveShare(shareGuid));
+            if (!reusingShare)
```
benhillis: same here

src/windows/wslcsession/WSLCVirtualMachine.h
```diff
@@ -227,6 +227,11 @@ class WSLCVirtualMachine
 
     std::map<ULONG, AttachedDisk> m_attachedDisks;
     std::map<std::string, GUID> m_mountedWindowsFolders;
+
+    // VirtioFs share cache: maps (normalized WindowsPath, readOnly) to share GUID.
+    // Shares are kept alive after unmount for reuse on subsequent mounts of the same folder.
+    std::map<std::pair<std::wstring, bool>, GUID> m_virtioFsShares;
```
benhillis, line 233: Do we need this separate collection? We already have m_mountedWindowsFolders.

benhillis (changes requested):

src/windows/wslcsession/WSLCVirtualMachine.cpp
OneBlue: My thinking was that if we're not in virtiofs mode,  `reusingShare` will be false, so we'll enter this block. At that point, we'll only want to add to `m_virtioFsShares` if we actually created a virtiofs share

src/windows/wslcsession/WSLCVirtualMachine.h
OneBlue: We can't easily reuse that one because the key is the linux path (it was originally written that way because linux paths are easier to compare). 

This is a temporary solution, so probably not worth doing much bigger change

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -971,7 +994,15 @@ try
...
+            {
+                if (FeatureEnabled(WslcFeatureFlagsVirtioFs))
+                {
+                    m_virtioFsShares.erase({normalizedPath, readOnly});
+                }
+
```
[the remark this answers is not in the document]
benhillis: Agreed, if just the mount fails we shouldn't remove the share.

benhillis (approved):

PR #40311, opened 2026-04-24
Add WSLCContainerNetworkTypeCustom support

beena352:
Add support for WSLCContainerNetworkTypeCustom, allowing containers to attach to user-created WSLC networks.

## PR Checklist

- Uncommented WSLCContainerNetworkTypeCustom = 3 in wslc.idl
- Extended ProcessPortMappings with a Custom branch that validates the network name, checks existence against the session networks map, and sets the network mode to the custom name. Custom networks allocate VM ports like bridged
- DockerNetworkModeToWSLCNetworkType now returns Custom for unrecognized modes instead of throwing, so container recovery works for custom networks
- Recovery port loop extended to allocate ports for Custom alongside Bridged

### DIFF
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -147,7 +147,7 @@ uint16_t AllocateEphemeralPort(int family, const char* address)
 
 // Builds port mapping list from container options and returns the network mode string.
 std::pair<std::vector<ContainerPortMapping>, std::string> ProcessPortMappings(
-    std::vector<_WSLCPortMapping>& requestedPorts, WSLCContainerNetworkType networkType, WSLCVirtualMachine& virtualMachine)
+    std::vector<_WSLCPortMapping>& requestedPorts, WSLCContainerNetworkType networkType, WSLCVirtualMachine& virtualMachine, WSLCSession& session, LPCSTR containerNetworkName)
 {
     // Determine network mode string.
     std::string networkMode;
@@ -162,6 +162,16 @@ std::pair<std::vector<ContainerPortMapping>, std::string> ProcessPortMappings(
     else if (networkType == WSLCContainerNetworkTypeNone)
     {
         networkMode = "none";
+    }
+    else if (networkType == WSLCContainerNetworkTypeCustom)
+    {
+        THROW_HR_WITH_USER_ERROR_IF(
+            E_INVALIDARG, Localization::MessageWslcContainerNetworkNameRequired(), !containerNetworkName || strlen(containerNetworkName) == 0);
+
+        THROW_HR_WITH_USER_ERROR_IF(
+            WSLC_E_NETWORK_NOT_FOUND, Localization::MessageWslcNetworkNotFound(containerNetworkName), !session.HasNetwork(containerNetworkName));
+
+        networkMode = containerNetworkName;
     }
     else
     {
@@ -186,8 +196,8 @@ std::pair<std::vector<ContainerPortMapping>, std::string> ProcessPortMappings(
 
         auto& entry = ports.emplace_back(VMPortMapping::FromWSLCPortMapping(e), e.ContainerPort);
 
-        // Only allocate port for bridged network. Host mode ports are allocated when the container starts.
-        if (networkType == WSLCContainerNetworkTypeBridged)
+        // Allocate VM ports for bridged and custom networks. Host mode ports are allocated when the container starts.
+        if (networkType == WSLCContainerNetworkTypeBridged || networkType == WSLCContainerNetworkTypeCustom)
         {
             entry.VmMapping.AssignVmPort(virtualMachine.AllocatePort(e.Family, e.Protocol));
         }
@@ -269,7 +279,8 @@ WSLCContainerNetworkType DockerNetworkModeToWSLCNetworkType(const std::string& m
         return WSLCContainerNetworkTypeNone;
     }
 
-    THROW_HR_MSG(E_INVALIDARG, "Invalid networking mode: %hs", mode.c_str());
+    // Any unrecognized mode is a custom user-defined network.
+    return WSLCContainerNetworkTypeCustom;
 }
 
 std::uint64_t ParseDockerTimestamp(const std::string& timestamp)
@@ -1362,7 +1373,12 @@ std::unique_ptr<WSLCContainerImpl> WSLCContainerImpl::Create(
     }
 
     // Process port mappings from container options.
-    auto [mappedPorts, networkMode] = ProcessPortMappings(ports, containerOptions.ContainerNetwork.ContainerNetworkType, virtualMachine);
+    auto [mappedPorts, networkMode] = ProcessPortMappings(
+        ports,
+        containerOptions.ContainerNetwork.ContainerNetworkType,
+        virtualMachine,
+        wslcSession,
+        containerOptions.ContainerNetwork.ContainerNetworkName);
 
     request.HostConfig.NetworkMode = networkMode;
 
@@ -1470,7 +1486,7 @@ std::unique_ptr<WSLCContainerImpl> WSLCContainerImpl::Open(
     {
         auto& inserted = ports.emplace_back(ContainerPortMapping{VMPortMapping::FromContainerMetaData(e), e.ContainerPort});
 
-        if (networkingMode == WSLCContainerNetworkTypeBridged)
+        if (networkingMode == WSLCContainerNetworkTypeBridged || networkingMode == WSLCContainerNetworkTypeCustom)
         {
             auto allocation = virtualMachine.TryAllocatePort(e.VmPort, e.Family, e.Protocol);
 
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -279,7 +279,10 @@ WSLCContainerNetworkType DockerNetworkModeToWSLCNetworkType(const std::string& m
         return WSLCContainerNetworkTypeNone;
     }
 
-    // Any unrecognized mode is a custom user-defined network.
+    // Reject Docker special syntaxes (container:<id>, service:<name>, etc.);
+    // any plain name is treated as a user-defined custom network.
+    THROW_HR_IF_MSG(E_INVALIDARG, mode.empty() || mode.find(':') != std::string::npos, "Unsupported Docker network mode: %hs", mode.c_str());
+
     return WSLCContainerNetworkTypeCustom;
 }
 
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -32,6 +32,7 @@ using wsl::windows::common::relay::ReadHandle;
 using wsl::windows::common::relay::RelayHandle;
 using wsl::windows::service::wslc::ContainerPortMapping;
 using wsl::windows::service::wslc::IWSLCVolume;
+using wsl::windows::service::wslc::NetworkEntry;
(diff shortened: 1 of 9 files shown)


src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -146,7 +146,7 @@ uint16_t AllocateEphemeralPort(int family, const char* address)
 
 // Builds port mapping list from container options and returns the network mode string.
 std::pair<std::vector<ContainerPortMapping>, std::string> ProcessPortMappings(
-    std::vector<_WSLCPortMapping>& requestedPorts, WSLCContainerNetworkType networkType, WSLCVirtualMachine& virtualMachine)
+    std::vector<_WSLCPortMapping>& requestedPorts, WSLCContainerNetworkType networkType, WSLCVirtualMachine& virtualMachine, WSLCSession& session, LPCSTR containerNetworkName)
 {
     // Determine network mode string.
```
OneBlue, line 162: We should return E_INVALIDARG if `containerNetworkName` is non-null and `networkType` is not `WSLCContainerNetworkTypeCustom`

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -162,6 +162,16 @@ std::pair<std::vector<ContainerPortMapping>, std::string> ProcessPortMappings(
     {
         networkMode = "none";
     }
+    else if (networkType == WSLCContainerNetworkTypeCustom)
+    {
+        THROW_HR_WITH_USER_ERROR_IF(
+            E_INVALIDARG, Localization::MessageWslcContainerNetworkNameRequired(), !containerNetworkName || strlen(containerNetworkName) == 0);
+
+        THROW_HR_WITH_USER_ERROR_IF(
+            WSLC_E_NETWORK_NOT_FOUND, Localization::MessageWslcNetworkNotFound(containerNetworkName), !session.HasNetwork(containerNetworkName));
```
OneBlue: Checking `HasNetwork()` creates a race condition, because the network locks is dropped when this method returns. 

To fix this, I recommend passing in the networks while holding the networks lock, like we do for volumes

test/windows/WSLCTests.cpp
```diff
@@ -5284,6 +5284,57 @@ class WSLCTests
...
 
+    WSLC_TEST_METHOD(ContainerCustomNetworkTest)
+    {
+        const std::string networkName = "custom-net-test";
+
+        LOG_IF_FAILED(m_defaultSession->DeleteNetwork(networkName.c_str()));
+
+        WSLCDriverOption opts[] = {{"Subnet", "172.30.0.0/16"}};
+
+        WSLCNetworkOptions networkOptions{};
```
OneBlue, line 5361: I recommend adding test cases for: 

- Multiple containers in the same network
- Trying to delete the network while a container is running in it
- Mapping container ports when the container is running in a custom network
- Recovering a container from a the session VHD that has a custom network

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -268,7 +288,11 @@ WSLCContainerNetworkType DockerNetworkModeToWSLCNetworkType(const std::string& m
         return WSLCContainerNetworkTypeNone;
     }
 
-    THROW_HR_MSG(E_INVALIDARG, "Invalid networking mode: %hs", mode.c_str());
+    // Reject Docker special syntaxes (container:<id>, service:<name>, etc.);
+    // any plain name is treated as a user-defined custom network.
+    THROW_HR_IF_MSG(E_INVALIDARG, mode.empty() || mode.find(':') != std::string::npos, "Unsupported Docker network mode: %hs", mode.c_str());
+
+    return WSLCContainerNetworkTypeCustom;
```
kvega005, line 302: Do we want to at least eventualy support something like `container:<name|id>`, (this container shares another container's network namespace)?

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -268,7 +288,11 @@ WSLCContainerNetworkType DockerNetworkModeToWSLCNetworkType(const std::string& m
         return WSLCContainerNetworkTypeNone;
     }
 
-    THROW_HR_MSG(E_INVALIDARG, "Invalid networking mode: %hs", mode.c_str());
+    // Reject Docker special syntaxes (container:<id>, service:<name>, etc.);
+    // any plain name is treated as a user-defined custom network.
+    THROW_HR_IF_MSG(E_INVALIDARG, mode.empty() || mode.find(':') != std::string::npos, "Unsupported Docker network mode: %hs", mode.c_str());
+
```
kvega005, line 301: Docker treats mode.empty() as "default" which is bridged. Do we want to handle that here?

kvega005: It may be work adding that handling in case we ever want to try to support loading containers created via the docker CLI inside the session shell.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -146,8 +147,17 @@ uint16_t AllocateEphemeralPort(int family, const char* address)
...
+    std::vector<_WSLCPortMapping>& requestedPorts,
+    WSLCContainerNetworkType networkType,
+    WSLCVirtualMachine& virtualMachine,
+    const std::unordered_map<std::string, NetworkEntry>& sessionNetworks,
+    LPCSTR containerNetworkName)
 {
+    THROW_HR_IF_MSG(
+        E_INVALIDARG,
+        containerNetworkName != nullptr && networkType != WSLCContainerNetworkTypeCustom,
+        "ContainerNetworkName must be null when network type is not Custom");
```
kvega005, line 160: Should this be a localized message?

beena352: This uses THROW_HR_IF_MSG, so the message only goes to the trace log and is not surfaced to the caller through IErrorInfo. Since this is meant as an internal guard for invalid struct usage, I left it unlocalized to match the existing checks nearby. Happy to localize it if you would prefer.

src/windows/wslcsession/WSLCContainer.cpp
beena352: Yes, we can do this. Let me take this as a follow-up, container:<id> is a per-container attachment mode rather than a named network.

test/windows/WSLCTests.cpp
```diff
@@ -5335,6 +5335,177 @@ class WSLCTests
...
+
+        auto container1 = launcher1.Launch(*m_defaultSession);
+        VERIFY_ARE_EQUAL(container1.State(), WslcContainerStateRunning);
+        VERIFY_ARE_EQUAL(container1.Inspect().HostConfig.NetworkMode, networkName);
+
+        auto container2 = launcher2.Launch(*m_defaultSession);
+        VERIFY_ARE_EQUAL(container2.State(), WslcContainerStateRunning);
+        VERIFY_ARE_EQUAL(container2.Inspect().HostConfig.NetworkMode, networkName);
+
+        VERIFY_SUCCEEDED(container1.Get().Stop(WSLCSignalSIGTERM, 0));
```
OneBlue: nit: This isn't needed since `WSLCContainerLauncher ` automatically deletes containers on scope exit

OneBlue: (Same for below)

test/windows/WSLCTests.cpp
```diff
@@ -5335,6 +5335,177 @@ class WSLCTests
...
+        auto networkCleanup = wil::scope_exit([&]() { LOG_IF_FAILED(m_defaultSession->DeleteNetwork(networkName.c_str())); });
+
+        WSLCContainerLauncher launcher(
+            "debian:latest", "test-custom-net-inuse", {"sleep", "99999"}, {}, WSLCContainerNetworkType::WSLCContainerNetworkTypeCustom);
+        launcher.SetContainerNetworkName(std::string(networkName));
+
+        auto container = launcher.Launch(*m_defaultSession);
+        VERIFY_ARE_EQUAL(container.State(), WslcContainerStateRunning);
+
+        VERIFY_FAILED(m_defaultSession->DeleteNetwork(networkName.c_str()));
```
OneBlue: I recommend validating the exact error code that we expect here

test/windows/WSLCTests.cpp
```diff
@@ -5335,6 +5335,177 @@ class WSLCTests
...
+            VERIFY_ARE_EQUAL(container.State(), WslcContainerStateCreated);
+            VERIFY_SUCCEEDED(container.Get().Start(WSLCContainerStartFlagsAttach, nullptr));
+            VERIFY_ARE_EQUAL(container.State(), WslcContainerStateRunning);
+
+            VERIFY_ARE_EQUAL(container.Inspect().HostConfig.NetworkMode, networkName);
+
+            VERIFY_SUCCEEDED(container.Get().Stop(WSLCSignalSIGKILL, 0));
+            VERIFY_SUCCEEDED(container.Get().Delete(WSLCDeleteFlagsNone));
+
+            LOG_IF_FAILED(session->DeleteNetwork(networkName.c_str()));
```
OneBlue: This should be in a scope_exit so the network is deleted on failure

benhillis:

test/windows/WSLCTests.cpp
```diff
@@ -5350,6 +5350,221 @@ class WSLCTests
...
+        auto networkCleanup = wil::scope_exit([&]() { LOG_IF_FAILED(m_defaultSession->DeleteNetwork(networkName.c_str())); });
+
+        WSLCContainerLauncher launcher(
+            "debian:latest", "test-custom-network", {"sleep", "99999"}, {}, WSLCContainerNetworkType::WSLCContainerNetworkTypeCustom);
+        launcher.SetContainerNetworkName(std::string(networkName));
+
+        auto container = launcher.Launch(*m_defaultSession);
+        VERIFY_ARE_EQUAL(container.State(), WslcContainerStateRunning);
+        VERIFY_ARE_EQUAL(container.Inspect().HostConfig.NetworkMode, networkName);
+        VERIFY_SUCCEEDED(container.Get().Stop(WSLCSignalSIGTERM, 0));
```
OneBlue: nit: This isn't needed since `RunningWSLCContainer` stops and deletes contains on scope exit by default

test/windows/WSLCTests.cpp
```diff
@@ -5350,6 +5350,221 @@ class WSLCTests
...
+
+            networkCleanup.release();
+        }
+
+        {
+            auto session = CreateSession(GetDefaultSessionSettings(L"recovery-test-custom-net", true, WSLCNetworkingModeNAT));
+            auto networkCleanup = wil::scope_exit([&]() { LOG_IF_FAILED(session->DeleteNetwork(networkName.c_str())); });
+
+            auto container = OpenContainer(session.get(), containerName);
+            container.SetDeleteOnClose(false);
```
OneBlue: nit: We could remove this and let the container be automatically deleted when this test completes to simplify things

test/windows/WSLCTests.cpp
```diff
@@ -5350,6 +5350,221 @@ class WSLCTests
...
+
+    WSLC_TEST_METHOD(ContainerCustomNetworkRecoveryTest)
+    {
+        auto restore = ResetTestSession();
+
+        const std::string networkName = "custom-net-recovery";
+        const std::string containerName = "test-custom-net-recovery";
+
+        {
+            auto session = CreateSession(GetDefaultSessionSettings(L"recovery-test-custom-net", true, WSLCNetworkingModeNAT));
```
OneBlue: nit: This test could be done in `m_defaultSession` (We could create the network and then call             ResetTestSession() to force a session reload).

OneBlue (dismissed):
LGTM, minor comments

OneBlue (approved):

PR #40312, opened 2026-04-24
CLI: Unit test to ensure no argument name or alias collisions exist

dkbennett:
Simple test that walks the CLI command tree and ensures no commands have ambiguous names or aliases that map to different arguments.

Example of finding a collision (I intentionally gave session the same alias as help to force a collision)
image

Example of no collisions (the above alias removed)
image

## PR Checklist

Walks the command tree from the Root and examines arguments of each command and ensures there are no duplicate names or aliases in any of them.

## Validation Steps Performed
Test passes!

### DIFF
--- a/test/windows/wslc/WSLCCLICommandUnitTests.cpp
+++ b/test/windows/wslc/WSLCCLICommandUnitTests.cpp
@@ -170,45 +170,87 @@ class WSLCCLICommandUnitTests
         VERIFY_IS_TRUE(found, L"RootCommand should contain VersionCommand");
     }
 
-    // Walk every command in the root tree and verify no name or alias argument collisions.
+    // Walk every command in the root tree and verify no argument collisions.
     TEST_METHOD(AllCommands_NoAmbiguousArgumentNamesOrAliases)
     {
-        // Starting with the Root command, verify no argument collisions.
-        std::vector<std::unique_ptr<Command>> queue;
-        queue.push_back(std::make_unique<RootCommand>());
+        // Build a lookup table from ArgType -> enum name string using the same X-macro.
+        static constexpr const wchar_t* c_argTypeNames[] = {
+#define WSLC_ARG_ENUM(EnumName, Name, Alias, Kind, Desc) L## #EnumName,
+            WSLC_ARGUMENTS(WSLC_ARG_ENUM)
+#undef WSLC_ARG_ENUM
+        };
 
-        while (!queue.empty())
+        const auto ArgTypeName = [](argument::ArgType type) -> std::wstring_view {
+            const auto index = static_cast<size_t>(type);
+            const auto max = static_cast<size_t>(argument::ArgType::Max);
+            if (index < max)
+            {
+                return c_argTypeNames[index];
+            }
+
+            return L"<unknown>";
+        };
+
+        // Starting with the Root command, verify no argument collisions.
+        std::vector<std::unique_ptr<Command>> commands;
+        commands.push_back(std::make_unique<RootCommand>());
+
+        while (!commands.empty())
         {
-            auto current = std::move(queue.back());
-            queue.pop_back();
+            auto current = std::move(commands.back());
+            commands.pop_back();
             VERIFY_IS_NOT_NULL(current.get());
 
             const std::wstring commandFullName(current->FullName());
-            std::unordered_set<std::wstring> seenNames;
-            std::unordered_set<std::wstring> seenAliases;
+            std::unordered_set<size_t> seenTypes;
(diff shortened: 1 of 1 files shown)


test/windows/wslc/WSLCCLICommandUnitTests.cpp
```diff
@@ -169,6 +169,47 @@ class WSLCCLICommandUnitTests
...
+                const auto& name = arg.Name();
+                auto [it, inserted] = seenNames.emplace(name, name);
+                VERIFY_IS_TRUE(inserted, std::format(L"Command '{}' has no duplicate name '--{}'", commandFullName, name).c_str());
+
+                // Check alias collision; skip empty aliases (NO_ALIAS).
+                const auto& alias = arg.Alias();
+                if (!alias.empty())
+                {
+                    auto [it, inserted] = seenAliases.emplace(alias, name);
+                    VERIFY_IS_TRUE(inserted, std::format(L"Command '{}' has no duplicate alias '-{}'", commandFullName, alias).c_str());
```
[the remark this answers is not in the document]
dkbennett: This is good feedback and its unnecessary, it should be obvious if there's a collision what the collision is from the verify statement

I've updated to unordered map.

test/windows/wslc/WSLCCLICommandUnitTests.cpp
```diff
@@ -169,6 +169,47 @@ class WSLCCLICommandUnitTests
...
+                // Check name collision.
+                const auto& name = arg.Name();
+                auto [it, inserted] = seenNames.emplace(name, name);
+                VERIFY_IS_TRUE(inserted, std::format(L"Command '{}' has no duplicate name '--{}'", commandFullName, name).c_str());
+
+                // Check alias collision; skip empty aliases (NO_ALIAS).
+                const auto& alias = arg.Alias();
+                if (!alias.empty())
+                {
+                    auto [it, inserted] = seenAliases.emplace(alias, name);
```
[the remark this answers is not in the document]
dkbennett: Fixed this with unordered set.

test/windows/wslc/WSLCCLICommandUnitTests.cpp
```diff
@@ -169,6 +169,49 @@ class WSLCCLICommandUnitTests
...
+            std::unordered_set<std::wstring> seenNames;
+            std::unordered_set<std::wstring> seenAliases;
+
+            for (const auto& arg : current->GetAllArguments())
+            {
+                // Check name collision.
+                const auto& name = arg.Name();
+                VERIFY_IS_TRUE(
+                    seenNames.emplace(name).second,
+                    std::format(L"Command '{}' has no duplicate name '--{}'", commandFullName, name).c_str());
```
AmelBawa-msft: Should the message say "Command has duplicate name" instead?

test/windows/wslc/WSLCCLICommandUnitTests.cpp
```diff
@@ -169,6 +169,49 @@ class WSLCCLICommandUnitTests
 
         VERIFY_IS_TRUE(found, L"RootCommand should contain VersionCommand");
     }
+
+    // Walk every command in the root tree and verify no name or alias argument collisions.
+    TEST_METHOD(AllCommands_NoAmbiguousArgumentNamesOrAliases)
+    {
+        // Starting with the Root command, verify no argument collisions.
+        std::vector<std::unique_ptr<Command>> queue;
+        queue.push_back(std::make_unique<RootCommand>());
```
AmelBawa-msft: I initially thought we are doing BFS, but I believe the queue here is behaving more like a stack (DFS). Consider renaming to 'stack' or 'commands'

AmelBawa-msft (approved):
Approve with suggestions

test/windows/wslc/WSLCCLICommandUnitTests.cpp
dkbennett: When I ran the test it reads "Verify.IsTrue(.... " with the message command has no duplicate name. When it fails, the verify fails and so it is not true, thus there is a duplicate.  It just reads awkwardly that we are verifying something is true (no duplicates) but indicating in the message that there is a duplicate. The messaging as-is is clear when looking at the test output.

dkbennett: And I've updated it to only output message on failure, and now it includes the ArgType names that collide for very easy identification of problems.

test/windows/wslc/WSLCCLICommandUnitTests.cpp
dkbennett: Updated

yao-msft (approved):

PR #40315, opened 2026-04-24
CLI: Enhance image and container inspection behavior

AmelBawa-msft:
- Graceful not-found handling for inspect commands, container inspect and image inspect now 
catch not-found errors and print a friendly localized message to stderr instead of throwing an 
unhandled exception.
 - Partial results on multi-inspect. When inspecting multiple items, found items are still returned in JSON output; missing items log an error and set exit code 1.
 - New MessageWslcImageNotFound localization string added to Resources.resw.
 - New E2E test suite for container inspect (help, missing arg, not-found, single, multi, mixed found/not-found).
 - Updated image inspect not-found test to expect the new friendly error + empty JSON array output.

## PR Checklist

### DIFF
--- a/src/windows/wslc/tasks/ContainerTasks.cpp
+++ b/src/windows/wslc/tasks/ContainerTasks.cpp
@@ -24,6 +24,7 @@ Abstract:
 #include <wslc_schema.h>
 
 using namespace wsl::shared;
+using namespace wsl::windows::common;
 using namespace wsl::windows::common::string;
 using namespace wsl::windows::common::wslutil;
 using namespace wsl::windows::wslc::execution;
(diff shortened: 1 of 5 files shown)


src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -24,13 +24,34 @@ Module Name:
...
 namespace wsl::windows::wslc::task {
+
+static bool TryInspectContainer(Session& session, const std::string& containerId, std::optional<wslc_schema::InspectContainer>& inspectData)
+{
+    try
+    {
+        inspectData = ContainerService::Inspect(session, containerId);
+        return true;
+    }
+    catch (const wil::ResultException& ex)
```
OneBlue, line 43: I recommend using `catch (...)` so these blocks also catch C++/ winrt  exceptions

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -71,8 +92,15 @@ void InspectContainers(CLIExecutionContext& context)
     std::vector<wsl::windows::common::wslc_schema::InspectContainer> result;
     for (const auto& id : containerIds)
     {
-        auto inspectData = ContainerService::Inspect(session, WideToMultiByte(id));
-        result.push_back(inspectData);
+        std::optional<wslc_schema::InspectContainer> inspectData;
+        if (TryInspectContainer(session, WideToMultiByte(id), inspectData))
+        {
+            result.push_back(*inspectData);
```
OneBlue, line 98: nit: `inspectData.value()` could be moved to avoid a copy

OneBlue: (Same for below)

OneBlue (approved):

PR #40317, opened 2026-04-24
merge master -> feature/wsl-for-apps

benhillis:


OneBlue (approved):

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -567,15 +568,18 @@ WSLCVirtualMachine::ConnectedSocket WSLCVirtualMachine::ConnectSocket(wsl::share
 {
     WSLC_ACCEPT message{};
     message.Fd = Fd;
-    const auto& response = Channel.Transaction(message);
+
+    auto transaction = Channel.StartTransaction();
```
OneBlue, line 572: Unless I'm missing something, the previous implementation should already do the right thing (`Transaction()`) internally calls `StartTransaction()` (applies in other places)

benhillis: You're right, I think I went overboard with this. Only a few changes should be needed.

PR #40319, opened 2026-04-25
fix: replace bare except with except Exception in Python scripts

srpatcha:
## Changes
fix: replace bare except with except Exception in Python scripts

Signed-off with GPG.


OneBlue:
This pull request seem to contain various unrelated changes. There's also no context on what the benefit of catching `Exception` explicitly is over generic `except` block.

Closing since the benefit from that change is unclear, and the diff contains unrelated changes.

PR #40329, opened 2026-04-25
Check WSLCHandle discriminant before accessing union member

benhillis:
DmesgOutput is a discriminated union (WSLCHandle). Accessing Handle.File without verifying Type == WSLCHandleTypeFile is undefined behavior when the active member is Pipe or Socket. Adds the type check before accessing the File member.

### DIFF
--- a/src/windows/service/exe/HcsVirtualMachine.cpp
+++ b/src/windows/service/exe/HcsVirtualMachine.cpp
@@ -112,6 +112,11 @@ HcsVirtualMachine::HcsVirtualMachine(_In_ const WSLCSessionSettings* Settings)
     // TODO: move dmesg collector to user session process.
     // N.B. 'DmesgOutput' needs to be duplicated since COM will close it when this call completes.
     wil::unique_handle dmesgOutputHandle;
+    // All WSLCHandle union members are HANDLE typedefs with the same size and offset,
+    // so accessing .File is safe regardless of the discriminant.
+    static_assert(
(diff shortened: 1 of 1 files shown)


JohnMcPMS (approved):

src/windows/service/exe/HcsVirtualMachine.cpp
```diff
@@ -112,7 +112,8 @@ HcsVirtualMachine::HcsVirtualMachine(_In_ const WSLCSessionSettings* Settings)
     // TODO: move dmesg collector to user session process.
     // N.B. 'DmesgOutput' needs to be duplicated since COM will close it when this call completes.
     wil::unique_handle dmesgOutputHandle;
-    if (Settings->DmesgOutput.Handle.File != nullptr && Settings->DmesgOutput.Handle.File != INVALID_HANDLE_VALUE)
+    if (Settings->DmesgOutput.Type == WSLCHandleTypeFile && Settings->DmesgOutput.Handle.File != nullptr &&
+        Settings->DmesgOutput.Handle.File != INVALID_HANDLE_VALUE)
```
OneBlue: I don't think we should require this handle to actually be a file. This will most likely fail in the tests, since we use a pipe there. 

It was originally written that way because all union fields have the same side and offset. We could probably make this safe by static_asserting that all fields are equivalent to access the underlying through

OneBlue (changes requested):

PR #40330, opened 2026-04-25
Remove unreachable else branch in TTY stdin relay

benhillis:
write() on success returns a value in [0, count]. The condition (bytesWritten <= pendingStdin.size()) is therefore always true, making the else branch dead code. Simplify to unconditional erase.

### DIFF
--- a/src/linux/init/WSLCInit.cpp
+++ b/src/linux/init/WSLCInit.cpp
@@ -277,14 +277,9 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLC_TTY_RELAY
             }
             else
             {
-                if (bytesWritten <= pendingStdin.size()) // Partial or complete write
-                {
-                    pendingStdin.erase(pendingStdin.begin(), pendingStdin.begin() + bytesWritten);
-                }
-                else
-                {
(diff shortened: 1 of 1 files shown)


src/linux/init/WSLCInit.cpp
```diff
@@ -277,14 +277,9 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, const WSLC_TTY_RELAY
...
             {
-                if (bytesWritten <= pendingStdin.size()) // Partial or complete write
-                {
-                    pendingStdin.erase(pendingStdin.begin(), pendingStdin.begin() + bytesWritten);
-                }
-                else
-                {
-                    LOG_ERROR("Unexpected write result {}, pending={}", bytesWritten, pendingStdin.size());
-                }
+                // write() returns 0..pendingStdin.size(), so this is always a
```
JohnMcPMS: `WIL_ASSERT` as a defensive measure?

JohnMcPMS (approved):

src/linux/init/WSLCInit.cpp
OneBlue: Yeah I originally wrote this as a "broken-logic" detection, since relays are hard to get right

benhillis: ok, I can switch to WI_ASSERT

Issue #7456, opened 2021-09-23
Zone.Identifier Files when downloading from Windows to WSL file structure


ameya-jain:
When downloading a file from an online source lets say [this image](https://venturebeat.com/wp-content/uploads/2021/07/windows11-e1625655697329.png?fit=2879%2C1439&strip=all) or a zip folder and directly saving it to the WSL file structure using the Microsoft Edge 'save as' option ( I have turned off the default download location so edge asks me every time where I should save something downloaded), when the download is complete I see these Zone Identifier files.

I read about a similar #4609 issue and from what I understand is that these files shouldn't be downloaded or should be removed after download.

![image](https://user-images.githubusercontent.com/43452398/134534516-7c1390ed-18e4-4f50-b49c-069145a4db61.png)

My Ubuntu specifications : 
![image](https://user-images.githubusercontent.com/43452398/134533677-c1ea4b26-9ac6-40a0-806a-ffd45f6d6777.png)

My windows : 
![image](https://user-images.githubusercontent.com/43452398/134531262-16b83ac8-c35f-42b0-9c26-e6e0013fc321.png)

rehanhaider:
I can confirm this issue still exists on Windows 11, even though it was supposed to be fixed. Steps to reproduce are
1. Open any webpage and download an image, similar to [this]
2. In the "Save As" dialogue, choose "Linux -> Ubuntu" and browse to folder of choice, and save
3. Creates a file with name "<download-file-name>:Zone.Identifier"

JacobDB:
I'm also still seeing this in Windows 11, any time I download a file and copy it over to WSL via File Explorer, along comes a `:Zone.Identifier`.

Kai-Richardson:
As the filer of the [original issue](https://github.com/microsoft/WSL/issues/4609), I can confirm this is still not fixed for either Windows 10 or Windows 11. 
Exact same reproduction steps I outlined [here](https://github.com/microsoft/WSL/issues/4609#issuecomment-948322716) still work:
> 1. Navigate to [an image url](https://i.imgur.com/7uEP6ds.png).
> 2. Right click and Save Image As
> 3. Choose location on my Windows NTFS C: Drive, such as `C:\Users\zekai\Desktop`
> 4. Drag (Click and hold) or copy/paste (CTRL+C & CTRL+P) PNG file to a WSL filespace, such as `\\wsl$\Ubuntu\home\cassini\github`
[...]

Issue #4609, opened 2019-10-21
Zone.Identifier Files when copying from Windows to WSL filestructure


Kai-Richardson:
* Windows build number:  Microsoft Windows [Version 10.0.18990.1]

`uname -a` produces: Linux 4.19.67-microsoft-standard #1 SMP Sun Aug 18 13:37:54 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

* What you're doing and what's happening:  This occurs when using Windows Explorer.
1.  Download a file from an online resource into a windows folder. For example, [this image](https://i.imgur.com/7uEP6ds.png) downloaded into your Downloads folder.
2. Drag/Copy/Move said file into a WSL directory.
3. Notice the file that is also created, named like "movedfile•Zone.Identifier", but instead of a bullet it is the character 0xF03A (Unicode: U+F03A). It does render like a bullet in Windows Explorer though.

* What's wrong / what should be happening instead: The filename•Zone.Identifier NFTS files should not be created.

Contents of the Zone Identifier file are as to be expected, my sample having the contents:
```
[ZoneTransfer]
ZoneId=3
ReferrerUrl=https://i.imgur.com/7uEP6ds.png
HostUrl=https://i.imgur.com/7uEP6ds.png
```

Image:
![image](https://user-images.githubusercontent.com/49656058/67179151-4d558f00-f38a-11e9-9708-ea2ac2b9edb1.png)

therealkenc:
Can't repro here. 

![image](https://user-images.githubusercontent.com/7049700/67180567-d969b580-f38e-11e9-972e-30f5f139a39e.png)


This is on 19002; whether the small version bump makes a difference I doubt. [Zone.Identifier](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/6e3f7352-d11c-4d76-8c39-2516a9df36e8) does get a quick search hit. Said [functionality](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms537183(v=vs.85)?redirectedfrom=MSDN) is news to me.

On a lark, from your WSL2 bash prompt, what is the output of:
```
$ powershell.exe "Get-CimInstance -Namespace root/SecurityCenter2 -ClassName AntivirusProduct" | grep displayName
```

0xbadfca11:
I can reproduce with both WSL1 and WSL2 at 19002.
[...]

PR #40335, opened 2026-04-27
Add option to block zone identifier creation when accessing Linux files from Windows

chemwolf6922:

When copying a file from NTFS with the zone identifier file stream to the p9 share, a `xxx:Zone.Identifier` file will be created in the Linux file system. This has been a problem for WSL users for a long time.
This PR adds a wsl.conf option `fileServer.blockZoneIdentifier` (default false) to block the file creation in the p9 server. So users can opt in to avoid its creation.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/7456 https://github.com/microsoft/WSL/issues/4609

## Validation Steps Performed

* Add test Plan9Tests::Plan9Tests::TestBlockZoneIdentifier.
* Validated with browser file downloads.

### DIFF
--- a/src/linux/init/WslDistributionConfig.cpp
+++ b/src/linux/init/WslDistributionConfig.cpp
@@ -45,6 +45,7 @@ WslDistributionConfig::WslDistributionConfig(const char* configFilePath)
         ConfigKey("fileServer.logFile", Plan9LogFile),
         ConfigKey("fileServer.logLevel", Plan9LogLevel),
         ConfigKey("fileServer.logTruncate", Plan9LogTruncate),
+        ConfigKey("fileServer.blockZoneIdentifier", Plan9BlockZoneIdentifier),
 
         ConfigKey(c_ConfigGpuEnabledOption, GpuEnabled),
         ConfigKey(c_ConfigAppendGpuLibPathOption, AppendGpuLibPath),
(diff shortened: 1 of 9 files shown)


benhillis:
Thanks for trying this, I think the main issue I see is SymLink, Link, MkNod, MkDir not blocked. Test cases that cover this should be added as well.

chemwolf6922:
> Thanks for trying this, I think the main issue I see is SymLink, Link, MkNod, MkDir not blocked. Test cases that cover this should be added as well.

Hi @benhillis , I added the block logic for these. And added test for mkdir.
However, tests for symlink, link, rename and mknod are not added. Windows does not initiate mknod. And Symlink, link and rename with ":" in the filename do not work on Windows. I tried SetFileInformationByHandle and that did not work either . Please let me know if you have more insights or suggestions on this. Thanks.

OneBlue (commented):
Thank you for doing this, however I'm not sure that this is the right way to solve this. This change would require a change to the per-distro plan9 config file  (which most users don't know about), and blindly returning access denied based on filenames will most likely have other side effects. 

If we want to fix this, I would suggest to investigate this from the Windows side, and understand what's causing Windows to create these files on a non-NTFS volume (\\wsl.localhost). Ideally, these files should never be created on WSL

benhillis:
> Thank you for doing this, however I'm not sure that this is the right way to solve this. This change would require a change to the per-distro plan9 config file (which most users don't know about), and blindly returning access denied based on filenames will most likely have other side effects.
> 
> If we want to fix this, I would suggest to investigate this from the Windows side, and understand what's causing Windows to create these files on a non-NTFS volume (\wsl.localhost). Ideally, these files should never be created on WSL

Agreed, this seems to be fixing the symptom, not the cause. There is nothing inherently wrong with files that have this naming.

PR #40338, opened 2026-04-27
Use auto reset event for container stop

kvega005:

Refactor the `WSLCContainerImpl` class to utilize an `auto_reset_event` for managing container stop notifications. This change simplifies the signaling mechanism and eliminates the need for a separate promise to handle stop state transitions.

## PR Checklist

**Problem.** The previous implementation used a mutex and a lazily-created `std::optional<std::promise<uint64_t>>` to deliver the docker stop-event timestamp from `OnEvent()` to` Stop()`, which exposed a TOCTOU race when the container exited concurrently with a caller invoking `Stop()`: if `OnEvent()` acquired the lock and observed the optional as empty before `Stop()` had a chance to create it, `OnEvent()` would perform the state transition and return without ever fulfilling the promise; `Stop()` would then create the promise, wait on its future, and hang for the full 60-second timeout because the only producer had already run.

**Change.** Introduced a `StopNotification` struct containing an `atomic` timestamp and a `wil::slim_event_auto_reset` for signaling. This change allows `OnEvent` to signal the event directly, simplifying the stop notification process.

## Validation Steps Performed

- Automated tests added or updated to cover the new signaling mechanism.
- Manual validation performed by running container stop scenarios to ensure correct event signaling and state transitions.

### DIFF
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -725,8 +725,7 @@ void WSLCContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCod
 bool WSLCContainerImpl::WaitForEvent(const wil::unique_event& Event, std::chrono::milliseconds Timeout) const
 {
     const HANDLE waitHandles[] = {Event.get(), m_wslcSession.SessionTerminatingEvent()};
-    const DWORD waitResult =
-        WaitForMultipleObjects(RTL_NUMBER_OF(waitHandles), waitHandles, FALSE, gsl::narrow<DWORD>(Timeout.count()));
+    const DWORD waitResult = WaitForMultipleObjects(RTL_NUMBER_OF(waitHandles), waitHandles, FALSE, gsl::narrow<DWORD>(Timeout.count()));
 
     switch (waitResult)
     {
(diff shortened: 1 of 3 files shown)


src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -809,11 +783,11 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill)
     }
 
     // Wait for the stop event to get the Docker timestamp.
-    // Safe while holding m_lock since OnEvent() uses m_stopStateLock on this path.
+    // OnEvent() signals the event before taking m_lock, so this won't deadlock.
     std::optional<std::uint64_t> stopTimestamp;
-    if (stopFuture.wait_for(60s) == std::future_status::ready)
+    if (waitForStop && m_stopNotification.Event.wait(60'000))
```
OneBlue: This was already a problem with the previous implementation, but in situations like that if would be nice to wait for either `m_stopNotification.Event` or `WSLCSession::m_sessionTerminatingEvent`, so we can exit early if the session is being terminated. 

If the second is signaled, we could just throw E_ABORT like we do in other places

OneBlue: (We might need to switch to a wil::unique_event for that though)

OneBlue (approved):

src/windows/wslcsession/WSLCContainer.cpp
OneBlue: Feel free to merge as-is though, since this is a pre-existing issue

OneBlue (approved):

PR #40345, opened 2026-04-28
fedora: Add Fedora 44 and drop Fedora 42

jeremycline:
Fedora 44 releases on April 28th; add its images and make it the default. Additionally, Fedora 42 will shortly go EOL so it is removed from the list.

### DIFF
--- a/distributions/DistributionInfo.json
+++ b/distributions/DistributionInfo.json
@@ -199,9 +199,22 @@
         ],
         "Fedora": [
             {
+                "Name": "FedoraLinux-44",
+                "FriendlyName": "Fedora Linux 44",
+                "Default": true,
+                "Amd64Url": {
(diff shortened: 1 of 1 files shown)


OneBlue:
@jeremycline: From the tests I can see that that tars have selinux extended attributes set (security.selinux), which will cause various warnings if users try to use these distros in wsl1 mode. 

This isn't a blocker, so we can either merge as-is, or you could push another iteration with updated tars. Let me know what you want to do here.

jeremycline:
> @jeremycline: From the tests I can see that that tars have selinux extended attributes set (security.selinux), which will cause various warnings if users try to use these distros in wsl1 mode.
> 
> This isn't a blocker, so we can either merge as-is, or you could push another iteration with updated tars. Let me know what you want to do here.

Yeah, this has been an issue for a while and I had made some build tweaks to see about fixing this, but I've been slammed with other things this release cycle and never followed up on it. There's generally not much appetite that I'm aware of to support WSL1. I'm fine merging it as-is. Thanks!

OneBlue (approved):

PR #40346, opened 2026-04-28
Use session display name as WSLC VM owner

benhillis:
Use the session display name as the HCS VM owner field so that `hcsdiag list` output shows which session owns each VM (e.g. `wslc-cli-admin-benhillis`). This aids debugging and allows tests to reliably identify their VM.

Falls back to `WSLC` if DisplayName is null (defensive, should not happen in practice).

### Changes
- `HcsVirtualMachine.cpp`: Set owner to `Settings->DisplayName` instead of `L"WSL"`
- `WSLCTests.cpp`: Add `VmOwnerMatchesSessionDisplayName` test with `ListVmOwners()` helper that parses `hcsdiag list` output

### DIFF
--- a/src/windows/service/exe/HcsVirtualMachine.cpp
+++ b/src/windows/service/exe/HcsVirtualMachine.cpp
@@ -49,7 +49,7 @@ HcsVirtualMachine::HcsVirtualMachine(_In_ const WSLCSessionSettings* Settings)
 
     // Build HCS settings
     hcs::ComputeSystem systemSettings{};
-    systemSettings.Owner = L"WSL";
+    systemSettings.Owner = Settings->DisplayName ? Settings->DisplayName : L"WSLC";
     systemSettings.ShouldTerminateOnLastHandleClosed = true;
 
     // Determine which schema version to use based on the Windows version. Windows 10 does not support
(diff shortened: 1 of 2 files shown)


test/windows/WSLCTests.cpp
```diff
@@ -437,6 +438,50 @@ class WSLCTests
         }
     }
 
+    // Returns the set of VM owner names currently reported by hcsdiag.
+    static std::vector<std::wstring> ListVmOwners()
```
benhillis, line 442: This utility will also be used in an upcoming change I'm working on to use `hcsdiag kill <vm_id>` to test unexpected VM termination.

src/windows/service/exe/HcsVirtualMachine.cpp
```diff
@@ -49,7 +49,7 @@ HcsVirtualMachine::HcsVirtualMachine(_In_ const WSLCSessionSettings* Settings)
 
     // Build HCS settings
     hcs::ComputeSystem systemSettings{};
-    systemSettings.Owner = L"WSL";
+    systemSettings.Owner = Settings->DisplayName ? Settings->DisplayName : L"WSLC";
```
OneBlue, line 52: `Settings->DisplayName` should never be null at that point

OneBlue: Also, do we know if there are any limitations on `systemSettings.Owner` ? Are string size, or special characters limited ?

kvega005 (approved):

PR #40348, opened 2026-04-28
Add a temporary fix to prevent VS project load failures due to 'AnyCpu' targets

OneBlue:

This change introduces a temporary opt-out flag to skip the C# version of the SDK, since it breaks loading the project with visual studio: 
image

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,6 +172,10 @@ if (NOT DEFINED WSL_BUILD_WSL_SETTINGS)
     set(WSL_BUILD_WSL_SETTINGS false)
 endif ()
 
+if (NOT DEFINED WSL_BUILD_SDKCS)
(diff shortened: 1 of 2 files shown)


AmelBawa-msft (approved):

JohnMcPMS (approved):
🤖 comment on the commented out user config should be applied, but isn't critical as I assume this will be reverted with the fix soon enough.

OneBlue:
> 🤖 comment on the commented out user config should be applied, but isn't critical as I assume this will be reverted with the fix soon enough.

Agreed, and since we don't actually that binary anywhere yet, just switched to disabled by default instead, so contributors don't have to touch `UserConfig.cmake` for the fix to work

I **think** this won't break the existing nuspec, but we'll see what the CI says

florelis (approved):

PR #40360, opened 2026-04-29
Implement WinRT / C# projection for the WSLC SDK API surface

florelis:

This PR the C++/WinRT wrapper for the WSLC SDK, which is then projected to C#.

## PR Checklist

This builds on top of #40121 and #40212. The relevant changes for this are only the `.h` and `.cpp` files.

The implementation for all types follows a similar pattern, but there are some differences between types that are meant to be constructed by the consumer (e.g., `SessionSettings`) and types that can only be obtained by a function call (e.g., `Session`).

For types constructed by the consumer:
* Added constructors to the `implementation` types that don't have one in the public API
* Declared private fields to back all the properties
* All getters and setters use the backing fields directly
* Setters perform minimal validation (e.g., non-empty strings)
* String values are not stored as a WinRT `hstring` but as a native `std::(w)string` to directly back any string pointers passed to the C API.
* Properties that in the C API use 0 to represent the default value, here use instead an `IReference<>` (projected to C# as a nullable type) and `nullptr` to represent the default.
* Added a `ToStruct()` or `ToStructPointer()` (depending on usage) to get and retrieve the equivalent `struct` for the C API. This native `struct` is cached so that we only create it once. To avoid dealing with updating the native object if a property is changed, I blocked all setters if the native `struct` was already created.

For the types that are only constructed by the API:
* Added a backing field for pointer to the handle returned by the API
* Most functions are just forwarding parameters to the handle
* The destructor takes care of releasing them

For the few properties that are lists (e.g., `ContainerPortMapping`), I created two backing vectors. One is for the WinRT objects that are used when interacting with the consumer, and the other is for backing the pointers passed to the C API.

For callbacks, we use events that are configured before calling `Start()` on the object, and we pass the C API a callback that invokes the event handler.

The tests are more or less a direct translation of the tests for the C API. The tests show some issues that I have not resolved yet.


benhillis:

src/windows/WslcSDK/winrt/Container.cpp
```diff
@@ -15,38 +15,125 @@ Module Name:
...
+    return static_cast<winrt::Microsoft::WSL::Containers::ContainerState>(state);
+}
+
+WslcContainer Container::ToHandle()
+{
+    EnsureCreated();
+    return m_container.get();
+}
+
+void Container::EnsureCreated()
```
JohnMcPMS: Per-discussion, want to add a ProcessSettings property that specifies the output handling type so that Container can be created after C SDK call always having the handle.  Possibly as a follow up PR, although I'm liking waiting less as I go through this one.

src/windows/WslcSDK/winrt/Container.cpp
```diff
@@ -15,38 +15,125 @@ Module Name:
...
+    {
+        WslcProcess initHandle;
+        winrt::check_hresult(WslcGetContainerInitProcess(m_container.get(), &initHandle));
+        m_initProcess->AttachHandle(initHandle);
+    }
+
+    releaseRef.release();
 }
+
 void Container::Stop(winrt::Microsoft::WSL::Containers::Signal const& signal, uint32_t timeoutSeconds)
```
JohnMcPMS: Missed in IDL review: There is a TimeSpan type that we should use and can document the second resolution if needed.  Applies to any other time spans.

src/windows/WslcSDK/winrt/ContainerNamedVolume.cpp
```diff
@@ -17,32 +17,87 @@ Module Name:
...
 namespace winrt::Microsoft::WSL::Containers::implementation {
-ContainerNamedVolume::ContainerNamedVolume(hstring const& name, hstring const& containerPath, bool readOnly)
+
+ContainerNamedVolume::ContainerNamedVolume(hstring const& name, hstring const& containerPath, bool readOnly) :
+    m_name(winrt::to_string(name)), m_containerPath(winrt::to_string(containerPath)), m_readOnly(readOnly)
 {
-    throw hresult_not_implemented();
+    if (name.empty())
+    {
+        throw hresult_invalid_argument(L"Volume name cannot be empty");
```
JohnMcPMS, line 26: May need to be localized.  I would suggest at least tagging these with some sort of macro for now, like `WSLC_LOC_DEBT(L"...")` so that they can be found in the future easily.

src/windows/WslcSDK/winrt/ContainerNamedVolume.cpp
```diff
@@ -17,32 +17,87 @@ Module Name:
...
-    throw hresult_not_implemented();
+    return winrt::to_hstring(m_name);
 }
+
 void ContainerNamedVolume::Name(hstring const& value)
 {
-    throw hresult_not_implemented();
+    if (m_containerNamedVolume)
+    {
+        throw hresult_illegal_state_change(L"Cannot change volume name after the options have been applied");
```
JohnMcPMS:
```suggestion
        throw hresult_illegal_state_change(L"Cannot change value after options have been applied");
```
If these are to be localized, it would probably be better if all of these were the same string, like.

JohnMcPMS (commented):
Still need to look over the 3 complex files tomorrow.

src/windows/WslcSDK/winrt/Streams.cpp
```diff
@@ -0,0 +1,76 @@
...
+    if (options != InputStreamOptions::None)
+    {
+        throw winrt::hresult_error(HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED), L"Only InputStreamOptions::None is supported");
+    }
+
+    // Move to a background thread, ensuring that this object stays alive until the async operation completes.
+    auto self = get_strong();
+    co_await winrt::resume_background();
+
+    THROW_HR_IF(E_BOUNDS, count > buffer.Capacity());
```
JohnMcPMS: Check for parameter errors before first suspension.

Also check for null buffer as it seems like it is allowed by the interface contract.

src/windows/WslcSDK/winrt/Session.cpp
```diff
@@ -14,73 +14,309 @@ Module Name:
...
+    winrt::check_hresult(WslcSetSessionSettingsTerminationCallback(GetStructPointer(m_settings), TerminatedCallback, /* context */ this));
+
+    wil::unique_cotaskmem_string errorMessage;
+    auto hr = WslcCreateSession(GetStructPointer(m_settings), m_session.put(), errorMessage.put());
+    THROW_MSG_IF_FAILED(hr, errorMessage);
+    m_settings = nullptr;
+
+    // This object needs to stay alive for as long as the callbacks may be invoked even if all other references to it are dropped.
+    // We increase its ref count here, and decrease it once the session terminates.
+    AddRef();
```
JohnMcPMS: If there is a race between releasing the session handle and the termination callback, we should fix that in the C SDK or below.

I traced into the code and the callback invocation cannot happen after the session object has been destroyed without a bug in the runtime (it would be receiving a callback into a destroyed object).  I think there actually IS a race there, but I will create a PR to fix it separately.

I don't think we need to do this dance; as long as we release the session handle in the destructor.

src/windows/WslcSDK/winrt/Session.cpp
```diff
@@ -14,73 +14,309 @@ Module Name:
...
+
+    auto self = get_strong(); // keep session alive across suspension
+    co_await winrt::resume_background();
+
+    auto context = ProgressCallbackHelper<winrt::Microsoft::WSL::Containers::ImageProgress>{co_await winrt::get_progress_token()};
+
+    auto uri = winrt::to_string(options.Uri());
+    auto auth = winrt::to_string(options.RegistryAuth());
+
+    WslcPullImageOptions pullOptions{};
```
JohnMcPMS: I get that we do not want to leave callbacks on the internal options from the WinRT type, but we could start with a copy of its struct and set only the callback data.  That avoids the round trip conversions for the URI and auth.

src/windows/WslcSDK/winrt/Session.cpp
```diff
@@ -14,73 +14,309 @@ Module Name:
...
+
+IVectorView<winrt::Microsoft::WSL::Containers::ImageInfo> Session::Images()
+{
+    EnsureStarted();
+
+    WslcImageInfo* imagesArrayPtr = nullptr;
+    uint32_t count = 0;
+    winrt::check_hresult(WslcListSessionImages(ToHandle(), &imagesArrayPtr, &count));
+
+    // We can't pass this directly to WslcListSessionImages because the field for size is of a different type.
```
JohnMcPMS: See `wil::unique_cotaskmem_array_ptr::size_address<T>`.

src/windows/WslcSDK/winrt/Session.cpp
```diff
@@ -14,73 +14,309 @@ Module Name:
...
+    wil::unique_cotaskmem_array_ptr<WslcImageInfo> imagesArray{imagesArrayPtr, count};
+
+    auto images = winrt::single_threaded_vector<winrt::Microsoft::WSL::Containers::ImageInfo>();
```
JohnMcPMS: I believe it is much more efficient to create a `std::vector` with the appropriate capacity up front, then move that into the `winrt::single_threaded_vector` when done.

src/windows/WslcSDK/winrt/Process.cpp
```diff
@@ -14,59 +14,291 @@ Module Name:
...
 #include "precomp.h"
 #include "Process.h"
+#include "Streams.h"
 #include "Microsoft.WSL.Containers.Process.g.cpp"
 
 namespace winrt::Microsoft::WSL::Containers::implementation {
+
+Process::Process(WslcProcess process) : m_process(process)
+{
+    StartExitThread();
```
JohnMcPMS: Who is creating a process that is already running?  The container initial process should exist before the container is started, but in a special form that doesn't allow the caller to start the process directly.  It needs to have the process handle applied to it on container start.

Any later processes should have the container and settings via the other constructor.  I don't see any point at which we will need to construct a process object from the handle, as we don't have an API to attach to a running process.

src/windows/WslcSDK/winrt/Process.cpp
```diff
@@ -14,59 +14,291 @@ Module Name:
...
+    if (m_errorReceivedEvent)
+    {
+        callbacks.onStdErr = OutputCallback;
+    }
+
+    winrt::check_hresult(WslcSetProcessSettingsCallbacks(settings, &callbacks, this));
+
+    // This object needs to stay alive for as long as the callbacks may be invoked even if all other references to it are dropped.
+    // We increase its ref count here, and decrease it once the process exits.
+    AddRef();
```
JohnMcPMS: Same thing as the session, we shouldn't need to do this dance.  And I already know that this one is good to rely on the process release.

src/windows/WslcSDK/winrt/Process.cpp
```diff
@@ -14,59 +14,291 @@ Module Name:
...
+    }
+}
+
+void Process::StartExitThread()
+{
+    winrt::check_hresult(WslcGetProcessExitEvent(m_process.get(), m_exitEventHandle.put()));
+
+    // Start a background thread that waits for exit and fires the WinRT event.
+    m_destructedEvent.create(wil::EventOptions::ManualReset);
+    m_exitThread = std::thread(
```
JohnMcPMS: Use a thread pool wait to reduce the need for a private thread just for waiting.

src/windows/WslcSDK/winrt/Process.cpp
```diff
@@ -14,59 +14,291 @@ Module Name:
...
-    throw hresult_not_implemented();
+    wil::unique_handle handle;
+    winrt::check_hresult(WslcGetProcessIOHandle(ToHandle(), WSLC_PROCESS_IO_HANDLE_STDIN, handle.put()));
+    return winrt::make<IOHandleOutputStream>(std::move(handle));
 }
+
 winrt::event_token Process::OutputReceived(winrt::Microsoft::WSL::Containers::ProcessOutputHandler const& handler)
 {
-    throw hresult_not_implemented();
+    // Callbacks can only be registered before the process is started.
```
JohnMcPMS: Remove with explicit use property.

src/windows/WslcSDK/winrt/Session.cpp
florelis: I never noticed anything about a race in the C SDK or below.

I'm doing this because I wanted to have the behavior of "if the WinRT object is destroyed, I do not necessarily want the C object to terminate the session", but maybe we don't actually want that. But I think we want that at least for processes, to be able to do some kind of "fire and forget" and not have to keep the WinRT process object around.

If we terminate the C Session during the WinRT Session destructor, we get some lifetime issues even if there is no race at the C API. The flow would be:
1. WinRT Session destructor starts; WinRT Session is no longer valid
2. C Session starts to be terminated
3. C Session calls its termination callback. The callback had the WinRT Session as the context, but it is no longer valid
4. C Session finishes termination / destruction

But I guess this can be resolved by having a separate object work as the context for the callback, and just ensuring that it is destroyed after the C session.

src/windows/WslcSDK/winrt/ContainerNamedVolume.cpp
florelis: Does it need to be localized? I get the need to localize errors that may need to be shown to end users, but all of the errors at this level are going to be programmer errors where they pass in bad parameters. I would expect these to almost always happen during development, where it may not be as big of a deal if they are not localized.

src/windows/WslcSDK/winrt/Session.cpp
JohnMcPMS: > I'm doing this because I wanted to have the behavior of "if the WinRT object is destroyed, I do not necessarily want the C object to terminate the session", but maybe we don't actually want that.

I don't think we want that.  If we needed it for safety reasons, then I would agree that we should do it, not that we want it.  Even then my goal would be that the session has been signaled on WinRT object destruction and the "leak" is just until it tears down all the way.

Based on my looking into the code, I don't believe that session release will return until the termination callback has completed.  The WinRT object will still exist in memory at that time, but its state may be indeterminate depending on implementation.  The important thing would be to ensure that the session handle is released *before* the termination event (and anything we use in the callback path) is destructed.

src/windows/WslcSDK/winrt/ContainerNamedVolume.cpp
JohnMcPMS: @benhillis , what is your/the WSL stance on localizing for this situation?

src/windows/WslcSDK/winrt/Session.cpp
florelis: The exit callback will signal the WinRT event, and the WinRT event handler takes the WinRT Session as a parameter. So the callback has to be called while the object is still alive, which has to be before the destructor and even before the `final_release`.

For Session, I think I'll remove the sender from the handler as there is nothing that we can use from it at that point AFAICT

For Process, I'm thinking we could have the process continue if the WinRT Process is destroyed, and just stop handling the exit event.

JohnMcPMS: I see, the sender portion.  Yes, that is a complicating factor.  If we wanted to keep it present, we probably would need something like a weak ref to the WinRT Session held inside the implementation session so that we could detect the pre or post object validity states.  It seems easier to just not have it and rely on callers to attach their own context to the callback state.

JohnMcPMS: The process will already continue on if the handle is released and the C SDK will disconnect the callbacks safely.  I think that one doesn't require any extra consideration, unless again we have the sender.  I would still suggest handling it by requiring us to get the lifetime correct without extra add/release and then detect the pre and post WinRT validity if needed for sender.

src/windows/WslcSDK/winrt/Container.cpp
```diff
@@ -15,38 +15,92 @@ Module Name:
...
+#include "Process.h"
+#include "ProcessSettings.h"
+
+using namespace winrt::Windows::Foundation;
 
 namespace winrt::Microsoft::WSL::Containers::implementation {
-void Container::Start(winrt::Microsoft::WSL::Containers::ContainerStartFlags const& flags)
+Container::Container(WslcSession session, winrt::Microsoft::WSL::Containers::ContainerSettings const& settings) :
+    m_initProcess(winrt::make_self<implementation::Process>(settings.InitProcess()))
 {
```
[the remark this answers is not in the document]
JohnMcPMS: The C SDK allows for no init process, as does I think the runtime.  The C SDK sends a default initialized process settings which appears to result in a bunch of things simply not being set in the internal request.  I suspect that leads to no init process being launched on container start.

It feels like we should model this as: settings.InitProcess is null, so m_initProcess is null.

src/windows/WslcSDK/winrt/ContainerSettings.h
```diff
@@ -49,7 +49,7 @@ struct ContainerSettings : ContainerSettingsT<ContainerSettings>
 private:
     std::string m_imageName;
     std::string m_name;
-    winrt::Microsoft::WSL::Containers::ProcessSettings m_initProcess{nullptr};
+    winrt::Microsoft::WSL::Containers::ProcessSettings m_initProcess{};
```
JohnMcPMS: Empty (nothing set) process settings are indistinguishable from null in the runtime, so if we want to support null means no process then we have to support "no cmdline means no process" as well I think.

I would argue that it comes down to how the attempt to get the init process handle goes when we don't have one.  Either we have to:

1. Predict when process will actually be created so we can match the Process object creation state to match
2. Always create a Process object and handle the error/null handle state of no process being created

src/windows/WslcSDK/winrt/PushImageOptions.cpp
```diff
@@ -72,19 +62,13 @@ void PushImageOptions::RegistryAuth(hstring const& value)
     m_registryAuth = winrt::to_string(value);
 }
 
-WslcPushImageOptions* PushImageOptions::ToStructPointer()
+WslcPushImageOptions PushImageOptions::ToStruct()
```
JohnMcPMS, line 75: We should keep the member and return a copy of it to prevent future changes as you had previously.

Applies to other places like this.

src/windows/WslcSDK/winrt/Process.cpp
```diff
@@ -19,44 +19,59 @@ Module Name:
...
-    winrt::check_hresult(WslcSetProcessSettingsCallbacks(settings, &callbacks, this));
+    wil::unique_handle exitEventHandle;
+    winrt::check_hresult(WslcGetProcessExitEvent(ToHandle(), exitEventHandle.put()));
 
-    // This object needs to stay alive for as long as the callbacks may be invoked even if all other references to it are dropped.
-    // We increase its ref count here, and decrease it once the process exits.
-    AddRef();
-    m_hasExitCallback = true;
+    auto weak_this = get_weak();
+    co_await winrt::resume_on_signal(exitEventHandle.get());
```
JohnMcPMS, line 77: The bad thing about this is that we cannot cancel it, so if I create a large number of processes that I release without waiting for exit, I can have a large number of these (presumably threadpool waits).  The event may be kept alive, but no one will be signaling it when the process does exit.

Creating manual thread pool waits allows us to cancel them when the process object is destructed.  If there is some other form of coroutine output object that similarly allows us to cancel the wait, I'm fine with that as well.

florelis: Added cancellation with `enable_propagation()` as discussed offline.

src/windows/WslcSDK/winrt/ContainerSettings.h
florelis: Changed container to not create initProcess if the settings are null, and the process settings to require a non empty cmdline

src/windows/WslcSDK/winrt/Streams.cpp
JohnMcPMS, line 54: Check for `m_handle` before using now that it can be closed.

src/windows/WslcSDK/winrt/ProcessSettings.h
```diff
@@ -47,7 +47,7 @@ struct ProcessSettings : ProcessSettingsT<ProcessSettings>
 
 private:
     std::string m_workingDirectory;
-    winrt::Windows::Foundation::Collections::IVector<hstring> m_cmdLine{winrt::single_threaded_vector<hstring>()};
+    winrt::Windows::Foundation::Collections::IVector<hstring> m_cmdLine{nullptr};
```
JohnMcPMS: Feels like we should still have the default empty vector, it is odd that environment variables has the empty container but cmdline doesn't.

src/windows/WslcSDK/winrt/ProcessSettings.cpp
```diff
@@ -68,6 +68,11 @@ void ProcessSettings::CmdLine(winrt::Windows::Foundation::Collections::IVector<h
         throw winrt::hresult_error(E_POINTER, L"CmdLine cannot be null");
     }
 
+    if (value.Size() == 0)
```
JohnMcPMS: We shouldn't block resetting to the original state.  This also doesn't prevent an empty vector later: the user can set a non-empty vector then clear it out.

src/windows/WslcSDK/winrt/ContainerSettings.h
```diff
@@ -49,7 +49,7 @@ struct ContainerSettings : ContainerSettingsT<ContainerSettings>
 private:
     std::string m_imageName;
     std::string m_name;
-    winrt::Microsoft::WSL::Containers::ProcessSettings m_initProcess{};
+    winrt::Microsoft::WSL::Containers::ProcessSettings m_initProcess{nullptr};
```
JohnMcPMS, line 52: I prefer that we pre-existing objects for everything so that the user can just set properties without an intermediate step to create the settings object.

src/windows/WslcSDK/winrt/Container.cpp
```diff
@@ -22,9 +22,13 @@ Module Name:
 using namespace winrt::Windows::Foundation;
 
 namespace winrt::Microsoft::WSL::Containers::implementation {
-Container::Container(WslcSession session, winrt::Microsoft::WSL::Containers::ContainerSettings const& settings) :
-    m_initProcess(winrt::make_self<implementation::Process>(settings.InitProcess()))
+Container::Container(WslcSession session, winrt::Microsoft::WSL::Containers::ContainerSettings const& settings)
 {
+    if (settings.InitProcess())
```
JohnMcPMS, line 27: Maybe something like `if (settings.InitProcess() && get_impl(settings.InitProcess()).WillCreateInitProcess())`

src/windows/WslcSDK/winrt/Process.cpp
```diff
@@ -82,7 +89,7 @@ void Process::AttachHandle(WslcProcess handle)
     }
 
     m_process.reset(handle);
-    StartWaitingForExit();
+    m_waitForExitAction = StartWaitingForExitAsync();
```
JohnMcPMS: Maybe `StartWaitingForExit` that is only `m_waitAction = StartWaitingForExitAsync();` so we don't accidentally not set the member someplace in the future.

src/windows/WslcSDK/winrt/Process.cpp
```diff
@@ -99,14 +106,20 @@ void Process::Start()
         throw winrt::hresult_illegal_method_call(L"Start() cannot be called on the init process, it is started by the container");
     }
 
+    auto cmdLine = GetImplementation(m_settings)->CmdLine();
```
JohnMcPMS: This could use the same implementation function as the choice of whether to create the init process or not.  A semantically defined function will be easier to maintain than manual checks for the presence of a cmd line.

florelis: This isn't in the init process path. For init, it is started by the container.
I added a function `EnsureCanStart()` that checks this.

src/windows/WslcSDK/winrt/Container.cpp
florelis: What would `WillCreateInitProcess()` check? Just cmdline? I thought containers could have a default init cmdline, but we would still want to set callbacks and stuff

JohnMcPMS (approved):
I think we are at the point where we will need samples / use to find rough edges.

PR #40373, opened 2026-05-01
Fix CodeSign validation for wslcsdkcs.dll in SDK output

benhillis:
**Problem:** Guardian CodeSign validation fails for \sdk/arm64/wslcsdkcs.dll\ with \CodeSign.MissingSigningCert\.

**Root cause:** \wslcsdkcs\ was in the first target group alongside \wslcsdk\. After that group is built and signed, the \Build remaining targets\ step (\cmake --build . --config Release\) rebuilds \wslcsdkcs.dll\ because MSBuild detects its dependency (\wslcsdk.dll\) was modified by signing. The unsigned rebuild overwrites the signed version in \in/\.

**Fix:** Move \wslcsdkcs\ to the \msipackage\ target stage. Since this stage runs after the first group is signed, \wslcsdkcs\ gets built against the already-signed \wslcsdk.dll\. When \Build remaining targets\ runs later, nothing has changed for \wslcsdkcs\ so it won't be rebuilt — the signed version stays intact.

Same class of issue as the wslsettings signing fix.

### DIFF
--- a/.pipelines/build-job.yml
+++ b/.pipelines/build-job.yml
@@ -212,8 +212,13 @@ jobs:
               if (Test-Path "generated\dev-cert.pfx") { Copy-Item "generated\dev-cert.pfx" "$(ob_outputDirectory)\installer\dev-cert.pfx" }
               New-Item -ItemType Directory -Path "$(ob_outputDirectory)\sdk\${{ parameters.platform }}" -Force
               Copy-Item -Path "bin\${{ parameters.platform }}\release\wslcsdk.lib" -Destination "$(ob_outputDirectory)\sdk\${{ parameters.platform }}\wslcsdk.lib"
-              Copy-Item -Path "bin\${{ parameters.platform }}\release\wslcsdk.dll" -Destination "$(ob_outputDirectory)\sdk\${{ parameters.platform }}\wslcsdk.dll"
-              Copy-Item -Path "bin\${{ parameters.platform }}\release\wslcsdkcs.dll" -Destination "$(ob_outputDirectory)\sdk\${{ parameters.platform }}\wslcsdkcs.dll"
+              if ('${{ parameters.isRelease }}' -eq 'True') {
+                  Copy-Item -Path "$(packageStagingDir)\${{ parameters.platform }}\wslcsdk.dll" -Destination "$(ob_outputDirectory)\sdk\${{ parameters.platform }}\wslcsdk.dll"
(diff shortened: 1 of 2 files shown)


.pipelines/build-job.yml
```diff
@@ -197,6 +197,16 @@ jobs:
                           }
                       }
 
+      - task: PowerShell@2
+        displayName: "Stage SDK outputs (${{ parameters.platform }})"
+        inputs:
+          targetType: inline
+          script: |
+              New-Item -ItemType Directory -Path "$(ob_outputDirectory)\sdk\${{ parameters.platform }}" -Force
```
OneBlue: Is this hitting the same issue as wslsettings ?

OneBlue: Given that wslcsdkcs doesn't need to be included in the msi package, we could also just move it to the last stage in `targets`, alongside `msipackage`. That way we won't need to manually stage it

benhillis: Yes it seems that way, true I'll do that.

.pipelines/build-stage.yml
```diff
@@ -32,12 +32,12 @@ parameters:
...
+      - target: "wsl;libwsl;wslg;wslservice;wslhost;wslrelay;wslinstaller;wslinstall;initramfs;wslserviceproxystub;wslsettings;wslinstallerproxystub;testplugin;wslcsession;wslc;wsltests;wslcsdk"
+        pattern: "wsl.exe,libwsl.dll,wslg.exe,wslservice.exe,wslhost.exe,wslrelay.exe,wslinstaller.exe,wslinstall.dll,wslserviceproxystub.dll,wslsettings/wslsettings.dll,wslsettings/wslsettings.exe,wslinstallerproxystub.dll,WSLDVCPlugin.dll,testplugin.dll,wsldeps.dll,wslcsession.exe,wslc.exe,wslcsdk.dll"
       - target: "msixgluepackage"
         pattern: "gluepackage.msix"
-      - target: "msipackage"
-        pattern: "wsl.msi"
+      - target: "msipackage;wslcsdkcs"
+        pattern: "wsl.msi,wslcsdkcs.dll"
 
   - name: pool
```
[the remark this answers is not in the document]
benhillis: This is addressed by the approach change. \wslcsdkcs\ now builds in the \msipackage\ stage — after \wslcsdk.dll\ is signed. So when 'Build remaining targets' runs, \wslcsdkcs.dll\ was already built against the signed dependency. MSBuild won't rebuild it (nothing changed), so the signed version stays in \in/\ when the SDK copy happens.

OneBlue (approved):

PR #40374, opened 2026-05-01
Set page_reporting_order via kernel command line

benhillis:
Move page reporting configuration from post-boot process launches and init messages to the kernel command line parameter (`page_reporting.page_reporting_order=N`). This ensures the correct order is set at boot time for both WSLC and WSL2.

- Remove `PageReportingOrder` from the early config message struct
- Simplify the memory reduction thread in init to always perform compaction when idle, removing the `PageReportingOrder` gate which was effectively dead code (the value was always non-zero in practice)
- Remove the post-boot echo process from WSLC
- Rename `m_coldDiscardShiftSize` to `m_pageReportingOrder` for clarity
- Align HcsVirtualMachine page reporting logic with WslCoreVm (derive order from the same version-gated block that sets BackingPageSize/FaultClusterSizeShift)

Verified via ETL trace that the parameter takes effect at boot (`hv_balloon: Cold memory discard hint enabled with order 5`).

### DIFF
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -290,7 +290,7 @@ try
     // and memory reclaim. This ensures that the maximum number of pages can be discarded to the host.
     //
 
-    std::thread([Mode = Mode]() mutable {
+    std::thread([Mode]() mutable {
         try
         {
             //
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -122,7 +122,7 @@ std::optional<bool> g_EnableSocketLogging;
 
 int Chroot(const char* Target);
 
-void ConfigureMemoryReduction(int PageReportingOrder, LX_MINI_INIT_MEMORY_RECLAIM_MODE Mode);
+void ConfigureMemoryReduction(LX_MINI_INIT_MEMORY_RECLAIM_MODE Mode);
 
 void CreateSwap(unsigned int Lun);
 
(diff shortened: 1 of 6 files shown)


src/linux/init/main.cpp
```diff
@@ -290,32 +286,11 @@ Return Value:
...
-    // N.B. Compaction is not needed if page reporting order is set to single page mode.
-    //
-
-    if (PageReportingOrder == 0 && Mode == LxMiniInitMemoryReclaimModeDisabled)
-    {
-        return;
-    }
 
-    std::thread([PageReportingOrder = PageReportingOrder, Mode = Mode]() mutable {
+    std::thread([Mode = Mode]() mutable {
```
benhillis: I think just [Mode] should be fine?

src/windows/service/exe/HcsVirtualMachine.cpp
```diff
@@ -108,6 +108,10 @@ HcsVirtualMachine::HcsVirtualMachine(_In_ const WSLCSessionSettings* Settings)
     // Enable timesync workaround to sync on resume from sleep in modern standby.
     kernelCmdLine += L" hv_utils.timesync_implicit=1";
 
+    // Configure page reporting order - minimum order of pages reported as free to the hypervisor.
+    int pageReportingOrder = (windowsVersion.BuildNumber >= WindowsBuildNumbers::Germanium) ? 5 : 9;
```
benhillis: Look at how WslCoreVm.cpp handles ComputeTopology.Memory, this should be the same I think.

src/shared/inc/lxinitshared.h
```diff
@@ -1251,7 +1251,6 @@ typedef struct _LX_MINI_INIT_EARLY_CONFIG_MESSAGE
     unsigned int SwapLun;
     LX_MINI_INIT_MOUNT_DEVICE_TYPE SystemDistroDeviceType;
     unsigned int SystemDistroDeviceId;
-    int PageReportingOrder;
     LX_MINI_INIT_MEMORY_RECLAIM_MODE MemoryReclaimMode;
     // IPv4 address stored in network byte order
     uint32_t DnsTunnelingIpAddress = 0;
```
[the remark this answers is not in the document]
benhillis: the components are shipped together, they cannot get out of sync.

OneBlue (approved):

PR #40375, opened 2026-05-01
Add swap support for WSLC virtual machines (separate ephemeral VHD)

benhillis:
Add swap support for WSLC virtual machines using a dedicated ephemeral VHDX.

## Design
- Swap size defaults to VM memory size (`MemoryMb`) — no additional SDK setting needed
- Creates a dynamic VHDX at `StoragePath/swap.vhdx` (low IO cost on boot since swap is rarely written)
- Attaches as a separate SCSI disk; `mkswap + swapon` runs fire-and-forget (async, best-effort)
- VHD is deleted on session teardown in `Terminate()`
- Completely separate from the storage VHD — no sizing interactions or shared I/O

## Why separate VHD?
Putting swap on the storage VHD had many issues: VHD sizing assumptions, pre-existing VHD compatibility, shared I/O bandwidth under memory pressure, disk-full cascades, and sync `fallocate` blocking session init. A dedicated ephemeral VHD avoids all of these.

## Why dynamic (not fixed) VHD?
A fixed VHD would do a large sequential write on every boot even though swap pages are rarely touched. A dynamic VHD allocates blocks on demand, making session startup faster while still providing the same swap capacity.

## Changes
- `wslc.idl`: Add `SwapSizeMb` to both `WSLCSessionSettings` and `WSLCSessionInitSettings`
- `WSLCSession.cpp`: Create/attach swap VHD in `ConfigureStorage()`, delete in `Terminate()`
- `WSLCSessionManager.cpp`: Plumb `MemoryMb` → `SwapSizeMb` in init settings
- `HcsVirtualMachine.cpp`: Move page reporting to runtime sysfs write (out of kernel cmdline)
- `WSLCVirtualMachine.cpp`: Configure page_reporting_order via sysfs after boot
- `WSLCTests.cpp`: Verify swap is active on a block device (with retry for async setup)

### DIFF
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -357,6 +357,25 @@ void WSLCSession::ConfigureStorage(const WSLCSessionInitSettings& Settings, PSID
     // Mount the device to /root.
     m_virtualMachine->Mount(diskDevice.c_str(), c_containerdStorage, "ext4", "", 0);
 
+    // Configure swap on a separate ephemeral VHD.
+    if (Settings.SwapSizeMb > 0)
+    {
+        try
+        {
+            m_swapVhdPath = storagePath / "swap.vhdx";
+            DeleteFileW(m_swapVhdPath.c_str()); // Remove stale swap from prior run
+            wsl::core::filesystem::CreateVhd(m_swapVhdPath.c_str(), static_cast<ULONGLONG>(Settings.SwapSizeMb) * _1MB, UserSid, false, true);
+
+            auto [_, swapDevice] = m_virtualMachine->AttachDisk(m_swapVhdPath.c_str(), false);
+
+            // Fire-and-forget: mkswap + swapon runs asynchronously since swap is best-effort.
+            auto cmd = std::format("/usr/sbin/mkswap {0} && /usr/sbin/swapon {0}", swapDevice);
+            ServiceProcessLauncher launcher("/bin/sh", {"/bin/sh", "-c", cmd});
+            launcher.Launch(*m_virtualMachine);
+        }
+        CATCH_LOG()
+    }
+
     deleteVhdOnFailure.release();
 }
 
@@ -2417,6 +2436,13 @@ try
     m_dockerdProcess.reset();
     m_containerdProcess.reset();
     m_virtualMachine.reset();
+
+    // Delete the ephemeral swap VHD now that the VM is gone.
+    if (!m_swapVhdPath.empty())
+    {
+        LOG_IF_WIN32_BOOL_FALSE(DeleteFileW(m_swapVhdPath.c_str()));
+        m_swapVhdPath.clear();
+    }
 
     m_terminated = true;
     return S_OK;
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -364,7 +364,7 @@ void WSLCSession::ConfigureStorage(const WSLCSessionInitSettings& Settings, PSID
         {
             m_swapVhdPath = storagePath / "swap.vhdx";
             DeleteFileW(m_swapVhdPath.c_str()); // Remove stale swap from prior run
-            wsl::core::filesystem::CreateVhd(m_swapVhdPath.c_str(), static_cast<ULONGLONG>(Settings.SwapSizeMb) * _1MB, UserSid, false, true);
+            wsl::core::filesystem::CreateVhd(m_swapVhdPath.c_str(), static_cast<ULONGLONG>(Settings.SwapSizeMb) * _1MB, UserSid, false, false);
 
             auto [_, swapDevice] = m_virtualMachine->AttachDisk(m_swapVhdPath.c_str(), false);
 
--- a/src/windows/wslcsession/WSLCVirtualMachine.cpp
+++ b/src/windows/wslcsession/WSLCVirtualMachine.cpp
@@ -295,6 +295,18 @@ void WSLCVirtualMachine::Initialize()
     // Configure GPU mounts if enabled
     MountGpuLibraries(c_gpuLibrariesPath, c_gpuDriversPath);
 
+    // Configure page reporting - set the minimum order of pages reported as free to the hypervisor.
+    {
+        const auto windowsVersion = wsl::windows::common::helpers::GetWindowsVersion();
+        int pageReportingOrder = (windowsVersion.BuildNumber >= wsl::windows::common::helpers::WindowsBuildNumbers::Germanium) ? 5 : 9; // 128k or 2MB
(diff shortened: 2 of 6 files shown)


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -357,6 +357,25 @@ void WSLCSession::ConfigureStorage(const WSLCSessionInitSettings& Settings, PSID
...
+    // Configure a swapfile on the storage filesystem.
+    try
+    {
+        auto swapFile = std::format("{}/swap", c_containerdStorage);
+        auto cmd = std::format(
+            "expected=$({1} * 1024 * 1024); "
+            "if [ ! -f {0} ] || [ \"$(stat -c %s {0} 2>/dev/null)\" != \"$expected\" ]; then "
+            "swapoff {0} 2>/dev/null || true; "
+            "rm -f {0} && fallocate -l {1}M {0} && chmod 600 {0} && /usr/sbin/mkswap {0}; "
+            "fi && /usr/sbin/swapon {0}",
```
benhillis: I think this should be tweaked, maybe deleting the existing swap file and create from scratch is a better approach.

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -295,21 +295,21 @@ void WSLCVirtualMachine::Initialize()
...
-    // Configure cold discard hint size for page reporting.
-    // This sets the minimum order of pages that will be reported as free to the hypervisor.
+    // Configure networking. This must happen after all filesystems are mounted since /gns needs to access /sys.
+    ConfigureNetworking();
+
+    // Configure page reporting - set the minimum order of pages reported as free to the hypervisor.
     {
         const auto windowsVersion = wsl::windows::common::helpers::GetWindowsVersion();
-        int pageReportingOrder = (windowsVersion.BuildNumber >= wsl::windows::common::helpers::WindowsBuildNumbers::Germanium) ? 5 : 9; // 128k or 2MB
+        int pageReportingOrder =
```
benhillis: undo this change.

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -295,21 +295,21 @@ void WSLCVirtualMachine::Initialize()
     // Configure GPU mounts if enabled
     MountGpuLibraries("/usr/lib/wsl/lib", "/usr/lib/wsl/drivers");
 
-    // Configure cold discard hint size for page reporting.
-    // This sets the minimum order of pages that will be reported as free to the hypervisor.
+    // Configure networking. This must happen after all filesystems are mounted since /gns needs to access /sys.
+    ConfigureNetworking();
```
benhillis, line 299: move this back where it was.

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -1197,7 +1197,7 @@ void WSLCVirtualMachine::CollectCrashDumps(wil::unique_socket&& listenSocket)
     // No impersonation needed - the session process already runs as the user.
     wslutil::SetThreadDescription(L"CrashDumpCollection");
 
-    const auto crashDumpFolder = filesystem::GetTempFolderPath(GetCurrentProcessToken()) / L"wslc-crashes";
+    const auto crashDumpFolder = filesystem::GetTempFolderPath(GetCurrentProcessToken()) / L"wslc" / L"crashes";
```
benhillis: go back to wslc-crashes here and in the tests.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2418,6 +2436,13 @@ try
     m_containerdProcess.reset();
     m_virtualMachine.reset();
 
+    // Delete the ephemeral swap VHD now that the VM is gone.
+    if (!m_swapVhdPath.empty())
+    {
+        LOG_IF_WIN32_BOOL_FALSE(DeleteFileW(m_swapVhdPath.c_str()));
+        m_swapVhdPath.clear();
```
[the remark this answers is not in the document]
benhillis: eww

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -357,6 +357,25 @@ void WSLCSession::ConfigureStorage(const WSLCSessionInitSettings& Settings, PSID
...
     m_virtualMachine->Mount(diskDevice.c_str(), c_containerdStorage, "ext4", "", 0);
 
+    // Configure swap on a separate ephemeral VHD.
+    if (Settings.SwapSizeMb > 0)
+    {
+        try
+        {
+            m_swapVhdPath = storagePath / "swap.vhdx";
+            DeleteFileW(m_swapVhdPath.c_str()); // Remove stale swap from prior run
+            wsl::core::filesystem::CreateVhd(m_swapVhdPath.c_str(), static_cast<ULONGLONG>(Settings.SwapSizeMb) * _1MB, UserSid, false, true);
```
OneBlue: I'm not sure that we want a fixed VHD here. It will make swap access faster, but allocated a fixed VHD on every boot will do a lot of IO. I would have a preference for a dynamic VHD here (especially since we expect swap to be rarely written to)

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -357,6 +357,25 @@ void WSLCSession::ConfigureStorage(const WSLCSessionInitSettings& Settings, PSID
     // Mount the device to /root.
     m_virtualMachine->Mount(diskDevice.c_str(), c_containerdStorage, "ext4", "", 0);
 
+    // Configure swap on a separate ephemeral VHD.
+    if (Settings.SwapSizeMb > 0)
+    {
+        try
+        {
+            m_swapVhdPath = storagePath / "swap.vhdx";
+            DeleteFileW(m_swapVhdPath.c_str()); // Remove stale swap from prior run
```
OneBlue, line 366: Would it be possible to pass `OPEN_VIRTUAL_DISK_FLAG_BLANK_FILE` when we create the VHD to avoid having to delete the file before recreating it ?

src/windows/wslcsession/WSLCSession.cpp
OneBlue: Looking through the implementation of CreateVhd, we might want to plumb through `OPEN_VIRTUAL_DISK_FLAG_CACHED_IO` to make the swap a bit faster, since we don't care about data consitency after a reboot

src/windows/wslcsession/WSLCSession.cpp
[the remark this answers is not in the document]
benhillis: swap is best effort

OneBlue (approved):
LGTM, I wonder if we'll ever need a flag to enable / disable swap, but we can add support for that when the need comes

PR #40376, opened 2026-05-01
Share memory reduction logic and enable compaction for WSLC

benhillis:
## Summary

This PR refactors the Linux init memory-reduction logic so it can be shared between the WSL2 init (`main.cpp`) and WSLC init (`WSLCInit.cpp`), and enables idle-time compaction + cache dropping for WSLC VMs.

## Changes

- Extracts the memory reduction thread logic into a shared `StartMemoryReductionThread(LX_MINI_INIT_MEMORY_RECLAIM_MODE)` entry point in `util.cpp`.
- Moves `GetUserCpuTime` and `GetMemoryInUse` helpers to `util.cpp` as file-local (`static`) functions.
- WSL2 init calls `StartMemoryReductionThread` directly instead of going through `ConfigureMemoryReduction()`.
- Starts the memory reduction thread in WSLC init with drop-caches + compaction behavior when idle.
- Uses `LX_MINI_INIT_MEMORY_RECLAIM_MODE` wire enum directly rather than introducing a separate enum.

## Notes

- Memory compaction always runs when the VM is idle, regardless of the reclaim mode setting. Compaction and reclaim are independent concerns — compaction coalesces free pages for efficient page reporting, while reclaim controls cache/memory pressure behavior.


src/linux/init/WSLCInit.cpp
```diff
@@ -1051,6 +1051,12 @@ int WSLCEntryPoint(int Argc, char* Argv[])
         SetCloseOnExec(fd, true);
     }
 
+    //
+    // Start memory reduction thread (compaction + drop_caches when idle).
+    //
+
+    StartMemoryReductionThread(LxMiniInitMemoryReclaimModeDropCache);
```
OneBlue: I think we should wait after we chroot to start this since we move the procfs when we do that (otherwise the memory reduction thread could fail to access it after the mount is moved but before we chroot)

benhillis: Good call.

benhillis: Done in 6ebc3140 — moved the call inside the WSLC_MOUNT chroot block, guarded by a once_flag so it only fires after the first successful chroot.

ptrivedi (approved):

PR #40380, opened 2026-05-01
Move printk.devkmsg and page_reporting_order to kernel command line

benhillis:
Move `printk.devkmsg=on` from post-boot runtime writes in init to the kernel command line, ensuring it is applied at boot for both WSL2 and WSLC VMs.

Also adds `AppendCommonKernelCommandLine()` helper in `helpers.hpp`/`helpers.cpp` to share common kernel parameters (`hv_utils.timesync_implicit=1`, `printk.devkmsg=on`, and `page_reporting.page_reporting_order`) between WslCoreVm and HcsVirtualMachine.

Verified via `/proc/cmdline` and `/proc/sys/kernel/printk_devkmsg` that the parameter takes effect at boot.

### DIFF
--- a/src/linux/init/WSLCInit.cpp
+++ b/src/linux/init/WSLCInit.cpp
@@ -948,15 +948,6 @@ int WSLCEntryPoint(int Argc, char* Argv[])
     }
 
     //
-    // Disable rate limiting of user writes to dmesg.
-    //
-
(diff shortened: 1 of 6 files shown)


benhillis:
This will have to be merged with https://github.com/microsoft/WSL/pull/40374 (move the page_reporting_order kernel command line to this helper)

OneBlue (dismissed):

OneBlue (approved):

PR #40384, opened 2026-05-01
Add per-container resource limits and upstream sync workflow

kvega005:
Introduce functionality for setting per-container memory limits, CPU quotas, and ulimits.

## PR Checklist


src/windows/inc/wslc_schema.h
```diff
@@ -49,11 +49,23 @@ struct InspectState
     NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(InspectState, Status, Running, ExitCode, StartedAt, FinishedAt);
 };
 
+struct InspectUlimit
```
OneBlue: nit: Could we match the type names between `docker_schema` and `wslc_schema` ? That would make comparing them easier (looks like the schemas already deviated a bit, but we can fix this in a separate change)

OneBlue (dismissed):

OneBlue (approved):

PR #40385, opened 2026-05-01
CLI: Initialize shm-size and configure stop-signal

AmelBawa-msft:
- Added --shm-size and --stop-signal options to container create and container run
- Input validation
- E2E tests for both options on create and run, including verifying the shm mount size inside the container.

## PR Checklist


src/windows/wslc/services/ContainerModel.h
```diff
@@ -61,7 +63,7 @@ struct StopContainerOptions
 {
     static constexpr LONG DefaultTimeout = -1;
 
-    WSLCSignal Signal = WSLCSignalSIGTERM;
```
AmelBawa-msft, line 64: Instead of hardcoding a default signal value, `stop` command should stop the container with the pre-configured `stop-signal`

localization/strings/cs-CZ/Resources.resw
OneBlue, line 2632: This part of the diff seems unrelated (same for other resw files)

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -206,4 +214,24 @@ void ValidateGpus(const std::vector<std::wstring>& values, const std::wstring& a
...
+{
+    for (const auto& value : values)
+    {
+        std::ignore = GetMemorySizeFromString(value, argName);
+    }
+}
+
+ULONGLONG GetMemorySizeFromString(const std::wstring& input, const std::wstring& argName)
+{
+    auto narrowInput = wsl::windows::common::string::WideToMultiByte(input);
```
OneBlue: This conversion isn't needed (ParseMemorySize is templated)

benhillis (dismissed):
Looks like you picked up a bunch of localization changes in the merge.

localization/strings/cs-CZ/Resources.resw
AmelBawa-msft: @copilot remove WSLCCLI_SignalArgDescription entry from the resw files in the localization directoy, keep it only in the 'localization/strings/en-US/Resources.resw‎'. The string entry will be re-added in the other resw files after the PR is merged.

AmelBawa-msft:
> Looks like you picked up a bunch of localization changes in the merge.

@benhillis I modified an existing localized string that previously accepted an argument, but it no longer does. Without updating the corresponding entries in the other localization files, the project would fail to build. After discussing this offline with @OneBlue, I’ll remove the entire entry for now so the localization can be reapplied later.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -206,4 +214,24 @@ void ValidateGpus(const std::vector<std::wstring>& values, const std::wstring& a
...
+    auto parsed = wsl::shared::string::ParseMemorySize(narrowInput.c_str());
```
AmelBawa-msft:
```suggestion
    auto parsed = wsl::shared::string::ParseMemorySize(input.c_str());
```

benhillis (approved):

PR #40386, opened 2026-05-01
Show only the last few log lines while building

johnstep:
Reduce the build output in non-verbose mode by showing only the last few lines of the container log, in a scrolling window that displays the lines dimmed and collapses on completion.


src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -16,14 +16,195 @@ Module Name:
...
 }
 CATCH_RETURN();
+
+void BuildImageCallback::Redraw()
+{
+    CONSOLE_SCREEN_BUFFER_INFO info{};
+    THROW_IF_WIN32_BOOL_FALSE(GetConsoleScreenBufferInfo(m_console, &info));
+    // Use the visible window width (not buffer width), minus one column to avoid the
+    // deferred-wrap edge case when a line is exactly the window width.
+    const SHORT consoleWidth = std::max<SHORT>(0, info.srWindow.Right - info.srWindow.Left);
```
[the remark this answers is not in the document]
johnstep: This is not worth changing as the output is not useful in either case.

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -16,14 +16,199 @@ Module Name:
...
+        {
+            WriteTerminal(MultiByteToWide(line) + L'\n');
+        }
+    }
+}
+
+void BuildImageCallback::WriteTerminal(std::wstring_view content) const
+{
+    DWORD written;
+    LOG_IF_WIN32_BOOL_FALSE(WriteConsoleW(m_console, content.data(), static_cast<DWORD>(content.size()), &written, nullptr));
```
OneBlue: Out of curiosity, what's the reason for using WriteConsoleW here ? 

Asking because we usually use `wprintf()`, writing to the handle directly could lead to weird behavior if `wprintf` had unflushed content (even though I don't think that's the case in this change)

src/windows/wslc/services/BuildImageCallback.h
```diff
@@ -12,13 +12,51 @@ Module Name:
...
+    ~BuildImageCallback();
     HRESULT OnProgress(LPCSTR status, LPCSTR id, ULONGLONG current, ULONGLONG total) override;
+
+    void SetVerbose(bool verbose)
+    {
+        m_verbose = verbose;
+    }
+
+    // The handle must remain valid for the lifetime of this callback.
+    void SetCancelEvent(HANDLE cancelEvent)
```
OneBlue: Given that the cancel event is always set, I think it would be OK to take it as a required constructor parameter and assume that it's always non-null

OneBlue: (From a UX perspective, we probably won't want an un-cancellable build)

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -16,14 +16,199 @@ Module Name:
...
+        WriteTerminal(MultiByteToWide(status));
+        return S_OK;
+    }
+
+    // Log line: add to the scrolling window.
+    for (const char* p = status; *p != '\0'; ++p)
+    {
+        if (*p == '\n')
+        {
+            m_allLines.push_back(m_pendingLine);
```
OneBlue: nit: We might want to turn this into a single string that already contains the line feed, that way it would be easier to have a hard limit on how much output we keep in memory, so we don't run out in the case of an extremely verbose build failure. 

Capping at something like `10MB` feels like a reasonable default

OneBlue (dismissed):
LGTM, minor comments

OneBlue (approved):

PR #40406, opened 2026-05-04
Don't run the VM owner tests since custom VM owners are only support on Windows >= 11

OneBlue:

This change skips the VM owner tests since they fail on vb and fe because that feature isn't supported there.

## PR Checklist


benhillis:
The issue is just that hcsdiag.exe does not support raw:
https://github.com/microsoft/WSL/pull/40407

PR #40407, opened 2026-05-04
Fix WSLCTests::ListVms parser for older Windows hcsdiag

benhillis:
`WSLCTests::ListVms()` originally called `hcsdiag list -raw` and parsed the JSON output. That fails on older Windows builds (e.g. Win10 22H2 / build 19041), where the in-box `hcsdiag` doesn't recognize `-raw` and just prints its help text — so the JSON parse silently produced a discarded value, `ListVms()` returned an empty vector, and three tests failed in the `vb_release` lab image with `VM with owner 'X' not found`:

- `WSLCTests::VmOwnerMatchesSessionDisplayName`
- `WSLCTests::VmKillTerminatesSession`
- `WSLCTests::VmKillFailsInFlightOperations`

Per review feedback from @OneBlue, replace the `hcsdiag.exe` shellout with a direct call to `HcsEnumerateComputeSystems` (which is what `hcsdiag` does internally anyway). This avoids the older-Windows compatibility problem entirely, gives us a guaranteed JSON contract from the HCS API, drops a process launch, and matches the existing `wsl::windows::common::hcs` helper conventions (`CreateOperation` + `HcsWaitForOperationResult` + `ExecutionContext(Context::HCS)`). Also adds a `LogInfo` of the HCS result document so future failures have the raw enumeration captured next to the assertion.

### DIFF
--- a/test/windows/WSLCTests.cpp
+++ b/test/windows/WSLCTests.cpp
@@ -18,6 +18,8 @@ Abstract:
 #include "WSLCProcessLauncher.h"
 #include "WSLCContainerLauncher.h"
 #include "WslCoreFilesystem.h"
+#include "hcs.hpp"
+#include <nlohmann/json.hpp>
 
 using namespace std::literals::chrono_literals;
 using namespace wsl::windows::common::registry;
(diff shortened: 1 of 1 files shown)


test/windows/WSLCTests.cpp
```diff
@@ -445,25 +444,60 @@ class WSLCTests
...
+    //   <id>
+    //       <Type>, <State>, <Id>, <Owner>
+    //
+    // where the indented line is a comma-separated list (with arbitrary
+    // whitespace padding) common to every supported Windows version.
     static std::vector<VmInfo> ListVms()
     {
-        wsl::windows::common::SubProcess process(nullptr, L"hcsdiag list -raw");
-        auto output = process.RunAndCaptureOutput(10000);
+        wsl::windows::common::SubProcess process(nullptr, L"hcsdiag list");
```
OneBlue: This works, although given that the parsing gets a bit nasty with the commas, I would have a preference for calling [HcsEnumerateComputeSystems](https://learn.microsoft.com/en-us/virtualization/api/hcs/reference/hcsenumeratecomputesystems) directly, since this is what hcsdiag.exe does anyway

OneBlue (dismissed):
This works, although

OneBlue (approved):

PR #40414, opened 2026-05-04
Replace FAIL_FAST() with error returns in public WslConfigSetting API

benhillis:
Replace `FAIL_FAST()` with graceful error returns in two public DLL config exports.

## Problem

`GetWslConfigSetting()` and `SetWslConfigSetting()` in `src/windows/libwsl/WslCoreConfigInterface.cpp` are public `STDAPI_` exports consumed by external clients (declared in `src/windows/inc/WslCoreConfigInterface.h`). Both used `FAIL_FAST()` for unrecognized `WslConfigEntry` values.

When a newer client passes a `WslConfigEntry` value that isn't present in an older DLL build, `FAIL_FAST()` crashes the calling process. That is inappropriate for a public ABI boundary that should tolerate version skew.

## Fix

- `GetWslConfigSetting`: return `{.ConfigEntry = WslConfigEntry::NoEntry}`. `NoEntry = 0` is the existing sentinel (`WslCoreConfigInterface.h:23`); the pre-existing line-72 case already returns this same shape for the zero-init request scenario.
- `SetWslConfigSetting`: return `ERROR_INVALID_PARAMETER` (`unsigned long` Win32 error - matches the function's return type and the pattern already used elsewhere in this function).
- Also adds the missing trailing newline at end of file.

## Compatibility

Existing callers passing values from the same-version header never hit the default branch, so no behavior change for them. Cross-version callers now get a sentinel/error code instead of a process crash.

### DIFF
--- a/src/windows/libwsl/WslCoreConfigInterface.cpp
+++ b/src/windows/libwsl/WslCoreConfigInterface.cpp
@@ -196,7 +196,7 @@ WslConfigSetting GetWslConfigSetting(WslConfig_t wslConfig, WslConfigEntry wslCo
         wslConfigSetting.StringValue = wslConfig->Config.KernelModulesPath.c_str();
         break;
     default:
-        FAIL_FAST();
+        return {.ConfigEntry = WslConfigEntry::NoEntry};
     }
 
     return wslConfigSetting;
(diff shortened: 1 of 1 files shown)


OneBlue (commented):
A FAIL_FAST() here makes sense to me since that would be a programmer error, and if it ever happened, it would make it easy to catch

benhillis:
fair enough, the API isn't really public.

PR #40422, opened 2026-05-05
Updated ADMX for WSLC

craigloewen-msft:

Updated ADMX values for WSLC controls.

## PR Checklist

Completed PR checklist, since it's just an ADMX no big changes or effort needed from the checklist.

- [X] **Closes:** Link to issue #xxx
- [X] **Dev docs:** Added/updated if needed
- [X] **Documentation updated:** If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/wsl/) and link it here: #xxx

Simple ADMX change.

## Validation Steps Performed

Validated ADMX compiles (It's known that the logic isn't implemented yet)


intune/WSL.admx
```diff
@@ -171,5 +180,60 @@
       </elements>
     </policy>
 
+    <!-- WSL Container CLI Policies -->
+
+    <policy name="AllowWSLContainerCLI" class="Machine" displayName="$(string.AllowWSLContainerCLI)" explainText="$(string.AllowWSLContainerCLIExplain)" key="Software\Policies\WSL" valueName="AllowWSLContainerCLI">
+      <parentCategory ref="WSLContainerCLI" />
+      <supportedOn ref="windows:SUPPORTED_Windows10" />
```
ptrivedi, line 181: Is this is min version supported? So win 11 is covered?

intune/en-US/WSL.adml
```diff
@@ -49,11 +49,46 @@
         <string id="NetworkingModeNAT">NAT</string>
         <string id="NetworkingModeMirrored">Mirrored</string>
         <string id="NetworkingModeVirtioProxy">VirtioProxy</string>
+
+        <string id="WSLContainers">WSL Containers</string>
+        <string id="WSLContainerCLI">WSL Container CLI</string>
+        <string id="WSLContainerAPI">WSL Container API</string>
+
+        <string id="AllowWSLContainerCLI">Allow WSL Container CLI</string>
+        <string id="AllowWSLContainerCLIExplain">This policy controls whether the WSL Container CLI (wslc.exe) can be used on this machine. When enabled or not configured, users can use the WSL Container CLI. When set to disabled, the WSL Container CLI is blocked for all users.</string>
```
benhillis: do we also want to call out container.exe?

intune/en-US/WSL.adml
```diff
@@ -49,11 +49,46 @@
...
+
+        <string id="WSLContainerCLIRegistryAllowlist">Allowlist for WSL container CLI registries</string>
+        <string id="WSLContainerCLIRegistryAllowlistExplain">When enabled, the WSL Container CLI will only be allowed to pull images from the registries listed here. Specify registries as a semicolon-separated list of hostnames (e.g. "registry1.example.com;registry2.example.com").
+
+If the allowlist is non-empty, only images from listed registries can be pulled. If an image's registry appears in both the allowlist and the blocklist, it will be blocked.</string>
```
benhillis: maybe "populated" versus non-empty

intune/en-US/WSL.adml
```diff
@@ -49,11 +49,46 @@
...
+
+        <string id="WSLContainerCLIRegistryBlocklist">Blocklist for WSL container CLI registries</string>
```
benhillis: mayve deny instead of block to match the allow verb?

intune/en-US/WSL.adml
```diff
@@ -49,11 +49,46 @@
...
+        <string id="WSLContainerCLIRegistryBlocklistExplain">When enabled, the WSL Container CLI will be unable to pull images from the registries listed here. Specify registries as a semicolon-separated list of hostnames (e.g. "registry1.example.com;registry2.example.com").
+
+If an image's registry appears in both the allowlist and the blocklist, it will be blocked.</string>
+
+        <string id="AllowWindowsAppsRunLinuxContainers">Allow Windows apps to run Linux containers</string>
```
benhillis: Is this for the SDK? Might want to be more explicit

"Allow Windows apps to create and manage Linux containers via the WSL Containers API" or something.

intune/en-US/WSL.adml
```diff
@@ -49,11 +49,46 @@
...
+        <string id="AllowWindowsAppsRunLinuxContainersExplain">This policy controls whether Windows applications can run Linux containers as part of their runtime. When enabled or not configured, Windows apps can run Linux containers. When set to disabled, Windows apps cannot run Linux containers.
+
+Warning: Setting this to 'Disabled' can break Windows apps that depend on Linux containers.</string>
+
+        <string id="AllowWSLDistributionAccess">Allow WSL distribution access</string>
```
benhillis: This looks unrelated to the WSLC changes.

benhillis: We already have "AllowWSL"

benhillis (changes requested):
Some suggestions / questions.

intune/WSL.admx
```diff
@@ -171,5 +180,60 @@
...
+      <enabledValue>
+        <decimal value="1" />
+      </enabledValue>
+      <disabledValue>
+        <decimal value="0" />
+      </disabledValue>
+    </policy>
+
```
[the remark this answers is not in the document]
benhillis: Intentional, code changes will come later.

intune/WSL.admx
florelis, line 181: I think this was supposed to be the product (WSL) version that supports the policy, not the min OS version supported by WSL. But most policies seem to state an OS version anyways..

intune/WSL.admx
```diff
@@ -171,5 +180,60 @@
...
+    <policy name="WSLContainerCLIRegistryAllowlist" class="Machine" displayName="$(string.WSLContainerCLIRegistryAllowlist)" explainText="$(string.WSLContainerCLIRegistryAllowlistExplain)" presentation="$(presentation.WSLContainerCLIRegistryAllowlist)" key="Software\Policies\WSL">
+      <parentCategory ref="WSLContainerCLI" />
+      <supportedOn ref="windows:SUPPORTED_Windows10" />
+      <elements>
+        <text id="WSLContainerCLIRegistryAllowlist_Input" valueName="WSLContainerCLIRegistryAllowlist" />
```
florelis: Why not `<list>`?

intune/en-US/WSL.adml
```diff
@@ -49,11 +49,46 @@
...
+
+        <string id="WSLContainers">WSL Containers</string>
+        <string id="WSLContainerCLI">WSL Container CLI</string>
+        <string id="WSLContainerAPI">WSL Container API</string>
+
+        <string id="AllowWSLContainerCLI">Allow WSL Container CLI</string>
+        <string id="AllowWSLContainerCLIExplain">This policy controls whether the WSL Container CLI (wslc.exe) can be used on this machine. When enabled or not configured, users can use the WSL Container CLI. When set to disabled, the WSL Container CLI is blocked for all users.</string>
+
+        <string id="WSLContainerCLIRegistryAllowlist">Allowlist for WSL container CLI registries</string>
+        <string id="WSLContainerCLIRegistryAllowlistExplain">When enabled, the WSL Container CLI will only be allowed to pull images from the registries listed here. Specify registries as a semicolon-separated list of hostnames (e.g. "registry1.example.com;registry2.example.com").
```
florelis: What about when the policy is disabled or not configured?

intune/en-US/WSL.adml
```diff
@@ -49,11 +49,46 @@
...
+
+If the allowlist is non-empty, only images from listed registries can be pulled. If an image's registry appears in both the allowlist and the blocklist, it will be blocked.</string>
+
+        <string id="WSLContainerCLIRegistryBlocklist">Blocklist for WSL container CLI registries</string>
+        <string id="WSLContainerCLIRegistryBlocklistExplain">When enabled, the WSL Container CLI will be unable to pull images from the registries listed here. Specify registries as a semicolon-separated list of hostnames (e.g. "registry1.example.com;registry2.example.com").
+
+If an image's registry appears in both the allowlist and the blocklist, it will be blocked.</string>
+
+        <string id="AllowWindowsAppsRunLinuxContainers">Allow Windows apps to run Linux containers</string>
+        <string id="AllowWindowsAppsRunLinuxContainersExplain">This policy controls whether Windows applications can run Linux containers as part of their runtime. When enabled or not configured, Windows apps can run Linux containers. When set to disabled, Windows apps cannot run Linux containers.
```
florelis: This should be exposed through the API so that an app can fail gracefully if it is blocked.

The API currently only exposes `WslcGetMissingComponents()` to determine if you can it, but the policy does not fit there. Maybe we need to bring back `CanRun()`?

intune/en-US/WSL.adml
```diff
@@ -49,11 +49,46 @@
         <string id="NetworkingModeNAT">NAT</string>
         <string id="NetworkingModeMirrored">Mirrored</string>
         <string id="NetworkingModeVirtioProxy">VirtioProxy</string>
+
+        <string id="WSLContainers">WSL Containers</string>
+        <string id="WSLContainerCLI">WSL Container CLI</string>
```
OneBlue: I'm not sure that it's a good idea to separate the CLI & API here. Unfortunately won't have a secured way to differentiate between those, so I would recommend just having one flag for all of wslc usage

intune/en-US/WSL.adml
OneBlue: Do we need both an allow list and a block list for those ? The UX becomes tricky if both are set.

If possible I would recommend scoping down to only allow-list at first

intune/WSL.admx
craigloewen-msft: Yes! :)

craigloewen-msft: It's a ADMX nuance. This is just saying this works on Win10.

intune/WSL.admx
craigloewen-msft: No idea if ADMX supports that or not. But a string seems easier for us to parse! So I'd opt for keeping it as a string

intune/en-US/WSL.adml
craigloewen-msft: No since `container.exe` is just an alias, our product name and branding is still `wslc.exe`

intune/en-US/WSL.adml
craigloewen-msft: Made the verbage clearer

intune/WSL.admx
florelis: ADMX supports having a list element. I don't know how this looks in Intune, but for the local Group Policy Editor, the experience for this with a list is IMO better than with a string. Especially once you have many or long values.

As an example, these are the first two policies that I found using string and list.
image
image

intune/en-US/WSL.adml
```diff
@@ -50,30 +50,17 @@
         <string id="NetworkingModeMirrored">Mirrored</string>
         <string id="NetworkingModeVirtioProxy">VirtioProxy</string>
 
-        <string id="WSLContainers">WSL Containers</string>
-        <string id="WSLContainerCLI">WSL Container CLI</string>
-        <string id="WSLContainerAPI">WSL Container API</string>
+        <string id="WSLContainer">WSL Container</string>
```
florelis, line 53: Is it Containers or Container?

intune/en-US/WSL.adml
florelis: I don't think it's clear what happens when you disable the policy. Does it mean "there is no allowlist, so you _cannot_ use any registry" or "there is no allowlist, so you _can_ use any registry"

intune/WSL.admx
benhillis: list seems better

intune/en-US/WSL.adml
craigloewen-msft: It's WSL container

intune/WSL.admx
craigloewen-msft: Updated to a list!

intune/en-US/WSL.adml
craigloewen-msft: I've made it shorter and made it clearer that 'when it's enabled X happens'. 
I think it gives it the right level of clarity

build/.cmake/api/v1/query/client-vscode/query.json
```diff
@@ -0,0 +1 @@
+{"requests":[{"kind":"cache","version":2},{"kind":"codemodel","version":2},{"kind":"toolchains","version":1},{"kind":"cmakeFiles","version":1}]}
```
OneBlue: This looks unrelated

benhillis: ???

benhillis (approved):

PR #40432, opened 2026-05-06
Generate custom container name on empty input

yao-msft:
When containerOptions.Name is empty, generate our own container name.

## PR Checklist

## Validation Steps Performed
Added tests.


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -109,6 +110,29 @@ wslc_schema::InspectImage ConvertInspectImage(const docker_schema::InspectImage&
...
 }
 
+using wsl::windows::service::wslc::c_descriptors;
+using wsl::windows::service::wslc::c_mountains;
+
+// Generate a random container name in the format "descriptor_mountain".
+// When retry > 0, appends a random digit (0-9) to reduce collisions.
+std::string GenerateContainerName(int retry)
+{
+    static std::mt19937 s_gen(std::random_device{}());
```
OneBlue: I can't find a definite mention that this class is safe to use from multiple threads without synchronization.


Given that, I would recommend not marking it static unless we find a guarantee that it's thread safe

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -109,6 +110,29 @@ wslc_schema::InspectImage ConvertInspectImage(const docker_schema::InspectImage&
...
+
+    std::uniform_int_distribution<size_t> leftDist(0, c_descriptors.size() - 1);
+    std::uniform_int_distribution<size_t> rightDist(0, c_mountains.size() - 1);
+
+    auto name = std::string(c_descriptors[leftDist(s_gen)]) + "_" + c_mountains[rightDist(s_gen)];
```
OneBlue: nit: This could be simplified to: 

```
auto name = std::format("{}_{}", c_descriptors[leftDist(s_gen)], c_mountains[rightDist(s_gen)]);
```

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1636,8 +1660,41 @@ try
...
 
+        // Generate a unique container name if the user didn't provide one.
+        std::string generatedName;
+        auto options = *containerOptions;
+        if (options.Name == nullptr || options.Name[0] == '\0')
+        {
+            std::unordered_set<std::string> existingNames;
+            for (const auto& c : m_containers)
+            {
+                existingNames.insert(c->Name());
```
OneBlue: Looping through all the containers will make us create a copy of each container name. Instead of doing this, I think just iterating through `m_containers` after generating the name would be OK, since we expect conflicts to be very uniquely

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1636,8 +1660,41 @@ try
...
+                {
+                    generatedName = randomName;
+                    break;
+                }
+            }
+
+            // Fallback to a GUID name
+            if (generatedName.empty())
+            {
+                GUID guid{};
```
OneBlue, line 1686: Probably a good idea to log an event if we ever hit this

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1636,8 +1660,41 @@ try
     {
         std::scoped_lock lock(m_containersLock, m_volumesLock, m_networksLock);
 
+        // Generate a unique container name if the user didn't provide one.
+        std::string generatedName;
+        auto options = *containerOptions;
```
OneBlue: Copying the entire structure feels a bit rough here. Going through it, I *think* that the default copy behavior for this struct will do the right thing, but I would have a preference for adding an additional name argument to `Create()` so we don't have to do this

test/windows/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ add_compile_definitions(INLINE_TEST_METHOD_MARKUP)
 
 add_library(wsltests SHARED ${SOURCES} ${HEADERS})
 
-target_include_directories(wsltests PRIVATE ${CMAKE_SOURCE_DIR}/src/windows/WslcSDK)
+target_include_directories(wsltests PRIVATE ${CMAKE_SOURCE_DIR}/src/windows/WslcSDK ${CMAKE_SOURCE_DIR}/src/windows/wslcsession)
```
OneBlue: Instead of adding an include path, I would recommend moving that header to `src/windows/common` so we don't need to do a special case for it

OneBlue (approved):
LGTM

PR #40433, opened 2026-05-06
Avoid distro zombie state when wsl init dies in systemd mode

chemwolf6922:

This PR adds a wsl-init-watcher process to monitor the wsl init when systemd mode is enabled. 
The mini_init process monitors the process 1 of a distro to determine if the distro is alive. In non-systemd mode, that is the wsl init. In systemd mode, that will be the systemd init. And if the wsl init dies in systemd mode, the distro ends up in a zombie state. Where all wsl calls from Windows fail with "Catastrophic failure". 
This PR works by adding a new process that monitors the wsl init in systemd mode. And requests a shutdown if the wsl init process died.

## PR Checklist

## Validation Steps Performed

Add test: UnitTests::UnitTests::SystemdKillInitTerminatesDistro


test/windows/UnitTests.cpp
```diff
@@ -393,6 +393,21 @@ class UnitTests
...
+        auto revert = EnableSystemd();
+        VERIFY_IS_TRUE(IsSystemdRunning(L"--system"));
+
+        // Kill the WSL init process
+        LxsstuLaunchWsl(L"kill -9 2");
+
+        Sleep(3000);
+
+        // Verify that a new WSL command succeeds (the distro restarts cleanly).
+        auto [out, err] = LxsstuLaunchWslAndCaptureOutput(L"echo hello");
```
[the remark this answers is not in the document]
chemwolf6922: Not a good idea.

test/windows/UnitTests.cpp
```diff
@@ -393,6 +393,21 @@ class UnitTests
...
     }
 
+    WSL2_TEST_METHOD(SystemdKillInitTerminatesDistro)
+    {
+        auto revert = EnableSystemd();
+        VERIFY_IS_TRUE(IsSystemdRunning(L"--system"));
+
+        // Kill the WSL init process
+        LxsstuLaunchWsl(L"kill -9 2");
+
```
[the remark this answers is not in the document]
chemwolf6922: The wsl init having pid 2 in systemd mode is a stable behavior. This suggested check is not correct nor necessary.

src/linux/init/init.cpp
```diff
@@ -3488,4 +3512,44 @@ void WaitForBootProcess(wsl::linux::WslDistributionConfig& Config)
             LOG_ERROR("{} failed to start within {}ms", INIT_PATH, Config.BootInitTimeout);
         }
     }
-}
\ No newline at end of file
+}
+
+int WslInitWatcher(int Argc, char** Argv)
```
OneBlue, line 3509: Do we need a separate entrypoint for this, or could we simplify fork and run that logic inside the forked process ?

test/windows/UnitTests.cpp
```diff
@@ -393,6 +393,21 @@ class UnitTests
...
+        Sleep(3000);
```
OneBlue: Hard coded sleeps will usually make our tests slow / unreliable. Instead of doing this, I recommend using `wsl::shared::retry::RetryWithTimeout` until we see the distro being terminated

test/windows/UnitTests.cpp
```diff
@@ -393,6 +393,21 @@ class UnitTests
         }
     }
 
+    WSL2_TEST_METHOD(SystemdKillInitTerminatesDistro)
+    {
+        auto revert = EnableSystemd();
+        VERIFY_IS_TRUE(IsSystemdRunning(L"--system"));
+
+        // Kill the WSL init process
+        LxsstuLaunchWsl(L"kill -9 2");
```
OneBlue: We should validate that the exit code is 0 here

src/linux/init/init.cpp
```diff
@@ -2396,6 +2402,24 @@ Return Value:
             _exit(1);
         }
 
+        //
+        // Fork a watcher process that monitors WSL init and tears down
```
OneBlue, line 2406: Given that we're creating a child of init, wouldn't killing init also kill the watcher process ?

OneBlue (changes requested):

src/linux/init/init.cpp
chemwolf6922: This mainly is for getting a clean state with exec and avoid inheriting states from the init process. Running the logic directly after fork requires carefully managing the inherited stuff. Which may cause problems in the future.

src/linux/init/init.cpp
chemwolf6922: image
This child process do not have `prctl(PR_SET_PDEATHSIG, xxx)` set on purpose, so it won't die if the parent dies. (Above is a test replacing the reboot with sleep)

src/linux/init/init.cpp
```diff
@@ -2396,6 +2402,24 @@ Return Value:
...
+        const auto wslInitPid = std::to_string(getpid());
+        const int WatcherPid = fork();
+        if (WatcherPid < 0)
+        {
+            FATAL_ERROR("fork failed {}", errno);
+        }
+        else if (WatcherPid == 0)
+        {
+            execl(LX_INIT_PATH, LX_INIT_WSL_INIT_WATCHER, wslInitPid.c_str(), static_cast<char*>(nullptr));
+            LOG_ERROR("execl({}) failed {}", LX_INIT_WSL_INIT_WATCHER, errno);
```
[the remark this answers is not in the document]
chemwolf6922: No, this is redundant.

src/linux/init/init.cpp
```diff
@@ -3488,4 +3512,44 @@ void WaitForBootProcess(wsl::linux::WslDistributionConfig& Config)
...
+    {
+    }
+    if (rc <= 0 || (pfd.revents & POLLIN) == 0)
+    {
+        _exit(1);
+    }
+
+    // Teardown the current PID namespace. Not shutting down the VM.
+    reboot(RB_POWER_OFF);
+    _exit(1);
```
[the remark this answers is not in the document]
chemwolf6922: Intentionally keep this simple.

chemwolf6922:
Hi @OneBlue , could you please take another look? Thanks.

src/linux/init/init.cpp
```diff
@@ -3488,4 +3512,44 @@ void WaitForBootProcess(wsl::linux::WslDistributionConfig& Config)
...
     }
-}
\ No newline at end of file
+}
+
+int WslInitWatcher(int Argc, char** Argv)
+{
+    // Keep this as simple as possible. Log is not used.
+
+    prctl(PR_SET_NAME, LX_INIT_WSL_INIT_WATCHER, 0, 0, 0);
```
OneBlue: nit: We could use `UtilSetThreadName` for this

src/linux/init/init.cpp
```diff
@@ -3488,4 +3512,44 @@ void WaitForBootProcess(wsl::linux::WslDistributionConfig& Config)
...
+    int rc;
+    while ((rc = poll(&pfd, 1, -1)) < 0 && errno == EINTR)
+    {
+    }
+    if (rc <= 0 || (pfd.revents & POLLIN) == 0)
+    {
+        _exit(1);
+    }
+
+    // Teardown the current PID namespace. Not shutting down the VM.
```
OneBlue, line 3537: nit: If we ever hit this, it would be a good idea to log something to dmesg

src/linux/init/init.cpp
```diff
@@ -2396,6 +2402,24 @@ Return Value:
             _exit(1);
         }
 
+        //
+        // Fork a watcher process that monitors WSL init and tears down
+        // the PID namespace if it exits unexpectedly.
+        //
+
+        const auto wslInitPid = std::to_string(getpid());
+        const int WatcherPid = fork();
```
OneBlue: nit: We could use `UtilCreateChildProcess()` to simplify this is a bit

src/linux/init/init.cpp
```diff
@@ -3488,4 +3512,44 @@ void WaitForBootProcess(wsl::linux::WslDistributionConfig& Config)
...
+
+    prctl(PR_SET_NAME, LX_INIT_WSL_INIT_WATCHER, 0, 0, 0);
+
+    if (Argc < 2)
+    {
+        _exit(1);
+    }
+
+    char* end = nullptr;
+    const long parsed = strtol(Argv[1], &end, 10);
```
OneBlue: Instead of having to pass the pid through the command line, could we just use `getppid()` here ?

test/windows/UnitTests.cpp
```diff
@@ -393,6 +393,34 @@ class UnitTests
         }
     }
 
+    WSL2_TEST_METHOD(SystemdKillInitTerminatesDistro)
+    {
+        WslShutdown();
```
OneBlue: `WslShutdown()` isn't needed here (WslConfigChange does automatically when the config is changed)

test/windows/UnitTests.cpp
```diff
@@ -393,6 +393,34 @@ class UnitTests
...
+        VERIFY_NO_THROW(wsl::shared::retry::RetryWithTimeout<void>(
+            [&]() { THROW_HR_IF(E_UNEXPECTED, !IsSystemdRunning(L"--system")); }, std::chrono::seconds(1), std::chrono::minutes(1)));
+
+        // Kill the WSL init process
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(L"kill -9 2"), 0L);
+
+        // Wait for the distro to exit.
+        VERIFY_NO_THROW(wsl::shared::retry::RetryWithTimeout<void>(
+            [&]() {
+                const auto commandLine = LxssGenerateWslCommandLine(L"--list --running");
```
OneBlue: nit: We could extract [this method](https://github.com/microsoft/WSL/blob/14a635f6d39777196aa4c2482d8e715a605b23d0/test/windows/UnitTests.cpp#L6160) to check if the distro is running (this will have the benefit of not returning false positives if there happens to be a distro with a name that contains the name of the test distro)

OneBlue (dismissed):
Thank you for answering my questions! 

Added a couple minor comments, but change LGTM overall

src/linux/init/init.cpp
chemwolf6922: UtilSetThreadName and UtilCreateChildProcess both call the log system in the child process. But if we want to add log anyways, it's better to use those.

src/linux/init/init.cpp
```diff
@@ -3488,4 +3504,37 @@ void WaitForBootProcess(wsl::linux::WslDistributionConfig& Config)
...
+
+    UtilSetThreadName(LX_INIT_WSL_INIT_WATCHER);
+
+    const pid_t wslInitPid = getppid();
+    const int pidfd = syscall(SYS_pidfd_open, wslInitPid, 0);
+    if (pidfd < 0)
+    {
+        LOG_ERROR("pidfd_open failed {}", errno);
+        _exit(1);
+    }
```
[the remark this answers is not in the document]
chemwolf6922: A watcher failure should not cause a system failure.

chemwolf6922:
Hi @OneBlue , The PR is updated. Could you please take a look? Thanks.

OneBlue (approved):
LGTM, thank you !

OneBlue:
@chemwolf6922: I think there might be an issue with that change. I hit a hang today while testing something else, and init gets into a defunct state, suggesting that the init watcher doesn't actually reap it: 

image

PR #40434, opened 2026-05-06
Add support for starting a container in multiple networks

beena352:
This adds support for attaching a container to more than one network when it is created. Before this change, a container could only be connected to a single network. Now additional network names can be passed alongside the primary network, and the container will be connected to all of them at create time.

## PR Checklist

-Replaced the flat AdditionalNetworks/ContainerNetworkName fields with a unified WSLCNetworkAttachment struct array on WSLCContainerNetwork. The struct has NetworkName and ContainerIpAddress (reserved for a future --ip followup). For Custom network mode, Networks[0] is the primary network; for Bridged, all entries are additional.
- Added AddAdditionalNetwork() method to WSLCContainerLauncher for easy use
- On container create, all networks (primary + additional) are sent to Docker in one request using NetworkingConfig.EndpointsConfig
- Container inspect now returns network details (IP, gateway, MAC address) for each attached network under NetworkSettings.Networks


src/windows/service/inc/wslc.idl
```diff
@@ -293,6 +293,9 @@ typedef struct _WSLCContainerOptions
 
     [unique, size_is(NamedVolumesCount)] WSLCNamedVolume* NamedVolumes;
     ULONG NamedVolumesCount;
+
+    [unique, size_is(AdditionalNetworksCount)] LPCSTR* AdditionalNetworks;
+    ULONG AdditionalNetworksCount;
```
OneBlue: Instead of adding an additional field, I recommend making updating `WSLCContainerNetwork::ContainerNetworkName` to be an array instead. I think we should probably have a structure to define a custom network attachment, something like: 

```
struct WSLCNetworkAttachment
{
    string NetworkName;
    optional<string> ContainerIpAddress;
};
```

(that will allow us to implement --ip later in a followup change)

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -405,6 +405,53 @@ void ValidateNamedVolumes(
...
+
+    THROW_HR_WITH_USER_ERROR_IF(
+        E_INVALIDARG,
+        Localization::MessageWslcAdditionalNetworksRequirePrimary(),
+        primaryNetworkMode == "host" || primaryNetworkMode == "none");
+
+    std::unordered_set<std::string> seenNetworks;
+
+    // Track the primary network name to detect duplicates in the additional networks list.
+    seenNetworks.insert(primaryNetworkMode);
```
OneBlue: nit: Instead of maintaining a list of the networks we saw, we can check use `insert()` when we add to `endpointConfig` and use its return value to check if a new entry was actually added or not

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -381,6 +381,71 @@ void ProcessNamedVolumes(const WSLCContainerOptions& containerOptions, wsl::wind
...
+        Localization::MessageWslcAdditionalNetworksRequirePrimary(),
+        primaryNetworkMode == "host" || primaryNetworkMode == "none");
+
+    // Add the primary and all additional networks to EndpointsConfig so Docker attaches them all at create time.
+    auto& endpointsConfig = request.NetworkingConfig.EndpointsConfig;
+    endpointsConfig[primaryNetworkMode] = {};
+
+    for (ULONG i = startIndex; i < network.NetworksCount; i++)
+    {
+        // NetworkName may arrive as null when the caller passed "" (MIDL marshals empty LPCSTR as null).
```
OneBlue: I'm surprised by this. Do we know what could cause an empty string to be turned into null through MIDL ? If that was the cause, that would cause us a lot of issues in other places

beena352: Good catch, not a MIDL issue. [unique] LPCSTR allows null, so that's where it comes from. Updated to explicitly reject null/empty with a clearer error.

OneBlue (approved):

PR #40441, opened 2026-05-06
Use overlapped pipes in WSLC session terminate tests

benhillis:
The `LoadImage` and `ImportImage` 'session terminate' sub-tests used `CreatePipe` which creates synchronous pipe handles. When the relay calls `ReadFile` with an `OVERLAPPED` structure on a synchronous handle, the call blocks the thread instead of returning `ERROR_IO_PENDING`. This prevents `WaitForMultipleObjects` from ever checking the session terminating event, causing a deadlock when `Terminate()` is called.

Replace `CreatePipe` with `OpenAnonymousPipe` (`FILE_FLAG_OVERLAPPED`) so `ReadFile` returns `ERROR_IO_PENDING` and the relay's event loop can detect the termination signal.

### DIFF
--- a/test/windows/WSLCTests.cpp
+++ b/test/windows/WSLCTests.cpp
@@ -1113,10 +1113,11 @@ class WSLCTests
         }
 
         // Validate that LoadImage is aborted when the session terminates.
+        // N.B. The read pipe must support overlapped IO so the relay's event-based cancellation works.
+        // CreatePipe creates synchronous pipes where ReadFile blocks the thread, preventing
+        // WaitForMultipleObjects from detecting the session terminating event.
         {
-            wil::unique_handle pipeRead;
-            wil::unique_handle pipeWrite;
-            VERIFY_WIN32_BOOL_SUCCEEDED(CreatePipe(&pipeRead, &pipeWrite, nullptr, 2));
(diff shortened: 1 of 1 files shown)


OneBlue (commented):
> This prevents WaitForMultipleObjects from ever checking the session terminating event, causing a deadlock when Terminate() is called

I don't think this is true. We specifically have code to cancel synchronous IO when the session terminates. Did we see a deadlock caused by this test ?

benhillis:
> > This prevents WaitForMultipleObjects from ever checking the session terminating event, causing a deadlock when Terminate() is called
> 
> I don't think this is true. We specifically have code to cancel synchronous IO when the session terminates. Did we see a deadlock caused by this test ?

This assertion is from copilot, I don't buy it either. However, I think switching away from the CreatePipe API is probably a good call.

OneBlue:
> This assertion is from copilot, I don't buy it either. However, I think switching away from the CreatePipe API is probably a good call.

Ideally I think we should test with both (the BlockingOperations tests use overlapped pipes), so we can validate that we behave correctly in both cases

yao-msft (approved):

OneBlue:
Closing since the service must support both overlapped and non-overlapped HANDLES (so the tests are OK to use either).

The hang on fe_releases testpasses repro'd on that specific test case, but the root cause was unrelated to these handles

PR #40443, opened 2026-05-06
Remove DnsTunnelingSocket flag from VirtioNetworking

benhillis:
Drops the `DnsTunnelingSocket` virtio networking flag (and the dedicated DNS hvsocket plumbing it carried) in favor of the in-built `DnsTunneling` path. In virtio proxy mode, DNS queries are now always forwarded by the host virtio proxy itself: Linux's `/etc/resolv.conf` points at the eth0 gateway IP rather than the listener IP, no DNS hvsocket is opened, and the Linux init does not start a `DnsTunnelingManager`.

Changes:
- `VirtioNetworking`: remove `DnsTunnelingSocket` enum value, `dnsHvsocket` constructor parameter, `m_dnsTunnelingResolver` field, and the `DnsResolver.h` include. `RefreshGuestConnection` now uses only the in-built `DnsTunneling` path.
- `WslCoreVm`: virtio proxy branch unconditionally sets the `DnsTunneling` flag when DNS tunneling is enabled. `message->EnableDnsTunneling` is suppressed for virtio proxy mode so Linux init won't open the DNS hvsocket. The pre-accepted `dnsTunnelingSocket` is dropped on the NAT->VirtioProxy fallback path.
- `HcsVirtualMachine`: WSLC virtio proxy branch likewise switches to the `DnsTunneling` flag and discards the unused `dnsSocketHandle`.
- Tests: add `VirtioProxyTests::DnsTunnelingResolvConfUsesGateway` which asserts `resolv.conf` contains the gateway IP and not the legacy listener IP, confirming the in-built path is in use.

### DIFF
--- a/src/windows/service/exe/HcsVirtualMachine.cpp
+++ b/src/windows/service/exe/HcsVirtualMachine.cpp
@@ -372,24 +372,26 @@ try
     // so we need our own copies to take ownership.
     wil::unique_socket gnsSocketHandle{reinterpret_cast<SOCKET>(wslutil::DuplicateHandle(GnsSocket))};
     wil::unique_socket dnsSocketHandle;
+
+    // Run the DnsResolver support check first so it can clear the feature flag before the validation below.
+    // The check still applies to virtio proxy because the host virtio proxy uses the same Windows DNS APIs.
     if (FeatureEnabled(WslcFeatureFlagsDnsTunneling))
     {
-        THROW_HR_IF(E_INVALIDARG, DnsSocket == nullptr);
-
         const auto result = wsl::core::networking::DnsResolver::LoadDnsResolverMethods();
         if (FAILED(result))
         {
             LOG_HR_MSG(result, "Failed to load DNS resolver methods, DNS tunneling will be disabled");
             WI_ClearFlag(m_featureFlags, WslcFeatureFlagsDnsTunneling);
         }
-        else
-        {
-            dnsSocketHandle.reset(reinterpret_cast<SOCKET>(wslutil::DuplicateHandle(*DnsSocket)));
-        }
     }
-    else
+
+    // The DNS hvsocket is only allocated for NAT mode.
+    const bool expectDnsSocket = FeatureEnabled(WslcFeatureFlagsDnsTunneling) && m_networkingMode == WSLCNetworkingModeNAT;
+    THROW_HR_IF(E_INVALIDARG, expectDnsSocket != (DnsSocket != nullptr));
+
+    if (DnsSocket != nullptr)
     {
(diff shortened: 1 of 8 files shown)


src/windows/service/exe/HcsVirtualMachine.cpp
```diff
@@ -372,24 +372,24 @@ try
     // so we need our own copies to take ownership.
     wil::unique_socket gnsSocketHandle{reinterpret_cast<SOCKET>(wslutil::DuplicateHandle(GnsSocket))};
     wil::unique_socket dnsSocketHandle;
+
+    // The DNS hvsocket is only allocated for NAT mode.
+    THROW_HR_IF(E_INVALIDARG, (FeatureEnabled(WslcFeatureFlagsDnsTunneling) && m_networkingMode == WSLCNetworkingModeNAT) != (DnsSocket != nullptr));
```
ptrivedi, line 377: nit: definitely room for cleanup and if/else consolidation for FeatureEnabled(WslcFeatureFlagsDnsTunneling), NAT mode, DnsSocket != nullptr conditions below

test/windows/NetworkTests.cpp
```diff
@@ -5083,5 +5083,22 @@ class VirtioProxyTests
...
+    {
+        VIRTIOPROXY_TEST_ONLY();
+        DNS_TUNNELING_TEST_ONLY();
+
+        m_config->Update(LxssGenerateTestConfig({.networkingMode = wsl::core::NetworkingMode::VirtioProxy, .dnsTunneling = true}));
+
+        const auto state = NetworkTests::GetInterfaceState(L"eth0");
+        VERIFY_IS_TRUE(state.Gateway.has_value());
+
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(L"cat /etc/resolv.conf | grep nameserver | grep -F " + state.Gateway.value()), static_cast<DWORD>(0));
```
ptrivedi, line 5098: nit: doesn't matter with the generated resolv.conf, but I do like the tightened awk version in WSLCTests. 'grep -F' here would do substring matches so maybe grep -w is better. Also like the ^nameserver in the WSLCTests better than the 'grep nameserver' below that could match comments etc.

ptrivedi (approved):

src/windows/service/exe/HcsVirtualMachine.cpp
benhillis: I thought this was pretty clean, what were you thinking? I can address this in a follow-up.

test/windows/NetworkTests.cpp
benhillis: Fair, can adjust tests in a follow-up.

PR #40473, opened 2026-05-08
CLI: Add image import command

AmelBawa-msft:
- Added wslc image import command to create an image from a tarball, with an optional image name/tag argument. Also registered as a top-level wslc import alias.
- Supports file path or stdin (-) as input
- E2E tests covering import with/without tag, invalid path, stdin-style usage, and the root-level import alias.

## Related issues:
- http://task.ms/62246732 (CLI parser rejecting `-`)
- http://task.ms/62249460 (Allow untagged images)

## PR Checklist


src/windows/wslc/services/ImageService.cpp
```diff
@@ -187,6 +187,31 @@ void ImageService::Load(wsl::windows::wslc::models::Session& session, const std:
     THROW_IF_FAILED(session.Get()->LoadImage(ToCOMInputHandle(imageFile.get()), nullptr, fileSize.QuadPart));
 }
 
+void ImageService::Import(wsl::windows::wslc::models::Session& session, const std::wstring& input, const std::string& imageName)
+{
+    HANDLE imageHandle = nullptr;
```
OneBlue: nit: Could we factor this block with `Load()` ? Those two methods essentially do the same (open a file or stdin based on path, and then get the file size

src/windows/wslc/services/ImageService.cpp
```diff
@@ -187,6 +187,31 @@ void ImageService::Load(wsl::windows::wslc::models::Session& session, const std:
...
+    wil::unique_hfile imageFile;
+    ULONGLONG contentLength = 0;
+
+    if (input == L"-")
+    {
+        imageHandle = GetStdHandle(STD_INPUT_HANDLE);
```
OneBlue: If we take this branch, `contentLength` will be set to zero, which will most likely fail in the service. 

For now, I would recommend calling GetFileSizeEx() on stdin, and returning an error if it fails (for the case where stdin is just a file handle)

test/windows/wslc/e2e/WSLCE2EImageImportTests.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    WSLCE2EImageImportTests.cpp
+
+Abstract:
+
+    This file contains end-to-end tests for WSLC image import.
```
OneBlue, line 11: I'd recommend adding test coverage from importing an image from stdin

AmelBawa-msft: Found a bug in the CLI which makes this challenging to test: http://task.ms/62246732 
I added a skipped stub test method with a TODO in it.

src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -210,6 +210,22 @@ void LoadImage(CLIExecutionContext& context)
...
+    WI_ASSERT(context.Data.Contains(Data::Session));
+    WI_ASSERT(context.Args.Contains(ArgType::ImportFile));
+    auto& session = context.Data.Get<Data::Session>();
+
+    std::string imageName;
+    if (context.Args.Contains(ArgType::ImageId))
+    {
+        imageName = WideToMultiByte(context.Args.Get<ArgType::ImageId>());
+    }
+
```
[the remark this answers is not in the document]
AmelBawa-msft: http://task.ms/62249460

src/windows/wslc/services/ImageService.cpp
```diff
@@ -71,6 +71,33 @@ std::string GetServerFromImage(const std::string& image)
     return server;
 }
 
+struct InputSource
+{
+    HANDLE Handle = nullptr;
```
OneBlue: nit: This could be simplified a bit by using `wsl::windows::common::relay::HandleWrapper`. It can either hold a "owned" handle in the form of a `wil::unique_handle`, or  an "unowned" HANDLE.

src/windows/wslc/services/ImageService.cpp
```diff
@@ -71,6 +71,33 @@ std::string GetServerFromImage(const std::string& image)
...
+    }
+    else
+    {
+        result.File.reset(CreateFileW(input.c_str(), GENERIC_READ, FILE_SHARE_READ, nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr));
+        THROW_LAST_ERROR_IF(!result.File);
+        result.Handle = result.File.get();
+    }
+
+    LARGE_INTEGER fileSize{};
+    THROW_LAST_ERROR_IF(!GetFileSizeEx(result.Handle, &fileSize));
```
OneBlue: It might be a good idea to return a specialized error code if this fails (because in the current implementation, this will fail if stdin is a pipe), just to make it clear that we don't support pipes

OneBlue (dismissed):
LGTM, minor comments

OneBlue (approved):
LGTM ! We might want to add test coverage to cover trying to import from a pipe, but that can be done in a future change

PR #40475, opened 2026-05-08
Add ContainerStats and IWSLCContainer Stats runtime implementation

dkbennett:
Adds Stats to IWSLCContainer
Implements Stats
Tests Stats

## PR Checklist

See Copilot summary :)

## Validation Steps Performed
Ran the new ContainerStats tests, which passed.

### DIFF
--- a/src/windows/service/inc/wslc.idl
+++ b/src/windows/service/inc/wslc.idl
@@ -527,6 +527,7 @@ interface IWSLCContainer : IUnknown
     HRESULT GetName([out, string] LPSTR* Name);
     HRESULT GetLabels([out, size_is(, *Count)] WSLCLabelInformation** Labels, [out] ULONG* Count);
     HRESULT Kill([in] WSLCSignal Signal);
+    HRESULT Stats([out] LPSTR* Output);
 }
 
 typedef enum _WSLCDeletedImageType
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -1732,6 +1732,19 @@ void WSLCContainerImpl::Logs(WSLCLogsFlags Flags, WSLCHandle* Stdout, WSLCHandle
     }
 }
 
+void WSLCContainerImpl::Stats(LPSTR* Output) const
+{
+    auto lock = m_lock.lock_shared();
+
+    try
+    {
+        auto stats = m_dockerClient.ContainerStats(m_id);
+        std::string json = wsl::shared::ToJson(stats);
+        *Output = wil::make_unique_ansistring<wil::unique_cotaskmem_ansistring>(json.c_str()).release();
+    }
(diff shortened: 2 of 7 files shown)


src/windows/service/inc/wslc.idl
```diff
@@ -527,6 +527,7 @@ interface IWSLCContainer : IUnknown
     HRESULT GetName([out, string] LPSTR* Name);
     HRESULT GetLabels([out, size_is(, *Count)] WSLCLabelInformation** Labels, [out] ULONG* Count);
     HRESULT Kill([in] WSLCSignal Signal);
```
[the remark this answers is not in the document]
dkbennett: I checked with Pierre prior to doing this, and adding to the existing IWSLCContainer is fine for now since we dont expect any clients taking a dependency on this yet.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1732,6 +1732,19 @@ void WSLCContainerImpl::Logs(WSLCLogsFlags Flags, WSLCHandle* Stdout, WSLCHandle
     }
 }
 
+void WSLCContainerImpl::Stats(LPSTR* Output) const
+{
+    auto lock = m_lock.lock_shared();
+
+    try
+    {
+        auto stats = m_dockerClient.ContainerStats(m_id);
```
OneBlue, line 1741: What we usually do for similar situations (like Inspect()) is that we have our own schema in `wslc_schema.h`, but given that Stats() is very self contained, directly returning the docker schema is probably OK, since we don't use it for anything else. 

We can always split those later if we need to

test/windows/WSLCTests.cpp
```diff
@@ -7797,6 +7797,124 @@ class WSLCTests
         }
     }
 
+    WSLC_TEST_METHOD(ContainerStats_RunningContainer)
+    {
+        // Start a long-lived detached container on a bridged network so network stats are populated.
+        WSLCContainerLauncher launcher("debian:latest", "wslc-test-stats", {"sleep", "60"}, {}, WSLCContainerNetworkTypeBridged);
+
+        auto runningContainer = launcher.Launch(*m_defaultSession, WSLCContainerStartFlagsNone);
+        auto cleanup = wil::scope_exit([&]() { runningContainer.SetDeleteOnClose(true); });
```
OneBlue, line 7806: nit: this isn't actually needed (DeleteOnClose is the default behavior)

OneBlue (approved):
LGTM (minor test comment). Thank you for contributing to the runtime !

src/windows/wslcsession/WSLCContainer.cpp
dkbennett: I had considered this, but it would have just been a passthrough, but yes happy to split it later if we need to.

test/windows/WSLCTests.cpp
dkbennett: I'll remove this in the CLI command implementation PR to prevent having to do another churn cycle for this.

PR #40476, opened 2026-05-08
Add Uid/Gid/Fixed driver options to WSLC VHD volumes and expose via SDK

benhillis:
## Summary

Adds POSIX `Uid`/`Gid` and pre-allocated `Fixed` driver options to the WSLC named-volume `vhd` driver, and plumbs them through the public C SDK and the WinRT projection.

Today, containers running as a non-root user cannot write to their own persistent volumes — `mkfs.ext4` leaves the root inode owned by `root:root`, and the `vhd` driver only accepted `SizeBytes`. The new options let the SDK consumer create a volume that is immediately usable by the in-container user.

`Fixed` lets workloads that need predictable I/O pre-allocate the VHD instead of growing it dynamically.

## Driver options (named volumes)

| Option | Type | Notes |
|---|---|---|
| `SizeBytes` | uint64 (decimal) | required; >0 |
| `Fixed` | bool (`true`/`false`/`1`/`0`) | optional; defaults to `false` |
| `Uid` | uint32 (decimal) | optional; **must be paired with `Gid`** |
| `Gid` | uint32 (decimal) | optional; **must be paired with `Uid`** |

Ownership is baked into the ext4 root inode at format time via `mkfs.ext4 -E root_owner=UID:GID`, not via a post-mount `chown`. The container user owns the root inode from the first mount and can `chmod` it with its own umask if it needs different permissions.

All option parsing goes through a shared `OptionParser` helper (`Required<T>`/`Optional<T>`/`OptionalBool`/`RejectUnknown`) with errno-capture, end-pointer validation, leading-sign rejection, and consumed-key tracking. The `Open` path uses the same parser so persisted metadata is validated identically on reload.

## C SDK (`wslcsdk.h` / `wslcsdk.dll`)

```c
typedef enum WslcVhdRequirementsFlags {
    WSLC_VHD_REQ_FLAG_NONE  = 0,
    WSLC_VHD_REQ_FLAG_OWNER = 1,
} WslcVhdRequirementsFlags;

typedef struct WslcVhdRequirements {
    PCSTR    name;
    uint64_t sizeBytes;
    WslcVhdType type;
    WslcVhdRequirementsFlags flags;
    uint32_t uid;
    uint32_t gid;
} WslcVhdRequirements;
```

`WslcCreateSessionVhdVolume`:
* honors `WSLC_VHD_TYPE_FIXED` (was previously `E_NOTIMPL`)
* dynamically builds the underlying `WSLCDriverOption[]` based on the flag bits
* rejects unknown `type` values and unknown flag bits with `E_INVALIDARG` so future additions cannot silently fail-open

`WslcSetSessionSettingsVhd` does not plumb owner/fixed through the session rootfs VHD path and rejects `flags != NONE` with `E_INVALIDARG` instead of silently ignoring those fields.

> [!IMPORTANT]
> **ABI break:** `WSLC_SESSION_OPTIONS_SIZE` bumps `80 → 88` to match the wider embedded `WslcVhdRequirements`. Callers of `wslcsdk.lib`/`wslcsdk.dll` must recompile against the new header.

### Caller usage

```c
WslcVhdRequirements vhd = {0};
vhd.name      = "data";
vhd.sizeBytes = 64ULL << 20;
vhd.type      = WSLC_VHD_TYPE_FIXED;
vhd.flags     = WSLC_VHD_REQ_FLAG_OWNER;
vhd.uid       = 65534;
vhd.gid       = 65534;
WslcCreateSessionVhdVolume(session, &vhd, &errorMsg);
```

## WinRT projection (`Microsoft.WSL.Containers`)

```idl
runtimeclass VhdRequirements {
    VhdRequirements(String name, UInt64 sizeInBytes, VhdType type);
    String  Name        { get; };
    UInt64  SizeInBytes { get; };
    VhdType Type        { get; };

    void SetOwner(UInt32 uid, UInt32 gid);
}
```

Pair-based `SetOwner` avoids the half-set foot-gun that per-property setters would create. Default-constructed `VhdRequirements` (or any path that does not call the new setter) gets `flags = NONE` = old behavior.

```csharp
var req = new VhdRequirements("data", 64UL << 20, VhdType.Fixed);
req.SetOwner(65534, 65534);
```

The WinRT projection has no test infrastructure in the repo (the existing `Name`/`SizeInBytes`/`Type` projection has never had tests either). Behavior is fully covered at the C SDK layer that the projection delegates to.

## Tests

* `WSLCTests.cpp`
  * `NamedVolumeVhdOptionsParseTest` — SizeBytes presence/range, unknown-key rejection (`Bogus=...`), sign rejection
  * `NamedVolumesVhdOwnership` — `mkfs -E root_owner` end-to-end (`stat -c "%u %g"` on mount root)
  * `NamedVolumesVhdFixed` — asserts on-disk `file_size >= requested`
* `WslcSdkTests.cpp`
  * Invalid type → `E_INVALIDARG`
  * `Fixed` positive (size verification on disk)
  * `OWNER` positive (uid=65534, gid=65534)
  * Unknown flag bit → `E_INVALIDARG`
  * `flags=NONE` with non-zero uid/gid silently ignored

## Validation status

* [x] `cmake --build` clean (`wslservice`, `wslcsession`, `wslcsdk`, `wslcsdkwinrt`, `wsltests`)
* [x] `FormatSource.ps1 -ModifiedOnly $true` — clang-format reports no changes

### DIFF
--- a/src/windows/WslcSDK/wslcsdk.h
+++ b/src/windows/WslcSDK/wslcsdk.h
@@ -40,7 +40,7 @@ EXTERN_C_START
 #define WSLC_E_SDK_UPDATE_NEEDED MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 11)         /* 0x8004060B */
 
 // Session values
-#define WSLC_SESSION_OPTIONS_SIZE 80
+#define WSLC_SESSION_OPTIONS_SIZE 96
 #define WSLC_SESSION_OPTIONS_ALIGNMENT 8
 
 typedef struct WslcSessionSettings
@@ -80,8 +80,21 @@ typedef enum WslcContainerNetworkingMode
 typedef enum WslcVhdType
 {
     WSLC_VHD_TYPE_DYNAMIC = 0, // Expanding VHDX (default)
-    WSLC_VHD_TYPE_FIXED = 1
+    WSLC_VHD_TYPE_FIXED = 1    // Fixed-allocation VHDX (only honored by WslcCreateSessionVhdVolume)
 } WslcVhdType;
+
+typedef enum WslcVhdRequirementsFlags
+{
+    WSLC_VHD_REQ_FLAG_NONE = 0x00000000,
+    // When set, WslcVhdRequirements::uid and gid are honored. When clear,
+    // those fields are ignored and the volume is left owned by root:root.
+    WSLC_VHD_REQ_FLAG_OWNER = 0x00000001,
+    // When set, WslcVhdRequirements::mode is honored. When clear, the field
+    // is ignored and the volume keeps the default mkfs.ext4 mode (0755).
+    WSLC_VHD_REQ_FLAG_MODE = 0x00000002,
+} WslcVhdRequirementsFlags;
+
+DEFINE_ENUM_FLAG_OPERATORS(WslcVhdRequirementsFlags);
 
 typedef struct WslcVhdRequirements
 {
@@ -89,6 +102,15 @@ typedef struct WslcVhdRequirements
     _In_z_ PCSTR name;
     _In_ uint64_t sizeBytes; // Desired size (for create/expand)
     _In_ WslcVhdType type;
+    // The remaining fields are only honored by WslcCreateSessionVhdVolume.
+    // WslcSetSessionSettingsVhd manages the session's root VHD where
+    // ownership/mode have no meaning, and rejects any non-NONE flags with
+    // E_INVALIDARG so callers do not silently miss the limitation.
+    _In_ WslcVhdRequirementsFlags flags;
+    _In_ uint32_t uid;  // honored iff (flags & WSLC_VHD_REQ_FLAG_OWNER)
+    _In_ uint32_t gid;  // honored iff (flags & WSLC_VHD_REQ_FLAG_OWNER)
+    _In_ uint32_t mode; // octal file mode (e.g. 0750), honored iff (flags & WSLC_VHD_REQ_FLAG_MODE).
+                        // Must be non-zero and <= 07777 when WSLC_VHD_REQ_FLAG_MODE is set.
 } WslcVhdRequirements;
 
 typedef enum WslcSessionFeatureFlags
--- a/src/windows/WslcSDK/wslcsdk.h
+++ b/src/windows/WslcSDK/wslcsdk.h
@@ -103,14 +103,12 @@ typedef struct WslcVhdRequirements
     _In_ uint64_t sizeBytes; // Desired size (for create/expand)
     _In_ WslcVhdType type;
     // The remaining fields are only honored by WslcCreateSessionVhdVolume.
-    // WslcSetSessionSettingsVhd manages the session's root VHD where
-    // ownership/mode have no meaning, and rejects any non-NONE flags with
-    // E_INVALIDARG so callers do not silently miss the limitation.
+    // WslcSetSessionSettingsVhd rejects non-NONE flags with E_INVALIDARG.
     _In_ WslcVhdRequirementsFlags flags;
     _In_ uint32_t uid;  // honored iff (flags & WSLC_VHD_REQ_FLAG_OWNER)
     _In_ uint32_t gid;  // honored iff (flags & WSLC_VHD_REQ_FLAG_OWNER)
-    _In_ uint32_t mode; // octal file mode (e.g. 0750), honored iff (flags & WSLC_VHD_REQ_FLAG_MODE).
-                        // Must be non-zero and <= 07777 when WSLC_VHD_REQ_FLAG_MODE is set.
+    _In_ uint32_t mode; // octal mode (e.g. 0750), non-zero and <= 07777,
+                        // honored iff (flags & WSLC_VHD_REQ_FLAG_MODE)
 } WslcVhdRequirements;
 
 typedef enum WslcSessionFeatureFlags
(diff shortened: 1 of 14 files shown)


src/windows/wslcsession/WSLCVhdVolume.cpp
```diff
@@ -119,6 +163,23 @@ std::unique_ptr<WSLCVhdVolumeImpl> WSLCVhdVolumeImpl::Create(
 
     auto mountCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { VirtualMachine.Unmount(virtualMachinePath.c_str()); });
 
+    // Apply ownership and mode to the volume root if requested. These are
+    // persisted on disk in the ext4 root inode and survive remount/recovery,
+    // so they are only applied at create time. Parse() guarantees Uid and Gid
+    // are either both present or both absent.
+    if (opts.Uid.has_value())
```
OneBlue: I don't think this is the way we should do this. Here the `chown` / `chmod` will only apply to the root directory (this will only apply to permission changes to the root, not the entire disk). 

I think we should use uid / gid when we do original mount, and then instead of offering a direct `mode` option, we should offer a `umask` option (if there's a use for it), but I think uid & gid should be enough for most usecases

src/windows/wslcsession/OptionParser.h
```diff
@@ -0,0 +1,131 @@
...
+    // MessageWslcMissingVolumeOption if Key is absent, or
+    // MessageWslcInvalidVolumeOption if the value is empty, has a leading
+    // sign, contains non-digit (for the given Base) characters, overflows, or
+    // exceeds Max.
+    template <typename T>
+    T Required(std::string_view Key, T Max = (std::numeric_limits<T>::max)(), int Base = 10);
+
+    // Same validation as Required, but returns nullopt when the key is absent.
+    template <typename T>
+    std::optional<T> Optional(std::string_view Key, T Max = (std::numeric_limits<T>::max)(), int Base = 10);
```
OneBlue: I think we can assume that `Base` is always 10 for our usecase

OneBlue (changes requested):

src/windows/wslcsession/OptionParser.h
benhillis: mode was octal, but that was removed.

src/windows/wslcsession/WSLCVhdVolume.cpp
benhillis: I think the right approach is to pass the uid / gid to mkfs.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -80,15 +80,30 @@ typedef enum WslcContainerNetworkingMode
...
+
+DEFINE_ENUM_FLAG_OPERATORS(WslcVhdRequirementsFlags);
+
 typedef struct WslcVhdRequirements
 {
     // Ignored by WslcSetSessionSettingsVhd
     _In_z_ PCSTR name;
     _In_ uint64_t sizeBytes; // Desired size (for create/expand)
     _In_ WslcVhdType type;
+    // The remaining fields are only honored by WslcCreateSessionVhdVolume.
```
JohnMcPMS, line 102: Should we just make the session creation setting value *only* the size, rather than taking in a struct in which every field is ignored except the size?  Or do we expect that these limitations are temporary and the type/flags/owner properties are eventually to be implemented?

Not saying you need to do it in this change, I will create one if it is deemed a good idea.  An option is another struct (`WslcSessionCreationVhdRequirements`) that contains only the values valid for session creation that `WslcVhdRequirements` derives from.  If/as fields are implemented for the primary VHD, they can be shifted without impacting code as written.

OneBlue (approved):
LGTM

src/windows/wslcsession/OptionParser.h
```diff
@@ -0,0 +1,121 @@
...
+        return std::nullopt;
+    }
+
+    return ParseUnsignedValue<T>(Key, *value, Max);
+}
+
+template <typename T>
+inline T OptionParser::ParseUnsignedValue(std::string_view Key, const std::string& Value, T Max)
+{
+    static_assert(std::is_unsigned_v<T>, "OptionParser numeric accessors only support unsigned integer types");
```
JohnMcPMS, line 99:
```suggestion
    static_assert(std::is_integral_v<T> && std::is_unsigned_v<T> && sizeof(T) <= sizeof(uint64_t), "OptionParser numeric accessors only support unsigned integer types up to 64 bits");
```
super-duper-nit: if an unsigned floating-point type were to exist (and `T(-1)` were meaningful), this would allow it.

nit: More realistically, a larger unsigned integral type might exist.

src/windows/wslcsession/OptionParser.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+namespace wsl::windows::service::wslc {
+
+OptionParser::OptionParser(const std::map<std::string, std::string>& Options) noexcept : m_options(Options)
+{
+}
+
+const std::string* OptionParser::Find(std::string_view Key)
+{
+    const auto it = m_options.find(std::string(Key));
```
JohnMcPMS, line 29: If you always have to convert to a `std::string`, consider a `const std::string&`.  If there are an annoying number of cases where you hold a `std::string_view`, have an override/view specific function.

JohnMcPMS (commented):
Went to lunch while it was open...

PR #40480, opened 2026-05-09
Fix shared ptr circular reference leaks

chemwolf6922:

Both loggers have their member thread holding a shared pointer of themselves. This renders the m_exitEvents logic useless. And could lead to resource leaks.
Since the thread's lifespan is always shorter than the logger's, changing the capture from shared_from_this to this will fix the leak.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/40470

### DIFF
--- a/src/windows/common/Dmesg.cpp
+++ b/src/windows/common/Dmesg.cpp
@@ -72,13 +72,13 @@ std::pair<std::wstring, std::thread> DmesgCollector::StartDmesgThread(InputSourc
 
     THROW_LAST_ERROR_IF(!pipe);
 
-    auto workerThread = std::thread([Self = shared_from_this(), Source, Pipe = std::move(pipe)]() {
+    auto workerThread = std::thread([this, Source, Pipe = std::move(pipe)]() {
         try
         {
             wsl::windows::common::wslutil::SetThreadDescription(L"Dmesg");
 
             // When the pipe connects, start reading data.
-            wsl::windows::common::helpers::ConnectPipe(Pipe.get(), INFINITE, Self->m_exitEvents);
+            wsl::windows::common::helpers::ConnectPipe(Pipe.get(), INFINITE, m_exitEvents);
 
             std::vector<char> buffer(LX_RELAY_BUFFER_SIZE);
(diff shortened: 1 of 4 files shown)


OneBlue (changes requested):
This change makes sense to break the reference cycle, but that means that `this` can now be deleted while the worker threads are still running. 

If we want to make this change, we need to synchronize the worker threads with the deletion of the instance that they reference (the easiest way would be for both class to keep track of their worker threads, and signal their threads to exit + wait for them to actually exit in the destructor)

JohnMcPMS:
> This change makes sense to break the reference cycle, but that means that `this` can now be deleted while the worker threads are still running.
> 
> If we want to make this change, we need to synchronize the worker threads with the deletion of the instance that they reference (the easiest way would be for both class to keep track of their worker threads, and signal their threads to exit + wait for them to actually exit in the destructor)

`weak_from_this` and any failure to resolve a `shared_ptr` indicates a deleted parent and thus the thread should just exit?

OneBlue:
> `weak_from_this` and any failure to resolve a `shared_ptr` indicates a deleted parent and thus the thread should just exit?

Yeah that could work,  although I would have a preference for synchronizing the thread exit with the VM exit, since that would guarantee that we're not leaking threads (and in the past runaway threads have caused various issues)

chemwolf6922:
Hi @OneBlue , @JohnMcPMS . Both logger's destructor will break and join the thread. I don't think delete this would cause a lifecycle problem.
https://github.com/microsoft/WSL/blob/4ee78818dbef570a8191e3dbbe36631ec2a8d1bb/src/windows/common/Dmesg.cpp#L33-L45

OneBlue (approved):
> Hi @OneBlue , @JohnMcPMS . Both logger's destructor will break and join the thread. I don't think delete this would cause a lifecycle problem.

That's true ! I assumed that the thread never exited since the original issue mentioned the thread being leaked, but that was actually because of the circular reference causing the logger itself to never be torn down. 

Sorry for the incorrect review, change LGTM !

PR #40501, opened 2026-05-11
Skip localization PR creation on file-change CI trigger

benhillis:
Gate the `Create pull request` step in the nightly localization pipeline on `Build.Reason in (Schedule, Manual)` so it does not run when the pipeline is triggered by a push that modified `localization/strings/en-US/Resources.resw`.

Without this, every change to `Resources.resw` (e.g., #40499) kicks off the pipeline, which then runs Touchdown and opens a follow-up localization PR even though there are no new strings to localize. Those PRs consume build resources without changing any localized strings.

After this change:
- Nightly scheduled runs still create localization PRs as before.
- Manual pipeline runs still create PRs (useful for testing).
- File-change CI runs still execute Touchdown but skip PR creation.

### DIFF
--- a/.pipelines/wsl-build-nightly-localization.yml
+++ b/.pipelines/wsl-build-nightly-localization.yml
@@ -43,6 +43,7 @@ steps:
     cultureMappingType: 'None'
 
 - task: PipAuthenticate@1
+  condition: and(succeeded(), in(variables['Build.Reason'], 'Schedule', 'Manual'))
   inputs:
     artifactFeeds: 'wsl'
 
--- a/.pipelines/wsl-build-nightly-localization.yml
+++ b/.pipelines/wsl-build-nightly-localization.yml
@@ -51,6 +51,7 @@ steps:
             python tools/devops/create-change.py . "$env:token" "WSL localization"  "Localization change from build: $env:buildId" "user/localization/$env:buildId" "$env:targetBranch"
 
   displayName: Create pull request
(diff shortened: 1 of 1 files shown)


.pipelines/wsl-build-nightly-localization.yml
```diff
@@ -51,6 +51,7 @@ steps:
             python tools/devops/create-change.py . "$env:token" "WSL localization"  "Localization change from build: $env:buildId" "user/localization/$env:buildId" "$env:targetBranch"
 
   displayName: Create pull request
+  condition: and(succeeded(), in(variables['Build.Reason'], 'Schedule', 'Manual'))
```
OneBlue, line 55: nit: We could also skip the two above tasks if we don't want to create a PR

OneBlue (dismissed):

.pipelines/wsl-build-nightly-localization.yml
benhillis: Good call - also gated PipAuthenticate since it's only there to support create-change.py. Left TouchdownBuildTask ungated so new source strings still get submitted to the translation service on file-change CI runs.

OneBlue (changes requested):
The latest iteration is missing the `condition` on the pip install invocation (which will fail without the pip authenticate part)

OneBlue (approved):
My previous review was wrong ! 

LGTM

PR #40505, opened 2026-05-12
Set SO_UPDATE_ACCEPT_CONTEXT on sockets accepted via AcceptEx in PortRelay

wangxin12:
After `AcceptEx` completes, set `SO_UPDATE_ACCEPT_CONTEXT` to mark the socket as connected. Without this, `getpeername`/`getsockname`/`shutdown` and a number of `setsockopt`/`getsockopt` options fail on the accepted socket.

## PR Checklist


src/windows/wslrelay/localhost.cpp
ptrivedi, line 403: nit: Would be good to add a comment here to indicate that CompleteAccept now also sets up accept context in addition to its previous functionality of completing overlapped I/O

benhillis:
Can you add a regression test please?

src/windows/wslrelay/localhost.cpp
wangxin12: Added.

wangxin12:
> Can you add a regression test please?

Hi @benhillis , I looked into this and ran into a few obstacles that make a meaningful regression test hard. Wanted to walk through them and get your input on the best path forward.

1. The accepted socket is not exposed, so we can't directly call APIs like `shutdown` on it to test.
2. The current `wsl::windows::common::relay::BidirectionalRelay` logs and swallows the failure, and relay loop tears down both directions on the first EOF either way. From the client's perspective the connection still closes cleanly — just via `closesocket` instead of orderly `shutdown`.
```
LOG_LAST_ERROR_IF(shutdown(reinterpret_cast<SOCKET>(LeftHandle), SD_SEND) == SOCKET_ERROR)
```

src/windows/wslcsession/PortRelayHandle.cpp
```diff
@@ -83,6 +83,12 @@ void PortRelayAcceptHandle::Collect()
         THROW_IF_WIN32_BOOL_FALSE(WSAGetOverlappedResult(ListenSocket.get(), &Overlapped, &bytesReceived, false, &flagsReturned));
     }
 
+    // Set the accept context to mark the socket as connected.
+    const SOCKET listenSocket = ListenSocket.get();
```
OneBlue: nit: We could we move this to a shared helper in `socket.hpp` ? This is done in multiple places

OneBlue (approved):
LGTM

PR #40515, opened 2026-05-12
Localize ADMX Group Policy strings via Touchdown

benhillis:
Adds the WSL ADMX policy strings to the nightly Touchdown localization pipeline so translated `.adml` files are produced for every locale already covered by `Resources.resw`. Translated files land at `intune/<locale>/WSL.adml`, matching the Group Policy on-disk layout (the language-neutral `WSL.admx` stays at `intune/WSL.admx`).

## Changes

- **`intune/en-US/WSL.adml`** — new baseline ADML file. Each translatable string is annotated with `{Locked="..."}` XML comments so translators preserve product names (Windows Subsystem for Linux, WSL, WSL1, WSL2, Store WSL), `.wslconfig` keys (`wsl2.kernel`, `wsl2.systemDistro`, `wsl2.kernelCommandLine`, `wsl2.nestedVirtualization`, `wsl2.kernelDebugPort`, `wsl2.networkingMode`, `wsl2.firewall`), command names (`wsl.exe`, `debug-shell`, `mount`), networking-mode enum values (`None`, `NAT`, `Mirrored`, `VirtioProxy`), and the policy state literal `Disabled`.
- **`.pipelines/wsl-build-nightly-localization.yml`** — adds `intune/en-US/WSL.adml` to the trigger paths and to the `TouchdownBuildTask` `resourceFilePath` block. The combined `;O:intune\` mapping puts translated files at `intune/<locale>/WSL.adml`.
- **`tools/devops/validate-localization.py`** — adds `validate_adml()` enforcing the same locked-token invariant as `.resw` (every `{Locked="X"}` token must appear verbatim in its target string value), id parity for both `<stringTable>` and `<presentationTable>` between baseline and translated files, and a `<!-- ... -->` based comment scan for locked tokens (since ElementTree drops comments). All CLI args are now `--option` form with module-level defaults (`RESOURCE_FOLDER`, `BASELINE_LANGUAGE='en-US'`, `ADML_FOLDER='intune'`); the CI fast path runs zero-arg with no `pip install click` dependency. Local overrides use the click flag form.
- **`.pipelines/build-stage.yml`** — invokes the validator with no args (was `localization/strings en-US`); the script now sources both folders and the baseline language from its module-level defaults.
- **`tools/devops/create-change.py`** — includes untracked files in the `Changed files:` report and stages them with `git add -A` (via `repo.git.add(A=True)`) before commit. Without this, Touchdown's first-time ADML outputs would land as untracked files in newly created `intune/<locale>/` directories and be silently dropped by `git commit -a`, which only stages modifications and deletions of already-tracked files.

## Verification

Manually queued `wsl-github-localization` against this branch (ADO build [`146781281`](https://dev.azure.com/microsoft/Microsoft.WSL/_build/results?buildId=146781281)). Touchdown returned `WSL.adml` translations for 20 locales (cs-CZ, da-DK, de-DE, en-GB, es-ES, fi-FI, fr-FR, hu-HU, it-IT, ja-JP, ko-KR, nb-NO, nl-NL, pl-PL, pt-BR, pt-PT, ru-RU, sv-SE, tr-TR, zh-CN, zh-TW) and saved them to disk at `intune/<locale>/WSL.adml`. A subsequent run with the `create-change.py` fix in this branch will produce a follow-up auto-PR containing those `.adml` files alongside the usual `.resw` updates. Validator was sanity-tested zero-arg under both Windows and WSL Linux (case-sensitive FS): 22 resw locales + ADML baseline pass; injecting bogus locked tokens or removing strings/presentations from a fake locale correctly fails with exit 1.

Note: depends on no other PR. The unrelated nightly-pipeline target-branch fix that surfaced during this debugging is in #40510.

### DIFF
--- a/.pipelines/build-stage.yml
+++ b/.pipelines/build-stage.yml
@@ -104,7 +104,7 @@ stages:
           pool: ${{ parameters.pool }}
 
         steps:
-          - script: python tools/devops/validate-localization.py localization/strings en-US
+          - script: python tools/devops/validate-localization.py
             displayName: Validate localization resources
 
           - script: python tools\devops\validate-copyright-headers.py src
--- a/intune/en-US/WSL.adml
+++ b/intune/en-US/WSL.adml
@@ -5,57 +5,83 @@
   <description>Windows Subsystem for Linux</description>
   <resources>
     <stringTable>
+        <!-- {Locked="Windows Subsystem for Linux"} -->
         <string id="WindowsSubsystemForLinux">Windows Subsystem for Linux</string>
 
+        <!-- {Locked="Windows Subsystem for Linux"} -->
         <string id="AllowWSL">Allow the Windows Subsystem for Linux</string>
+        <!-- {Locked="Windows Subsystem for Linux"} -->
         <string id="AllowWSLExplain">When set to disabled, this policy disables access to running Linux distributions in the Windows Subsystem for Linux for all users on the machine.</string>
 
+        <!-- {Locked="Windows Subsystem for Linux"} -->
         <string id="AllowInboxWSL">Allow the Inbox version of the Windows Subsystem for Linux</string>
+        <!-- {Locked="Windows Subsystem for Linux"}{Locked="WSL"} -->
         <string id="AllowInboxWSLExplain">When set to disabled, this policy disables the inbox version (optional component) of the Windows Subsystem for Linux. If this policy is disabled, only the store version of WSL can be used.</string>
 
+        <!-- {Locked="WSL1"} -->
         <string id="AllowWSL1">Allow WSL1</string>
+        <!-- {Locked="WSL1"}{Locked="WSL2"} -->
         <string id="AllowWSL1Explain">When set to disabled, this policy disables WSL1. When disabled, only WSL2 distributions can be used.</string>
 
         <string id="CustomKernelUserSettingConfigurable">Allow custom kernel configuration</string>
+        <!-- {Locked=".wslconfig"}{Locked="wsl2.kernel"}{Locked="Store WSL"} -->
         <string id="CustomKernelExplain">When set to disabled, this policy disables custom kernel configuration via .wslconfig (wsl2.kernel). This policy only applies to Store WSL.</string>
 
         <string id="CustomSystemDistroUserSettingConfigurable">Allow custom system distribution configuration</string>
+        <!-- {Locked=".wslconfig"}{Locked="wsl2.systemDistro"}{Locked="Store WSL"} -->
         <string id="CustomSystemDistroExplain">When set to disabled, this policy disables custom system distribution configuration via .wslconfig (wsl2.systemDistro). This policy only applies to Store WSL.</string>
 
         <string id="CustomKernelCommandLineUserSettingConfigurable">Allow kernel command line configuration</string>
+        <!-- {Locked=".wslconfig"}{Locked="wsl2.kernelCommandLine"}{Locked="Store WSL"} -->
         <string id="CustomKernelCommandLineExplain">When set to disabled, this policy disables kernel command line configuration via .wslconfig (wsl2.kernelCommandLine). This policy only applies to Store WSL.</string>
 
         <string id="AllowDebugShell">Allow the debug shell</string>
+        <!-- {Locked="wsl.exe"}{Locked="debug-shell"}{Locked="Store WSL"} -->
         <string id="AllowDebugShellExplain">When set to disabled, this policy disables the debug shell (wsl.exe --debug-shell). This policy only applies to Store WSL.</string>
 
         <string id="NestedVirtualizationUserSettingConfigurable">Allow nested virtualization</string>
+        <!-- {Locked=".wslconfig"}{Locked="wsl2.nestedVirtualization"}{Locked="Store WSL"} -->
         <string id="NestedVirtualizationExplain">When set to disabled, this policy disables nested virtualization configuration via .wslconfig (wsl2.nestedVirtualization). This policy only applies to Store WSL.</string>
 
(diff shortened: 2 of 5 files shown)


intune/en-US/WSL.adml
```diff
@@ -5,57 +5,83 @@
...
+        <!-- {Locked=".wslconfig"}{Locked="wsl2.nestedVirtualization"}{Locked="Store WSL"} -->
         <string id="NestedVirtualizationExplain">When set to disabled, this policy disables nested virtualization configuration via .wslconfig (wsl2.nestedVirtualization). This policy only applies to Store WSL.</string>
 
         <string id="KernelDebugUserSettingConfigurable">Allow kernel debugging</string>
+        <!-- {Locked=".wslconfig"}{Locked="wsl2.kernelDebugPort"}{Locked="Store WSL"} -->
         <string id="KernelDebugExplain">When set to disabled, this policy disables kernel debugging configuration via .wslconfig (wsl2.kernelDebugPort). This policy only applies to Store WSL.</string>
 
         <string id="CustomNetworkingUserSettingConfigurable">Allow custom networking configuration</string>
+        <!-- {Locked=".wslconfig"}{Locked="wsl2.networkingmode"}{Locked="Store WSL"} -->
         <string id="CustomNetworkingExplain">When set to disabled, this policy disables custom networking configuration via .wslconfig (wsl2.networkingmode). This policy only applies to Store WSL.</string>
```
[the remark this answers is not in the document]
benhillis: Good catch — fixed both occurrences in 8287582a (token + explain string).

tools/devops/validate-localization.py
```diff
@@ -194,15 +278,16 @@ def run(resource_folder: str, baseline_language: str, fix: bool):
...
-        run(sys.argv[1], sys.argv[2], False)
+        run(sys.argv[1], sys.argv[2], False, ADML_FOLDER)
     else:
         import click
 
         @click.command()
         @click.argument('resource-folder', default='localization/strings')
         @click.argument('baseline-language', default='en-us')
+        @click.option('--adml-folder', default=ADML_FOLDER, show_default=True)
         @click.option('--fix', is_flag=True)
```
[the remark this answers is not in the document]
benhillis: Already fixed in an earlier amend (88b8b6bd) — module-level default is now `BASELINE_LANGUAGE = 'en-US'`.

.pipelines/build-stage.yml
```diff
@@ -104,7 +104,7 @@ stages:
           pool: ${{ parameters.pool }}
 
         steps:
-          - script: python tools/devops/validate-localization.py localization/strings en-US
+          - script: python tools/devops/validate-localization.py
             displayName: Validate localization resources
```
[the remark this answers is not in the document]
benhillis: Good catch on the stale description — updated. The previous positional form (`localization/strings en-US`) was only used by `.pipelines/build-stage.yml` (now zero-arg) and is documented as zero-arg in `.github/copilot-instructions.md`; nothing else in the tree calls it positionally.

tools/devops/validate-localization.py
```diff
@@ -174,7 +174,106 @@ def fix_comments(comments: dict, path: str, strings: dict):
...
+    `<string>` element. Non-Locked comments are ignored.
+    """
+    with open(path, 'rb') as fd:
+        content = fd.read()
+
+    root = xml.etree.ElementTree.fromstring(content)
+
+    strings = {}
+    string_table = root.find(f'.//{ADML_NS}stringTable')
+    if string_table is not None:
```
OneBlue: `string_table` should never be None right ?

tools/devops/validate-localization.py
```diff
@@ -174,7 +174,106 @@ def fix_comments(comments: dict, path: str, strings: dict):
...
+        strings = {sid: (text, []) for sid, text in
+                   ((s.get('id'), s.text or '') for s in string_table.findall(f'{ADML_NS}string'))
+                   if sid is not None}
+
+        # ElementTree drops comments; re-scan raw text to associate Locked tokens with their <string>.
+        text = content.decode('utf-8').lstrip('\ufeff')
+        pattern = re.compile(r'<!--(.*?)-->\s*<string\s+id="([^"]+)"', re.DOTALL)
```
OneBlue: Hmm this is feeling a bit brittle (using a regex to scan the entire document and try to match the comment based on the entire node. 

Could we use what's suggested in https://stackoverflow.com/questions/59543824/how-to-read-commented-text-from-xml-file-in-python to get the comments in the tree instead ?

benhillis: yeah wasn't a huge fan of this approach either, let me try to massage.

benhillis: Reworked in 7e8c14a6. Now uses an XMLParser with TreeBuilder(insert_comments=True) and walks stringTable children in document order, accumulating Locked tokens from preceding comment nodes and resetting after each <string>. The whole-document regex scan is gone. Also dropped the defensive 'is not None' guards on stringTable / presentationTable per the other thread -- a missing table now raises with a clear error message.

JohnMcPMS (approved):
My python isn't great but it seems like it should work generally if it is running now.

tools/devops/create-change.py
```diff
@@ -33,7 +35,9 @@ def main(repo_path: str, token: str, committer: str, message: str, branch: str,
             config.set_value("user", "email", COMMITTER_EMAIL)
             config.set_value("user", "name", committer)
 
-        repo.git.commit('-a', m=message)
+        # 'git add -A' so newly created files in new directories are staged too.
+        repo.git.add(A=True)
```
ptrivedi, line 39: nit: won't this add untracked files as well? i guess we could assume there won't ever be any, but would it make sense to only git add intune and localization folders?

PR #40518, opened 2026-05-13
Improve VirtioFs performance with swiotlb

asherkariv:
## Summary

Improve VirtioFs performance by adding swiotlb (Software Translation Lookaside Buffer) bounce buffer support for virtio devices. Swiotlb enables more efficient DMA handling for VirtioFs, reducing overhead in file system operations.

## PR Checklist
- [x] Builds clean (Release x64)
- [x] Code review passed

### Commit 1: Add experimental.swiotlb .wslconfig setting
- Adds a new `[experimental]` config key `swiotlb` to specify custom swiotlb configuration
- Adds localization strings and validation

### Commit 2: Swiotlb gating, vmId plumbing, and devicehost override
- **Swiotlb gating for VirtioFs perf**: Automatically compute default swiotlb bounce buffer config when VirtioFs (or other virtio devices) are enabled, enabling efficient DMA for file system operations
- **Virtio9p deprecation ordering**: Move the virtio9p deprecation/disable block before swiotlb computation so deprecated devices don't trigger unnecessary bounce buffer config
- **vmId plumbing**: Inject vmId=<machineId> into all virtio device options via \GuestDeviceManager::AddHdvShareWithOptions\ for proper device-to-VM association
- **WSL_DEVICE_HOST_DLL override**: Add CMake option to replace wsldevicehost.dll without a NuGet update (dev-loop improvement)
- **Fix empty-options guard**: Prevent double semicolons when appending to device options
- **Package update**: Update packages.config

## Validation Steps
- Built successfully in Release x64 configuration
- Code reviewed for logic correctness (swiotlb gating order verified)

### DIFF
--- a/localization/strings/en-US/Resources.resw
+++ b/localization/strings/en-US/Resources.resw
@@ -802,6 +802,10 @@ For more information on Admin Protection, please visit https://aka.ms/apdevguide
     <value>Invalid IP value '{}' for key '{}' in {}:{}</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
+  <data name="MessageConfigInvalidSwiotlb" xml:space="preserve">
+    <value>Invalid SWIOTLB value '{}' for key '{}' in {}:{}. Expected format: '0x&lt;hex&gt;,&lt;size&gt;K|M'</value>
+    <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
+  </data>
   <data name="MessageConfigInvalidBoolean" xml:space="preserve">
     <value>Invalid boolean value '{}' for key '{}' in {}:{}</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
@@ -1125,6 +1129,14 @@ Attempting web download...</value>
   </data>
   <data name="MessageConfigVirtio9pDisabled" xml:space="preserve">
     <value>wsl2.virtio9p is disabled, falling back to 9p with vsock transport.</value>
+  </data>
+  <data name="MessageVirtioProxyRequiresVirtio" xml:space="preserve">
+    <value>VirtioProxy networking is not supported when wsl2.virtio is disabled; falling back to networkingMode {}.</value>
+    <comment>{Locked="wsl2.virtio"}{Locked="VirtioProxy"}{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
+  </data>
+  <data name="MessageSwiotlbKernelUnsupported" xml:space="preserve">
+    <value>The running kernel is missing a patch that significantly improves virtio device performance. Update to a more recent WSL kernel to enable this optimization.</value>
+    <comment>{Locked="virtio"}</comment>
   </data>
   <data name="MessageInstallationCorrupted" xml:space="preserve">
     <value>WSL installation appears to be corrupted (Error code: {}).
--- a/src/shared/inc/lxinitshared.h
+++ b/src/shared/inc/lxinitshared.h
@@ -1430,9 +1430,10 @@ typedef struct _LX_INIT_GUEST_CAPABILITIES
 
     MESSAGE_HEADER Header;
     bool SeccompAvailable;
+    bool KernelSupportsHvPciSwiotlb;
     char Buffer[]; // Contains the kernel version string
 
-    PRETTY_PRINT(FIELD(Header), FIELD(SeccompAvailable), BUFFER_FIELD(Buffer));
+    PRETTY_PRINT(FIELD(Header), FIELD(SeccompAvailable), FIELD(KernelSupportsHvPciSwiotlb), BUFFER_FIELD(Buffer));
 } LX_INIT_GUEST_CAPABILITIES, *PLX_INIT_GUEST_CAPABILITIES;
 
 typedef struct _LX_MINI_INIT_WAIT_FOR_PMEM_DEVICE_MESSAGE
--- a/src/windows/common/DeviceHostProxy.cpp
+++ b/src/windows/common/DeviceHostProxy.cpp
@@ -133,6 +133,20 @@ void DeviceHostProxy::Shutdown()
 {
     {
         auto lock = m_lock.lock_exclusive();
+
+        for (auto& entry : m_fileSystems)
+        {
+            try
+            {
+                wil::com_ptr<IPlan9FileSystem> instance;
+                THROW_IF_FAILED(
+                    m_git->GetInterfaceFromGlobal(entry.Cookie, __uuidof(IPlan9FileSystem), reinterpret_cast<void**>(instance.put())));
+
+                LOG_IF_FAILED(instance->Teardown());
+            }
+            CATCH_LOG()
+        }
+
         m_fileSystems.clear();
         m_shutdown = true;
     }
--- a/src/windows/common/WslCoreConfig.cpp
+++ b/src/windows/common/WslCoreConfig.cpp
@@ -14,6 +14,7 @@ Abstract:
 
 #include "precomp.h"
 #include "WslCoreConfig.h"
+#include "helpers.hpp"
 #include "Localization.h"
 #include "WslCoreFirewallSupport.h"
 #include "WslCoreNetworkingSupport.h"
@@ -61,6 +62,24 @@ void wsl::core::Config::ParseConfigFile(_In_opt_ LPCWSTR ConfigFilePath, _In_opt
         {
             DnsTunnelingIpAddress = address.S_un.S_addr;
         }
+    };
+
+    auto parseSwiotlb = [&](const char* name, const char* value, const wchar_t* fileName, unsigned long fileLine) {
+        // If the value does not conform to the expected format, SWIOTLB customization is disabled.
+        SwiotlbConfig.clear();
+        try
+        {
+            auto wideValue = wsl::shared::string::MultiByteToWide(value);
+            std::wregex swiotlbPattern(L"^(0x[0-9a-fA-F]+,[0-9]+[mk])$", std::regex::icase);
+            if (!std::regex_match(wideValue, swiotlbPattern))
+            {
+                EMIT_USER_WARNING(shared::Localization::MessageConfigInvalidSwiotlb(value, name, fileName, fileLine));
+                return;
+            }
+
+            SwiotlbConfig = std::move(wideValue);
+        }
+        CATCH_LOG()
     };
 
     ConfigKeyPresence earlyBootLoggingPresent{};
@@ -124,7 +143,8 @@ void wsl::core::Config::ParseConfigFile(_In_opt_ LPCWSTR ConfigFilePath, _In_opt
         ConfigKey(ConfigSetting::Experimental::InitialAutoProxyTimeout, InitialAutoProxyTimeout),
         ConfigKey(ConfigSetting::Experimental::IgnoredPorts, std::move(parseIgnoredPorts)),
         ConfigKey(ConfigSetting::Experimental::HostAddressLoopback, EnableHostAddressLoopback),
-        ConfigKey(ConfigSetting::Experimental::SetVersionDebug, SetVersionDebug)};
+        ConfigKey(ConfigSetting::Experimental::SetVersionDebug, SetVersionDebug),
+        ConfigKey(ConfigSetting::Experimental::Swiotlb, std::move(parseSwiotlb))};
 
     wil::unique_file ConfigFile;
     if (ConfigFilePath != nullptr)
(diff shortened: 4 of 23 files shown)


src/shared/inc/lxinitshared.h
```diff
@@ -162,6 +162,7 @@ Module Name:
 //
 
 #define LX_INIT_DEFAULT_PLAN9_MOUNT_OPTIONS ";uid=1000;gid=1000;symlinkroot=/mnt/"
+#define LX_INIT_SWIOTLB_MOUNT_OPTIONS ";swiotlb="
```
benhillis: if this is only used by the Windows side it does not need to be in the shared header.

src/windows/common/WslCoreConfig.cpp
```diff
@@ -63,6 +67,29 @@ void wsl::core::Config::ParseConfigFile(_In_opt_ LPCWSTR ConfigFilePath, _In_opt
...
+            auto wideValue = wsl::shared::string::MultiByteToWide(value);
+            std::wregex swiotlbPattern(c_swiotlbPattern, std::regex::icase);
+            if (!std::regex_match(wideValue, swiotlbPattern))
+            {
+                EMIT_USER_WARNING(shared::Localization::MessageConfigInvalidSwiotlb(value, name, fileName, fileLine));
+                return;
+            }
+
+            SwiotlbCfg = std::move(wideValue);
+
```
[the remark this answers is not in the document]
benhillis: This seems like valid feedback.

src/windows/common/WslCoreConfig.cpp
```diff
@@ -63,6 +67,29 @@ void wsl::core::Config::ParseConfigFile(_In_opt_ LPCWSTR ConfigFilePath, _In_opt
         }
     };
 
+    auto parseSwiotlb = [&](const char* name, const char* value, const wchar_t* fileName, unsigned long fileLine) {
```
benhillis: I don't love this UX. How is a user supposed to configure this? Why would they want to pick one value versus another? Why do we need this to be configurable?

Also as this is currently structured it will only work for WSL, not WSLC.

src/windows/service/inc/wslservice.idl
```diff
@@ -140,6 +140,8 @@ cpp_quote("#define LXSS_VM_MODE_INITRD_NAME L\"initrd.img\"")
 cpp_quote("#define LXSS_VM_MODE_KERNEL_NAME L\"kernel\"")
 cpp_quote("#define LXSS_VM_MODE_VHD_NAME L\"ext4.vhdx\"")
 
+cpp_quote("#define LXSS_VM_KERNEL_SWIOTLB_CONFIG L\"hv_pci_swiotlb=\"")
```
benhillis: is this the kernel config we are looking at adding?

src/windows/common/WslCoreConfig.cpp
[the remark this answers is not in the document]
asherkariv: If the input is invalid, SwiotlbCfg keeps its default value, which is the expected behaviour.

src/shared/inc/lxinitshared.h
asherkariv: I was contemplating with this, by you are right, this information is only used on the Windows side (wsldevicehost.dll). I will remove it.

src/windows/service/inc/wslservice.idl
asherkariv: Yes, this is the kernel command line we use when swiotlb is specified with a range (rather than lowmem), e.g.,
"hv_pci_swiotlb=0x100000000,64M"

src/windows/common/WslCoreConfig.cpp
asherkariv: In practice the user will not touch this configuration parameter. We want to be able to configure swiotlb with either a range, or using low-memory. This is also why I put it under experimental (see the .wslconfig snippet below). Using a range requires the kernel patch, and low-memory does not. After experimenting with it we can set the default value. 

The current behavior is that by default, if no "swiotlb=..." is supplied, the the system behaves exactly like today with the cache disabled. When "swiotlb=..." is provided, swiotlb is configured and cache is enabled.
Please keep in mind, that the "cache" is not really a cache anymore, it is a static set of apertures, that never get evicted.

Does that make sense?

[wsl2]
...

[experimental]
#swiotlb=0x100000000,64M
swiotlb=lowmem

src/windows/service/inc/wslservice.idl
benhillis: ok, that kernel update isn't included in this PR so this won't do anything, right?

src/shared/inc/lxinitshared.h
asherkariv: Fixed

src/windows/common/WslCoreConfig.cpp
asherkariv: Set lowmem as the default configuration for swiotlb, no explicit .wslconfig parameter is needed. swiotlb perf benefits enabled by default.

src/windows/common/WslCoreConfig.cpp
```diff
@@ -462,6 +490,13 @@ void wsl::core::Config::Initialize(_In_opt_ HANDLE UserToken)
         EnableVirtio9p = false;
     }
 
+    if (!SwiotlbCfg.empty() && !EnableVirtioFs)
+    {
+        EMIT_USER_WARNING(wsl::shared::Localization::MessageSwiotlbRequiresVirtioFs());
+        SwiotlbCfg.clear();
+        SwiotlbKernelCfg.clear();
+    }
```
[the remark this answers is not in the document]
asherkariv: test: add experimental.swiotlb coverage to WSL2 Warnings test

Extend WSL2_TEST_METHOD(Warnings) in test/windows/UnitTests.cpp with
validateWarnings cases for the new experimental.swiotlb .wslconfig
setting:

  * Valid gpa,size value (0x100000000,64M) with virtiofs disabled
    (default) -- expects the "requires virtiofs" warning and the
    setting to be cleared.
  * Valid "lowmem" value with virtiofs disabled -- same warning.
  * Malformed value ("garbage") -- expects the parser
    "Invalid SWIOTLB value ..." warning; the field stays empty so the
    virtiofs requirement check does not fire.
  * Unsupported size suffix ("0x100000000,64g") -- same parser warning.
  * Valid "lowmem" with virtiofs enabled (DrvFsMode::VirtioFs prefix)
    -- expects no warnings.
  * Valid gpa,size value with virtiofs enabled -- expects no warnings.

Placed next to the existing experimental.ignoredPorts cases; reuses
LxssGenerateTestConfig({.drvFsMode = DrvFsMode::VirtioFs}) to enable
virtiofs in the positive cases. No header changes required.

src/windows/common/WslCoreConfig.h
```diff
@@ -381,5 +382,8 @@ struct Config
 
     GUID NatNetworkId() const noexcept;
     LPCWSTR NatNetworkName() const noexcept;
+
+    std::wstring SwiotlbCfg = L"lowmem";
```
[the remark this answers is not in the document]
asherkariv: Fixed

benhillis:
I might be confused but I thought we didn’t need this setting until we have the updated kernel?

test/windows/UnitTests.cpp
```diff
@@ -2104,6 +2104,38 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
...
+        //
+        // The default test config has virtiofs disabled, so any (otherwise valid) value is
+        // parsed successfully but then cleared during Config::Initialize with a warning.
+        validateWarnings(
+            L"[experimental]\nswiotlb=lowmem",
+            L"wsl: experimental.swiotlb requires virtiofs to be enabled; the setting will be ignored.\r\n");
+
+        validateWarnings(
+            L"[experimental]\nswiotlb=0x100000000,64M",
+            L"wsl: experimental.swiotlb requires virtiofs to be enabled; the setting will be ignored.\r\n");
```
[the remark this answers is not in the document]
asherkariv: Fixed

src/windows/common/helpers.cpp
```diff
@@ -741,4 +742,24 @@ void wsl::windows::common::helpers::AppendCommonKernelCommandLine(_Inout_ std::w
 
     // Configure page reporting order - minimum order of pages reported as free to the hypervisor.
     kernelCmdLine += std::format(L" page_reporting.page_reporting_order={}", pageReportingOrder);
+
+    // Reserve a swiotlb bounce buffer for virtio devices.
+    if (!swiotlbConfig.empty())
+    {
+        kernelCmdLine += std::format(L" swiotlb=force hv_pci_swiotlb={}", swiotlbConfig);
```
benhillis: I think we should be adding both, right @asherkariv ?

asherkariv: Correct

localization/strings/en-US/Resources.resw
```diff
@@ -802,6 +802,10 @@ For more information on Admin Protection, please visit https://aka.ms/apdevguide
     <value>Invalid IP value '{}' for key '{}' in {}:{}</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
+  <data name="MessageConfigInvalidSwiotlb" xml:space="preserve">
+    <value>Invalid SWIOTLB value '{}' for key '{}' in {}:{}. Expected format: '0x&lt;hex&gt;,&lt;size&gt;K|M'</value>
```
ptrivedi: nit: Looks like your regex for the format allows for case-insensitive k|m as well. It lists only the uppercase as the expected format in other places in the code as well

src/windows/common/helpers.cpp
```diff
@@ -772,4 +773,27 @@ void wsl::windows::common::helpers::AppendCommonKernelCommandLine(_Inout_ std::w
...
```
ptrivedi: Don't you need a sizing parameter for swiotlb (especially) along with 'force'? Without the sizing parameter, I am reading that it will default to 64MB floor on x86_64 systems (a bit better on arm where it tries to cap the automatic allocation at a maximum of 6% of total system RAM or a hard ceiling). I do feel that's low and will cause you to run out of swiotlb memory...

asherkariv: For this feature, we are using the kernel "swiotlb engine", and providing our own swiotlb pool, which the kernel uses for hv_pci devices. The hv_pci swiotlb pool parameters are specified via hv_pci_swiotlb=<base gpa>,<size>. 
Thus, we can leave the swiotlb default region to use the default size.

localization/strings/en-US/Resources.resw
asherkariv: Correct, the example uses the more common use of K,M, but we still do allow a lowercase postfix.

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -415,6 +415,31 @@ void WslCoreVm::Initialize(const GUID& VmId, const wil::shared_handle& UserToken
...
+    // Accept a connection from mini_init with a receive timeout so the service does not get stuck waiting for a response from the VM.
+    m_miniInitChannel =
+        wsl::shared::SocketChannel{AcceptConnection(m_vmConfig.KernelBootTimeout), "mini_init", {m_terminatingEvent.get()}};
+
+    // Accept the connection from the Linux guest for notifications.
+    m_notifyChannel = AcceptConnection(m_vmConfig.KernelBootTimeout);
+
+    // Receive and parse the guest kernel version
+    ReadGuestCapabilities();
+
```
ptrivedi, line 427: Moving this before drvfs initialization would affect our startup latency? Before, the drvfs initialization could have proceeded concurrently with the kernel boot

src/windows/common/helpers.cpp
```diff
@@ -772,4 +773,27 @@ void wsl::windows::common::helpers::AppendCommonKernelCommandLine(_Inout_ std::w
...
+    // Skip swiotlb on VMs that cannot fit the reserved buffer. Users can still opt in explicitly
+    // via experimental.swiotlb in .wslconfig.
+    if (memoryBytes < c_swiotlbBase + c_swiotlbSize)
+    {
+        return {};
+    }
+
+    // 1 GiB base sits below 4 GiB (32-bit PCI BAR limit on ARM64 virtio), below the Hyper-V
+    // low PCI MMIO hole (starts at >= 3 GiB on both x64 and ARM64), and above early boot allocations.
+    return L"0x40000000,64M";
```
ptrivedi: why do we return a hardcoded string here instead of using the c_swiotlbBase and c_swiotlbSize constants to construct it?

src/windows/common/DeviceHostProxy.cpp
```diff
@@ -133,6 +133,20 @@ void DeviceHostProxy::Shutdown()
 {
     {
         auto lock = m_lock.lock_exclusive();
+
+        for (auto& entry : m_fileSystems)
+        {
+            try
+            {
+                wil::com_ptr<IPlan9FileSystem> instance;
```
ptrivedi: Hmm, is this a fix for an existing issue? or needed for swiotlb buffer release or something?

src/shared/inc/lxinitshared.h
```diff
@@ -1430,9 +1430,10 @@ typedef struct _LX_INIT_GUEST_CAPABILITIES
 
     MESSAGE_HEADER Header;
     bool SeccompAvailable;
+    bool KernelSupportsHvPciSwiotlb;
```
ptrivedi: thinking out loud here. We shouldn't have to worry about backward compat when adding to this struct, because init and service will always be deployed together...?

asherkariv: This is for the case where the user uses a private kernel, that may or may not have the path.

OneBlue: Yes wslservice and `init` always come from the same build so this isn't an issue

src/windows/common/WslCoreConfig.cpp
```diff
@@ -124,7 +143,8 @@ void wsl::core::Config::ParseConfigFile(_In_opt_ LPCWSTR ConfigFilePath, _In_opt
         ConfigKey(ConfigSetting::Experimental::InitialAutoProxyTimeout, InitialAutoProxyTimeout),
         ConfigKey(ConfigSetting::Experimental::IgnoredPorts, std::move(parseIgnoredPorts)),
         ConfigKey(ConfigSetting::Experimental::HostAddressLoopback, EnableHostAddressLoopback),
-        ConfigKey(ConfigSetting::Experimental::SetVersionDebug, SetVersionDebug)};
+        ConfigKey(ConfigSetting::Experimental::SetVersionDebug, SetVersionDebug),
+        ConfigKey(ConfigSetting::Experimental::Swiotlb, std::move(parseSwiotlb))};
```
OneBlue, line 147: What's the thinking for exposing a new .wslconfig value for this ? Could we set a hardcoded value first ? That would greatly simplify this change

asherkariv:
Canceling this PR, as we have an updated PR using a different kernel patch, which lets the wsldevicehost know where the swiotlb region has been allocated. 
https://github.com/microsoft/WSL/pull/40654

Issue #40458, opened 2026-05-07
WSL exits unexpectedly when compiling software.


kphillisjr:
### Windows Version

Microsoft Windows [Version 10.0.26200.8246]

### WSL Version

2.6.3.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

Linux version 6.6.87.2-microsoft-standard-WSL2 (root@439a258ad544) (gcc (GCC) 11.2.0, GNU ld (GNU Binutils) 2.37) #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025

### Distro Version

Ubuntu 24.04.4 

### Other Software

Visual Studio 2022
Command Prompt
other wsl instance ( running top)

### Repro Steps

# Initial Configuration
1) Install Ubuntu using the Visual Studio option to "Install WSL For Me" under the Project Menu 
2) Open a command line after the Installation completes sucessfully.
3) sign into root on the VM using the command:
```
wsl -u root
```
4) Install the Dependencies to build the Vulkan SDK for linux ( Documentation Here: https://vulkan.lunarg.com/doc/sdk/latest/linux/getting_started.html )
```
apt update
apt upgrade
apt install xz-utils

apt-get install libglm-dev cmake libxcb-dri3-0 libxcb-present0 libpciaccess0 \
libpng-dev libxcb-keysyms1-dev libxcb-dri3-dev libx11-dev g++ gcc \
libwayland-dev libxrandr-dev libxcb-randr0-dev libxcb-ewmh-dev \
git python-is-python3 bison libx11-xcb-dev liblz4-dev libzstd-dev \
ocaml-core ninja-build pkg-config libxml2-dev wayland-protocols python3-jsonschema \
clang-format qtbase5-dev qt6-base-dev
```
5) Create the folder ``` /opt/vulkansdk/custom_built ```.
6) Change the group and file ownership of the folder created in step 5 to allow the non-root wsl user the ability to create/edit files  in this path.
7) exit out of this session.
8) Use Visual Studio to download the Repository https://github.com/kphillisjr/VkSDKBuilder 
9) Issue the Configuration, and build commands on visual studio for wsl linux. The first build I am doing uses the default configuration for ```linux-debug```.
10) use the open command prompt to launch wsl.
11) run the command ```top```

Note: Once the first crash happens it can be reproduced with the following sequence
1) open wsl on command line
```wsl```
[...]

PR #40519, opened 2026-05-13
Protect critical WSL processes under heavy load with cgroup & isolate distro cgroups

chemwolf6922:
### Issue 1:
When under heavy memory load, the critical WSL processes can fail due to various OOM failures. And may end up in an error state after the memory storm ended. Causing "Catastrophic Failures" afterwards.

### Issue 2:
All distro's systemd instances shared the same sets of cgroup. This will cause conflicts. For example, when booting multiple wsl distros at the same time. Only one can create the systemd user session successfully.
```
systemd[1023]: user@1000.service: Failed to attach to cgroup /user.slice/user-1000.slice/user@1000.service: Device or resource busy
```

### Changes
This PR reorganize the wsl cgroup with the follow structure:
```
# When systemd is enabled
/                                                  # critical wsl processes live in root with no resource limit.
    --wsl-user                                     # resource limited cgroup
        -- non-distro                              # for processes not belonging to a distro, like plugins
        -- distro-<init pid>                       # per distro root
            -- non-systemd                         # for non systemd user processes
            -- systemd                             # for systemd and it's cgroup tree

# When systemd is not enabled
/                                                  # critical wsl processes live in root with no resource limit.
    --wsl-user                                     # resource limited cgroup
        -- non-distro                              # for processes not belonging to a distro, like plugins
        -- distro-<init pid>                       # per distro cgroup
```
The wsl-user cgroup has memory.max and cpu.max set so it can only take max - 32MiB of RAM and max - 0.01 CPU cores. Effectively reserving 32MiB RAM and 0.01 cores for processes not under this cgroup.

### Note
Since this uses cgroupv2. Enabling this will not allow the use of cgroupv1. A .wslconfig option IsolateDistroCgroup (default true) is added so users can opt-out of this and keep using cgorupv1.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/40458 https://github.com/microsoft/WSL/issues/40593 https://github.com/microsoft/WSL/issues/40865

### Issue 1
Validated by the OP of this issue: https://github.com/microsoft/WSL/issues/40458#issuecomment-4434265543
The original issue is not deterministic and is hard to repro. Thus, no new test is added.

### Issue 2
The test_distro won't repro the issue. No new test is added.
Validated manually with launching Ubuntu and Debian at the same time.

### General
Add tests to validate the cgroup isolation works:
UnitTests::UnitTests::IsolatedCgroupLayout
UnitTests::UnitTests::IsolatedCgroupLayoutSystemd
UnitTests::UnitTests::IsolatedCgroupLayoutDisabled


chemwolf6922:
If we want this change. Please help determine the reserved memory size. 128M might be too much.

src/linux/init/init.cpp
```diff
@@ -1208,6 +1208,9 @@ try
 
         SessionLeader = UtilCreateChildProcess(
             "SessionLeader", [SessionLeaderFd = std::move(SessionLeaderFd), TtyFd = std::move(TtyFd), &Channel, &Config]() mutable {
+                // Move session leader into the memory-limited user cgroup.
+                WriteToFile(WSL_USER_CGROUP_PROCS, "0");
```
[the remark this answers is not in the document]
chemwolf6922: This should not be considered a critical error.

src/linux/init/util.cpp
```diff
@@ -3417,6 +3417,9 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
     auto ControlPipe = wil::unique_pipe::create(O_CLOEXEC);
 
     const int ChildPid = UtilCreateChildProcess("CreateChildProcess", [&]() {
+        // Move child into the memory-limited user cgroup.
+        WriteToFile(WSL_USER_CGROUP_PROCS, "0");
```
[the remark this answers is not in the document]
chemwolf6922: Should not be considered critical. But log is helpful.

src/linux/init/init.cpp
```diff
@@ -2391,6 +2397,13 @@ Return Value:
 
             CreateWslSystemdUnits(Config);
 
+            //
+            // Move systemd into the memory-limited user cgroup so that all
+            // services and user processes run under the memory cap.
+            //
+
+            WriteToFile(WSL_USER_CGROUP_PROCS, "0");
```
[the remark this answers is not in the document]
chemwolf6922: Non-critical

src/linux/init/config.cpp
```diff
@@ -992,6 +992,9 @@ try
     if (Config.BootCommand.has_value())
     {
         UtilCreateChildProcess("BootCommand", [Command = Config.BootCommand.value(), SavedSignals = g_SavedSignalActions]() {
+            // Move boot command into the memory-limited user cgroup.
+            WriteToFile(WSL_USER_CGROUP_PROCS, "0");
```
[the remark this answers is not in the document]
chemwolf6922: ?

src/linux/init/util.cpp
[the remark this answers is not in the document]
chemwolf6922: (Also wrong return value comparison.)

src/linux/init/main.cpp
```diff
@@ -3888,6 +3891,71 @@ void EnableDebugMode(const std::string& Mode)
...
+    if (UtilMkdir(WSL_USER_CGROUP_PATH, 0755) < 0)
+    {
+        LOG_ERROR("Failed to create wsl-user cgroup directory {}", errno);
+        return;
+    }
+
+    if (WriteToFile(CGROUP_MOUNTPOINT "/cgroup.subtree_control", "+memory") < 0)
+    {
+        LOG_ERROR("Failed to enable memory controller {}", errno);
+        return;
```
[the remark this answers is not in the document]
chemwolf6922: image
https://www.kernel.org/doc/Documentation/admin-guide/cgroup-v2.rst

src/linux/init/config.cpp
```diff
@@ -992,6 +992,13 @@ try
     if (Config.BootCommand.has_value())
     {
         UtilCreateChildProcess("BootCommand", [Command = Config.BootCommand.value(), SavedSignals = g_SavedSignalActions]() {
+            // Move boot command into the memory-limited user cgroup.
+            if (WriteToFile(WSL_USER_CGROUP_PROCS, "0") != 0)
+            {
+                // Non-critical.
+                LOG_ERROR("Failed to move boot command into user cgroup: {}", errno);
```
OneBlue: nit: I would have a preference towards making this critical, otherwise we could miss places where this assignment fails, which would bring the OOM issue

src/linux/init/main.cpp
```diff
@@ -4095,7 +4163,14 @@ int main(int Argc, char* Argv[])
         }
     }
 
-    UtilMount(nullptr, CGROUP_MOUNTPOINT, CGROUP2_DEVICE, 0, nullptr);
+    if (UtilMount(nullptr, CGROUP_MOUNTPOINT, CGROUP2_DEVICE, 0, nullptr) != 0)
+    {
+        LOG_ERROR("Failed to mount cgroup2 filesystem: {}", errno);
```
OneBlue: I recommend hard-failing if we ever hit this

src/linux/init/main.cpp
```diff
@@ -3888,6 +3891,71 @@ void EnableDebugMode(const std::string& Mode)
...
+    auto userMemoryMax = std::to_string(totalRam - c_systemReservedMemory);
+    if (WriteToFile(WSL_USER_CGROUP_MEMORY_MAX, userMemoryMax.c_str()) < 0)
+    {
+        LOG_ERROR("Failed to set memory.max for wsl-user cgroup {}", errno);
+        return;
+    }
+
+    LOG_INFO("WSL user cgroup created with memory.max={} (totalram={}, reserved={})", userMemoryMax, totalRam, c_systemReservedMemory);
+}
+CATCH_LOG()
```
OneBlue: I recommend failing if we ever throw an exception here

src/linux/init/main.cpp
```diff
@@ -103,6 +103,7 @@ Module Name:
 #define syscall_arch (offsetof(struct seccomp_data, arch))
 
 constexpr auto c_trueString = "1";
+constexpr size_t c_systemReservedMemory = 128 * 1024 * 1024; // 128MB reserved for WSL system processes
```
benhillis: I’m not super familiar with this setting, I assume this is just a minimum? How did you pick this value? Is there some other functionality to mark processes as critical another way to protect against cpu starvation as well?

chemwolf6922: Yes this acts as a reserved value for the critical wsl processes. Those processes memory usage can go over this value.

The value is arbitrarily picked. I need inputs from the team to pick a more based value. 128MiB is likely too large.

From my personal experience (limiting a lxc's cpu usage), using cgroup cpu limit is not a good idea. It will greatly hammer the actual cpu performance. I'll do some test in wsl to see if that's also true here.

chemwolf6922: For the magic number. I calculated the sum of PSS of all processes in the root cgroup namespace. And it's around **9068KB** with one distro (with wslg) and the debug shell open. 
```
PSS_KB  PID     NAME    CGROUP
499     1       mini_init       0::/
174     281     DebugShell      0::/
1269    284     login   0::/
638     289     chronyd 0::/
1066    296     GnsEngine       0::/
723     298     init    0::/
808     300     init()  0::/
973     306     init-systemd(Ub 0::/
430     307     init-watcher    0::/
759     351     init    0::/
1729    3836    bash    0::/
root@feng-workstation [ ~ ]#
```
And it increases to **9140KB** with two distro (with wslg) and the debug shell open. Not much increase since most of the pages are shared. 
```
PSS_KB  PID     NAME    CGROUP
349     1       mini_init       0::/
140     281     DebugShell      0::/
1193    284     login   0::/
585     289     chronyd 0::/
795     296     GnsEngine       0::/
547     298     init    0::/
530     300     init()  0::/
609     306     init-systemd(Ub 0::/
301     307     init-watcher    0::/
535     351     init    0::/
1686    3836    bash    0::/
530     6336    init()  0::/
619     6341    init-systemd(De 0::/
305     6342    init-watcher    0::/
416     6370    init    0::/
```
So 128MiB is likely too much. 16MiB is enough and maybe 24 or 32MiB for some head room.

One thing to note is that since we are not providing isolated cgroup views to the distros, any process can move them out of the wsl-user namespace. One example is the Windows defender for endpoint. It moves itself to the /mde namespace. And it takes ~200MB of RAM.

chemwolf6922: For the CPU performance impact. I did the following test on a nproc=32 machine:
**cpu.max = max 100000  (default)**
Geekbench: single: 2917, multi: 17046
Compiling the kernel -j33: 7m28.953s
**cpu.max = 3190000 100000 (reserve 0.1 cpu for critical processes)**
Geekbench: single: 2929, multi: 17050
Compiling the kernel -j33: 7m25.078s

My system is not clean so the difference is within margin of error. I'm not seeing the performance degradation I experienced before.

I'll update the cgroup setting to add the cpu reservation.

src/linux/init/main.cpp
```diff
@@ -4095,7 +4198,14 @@ int main(int Argc, char* Argv[])
         }
     }
 
-    UtilMount(nullptr, CGROUP_MOUNTPOINT, CGROUP2_DEVICE, 0, nullptr);
+    if (UtilMount(nullptr, CGROUP_MOUNTPOINT, CGROUP2_DEVICE, 0, nullptr) != 0)
```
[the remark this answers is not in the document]
chemwolf6922: ?

src/linux/init/main.cpp
```diff
@@ -3888,6 +3908,98 @@ void EnableDebugMode(const std::string& Mode)
...
+    {
+        LOG_WARNING("Total RAM ({}) is too small to reserve {} for system processes", totalRam, c_systemReservedMemory);
+        return;
+    }
+
+    if (UtilEnableAllCgroupControllers(CGROUP_MOUNTPOINT) < 0)
+    {
+        LOG_ERROR("Failed to enable cgroup controllers for root {}", errno);
+        return;
+    }
```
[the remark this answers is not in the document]
chemwolf6922: Not true for the root cgroup

test/windows/UnitTests.cpp
```diff
@@ -6979,5 +6982,105 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
     }
 
+    void ValidateIsolatedCgroupLayout(bool Systemd)
+    {
+        constexpr auto secondDistroName = L"cgroup-test-distro";
+
+        // Ensure no stale state from a previous run.
+        TerminateDistribution();
+        LxsstuLaunchWsl(std::format(L"--unregister {}", secondDistroName));
+
```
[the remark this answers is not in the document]
chemwolf6922: yes, it should be. The default value is from the implementation, not from the test.

src/linux/init/main.cpp
```diff
@@ -3227,7 +3252,14 @@ try
         return ProcessMountFolderMessage(Transaction, Buffer);
 
     case LxInitCreateProcess:
-        return ProcessCreateProcessMessage(Transaction, Buffer);
+        if (access(WSL_USER_NON_DISTRO_CGROUP_PATH, F_OK) == 0)
+        {
```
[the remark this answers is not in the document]
chemwolf6922: ?

src/linux/init/main.cpp
```diff
@@ -2227,6 +2242,9 @@ void ProcessLaunchInitMessage(
 
         THROW_LAST_ERROR_IF(MountDevice(Message->MountDeviceType, Message->DeviceId, DISTRO_PATH, FsType, Message->Flags, MountOptions) < 0);
 
+        auto MiniInitDirectChildPidPath = std::filesystem::read_symlink(PROCFS_PATH "/self");
+        pid_t MiniInitDirectChildPid = std::stoul(MiniInitDirectChildPidPath.string());
+
```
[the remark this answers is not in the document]
chemwolf6922: Valid issue. Wrong concern. They are in the same process tree. The actual issue is that they may race.

src/linux/init/config.cpp
```diff
@@ -991,17 +991,23 @@ try
 
     if (Config.BootCommand.has_value())
     {
-        UtilCreateChildProcess("BootCommand", [Command = Config.BootCommand.value(), SavedSignals = g_SavedSignalActions]() {
-            //
-            // Restore default signal dispositions for the child process.
-            //
+        UtilCreateChildProcess(
+            "BootCommand", [Command = Config.BootCommand.value(), SavedSignals = g_SavedSignalActions, CgroupPathOpt = Config.CgroupPath]() {
+                if (CgroupPathOpt.has_value())
```
OneBlue: nit: I would recommend adding the cgroup as an optional argument to `UtilCreateChildProcess()` so this can be factored

src/linux/init/init.cpp
```diff
@@ -2299,6 +2309,20 @@ Return Value:
...
+    // Get the per-distro cgroup path.
+    //
+
+    const auto DistroCgroupPath = getenv(LX_WSL2_DISTRO_CGROUP_PATH);
+    if (DistroCgroupPath != nullptr)
+    {
+        if (access(DistroCgroupPath, F_OK) == 0)
+        {
+            Config.CgroupPath = DistroCgroupPath;
+        }
```
OneBlue, line 2318: we should log if `access` fails here

src/linux/init/main.cpp
```diff
@@ -102,6 +102,10 @@ Module Name:
 #define syscall_arch (offsetof(struct seccomp_data, arch))
 
 constexpr auto c_trueString = "1";
+constexpr size_t c_systemReservedMemory = 32 * 1024 * 1024; // 32MiB reserved for WSL system processes
+constexpr long c_cpuPeriodMicros = 100000;
+constexpr long c_systemReservedCpuMicros = 5000; // 0.05 CPU reserved for WSL system processes
```
OneBlue: Would this essentially "block" 5% of a CPU for WSL processes ? If so, that feels like quite a bit. I'd recommend dropping to 0.01, given that WSL processes are mostly IO bound and not CPU bound

src/linux/init/main.cpp
```diff
@@ -3657,6 +3732,104 @@ void EnableDebugMode(const std::string& Mode)
...
+        return;
+    }
+
+    if (UtilEnableAllCgroupControllers(CGROUP_MOUNTPOINT) < 0)
+    {
+        LOG_ERROR("Failed to enable cgroup controllers for root {}", errno);
+        return;
+    }
+
+    if (UtilMkdir(WSL_USER_CGROUP_PATH, 0755) < 0 && errno != EEXIST)
```
OneBlue: UtilMkdir() already handles EEXIST, so it doesn't need to be special cased (applies in other places in this change)

src/linux/init/main.cpp
```diff
@@ -3971,6 +4144,102 @@ int main(int Argc, char* Argv[])
 
                     sync();
 
+                    //
+                    // Clear the distro cgroup
+                    //
+
+                    auto CgroupDir = UtilGetDistroCgroupPath(Result);
```
OneBlue, line 4146: What's the reason for manually killing processes inside the cgroup ? 

Wouldn't `init` exiting automatically do that ?

src/linux/init/util.cpp
```diff
@@ -3726,3 +3730,56 @@ try
...
+
+int UtilEnableAllCgroupControllers(const std::string& CgroupPath)
+{
+    // Only cpu and memory are required for wsl's resource limit.
+    if (WriteToFile((CgroupPath + "/cgroup.subtree_control").c_str(), "+cpu +memory") < 0)
+    {
+        LOG_ERROR("Failed to enable cgroup controllers for {}: {}", CgroupPath, errno);
+        return -1;
+    }
+    const char* const OptionalControllers[] = {"+pids", "+io", "+cpuset", "+hugetlb", "+rdma", "+misc"};
```
OneBlue, line 3760: What's the reason for enabling those if we don't need them ?

src/linux/init/init.cpp
```diff
@@ -1180,6 +1180,11 @@ try
 
         SessionLeader = UtilCreateChildProcess(
             "SessionLeader", [SessionLeaderFd = std::move(SessionLeaderFd), TtyFd = std::move(TtyFd), &Channel, &Config]() mutable {
+                if (Config.CgroupPath.has_value())
+                {
+                    UtilTryMoveSelfToDistroCgroup(Config.CgroupPath.value(), false, "session leader");
```
OneBlue: Wouldn't the child inherit its parent cgroup here ?

test/windows/UnitTests.cpp
```diff
@@ -7314,5 +7317,105 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
+        // Import the second distro.
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(std::format(L"--import {} . \"{}\" --version 2", secondDistroName, g_testDistroPath)), 0L);
+
+        std::optional<decltype(EnableSystemd())> systemdCleanup;
+        if (systemd)
+        {
+            systemdCleanup.emplace(EnableSystemd());
+
+            LxssWriteWslDistroConfig("[boot]\nsystemd=true\n", secondDistroName);
```
OneBlue: nit: i'd recommend adding an optional distro argument to `EnableSystemd` so we don't need to duplicate the logic here

test/windows/UnitTests.cpp
```diff
@@ -7314,5 +7317,105 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
+            LxsstuLaunchWsl(std::format(L"--terminate {}", secondDistroName));
+        }
+
+        auto getCgroup = [](LPCWSTR distro) {
+            auto [out, _] = LxsstuLaunchWslAndCaptureOutput(std::format(L"-d {} --cd / -- cat /proc/self/cgroup", distro));
```
OneBlue: I don't think we need `--cd /` here

test/windows/UnitTests.cpp
```diff
@@ -7314,5 +7317,105 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
+            while (!out.empty() && (out.back() == L'\n' || out.back() == L'\r'))
```
OneBlue: I recommend using `cat -e` so we don't have to deal with the '\n' ourselves

src/linux/init/main.cpp
```diff
@@ -2227,6 +2242,51 @@ void ProcessLaunchInitMessage(
...
 
+        auto MiniInitDirectChildPidPath = std::filesystem::read_symlink(PROCFS_PATH "/self");
+        pid_t MiniInitDirectChildPid = std::stoul(MiniInitDirectChildPidPath.string());
+
+        //
+        // Set up the per-distro cgroup before potentially forking into two inits.
+        //
+
+        std::string DistroCgroupPath{};
+        if (access(WSL_USER_CGROUP_PATH, F_OK) == 0)
```
OneBlue, line 2263: Could we factor this check with the below guiApplications check ? Otherwise we're parsing wsl.conf twice

src/linux/init/main.cpp
```diff
@@ -3971,6 +4144,102 @@ int main(int Argc, char* Argv[])
...
+                        CATCH_LOG();
+
+                        //
+                        // Recursively rmdir the cgroup subtree.
+                        //
+
+                        try
+                        {
+                            std::vector<std::string> dirs;
+                            for (const auto& entry : std::filesystem::recursive_directory_iterator(
```
OneBlue, line 4158: Is there a potential race with the distribution being restarted while we're removing its cgroup ?

src/linux/init/main.cpp
chemwolf6922: This is 5% of one logical core. Drop to 1% should be enough though.

src/linux/init/util.cpp
chemwolf6922: This is for any program in the subgroups to be able to operate those. For example, systemd.

src/linux/init/main.cpp
chemwolf6922: The per distro cgroup's name is derived from the root pid of it's direct child of mini_init. If the pid is reused as the next distro's init. This could cause a race condition. 
However, since the distro creation and the waitpid + cleanup happens in the same loop. The pid cannot be reused to create a distro before or during the waitpid + cleanup process. So, there should not be a race window.

src/linux/init/main.cpp
chemwolf6922: Made this critical. Please note this error was ignored before. And I don't have much context on why.

src/linux/init/init.cpp
chemwolf6922: Yes, but the parent here is the distro init. Which is a critical wsl process and is in the root cgroup. So we need to move the session leader to the per-distro cgroup.

src/linux/init/config.cpp
chemwolf6922: IMHO, fail early here will increase the total failure rate. Since we will be increasing from P(move failure) * P(oom) to P(move failure).

pravic:
Hello, any updates on this? What stops from merging and how to unblock?

test/windows/UnitTests.cpp
```diff
@@ -7513,5 +7516,100 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
+        // Re-start the default test_distro and confirm that exactly one distro-<pid> cgroup remains:
+        // the one belonging to the distro we just started to perform the check.  The stale cgroups of
+        // the two terminated distros must have been removed.
+        //
+        // N.B. Mini_init cleans up per-distro cgroups asynchronously from its SIGCHLD reaper after
+        // wsl --terminate returns (it sync()s, then recursively rmdir's the distro cgroup subtree).

+        // On slower hosts (e.g. CI pipelines) the cleanup of the two terminated distros

+        // can still be in flight when this check runs, so retry until cleanup completes.
```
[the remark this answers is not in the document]
chemwolf6922: ? I think you introduced this one...

src/linux/init/main.cpp
```diff
@@ -3971,6 +4139,47 @@ int main(int Argc, char* Argv[])
...
+                    if (access(CgroupDir.c_str(), F_OK) == 0)
+                    {
+                        LOG_INFO("Process {} exited, removing cgroup {}", Result, CgroupDir);
+
+                        //
+                        // Recursively rmdir the cgroup subtree.
+                        //
+
+                        try
+                        {
```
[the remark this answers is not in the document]
chemwolf6922: As discussed above, we trust the init process for this.

test/windows/Common.cpp
```diff
@@ -1460,9 +1460,9 @@ std::wstring LxssWriteWslConfig(const std::wstring& Content)
 }
 
 // writes distro specific settings /etc/wsl.conf
-std::string LxssWriteWslDistroConfig(const std::string& Content)
+std::string LxssWriteWslDistroConfig(const std::string& Content, LPCWSTR DistributionName)
 {
-    std::string path = std::format("\\\\wsl.localhost\\{}\\etc\\wsl.conf", LXSS_DISTRO_NAME_TEST);
+    std::string path = std::format("\\\\wsl.localhost\\{}\\etc\\wsl.conf", wsl::shared::string::WideToMultiByte(DistributionName));
```
OneBlue, line 1465: nit: `WideToMultiByte` is not needed here

test/windows/UnitTests.cpp
```diff
@@ -6768,6 +6768,9 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
 
     WSL2_TEST_METHOD(CGroupv1)
     {
+        // cgroupv1 conflicts with the per-distro cgroup hierarchy
```
OneBlue, line 6771: If that's the case, I recommend adding a warning when we start a distribution with cgroupv1 enabled, so others know which setting to disable at the .wslconfig level

OneBlue (approved):
LGTM, couple minor comments that can be addressed in a followup. 

Sorry for the delay in review !

test/windows/Common.cpp
chemwolf6922: Will resolve in a follow up.

test/windows/UnitTests.cpp
chemwolf6922: Will resolve in a follow up.

linghengqian (commented):
Although the current PR corresponds to commits tagged with 2.9.5, 2.9.6, and 2.9.7, the latest version released on GitHub is still 2.9.4? Am I missing an issue?

DimNS:
> Although the current PR corresponds to commits tagged with 2.9.5, 2.9.6, and 2.9.7, the latest version released on GitHub is still 2.9.4? Am I missing an issue?

I confirm that the tags are already there, but the release descriptions are not yet available.

PR #40520, opened 2026-05-13
ci: declare GITHUB_TOKEN scope explicitly for three workflows

arpitjain099:
Three workflows in this repo currently leave `GITHUB_TOKEN` scope implicit (no top-level `permissions:` block), so the token inherits whatever the repository default is. This patch declares the minimum scope each workflow actually needs:

- **`distributions.yml`** and **`modern-distributions.yml`** — both run on `pull_request` against `distributions/**` and only check out the repo + run a Python validator. `contents: read` is all they need.
- **`winget.yml`** — runs on `release: released` and forwards `*.msi` asset URLs to `wingetcreate` using the `WINGET_TOKEN` secret. It does not check out the repo and does not call any GitHub API endpoint, so `permissions: {}` (no scopes at all) is correct here.

The change is mechanical — no job steps, no triggers, no conditions move. Files were validated with `python -c "import yaml; yaml.safe_load(...)"` before commit.


arpitjain099:
Hi @benhillis, gentle ping on this. PR has been open for 4 days without review. I noticed you've been on the recent-merger side of recent merges in this repo. When you have a moment, would you mind giving it a quick look? No urgency. Happy to address any feedback.

arpitjain099:
Hey, Bringing this back to the top of the queue. Happy to fix conflicts or adjust the approach. Thanks for your time.

OneBlue (approved):

PR #40521, opened 2026-05-13
Implement the WSLC plugin API

OneBlue:

This change implements the WSLC plugin API.

Notable design changes: 

- All plugins are only loaded once, by wslservice.exe
- Plugins remain loaded regardless of the WSL policies registry keys
- User errors originated from plugins clearly say which plugin the error comes from

## PR Checklist


src/windows/service/exe/PluginManager.h
```diff
@@ -37,12 +37,23 @@ class PluginManager
     PluginManager& operator=(PluginManager&&) = delete;
 
     void LoadPlugins();
+    void UnloadPlugins();
```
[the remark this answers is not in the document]
benhillis: This is accruate.

src/windows/service/inc/wslc.idl
```diff
@@ -803,20 +825,21 @@ interface IWSLCSessionFactory : IUnknown
     HRESULT CreateSession(
         [in] const WSLCSessionInitSettings* Settings,
         [in] IWSLCVirtualMachine* Vm,
+        [in, unique] IWSLCPluginNotifier* PluginNotifier,
```
benhillis: Is PluginNotifier ever NULL? If so I think we are missing some null checks.

src/windows/service/inc/wslc.idl
```diff
@@ -746,10 +767,11 @@ interface IWSLCSession : IUnknown
     // Returns a handle to this COM server process (used to add to job object).
     HRESULT GetProcessHandle([out, system_handle(sh_process)] HANDLE* ProcessHandle);
 
-    // Initializes the session with a pre-created VM.
+    // Initializes the session with a pre-created VM?
     HRESULT Initialize(
         [in] const WSLCSessionInitSettings* Settings,
-        [in] IWSLCVirtualMachine* Vm);
+        [in] IWSLCVirtualMachine* Vm,
+        [in] IWSLCPluginNotifier* PluginNotifier);
```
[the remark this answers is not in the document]
OneBlue: That's ok, this is an internal interface

src/windows/service/inc/wslc.idl
```diff
@@ -746,10 +767,11 @@ interface IWSLCSession : IUnknown
     // Returns a handle to this COM server process (used to add to job object).
     HRESULT GetProcessHandle([out, system_handle(sh_process)] HANDLE* ProcessHandle);
 
-    // Initializes the session with a pre-created VM.
+    // Initializes the session with a pre-created VM?
```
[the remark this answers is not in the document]
OneBlue: That's ok, this is an internal interface

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -881,6 +907,16 @@ __requires_exclusive_lock_held(m_lock) unique_com_disconnect WSLCContainerImpl::
 {
     unique_com_disconnect comWrapper;
 
+    // Notify plugin manager that the container is stopping. Errors are ignored.
+    if (m_state == WslcContainerStateRunning)
+    {
+        try
+        {
+            LOG_IF_FAILED(m_pluginNotifier->OnContainerStopping(m_id.c_str()));
+        }
```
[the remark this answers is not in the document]
OneBlue: m_pluginNotifier cannot be null

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -626,6 +627,20 @@ void WSLCSession::StreamImageOperation(DockerHTTPClient::HTTPRequestContext& req
...
+{
+    LOG_IF_FAILED(m_pluginNotifier->OnImageCreated(InspectImageLockHeld(ImageNameOrId).c_str()));
+}
+CATCH_LOG()
+
+void WSLCSession::OnImageDeleted(const std::string& ImageId) noexcept
+try
+{
+    LOG_IF_FAILED(m_pluginNotifier->OnImageDeleted(ImageId.c_str()));
+}
```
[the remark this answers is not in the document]
OneBlue: m_pluginNotifier cannot be null

src/windows/service/inc/wslc.idl
OneBlue: It should never be null. Will update

benhillis (commented):
🤖 **AI-generated review** (GitHub Copilot CLI, Claude Opus 4.7, run by @benhillis).

This is automated feedback — please use your judgment on every item. I focused on correctness / lifetime / API contract issues and skipped style. Three items from a prior pass were already addressed in `581b849e` / `af4affd8` (dead `UnloadPlugins`, nullable notifier contract, path-traversal in `WSLCMountFolder`) — thanks for those. Remaining items are inline.

Separately: I noted in another context that this design is compatible with #40120 (out-of-process plugin host); the lock-held-across-callbacks point below is the one place where the two PRs would interact and would benefit from a fix in-tree regardless of #40120.

src/windows/wslcsession/WSLCContainer.h
```diff
@@ -197,6 +202,7 @@ class WSLCContainerImpl
     std::uint64_t m_createdAt{};
     WSLCContainerState m_state = WslcContainerStateInvalid;
     WSLCSession& m_wslcSession;
+    IWSLCPluginNotifier* m_pluginNotifier;
```
benhillis, line 205: 🤖 **AI feedback (Copilot CLI)**: `m_pluginNotifier` is stored as a raw `IWSLCPluginNotifier*` while `WSLCSession::m_pluginNotifier` is a `wil::com_ptr`. The container's raw pointer aliases the session's `com_ptr` and only stays valid as long as container lifetime ⊆ session lifetime. Anywhere a `unique_ptr<WSLCContainerImpl>` could outlive the parent session — or any destruction ordering during shutdown that releases the session's `com_ptr` before its containers — would yield a UAF on the next `OnContainerStopping` call. Suggest making this a `wil::com_ptr<IWSLCPluginNotifier>` (one extra AddRef per container, negligible).

src/windows/service/exe/PluginManager.cpp
```diff
@@ -99,7 +100,256 @@ try
...
+
+    WSL_LOG(
+        "WslcPluginProcessGetFd",
+        TraceLoggingValue(static_cast<int>(Fd), "Fd"),
+        TraceLoggingValue(handle.Handle.Socket, "Handle"),
+        TraceLoggingValue(result, "Result"));
+
+    RETURN_IF_FAILED(result);
+
+    *Handle = handle.Handle.Socket;
```
benhillis, line 290: 🤖 **AI feedback (Copilot CLI)**: `WSLCHandle` is a tagged union (`switch_type(WSLCHandleType)`) over `File`/`Pipe`/`Socket`, but this always reads `.Handle.Socket` regardless of `handle.Type`. The public API doc for `WSLCPluginAPI_ProcessGetFd` promises *"the caller takes ownership and must close it with `CloseHandle()`"*, but a `WSLCHandleTypeSocket` value must be closed with `closesocket()` to release Winsock state correctly. Two options: (a) switch on `handle.Type` and either return the matching union member alongside its type or convert sockets to plain Win32 handles before exposing to plugins; (b) document that the returned HANDLE may be a SOCKET and require `closesocket` for those cases (this would contradict the current comment). As written the type tag is silently dropped.

src/windows/service/exe/PluginManager.cpp
```diff
@@ -322,3 +572,134 @@ void PluginManager::ThrowIfFatalPluginError() const
...
+            WSL_LOG(
+                "PluginOnWslcContainerStartedCall",
+                TraceLoggingValue(e.name.c_str(), "Plugin"),
+                TraceLoggingValue(Session->SessionId, "SessionId"),
+                TraceLoggingValue(result, "Result"));
+
+            if (FAILED(result))
+            {
+                LOG_HR_MSG(result, "Plugin '%ls' rejected container creation", e.name.c_str());
+                THROW_HR_WITH_USER_ERROR(result, wsl::shared::Localization::MessageFatalPluginError(e.name));
```
benhillis: 🤖 **AI feedback (Copilot CLI)**: `OnWslcSessionCreated` uses `ThrowIfPluginError(...)` (which records `m_pluginError` so subsequent calls short-circuit via `ThrowIfFatalPluginError()`); `OnWslcContainerStarted` throws via `THROW_HR_WITH_USER_ERROR` directly and bypasses that bookkeeping. The two veto-style hooks produce different observable behavior for plugin "fatal" errors. Either unify on `ThrowIfPluginError` here, or add a comment explaining why container rejection is intentionally non-sticky.

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -123,9 +137,31 @@ WSLCSessionManagerImpl::~WSLCSessionManagerImpl()
     std::lock_guard lock(m_wslcSessionsLock);
     for (auto& entry : m_sessions)
     {
+        NotifySessionStoppingLockHeld(entry);
         LOG_IF_FAILED(entry.Ref->Terminate());
     }
+
+    g_managerInstance.store(nullptr);
```
benhillis: 🤖 **AI feedback (Copilot CLI)**: Publication race on `g_managerInstance`:

- **Destructor**: `g_managerInstance.store(nullptr)` happens *after* iterating `m_sessions` and firing `NotifySessionStoppingLockHeld(entry)` for each (line 140). During those notifications, plugin callbacks see `Instance() != nullptr` and operate on a partially-destroyed object — specifically, `m_sessions` entries that have already had `Ref->Terminate()` called on them in earlier loop iterations.
- **Constructor** (line 128): `store(this)` happens before member init has fully completed; a concurrent plugin call into `WSLCMountFolder` → `Instance()` could see a half-built object.

Suggest: clear `g_managerInstance` *before* the destructor loop; publish `this` in the constructor *after* member init (or via a separate post-construction step / flag).

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -231,6 +284,33 @@ void WSLCSessionManagerImpl::CreateSession(const WSLCSessionSettings* Settings,
...
+        try
+        {
+            auto& entry = m_sessions.back();
+            WSLCSessionInformation info{};
+            info.SessionId = static_cast<WSLCSessionId>(entry.SessionId);
+            info.DisplayName = entry.DisplayName.c_str();
+            info.ApplicationPid = entry.CreatorPid;
+            info.UserToken = entry.UserToken.get();
+            info.UserSid = entry.UserSid.data();
+            g_pluginManager.OnWslcSessionCreated(&info);
```
benhillis, line 297: 🤖 **AI feedback (Copilot CLI)**: `OnWslcSessionCreated` fires while `m_wslcSessionsLock` (a `std::recursive_mutex`) is held. The plugin callback chain `WSLCMountFolder` → `ResolveWslcSession` → `Instance()->FindSession` re-enters the same lock; works today only because the callback runs on the *same* thread that holds the lock (recursive_mutex same-thread reentry).

Two consequences:
1. Concurrency — any other `CreateSession` / `OpenSession` / `ListSessions` is blocked for the duration of plugin work (which may include IO, network calls, spawning processes).
2. Out-of-process plugin host (per #40120): callbacks arrive on a *different* COM thread, so recursive_mutex's same-thread-reentry property no longer applies → hard deadlock if a plugin's `OnSessionCreated` calls `WSLCMountFolder`.

Suggested fix is the same `shared_mutex` pattern #40120 uses for VM hooks: a separate callback lock taken in shared mode while notifying, exclusive only for destructive ops. Worth fixing in-tree even independent of #40120 for reason (1).

src/windows/service/exe/PluginManager.cpp
```diff
@@ -99,7 +100,256 @@ try
...
+
+} // namespace
+
+extern "C" {
+
+HRESULT WSLCMountFolder(WSLCSessionId Session, LPCWSTR WindowsPath, BOOL ReadOnly, LPCWSTR Name, LPSTR Mountpoint)
+try
+{
+    RETURN_HR_IF(E_POINTER, WindowsPath == nullptr || Name == nullptr || Mountpoint == nullptr);
+    RETURN_HR_IF_MSG(E_INVALIDARG, StrStrW(Name, L"..") != nullptr, "Name cannot contain '..': %ls", WindowsPath);
```
benhillis: 🤖 **AI feedback (Copilot CLI)**: Nice fix for the obvious traversal vector. Two small follow-ups:

1. `StrStrW` is a substring match, so legitimate names containing `..` (e.g., `linux..test`) get rejected. A tokenized check (split `Name` on `/`/`\` and reject any segment that is exactly `..`) would be more precise.
2. The error message string is `"Name cannot contain '..': %ls"` but the `%ls` argument is `WindowsPath`, not `Name`. Probably want to log `Name` here for diagnosability.

src/windows/wslcsession/WSLCContainer.h
OneBlue: The session cannot be released before WSLCContainerImpl. All WSLCContainerImpl are cleared before the session terminates

src/windows/service/exe/PluginManager.cpp
OneBlue: Added an assert that the handle is indeed a socket. Regarding closing, this is the behavior that we have in the existing plugin API: Sockets are returned as handles, and closed by CloseHandle()

benhillis (dismissed):

benhillis (approved):

src/windows/service/exe/PluginManager.cpp
```diff
@@ -99,7 +100,264 @@ try
...
+        E_INVALIDARG,
+        nameLength == 0 ||
+            !std::ranges::all_of(Name, Name + nameLength, [&](auto c) { return c == '-' || c == '_' || iswalnum(c); }),
+        "Invalid mount name: %ls",
+        Name);
+
+    auto session = ResolveWslcSession(Session);
+
+    // Mount the folder under /mnt/wsl-plugin/<Name>. Convert Name to UTF-8 for the Linux path.
+    const auto linuxPath = std::format("/mnt/wsl-plugin/{}", Name);
```
atbagga, line 139: We allowed to mount on any path earlier like /etc/opt/microsoft/mdatp for binaries or /var/log/microsoft/mdatp for logs folder. 

Now it is restricted to specific prefix. Can we remove this restriction?

PR #40527, opened 2026-05-13
CLI: Fix parser treating stdin positional ('-') as an argument specifier error.

dkbennett:
The parser was treating '-' as an invalid argument specifier when it is a valid positional argument value when used to specify 'stdin'. This fixes it by checking first if there is a positional argument available, and if so then we treat it as the value and only consider it an argument error if there are no positionals remaining.  The assert in positional processing was also updated to allow a single '-' as input.

Parser test cases were added to verify this and the generic parser test cases test case was renamed for clarity.

## PR Checklist

## Validation Steps Performed
Manually verified argument is valid

Ran updated parser test cases and confirmed the - argument positional cases pass.

### DIFF
--- a/src/windows/wslc/arguments/ArgumentParser.cpp
+++ b/src/windows/wslc/arguments/ArgumentParser.cpp
@@ -67,31 +67,24 @@ void ParseArgumentsStateMachine::ThrowIfError() const
     }
 }
 
+void ParseArgumentsStateMachine::AdvanceToNextPositional(std::vector<Argument>::iterator& itr) const
+{
+    while (itr != m_positionalArgs.end() && (m_executionArgs.Count(itr->Type()) == itr->Limit()))
+    {
+        ++itr;
+    }
+}
+
 const Argument* ParseArgumentsStateMachine::NextPositional()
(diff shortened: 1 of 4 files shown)


src/windows/wslc/arguments/ArgumentParser.cpp
```diff
@@ -84,6 +84,17 @@ const Argument* ParseArgumentsStateMachine::NextPositional()
     return &*m_positionalSearchItr;
 }
 
+bool ParseArgumentsStateMachine::HasNextPositional() const
```
JohnMcPMS, line 84: Could refactor to share the advancement code into a `GetNextPositionalIterator` that this and `NextPositional` use so as to not duplicate the code.

JohnMcPMS (dismissed):

src/windows/wslc/arguments/ArgumentParser.cpp
dkbennett: Great suggestion, implemented!

JohnMcPMS (approved):

PR #40530, opened 2026-05-13
Made wslc build output more detailed

craigloewen-msft:

Changed `wslc build` to give more detailed build output. 

image

## PR Checklist

- [X] **Closes:**
- [X] **Dev docs:** Added/updated if needed
- [X] **Documentation updated:** If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/wsl/) and link it here: #xxx

Summary of changes:

- Change WSLCSession to send over the completed JSON bits
- BuildImageCallback then takes in these bits, and parses them with a 'BuildView' class
- This 'BuildView' object stores everything needed to do the rendering
- BuildImageCallback then has a thread that renders every few ms to the console
- Rendering is done by just appending the output, but once we hit the max viewport height we switch to a 'fixed view' mode which renders just to the screen size. 
- On destruction of that BuildImageCallback we print out the last result and any errors.

## Validation Steps Performed

Tested with various Containerfiles some simple and some complex. Here is a sufficiently complex one that you can use to test yourself:

<details>
<summary>View file</summary>
This containerfile does multi target builds that are copying from eachother and then finally gives an error right at the end. 

```
FROM alpine:latest AS builder
RUN apk add --no-cache gcc musl-dev
COPY hello.c /src/hello.c
RUN gcc -O2 -static -o /src/hello /src/hello.c && strip /src/hello


FROM python:3.12 AS assets
RUN pip install --no-cache-dir jinja2 markdown numpy matplotlib
COPY generate_page.py /work/generate_page.py
COPY template.html /work/template.html
COPY content.md /work/content.md
RUN python /work/generate_page.py

FROM python:3.13 AS systools
RUN apt-get update && apt-get install -y --no-install-recommends \
    curl ca-certificates jq file \
    && rm -rf /var/lib/apt/lists/*
RUN echo "System tools layer ready"

FROM alpine:latest AS final
RUN apk add --no-cache tini libstdc++

# Copy the compiled binary from builder
COPY --from=builder /src/hello /usr/local/bin/hello

# Copy generated HTML from assets stage
COPY --from=assets /work/output/index.html /var/www/index.html

# Copy a utility from systools
COPY --from=systools /usr/bin/jq /usr/local/bin/jq

RUN echho 123

# Add a health-check script
RUN echo '#!/bin/sh' > /healthcheck.sh \
    && echo '/usr/local/bin/hello && echo "OK"' >> /healthcheck.sh \
    && chmod +x /healthcheck.sh


EXPOSE 8080
ENTRYPOINT ["/sbin/tini", "--"]
CMD ["/usr/local/bin/hello"]
```

</details>

I notice that we don't get all the data coming in - so we don't get signals of every layer extraction completion. I've left the PR as is which just does best effort on showing it.
I also didn't program in support for changing the Terminal Window while it's running, which I think is OK.

### DIFF
--- a/src/windows/wslc/services/BuildImageCallback.cpp
+++ b/src/windows/wslc/services/BuildImageCallback.cpp
@@ -14,7 +14,6 @@ Abstract:
 
 #include "precomp.h"
 #include "BuildImageCallback.h"
-#include <docker_schema.h>
 
 namespace wsl::windows::wslc::services {
 
@@ -197,31 +196,44 @@ try
     }
 
     std::string statusStr(status);
-    std::string idStr(id ? id : "");
 
-    // BuildKit JSON messages are identified by the "buildkit" id
-    if (idStr == "buildkit")
+    try
     {
-        StartTimerThread();
-
         std::lock_guard lock(m_mutex);
+        m_buildView.ProcessMessage(statusStr);
 
-        try
+        if (!m_isConsole)
         {
-            auto json = nlohmann::json::parse(statusStr);
-            wsl::windows::common::docker_schema::BuildKitSolveStatus solveStatus{};
-            from_json(json, solveStatus);
-            m_buildView.ProcessMessage(solveStatus);
+            PrintPlainTextProgress();
         }
-        catch (...)
+        else 
         {
-            // Ignore malformed JSON
+            StartTimerThread();
         }
+    } 
+    catch (...)
+    {
+        LOG_CAUGHT_EXCEPTION();
     }
 
     return S_OK;
 }
 CATCH_RETURN();
+
+// ── Plain-text progress (non-console) ───────────────────────────────
+
+void BuildImageCallback::PrintPlainTextProgress()
+{
+    for (const auto* step : m_buildView.GetUnreportedSteps())
+    {
+        wprintf(L"%hs\n", step->stepLabel.c_str());
+
+        if (!step->error.empty())
+        {
+            wprintf(L"%hs\n", step->error.c_str());
+        }
+    }
+}
 
 // ── Step line formatter ─────────────────────────────────────────────
 
@@ -280,6 +292,51 @@ std::wstring BuildImageCallback::FormatStepLine(const ViewStep& step, SHORT cons
     }
 }
 
+// Synthesize cached FROM lines for base image steps that were skipped
+// because they were resolved from a shared stage.
+void BuildImageCallback::AppendCachedFromLines(
+    const ViewTarget& target, SHORT consoleWidth, std::vector<std::wstring>& lines) const
+{
+    if (target.steps.empty())
+    {
+        return;
+    }
+
+    const auto& firstStep = target.steps[0];
+    if (firstStep.stepNum <= 1 || firstStep.stepTotal == 0)
+    {
+        return;
+    }
+
+    for (uint32_t n = 1; n < firstStep.stepNum; n++)
+    {
+        std::string fromImage = " ...";
+        for (const auto& digest : firstStep.inputs)
+        {
+            auto* inputStep = m_buildView.StepByDigest(digest);
+            if (inputStep && inputStep->stepLabel.find("] FROM ") != std::string::npos)
+            {
+                auto pos = inputStep->stepLabel.find("] FROM ");
+                if (pos != std::string::npos)
+                {
+                    fromImage = inputStep->stepLabel.substr(pos + 6);
+                }
+                break;
+            }
+        }
+        auto label = std::format("[{} {}/{}] FROM{}", target.name, n, firstStep.stepTotal, fromImage);
+        auto wlabel = MultiByteToWide(label);
+        if (static_cast<SHORT>(wlabel.size()) > consoleWidth - 10)
+        {
+            wlabel.resize(consoleWidth - 13);
+            wlabel += L"...";
+        }
+        auto padLen =
+            (wlabel.size() + 6 < static_cast<size_t>(consoleWidth)) ? static_cast<size_t>(consoleWidth) - wlabel.size() - 6 : 0;
+        lines.push_back(std::format(L"\033[36m{}{}{}\033[0m", wlabel, std::wstring(std::max<size_t>(1, padLen), L' '), L"CACHED"));
+    }
+}
+
 // ── Frame line builder ──────────────────────────────────────────────
 // Builds all frame lines into a vector. Each entry is one visual line
 // (no embedded newlines). Used by both RenderFrame and the destructor's
@@ -324,42 +381,7 @@ std::vector<std::wstring> BuildImageCallback::BuildFrameLines(SHORT consoleWidth
         firstTarget = false;
 
         // Synthesize cached FROM steps for shared base images
-        if (!target.steps.empty())
-        {
-            const auto& firstStep = target.steps[0];
-            if (firstStep.stepNum > 1 && firstStep.stepTotal > 0)
-            {
-                for (uint32_t n = 1; n < firstStep.stepNum; n++)
-                {
-                    std::string fromImage = " ...";
-                    for (const auto& digest : firstStep.inputs)
-                    {
-                        auto* inputStep = m_buildView.StepByDigest(digest);
-                        if (inputStep && inputStep->stepLabel.find("] FROM ") != std::string::npos)
-                        {
-                            auto pos = inputStep->stepLabel.find("] FROM ");
-                            if (pos != std::string::npos)
-                            {
-                                fromImage = inputStep->stepLabel.substr(pos + 6);
-                            }
-                            break;
-                        }
-                    }
-                    auto label = std::format("[{} {}/{}] FROM{}", target.name, n, firstStep.stepTotal, fromImage);
-                    auto wlabel = MultiByteToWide(label);
-                    if (static_cast<SHORT>(wlabel.size()) > consoleWidth - 10)
-                    {
-                        wlabel.resize(consoleWidth - 13);
-                        wlabel += L"...";
-                    }
-                    auto padLen = (wlabel.size() + 6 < static_cast<size_t>(consoleWidth))
-                                      ? static_cast<size_t>(consoleWidth) - wlabel.size() - 6
-                                      : 0;
-                    lines.push_back(
-                        std::format(L"\033[36m{}{}{}\033[0m", wlabel, std::wstring(std::max<size_t>(1, padLen), L' '), L"CACHED"));
-                }
-            }
-        }
+        AppendCachedFromLines(target, consoleWidth, lines);
 
         // Steps
         for (const auto& step : target.steps)
--- a/src/windows/wslc/services/BuildImageCallback.cpp
+++ b/src/windows/wslc/services/BuildImageCallback.cpp
@@ -206,11 +206,11 @@ try
         {
             PrintPlainTextProgress();
         }
-        else 
+        else
         {
             StartTimerThread();
         }
-    } 
+    }
     catch (...)
     {
         LOG_CAUGHT_EXCEPTION();
@@ -294,8 +294,7 @@ std::wstring BuildImageCallback::FormatStepLine(const ViewStep& step, SHORT cons
 
 // Synthesize cached FROM lines for base image steps that were skipped
 // because they were resolved from a shared stage.
-void BuildImageCallback::AppendCachedFromLines(
-    const ViewTarget& target, SHORT consoleWidth, std::vector<std::wstring>& lines) const
+void BuildImageCallback::AppendCachedFromLines(const ViewTarget& target, SHORT consoleWidth, std::vector<std::wstring>& lines) const
 {
     if (target.steps.empty())
     {
@@ -331,8 +330,7 @@ void BuildImageCallback::AppendCachedFromLines(
             wlabel.resize(consoleWidth - 13);
             wlabel += L"...";
         }
-        auto padLen =
-            (wlabel.size() + 6 < static_cast<size_t>(consoleWidth)) ? static_cast<size_t>(consoleWidth) - wlabel.size() - 6 : 0;
+        auto padLen = (wlabel.size() + 6 < static_cast<size_t>(consoleWidth)) ? static_cast<size_t>(consoleWidth) - wlabel.size() - 6 : 0;
         lines.push_back(std::format(L"\033[36m{}{}{}\033[0m", wlabel, std::wstring(std::max<size_t>(1, padLen), L' '), L"CACHED"));
     }
 }
--- a/src/windows/wslc/services/BuildImageCallback.cpp
+++ b/src/windows/wslc/services/BuildImageCallback.cpp
@@ -14,31 +14,145 @@ Abstract:
 
 #include "precomp.h"
 #include "BuildImageCallback.h"
+#include <docker_schema.h>
 
 namespace wsl::windows::wslc::services {
 
 using wsl::windows::common::string::MultiByteToWide;
 
+// ── Formatting helpers ───────────────────────────────────────────────
+
+static std::wstring FormatBytes(int64_t b)
+{
+    if (b <= 0)
+    {
+        return L"0B";
+    }
+
+    constexpr double c_kB = 1024.0;
+    constexpr double c_MB = 1024.0 * 1024.0;
+    constexpr double c_GB = 1024.0 * 1024.0 * 1024.0;
+    auto val = static_cast<double>(b);
+
+    if (val >= c_GB)
+    {
+        return std::format(L"{:.2f}GB", val / c_GB);
+    }
+    if (val >= c_MB)
+    {
+        return std::format(L"{:.2f}MB", val / c_MB);
+    }
+    if (val >= c_kB)
+    {
+        return std::format(L"{:.2f}kB", val / c_kB);
+    }
+    return std::format(L"{}B", b);
+}
+
+static std::wstring FormatElapsed(std::chrono::steady_clock::time_point start)
+{
+    auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start);
+    auto totalSeconds = elapsed.count() / 1000;
+
+    if (totalSeconds >= 60)
+    {
+        auto minutes = totalSeconds / 60;
+        auto seconds = totalSeconds % 60;
+        return std::format(L"{}m{:.0f}s", minutes, static_cast<double>(seconds));
+    }
+
+    return std::format(L"{:.1f}s", elapsed.count() / 1000.0);
+}
+
+static std::wstring FormatDuration(std::chrono::milliseconds d)
+{
+    auto totalSeconds = d.count() / 1000;
+
+    if (totalSeconds >= 60)
+    {
+        auto minutes = totalSeconds / 60;
+        auto seconds = totalSeconds % 60;
+        return std::format(L"{}m{:.0f}s", minutes, static_cast<double>(seconds));
+    }
+
+    return std::format(L"{:.1f}s", d.count() / 1000.0);
+}
+
+// ── BuildImageCallback ──────────────────────────────────────────────
+
 BuildImageCallback::~BuildImageCallback()
 try
 {
-    // Capture any partial line so it's included in the error replay below; otherwise
-    // CollapseWindow() would discard it.
-    if (!m_pendingLine.empty())
-    {
-        m_pendingLine += '\n';
-        m_allLines.push_back(std::move(m_pendingLine));
-    }
+    StopTimerThread();
 
-    CollapseWindow();
-
-    // On build error (not cancellation), replay the full log output so the user can see what went wrong.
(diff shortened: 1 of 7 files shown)


src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -14,31 +14,145 @@ Module Name:
...
+    {
+        return std::format(L"{:.2f}kB", val / c_kB);
+    }
+    return std::format(L"{}B", b);
+}
+
+static std::wstring FormatElapsed(std::chrono::steady_clock::time_point start)
+{
+    auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start);
+    auto totalSeconds = elapsed.count() / 1000;
```
ptrivedi: There's code repetition here. FormatElapsed could simply call FormatDuration

src/windows/wslc/services/BuildView.cpp
```diff
@@ -0,0 +1,404 @@
...
+#include "precomp.h"
+#include "BuildView.h"
+#include <docker_schema.h>
+#include <wslutil.h>
+
+using namespace wsl::windows::common;
+
+namespace wsl::windows::wslc::services {
+
+// ── Utility functions ────────────────────────────────────────────────
```
ptrivedi, line 24: If ShortDigest, ParseElapsed etc. are just utility functions, it is cleaner to make them static or move them to anonymous namespace

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +168,380 @@ bool BuildImageCallback::IsCancelled() const
...
+        // Blank line separator between targets (not before first).
+        if (!firstTarget)
         {
-            // \r\n is a line ending; standalone \r overwrites the current line.
-            if (*(p + 1) != '\n')
+            lines.emplace_back();
+        }
+        firstTarget = false;
+
+        // Synthesize cached FROM steps for shared base images
```
ptrivedi, line 381: some things, for example the cached FROM steps logic, can be broken out into separate helper method for readability sake

src/windows/wslc/services/BuildView.cpp
```diff
@@ -0,0 +1,404 @@
...
+        while (std::getline(stream, line))
+        {
+            // Trim each sub-line
+            auto lineEnd = line.find_last_not_of(" \r\t");
+            line = (lineEnd != std::string::npos) ? line.substr(0, lineEnd + 1) : "";
+            auto lineStart = line.find_first_not_of(" \r\t");
+            line = (lineStart != std::string::npos) ? line.substr(lineStart) : "";
+            if (!line.empty())
+            {
+                step.logOutput.push_back(std::move(line));
```
ptrivedi, line 374: for a very long build step, the step.logOutput can grow pretty large and consume significant memory. Consider capping it here or in BuildFrameLines or could keep a small ring buffer of only the last n lines, where the cap of n could be somewhat generous...

src/windows/wslc/services/BuildView.cpp
```diff
@@ -0,0 +1,404 @@
...
+            newStep.stepLabel = parsed.stepLabel;
+            newStep.stepNum = parsed.stepNum;
+            newStep.stepTotal = parsed.stepTotal;
+            newStep.inputs = vertex.inputs;
+
+            stepIdx = m_targets[targetIdx].steps.size();
+            m_targets[targetIdx].steps.push_back(std::move(newStep));
+            m_digestIndex[vertex.digest] = {targetIdx, stepIdx};
+
+            if (std::find(targetsToSort.begin(), targetsToSort.end(), targetIdx) == targetsToSort.end())
```
ptrivedi: nit: number of targets is probably small, but my default inclination would have been to use unordered_set instead of a vector

src/windows/wslc/services/BuildView.h
```diff
@@ -0,0 +1,102 @@
...
+
+struct ViewTarget
+{
+    std::string name;
+    std::vector<ViewStep> steps;
+};
+
+class BuildView
+{
+public:
```
ptrivedi, line 66: for classes with non-trivial members like this one, the wsl codebase uses NON_COPYABLE  as a convention

ptrivedi (commented):
nice!

src/windows/wslc/services/BuildImageCallback.cpp
craigloewen-msft: Fixed! Thank you :)

craigloewen-msft:
Do we want to keep the `--verbose` option? Not sure what it would do given this new implementation.

dkbennett (approved):
Output looks good, I'm not entirely sure what it should look like, but I'd just normally ask Craig if it is good, and since this is Craig's PR I'm assuming it is correct. :)

benhillis:
Hi! This PR is approved but has merge conflicts with `master`. Could you rebase when you have a chance so we can land it? Thanks!

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -19,26 +19,129 @@ namespace wsl::windows::wslc::services {
...
+    if (val >= c_MB)
+    {
+        return std::format(L"{:.2f}MB", val / c_MB);
+    }
+    if (val >= c_kB)
+    {
+        return std::format(L"{:.2f}kB", val / c_kB);
+    }
+    return std::format(L"{}B", b);
+}
```
kvega005, line 49: Just a thought, have you considered `StrFormatByteSizeEx` ? It doesn't exactly match the docker style format, so I understand why you might not choose it. But it is more consistent with other windows components, and it has built-in localization for the decimal seperator.

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
+void BuildImageCallback::RenderFrame()
 {
     CONSOLE_SCREEN_BUFFER_INFO info{};
     THROW_IF_WIN32_BOOL_FALSE(GetConsoleScreenBufferInfo(m_console, &info));
-    // Use the visible window width (not buffer width), minus one column to avoid the
-    // deferred-wrap edge case when a line is exactly the window width. Clamp to at
-    // least zero so the value never goes negative (which would underflow when passed
-    // to std::wstring::resize).
-    const SHORT consoleWidth = std::max<SHORT>(0, info.srWindow.Right - info.srWindow.Left);
+    const SHORT consoleWidth = std::max<SHORT>(40, info.srWindow.Right - info.srWindow.Left);
```
kvega005, line 489: This is computed twice here and I think once in `TableOutput::GetConsoleWidth()`, we could probably create a hellper GetConsoleWidth that can be used in this  file and in `TableOutput.`

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -19,26 +19,129 @@ namespace wsl::windows::wslc::services {
...
+            if (hasError)
+            {
+                for (const auto& target : m_buildView.GetTargets())
+                {
+                    for (const auto& step : target.steps)
+                    {
+                        if (!step.error.empty())
+                        {
+                            buffer += std::format(L"\n\033[31m Error: {}\033[0m\n", MultiByteToWide(step.error));
+                            if (!step.logOutput.empty())
```
kvega005, line 125: Minor: each std::format(...) here allocates a temporary std::wstring that's immediately copied into buffer. Consider std::format_to(std::back_inserter(buffer), L"...", ...) to write directly into the buffer and skip the intermediate allocation. Same pattern shows up in RenderFrame() and BuildFrameLines() if you want to apply it more broadly — not blocking, just a small perf/cleanliness nit.

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
+    // Layout: step text + padding + timer, truncating the step text if needed.
+    auto timerWidth = static_cast<SHORT>(timer.size());
+    auto maxTextWidth = static_cast<size_t>(std::max<SHORT>(10, consoleWidth - timerWidth - 2));
+
+    if (wtext.size() > maxTextWidth && maxTextWidth > 3)
     {
-        if (*p == '\n')
+        wtext.resize(maxTextWidth - 3);
+        wtext += L"...";
+    }
```
kvega005, line 274: This pattern is repeated quite a bit. Would be nice to add a `TruncateWithEllipsis` helper like:

```
static void TruncateWithEllipsis(std::wstring& s, size_t maxWidth)
{
    if (s.size() > maxWidth && maxWidth > 3)
    {
        s.resize(maxWidth - 3);
        s += L"...";
    }
}
```

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
 
-            m_lines.push_back(std::move(m_pendingLine));
-            m_pendingLine.clear();
-            if (m_lines.size() > c_maxDisplayLines)
+    // Total build timer header (right-aligned, dimmed)
+    {
+        auto elapsed = FormatElapsed(m_buildStartTime);
+        auto label = std::format(L"Building {}", elapsed);
+        auto pad = static_cast<size_t>(std::max<SHORT>(0, consoleWidth - static_cast<SHORT>(label.size())));
+        lines.push_back(std::format(L"{}\033[2m{}\033[0m", std::wstring(pad, L' '), label));
```
kvega005, line 352: `std::format` supports dynamic width and alignment: `{:<{}}` (left-pad) and `{:>{}}` (right-pad). Several spots build a `std::wstring(pad, L' ')` just to splice in — could be folded into the format string. E.g. the header line:
`lines.push_back(std::format(L"\033[2m{:>{}}\033[0m", label, consoleWidth));`

Same applies to `FormatStepLine` and `AppendCachedFromLines`.

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
+        bool hasVisibleSteps = false;
+        for (const auto& step : target.steps)
+        {
+            if (step.started || step.completed)
             {
-                m_lines.pop_front();
+                hasVisibleSteps = true;
+                break;
             }
         }
```
kvega005, line 367: nit: `std::ranges::any_of(target.steps, [](const auto& s) { return s.started || s.completed; })`

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
+        for (const auto& digest : firstStep.inputs)
         {
-            // Store with the trailing newline so the byte count matches what is replayed.
-            // Cap retained log output to avoid unbounded growth on very long builds.
-            m_allLines.push_back(m_pendingLine + '\n');
-            m_allLinesBytes += m_allLines.back().size();
-            while (m_allLinesBytes > c_maxAllLinesBytes && !m_allLines.empty())
+            auto* inputStep = m_buildView.StepByDigest(digest);
+            if (inputStep && inputStep->stepLabel.find("] FROM ") != std::string::npos)
             {
```
kvega005, line 317: The string is searched twice. Once in the if condition, once to get pos. The second find is pure waste — its result is guaranteed to equal the first. Do one find, then test its result. Also the + 6 is a magic number. It's "skip past ] FROM" (6 characters). If someone changes the search string to "] FROM:" or "]  FROM " (extra space), they have to remember to update the 6 — and the compiler won't catch the mistake. 

```
auto* inputStep = m_buildView.StepByDigest(digest);
if (inputStep)
{
    auto pos = inputStep->stepLabel.find(c_fromMarker);
    if (pos != std::string::npos)
    {
        fromImage = inputStep->stepLabel.substr(pos + c_fromMarker.size());
        break;
    }
}
```

src/windows/wslc/services/BuildView.cpp
```diff
@@ -0,0 +1,425 @@
...
+        auto& step = m_targets[it->second.first].steps[it->second.second];
+        std::string text = wslutil::Base64Decode(logEntry.data);
+
+        // Trim trailing whitespace.
+        auto endPos = text.find_last_not_of(" \n\r\t");
+        text = (endPos != std::string::npos) ? text.substr(0, endPos + 1) : "";
+
+        // Trim leading whitespace.
+        auto startPos = text.find_first_not_of(" \n\r\t");
+        text = (startPos != std::string::npos) ? text.substr(startPos) : "";
```
kvega005, line 355: This pattern is repeated quite a bit. There is a `StripLeadingWhitespace` in `wsl::windows::common::string`, I think adding a StripWhitespace or Trim() is fitting.

kvega005: Or just add a StripTrailingWhiteSpace and do both.

src/windows/wslc/services/BuildView.cpp
```diff
@@ -0,0 +1,425 @@
...
+            auto lineEnd = line.find_last_not_of(" \r\t");
+            line = (lineEnd != std::string::npos) ? line.substr(0, lineEnd + 1) : "";
+            auto lineStart = line.find_first_not_of(" \r\t");
+            line = (lineStart != std::string::npos) ? line.substr(lineStart) : "";
+            if (!line.empty())
+            {
+                step.logOutput.push_back(std::move(line));
+                if (step.logOutput.size() > c_maxLogLinesPerStep)
+                {
+                    step.logOutput.erase(step.logOutput.begin());
```
kvega005, line 377: Erasing from the front of std::vector is O(n) — and you do it on every overflow log line, so writing N lines into a step is O(N · c_maxLogLinesPerStep). I think you can switch logOutput to std::deque<std::string>std::string — pop_front() is O(1).
Realistically, the perf hit is probably not very big, so it shouldn't be a huge issue, but its worth considering.

src/windows/wslc/services/BuildView.cpp
```diff
@@ -0,0 +1,425 @@
...
+using namespace wsl::windows::common;
+
+namespace wsl::windows::wslc::services {
+
+// ── Utility functions ────────────────────────────────────────────────
+
+std::string ShortDigest(const std::string& id)
+{
+    constexpr auto prefix = "sha256:";
+    constexpr size_t prefixLen = 7;
```
kvega005, line 29: Nit: magic number here. use `constexpr std::string_view prefix = "sha256:";` and `prefix.size()` or  strlen.

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
-    // Determine how many completed lines to show, leaving room for the pending line.
-    const bool showPending = !m_pendingLine.empty();
-    SHORT completedCount = static_cast<SHORT>(m_lines.size());
-    if (showPending && completedCount >= c_maxDisplayLines)
+    // Snapshot viewport height on first render.
+    if (m_viewportHeight == 0)
     {
-        completedCount = c_maxDisplayLines - 1;
+        m_viewportHeight = info.srWindow.Bottom - info.srWindow.Top + 1;
     }
```
kvega005, line 495: Viewport height is captured once on the first render and never updated. If the user resizes their terminal mid-build (especially shrinks it), the fixed-mode layout will render past the visible area or leave stale content. Width is re-read every frame, so this is the only resize blind spot. Probably not worth fixing now — just worth a comment here noting the limitation, or re-querying m_viewportHeight each frame.

kvega005 (commented):
Nit: Might make sense to add unit tests for `BuildView::ProcessMessage`. See `test/wslc/*UnitTests.cpp`

src/windows/wslc/services/BuildView.cpp
```diff
@@ -0,0 +1,425 @@
...
+{
+    auto it = m_digestIndex.find(digest);
+    if (it == m_digestIndex.end())
+    {
+        return nullptr;
+    }
+    return &m_targets[it->second.first].steps[it->second.second];
+}
+
+bool BuildView::HasErrors() const
```
kvega005, line 410: I don't see any callers for this. Did I miss something? If its not being used, best to remove.

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
-    if (status == nullptr || *status == '\0')
+    if (status == nullptr || *status == '\0' || IsCancelled())
     {
         return S_OK;
     }
 
-    // When cancellation is pending, skip all processing so the server's IO loop can
-    // return to its event wait and detect the cancel event promptly.
-    if (IsCancelled())
+    std::string statusStr(status);
```
ptrivedi, line 198: This could be a big json blob. We could avoid constructing a std::string here by changing ProcessMessage to take either string_view or passing it the raw LPCSTR. Even with string_view, I think a strlen() would be needed, which json::parse would have to do internally anyway. So better to have ProcessMessage take const char * here

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
+}
+
+// ── Frame line builder ──────────────────────────────────────────────
+// Builds all frame lines into a vector. Each entry is one visual line
+// (no embedded newlines). Used by both RenderFrame and the destructor's
+// final permanent output.
+
+std::vector<std::wstring> BuildImageCallback::BuildFrameLines(SHORT consoleWidth) const
+{
+    std::vector<std::wstring> lines;
```
ptrivedi, line 345: RenderFrame(), which calls BuildFrameLines() seems to be getting called every 200ms. Which means this will get called 5x per second and each time allocate a fresh wstring vector plus another dozen or so wstring objects. At least for the vector, consider keeping a member variable, then clear and reuse it on each frame to avoid repeated heap allocations

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
+        lines.push_back(std::format(L"\033[36m{}{}{}\033[0m", wlabel, std::wstring(std::max<size_t>(1, padLen), L' '), L"CACHED"));
+    }
+}
+
+// ── Frame line builder ──────────────────────────────────────────────
+// Builds all frame lines into a vector. Each entry is one visual line
+// (no embedded newlines). Used by both RenderFrame and the destructor's
+// final permanent output.
+
+std::vector<std::wstring> BuildImageCallback::BuildFrameLines(SHORT consoleWidth) const
```
ptrivedi, line 343: Instead of returning a vector, consider appending to a pre-allocated buffer or use a reusable member variable

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
+        }
+        auto label = std::format("[{} {}/{}] FROM{}", target.name, n, firstStep.stepTotal, fromImage);
+        auto wlabel = MultiByteToWide(label);
+        if (static_cast<SHORT>(wlabel.size()) > consoleWidth - 10)
+        {
+            wlabel.resize(consoleWidth - 13);
+            wlabel += L"...";
+        }
+        auto padLen = (wlabel.size() + 6 < static_cast<size_t>(consoleWidth)) ? static_cast<size_t>(consoleWidth) - wlabel.size() - 6 : 0;
+        lines.push_back(std::format(L"\033[36m{}{}{}\033[0m", wlabel, std::wstring(std::max<size_t>(1, padLen), L' '), L"CACHED"));
```
ptrivedi, line 334: The line above has repeated std::format + temporary wstring creation in a tight loop. Each std::format will allocate a temporary wstring, the push_back will then move it. Since this could be in a hot rendering path, I would consider pre-reserving the buffer or using std::format_to with a back inserter into a single output string to reduce repeated allocations

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -54,173 +157,401 @@ bool BuildImageCallback::IsCancelled() const
...
+}
 
-    // Match the specific "log" sentinel sent by WSLCSession::BuildImage rather than
-    // accepting any non-empty id, so future or unrelated id usage defaults to permanent.
-    const bool isLog = (id != nullptr && std::string_view{id} == "log");
+// ── Step line formatter ─────────────────────────────────────────────
+
+std::wstring BuildImageCallback::FormatStepLine(const ViewStep& step, SHORT consoleWidth) const
+{
+    auto wtext = MultiByteToWide(step.stepLabel);
```
ptrivedi, line 242: nit: this narrow to wide conversion will happen on every render. I don't think the step labels change after creation, so could pre-convert and store

PR #40534, opened 2026-05-13
Move IO related classes to their own file & namespace

OneBlue:
## PR Checklist


dkbennett (approved):
Looks like PartialHandleRead in the windows test common is not included, there may be some other useful IO classes in the test code that could be moved to this new io namespace.

OneBlue:
> Looks like PartialHandleRead in the windows test common is not included, there may be some other useful IO classes in the test code that could be moved to this new io namespace.

We could consider that, but I have a preference towards separating product & test-only code, so I'll leave those in the tests for now. We can revisit later if needed

Issue #40488, opened 2026-05-10
Cannot start WSL: Wsl/Service/CreateInstance/CreateVm/HCS/ERROR_FILE_NOT_FOUND


Hai-Ren:
### Solution

1. Download the installer from https://github.com/microsoft/WSL/releases/download/2.7.13/wsl.2.7.13.0.x64.msi. Please check this is your current version.
2. Exit docker desktop.
3. Right click on the installer -> show more options -> repair<img width="450" height="76" alt="Image" src="https://github.com/user-attachments/assets/905a18e2-bcaf-4812-a52f-f06d422cf6ec" />

**Roll back also works but is not required**

The issue was fixed in 2.9.3 #40625. However, the fix only works when updating from a fixed version.

---

### Windows Version

Microsoft Windows [Version 10.0.26200.8246]

### WSL Version

2.7.3.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.114.1-1

### Distro Version

Ubuntu 22.04

### Other Software

- Visual Studio Code 1.109.2 with WSL extension
- CERN ROOT (C++ Data Analysis Framework), version 6.26
- WSLg (GUI support for WSL)

### Repro Steps

The issue started after an abrupt crash, and now the broken state is permanent for any WSL2 operation.

1. I was running a C++ plotting task using CERN ROOT inside WSL2 (Ubuntu 22.04) via VS Code Remote.
2. The ROOT drawing failed to render, ROOT exited unexpectedly, and VS Code immediately lost connection to WSL.
3. I opened a new Windows PowerShell and ran `wsl`. It failed with Wsl/Service/CreateInstance/CreateVm/HCS/ERROR_FILE_NOT_FOUND
4. I rebooted the Windows host machine completely and ran `wsl` again. The same error persists.
5. Attempted to bypass the broken distro by installing a brand new WSL2 instance: `wsl --install -d Ubuntu-24.04`. This ALSO fails with the exact same HCS error.
6. Switched default version to WSL1 (`wsl --set-default-version 1`) and ran the install command again. The installation SUCCEEDED and WSL1 works perfectly.

**Troubleshooting steps I have already taken (None of these worked for WSL2):**
- Verified the VHDX file exists, is on an NTFS drive, and the folder is NOT compressed.
- Checked `diskpart` and verified the original ext4.vhdx file is not corrupted.
- Unregistered the old distro and tried `wsl --import-in-place`.
[...]

PR #40535, opened 2026-05-14
Fix system.vhd missing after update

chemwolf6922:

When these events happen during an update. The system.vhd could be missing afterwards:
1. During uninstall, the system.vhd is held. So, it's removal is delayed to next reboot.
2. During install, the system.vhd is no longer held. So, it's replacement is done without a reboot.
3. The installer requests for a reboot.
4. After the reboot. The system executes the scheduled delete and delete the newly installed system.vhd.

This PR fixes this issue by adding two steps in the installer that:
1. Get the file list in the new installation.
2. Cancel any scheduled reboot delete on files that's in the new installation. 

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/40488

According to this user provided log: https://gist.github.com/BBQ4ever/52e437af3324aafe02bf511027f7573e
line 1608: MSIRMSHUTDOWN was set to 1.
line 5014 - 5025: system.vhd was still held. And it was scheduled to be reboot deleted.
line 9505 - 9509: system.vhd was no longer held. Copying the new file succeeded.
line 12195: installer returned with 1641, telling the system to reboot.

## Validation Steps Performed

Add new test InstallerTests::InstallScrubsPendingDeletesForInstalledFiles.

### DIFF
--- a/msipackage/package.wix.in
+++ b/msipackage/package.wix.in
@@ -611,6 +611,29 @@
         <CustomAction Id="InstallMsix.SetProperty" Return="check" Property="InstallMsix" Value='[DATABASE]' Execute='immediate' />
         <CustomAction Id="CalculateWslSettingsProtocolIds" Impersonate="no" BinaryRef="wslinstall.dll" DllEntry="CalculateWslSettingsProtocolIds" Return="check" Execute='immediate' />
 
+        <!-- Builds the list of files this MSI will install (resolved to absolute target
+             paths) and stashes it into the property whose name matches the deferred CA
+             below. MSI propagates that property as CustomActionData. -->
+        <CustomAction Id="BuildPendingDeleteScrubList"
+            Impersonate="no"
+            BinaryRef="wslinstall.dll"
+            DllEntry="BuildPendingDeleteScrubList"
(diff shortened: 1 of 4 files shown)


benhillis:
I am extremely hesitant to add a huge custom install action to work around something that is almost certainly a MSIX deployment issue. I would very strongly prefer if you guys could work on getting a reliable repro of the issue and root cause the MSIX upgrade bug so it can be fixed and not worked around.

benhillis (changes requested):
I do not want to add a huge custom install action to work around a deployment stack bug.

benhillis:
Another option, if we can detect we are in this "reboot needed" state, should we just return an error to the user telling them they need to reboot?

OneBlue (changes requested):
Agreed with @benhillis. To fix think I think we should get a repro and understand what's causing the installation to stop half-transaction. My personal theory is that this is something that only happens during Store updates (since we run the MSIX installer path many times in the tests, and we never saw this). 

Overall custom MSI actions are pretty high-risk, so unless we're 100% certain that this is the only solution, I would recommend against it

PR #40542, opened 2026-05-14
p9fs: hold an O_PATH fd per fid in File::Walk

benhillis:
The Plan9 server's `File::Walk` stored each fid's location as a path string
and revalidated it by path on every subsequent operation
(`fstatat(RootFd, m_FileName, ...)`, etc.). Between the Walk-time validation
and any later use, the named entry on disk could be replaced (e.g. swapped
for a symlink), and subsequent ops would re-resolve the path against the new
target.

Pin the validated inode per fid with an `O_PATH | O_NOFOLLOW` fd:

- Add `wil::unique_fd m_PathFd` and a `PathFd()` helper that falls back to
  `m_Root->RootFd` for the share-root fid.
- `Walk` opens the child via `openat(parent_PathFd, leaf, O_PATH | O_NOFOLLOW | O_CLOEXEC)`,
  stats it with `AT_EMPTY_PATH`, runs the existing mount-boundary check,
  then commits `m_PathFd`.
- `Stat` / `GetAttr` `fstatat` through the held fd.
- `Open` / `OpenFile` reject symlink qids with `ELOOP` and `Reopen` via
  `/proc/self/fd/N` (`O_NOFOLLOW` stripped because `util::Reopen`'s
  `O_NOFOLLOW` branch is path-based and would defeat the fix).
- `Create` issues `openat` against the parent's held fd with `O_NOFOLLOW`
  and dups the resulting fd to populate the new fid's `m_PathFd`.
- The copy constructor dups `m_PathFd` and propagates `m_Device`.

The mount-boundary check (drvfs/9p/virtiofs) is preserved. Path-based child
ops (chmod, chown, utimensat, unlinkat, rename, mkdir, mknod, symlink, link,
readlink, xattr, access) are intentionally left untouched.

## Test
`Plan9Tests::*` pass.

### DIFF
--- a/src/linux/plan9/p9file.cpp
+++ b/src/linux/plan9/p9file.cpp
@@ -144,8 +144,10 @@ bool File::IsOnRoot(const std::shared_ptr<const IRoot>& root)
     return m_Root == root;
 }
 
-// Returns the fd that pins this fid's inode (O_PATH fd from Walk/Create, or
-// the share RootFd for the root fid). Caller must hold m_Lock.
+// Returns the fd that pins this fid's inode (installed by Walk/Create, or the
+// share RootFd for the root fid). Callers must hold m_Lock once the fid is
+// published; Walk() may call this before publication, while it still has
+// exclusive access to the fid.
 int File::PathFd() const noexcept
 {
     return m_PathFd ? m_PathFd.get() : m_Root->RootFd;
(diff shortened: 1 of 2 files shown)


kvega005 (dismissed):
The copilot callouts, i think, are valid, but I don't think either warrants a code change. I would just update the comments.

PR #40544, opened 2026-05-14
Use inline _locComment syntax for locked tokens in en-US WSL.adml

benhillis:
The standalone `<!-- {Locked="..."} -->` XML comments above each `<string>` in `intune/en-US/WSL.adml` are silently ignored by the POMXML parser used by `TouchdownBuildTask`, so locked tokens were getting translated in every locale (e.g. `NetworkingMode` values `Mirrored`/`None`/`NAT` were translated; `.wslconfig` key names were translated; etc.).

Move the directives inside each `<string>` using the inline form the parser actually honors:

```xml
<string id="AllowWSL"><!-- _locComment='{Locked="Windows Subsystem for Linux"}' -->Allow the Windows Subsystem for Linux</string>
<string id="NetworkingModeMirrored"><!-- _locComment="{Locked}" -->Mirrored</string>
```

Also update `tools/devops/validate-localization.py` so its ADML parser extracts locked tokens from the inline form (instead of standalone sibling comments) — keeps the baseline-token-presence check catching authoring mistakes in en-US.

Localized `intune/<locale>/WSL.adml` files are intentionally not touched here; Touchdown will regenerate them on the next nightly run, after which their existing translations of locked tokens will get cleaned up. The validator continues to emit warnings (not errors) for translated-but-not-preserved tokens so the transition period doesn't fail CI.

### DIFF
--- a/intune/en-US/WSL.adml
+++ b/intune/en-US/WSL.adml
@@ -5,84 +5,58 @@
   <description>Windows Subsystem for Linux</description>
   <resources>
     <stringTable>
-        <!-- {Locked="Windows Subsystem for Linux"} -->
-        <string id="WindowsSubsystemForLinux">Windows Subsystem for Linux</string>
+        <string id="WindowsSubsystemForLinux"><!-- _locComment='{Locked="Windows Subsystem for Linux"}' -->Windows Subsystem for Linux</string>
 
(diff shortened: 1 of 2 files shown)


benhillis:
test pipeline run: https://github.com/microsoft/WSL/pull/40545

OneBlue (approved):

PR #40549, opened 2026-05-15
Support connecting and disconnecting containers to networks after creation

beena352:
Implements ConnectToNetwork and DisconnectFromNetwork methods on IWSLCContainer, enabling containers to join/leave networks after they are created and running.

## PR Checklist

- [X ] **Tests:** Added/updated if needed and all pass

- Acquires m_lock.lock_shared() for thread safety
- Validates input before calling Docker: empty network name, unsupported IP address, Host/None network mode
- Docker errors surfaced via THROW_DOCKER_USER_ERROR_MSG

## Validation Steps Performed
- ConnectDisconnectContainerNetworkRoundTripTest: Create bridged container → create network → connect → verify via inspect (network appears with IP) → disconnect → verify gone
- ConnectDisconnectContainerValidationTest:

- Empty NetworkName → E_INVALIDARG (both connect and disconnect)
- Nonexistent network → WSLC_E_NETWORK_NOT_FOUND (both connect and disconnect)
- Host and None mode containers → E_INVALIDARG (both connect and disconnect)
- ContainerIpAddress set → E_NOTIMPL


src/windows/service/inc/wslc.idl
```diff
@@ -790,6 +790,8 @@ interface IWSLCSession : IUnknown
     HRESULT DeleteNetwork([in] LPCSTR Name);
     HRESULT ListNetworks([out, size_is(, *Count)] WSLCNetworkInformation** Networks, [out] ULONG* Count);
     HRESULT InspectNetwork([in] LPCSTR Name, [out] LPSTR* Output);
+    HRESULT AttachContainerToNetwork([in] LPCSTR ContainerId, [in] const WSLCNetworkAttachment* Attachment);
```
OneBlue: I would recommend moving those methods to `WSLCContainer` instead. That way we don't need to perform the container lookup in the method's implementation. That will also take care of container names and id prefixes for us

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2295,6 +2295,115 @@ try
...
+    THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_virtualMachine);
+
+    std::scoped_lock locks(m_containersLock, m_networksLock);
+    std::erase_if(m_containers, [](const auto& entry) { return entry.second->State() == WslcContainerStateDeleted; });
+
+    auto containerIt = m_containers.find(containerId);
+    THROW_HR_WITH_USER_ERROR_IF(
+        WSLC_E_CONTAINER_NOT_FOUND, Localization::MessageWslcContainerNotFound(containerId), containerIt == m_containers.end());
+
+    auto networkIt = m_networks.find(networkName);
```
OneBlue: Could we just let docker handle the network not found error for us ? That would allow us to easily support the `container:<id>` case

dkbennett (dismissed):
LGTM, the copilot comments on the ABI can be ignored since we haven't shipped it yet, but soon we can't ignore that. The comments about sending to docker without checking may have some merit, though I believe the model is send it to docker and then we report the docker error if there's a problem, so that's probably OK, and something easy to fix later if not.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -2348,6 +2348,66 @@ void WSLCContainerImpl::GetLabels(WSLCLabelInformation** Labels, ULONG* Count) c
     *Labels = labelsArray.release();
 }
 
+void WSLCContainerImpl::AttachToNetwork(const WSLCNetworkAttachment* Attachment)
```
kvega005: Nit: I think we should remain consistent with naming. If docker calls it "connect", we should also call it connect. (Referriing to `m_dockerClient.ConnectContainerToNetwork(networkName, request);`)

test/windows/WSLCTests.cpp
```diff
@@ -6248,6 +6248,117 @@ class WSLCTests
...
+    WSLC_TEST_METHOD(DetachContainerEmptyNetworkNameTest)
+    {
+        const std::string containerName = "test-detach-empty-net";
+
+        WSLCContainerLauncher launcher("debian:latest", containerName, {"sleep", "99999"}, {});
+        auto container = launcher.Launch(*m_defaultSession);
+
+        VERIFY_ARE_EQUAL(E_INVALIDARG, container.Get().DetachFromNetwork(""));
+        ValidateCOMErrorMessage(L"Network name cannot be empty.");
+    }
```
kvega005: I think the general convention in this test file is to keep related scenarios in a single WSLC_TEST_METHOD and use local helper lambdas for shared setup/assertions. Could we follow that pattern here to reduce duplication and keep the matrix easy to scan?

Examples in this file:
- ContainerListFilter uses listContainers and expectContainers helpers: WSLCTests.cpp:5587
- ListImages uses shared setup/cleanup with multiple scenario blocks in one method: WSLCTests.cpp:760

test/windows/WSLCTests.cpp
```diff
@@ -6248,6 +6248,117 @@ class WSLCTests
...
+        const std::string containerName = "test-attach-empty-net";
+
+        WSLCContainerLauncher launcher("debian:latest", containerName, {"sleep", "99999"}, {});
+        auto container = launcher.Launch(*m_defaultSession);
+
+        WSLCNetworkAttachment attachment{};
+        attachment.NetworkName = "";
+        VERIFY_ARE_EQUAL(E_INVALIDARG, container.Get().AttachToNetwork(&attachment));
+        ValidateCOMErrorMessage(L"Network name cannot be empty.");
+    }
```
kvega005: Could we add explicit negative tests for AttachToNetwork and DetachFromNetwork with non-existent network name that assert we get the network not found hresult and a good COM error message.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -2348,6 +2348,66 @@ void WSLCContainerImpl::GetLabels(WSLCLabelInformation** Labels, ULONG* Count) c
...
+    request.Container = m_id;
+
+    try
+    {
+        m_dockerClient.DisconnectContainerFromNetwork(networkName, request);
+    }
+    catch (const DockerHTTPException& e)
+    {
+        THROW_DOCKER_USER_ERROR_MSG(e, "Failed to detach container '%hs' from network '%hs'", m_id.c_str(), networkName.c_str());
+    }
```
kvega005, line 2371: If the error code is 404, we should return the specfic `WSLC_E_NETWORK_NOT_FOUND`. There is an example of this in line ~207 (`ProcessPortMappings`).

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -2361,6 +2421,24 @@ try
 }
 CATCH_RETURN();
 
+HRESULT WSLCContainer::AttachToNetwork(const WSLCNetworkAttachment* Attachment)
+try
+{
+    COMServiceExecutionContext context;
+    RETURN_HR_IF(E_POINTER, Attachment == nullptr);
```
kvega005: I prefer having this check inside the Impl method where Attachment is actually being used.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -2361,6 +2421,24 @@ try
...
+    return CallImpl(&WSLCContainerImpl::AttachToNetwork, Attachment);
+}
+CATCH_RETURN();
+
+HRESULT WSLCContainer::DetachFromNetwork(LPCSTR NetworkName)
+try
+{
+    COMServiceExecutionContext context;
+    RETURN_HR_IF(E_POINTER, NetworkName == nullptr);
+    return CallImpl(&WSLCContainerImpl::DetachFromNetwork, NetworkName);
```
kvega005: Ditto: prefer the nullcheck where the pointer is actually being dereferenced.

kvega005 (changes requested):
I think this stuff uint the PR description got copied from another PR:
- Uses existing lock pattern: m_lock.lock_shared() then scoped_lock(m_containersLock, m_networksLock)
- Cleans up deleted containers via erase_if before lookup
- Error handling follows DeleteNetwork pattern: local validation with exact errors, Docker errors surfaced via catch-all

src/windows/wslcsession/WSLCContainer.cpp
beena352: This was intentional- I used Attach/Detach here to stay consistent with the existing WSL naming (AttachDisk/DetachDisk in wslc.idl). In this codebase, Connect is mostly used for networking/socket-related operations. The Docker wrapper still uses ConnectContainerToNetwork since that matches Docker’s own API endpoint. Later, when we add the CLI commands, those will probably use connect/disconnect since that’s the terminology Docker users are already familiar with.

beena352:
> I think this stuff uint the PR description got copied from another PR:
> 
> * Uses existing lock pattern: m_lock.lock_shared() then scoped_lock(m_containersLock, m_networksLock)
> * Cleans up deleted containers via erase_if before lookup
> * Error handling follows DeleteNetwork pattern: local validation with exact errors, Docker errors surfaced via catch-all

Those were from the original design, I moved the methods from IWSLCSession to IWSLCContainer after review feedback but didn't update the description. Fixed now.

src/windows/wslcsession/WSLCContainer.cpp
kvega005: I have to disagree here. The Attach/Detach was used for disks because that's nomenclature that the lower level API (HCS) and existing WSL APIs use. However, network is container network is an existing docker concept.

**Sidenote:** There is also a difference at the HCS level between `HcsNetworkAttach` and `HcsConnectComputeSystemNetworkEndpoint`. Docker's `network connect` is semantically analogous to both Attach and Connect.

Additionally, consider, we will eventually add this to the CLI and SDK, and at the CLI it will be called `wslc network connect` copying `docker network connect`.

It is awkward when we have a CLI argument called connect which calls `Container.AttachNetwork()` which calls `docker.connectNetwork`. This creates two separate names for the same action.

Same issue at the SDK level. We want to remain consistent with docker nomenclature, so the usage is familiar to developers that use the docker CLI and APIs. We would either have to expose a SDK API called `AttachNetwork` which contradicts the already established CLI `connect` nomenclature, or call the SDK API `ConnectNetwork` which will call the runtimes `AttachNetwork`. Also, awkward and creates multiple names for the same concept. 

This is a minor detail, but it is important to remain consistent with naming. It keeps the code clean and avoids future confusion.

**In short**: In my opinion, the benefit of keeping the naming of a concept consistent across all levels of the call stack is greater than keeping the naming parallel to other concepts at the same level of the call stack.

test/windows/WSLCTests.cpp
```diff
@@ -6248,6 +6248,113 @@ class WSLCTests
...
+
+            attachment.NetworkName = nonExistentNetwork.c_str();
+            VERIFY_ARE_EQUAL(WSLC_E_NETWORK_NOT_FOUND, container.Get().ConnectToNetwork(&attachment));
+            ValidateCOMErrorMessage(expectedError.c_str());
+
+            VERIFY_ARE_EQUAL(WSLC_E_NETWORK_NOT_FOUND, container.Get().DisconnectFromNetwork(nonExistentNetwork.c_str()));
+            ValidateCOMErrorMessage(expectedError.c_str());
+        }
+
+        LogInfo("Test: Host and none mode rejection");
```
kvega005: Do we need these? No other test in this file does this type of logging.

src/windows/service/inc/wslc.idl
```diff
@@ -536,6 +536,8 @@ interface IWSLCContainer : IUnknown
     HRESULT GetLabels([out, size_is(, *Count)] WSLCLabelInformation** Labels, [out] ULONG* Count);
     HRESULT Kill([in] WSLCSignal Signal);
     HRESULT Stats([out] LPSTR* Output);
+    HRESULT ConnectToNetwork([in] const WSLCNetworkAttachment* Attachment);
```
kvega005: If we are renaming from Attachment to Connection, we should consider renaming this argument as well. Maybe something like NetworkConnection? Docker calls the argument for their equivalent `EndpointConfig`. There is also precedent in the COM struct to use "options" struct.

test/windows/WSLCTests.cpp
```diff
@@ -6248,6 +6248,113 @@ class WSLCTests
         ValidateCOMErrorMessage(L"Container network name is required for custom network type.");
     }
 
+    WSLC_TEST_METHOD(ConnectDisconnectContainerNetworkRoundTripTest)
```
kvega005: There is a lot of repeated code in this method. 

Echoing [this comment](https://github.com/microsoft/WSL/pull/40549#discussion_r3284063519), use local helper lambdas for shared setup/assertions. Could we follow that pattern here to reduce duplication and keep the matrix easy to scan?

Examples in this file:

- ContainerListFilter uses listContainers and expectContainers helpers: WSLCTests.cpp:5587
- ListImages uses shared setup/cleanup with multiple scenario blocks in one method: WSLCTests.cpp:760

src/windows/inc/docker_schema.h
```diff
@@ -158,6 +158,22 @@ struct Network
     NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(Network, Id, Name, Driver, Scope, Internal, IPAM, Labels);
 };
 
+struct ConnectNetworkRequest
```
OneBlue: Looks like `ConnectNetworkRequest`  and `DisconnectNetworkRequest` have the same fields. Could we only use one type for both ?

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -2348,6 +2348,75 @@ void WSLCContainerImpl::GetLabels(WSLCLabelInformation** Labels, ULONG* Count) c
...
 
+void WSLCContainerImpl::ConnectToNetwork(const WSLCNetworkConnectionOptions* Options)
+{
+    THROW_HR_IF(E_POINTER, Options == nullptr);
+    THROW_HR_WITH_USER_ERROR_IF(E_NOTIMPL, Localization::MessageWslcContainerIpAddressNotSupported(), Options->ContainerIpAddress != nullptr);
+
+    THROW_HR_WITH_USER_ERROR_IF(
+        E_INVALIDARG, Localization::MessageWslcNetworkNameRequired(), !Options->NetworkName || strlen(Options->NetworkName) == 0);
+
+    const std::string networkName = Options->NetworkName;
```
OneBlue: nit: I think we can skip this additional variable (same for the disconnect path)

src/windows/wslcsession/WSLCContainer.h
```diff
@@ -125,6 +127,11 @@ class WSLCContainerImpl
         return m_containerFlags;
     }
 
+    WSLCContainerNetworkType NetworkMode() const noexcept
```
OneBlue: Looks like this is unused

test/windows/WSLCTests.cpp
```diff
@@ -6248,6 +6248,101 @@ class WSLCTests
         ValidateCOMErrorMessage(L"Container network name is required for custom network type.");
     }
 
+    WSLC_TEST_METHOD(ConnectDisconnectContainerNetworkTest)
+    {
+        auto launchContainer = [&](const std::string& name,
```
OneBlue: I recommend adding test coverage for: 

- Trying to connect / disconnect from the container's primary network
- Trying to connect to the same network twice

OneBlue (dismissed):
LGTM

kvega005 (approved):
LGTM

PR #40551, opened 2026-05-15
Bundle and mount Linux kernel headers in WSL2 distros

benhillis:
## Summary

Bundles the Linux kernel UAPI headers shipped in the `Microsoft.WSL.Kernel` nuget package and mounts them read-only into every WSL2 distro by default. Pattern mirrors the existing `kernelModules` plumbing.

## What it does

- **Build**: stages `linux-headers/` next to the kernel image in the dev binary path (`CMakeLists.txt`, `UserConfig.cmake.sample`) and packages them in the MSI (`msipackage/package.wix.in`).
- **Service**: new `kernelHeaders=` setting in `.wslconfig` (empty = bundled default, or a custom path). Validated alongside `kernel`/`kernelModules` in `WslCoreVm`.
- **Guest**: mounted via 9p at `/usr/src/linux-headers-$(uname -r)` with a `/lib/modules/$(uname -r)/build` symlink so out-of-tree module builds and tools that key off `uname -r` find them.
- **UI**: Developer page in wslsettings exposes the path with the same UX as the kernel/kernelModules controls.

## Testing

- New `KernelHeaders` TAEF unit test covers default-mount, custom path, and validation errors.
- Manually verified a fresh distro shows `/usr/src/linux-headers-*` populated and `make headers_check` works against it.

### DIFF
--- a/src/linux/init/config.cpp
+++ b/src/linux/init/config.cpp
@@ -1121,6 +1121,68 @@ Return Value:
     CATCH_LOG()
 
     //
+    // Move the kernel headers temporary mount into its final location and create the symlinks
+    // expected by Debian/Ubuntu tooling. The target path is supplied by mini_init as
+    // /usr/src/linux-headers-<uname -r>/include.
+    //
+    // N.B. Failure to install the headers is non-fatal; the distro will simply boot without
+    //      them, mirroring the kernel modules behavior above.
+    //
+
+    try
+    {
+        auto tempMount = RemoveMountAndEnvironmentOnScopeExit(LX_WSL2_KERNEL_HEADERS_MOUNT_ENV);
+        if (tempMount)
+        {
+            const char* target = getenv(LX_WSL2_KERNEL_HEADERS_PATH_ENV);
+            if (target)
+            {
+                std::string targetPath{target};
+                unsetenv(LX_WSL2_KERNEL_HEADERS_PATH_ENV);
+
+                if (tempMount.MoveMount(targetPath.c_str()))
+                {
+                    // Derive the headers root directory by stripping the trailing "/include".
+                    constexpr std::string_view c_includeSuffix = "/include";
+                    if (targetPath.ends_with(c_includeSuffix))
+                    {
+                        const std::string headersRoot = targetPath.substr(0, targetPath.size() - c_includeSuffix.size());
+
+                        utsname unameBuffer{};
+                        if (uname(&unameBuffer) == 0)
+                        {
+                            // See main.cpp; convert release to std::string so subsequent appends
+                            // don't follow embedded NUL bytes from the char[65] buffer.
+                            const std::string release{unameBuffer.release};
+                            const std::string modulesDir = std::format("/lib/modules/{}", release);
+                            if (UtilMkdirPath(modulesDir.c_str(), 0755) == 0)
+                            {
+                                for (const auto* linkName : {"build", "source"})
+                                {
+                                    const std::string linkPath = modulesDir + "/" + linkName;
+                                    if ((symlink(headersRoot.c_str(), linkPath.c_str()) < 0) && (errno != EEXIST))
+                                    {
+                                        LOG_ERROR("symlink({}, {}) failed {}", headersRoot, linkPath, errno);
+                                    }
+                                }
+                            }
+                            else
+                            {
+                                LOG_ERROR("UtilMkdirPath({}) failed {}", modulesDir, errno);
+                            }
+                        }
+                        else
+                        {
+                            LOG_ERROR("uname failed {}", errno);
+                        }
+                    }
+                }
+            }
+        }
+    }
+    CATCH_LOG()
+
+    //
     // Change the permission of some devtmpfs devices to be more permissive.
     //
     // N.B. These devices may not be present with a custom kernel config.
--- a/src/linux/init/config.cpp
+++ b/src/linux/init/config.cpp
@@ -1120,15 +1120,6 @@ Return Value:
     }
     CATCH_LOG()
 
-    //
-    // Move the kernel headers temporary mount into its final location and create the symlinks
-    // expected by Debian/Ubuntu tooling. The target path is supplied by mini_init as
-    // /usr/src/linux-headers-<uname -r>/include.
-    //
-    // N.B. Failure to install the headers is non-fatal; the distro will simply boot without
-    //      them, mirroring the kernel modules behavior above.
-    //
-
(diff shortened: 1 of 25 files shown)


src/linux/init/config.cpp
```diff
@@ -1120,6 +1120,68 @@ Return Value:
     }
     CATCH_LOG()
 
+    //
+    // Move the kernel headers temporary mount into its final location and create the symlinks
+    // expected by Debian/Ubuntu tooling. The target path is supplied by mini_init as
```
benhillis: is this distro agnostic?

src/linux/init/config.cpp
```diff
@@ -1120,6 +1120,68 @@ Return Value:
...
+                    // Derive the headers root directory by stripping the trailing "/include".
+                    constexpr std::string_view c_includeSuffix = "/include";
+                    if (targetPath.ends_with(c_includeSuffix))
+                    {
+                        const std::string headersRoot = targetPath.substr(0, targetPath.size() - c_includeSuffix.size());
+
+                        utsname unameBuffer{};
+                        if (uname(&unameBuffer) == 0)
+                        {
+                            // See main.cpp; convert release to std::string so subsequent appends
```
benhillis: get rid of this comment

src/linux/init/config.cpp
```diff
@@ -1120,6 +1120,68 @@ Return Value:
...
+                            // don't follow embedded NUL bytes from the char[65] buffer.
+                            const std::string release{unameBuffer.release};
+                            const std::string modulesDir = std::format("/lib/modules/{}", release);
+                            if (UtilMkdirPath(modulesDir.c_str(), 0755) == 0)
+                            {
+                                for (const auto* linkName : {"build", "source"})
```
benhillis: we need two symlinks?

src/linux/init/main.cpp
```diff
@@ -3439,6 +3455,37 @@ try
...
+            // Mount the kernel headers 9p share at a temporary path. The headers will be moved
+            // to /usr/src/linux-headers-<uname -r>/include by the distro init process. Failure
+            // here is non-fatal; if the headers cannot be mounted the distro will simply boot
+            // without them.
+            //
+            // N.B. uname() is called first so a (highly unlikely) failure does not leave behind
+            //      an orphaned mount at KERNEL_HEADERS_TEMP_PATH that the distro init wouldn't
+            //      know to clean up.
+            utsname UnameBuffer{};
+            if (uname(&UnameBuffer) < 0)
```
benhillis: maybe switch to THROW? really shouldn't fail...

src/linux/init/main.cpp
```diff
@@ -3439,6 +3455,37 @@ try
...
+            {
+                LOG_ERROR("uname failed, {}", errno);
+            }
+            else if (MountPlan9(LXSS_KERNEL_HEADERS_SHARE, KERNEL_HEADERS_TEMP_PATH, true) < 0)
+            {
+                LOG_ERROR("Failed to mount kernel headers 9p share, {}", errno);
```
benhillis: helper already logs an error?

src/linux/init/main.cpp
```diff
@@ -3439,6 +3455,37 @@ try
...
+            }
+            else
+            {
+                // N.B. Convert UnameBuffer.release (a char[65] array) to std::string first so
```
benhillis: get rid of this comment.

src/linux/init/config.cpp
benhillis: Reworded out of existence — only the universal `/lib/modules/$(uname -r)/build` symlink remains, which is what every distro's out-of-tree build tooling queries. The `/usr/src/linux-headers-*` mount path itself is the Debian/Ubuntu convention but harmless on other distros.

src/linux/init/config.cpp
benhillis: Done.

src/linux/init/config.cpp
benhillis: Dropped `source` — we only ship UAPI headers, not a full source tree, so `source` was misleading. Just `build` now.

src/linux/init/main.cpp
benhillis: Done — switched to `THROW_LAST_ERROR_IF`.

src/linux/init/main.cpp
benhillis: Right, dropped the redundant `LOG_ERROR` (UtilMount logs via WIL on throw).

src/linux/init/main.cpp
benhillis: Done.

src/linux/init/main.cpp
```diff
@@ -3439,6 +3455,18 @@ try
             }
         }
 
+        if (ConfigMessage->MountKernelHeaders)
+        {
+            utsname UnameBuffer{};
+            THROW_LAST_ERROR_IF(uname(&UnameBuffer) < 0);
```
ptrivedi, line 3229: nit: config.cpp also seems to have a uname syscall being made to get the release. We should try to share the result instead of making another syscall, even though it is a cheaper syscall

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -407,6 +447,17 @@ void WslCoreVm::Initialize(const GUID& VmId, const wil::shared_handle& UserToken
...
     }
 
+    // Add a read-only 9p share for the kernel headers directory. Mini-init mounts this share at a
+    // temporary path and hands it off to distro init, which moves it under
+    // /usr/src/linux-headers-$(uname -r)/include and creates the /lib/modules/$(uname -r)/build
+    // symlink expected by out-of-tree module tooling.
+    if (m_mountKernelHeaders)
+    {
+        constexpr auto flags = (hcs::Plan9ShareFlags::ReadOnly | hcs::Plan9ShareFlags::AllowOptions);
+        wsl::windows::common::hcs::AddPlan9Share(
```
ptrivedi, line 469: why not use the same addShare lambda that is being used above for GPU shares?

ptrivedi:
A few minor comments, LGTM otherwise

ptrivedi (dismissed):
a couple of minor comments, but will approve

benhillis:
Waiting on this one until we bundle the headers in the modules vhd.

benhillis:
I think we should hold off until the headers are in a single VHD. We are working on this (and also adding support for the perf binary).

PR #40552, opened 2026-05-15
wslc: align docker_schema with bundled dockerd v25.0.3 (API v1.44)

benhillis:
Follow-up to #40550. Audits the rest of `docker_schema.h` against the moby version we actually ship — **dockerd v25.0.3**, Docker API **v1.44** (verified via the version banner in the dockerd ETL trace from the linked test failure). Authoritative source: <https://github.com/moby/moby/tree/v25.0.3/api>.

None of these are observable user-facing bugs on a stock Docker daemon today, but each is a latent crash or wrong-behavior path waiting on the right wire payload from a third-party driver, future daemon, or TTY exec.

## Schema fixes

| # | Field | Before | After | Why |
|---|---|---|---|---|
| 1 | `ContainerState` enum | `Created` was the first enum entry, so `NLOHMANN_JSON_SERIALIZE_ENUM` mapped any unrecognized state string to `Created` | `Unknown` is first, mapped to `nullptr` | Daemon can introduce new states (e.g. `configured` for swarm). Old code silently labels them `Created`. |
| 2 | `HostConfig.ShmSize` | `std::optional<ULONGLONG>` | `std::int64_t` | Wire type is signed `int64`; `0` already means "use daemon default". The `optional` added nothing. |
| 3 | `Volume.Status` | `optional<map<string, string>>` | `optional<map<string, nlohmann::json>>` | Docker's wire schema is `map[string]any`. Third-party volume drivers may publish numbers, bools, or nested objects — currently throws `type_error`. |
| 4 | `Image.Size`, `InspectImage.Size` | `uint64_t` | `int64_t` | Daemon serializes `int64`. |
| 5 | `CreateExec` / `StartExec` `ConsoleSize` | Macro-generated `to_json` that always wrote `"ConsoleSize": []` | Custom `to_json` that omits the field when empty | Docker treats an empty array as an explicit 0×0 console for TTY execs. We never set `ConsoleSize`, so every exec was sending `[]`. |
| 6 | `CreatedContainer.Name` | Field present | Removed | Docker's `POST /containers/create` response is only `{Id, Warnings}`; `Name` is a query parameter, not echoed in the response body. Switched to `WITH_DEFAULT` for parity. |
| 7 | API doc URLs | `v1.52` (aspirational) | `v1.44` + comment about bundled dockerd version | Match the actual shipped daemon. |

## Consumer cascade

`Image.Size` going `uint64_t -> int64_t` requires explicit casts where the value flows back to ULONGLONG ABI fields:
- `WslcSDK\wslcsdk.cpp:1480`
- `wslc\services\ImageService.cpp:207`
- `wslcsession\WSLCSession.cpp:112, 1314, 1340`

`HostConfig.ShmSize` becoming `int64_t` simplifies the consumer at `WSLCContainer.cpp:1419` (drops the `> 0` guard, just direct assignment with cast).

## Not in scope

- **InspectExec polling-loop tightening**: deferred until #40550 lands. The change only becomes safe once `state.Pid` is no longer always `has_value() == true`, otherwise we'd race-bug on the transient `Running=false, ExitCode=null` state.
- **moby v28+ schema changes** (e.g., `ExitCode` becoming non-nullable): not exercised today; left for a future PR after the daemon bumps.

## Verification

- `clang-format --dry-run --Werror` passes on all modified files.
- Each consumer of every changed field was inspected by hand. No code reads `Volume.Status`, `CreatedContainer.Name`, or any `ConsoleSize` field; the rest are listed above.
- Cross-checked against `https://raw.githubusercontent.com/moby/moby/v25.0.3/api/types/...` to confirm wire types.

### DIFF
--- a/src/windows/WslcSDK/wslcsdk.cpp
+++ b/src/windows/WslcSDK/wslcsdk.cpp
@@ -1477,7 +1477,7 @@ try
                 memcpy_s(currentResult.name, sizeof(decltype(WslcImageInfo::name)), currentImage.Image, sizeof(decltype(WSLCImageInformation::Image))) !=
                     0);
             ConvertSHA256Hash(currentImage.Hash, currentResult.sha256);
-            currentResult.sizeBytes = currentImage.Size;
+            currentResult.sizeBytes = static_cast<uint64_t>(currentImage.Size);
             currentResult.createdUnixTime = currentImage.Created;
         }
 
--- a/src/windows/WslcSDK/wslcsdk.cpp
+++ b/src/windows/WslcSDK/wslcsdk.cpp
@@ -1477,7 +1477,7 @@ try
                 memcpy_s(currentResult.name, sizeof(decltype(WslcImageInfo::name)), currentImage.Image, sizeof(decltype(WSLCImageInformation::Image))) !=
                     0);
             ConvertSHA256Hash(currentImage.Hash, currentResult.sha256);
-            currentResult.sizeBytes = static_cast<uint64_t>(currentImage.Size);
+            currentResult.sizeBytes = currentImage.Size;
             currentResult.createdUnixTime = currentImage.Created;
         }
 
--- a/src/windows/inc/docker_schema.h
+++ b/src/windows/inc/docker_schema.h
@@ -9,7 +9,8 @@ Module Name:
 Abstract:
 
     JSON schema for the docker API.
-    The documentation for the API can be found at: https://docs.docker.com/reference/api/engine/version/v1.52/#tag/Container
+    Targets the daemon API version bundled with WSLC's dockerd (currently v25.0.3, API v1.44).
+    The documentation for the API can be found at: https://docs.docker.com/reference/api/engine/version/v1.44/#tag/Container
 
 --*/
 
@@ -22,10 +23,9 @@ namespace wsl::windows::common::docker_schema {
 struct CreatedContainer
 {
     std::string Id;
-    std::string Name;
     std::vector<std::string> Warnings;
 
-    NLOHMANN_DEFINE_TYPE_INTRUSIVE(CreatedContainer, Id, Warnings);
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(CreatedContainer, Id, Warnings);
 };
 
 struct ErrorResponse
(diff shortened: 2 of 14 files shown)


src/windows/wslc/services/ImageService.cpp
```diff
@@ -204,7 +204,7 @@ std::vector<ImageInformation> ImageService::List(wsl::windows::wslc::models::Ses
 
         info.Id = image.Hash;
         info.Created = image.Created;
-        info.Size = image.Size;
```
benhillis, line 207: Should probably update type instead of cast

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -1477,7 +1477,7 @@ try
                 memcpy_s(currentResult.name, sizeof(decltype(WslcImageInfo::name)), currentImage.Image, sizeof(decltype(WSLCImageInformation::Image))) !=
                     0);
             ConvertSHA256Hash(currentImage.Hash, currentResult.sha256);
-            currentResult.sizeBytes = currentImage.Size;
+            currentResult.sizeBytes = static_cast<uint64_t>(currentImage.Size);
```
benhillis: Update type?

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1413,10 +1413,7 @@ std::unique_ptr<WSLCContainerImpl> WSLCContainerImpl::Create(
         request.HostConfig.Ulimits = std::move(ulimits);
     }
 
-    if (containerOptions.ShmSize > 0)
-    {
-        request.HostConfig.ShmSize = containerOptions.ShmSize;
-    }
+    request.HostConfig.ShmSize = static_cast<int64_t>(containerOptions.ShmSize);
```
benhillis: Update type

src/windows/inc/docker_schema.h
```diff
@@ -580,7 +617,7 @@ struct CreateImageProgress
 };
 
 // Container stats (GET /containers/{id}/stats?stream=false)
-// See: https://docs.docker.com/reference/api/engine/version/v1.52/#tag/Container/operation/ContainerStats
+// See: https://docs.docker.com/reference/api/engine/version/v1.44/#tag/Container/operation/ContainerStats
```
dkbennett, line 620: The title clearly says this is intended to be v1.44, but it also appears to be a downgrade from 1.52? Assuming tests will catch most regressions here, but seems like this might be a risky change if test coverage has a gap.

dkbennett (approved):

src/windows/inc/docker_schema.h
dkbennett: Nevermind this is just for Container Stats, which I just added and has good test coverage so should be fine.

benhillis: This schema comment was written before we switched (downgraded) our docker version to the version that's part of azure linux 3

PR #40558, opened 2026-05-15
wslc service com api support for emit warning message as optional com callback

yao-msft:
Added IWarningCallback to COM api that may emit warning messages. Scanned through wslcsession to add warning messages as applicable.

## PR Checklist

## Validation Steps Performed
Added 2 tests that test the warning callback where we can create a warning situation deterministically. Other places we do not have a deterministic way to get to the warning state.


src/windows/common/ExecutionContext.h
```diff
@@ -215,6 +215,18 @@ class COMServiceExecutionContext : public ExecutionContext
...
+    // Set a process-wide warnings pipe for streaming warnings to the CLI.
+    // The pipe is owned by the caller and must outlive all COMServiceExecutionContext instances.
+    // Must be called before any COM methods are serviced (e.g., during Initialize).
+    static void SetWarningsPipe(HANDLE pipe);
+
+protected:
+    bool CollectUserWarning(const std::wstring& warning) override;
+
+private:
+    static std::atomic<HANDLE> s_warningsPipe;
```
OneBlue: Unfortunately I don't think a global handle will work here (long term, we should add support for warnings when creating a container, and in that code paths, multiple containers can start at the same time), so I would recommend making the warning pipe / COM pointer a regular field, so only the thread that sets it can use it

OneBlue (commented):
I like this structure because it will nicely integrate with the EMIT_USER_WARNING() macros that we have.

Sadly there's one tricky that we need to consider: If the caller doesn't read from the handle, we need to be able to Cancel the IO so that we don't get stuck when trying to terminate a session if the caller doesn't read from the HANDLE and the pipe gets full. 

([This is how we do it on the regular session handles](https://github.com/microsoft/WSL/blob/d4359f1fdf0c0a43d1b9859c6467bd31fd353858/src/windows/wslcsession/WSLCSession.cpp#L2691-L2692) 

The issue would be the same if we decided to use a COM callback, we would need to keep track of the callback, and cancel it if the session terminates ([pointer to how we terminate COM callbacks today](https://github.com/microsoft/WSL/blob/d4359f1fdf0c0a43d1b9859c6467bd31fd353858/src/windows/wslcsession/WSLCSession.cpp#L2731) 

One way we could make this work would be by adding a pointer to the WSLCSession to `COMServiceExecutionContext`, and do something like this: 

```
HRESULT WSLCSession::CreateContainer(const WSLCContainerOptions* containerOptions, IWSLCContainer** Container)
{
    COMServiceExecutionContext context(this);
[...]
}
```

So that way, we can register the COM callback / IO with the session when we emit a warning, and so it will be correctly cancelled if the session terminates. 


Regarding the pipe vs COM callback, I think I have a small preference for the COM callback, since we already do that in other places (like for the pull / build callbacks), and at least it won't force the caller to create a thread if no warning are emitted

src/windows/wslcsession/WSLCExecutionContext.h
```diff
@@ -0,0 +1,57 @@
...
+        m_session(session), m_warningCallback(warningCallback)
+    {
+    }
+
+    ~WSLCExecutionContext() override = default;
+
+protected:
+    bool CollectUserWarning(const std::wstring& warning) override
+    {
+        if (m_warningCallback != nullptr)
```
OneBlue, line 30: Could we register the callback at the method level ? We're unlikely to emit multiple warnings in the same execution context so that could simplify things a bit

OneBlue: un-registering and re-registering later wouldn't be a big deal anyway

OneBlue (commented):
This approach looks great and integrate nicely into our existing warning infrastructure. 

I think the next step would be to add logic to print the warnings on stderr in wslc, and this should be ready to merge

src/windows/wslc/services/WarningCallback.h
```diff
@@ -0,0 +1,25 @@
...
+
+namespace wsl::windows::wslc::services {
+
+class DECLSPEC_UUID("A7E3F8B2-4D19-4C6A-9E5B-8F2A1D3C7E90") WarningCallback
+    : public Microsoft::WRL::RuntimeClass<Microsoft::WRL::RuntimeClassFlags<Microsoft::WRL::ClassicCom>, IWarningCallback, IFastRundown>
+{
+public:
+    HRESULT OnWarning(LPCWSTR Message) override
+    {
+        if (Message)
```
OneBlue: Can message ever be null ? If not, I'd just add a WI_ASSERT here

src/windows/wslc/services/WarningCallback.h
```diff
@@ -0,0 +1,25 @@
...
+        {
+            wsl::windows::common::wslutil::PrintMessage(Message, stderr);
```
OneBlue: I recommend prefix the message with `wsl: ` like we do for WSL warnings so it' clear that the warning comes from us and not from the user's command

OneBlue (dismissed):
LGTM. Couple minor comments. 

Could we also add one test that validates that warnings are wired in wslc.exe ? We could create a broken container metadata and then run "wslc ps" and validate that the warning is printed on stderr

src/windows/wslc/services/WarningCallback.h
yao-msft: wsl: is already automatically injected in in ExecutionContext::EmitUserWarning

yao-msft:
> LGTM. Couple minor comments.
> 
> Could we also add one test that validates that warnings are wired in wslc.exe ? We could create a broken container metadata and then run "wslc ps" and validate that the warning is printed on stderr

added wslce2e test

test/windows/wslc/e2e/WSLCE2EWarningTests.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+        // Run the CLI: recovery of the corrupt container fails and the warning is printed on stderr.
+        auto result = RunWslc(L"container list");
+        VERIFY_IS_TRUE(result.ExitCode.has_value());
+        VERIFY_ARE_EQUAL(0u, result.ExitCode.value());
+        VERIFY_IS_TRUE(result.Stderr.has_value());
+
+        const auto expectedWarning =
+            wsl::shared::Localization::MessageWslcFailedToRecoverContainer(string::MultiByteToWide(corruptContainerId));
+        VERIFY_ARE_NOT_EQUAL(std::wstring::npos, result.Stderr.value().find(expectedWarning));
```
OneBlue: nit: I recommend validating the exact warning so we know that we don't have unexpected chars in there

test/windows/wslc/e2e/WSLCE2EWarningTests.cpp
```diff
@@ -0,0 +1,113 @@
...
+                {"/usr/bin/docker",
+                 "create",
+                 "--label",
+                 "wslc.container.metadata=INVALID_JSON",
+                 string::WideToMultiByte(AlpineImage.NameAndTag())});
+            VERIFY_IS_FALSE(corruptContainerId.empty());
+        }
+
+        // cleanup: remove the corrupt container
+        auto cleanup = wil::scope_exit([&]() { RemoveContainerNoThrow(corruptContainerId); });
```
OneBlue, line 96: nit: We could pass `session` to RemoveContainerNoThrow so we don't need to have all the logic to open a session there

OneBlue (dismissed):
LGTM. One minor test comment

test/windows/wslc/e2e/WSLCE2EWarningTests.cpp
yao-msft: Nice catch, actually I found 2 \r\n being printed

test/windows/wslc/e2e/WSLCE2EWarningTests.cpp
yao-msft: There's code below to terminate the default session so the next wslc command recreates it and runs recovery. So we have to open it again internally for cleanup.

OneBlue: Good point! We should still be able to reuse OpenDefaultElevatedSession() to open it though

OneBlue (approved):

test/windows/wslc/e2e/WSLCE2EWarningTests.cpp
yao-msft: OpenDefaultElevatedSession() expects a clean session, there's code to verify no errors. that's why we had this code logic to open a session with corrupt data then clean it up.

PR #40582, opened 2026-05-18
Build the C# projection for AnyCPU

florelis:

The C# projection binary should be built for AnyCPU. I intended for it to do that by not setting the platform explicitly and relying on the default, but that still built for a specific platform due to the global settings. This sets the target to AnyCPU explicitly.

## PR Checklist

### DIFF
--- a/src/windows/WslcSDK/csharp/CMakeLists.txt
+++ b/src/windows/WslcSDK/csharp/CMakeLists.txt
@@ -22,6 +22,9 @@ set_target_properties(
   wslcsdkcs PROPERTIES
   FOLDER windows
   LINKER_LANGUAGE CSharp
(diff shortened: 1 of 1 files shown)


JohnMcPMS (approved):
Is this a shotgun approach or do we need to set all of these values?

florelis:
> Is this a shotgun approach or do we need to set all of these values?

Shotgun approach. It matches what [wslsettings](https://github.com/microsoft/WSL/blob/master/src/windows/wslsettings/CMakeLists.txt) does. It's not immediately clear to me which of these are set by the VS generator from `${TARGET_PLATFORM}` or `${CMAKE_GENERATOR_PLATFORM}`, so it's just easier to set all.

PR #40583, opened 2026-05-18
Add CDI for WSLC GPU

kvega005:
This pull request implements a CDI spec for the WSL GPU device, adds a runtime hook for container setup, and updates the Docker integration to request GPU resources via CDI. The changes also include schema additions, refactoring, and updates to tests to align with the new approach.

**Container Device Interface (CDI) Integration for GPU:**

- Added CDI schema definitions in `cdi_schema.h` and logic in `WSLCInit.cpp` to generate a CDI spec for the WSL GPU device, including device nodes, mounts, and a runtime hook (`wsl-gpu-hook`). This spec is written to `/etc/cdi/microsoft.com-wslc.json` for use by container runtimes.

- Implemented the `wsl-gpu-hook` entrypoint and registration, which configures the container's dynamic linker (`ld.so.conf.d/ld.wsl.conf` and `ld.so.cache`) to ensure GPU libraries are discoverable inside the container.

**Docker Integration and Schema Updates:**

- Updated the Docker host config schema to support `DeviceRequests`, enabling containers to request the WSL GPU device via CDI rather than through direct device and bind mounts.

**Refactoring and Removal of Legacy GPU Setup:**

- Removed previous logic that manually set `LD_LIBRARY_PATH` and bind-mounted GPU libraries/drivers, as this is now handled by the CDI spec and runtime hook.

**Test Updates:**

- Updated tests to validate the new CDI-based GPU setup, checking for the presence of the `ld.so.conf` fragment and correct linker cache population, instead of checking `LD_LIBRARY_PATH` and manual mounts. [[1]](diffhunk://#diff-26c08ea0432c516d56db85fca4dad800c9f8992d54affb2a56cc66de906fcc1eL3382-L3392) [[2]](diffhunk://#diff-26c08ea0432c516d56db85fca4dad800c9f8992d54affb2a56cc66de906fcc1eL3421-R3413)
## PR Checklist


src/linux/init/WSLCInit.cpp
```diff
@@ -56,25 +58,117 @@ extern int g_LogFd;
 
 extern void WSLCEnableCrashDumpCollection();
 
+#define WSLC_GPU_LIB_PATH "/usr/lib/wsl/lib"
+#define WSLC_GPU_DRIVERS_PATH "/usr/lib/wsl/drivers"
```
benhillis: these defines exist elsewhere

src/linux/init/WSLCInit.cpp
```diff
@@ -56,25 +59,134 @@ extern int g_LogFd;
 
 extern void WSLCEnableCrashDumpCollection();
 
+#define WSLC_GPU_DRIVERS_PATH LXSS_LIB_PREFIX "/drivers"
```
benhillis: Let's add another LXSS_GPU_DRIVER_PATH #define, since with this addition it will be used twice.

src/linux/init/WSLCInit.cpp
```diff
@@ -56,25 +59,134 @@ extern int g_LogFd;
...
+
 struct WSLCState
 {
     std::optional<std::filesystem::path> ModulesMountPoint;
 };
 
 static WSLCState g_state;
 
-int CreateCaptureCrashSymlink()
+int CreateSymlink(PCSTR LinkName)
```
benhillis: I don't like the name of this method, way too generic. Do we even need a helper?

src/linux/init/WSLCInit.cpp
```diff
@@ -56,25 +59,134 @@ extern int g_LogFd;
...
+{
+    THROW_LAST_ERROR_IF(symlink("/init", LinkName) < 0 && errno != EEXIST);
+    return 0;
+}
+CATCH_RETURN_ERRNO()
+
+void WriteWslcCdiSpec()
+try
+{
+    wsl::shared::cdi::DeviceNode dxg;
```
benhillis: nit, default init {};

benhillis: this and all structs in this method

src/linux/init/WSLCInit.cpp
```diff
@@ -56,25 +59,134 @@ extern int g_LogFd;
...
+    gpu.containerEdits.hooks.push_back(std::move(hook));
+
+    wsl::shared::cdi::Spec spec;
+    spec.cdiVersion = "0.6.0";
+    spec.kind = LX_WSLC_CDI_KIND;
+    spec.devices.push_back(std::move(gpu));
+
+    THROW_LAST_ERROR_IF(UtilMkdirPath("/etc/cdi", 0755) < 0);
+    THROW_LAST_ERROR_IF(WriteToFile("/etc/cdi/microsoft.com-wslc.json", nlohmann::json(spec).dump().c_str()) < 0);
+}
```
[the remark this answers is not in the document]
benhillis: these look valid.

src/linux/init/WSLCInit.cpp
```diff
@@ -56,25 +59,134 @@ extern int g_LogFd;
...
+                config = std::move(existing);
+            }
+        }
+        CATCH_LOG()
+    }
+
+    config["features"]["cdi"] = true;
+
+    THROW_LAST_ERROR_IF(UtilMkdirPath("/etc/docker", 0755) < 0);
+    THROW_LAST_ERROR_IF(WriteToFile(c_daemonConfigPath, config.dump().c_str()) < 0);
```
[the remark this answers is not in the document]
benhillis: I wouldn't write this merging code yet if it isn't needed. Just fail loudly so we know when we update to a system distro vhd that has the unexpected file.

src/linux/init/WSLCInit.cpp
```diff
@@ -56,25 +59,134 @@ extern int g_LogFd;
...
+
+int WslcGpuHookEntry()
 try
 {
-    THROW_LAST_ERROR_IF(symlink("/init", "/" LX_INIT_WSL_CAPTURE_CRASH) < 0);
+    // OCI runtime hooks receive the container state as JSON on stdin.
+    std::string stateJson;
+    char buf[4096];
+    ssize_t n;
+    while ((n = TEMP_FAILURE_RETRY(read(STDIN_FILENO, buf, sizeof(buf)))) > 0)
```
benhillis: this is kind of gross, do we not have a helper that does this?

src/linux/init/WSLCInit.cpp
```diff
@@ -56,25 +59,134 @@ extern int g_LogFd;
...
+    // Read the OCI spec's root.path from <bundle>/config.json. For Docker/containerd this is an
+    // absolute path to the overlay-merged rootfs accessible from the runtime namespace.
+    const auto spec = nlohmann::json::parse(UtilReadFileContent(bundle + "/config.json"));
+    std::string rootfs = spec.at("root").at("path").get<std::string>();
+    if (!rootfs.empty() && rootfs.front() != '/')
+    {
+        rootfs = bundle + "/" + rootfs;
+    }
+
+    const std::string confDir = rootfs + "/etc/ld.so.conf.d";
```
benhillis: std::filesystem for paths

src/linux/init/WSLCInit.cpp
```diff
@@ -56,25 +59,134 @@ extern int g_LogFd;
...
+    const std::string confPath = confDir + "/ld.wsl.conf";
+
+    THROW_LAST_ERROR_IF(UtilMkdirPath(confDir.c_str(), 0755) < 0);
+    THROW_LAST_ERROR_IF(WriteToFile(confPath.c_str(), LXSS_LIB_PATH "\n") < 0);
+
+    // chroots into the rootfs and uses the container's own /etc/ld.so.conf chain,
+    // writing /etc/ld.so.cache inside the container.
+    const char* const ldArgv[] = {LDCONFIG_COMMAND, "-r", rootfs.c_str(), nullptr};
+    THROW_LAST_ERROR_IF(UtilCreateProcessAndWait(ldArgv[0], ldArgv) < 0);
```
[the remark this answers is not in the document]
benhillis: seems legit.

src/linux/init/WSLCInit.cpp
```diff
@@ -620,7 +732,12 @@ void HandleMessageImpl(
             THROW_LAST_ERROR_IF(Chroot(target) < 0);
 
             // Recreate the crash dump symlink inside the new root.
-            CreateCaptureCrashSymlink();
+            CreateSymlink("/" LX_INIT_WSL_CAPTURE_CRASH);
```
benhillis: you are ignoring failures, if you fix my above comment about just using symlink directly it will be better.

benhillis (dismissed):

src/linux/init/WSLCInit.cpp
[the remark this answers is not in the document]
kvega005: For Docker/containerd, the bundle is something like `/run/containerd/io.containerd.runtime.v2.task/moby/<id>/` and contains just `config.json` plus a small rootfs mount stub, but the actual merged rootfs lives at `/var/lib/docker/overlay2/<id>/merged/`, outside the bundle.

The OCI runtime spec explicitly allows `root.path` to be either:

- relative to the bundle (typical for runc run standalone bundles), or
- absolute (typical for runtimes like Docker/containerd that pre-mount the overlay).

So enforcing "rootfs is in bundle" would break the Docker case entirely.

If we wanted to tighten further, the only realistic additional check would be an allowlist of known-safe parent prefixes (e.g. `/var/lib/docker`, `/var/lib/containerd`, `/run/containerd`, plus the bundle itself for standalone runc), but that's a bit brittle.

src/linux/init/util.h
```diff
@@ -313,6 +313,6 @@ std::string UtilReadFileContent(std::string_view path);
 
 uint16_t UtilWinAfToLinuxAf(uint16_t AddressFamily);
 
-int WriteToFile(const char* Path, const char* Content, int permissions = 0644);
+int WriteToFile(const char* Path, const char* Content, int OpenFlags = O_WRONLY | O_CLOEXEC | O_CREAT, int Permissions = 0644);
```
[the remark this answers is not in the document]
kvega005: No call sites pass 0644 (or any octal mode) as the third argument. The only match is the default parameter in the declaration in `util.h`

src/linux/init/WSLCInit.cpp
```diff
@@ -63,19 +67,105 @@ struct WSLCState
...
+    rootfsPath = std::filesystem::canonical(rootfsPath);
+    THROW_ERRNO_IF(EINVAL, rootfsPath == "/");
+    THROW_ERRNO_IF(ENOTDIR, !std::filesystem::is_directory(rootfsPath));
+
+    const auto confDir = rootfsPath / "etc/ld.so.conf.d";
+    const auto confPath = confDir / "ld.wsl.conf";
+
+    THROW_LAST_ERROR_IF(UtilMkdirPath(confDir.c_str(), 0755) < 0);
+    THROW_LAST_ERROR_IF(WriteToFile(confPath.c_str(), LXSS_LIB_PATH "\n", O_WRONLY | O_CLOEXEC | O_CREAT | O_TRUNC | O_NOFOLLOW) < 0);
+
```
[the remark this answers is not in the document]
kvega005: The trust boundary is "we trust the container image's filesystem layout for the GPU hook", which is consistent with how `ldconfig -r` itself behaves (it also walks `/etc/ld.so.conf` chains via normal path resolution).

src/linux/init/util.h
```diff
@@ -313,6 +313,6 @@ std::string UtilReadFileContent(std::string_view path);
...
```
benhillis, line 316: I know why you did things in this order (to match real open argument order), but we should be careful that any callers that were previously passing optional vales for permission are not now passing them as OpenFlags.

benhillis: I assume you did an audit of everybody calling WriteToFile?

benhillis: Maybe safest to just remove the default parameters from the def and chase down compiler errors to make sure they are all right.

src/linux/init/WSLCInit.cpp
```diff
@@ -63,19 +67,105 @@ struct WSLCState
...
+    // Read the OCI spec's root.path from <bundle>/config.json. This is either an absolute path to
+    // the overlay-merged rootfs or a path relative to the bundle directory.
+    const auto spec = nlohmann::json::parse(UtilReadFileContent((bundle / "config.json").native()));
+    std::filesystem::path rootfsPath = spec.at("root").at("path").get<std::string>();
+    if (rootfsPath.is_relative())
+    {
+        rootfsPath = bundle / rootfsPath;
+    }
+
+    rootfsPath = std::filesystem::canonical(rootfsPath);
```
benhillis, line 143: this is fine, but is there not a canonicalize that modifies the existing path?

benhillis (commented):
Looking close, just please make sure none of the write helpers are accidentally passing the wrong thing now.

src/linux/init/WSLCInit.cpp
kvega005: I think canonicalization is only provided as non-member functions that return a new path. There is no `path::canonical` method.

src/linux/init/util.h
kvega005: Yes, I checked. There were actually no callers using the third (permissions) parameter at all. We can actually remove that entirely if you want.

src/linux/init/WSLCInit.cpp
```diff
@@ -64,19 +68,105 @@ struct WSLCState
...
+    THROW_LAST_ERROR_IF(UtilMkdirPath("/etc/docker", 0755) < 0);
+    THROW_LAST_ERROR_IF(WriteToFile(c_daemonConfigPath, config.dump().c_str(), O_WRONLY | O_CLOEXEC | O_CREAT | O_TRUNC) < 0);
+}
+CATCH_LOG()
+
+int WslcGpuHookEntry()
+try
+{
+    // OCI runtime hooks receive the container state as JSON on stdin.
+    const std::string stateJson{std::istreambuf_iterator<char>(std::cin), {}};
```
OneBlue: nit: This could be simplified to: 

```
const auto state = nlohmann::json::parse(std::cin)
```

src/linux/init/WSLCInit.cpp
```diff
@@ -64,19 +68,105 @@ struct WSLCState
...
+
+    rootfsPath = std::filesystem::canonical(rootfsPath);
+    THROW_ERRNO_IF(EINVAL, rootfsPath == "/");
+    THROW_ERRNO_IF(ENOTDIR, !std::filesystem::is_directory(rootfsPath));
+
+    const auto confDir = rootfsPath / "etc/ld.so.conf.d";
+    const auto confPath = confDir / "ld.wsl.conf";
+
+    THROW_LAST_ERROR_IF(UtilMkdirPath(confDir.c_str(), 0755) < 0);
+    THROW_LAST_ERROR_IF(WriteToFile(confPath.c_str(), LXSS_LIB_PATH "\n", O_WRONLY | O_CLOEXEC | O_CREAT | O_TRUNC | O_NOFOLLOW) < 0);
```
OneBlue: nit: Instead of setting `O_NOFOLLOW` here, I would recommend chrooting into `rootfsPath` and executing the container's `ldconfig` inside the chroot. 

That way we have the guarantee that symlinks won't escape the container (the best way to do the chroot would be to move main.cpp's `Chroot()` to util.cpp and call it from here

src/linux/init/WSLCInit.cpp
```diff
@@ -64,19 +68,105 @@ struct WSLCState
...
+try
+{
+    constexpr auto c_daemonConfigPath = "/etc/docker/daemon.json";
+
+    THROW_ERRNO_IF(EEXIST, std::filesystem::exists(c_daemonConfigPath));
+
+    nlohmann::json config = nlohmann::json::object();
+    config["features"]["cdi"] = true;
+
+    THROW_LAST_ERROR_IF(UtilMkdirPath("/etc/docker", 0755) < 0);
```
OneBlue, line 121: not a blocker for this PR, but I think we should write this file (and the CDI config file) in the system VHD so we don't need to rewrite them at every boot

src/linux/init/WSLCInit.cpp
OneBlue: (That should also protect us from mismatches between our ldconfig and the container's ldconfig)

OneBlue (commented):
Change looks great. Minor comments

src/linux/init/WSLCInit.cpp
kvega005: My only concern with that is what happens if we want to update the CDI or Docker config. We can't just trust that if the file exists, it has the right content. This means we would still have to ensure it has the appropriate settings for our current version of WSLC. How is this handled for the WSL init?

Certainly, worth thinking about. But it's not something that I wanted to deal with in this PR.

One thing I considered is shipping these config files as part of the WSLC package. Maybe as something that just lives in the WSLC installation and gets mapped into each storage, so we don't need a bunch of copies of the file. But then we'd have to consider the performance cost of that too.

OneBlue: > This means we would still have to ensure it has the appropriate settings for our current version of WSLC. How is this handled for the WSL init?

The system VHD ships with the package, so we have the guarantee that there's always a match. The only downside of having those directly inside the VHD would be that every change to these files requires a building a new VHD, but I don't expect those to change very often

OneBlue (dismissed):

src/linux/init/WSLCInit.cpp
benhillis: I discussed this with Kevin as well, I think this is fine for now, seems like we are always going to have to write at least one of these files though because it depends on feature bits being set?

src/linux/init/WSLCInit.cpp
```diff
@@ -64,19 +68,99 @@ struct WSLCState
...
+    {
+        rootfsPath = bundle / rootfsPath;
+    }
+
+    rootfsPath = std::filesystem::canonical(rootfsPath);
+    THROW_ERRNO_IF(EINVAL, rootfsPath == "/");
+
+    THROW_LAST_ERROR_IF(chroot(rootfsPath.c_str()) < 0);
+
+    THROW_LAST_ERROR_IF(UtilMkdirPath("/etc/ld.so.conf.d", 0755) < 0);
```
benhillis, line 148: My recollection is that ldconfig does not work for some distros. For example, I think this will not work for alpine?

I think that's ok for now as we have the same limitation for WSL2, but might be worth investigating as a follow-up.

benhillis (approved):
LGTM - thanks for doing this.  One follow-up non blocking.

PR #40586, opened 2026-05-19
Fix _locComment syntax in en-US WSL.adml

benhillis:
## Summary
Fix the `_locComment` syntax in `intune/en-US/WSL.adml` so the Touchdown/lsbuild localization pipeline correctly parses the `{Locked="..."}` rules.

## PR Checklist
- [x] Tests added/passed (n/a – localization metadata only)
- [x] Documentation updated (n/a)
- [x] Schema updated (n/a)

## Detailed Description
Per feedback from the localization team, the previous inline syntax was being parsed incorrectly by lsbuild. The produced `.lcg` intermediary contained malformed Dev comments – e.g. `Unbalanced closing '}' rule delimiter at position 57` warnings and truncated `{Locked=` rules – so the locked-token rules were not being honored.

The correct directive requires the `_text` suffix:

- Before: `<!-- _locComment='{Locked="WSL"}' -->`
- After:  `<!-- _locComment_text='{Locked="WSL"}' -->`

This PR renames every `_locComment=` to `_locComment_text=` in the en-US baseline (the only source the loc pipeline reads — translated locales will be regenerated by the next loc cycle).

It also adds the missing lock comments to `<displayName>` and `<description>` at the top of the file. These were previously emitted with no locking applied (verified non-locked in the produced `.lcg`):

- `<displayName>` → `{Locked="WSL"}`
- `<description>` → `{Locked="Windows Subsystem for Linux"}`

## Validation Steps
- `python3 tools/devops/validate-localization.py` passes (exit 0).
- Visual inspection of the resulting `.lcg` should now show well-formed `{Locked="..."}` Dev comments for every string and for the new `displayName`/`description` entries.

### DIFF
--- a/intune/en-US/WSL.adml
+++ b/intune/en-US/WSL.adml
@@ -1,62 +1,62 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--  (c) 2006 Microsoft Corporation  -->
 <policyDefinitionResources xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" revision="0.1" schemaVersion="0.1" xmlns="http://schemas.microsoft.com/GroupPolicy/2006/07/PolicyDefinitions">
-  <displayName>WSL</displayName>
-  <description>Windows Subsystem for Linux</description>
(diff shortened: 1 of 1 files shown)


benhillis:
nice, this is working https://github.com/microsoft/WSL/pull/40588

JohnMcPMS (approved):

PR #40596, opened 2026-05-19
Improved details for build and pull

craigloewen-msft:

Added some small details to `wslc build` and `wslc pull`. 

Specifically this just adds a 'total MB' to build and progress bars and total layer sizes to pull. 

### Pull

Before:
image

After:
image

### Build

Before:
image

After:
image

## PR Checklist

Scoped it to be as small as possible. 

## Validation Steps Performed

Tested behaviour before and after.

### DIFF
--- a/src/shared/inc/stringshared.h
+++ b/src/shared/inc/stringshared.h
@@ -825,6 +825,30 @@ struct CaseInsensitiveCompare
     }
 };
 
+inline std::wstring FormatBytes(uint64_t bytes)
+{
+    constexpr double c_kB = 1024.0;
+    constexpr double c_MB = 1024.0 * 1024.0;
+    constexpr double c_GB = 1024.0 * 1024.0 * 1024.0;
+
+    if (bytes >= static_cast<uint64_t>(c_GB))
+    {
+        return std::format(L"{:.2f} GB", bytes / c_GB);
+    }
+    else if (bytes >= static_cast<uint64_t>(c_MB))
+    {
+        return std::format(L"{:.2f} MB", bytes / c_MB);
+    }
+    else if (bytes >= static_cast<uint64_t>(c_kB))
+    {
+        return std::format(L"{:.2f} KB", bytes / c_kB);
+    }
+    else
+    {
+        return std::format(L"{} B", bytes);
+    }
+}
+
 } // namespace wsl::shared::string
 
 template <>
--- a/src/shared/inc/stringshared.h
+++ b/src/shared/inc/stringshared.h
@@ -827,9 +827,9 @@ struct CaseInsensitiveCompare
 
 inline std::wstring FormatBytes(uint64_t bytes)
 {
-    constexpr double c_kB = 1024.0;
-    constexpr double c_MB = 1024.0 * 1024.0;
-    constexpr double c_GB = 1024.0 * 1024.0 * 1024.0;
+    constexpr double c_kB = 1000.0;
+    constexpr double c_MB = 1000.0 * 1000.0;
+    constexpr double c_GB = 1000.0 * 1000.0 * 1000.0;
 
     if (bytes >= static_cast<uint64_t>(c_GB))
     {
--- a/src/windows/wslc/services/BuildImageCallback.cpp
+++ b/src/windows/wslc/services/BuildImageCallback.cpp
@@ -253,7 +253,7 @@ void BuildImageCallback::Redraw()
     buffer += L"\033[22m\033[?25h";
 
     WriteTerminal(buffer);
-    m_displayedLines = displayCount + static_cast<SHORT>(m_pullLines.size());
+    m_displayedLines = displayCount;
 }
 
 } // namespace wsl::windows::wslc::services
--- a/src/windows/wslc/services/BuildImageCallback.cpp
+++ b/src/windows/wslc/services/BuildImageCallback.cpp
@@ -125,7 +125,7 @@ try
             terminator.insert(terminator.begin(), wide.back());
             wide.pop_back();
         }
-        wide = std::format(L"\033[36m{}\033[0m{}", wide, terminator);
+        wide = std::format(L"\033[92m{}\033[0m{}", wide, terminator);
         WriteTerminal(wide);
         return S_OK;
     }
--- a/src/windows/wslc/services/BuildImageCallback.cpp
+++ b/src/windows/wslc/services/BuildImageCallback.cpp
@@ -64,9 +64,10 @@ void BuildImageCallback::CollapseWindow()
 
     m_lines.clear();
     m_pendingLine.clear();
+    m_pullLines.clear();
 }
 
-HRESULT BuildImageCallback::OnProgress(LPCSTR status, LPCSTR id, ULONGLONG /*current*/, ULONGLONG /*total*/)
+HRESULT BuildImageCallback::OnProgress(LPCSTR status, LPCSTR id, ULONGLONG /*current*/, ULONGLONG total)
 try
 {
     if (status == nullptr || *status == '\0')
@@ -91,11 +92,37 @@ try
     // accepting any non-empty id, so future or unrelated id usage defaults to permanent.
     const bool isLog = (id != nullptr && std::string_view{id} == "log");
 
+    // Pull/download progress: update the per-entry map so Redraw can show each entry
+    // on a single line that updates in place.
+    if (!isLog && id != nullptr && *id != '\0' && total > 0)
+    {
+        m_pullLines[id] = status;
+
+        auto now = std::chrono::steady_clock::now();
+        if (now - m_lastRedraw >= c_redrawInterval)
+        {
+            Redraw();
+            m_lastRedraw = now;
+        }
+
+        return S_OK;
+    }
+
     if (!isLog)
     {
-        // Permanent line: collapse the scrolling window then print directly.
+        // Permanent build-step line: collapse the scrolling window then print in teal.
         CollapseWindow();
-        WriteTerminal(MultiByteToWide(status));
+        auto wide = MultiByteToWide(status);
+        if (!wide.empty() && wide.back() == L'\n')
+        {
+            wide.insert(0, L"\033[36m");
+            wide.insert(wide.size() - 1, L"\033[0m");
+        }
+        else
+        {
+            wide = std::format(L"\033[36m{}\033[0m", wide);
+        }
+        WriteTerminal(wide);
         return S_OK;
     }
 
@@ -217,10 +244,16 @@ void BuildImageCallback::Redraw()
         appendLine(m_pendingLine);
     }
 
+    // Render per-entry pull progress (each entry updates in place via the map).
+    for (const auto& [key, line] : m_pullLines)
+    {
+        appendLine(line);
+    }
+
     buffer += L"\033[22m\033[?25h";
 
     WriteTerminal(buffer);
-    m_displayedLines = displayCount;
+    m_displayedLines = displayCount + static_cast<SHORT>(m_pullLines.size());
 }
 
 } // namespace wsl::windows::wslc::services
--- a/src/windows/wslc/services/BuildImageCallback.cpp
+++ b/src/windows/wslc/services/BuildImageCallback.cpp
@@ -67,7 +67,7 @@ void BuildImageCallback::CollapseWindow()
     m_pullLines.clear();
 }
 
-HRESULT BuildImageCallback::OnProgress(LPCSTR status, LPCSTR id, ULONGLONG /*current*/, ULONGLONG total)
+HRESULT BuildImageCallback::OnProgress(LPCSTR status, LPCSTR id, ULONGLONG current, ULONGLONG total)
 try
 {
     if (status == nullptr || *status == '\0')
@@ -84,6 +84,12 @@ try
 
     if (m_verbose || !m_isConsole)
     {
+        // Skip pull progress updates when output is redirected, show only major steps
+        if (id != nullptr && *id != '\0' && total > 0)
+        {
+            return S_OK;
+        }
+
         wprintf(L"%hs", status);
         return S_OK;
     }
@@ -113,15 +119,13 @@ try
         // Permanent build-step line: collapse the scrolling window then print in teal.
         CollapseWindow();
         auto wide = MultiByteToWide(status);
-        if (!wide.empty() && wide.back() == L'\n')
+        std::wstring terminator;
+        while (!wide.empty() && (wide.back() == L'\n' || wide.back() == L'\r'))
         {
-            wide.insert(0, L"\033[36m");
-            wide.insert(wide.size() - 1, L"\033[0m");
+            terminator.insert(terminator.begin(), wide.back());
+            wide.pop_back();
         }
-        else
-        {
-            wide = std::format(L"\033[36m{}\033[0m", wide);
-        }
+        wide = std::format(L"\033[36m{}\033[0m{}", wide, terminator);
         WriteTerminal(wide);
         return S_OK;
     }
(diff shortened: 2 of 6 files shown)


src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -217,6 +244,12 @@ void BuildImageCallback::Redraw()
         appendLine(m_pendingLine);
     }
 
+    // Render per-entry pull progress (each entry updates in place via the map).
+    for (const auto& [key, line] : m_pullLines)
```
benhillis, line 255: The new pull-progress lines are appended to the output buffer but displayCount (which drives m_displayedLines) is calculated before them. Next time Redraw() fires, the cursor-up escape won't go back far enough, leaving stale ghost lines on screen.Same issue causes CollapseWindow() to incompletely erase when a permanent build-step line arrives.

Fix: add m_pullLines.size() to displayCount after the pull-lines loop, and make sure the total respects c_maxDisplayLines.

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -91,11 +92,37 @@ try
     // accepting any non-empty id, so future or unrelated id usage defaults to permanent.
     const bool isLog = (id != nullptr && std::string_view{id} == "log");
 
+    // Pull/download progress: update the per-entry map so Redraw can show each entry
```
benhillis, line 118: I'm not sure we are handling this correctly when m_isConsole is false. We should probably skip the whole callback stuff when non in an interactive terminal.

craigloewen-msft: I think this would break AIs who are scraping `wslc.exe build` output. Right now we output something like this:

```
PS C:\cDev\copilot\build-test> cat .\dockerfilesimple.log
Building image from directory: .

[1/2] FROM docker.io/library/ubuntu:24.04@sha256:c4a8d5503dfb2a3eb8ab5f807da5bc69a85730fb49b5cfca2330194ebcc41c7b
[2/2] RUN echo 123
exporting to image
  | exporting layers
  | writing image sha256:780c43a5ab95735b080898e63db96c3feebdd05591a0a934bc3252fe70201004
  | naming to docker.io/library/test
PS C:\cDev\copilot\build-test>
```

I think this is pretty reasonable and gives the right info on all the steps.

src/shared/inc/stringshared.h
```diff
@@ -825,6 +825,30 @@ struct CaseInsensitiveCompare
     }
 };
 
+inline std::wstring FormatBytes(uint64_t bytes)
+{
```
kvega005, line 829: I am wondering if we could move this to `wsl::windows::common::string` and use the windows `StrFormatByteSizeEx`. It does have a slightly different oputput. StrFormatByteSizeEx uses 1024-based math with the Windows-conventional labels (KB, MB, GB, TB) — i.e. 1.00 KB means 1024 bytes. The current helper is 1000-based. So switching will change the numbers users see during pulls which would diverge from docker's behavior. However, it does give localized decimal seperator. I could see this going either way. 

ALSO, if we are making a helper for this, we should probably get rid of the other FormatBytes method in ContainerTasks.cpp:37 and just use a common helper.

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -91,11 +98,35 @@ try
...
+        // Permanent build-step line: collapse the scrolling window then print in teal.
         CollapseWindow();
-        WriteTerminal(MultiByteToWide(status));
+        auto wide = MultiByteToWide(status);
+        std::wstring terminator;
+        while (!wide.empty() && (wide.back() == L'\n' || wide.back() == L'\r'))
+        {
+            terminator.insert(terminator.begin(), wide.back());
+            wide.pop_back();
+        }
```
kvega005: This is a bit hard to reason about. I think this can be simplified with `std::wstring::find_last_not_of(L"\r\n")`.
```
    const auto bodyLength = wide.find_last_not_of(L"\r\n") + 1; // or textLength or whatever you prefer (I am bad at naming stuff)
    const auto newlines = wide.substr(bodyLength);
    wide.resize(bodyLength);

    WriteTerminal(std::format(L"\033[36m{}\033[0m{}", wide, newlines));
```

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -91,11 +98,35 @@ try
     // accepting any non-empty id, so future or unrelated id usage defaults to permanent.
     const bool isLog = (id != nullptr && std::string_view{id} == "log");
 
+    // Pull/download progress: update the per-entry map so Redraw can show each entry
+    // on a single line that updates in place.
```
kvega005, line 119: Nit: The OnProgress body asks the same three questions in two different places — "does the message have an id?", "is it the log sentinel?", "is it per-layer pull progress?" — and the `id != nullptr && *id != '\0'` check appears several times. The verbose/non-TTY branch and the interactive branch also independently re-derive what kind of message this is.

```
const std::string_view idView = (id != nullptr) ? id : std::string_view{};
const bool isLog = (idView == "log");
const bool isPullProgress = (!idView.empty() && total > 0 && !isLog);

if (m_verbose || !m_isConsole)
{
    if (!isPullProgress)
    {
        wprintf(L"%hs", status);
    }
    return S_OK;
}

if (isPullProgress) { return HandlePullProgress(idStr, status); }
if (isLog)          { return HandleLogChunk(status); }

return HandleStepHeader(status);
```

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -91,11 +98,35 @@ try
...
         CollapseWindow();
-        WriteTerminal(MultiByteToWide(status));
+        auto wide = MultiByteToWide(status);
+        std::wstring terminator;
+        while (!wide.empty() && (wide.back() == L'\n' || wide.back() == L'\r'))
+        {
+            terminator.insert(terminator.begin(), wide.back());
+            wide.pop_back();
+        }
+        wide = std::format(L"\033[36m{}\033[0m{}", wide, terminator);
```
kvega005: Really nit, don't hate me for this. There are a lot of these ANSI escape sequences scattered throught the file, it would be nice to have to have these defined as constants for readability. Either at the top of this file or a common header. (IDK if its used elsewhere).

For example, this format could read like: `wide = std::format(L"{}{}{}{}", ansi::Cyan, wide, ansi::ResetColor, terminator);`
Then its a lot easier to understand what is happening.

kvega005: Actually, I did a brief search with copilot, and alledgedly its also used in `WSLCTests.cpp`, `WSLCE2EHelpers.h`, `ImageProgressCallback.cpp`, `BuildImageCallback.cpp`. So this could warrant placing in `stringshared.h` or `deps.h`. The test files i mentioned use the VT_* constants.

src/windows/wslc/services/BuildImageCallback.h
```diff
@@ -54,6 +55,8 @@ class DECLSPEC_UUID("3EDD5DBF-CA6C-4CF7-923A-AD94B6A732E5") BuildImageCallback
     std::string m_pendingLine;
     SHORT m_displayedLines = 0;
     std::chrono::steady_clock::time_point m_lastRedraw{};
+    // Per-entry pull progress lines, keyed by entry id. Updated in place by Redraw.
+    std::map<std::string, std::string> m_pullLines;
```
kvega005: The comment doesn't capture the load-bearing detail here: std::map's ordered iteration is what makes the visual "in place" rendering work. Redraw walks the map in key order each frame, so each layer id consistently lands on the same row — swap this for `unordered_map` and entries would shuffle between frames. Worth calling that out.

src/windows/wslc/services/ImageProgressCallback.cpp
```diff
@@ -84,19 +84,62 @@ std::wstring ImageProgressCallback::GenerateStatusLine(LPCSTR status, LPCSTR id,
...
+        std::wstring bar(c_progressBarWidth, L' ');
+        for (int i = 0; i < filled; ++i)
+        {
+            bar[i] = L'=';
+        }
+
+        if (filled < c_progressBarWidth)
+        {
+            bar[filled] = L'>';
+        }
```
kvega005: This pre-fills with spaces and then overwrites characters by index which is fine, but std::wstring already has constructors and append(n, ch) that express this directly without an explicit loop or index arithmetic. The intent reads more clearly when the bar is constructed segment-by-segment:

```
std::wstring bar;
bar.reserve(c_progressBarWidth);
bar.append(filled, L'=');
bar.append(L">");
bar.resize(c_progressBarWidth, L' ');
```

src/shared/inc/stringshared.h
dkbennett: Opportunity to centralize this, such as moving the one from ContainerTasks into stringshared.h

I'm not sure if Craig's PR is the right place to do that consolidation, but at the very least using the same method for it is appropriate.

src/windows/wslc/services/ImageProgressCallback.cpp
```diff
@@ -84,19 +84,62 @@ std::wstring ImageProgressCallback::GenerateStatusLine(LPCSTR status, LPCSTR id,
...
+            line.pop_back();
+        }
     }
 
     // Erase any previously written char on that line.
-    while (line.size() < info.dwSize.X)
+    while (line.size() < static_cast<size_t>(visibleWidth))
     {
         line += L' ';
     }
```
kvega005: nit: `line.resize(visibleWidth, L' ');` - no need for while loop

src/windows/wslc/services/BuildImageCallback.cpp
dkbennett: I was about to make the same comment. I'm not sure this PR should do a lot of refactoring, which could make it a lot bigger than it should be, but at the very least define the VT sequences as constants so the construction is a lot easier to read and so we can do that refactoring more easily later.

dkbennett (commented):
Echo a lot of the great feedback Kevin had. I like that this PR is relatively small though, and if refactoring things would lead to it growing significantly I would recommend not doing that and having a separate PR for that refactoring. But definitely all for things that make the refactoring easier at the very least.

src/windows/wslc/services/ImageProgressCallback.cpp
dkbennett: Recommend to use a std::wostringstream for these constructions rather than doing repeated manipulation of the same string. The stringstreams were designed to do that efficiently. Then convert once to a wstring when all manipulations are complete.

kvega005 (commented):
On a successful build, the dim scrolling window gets erased by the final `CollapseWindow()`, so once the build finishes only the teal step headers (`[1/4] FROM …`, `[2/4] RUN …`) remain in scrollback. The replay branch in the destructor only fires when `std::uncaught_exceptions()` indicates failure.

For comparison, `DOCKER_BUILDKIT=0 docker build` keeps all `RUN` stdout in scrollback interleaved with step headers. So things like `npm WARN …`, `pip` resolver notes, or a `BUILD SUCCESSFUL in 47s` line from a tool inside the container aren't recoverable after a successful wslc build without re-running it.

Not sure if that's intentional (the compact view is genuinely nice) — flagging in case it isn't. A couple of options if you do want to address it:

- Drop the `std::uncaught_exceptions()` check so the log replays on success too. `m_allLines` is already populated and bounded by `c_maxAllLinesBytes`.
- Or add a `--progress=plain` / `--quiet`-style opt-out (could just reuse the existing `m_verbose` path).

craigloewen-msft:
> On a successful build, the dim scrolling window gets erased by the final `CollapseWindow()`, so once the build finishes only the teal step headers (`[1/4] FROM …`, `[2/4] RUN …`) remain in scrollback. The replay branch in the destructor only fires when `std::uncaught_exceptions()` indicates failure.
> 
> For comparison, `DOCKER_BUILDKIT=0 docker build` keeps all `RUN` stdout in scrollback interleaved with step headers. So things like `npm WARN …`, `pip` resolver notes, or a `BUILD SUCCESSFUL in 47s` line from a tool inside the container aren't recoverable after a successful wslc build without re-running it.
> 
> Not sure if that's intentional (the compact view is genuinely nice) — flagging in case it isn't. A couple of options if you do want to address it:
> 
> * Drop the `std::uncaught_exceptions()` check so the log replays on success too. `m_allLines` is already populated and bounded by `c_maxAllLinesBytes`.
> * Or add a `--progress=plain` / `--quiet`-style opt-out (could just reuse the existing `m_verbose` path).

I'm not seeing this behaviour in docker when I do a build. For example here's an output where I run 'apt update' you can see it compacts it. So I'll leave the compacting behaviour :) 

image

src/shared/inc/stringshared.h
craigloewen-msft: I'm trying to leave this PR as small as possible, so I will refrain for now from doing refactoring changes. If it blocks this PR from going in then I can do it!

src/windows/wslc/services/BuildImageCallback.cpp
craigloewen-msft: Good suggestion, but in the interest of leaving this PR as small as possible I will leave it in as magic numbers. Can change that though if it's a blocker.

kvega005: I am ok with either choice. Definitely, pro not touching the test files here that are unrelated to this PR.

However, Ideally, I would like to see at least local constants defined (or at least a comment), so we know what each of the ansi strings mean without having to do google search. It would really just add a couple of lines at the top of the `.cpp` file. Then later we can go back and refactor them all into a common header.

Either way, let's make sure we don't forget to do the refactoring that was discussed in the PR. Please create ADO work items for anything that we decide to defer.

src/windows/wslc/services/ImageProgressCallback.cpp
```diff
@@ -84,23 +84,58 @@ std::wstring ImageProgressCallback::GenerateStatusLine(LPCSTR status, LPCSTR id,
...
+        line = std::format(L"{}: {}", id, status);
     }
 
-    // Erase any previously written char on that line.
-    while (line.size() < info.dwSize.X)
+    // Use the visible window width (not the buffer width) to prevent wrapping.
+    const auto visibleWidth = std::max<SHORT>(0, info.srWindow.Right - info.srWindow.Left + 1);
+
+    // Truncate to console width to prevent wrapping that would break cursor repositioning.
+    if (line.size() > static_cast<size_t>(visibleWidth))
```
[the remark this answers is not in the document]
kvega005: Is this valid? If both places are intending the same behavior, it might be a good idea to consolidate into a helper.

src/shared/inc/stringshared.h
```diff
@@ -825,6 +825,30 @@ struct CaseInsensitiveCompare
...
+    constexpr double c_kB = 1000.0;
+    constexpr double c_MB = 1000.0 * 1000.0;
+    constexpr double c_GB = 1000.0 * 1000.0 * 1000.0;
+
```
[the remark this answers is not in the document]
dkbennett: I like the SI / IEC distinction on format bytes. Or, merge the two together with a FormatUnitType enum as a parameter (and default to one).

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -81,78 +82,103 @@ try
...
     }
 
+    // Else is a build step
+    CollapseWindow();
+    auto wide = MultiByteToWide(status);
+    const auto bodyLength = wide.find_last_not_of(L"\r\n") + 1;
+    const auto newlines = wide.substr(bodyLength);
+    wide.resize(bodyLength);
+
+    WriteTerminal(std::format(L"\033[92m{}\033[0m{}", wide, newlines));
```
OneBlue: nit: I think it would be nice to have a set of constexpr strings for terminal escape codes like this (I had to look up that this was light green :) )

src/windows/wslc/services/BuildImageCallback.cpp
```diff
@@ -81,78 +82,103 @@ try
...
-                if (!m_pendingLine.empty() && now - m_lastRedraw >= c_redrawInterval)
+                // \r\n is a line ending; standalone \r overwrites the current line.
+                if (*(p + 1) != '\n')
                 {
-                    Redraw();
-                    m_lastRedraw = now;
+                    // Flush a throttled redraw before clearing so \r-based progress
+                    // updates are visible even when batched in a single OnProgress call.
+                    auto now = std::chrono::steady_clock::now();
+                    if (!m_pendingLine.empty() && now - m_lastRedraw >= c_redrawInterval)
```
OneBlue: We're doing this in 3 places, so I would recommend factoring that logic in a method like `RedrawIfNeeded()`

OneBlue (dismissed):
LGTM. Minor comments

src/windows/wslc/services/BuildImageCallback.cpp
dkbennett: Agree they should be constants, can copy the stuff from the test header  and we can refactor afterwards.

OneBlue (approved):

PR #40606, opened 2026-05-20
Refactor IsContainerNetworkMode into ParseContainerNetworkMode

beena352:
Replaces the IsContainerNetworkMode boolean helper with ParseContainerNetworkMode which returns the parsed target name directly. This eliminates duplicated prefix+substr+validate logic that was hand-rolled in multiple places.

## PR Checklist

This is a followup PR to #40601 addressing review comment to consolidate the container network mode parsing logic.
The container:<name|id> prefix parsing was done in two different ways:

-  IsContainerNetworkMode() check followed by manual substr()
- Direct starts_with() + substr() without using the helper at all
ParseContainerNetworkMode consolidates both patterns into a single function that returns std::optional<std::string> - the target name if the mode matches, std::nullopt otherwise. Two overloads are provided: one for LPCSTR (user-input paths) and one for std::string_view (Docker inspect recovery path).

### DIFF
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -151,32 +151,20 @@ uint16_t AllocateEphemeralPort(int family, const char* address)
 
 constexpr std::string_view c_containerNetworkPrefix = "container:";
 
-// Returns the target container name (the part after "container:") if Mode is a
-// container-mode network spec; std::nullopt otherwise.
-std::optional<std::string> ParseContainerNetworkMode(LPCSTR Mode)
+// Returns the target name after "container:" if present, std::nullopt otherwise.
+std::optional<std::string> ParseContainerNetworkMode(std::string_view mode)
 {
-    if (Mode == nullptr)
+    if (!mode.starts_with(c_containerNetworkPrefix))
     {
         return std::nullopt;
     }
 
-    const std::string_view view{Mode};
-    if (!view.starts_with(c_containerNetworkPrefix))
-    {
-        return std::nullopt;
-    }
-
-    return std::string{view.substr(c_containerNetworkPrefix.size())};
+    return std::string{mode.substr(c_containerNetworkPrefix.size())};
 }
 
-std::optional<std::string> ParseContainerNetworkMode(std::string_view Mode)
+std::optional<std::string> ParseContainerNetworkMode(LPCSTR mode)
 {
-    if (!Mode.starts_with(c_containerNetworkPrefix))
-    {
-        return std::nullopt;
-    }
-
-    return std::string{Mode.substr(c_containerNetworkPrefix.size())};
+    return mode != nullptr ? ParseContainerNetworkMode(std::string_view{mode}) : std::nullopt;
 }
 
 // Builds port mapping list from container options and returns the network mode string.
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -152,7 +152,7 @@ uint16_t AllocateEphemeralPort(int family, const char* address)
 constexpr std::string_view c_containerNetworkPrefix = "container:";
 
 // Returns the target name after "container:" if present, std::nullopt otherwise.
-std::optional<std::string> ParseContainerNetworkMode(std::string_view mode)
+std::optional<std::string> ParseContainerTarget(std::string_view mode)
 {
     if (!mode.starts_with(c_containerNetworkPrefix))
     {
@@ -162,9 +162,9 @@ std::optional<std::string> ParseContainerNetworkMode(std::string_view mode)
     return std::string{mode.substr(c_containerNetworkPrefix.size())};
 }
 
-std::optional<std::string> ParseContainerNetworkMode(LPCSTR mode)
+std::optional<std::string> ParseContainerTarget(LPCSTR mode)
 {
-    return mode != nullptr ? ParseContainerNetworkMode(std::string_view{mode}) : std::nullopt;
+    return mode != nullptr ? ParseContainerTarget(std::string_view{mode}) : std::nullopt;
 }
 
 // Builds port mapping list from container options and returns the network mode string.
@@ -183,6 +183,7 @@ std::pair<std::vector<ContainerPortMapping>, std::string> ProcessPortMappings(
 
     // Determine network mode string.
     std::string networkMode;
+    std::optional<std::string> containerTarget;
     if (networkType == WSLCContainerNetworkTypeBridged)
     {
         networkMode = "bridge";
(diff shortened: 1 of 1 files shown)


src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -241,7 +251,7 @@ std::pair<std::vector<ContainerPortMapping>, std::string> ProcessPortMappings(
 
         // Allocate VM ports for bridged and custom networks. Host mode ports are allocated when the container starts.
         if (networkType == WSLCContainerNetworkTypeBridged ||
-            (networkType == WSLCContainerNetworkTypeCustom && !IsContainerNetworkMode(containerNetworkName)))
+            (networkType == WSLCContainerNetworkTypeCustom && !ParseContainerNetworkMode(containerNetworkName)))
         {
             entry.VmMapping.AssignVmPort(virtualMachine.AllocatePort(e.Family, e.Protocol));
         }
```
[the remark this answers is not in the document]
beena352: This can't happen. Earlier in the function (line 207), container mode throws if requestedPorts is not empty. So the loop only executes for non-container modes, where ParseContainerNetworkMode returns std::nullopt immediately with no allocation.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -151,9 +151,20 @@ uint16_t AllocateEphemeralPort(int family, const char* address)
...
+std::optional<std::string> ParseContainerNetworkMode(std::string_view mode)
 {
-    return name != nullptr && std::string_view(name).starts_with(c_containerNetworkPrefix);
+    if (!mode.starts_with(c_containerNetworkPrefix))
+    {
+        return std::nullopt;
+    }
+
+    return std::string{mode.substr(c_containerNetworkPrefix.size())};
+}
```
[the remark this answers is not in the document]
beena352: The allocation only happens when the prefix matches. The two call sites that use the result need to own the string (one passes it to InspectContainer, the other stores it in a struct). Returning a string_view would mean callers have to worry about the original string being valid for long enough, which adds complexity for no real benefit.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -151,9 +151,20 @@ uint16_t AllocateEphemeralPort(int family, const char* address)
 
 constexpr std::string_view c_containerNetworkPrefix = "container:";
 
-bool IsContainerNetworkMode(LPCSTR name)
+// Returns the target name after "container:" if present, std::nullopt otherwise.
+std::optional<std::string> ParseContainerNetworkMode(std::string_view mode)
 {
```
kvega005: Nit: consider renaming for clarity. `ParseContainerNetworkMode` is a bit misleading because it doesn’t parse a general network mode; it only recognizes the `container:` prefix and extracts the target container identifier. A name like `ParseContainerTarget` would be more accurate.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -189,22 +200,21 @@ std::pair<std::vector<ContainerPortMapping>, std::string> ProcessPortMappings(
         THROW_HR_WITH_USER_ERROR_IF(
             E_INVALIDARG, Localization::MessageWslcContainerNetworkNameRequired(), !containerNetworkName || strlen(containerNetworkName) == 0);
 
-        if (IsContainerNetworkMode(containerNetworkName))
+        if (auto target = ParseContainerNetworkMode(containerNetworkName))
         {
```
kvega005: Nit: This is being used twice for the same input. Once here, and once in the if- check below (lin ~254). You could probably initialize the std::optional above (after `networkMode`). Set it here, if the networkType is custom, and use it again below so there is only one call to `ParseContainerNetworkMode`

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -151,9 +151,20 @@ uint16_t AllocateEphemeralPort(int family, const char* address)
...
+        return std::nullopt;
+    }
+
+    return std::string{mode.substr(c_containerNetworkPrefix.size())};
+}
+
+std::optional<std::string> ParseContainerTarget(LPCSTR mode)
+{
+    return mode != nullptr ? ParseContainerTarget(std::string_view{mode}) : std::nullopt;
 }
```
kvega005, line 168: Suggestion: reduce null-check duplication around GetPrimaryNetworkName / ParseContainerTarget.

I am looking at this overload, and I am trying to think "is there a way to refactor so we do not need to overload this?" It just feels like a lot of extra branching around the null checks. For example, in the `ProcessPortMappings` below, we already throw an error if `!containerNetworkName || strlen(containerNetworkName) == 0` then call `ParseContainerTarget(LPCSTR)` which checks for null again. 

The `ParseContainerTarget(LPCSTR)` overload and the scattered null checks (`!name || strlen(name) == 0`, `Networks == nullptr` repeated in two functions, branch-on-null at every call site) all exist because `GetPrimaryNetworkName` returns a nullable LPCSTR. A few targeted changes collapse all of this:

1. Change `GetPrimaryNetworkName` to return `std::optional<std::string>` or just `std::string`, throw for structural violations (Networks null with count > 0, Networks[0].NetworkName null). One place owns the contract; presence is encoded in the type. You could probably gate it so its only called for `WSLCContainerNetworkTypeCustom`, so you don't even need to deal with std::optional
2. Drop the ParseContainerTarget(LPCSTR) overload; keep only `ParseContainerTarget(const std::string&)` or `ParseContainerTarget(const std::string_view)`. All remaining call sites already have a `std::string`.
3. Change ProcessPortMappings’s `LPCSTR containerNetworkName` param to `const std::optional<std::string>& primaryNetworkName`. The null+empty check becomes `!primaryNetworkName || primaryNetworkName->empty();` the "must be null when not Custom" check becomes `.has_value() && type != Custom;` and the Custom branch dereferences safely.
4. In Create, call `GetPrimaryNetworkName` once, bind the optional, and pass it to both `ProcessPortMappings` and `ProcessAdditionalNetworks` instead of recomputing it. The duplicated `Networks == nullptr` check in `ProcessAdditionalNetworks` then goes away.

Net result: one overload deleted, one duplicated structural check removed, every if (name) / name != nullptr branch at the call sites disappears, and the nullable contract is expressed once in the type system rather 
than re-checked at each layer.

beena352: Good suggestion- agree the nullable LPCSTR chain could be cleaner with typed optionals. That's a bigger change than this followup nit though (touches GetPrimaryNetworkName, ProcessPortMappings, ProcessAdditionalNetworks, and Create). I'll do it as a separate follow-up PR.

kvega005:
LGTM

kvega005 (approved):

PR #40607, opened 2026-05-20
Add job objects to terminate child processes on VM shutdown

benhillis:
## Summary

Add `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` job objects to ensure that child processes spawned by the service (wsldevicehost, wslhost.exe, wslrelay.exe) are terminated when the VM shuts down. This prevents lingering processes from holding DLLs locked during package upgrades.

## PR Checklist

- [x] Applies to: WSL Windows components
- [x] Tests: Existing tests pass (`cmake --build . -- -m` succeeds)

### Problem
During package upgrade, `wsldevicehost.dll` (and potentially `wslhost.exe`/`wslrelay.exe`) can remain running after the service exits, holding file locks that prevent the installer from replacing binaries.

### Solution
Create per-VM job objects with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`:

- **DeviceHostProxy**: Creates a job object and assigns the COM device host process (received via `RegisterDeviceHost` PID) using `AssignProcessToJobObject`.
- **WslCoreVm**: Creates a job object and passes it to child process launches (debug console, kd relay, port relay, interop server) via `PROC_THREAD_ATTRIBUTE_JOB_LIST`.
- **SubProcess**: New `SetJobObject()` method that uses `PROC_THREAD_ATTRIBUTE_JOB_LIST` for clean job assignment at process creation time.
- **helpers**: All `Launch*` functions gain an optional `HANDLE JobObject` parameter (defaults to `nullptr` for backward compatibility).

When the VM is destroyed, the job handle closes and Windows automatically terminates all associated processes.

## Validation Steps

1. Build succeeds: `cmake --build . -- -m` (wslservice.exe, wsl.exe)
2. Verified no existing callers are broken (all new parameters default to nullptr)
3. Code review confirmed lifetime safety and correctness

### DIFF
--- a/src/windows/wslcsession/WSLCVirtualMachine.cpp
+++ b/src/windows/wslcsession/WSLCVirtualMachine.cpp
@@ -263,6 +263,16 @@ void WSLCVirtualMachine::Initialize()
 {
     THROW_IF_FAILED(m_vm->GetId(&m_vmId));
 
+    // Create a job object that will terminate child processes (wslrelay.exe)
+    // when the VM is destroyed.
+    m_processJobObject.reset(CreateJobObjectW(nullptr, nullptr));
+    THROW_LAST_ERROR_IF(!m_processJobObject);
+
+    JOBOBJECT_EXTENDED_LIMIT_INFORMATION jobInfo{};
+    jobInfo.BasicLimitInformation.LimitFlags = JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE;
+    THROW_IF_WIN32_BOOL_FALSE(
+        SetInformationJobObject(m_processJobObject.get(), JobObjectExtendedLimitInformation, &jobInfo, sizeof(jobInfo)));
+
     // Start crash dump collection thread.
     auto crashDumpSocket = hvsocket::Listen(m_vmId, LX_INIT_UTILITY_VM_CRASH_DUMP_PORT);
     THROW_LAST_ERROR_IF(!crashDumpSocket);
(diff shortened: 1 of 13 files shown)


JohnMcPMS (dismissed):
Looks reasonable, but I definitely don't know enough about the process graph to speak to its completeness.

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -263,6 +263,15 @@ void WSLCVirtualMachine::Initialize()
...
     THROW_IF_FAILED(m_vm->GetId(&m_vmId));
 
+    // Create a job object that will terminate child processes (wslrelay.exe)
+    // when the VM is destroyed.
+    m_processJobObject.reset(CreateJobObjectW(nullptr, nullptr));
+    THROW_LAST_ERROR_IF(!m_processJobObject);
+
+    JOBOBJECT_EXTENDED_LIMIT_INFORMATION jobInfo{};
+    jobInfo.BasicLimitInformation.LimitFlags = JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE;
+    THROW_IF_WIN32_BOOL_FALSE(SetInformationJobObject(m_processJobObject.get(), JobObjectExtendedLimitInformation, &jobInfo, sizeof(jobInfo)));
```
JohnMcPMS: This is the third instance of this code; probably worth a helper.

JohnMcPMS (approved):

PR #40609, opened 2026-05-20
Surface MSI reboot-required failures instead of silently dropping VHDs

benhillis:

Stop dropping `ERROR_SUCCESS_REBOOT_REQUIRED` (3010) on the floor when the embedded MSI inside the Microsoft Store MSIX can't replace a locked file (most commonly `system.vhd` while a WSL2 distro is running). Today this manifests to users as **`system.vhd` disappearing after a Store update** — Windows Installer has renamed it to `C:\Windows\Installer\Config.Msi\*.rbf` and scheduled the new file via `MoveFileEx(MOVEFILE_DELAY_UNTIL_REBOOT)`, but `wslinstaller` silently turns 3010 into success, so the user is never told to reboot and just sees a broken WSL.

## PR Checklist

- [x] **Communication:** Discussed offline
- [x] **Tests:** Added `InstallerTests::UpgradeWithLockedFileReportsRebootRequired`
- [x] **Localization:** New `MessageUpdateRebootRequired` string added with `{Locked="WSL"}` comment
- [ ] **Dev docs:** N/A

Defense-in-depth fix across three layers:

1. `WslInstaller::InstallMsipackageImpl` no longer converts 3010 → 0. Propagates the real HRESULT and now keeps the MSI log on 3010 for diagnostics.
2. New `SetRebootRequiredMarker()` / `IsRebootRequired()` in `install.cpp` write a **volatile** registry key (`REG_OPTION_VOLATILE`) at `HKLM\Software\Microsoft\Windows\CurrentVersion\Lxss\MSI\RebootPending`. The kernel auto-clears it on reboot, so there is no cleanup code path and no stuck-marker failure mode.
3. `LxssUserSession::_CreateInstance` (the single funnel for all distro-starting service calls) checks the marker on entry and throws `THROW_HR_WITH_USER_ERROR(HRESULT_FROM_WIN32(ERROR_SUCCESS_REBOOT_REQUIRED), MessageUpdateRebootRequired())`. This catches every client path uniformly — lifted `wsl.exe`, MSI-installed `wsl.exe`, `wslg.exe`, `bash.exe`, VS Code Remote-WSL, Terminal, etc. — without any new IDL/ABI surface. `CallMsiPackage` also checks the marker as a belt-and-braces measure for the lifted client.

### UX

During the Store auto-update: nothing visible (no toast, no prompt). On the user's next `wsl <anything>` invocation:

```
> wsl
WSL was updated but a system restart is required to complete the installation. Please reboot your machine and try again.
Error code: Wsl/Service/CreateInstance/0x80070bc2
```

After reboot: kernel destroys the volatile key, Windows pending-rename queue swaps the staged file into place, `wsl` works again. Zero cleanup code, no follow-up state to manage.

Commands that don't start a distro (`wsl --version`, `wsl --update`, `wsl --shutdown`, `wsl --list`) still work in this state — which is what we want, particularly for `--update` and `--shutdown`.

### Out of scope (follow-ups)

- **Defer the silent update when WSL is actively in use** so we don't have to reach this code path at all. Worth investigating as a follow-up.
- **Stop shipping `system.vhd` via MSI** (architectural — generate/repair it from a sidecar at first-run instead). Would eliminate the class of bug entirely.

The fix in this PR is necessary as a backstop regardless of either follow-up: silent data-loss-shaped bugs are unacceptable.

## Validation Steps Performed

- New automated test `InstallerTests::UpgradeWithLockedFileReportsRebootRequired` passes end-to-end:
  - Uninstalls the MSI, places a dummy file at the install path, memory-maps it (`MapViewOfFile` — required to actually defeat SYSTEM-level renames; plain `CreateFileW` without `FILE_SHARE_DELETE` is not sufficient), fakes the version to be older than the MSIX-embedded MSI, then drives the install through `InstallMsix()` → `WslInstaller` service.
  - Polls for the volatile marker, then verifies `wsl echo OK` fails with output containing "restart".
  - Cleans up (deletes marker, reinstalls MSI). End state is identical to start.
- Full Windows build (`cmake --build . -- -m`) succeeds.
- Manually verified the error code (`0x80070bc2` = `HRESULT_FROM_WIN32(ERROR_SUCCESS_REBOOT_REQUIRED)`) is consistent across all paths in this change.

### DIFF
--- a/localization/strings/en-US/Resources.resw
+++ b/localization/strings/en-US/Resources.resw
@@ -1008,6 +1008,10 @@ Falling back to NAT networking.</value>
   <data name="MessageFinishMsiInstallation" xml:space="preserve">
     <value>WSL is finishing an upgrade...</value>
   </data>
+  <data name="MessageUpdateRebootRequired" xml:space="preserve">
+    <value>WSL was updated but a system restart is required to complete the installation. Please reboot your machine and try again.</value>
+    <comment>{Locked="WSL"}</comment>
+  </data>
   <data name="MessageUpdateFailed" xml:space="preserve">
     <value>Update failed (exit code: {}).</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
--- a/src/windows/common/install.cpp
+++ b/src/windows/common/install.cpp
@@ -175,6 +175,16 @@ void WaitForMsiInstall()
         wprintf(L"\n%ls\n", message.get());
     }
 
+    if (exitCode == ERROR_SUCCESS_REBOOT_REQUIRED)
+    {
+        // The MSI completed but one or more files (typically system.vhd or wslservice.exe)
+        // were in use and have been scheduled for replacement on the next reboot. Surface
+        // this distinctly so the caller does not proceed to launch WSL against a
+        // half-installed package — notably, the previous system.vhd has been renamed away
+        // to %WINDIR%\Installer\Config.Msi\*.rbf and the new one is not yet in place.
+        THROW_HR_WITH_USER_ERROR(HRESULT_FROM_WIN32(ERROR_SUCCESS_REBOOT_REQUIRED), wsl::shared::Localization::MessageUpdateRebootRequired());
+    }
+
     if (exitCode != 0)
     {
         THROW_HR_WITH_USER_ERROR(HRESULT_FROM_WIN32(exitCode), wsl::shared::Localization::MessageUpdateFailed(exitCode));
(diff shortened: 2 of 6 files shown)


benhillis:
Good catch — fixed in 02969ca9. Removed the early IsRebootRequired() throw in CallMsiPackage(); the service-side check in _CreateInstance covers the distro-launching paths that actually depend on the half-installed files, and read-only / cleanup commands (--version, --shutdown, --list, --update) now work as expected.

Also added ClearRebootRequiredMarker() and wired it into the ERROR_SUCCESS paths of both the COM-driven and wsl --update install flows, so a user who shuts WSL down and re-runs the install can self-recover without an actual reboot.

Extended the integration test to verify wsl --version succeeds with the marker set and that the marker is cleared after a clean reinstall.

dkbennett:
This looks like almost the same PR as the newer one. Assuming the newer one is the intended reboot required messaging/behavior?

Issue #40537, opened 2026-05-14
WSL2 - Microsoft VS Code/Code.exe: Exec format error


Grufoony:
### Windows Version

Microsoft Windows [Version 10.0.26200.8457]

### WSL Version

2.6.3.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.87.2-1

### Distro Version

Ubuntu 26.04 / Ubuntu 24.04

### Repro Steps

If you open VS code with the classic command
```shell
code .
```
on a distro, then, if you switch distro, you get the error
```shell
grufoony@Gregorio:~$ code .
/mnt/c/Users/Gregorio/AppData/Local/Programs/Microsoft VS Code/bin/code: 62: /mnt/c/Users/Gregorio/AppData/Local/Programs/Microsoft VS Code/Code.exe: Exec format error
```

I noticed that if you do not call code on the other distro or you:
- call `code` command on distro 1
- reboot
- call `code` command on distro 2

it is apparently working

### Expected Behavior

See above

### Actual Behavior

See above

### Diagnostic Logs

[WslLogs-2026-05-14_10-27-14.tar.gz](https://github.com/user-attachments/files/27750937/WslLogs-2026-05-14_10-27-14.tar.gz)

benhillis:
I've seen a few reports of this, but I haven't been able to determine a repro. Do you have repeatable steps to get into this state?

Grufoony:
The only way I was able to reproduce this is:
- Have both Ubuntu 24.04 and 26.04 installed
- Open Ubuntu 24.04
- Launch `code .` command (it should work)
- Open Ubuntu 26.04
- Launch `code .` command -> now it should be broken

It has been a week since I have both distro installed and I'm dealing with this issue every day when I start the PC... once fixed it seems to be good until the next reboot

gorgsenegger:
I get exactly the same error on Windows 11, WSL version 2.7.3.0 after installing Ubuntu 26.04 in addition to 22.04.

benhillis:
Tracked this down. `WSLInterop` lives in `binfmt_misc`, which is kernel-global across the whole WSL2 VM, so all your distros share one registry. When a systemd distro shuts down cleanly, `systemd-shutdown` runs `disable_binfmt()`, which wipes every entry including `WSLInterop` for all running distros. That's why running `code` in a second distro breaks the first, and why a reboot fixes it until the next time.

[...]

Issue #10883, opened 2023-12-06
Problem launching Windows processes from inside WSL2 Docker


nrclark:
### Windows Version

Microsoft Windows [Version 10.0.19045.3448]

### WSL Version

2.0.9.0

### Are you using WSL 1 or WSL 2?

- [X] WSL 2
- [ ] WSL 1

### Kernel Version

5.15.133.1-1

### Distro Version

Ubuntu 22.04

### Other Software

Docker CE, version 24.0.7. Installed on WSL2 from Docker's apt repo.
Notepad.exe

### Repro Steps

I am working on a project that needs to run Windows applications from inside of a WSL2 Docker container. When I try to launch a Windows application directly (notepad.exe used as an example), I get a strange failure. The failure _doesn't_ occur if I launch it from inside of a shell instead.

Inside of WSL2 (with dockerd running as `sudo` in another tab),  this command fails:
```
docker run --rm -it --privileged --mount type=bind,source=/mnt,target=/mnt \
    --mount type=bind,source=/run/WSL,target=/run/WSL \
    --mount type=bind,source=/etc/passwd,target=/etc/passwd,readonly \
    --mount type=bind,source=/etc/group,target=/etc/group,readonly \
    --mount "type=bind,source=$HOME,target=$HOME" \
    --workdir "$HOME" \
    -u $(id -u):$(id -g) \
    --env WSLENV="${WSLENV}" \
    --env WSL_DISTRO_NAME="${WSL_DISTRO_NAME}" \
    --env WSL_INTEROP="${WSL_INTEROP}" \
    --env WT_PROFILE_ID="${WT_PROFILE_ID}" \
    --env WT_SESSION="${WT_SESSION}" \
    ubuntu \
    '/mnt/c/Windows/system32/notepad.exe'
 ```
with the following error:

```
<3>WSL (1) ERROR: ConfigInitializeEntry:1621: Failed to mount (null) at /dev as devtmpfs 1
```

but this one succeeds OK:
```
docker run --rm -it --privileged --mount type=bind,source=/mnt,target=/mnt \
    --mount type=bind,source=/run/WSL,target=/run/WSL \
    --mount type=bind,source=/etc/passwd,target=/etc/passwd,readonly \
    --mount type=bind,source=/etc/group,target=/etc/group,readonly \
    --mount "type=bind,source=$HOME,target=$HOME" \
    --workdir "$HOME" \
    -u $(id -u):$(id -g) \
[...]

Issue #8843, opened 2022-09-22
[WSL2][systemd][interop] Unable to Execute Windows Binary when systemd enabled


thelastlin:
### Version

 0.67.6.0

### WSL Version

- [X] WSL 2
- [ ] WSL 1

### Kernel Version

```
Linux version 5.15.62.1-microsoft-standard-WSL2 (oe-user@oe-host) (x86_64-msft-linux-gcc (GCC) 9.3.0, GNU ld (GNU Binutils) 2.34.0.20200220) #1 SMP Wed Aug 24 22:24:20 UTC 2022
```

### Distro Version

Ubuntu 22.04

### Repro Steps

1. Edit `/etc/wsl.conf`, enable systemd support. (https://devblogs.microsoft.com/commandline/systemd-support-is-now-available-in-wsl/)
2. Run `wsl.exe --shutdown` in wsl2 shell, and reopen the shell.
3. Run `wsl.exe --version` or some other PE binaries, and shell will warn you that the "exec format error" or "MZ: not found"

### Expected Behavior

Able to execute any valid PE binaries.

### Actual Behavior

"exec format error" or "MZ: not found"

### Diagnostic Logs

```shell
$ execve("/mnt/c/WINDOWS/system32/calc.exe", ["calc.exe"], 0x7ffc12cdd038 /* 37 vars */) = -1 ENOEXEC (Exec format error)
strace: exec: Exec format error
+++ exited with 1 +++
```
```shell
$ echo $WSL_INTEROP
/run/WSL/410_interop

$ ll /run/WSL
total 0
drwxrwxrwx  2 root root 100  9月 22 12:29 ./
drwxr-xr-x 26 root root 760  9月 22 12:29 ../
srwxrwxrwx  1 root root   0  9月 22 12:29 2_interop=
srwxrwxrwx  1 root root   0  9月 22 12:29 410_interop=
srwxrwxrwx  1 root root   0  9月 22 12:29 549_interop=

$ ll /proc/sys/fs/binfmt_misc/
total 0
drwxr-xr-x 2 root root 0  9月 22 12:29 ./
dr-xr-xr-x 1 root root 0  9月 22 12:29 ../
-rw-r--r-- 1 root root 0  9月 22 12:29 llvm-12-runtime.binfmt
-rw-r--r-- 1 root root 0  9月 22 12:29 llvm-14-runtime.binfmt
-rw-r--r-- 1 root root 0  9月 22 12:29 python2.7
[...]

PR #40610, opened 2026-05-20
fix(binfmt): prevent WSL interop handler loss across distro lifecycle

benhillis:

Fixes the WSL `binfmt_misc` interpreter being lost or corrupted in several scenarios, resulting in `exec format error` when running Windows executables (`cmd.exe`, `notepad.exe`, etc.) from inside a WSL2 distro.

Builds on top of @gordonlam-msft's branch `workitem/13885-fix-binfmt-config-cleared-on-termination` (commits 57ca57c8..0f2a71bd) and adds polish based on code review feedback.

## PR Checklist

- [x] **Closes:** #13885 (also addresses related reports #40537, #10883, #8843)
- [ ] **Communication:** Builds on existing work by @gordonlam-msft; please loop them in for review.
- [x] **Tests:** Added `BinfmtSurvivesDistroTermination` (multi-distro termination + F flag retention) and `SystemdBinfmtIsRestored` (generated unit overrides).
- [x] **Localization:** No new user-facing strings.
- [ ] **Dev docs:** Not needed.
- [ ] **Documentation updated:** Not needed.

### Root cause

`binfmt_misc` is **not** namespaced in the Linux kernel — there is a single global registration table shared across all WSL2 distros in the same utility VM. The previous design had several failure modes:

1. **Distro shutdown wiped the table.** When a distro with systemd shut down, systemd unmounted `/proc/sys/fs/binfmt_misc`, which flushes the global table for **all** distros in the VM.
2. **`systemd-binfmt.service` re-registered without the `F` flag.** Distros that included an `/etc/binfmt.d/WSLInterop.conf` would clear and re-register the entry, losing the fix-binary flag. Without `F`, the kernel resolves `/init` in the calling mount namespace at exec time, which breaks interop from Docker containers, chroots, and other altered mount namespaces.
3. **Competing MZ handlers (mono, wine, cli) could take priority** when registered via `binfmt.d` after WSL's in-mini-init registration.
4. **Stale config could persist** across config changes (`protectBinfmt=true` → `false`).

### Fixes (per commit)

- **57ca57c8** `prevent binfmt_misc handler loss on distro termination` — drops in `proc-sys-fs-binfmt_misc.mount.d/override.conf` with `DefaultDependencies=no` + `Before=umount.target` so the mount survives shutdown.
- **55644b8c** `split binfmt registration macros for WSL1/WSL2 compatibility` — introduces `BINFMT_INTEROP_REGISTRATION_STRING_VM` with `:FP`; the WSL1 (lxcore) path keeps `:P` because lxcore doesn't support `F`.
- **3e832106** `add ExecStartPost to ensure WSL binfmt handler priority` — generates `/run/binfmt.d/WSLInterop.conf` (highest-priority binfmt.d directory) and replaces the `ExecStart=` override on `systemd-binfmt.service` / `binfmt-support.service` with an `ExecStartPost=` that re-asserts WSL's handler.
- **7e38183b** `clean up stale binfmt config when protectBinfmt is disabled` — unlinks `/run/binfmt.d/WSLInterop.conf` in the else branch.
- **0f2a71bd** `log error on binfmt config cleanup failure` — logs cleanup `unlink` errors other than `ENOENT`.
- **0a2d4b27** `add multi-distro binfmt handler survival test` — TAEF test that imports a second WSL2 distro, enables systemd in both, terminates the second, and verifies that interop still works in the first and that the live kernel entry retains the `F` flag.
- **4283c9d3** `make ExecStartPost race-tolerant and strengthen test` — review polish: tolerate concurrent registration from another distro by allowing the echo writes to fail silently, then verify the live kernel entry via grep; wait for systemd to actually be running in both distros before the termination assertion.

### ExecStartPost command

The `ExecStartPost` looks like:

```
/bin/sh -c '(echo -1 > /proc/sys/fs/binfmt_misc/WSLInterop 2>/dev/null || true) ; \
            (echo ":WSLInterop:M::MZ::/init:FP" > /proc/sys/fs/binfmt_misc/register 2>/dev/null || true) ; \
            grep -q "^interpreter /init$" /proc/sys/fs/binfmt_misc/WSLInterop && \
            grep -q "^flags:.*F" /proc/sys/fs/binfmt_misc/WSLInterop'
```

The two writes are best-effort so systemd-binfmt.service isn't marked failed in races; the trailing greps verify the live kernel entry has the correct interpreter and the `F` flag.

## Validation Steps Performed

- `clang-format --dry-run --Werror` clean on all changed files.
- `python3 tools/devops/validate-copyright-headers.py` no new issues introduced.
- Existing `SystemdBinfmtIsRestored` test continues to validate generated unit overrides.
- New `BinfmtSurvivesDistroTermination` test covers the multi-distro scenario and asserts the `F` flag is preserved on the live kernel entry.
- Pending: full local Windows build + test run on hardware.


benhillis:
Closing — after a closer look, this is essentially a re-post of @yeelam-gordon's existing branch `workitem/13885-fix-binfmt-config-cleared-on-termination` with only minor editorial polish on top. Gordon's branch already has the substantive fixes; opening a separate PR steps on his work for no real gain.

I'll leave my suggestions as feedback on the underlying issue so @yeelam-gordon can take them (or not) when opening the PR from their own branch.

PR #40612, opened 2026-05-21
Prevent binfmt_misc unmount propagation across distros

### LINKED ISSUE #13885: binfmt config cleared on all distro when distro with systemd is terminated

benhillis:
## Summary

Fixes an `Exec format error` bug where Windows interop breaks in all running distros when a peer distro with systemd shuts down.

## Root Cause

`/proc` is mounted with `MS_SHARED`, so `binfmt_misc` (mounted under `/proc/sys/fs/binfmt_misc`) inherits shared mount propagation. When a peer distro's systemd shuts down and unmounts `binfmt_misc` in its mount namespace, the unmount propagates back through the shared peer group to init's namespace, wiping ALL binfmt_misc registrations globally.

## Fix

Mark `binfmt_misc` as `MS_PRIVATE` immediately after mounting so that unmounts in child namespaces are isolated and cannot propagate back. Also extends `UtilMount` to properly handle all mount propagation flags (`MS_SHARED`, `MS_PRIVATE`, `MS_SLAVE`, `MS_UNBINDABLE`) with mutual exclusivity validation and `MS_REC` support.

## PR Checklist

- [x] Builds cleanly
- [x] Includes regression test (`InteropSurvivesPeerDistroShutdown`)
- [x] Passes clang-format
- [x] Manually verified: entry survives peer distro termination, interop works, F-flag works in new mount namespaces

## Validation Steps

1. Import a second distro with systemd enabled
2. Start it and wait for systemd to come up
3. Terminate the peer distro (`wsl --terminate`)
4. Verify `cmd.exe /c echo test` still works in the primary distro
5. Verify `/proc/sys/fs/binfmt_misc/WSLInterop` still exists with F flag

Fixes #13885

### DIFF
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -1515,10 +1515,12 @@ Return Value:
     }
 
     //
-    // Mount the binfmt_misc filesystem.
+    // Mount the binfmt_misc filesystem as private so that unmounts in child
+    // mount namespaces (e.g. when a peer distro's systemd shuts down) do not
+    // propagate back and wipe the global registration table.
     //
 
-    if (UtilMount(nullptr, BINFMT_PATH, "binfmt_misc", MS_RELATIME, nullptr) < 0)
+    if (UtilMount(nullptr, BINFMT_PATH, "binfmt_misc", MS_RELATIME | MS_PRIVATE, nullptr) < 0)
     {
         return -1;
     }
--- a/src/linux/init/util.cpp
+++ b/src/linux/init/util.cpp
@@ -1773,11 +1773,31 @@ Return Value:
     //      - For Plan9 (9p): device is busy or not found
     //      - For VirtioFS: invalid tag (device not ready)
     //
-    // N.B. MS_SHARED must be applied in a separate mount() call, so it is
-    //      stripped from the initial mount flags and applied after the mount.
+    // N.B. Propagation flags (MS_SHARED, MS_PRIVATE, MS_SLAVE, MS_UNBINDABLE) are
+    //      mutually exclusive and must be applied in a separate mount() call after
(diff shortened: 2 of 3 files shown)


benhillis:
Closing in favor of #14443, which is the correct fix and has CI passing.

Our MS_PRIVATE approach was wrong: binfmt_misc entries are scoped per user-namespace (or effectively global), not per mount-namespace. WSL peer distros share the root user-ns (no CLONE_NEWUSER), so peer's `systemd-binfmt --unregister` -> write `-1` to `/proc/sys/fs/binfmt_misc/status` clears the global entry table regardless of mount propagation. Gordon's PR attacks the actual trigger (suppresses `ExecStop`, declares `/run/binfmt.d/WSLInterop.conf`, keeps the mount unit alive through shutdown, fixes `:P` -> `:FP`).

PR #40620, opened 2026-05-21
Refactor: Use std::optional<std::string> to eliminate nullable LPCSTR chain

beena352:
Changes GetPrimaryNetworkName to return std::optional<std::string> instead of nullable LPCSTR. Eliminates the ParseContainerTarget(LPCSTR) overload, updates ProcessPortMappings parameter to const std::optional<std::string>& primaryNetworkName, threads the primary network name through ProcessAdditionalNetworks as a parameter, and calls GetPrimaryNetworkName once in Create instead of twice. This pushes the null/empty contract into the type system - no more re-checking for null at each call site.

## PR Checklist

- [X ] **Localization:** All end user facing strings can be localized

When a helper returns a nullable raw pointer and multiple callers each null-check it separately, we change the return type to std::optional so the contract lives in one place.
What changed:
- GetPrimaryNetworkName throws for structural violations (Networks == nullptr with count > 0, NetworkName == nullptr) and returns std::nullopt for non-Custom or zero networks
- ProcessPortMappings parameter is now typed: !primaryNetworkName replaces !containerNetworkName, ->empty() replaces strlen(), .has_value() replaces != nullptr
- ProcessAdditionalNetworks receives the precomputed optional instead of calling GetPrimaryNetworkName internally — eliminates the redundant call
- Create calls GetPrimaryNetworkName once and passes to both functions

### DIFF
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -162,24 +162,19 @@ std::optional<std::string> ParseContainerTarget(std::string_view mode)
     return std::string{mode.substr(c_containerNetworkPrefix.size())};
 }
 
-std::optional<std::string> ParseContainerTarget(LPCSTR mode)
-{
-    return mode != nullptr ? ParseContainerTarget(std::string_view{mode}) : std::nullopt;
-}
-
 // Builds port mapping list from container options and returns the network mode string.
 std::pair<std::vector<ContainerPortMapping>, std::string> ProcessPortMappings(
     std::vector<_WSLCPortMapping>& requestedPorts,
     WSLCContainerNetworkType networkType,
     WSLCVirtualMachine& virtualMachine,
     const std::unordered_map<std::string, NetworkEntry>& sessionNetworks,
-    LPCSTR containerNetworkName,
+    const std::optional<std::string>& primaryNetworkName,
     DockerHTTPClient& dockerClient)
 {
     THROW_HR_IF_MSG(
         E_INVALIDARG,
-        containerNetworkName != nullptr && networkType != WSLCContainerNetworkTypeCustom,
-        "ContainerNetworkName must be null when network type is not Custom");
+        primaryNetworkName.has_value() && networkType != WSLCContainerNetworkTypeCustom,
+        "Primary network name must not be set when network type is not Custom");
 
     // Determine network mode string.
     std::string networkMode;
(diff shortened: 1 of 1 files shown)


src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -199,9 +194,9 @@ std::pair<std::vector<ContainerPortMapping>, std::string> ProcessPortMappings(
     else if (networkType == WSLCContainerNetworkTypeCustom)
     {
         THROW_HR_WITH_USER_ERROR_IF(
-            E_INVALIDARG, Localization::MessageWslcContainerNetworkNameRequired(), !containerNetworkName || strlen(containerNetworkName) == 0);
+            E_INVALIDARG, Localization::MessageWslcContainerNetworkNameRequired(), !primaryNetworkName || primaryNetworkName->empty());
```
dkbennett, line 197: nit: while valid, would prefer using has_value() and value() with optional types for clarity and avoid accidental undefined behavior. Also consistent with the has_value() usage above.

dkbennett (approved):
Looks good, some nits about consistent std::optional use.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -442,20 +437,22 @@ constexpr ULONG GetAdditionalStartIndex(WSLCContainerNetworkType type)
...
     const auto type = network.ContainerNetworkType;
     if (type != WSLCContainerNetworkTypeCustom || network.NetworksCount == 0)
     {
-        return nullptr;
+        return std::nullopt;
     }
     THROW_HR_IF_MSG(E_INVALIDARG, network.Networks == nullptr, "Networks pointer is null but NetworksCount=%lu", network.NetworksCount);
-    return network.Networks[0].NetworkName;
+    THROW_HR_IF_MSG(E_INVALIDARG, network.Networks[0].NetworkName == nullptr, "NetworkName is null for first network entry");
+    return std::string{network.Networks[0].NetworkName};
```
kvega005, line 449: NIT:
Before the PR (baseline), the code path for a Custom-network container with a null NetworkName looked like this:

1. GetPrimaryNetworkName(network) returned network.Networks[0].NetworkName directly — i.e. it returned nullptr (an LPCSTR).
2. That nullptr was passed down to ProcessPortMappings, where the first thing it hit was:
```
THROW_HR_WITH_USER_ERROR_IF(
    E_INVALIDARG,
    Localization::MessageWslcContainerNetworkNameRequired(),
    !containerNetworkName || strlen(containerNetworkName) == 0);
```

That throws an E_INVALIDARG with a localized, user-facing message ("A network name is required …") that wslc.exe prints to the user.


After the PR, the new GetPrimaryNetworkName helper looks like:

```
THROW_HR_IF_MSG(E_INVALIDARG, network.Networks[0].NetworkName == nullptr,
                "NetworkName is null for first network entry");
return std::string{network.Networks[0].NetworkName};
```

So now the null case throws earlier, but with THROW_HR_IF_MSG — which produces only an internal diagnostic string in the trace log. No localized message is attached to the error, so the user just sees a bare E_INVALIDARG (0x80070057) with no explanation.

Option A — keep the early check, but make it user-facing with `THROW_HR_WITH_USER_ERROR_IF` instead of `THROW_HR_IF_MSG` OR,
Option B — drop the check, return std::nullopt, let the downstream user-error fire

kvega005 (commented):
I don't think this really solves the probem we were trying to address of reducing all the duplicated logic. It is neater because we have typed strings, but we still end up with a bunch of `.has_value()` checks where the null checks were.

I think if we are going to do a refactor, we should try to maximize the wins we get. I've been thinking about this, and I think we should take a step back and workshop this a bit offline.

Here is some of the code duplication that I think we should target:
- ParseContainerTarget(*primaryNetworkName) is still called multiple times. 
- GetAdditionalStartIndex exists only because "is the first entry the primary?" is asked in three different places.
- The "should I allocate a VM port for this entry?" check (Bridged || (Custom && !containerTarget)) is open-coded in ProcessPortMappings and again in Open's per-port loop.
- ProcessAdditionalNetworks re-walks Networks[] and re-derives networkMode == "host"/"none" instead of being told.

beena352:
Closing this PR and continuing the work in #40672

PR #40622, opened 2026-05-21
install: emit warning instead of error when reboot is required after MSI update

benhillis:
## Summary

When the MSI installer returns `ERROR_SUCCESS_REBOOT_REQUIRED` (files in use, delayed rename pending), `WaitForMsiInstall` previously threw an error that blocked all subsequent `wsl.exe` invocations. This was unnecessarily harsh - the service's `_CreateInstance` gate already emits the same reboot-required warning when a distro launch is attempted.

## Changes

- **`WaitForMsiInstall()`**: emit `EMIT_USER_WARNING` and return instead of `THROW_HR_WITH_USER_ERROR`
- **`CallMsiPackage()`**: remove the now-dead reboot-specific rethrow

## Validation

- All 25 `InstallerTests` pass, including `UpgradeWithLockedFileReportsRebootRequired`

### DIFF
--- a/localization/strings/en-US/Resources.resw
+++ b/localization/strings/en-US/Resources.resw
@@ -1008,6 +1008,10 @@ Falling back to NAT networking.</value>
   <data name="MessageFinishMsiInstallation" xml:space="preserve">
     <value>WSL is finishing an upgrade...</value>
   </data>
+  <data name="MessageUpdateRebootRequired" xml:space="preserve">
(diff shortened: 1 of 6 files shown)


dkbennett (dismissed):
We should probably add similar checking of the reboot key to the WSLC CLI, but perhaps check the list of files that are holding up the install if possible (if they are WSLC-related, note it in the CLI and/or SDK API). Though I have never seen any issues with something of this nature with the CLI so probably very low priority for htat.

PR #40629, opened 2026-05-22
Add openvmm as vmm backend option for wslc vms

damanm24:
This PR adds [OpenVMM](https://github.com/microsoft/openvmm) as a possible VMM backing WSLC VMs.

This feature is gated on both a compile time gate and a WSLC feature flag. Non-exhaustive list of changes made to implement this:

1. Bump WSL.DeviceHost nuget package which now includes: OpenVMM and a corresponding VMService.proto file (used to generate protobuf C++ files to allow the wslc service to orchestrate VM management via [ttRPC](https://github.com/containerd/ttrpc) 
2. Implemented a ttrpc client that can communicate with OpenVMM's RPC server to configure and orchestrate the VM backing the WSLC session.
3. WSLCInit handles a new `WSLC_CONFIGURE_NETWORKING` message that allows us to configure networking within the VM using OpenVMM's [consomme networking stack](https://openvmm.dev/guide/reference/backends/consomme.html).
4. Introduced a new `OpenVmmVirtualMachine` class that implements the `IWSLCVirtualMachine` allowing users to switch between traditional HCS VMs and OpenVMM VMs. 
5. Extended `SocketChannel` class to support non-overlapped I/O which is not supported when using AF_UNIX sockets. This is required to bridge the connection between the `mini_init` process from the guest to the `wslcservice`.


msipackage/CMakeLists.txt
```diff
@@ -27,6 +27,13 @@ foreach(binary ${WINDOWS_BINARIES})
     list(APPEND BINARIES_DEPENDENCIES "${PACKAGE_INPUT_DIR}/${binary}")
 endforeach()
 
+if (INCLUDE_OPENVMM)
+    set(WSL_DEVICE_HOST_BINARIES wsldevicehost.dll;openvmm.exe)
```
benhillis, line 31: wsldevicehost.dll should be included all the time.

benhillis: maybe that's an existing issue?

src/linux/init/WSLCInit.cpp
```diff
@@ -675,6 +670,94 @@ void HandleMessageImpl(
     Transaction.SendResultMessage(result < 0 ? errno : 0);
 }
 
+void HandleMessageImpl(
```
benhillis, line 673: I think we probably want to use the existing GNS implementation to do network configuration. it should support all of this stuff.

src/windows/common/CMakeLists.txt
```diff
@@ -147,6 +149,19 @@ target_precompile_headers(common PRIVATE precomp.h)
 set_target_properties(common PROPERTIES FOLDER windows)
 target_include_directories(common PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/../service/mc/${TARGET_PLATFORM}/${CMAKE_BUILD_TYPE})
 
+# ATL headers (atlsafe.h) are needed by precomp.h but not automatically included
```
benhillis, line 152: why is this needed?

src/windows/service/exe/OpenVmmVirtualMachine.cpp
```diff
@@ -0,0 +1,730 @@
...
+
+    std::lock_guard lock(m_lock);
+
+    THROW_IF_FAILED(CoCreateGuid(&m_vmId));
+    m_vmIdString = wsl::shared::string::GuidToString<wchar_t>(m_vmId, wsl::shared::string::GuidToStringFlags::Uppercase);
+    m_featureFlags = Settings->FeatureFlags;
+
+    // Disable features not yet supported by the OpenVMM backend.
+    WI_ClearFlag(m_featureFlags, WslcFeatureFlagsGPU);
+    WI_ClearFlag(m_featureFlags, WslcFeatureFlagsVirtioFs);
```
benhillis: what does openvmm use for host shares if not virtiofs?

PR #40645, opened 2026-05-26
Add --timestamps, --since, and --until flags to wslc container logs

ptrivedi:
﻿## Summary

Wire up the existing `WSLCLogsFlagsTimestamps` flag and `Since`/`Until` parameters from the `IWSLCContainer::Logs` COM interface to the CLI. Previously these were hardcoded to 0/disabled despite the backend already supporting them.

## Changes

- Add `Timestamps` (flag, `-t`), `Since` (value), `Until` (value) argument types to `ArgumentDefinitions.h`
- Add `ULONGLONG` validation for `--since` and `--until` in `ArgumentValidation.cpp`
- Register new args on the container logs command
- Pass timestamps flag and since/until values through to the COM call in `ContainerService::Logs()`
- Add localization strings with `{Locked}` comments for the new flags
- Add 8 command-line parsing unit test cases in `CommandLineTestCases.h`
- Add 6 end-to-end test methods in `WSLCE2EContainerLogsTests.cpp` covering timestamps, since, until, combined usage, and short flag (`-t`)

## Usage

```
wslc container logs --timestamps mycontainer
wslc container logs --since 1700000000 mycontainer
wslc container logs --until 1700001000 mycontainer
wslc container logs -t -f --since 0 --tail 50 mycontainer
```

## Validation Steps

- Unit tests validate CLI argument parsing for valid and invalid inputs
- E2E tests verify timestamp output format, since/until filtering, and combined flag behavior
- The `-t` alias is safe as it is only used by `--tty` on create/exec/run commands, not logs
- Backend treats `since=0` and `until=0` as no filter, meaning the parameter is not sent to Docker API


localization/strings/cs-CZ/Resources.resw
```diff
@@ -2826,6 +2826,15 @@ Při prvním spuštění vytvoří soubor se všemi nastaveními okomentovanými
   <data name="WSLCCLI_TailArgDescription" xml:space="preserve">
     <value>Počet řádků, které se mají zobrazit od konce protokolů</value>
   </data>
+  <data name="WSLCCLI_TimestampsArgDescription" xml:space="preserve">
```
OneBlue, line 2829: These files don't need to be edited by hand (The `en-US` one is enough). The pipeline will apply those changes once merged.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -686,14 +686,27 @@ void ViewContainerLogs(CLIExecutionContext& context)
...
     if (context.Args.Contains(ArgType::Tail))
     {
         tail = validation::GetIntegerFromString<ULONGLONG>(context.Args.Get<ArgType::Tail>());
     }
 
-    ContainerService::Logs(session, WideToMultiByte(containerId), follow, tail);
+    ULONGLONG since = 0;
+    if (context.Args.Contains(ArgType::Since))
+    {
+        since = validation::GetIntegerFromString<ULONGLONG>(context.Args.Get<ArgType::Since>());
```
OneBlue, line 700: This works for now, but I think users will expect us to support RFC3339 timestamps here, so we should support formats like: 

```
wslc logs <id> --since  2013-01-02T13:23:37Z
```

And,  probably later, relative timestamps like: 

```
wslc logs <id> --since 1h
```

ptrivedi: good point, let me see in a follow up

localization/strings/cs-CZ/Resources.resw
ptrivedi: By hand no, my bot did it with its own claw :) But I do know that my pipeline run did fail with localization errors and then I asked the bot to apply the localization changes.

OneBlue: That's usually because of locked strings issue. The easiest way to fix it is by running ` python tools\devops\validate-localization.py  --fix`

Overall PR's shouldn't touch the resw files outside of the en-us one

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -509,7 +509,7 @@ InspectContainer ContainerService::Inspect(Session& session, const std::string&
     return wsl::shared::FromJson<InspectContainer>(output.get());
 }
 
-void ContainerService::Logs(Session& session, const std::string& id, bool follow, ULONGLONG tail)
+void ContainerService::Logs(Session& session, const std::string& id, bool follow, bool timestamps, ULONGLONG since, ULONGLONG until, ULONGLONG tail)
```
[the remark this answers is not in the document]
dkbennett: I would expect the clang formatter to do this as-needed, and if it decides it needs formatting then it would fail the formatting checks, so I believe this feedback can be ignored.

test/windows/wslc/e2e/WSLCE2EContainerLogsTests.cpp
```diff
@@ -57,6 +59,119 @@ class WSLCE2EContainerLogsTests
...
+        // Should have at most 2 lines (from --tail 2) and each should have a timestamp
+        auto lines = result.GetStdoutLines();
+        VERIFY_IS_TRUE(lines.size() <= 2);
+        for (const auto& line : lines)
+        {
+            if (!line.empty())
+            {
+                VERIFY_IS_TRUE(line.find(L'T') != std::wstring::npos);
+            }
+        }
```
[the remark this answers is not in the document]
dkbennett: This seems like a good catch.

dkbennett (approved):
LGTM, seems the runtime changes are already done to support the added parameters so this is essentially just CLI and tests w/ loc, and the CLI changes follow existing conventions very well and there are updated E2E tests and command line tests for validation. The existing tests should ensure there aren't argument ambiguity and there's no runtime changes here to worry about so approving.

The Copilot comments on one of the tests looks worth adjusting but its minor.

PR #40647, opened 2026-05-26
Add C# files to nuspec used in package stage

florelis:

In #40623 I made the inclusion of files in the SDK nuget conditional on what was configured to build. The package stage of the build pipeline does not use the .nuspec from the build stage, so by default it does not include the C# files. This change tells the package stage to include them.

## PR Checklist

### DIFF
--- a/.pipelines/package-stage.yml
+++ b/.pipelines/package-stage.yml
@@ -98,7 +98,7 @@ stages:
             displayName: "CMake configure (bundle-only)"
             inputs:
               workingDirectory: "."
-              cmakeArgs: . -DBUNDLE_ONLY=TRUE -DCMAKE_BUILD_TYPE=Release -DPACKAGE_VERSION=$(version.WSL_PACKAGE_VERSION) -DWSL_NUGET_PACKAGE_VERSION=$(version.WSL_NUGET_PACKAGE_VERSION) -DWSL_BUILD_SDKCS=true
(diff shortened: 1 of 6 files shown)


.pipelines/package-stage.yml
```diff
@@ -98,7 +98,7 @@ stages:
             displayName: "CMake configure (bundle-only)"
             inputs:
               workingDirectory: "."
-              cmakeArgs: . -DBUNDLE_ONLY=TRUE -DCMAKE_BUILD_TYPE=Release -DPACKAGE_VERSION=$(version.WSL_PACKAGE_VERSION) -DWSL_NUGET_PACKAGE_VERSION=$(version.WSL_NUGET_PACKAGE_VERSION)
+              cmakeArgs: . -DBUNDLE_ONLY=TRUE -DCMAKE_BUILD_TYPE=Release -DPACKAGE_VERSION=$(version.WSL_PACKAGE_VERSION) -DWSL_NUGET_PACKAGE_VERSION=$(version.WSL_NUGET_PACKAGE_VERSION) -DWSL_BUILD_SDKCS=true
```
benhillis: do we want to build in this phase? will that mean the dll isn't signed?

florelis: This won't make it build anything. For this stage, it only make the `.nuspec` list the C# DLL as one of its files. I probably should have renamed the variable, but I wanted to have a minimal change and get the fix asap.

I'll rename it to something like `WSL_INCLUDE_SDK_CSHARP` so it makes sense for both building and packaging

benhillis (approved):

PR #40541, opened 2026-05-14
Pipelines: exclude in-repo .ps1 scripts from CodeSign post-analysis

benhillis:
## Summary

The Guardian CodeSign post-analysis task has been failing the release build on `.ps1` files in the source tree (diagnostic / dev / test helpers) that we don't ship. The OneBranch `targetGlob` for actual signing already restricts signing to `.dll`/`.exe`/`.sys`/`.msi`/`.msix`/`.appx`/`.nupkg`, so PowerShell scripts being flagged here is purely a CodeSign validation noise issue.

## Change

Set `ob_sdl_codeSignValidation_excludes` to skip `**\*.ps1` on the three OneBranch pipelines (release, nightly, PR), and combine with the existing `testbin\**` exclusion in `build-job.yml` so neither rule clobbers the other.

## Files
- `.pipelines/build-job.yml` — extend existing exclude list with `**\*.ps1`
- `.pipelines/wsl-build-release-onebranch.yml` — add pipeline-level exclude
- `.pipelines/wsl-build-nightly-onebranch.yml` — add pipeline-level exclude
- `.pipelines/wsl-build-pr-onebranch.yml` — add pipeline-level exclude

## Validation

Dry-run pipeline (with test+flight stages skipped) is running against the same change on a sibling branch in parallel; opening this PR now to get review going while that finishes.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

PR #40653, opened 2026-05-27
Pipelines: exclude .ps1 from CodeSign in package stage

benhillis:
## Summary

The Guardian CodeSign post-analysis in the **package** job (build [147884488](https://dev.azure.com/microsoft/Microsoft.WSL/_build/results?buildId=147884488)) is failing on in-repo .ps1 scripts (`diagnostics/collect-wsl-logs.ps1`, `tools/deploy/*.ps1`, `triage/install-latest-wsl.ps1`, etc.) that are not shipped and don't need signing.

PR #40541 fixed this for the build job and added `ob_sdl_codeSignValidation_excludes: -|**\*.ps1` as a pipeline-level variable in the three `wsl-build-*-onebranch.yml` files. However, the `package` job in `package-stage.yml` declares its own job-level `variables:` block, and OneBranch's SDL injection only honors `ob_sdl_*` variables at job scope — so the pipeline-level value isn't being applied to the package job.

## PR Checklist

- [x] Issue exists in the WSL repo for this change
- [x] Validation steps below

## Detailed Description of the PR

Add `ob_sdl_codeSignValidation_excludes: -|**\*.ps1` to the `package` job's `variables:` block in `.pipelines/package-stage.yml`, mirroring what `build-job.yml` already does.

## Validation Steps

Run the release / nightly pipeline and confirm the package stage's 🛡 Guardian: Post Analysis step no longer reports `CodeSign.MissingSigningCert` errors for in-repo `.ps1` files.

### DIFF
--- a/.pipelines/package-stage.yml
+++ b/.pipelines/package-stage.yml
@@ -45,6 +45,7 @@ stages:
           ob_outputDirectory: '$(Build.SourcesDirectory)\out'
           ob_artifactBaseName: 'drop_wsl'
           ob_artifactSuffix: '_package'
+          ob_sdl_codeSignValidation_excludes: -|**\*.ps1
           buildStagePackageVersion: $[ stageDependencies.build_x64.build_x64.outputs['version.WSL_PACKAGE_VERSION'] ]
           buildStageNugetVersion: $[ stageDependencies.build_x64.build_x64.outputs['version.WSL_NUGET_PACKAGE_VERSION'] ]
 


.pipelines/package-stage.yml
```diff
@@ -45,6 +45,7 @@ stages:
           ob_outputDirectory: '$(Build.SourcesDirectory)\out'
           ob_artifactBaseName: 'drop_wsl'
           ob_artifactSuffix: '_package'
+          ob_sdl_codeSignValidation_excludes: -|**\*.ps1
```
OneBlue, line 48: nit: I don't think we need `-|` here since it's a single line string

OneBlue (approved):

PR #40655, opened 2026-05-27
Reject non-empty directories for new WSLC session storage

beena352:
Prevent users from accidentally reusing existing non-empty directories as WSLC session storage. For new sessions, validate that the storage directory is either empty or non-existent. Existing sessions identified by their storage.vhdx VHD bypass this check, so users can re-enter sessions freely even if additional files are present in the storage directory.

## PR Checklist

- [X] **Closes:** Link to issue #62259817

Validate session storage paths at initialization time:
- If the storage path exists but is not a directory, reject with E_INVALIDARG and a user-facing error message 
- If creating a new session and the directory already exists and is non-empty, reject with E_INVALIDARG and a user-facing error message
- If the directory is empty or doesn't exist, create it and proceed
- If a session's VHD already exists in the directory, skip the emptiness check (re-entry allowed)
This prevents data loss by ensuring new sessions don't accidentally overwrite unrelated user files placed in the storage directory. Once a session is created (VHD exists), the directory can contain other files without blocking re-entry.


OneBlue (commented):
I think refusing to create a session storage if the directory is non-empty makes sense (although I worry a bit about this putting us in a corner, like if there's ever a thumdb.db, desktop.ini or similar file created, this could "lock a user out" of using that folder permanently) 

I'm not sure about the marker file though. Whenever we create a storage folder, we always create a storage VHD, so could we use that directly instead ?

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -285,6 +286,52 @@ try
...
+                const bool empty = std::filesystem::is_empty(storagePath, ec);
+                THROW_IF_WIN32_ERROR_MSG(ec.value(), "is_empty failed for %ls", storagePath.c_str());
+
+                THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessageWslcSessionStorageMustBeEmpty(storagePath.c_str()), !empty);
+            }
+            else
+            {
+                std::filesystem::create_directories(storagePath, ec);
+                THROW_IF_WIN32_ERROR_MSG(ec.value(), "create_directories failed for %ls", storagePath.c_str());
+            }
```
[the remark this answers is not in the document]
beena352: Fixed. Added explicit validation that rejects paths existing as files with a clear localized error message and test case.

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -273,7 +273,18 @@ void WSLCSessionManagerImpl::CreateSession(
...
+        {
+            const auto factoryHr = factory->CreateSession(&sessionSettings, vm.Get(), notifier.Get(), WarningCallback, &session, &serviceRef);
+            if (FAILED(factoryHr))
+            {
+                if (auto comError = wslutil::GetCOMErrorInfo(); comError && comError->Message)
+                {
+                    THROW_HR_WITH_USER_ERROR(factoryHr, comError->Message.get());
+                }
+
+                THROW_IF_FAILED(factoryHr);
```
OneBlue: nit: This could be changed to just THROW_HR, since we know it's a failed HRESULT

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -285,6 +286,57 @@ try
...
+
+        THROW_HR_WITH_USER_ERROR_IF(
+            HRESULT_FROM_WIN32(ERROR_PATH_NOT_FOUND),
+            Localization::MessageWslcSessionStorageNotFound(Settings->StoragePath),
+            !vhdExists && WI_IsFlagSet(Settings->StorageFlags, WSLCSessionStorageFlagsNoCreate));
+
+        // Only check emptiness for new sessions. Existing sessions are identified by their VHD.
+        if (!vhdExists)
+        {
+            const bool isDir = std::filesystem::is_directory(storagePath, ec);
```
OneBlue: Could we always call is_empty, and look at is_empty's error on failure so we don't need to specifically handle the isDir vs !isDir case ourselves ?

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -285,6 +286,57 @@ try
...
+        // for non-existent paths. Filter those out — only throw on real failures.
+        auto throwOnRealFsError = [](const std::error_code& ec, const std::filesystem::path& path, const char* op) {
+            if (ec && ec.value() != ERROR_FILE_NOT_FOUND && ec.value() != ERROR_PATH_NOT_FOUND)
+            {
+                THROW_IF_WIN32_ERROR_MSG(ec.value(), "%hs failed for %ls", op, path.c_str());
+            }
+        };
+
+        std::error_code ec;
+        const auto vhdPath = storagePath / c_storageVhdFilename;
```
OneBlue: nit: We could set m_storageVhdPath  here so we don't need to duplicate this in ConfigureStorage()

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -285,6 +286,57 @@ try
         TraceLoggingValue(m_displayName.c_str(), "DisplayName"),
         TraceLoggingValue(Settings->CreatorPid, "CreatorPid"));
 
+    // Validate storage path before creating VM resources (Terminate() would clobber the error).
```
OneBlue: What causes Terminate() to clobber the error in that case ?

src/windows/wslcsession/WSLCSession.cpp
OneBlue: Asking because I think we should keep that logic in ConfigureStorage if possible, because there we already check if the VHD exists or not, so we could have the logic to fail if the directory is not empty there, which should help simplify this

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -437,11 +439,26 @@ void WSLCSession::ConfigureStorage(const WSLCSessionInitSettings& Settings, PSID
...
             WI_IsFlagSet(Settings.StorageFlags, WSLCSessionStorageFlagsNoCreate));
 
+        // Reject any non-empty existing path so we don't mix user files with session storage.
+        // is_empty's error_code distinguishes "directory doesn't exist yet" (OK, we'll create it)
+        // from "exists but unreadable / not a directory / other I/O error" (fail).
+        std::error_code ec;
+        const bool empty = std::filesystem::is_empty(storagePath, ec);
+        if (!ec)
+        {
+            THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessageWslcSessionStorageMustBeEmpty(storagePath.c_str()), !empty);
```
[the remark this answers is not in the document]
beena352: Good catch. std::filesystem::is_empty() returns true for empty files, so it can't be used to distinguish empty files from empty directories.
Added an is_directory check before calling is_empty, and now return a clear error if the storage path is not a directory.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2643,11 +2666,14 @@ try
             }
 
             // N.B. dockerd has exited by this point, so unmounting the VHD is safe since no container can be running.
-            try
+            if (m_storageMounted)
             {
-                m_virtualMachine->Unmount(c_containerdStorage);
+                try
```
OneBlue, line 2830: nit: Probably a good idea to set `m_storageMounted` after we umounted it successfully here

src/windows/wslcsession/WSLCSessionFactory.cpp
```diff
@@ -61,6 +57,11 @@ try
     RETURN_IF_FAILED(session->QueryInterface(IID_PPV_ARGS(Session)));
     *ServiceRef = serviceRef.Detach();
 
+    // N.B. The destruction callback must be installed last, after all fallible operations.
+    // If installed earlier, an unwinding session local would fire the exit callback and race
+    // with the COM stub marshaling IErrorInfo back to the caller.
```
OneBlue, line 62: Wow I'm surprised that didn't actually this earlier. @benhillis: I wonder if we should use LockServer()'s ref counting to decide when we exit instead of an exit event that might be signaled while a COM call is still pending

OneBlue (dismissed):
LGTM, minor comment

OneBlue (dismissed):

OneBlue (approved):

PR #40663, opened 2026-05-28
Fix potential channel corruption after cancelled ReceiveMessageOrClosed()

OneBlue:

This change fixes a potential channel corruption if a transaction is cancelled after its reply has been fully read. 

This solves the issue by keeping track of the receives bytes at the channel level, and injecting them in the next receive. Stale messages will be discarded based on transaction ids.

This issue could manifest as the following error: 

```
Microsoft.Windows.Wslc	Error	05-28-2026 10:05:41.134	"HRESULTString: 	E_UNEXPECTED
code: 	messageSize < sizeof(MESSAGE_HEADER)
failurecount: 	2
file: 	wsl\src\windows\common\HandleIO.cpp
function: 	wsl::windows::common::io::ReadSocketMessageHandle::ProcessRecvResult
hr: 	0x8000FFFF
linenumber: 	608
message: 	Unexpected message size: 1
threadid: 	63244
type: 	0"	wsl\src\windows\common\HandleIO.cpp		wsl::windows::common::io::ReadSocketMessageHandle::ProcessRecvResult	9812	63244	2		19a8ec70-4423-48d5-838d-c1b35aff6683		
```

Which causes the session termination logic to hang for up to a minute, since it fails to signal processes 

## PR Checklist

### DIFF
--- a/src/shared/inc/SocketChannel.h
+++ b/src/shared/inc/SocketChannel.h
@@ -636,9 +636,8 @@ private:
         auto io = CreateIO();
 
         gsl::span<gsl::byte> message;
-
         io.AddHandle(std::make_unique<windows::common::io::ReadSocketMessageHandle>(
-            m_socket.get(), m_buffer, [&message](auto& received) { message = received; }));
+            m_socket.get(), m_buffer, m_pendingBytes, [&message](auto& received) { message = received; }));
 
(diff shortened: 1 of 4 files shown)


benhillis (approved):
Looks good to me, thanks for adding the unit test.

PR #40671, opened 2026-05-29
Image list/prune filters

kvega005:
Exposes the `--filter` flag on the `wslc image list` and `wslc image prune` CLI commands.

## PR Checklist

- Added -f,--filter <key=value> to wslc image list (alias wslc image ls) and wslc image prune. The flag is repeatable and follows the same syntax/validation as wslc container list --filter (CLI rejects entries without =; key/value validity is left to the Docker daemon, matching docker's behavior and aligning with the model adopted across the rest of the CLI).
- `wslc image prune --all` continues to map to dangling=false under the hood, but a user-supplied --filter dangling=… now takes precedence so explicit user input always wins (also matching docker).

## Validation Steps Performed

Added e2e tests covering the new flag.

### DIFF
--- a/src/windows/wslc/commands/ImageListCommand.cpp
+++ b/src/windows/wslc/commands/ImageListCommand.cpp
@@ -28,6 +28,7 @@ namespace wsl::windows::wslc {
 std::vector<Argument> ImageListCommand::GetArguments() const
 {
     return {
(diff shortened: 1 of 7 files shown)


OneBlue (approved):
LGTM

PR #40672, opened 2026-05-29
Change WSLC container NetworkMode from enum to string

beena352:
This PR changes the WSLC container networking contract so NetworkMode is represented as a string instead of an enum.
The goal is to better align with Docker-compatible values and support modes that are not well represented by a fixed enum.

## PR Checklist

Background:
NetworkMode was previously modeled as an enum, which constrained accepted values and required mapping logic.
This made it harder to support broader/forward-compatible networking modes.
**Implementation notes:**
Replaced enum-based NetworkMode handling with string-based handling in WSLC-facing structures and SDK boundaries.
Preserved existing behavior for supported values while removing unnecessary enum translation friction.
Kept the change scoped to NetworkMode representation and related validation/consumption paths.


src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -99,14 +99,14 @@ WSLCSignal Convert(WslcSignal signal)
...
     {
     case WSLC_CONTAINER_NETWORKING_MODE_NONE:
-        return WSLCContainerNetworkTypeNone;
+        return "none";
     case WSLC_CONTAINER_NETWORKING_MODE_BRIDGED:
-        return WSLCContainerNetworkTypeBridged;
+        return "bridge";
     default:
         THROW_HR_MSG(E_INVALIDARG, "Invalid WslcContainerNetworkingMode: %i", mode);
     }
```
kvega005, line 112: If we are getting rid of the enum in the runtime, we should get rid of it at the SDK too. That can be in follow-up PR though, if needed.

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -822,7 +822,8 @@ try
         containerOptions.PortsCount = static_cast<ULONG>(internalContainerSettings->portsCount);
     }
 
-    containerOptions.ContainerNetwork.ContainerNetworkType = internalContainerSettings->networking;
+    // SDK only exposes the network mode (no additional endpoints today).
+    containerOptions.ContainerNetwork.NetworkMode = internalContainerSettings->networking;
```
kvega005: Really nit (and can be ignored), but if this is being passed to network mode, I would call it networkMode too. Just for consistency

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -162,86 +162,131 @@ std::optional<std::string> ParseContainerTarget(std::string_view mode)
...
-    DockerHTTPClient& dockerClient)
+bool IsNoneMode(std::string_view mode) noexcept
 {
-    THROW_HR_IF_MSG(
-        E_INVALIDARG,
-        containerNetworkName != nullptr && networkType != WSLCContainerNetworkTypeCustom,
-        "ContainerNetworkName must be null when network type is not Custom");
+    return mode == "none";
+}
+
```
kvega005: I don't think we need methods for this. `networkMode == "none"` reads the same or better than `IsNoneMode(networkMode)`

localization/strings/en-US/Resources.resw
```diff
@@ -2332,14 +2332,23 @@ For privacy information about this product please visit https://aka.ms/privacy.<
     <value>Network '{}' already exists.</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
-  <data name="MessageWslcAdditionalNetworksRequirePrimary" xml:space="preserve">
+
+  <data name="MessageWslcHostNoneNoAdditionalNetworks" xml:space="preserve">
```
kvega005: Why was this renamed? I suggest avoid renaming localized strings, because (i think) that makes the translation think its a new string that needs to be re-translated.

localization/strings/en-US/Resources.resw
```diff
@@ -2332,14 +2332,23 @@ For privacy information about this product please visit https://aka.ms/privacy.<
     <value>Network '{}' already exists.</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
-  <data name="MessageWslcAdditionalNetworksRequirePrimary" xml:space="preserve">
+
```
kvega005: nit: newline

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -162,86 +162,131 @@ std::optional<std::string> ParseContainerTarget(std::string_view mode)
...
+    // Reject `service:foo` and similar unsupported colon-prefixed modes before any further processing.
+    RejectUnsupportedColonSyntax(mode);
+
+    // host/none own the entire netns and cannot have additional endpoints.
+    THROW_HR_WITH_USER_ERROR_IF(
+        E_INVALIDARG, Localization::MessageWslcHostNoneNoAdditionalNetworks(), (IsHostMode(mode) || IsNoneMode(mode)) && hasEndpoints);
+
+    // container: shares the target's netns and cannot have additional endpoints.
+    THROW_HR_WITH_USER_ERROR_IF(
+        E_INVALIDARG, Localization::MessageWslcContainerModeNoAdditionalNetworks(), mode.starts_with(c_containerNetworkPrefix) && hasEndpoints);
```
kvega005: Should we just let docker validate the primary/additional networks mismatch? What are your thoughts? Or does that affect some of our downstream logic?

The docker engine does return nice error messages in these cases. If we are doing the validation ourselves, we should try to make the error message match:
- "container sharing network namespace with another container or host cannot be connected to any other network"
- "container cannot be connected to multiple networks with one of the networks in private (none) mode"

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -437,72 +439,26 @@ void ProcessNamedVolumes(const WSLCContainerOptions& containerOptions, wsl::wind
     }
 }
 
-constexpr ULONG GetAdditionalStartIndex(WSLCContainerNetworkType type)
-{
-    return (type == WSLCContainerNetworkTypeCustom) ? 1 : 0;
-}
-
-LPCSTR GetPrimaryNetworkName(const WSLCContainerNetwork& network)
+void ConfigureLdPathForGpu(std::vector<std::string>& Env)
```
kvega005: I think this migh be a bad merge or something. This was removed in #40583

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1777,15 +1744,23 @@ std::unique_ptr<WSLCContainerImpl> WSLCContainerImpl::Open(
...
     labels.erase(metadataIt);
 
-    auto networkMode = ParseDockerNetworkMode(dockerContainer.HostConfig.NetworkMode);
+    // Docker treats empty NetworkMode as the default (bridge).
+    std::string networkMode = dockerContainer.HostConfig.NetworkMode.empty() ? std::string{"bridge"} : dockerContainer.HostConfig.NetworkMode;
+
+    // Mirror Create()'s validation on recovery: refuse to take ownership of a container whose mode
+    // we never could have produced (e.g. `service:foo`), instead of silently letting downstream
+    // string comparisons accept it.
+    RejectUnsupportedColonSyntax(networkMode);
```
kvega005: Nit: this would be cleaner if we called the method something like `RejectUnsupportedNetworkModes()`. It would be easier to tell what the code is doing. Then the code speaks for itself, and we wouldn't need a long comment explaining it.

localization/strings/en-US/Resources.resw
```diff
@@ -2332,14 +2332,23 @@ For privacy information about this product please visit https://aka.ms/privacy.<
     <value>Network '{}' already exists.</value>
```
kvega005, line 2322: Nit: MessageWslcContainerIpAddressNotSupported and MessageWslcContainerNetworkNameRequired are no longer used. They should be removed.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -162,86 +162,131 @@ std::optional<std::string> ParseContainerTarget(std::string_view mode)
     return std::string{mode.substr(c_containerNetworkPrefix.size())};
 }
```
kvega005: Nit: This helper is no longer used.

kvega005 (commented):
Looks great, just some minor comments.

src/windows/WslcSDK/wslcsdk.cpp
beena352: Will do this in a follow-up PR

src/windows/wslcsession/WSLCContainer.cpp
beena352: Good point, I agree we should let Docker validate these combinations instead of duplicating that logic in WSLC.

I've removed the client-side checks for host/none/container modes with additional networks:

- MessageWslcHostNoneNoAdditionalNetworks
- MessageWslcContainerModeNoAdditionalNetworks
These cases now flow through Docker's native validation and error messages, which keeps the behavior aligned with Docker and avoids us maintaining duplicate validation logic. This shouldn't affect any downstream WSLC behavior, since we don't depend on those specific pre-check errors elsewhere.

I also removed the tests that were asserting the old E_INVALIDARG path and kept the tests for validations we still own, such as duplicate networks and network-not-found scenarios.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -151,97 +151,115 @@ uint16_t AllocateEphemeralPort(int family, const char* address)
...
+        return "host";
     }
-    else if (networkType == WSLCContainerNetworkTypeNone)
+
+    if (mode == "none")
     {
-        networkMode = "none";
+        THROW_HR_IF_MSG(E_INVALIDARG, hasRequestedPorts, "Port mappings are not supported without networking");
+        return "none";
     }
```
kvega005, line 192: discussion, not blocking — this caught my eye while reading through ResolveNetworkMode. Most of what it's doing is validation that Docker would do anyway:

- container: with no target → Docker returns "No such container"
- port mappings with none or container: mode → Docker returns a "conflicting options" error
- non-existent `container:<target>` → Docker returns "No such container"

The one check Docker can't do is `sessionNetworks.contains(...)` — that's WSLC's "you can only attach to networks WSLC manages" policy. So in principle, the whole function could collapse to something like:

```
std::string ResolveNetworkMode(
    LPCSTR networkMode,
    const std::unordered_map<std::string, NetworkEntry>& sessionNetworks)
{
    const std::string_view mode =
        (networkMode == nullptr || *networkMode == '\0') ? std::string_view{"bridge"} : networkMode;

    RejectUnsupportedNetworkModes(mode);

    // Session policy: WSLC only allows attaching to networks it manages.
    // Everything else (empty container:, ports + none, missing target, etc.)
    // is validated by Docker at CreateContainer time.
    if (mode != "host" && mode != "bridge" && mode != "none" && !mode.starts_with(c_containerNetworkPrefix))
    {
        THROW_HR_WITH_USER_ERROR_IF(
            WSLC_E_NETWORK_NOT_FOUND,
            Localization::MessageWslcNetworkNotFound(std::string{mode}),
            !sessionNetworks.contains(std::string{mode}));
    }

    return std::string{mode};
}
```

I'm honestly not sure if that's the right direction though — flagging it more for awareness than to push for a change. The trade-offs are:

- Pro: Way less code. Docker is the source of truth for what's a valid network configuration anyway, so we're not duplicating its rules (or risking them drifting from Docker's actual behavior over time).
- Con: We lose the localized `WSLC_E_CONTAINER_NOT_FOUND` and `MessageWslcContainerModeNoPorts` strings for those edge cases — the user sees Docker's English error wrapped in `CATCH_AND_THROW_DOCKER_USER_ERROR` instead.

Curious what you think. If you've already considered this and decided the localized errors are worth keeping, that's a totally reasonable call and you can disregard.

kvega005 (approved):
LGTM

src/windows/wslcsession/WSLCContainer.cpp
beena352: Good point, I agree with the direction.
I think it makes sense for "ResolveNetworkMode" to handle only the WSLC-specific checks and let Docker handle the network mode validation itself. That keeps us aligned with Docker's behavior and avoids maintaining duplicate validation logic.
The one thing we'd need to account for is "container:" mode normalization. Today we resolve it to "container:<id>", and a few downstream checks depend on that normalized value. We'd either keep that piece or update the downstream logic as part of the cleanup.
Overall, I agree with the approach and plan to make that additional simplification in a follow-up PR.

PR #40674, opened 2026-05-29
Add crash dump callback

kvega005:
Allows users to specify a callback that gets invoked after a crash dump is successfully written.

## PR Checklist


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2525,6 +2562,57 @@ try
...
+    // hold m_crashDumpLock (and can't deadlock with Register/Remove on the same thread that the
+    // callback might in turn use).
+    std::vector<wil::com_ptr<ICrashDumpCallback>> snapshot;
+    {
+        auto lock = m_crashDumpLock.lock_shared();
+        snapshot.assign(m_crashDumpCallbacks.begin(), m_crashDumpCallbacks.end());
+    }
+
+    for (const auto& callback : snapshot)
+    {
```
OneBlue, line 2713: We should call `RegisterUserCOMCallback()` to make sure that we don't get stuck if the user COM call hangs

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -1247,6 +1249,8 @@ void WSLCVirtualMachine::CollectCrashDumps(wil::unique_socket&& listenSocket)
     // No impersonation needed - the session process already runs as the user.
     wslutil::SetThreadDescription(L"CrashDumpCollection");
 
+    const auto coInit = wil::CoInitializeEx(COINIT_MULTITHREADED);
```
OneBlue, line 1254: What's the context for calling CoInitializeEx() here ? I'd expect COM to already be initialized here

kvega005: This isn't an RPC/COM pool thread. `CollectCrashDumps` runs on `m_crashDumpThread`, a raw `std::thread` we spawn ourselves in `WSLCVirtualMachine::Initialize()`.

We need COM initialized on this thread because it makes cross-process calls into the client's `ICrashDumpCallback` proxies (via `m_onCrashDump` → `WSLCSession::OnCrashDumpWritten` → callback->`OnCrashDump(...));` without `CoInitializeEx` those proxy calls fail with CO_E_NOTINITIALIZED.

test/windows/WslcSdkTests.cpp
```diff
@@ -319,6 +323,61 @@ class WslcSdkTests
...
+            auto* p = static_cast<std::promise<Invocation>*>(context);
+            p->set_value(Invocation{
+                info->dumpPath ? std::wstring{info->dumpPath} : std::wstring{},
+                info->processName ? std::string{info->processName} : std::string{},
+                info->pid,
+                info->signal,
+                info->timestamp});
+        };
+
+        std::filesystem::path extraStorage = m_storagePath / "wslc-crashcb-storage";
```
OneBlue: nit: We should delete that folder in a scope_exit_log when this test case exits

OneBlue (dismissed):
LGTM. This branch will need to be rebased once https://github.com/microsoft/WSL/pull/40756 is merged

OneBlue (approved):

PR #40675, opened 2026-05-29
CLI: Rewrite ForEachAsync to use threadpool, add timeout

dkbennett:
This is an improvement of the ForEachAsync generic method to use the windows thread pool and keep the pool full of workers whenever one completes instead of waiting for every worker to be complete before starting another batch.  This also adds a timeout to the method so workers do not execute endlessly and does some refactoring for easier debugging.

## PR Checklist

* Rewrites ForEachAsync, refactoring out classes and methods into details instead of lambdas for easier debugging and nicer stack frames.
* Uses windows thread pool and wil wrappers consistent with elsewhere in the codebase. 
* Fills the pool and keeps it full as work completes so longer running threads dont hold up work on other threads and we have full utilization for the duration.
* Handles timeouts and cancellation of all workers if one worker times out. 
* Adds to and updates tests.
* Updates container stats usage of this to support the cancellation event.

## Validation Steps Performed
* Manual running of stats with many containers.
* Updated unit tests run quickly and cleanly.

### DIFF
--- a/src/windows/wslc/core/AsyncExecution.h
+++ b/src/windows/wslc/core/AsyncExecution.h
@@ -9,91 +9,228 @@ Module Name:
 Abstract:
 
     Provides ForEachAsync, a generic helper for executing a work callback
-    over a collection concurrently in bounded batches using std::async.
+    over a collection concurrently using the Windows thread pool with bounded
+    concurrency and cooperative cancellation.
 
 --*/
 #pragma once
 
 #include <algorithm>
-#include <future>
+#include <chrono>
+#include <memory>
 #include <optional>
 #include <utility>
 #include <vector>
+#include <wil/resource.h>
 #include <wil/result_macros.h>
 
 namespace wsl::windows::wslc {
 
-// Invokes onWork for each element in items concurrently, in batches of batchSize.
-// Results are delivered serially to onSuccess. Errors are delivered serially to onError.
-//
-// This keeps wall time proportional to ceil(N / batchSize) rather than N for operations
-// that have inherent per-item latency (e.g. network or IPC calls).
-//
-// Note: worker threads have no guaranteed per-thread initialization (e.g. COM). Callers
-// whose onWork requires per-thread setup (such as CoInitializeEx) are responsible for
-// performing it at the start of the onWork lambda.
-//
-// TWork   : TItem -> TResult                              (called concurrently)
-// TSuccess: TResult -> void                               (called serially)
-// TError  : (TItem, wil::ResultException) -> void         (called serially)
-template <typename TItem, typename TWork, typename TSuccess, typename TError>
-void ForEachAsync(const std::vector<TItem>& items, TWork onWork, TSuccess onSuccess, TError onError, size_t batchSize = 10)
-{
-    WI_ASSERT(batchSize > 0);
-    THROW_HR_IF(E_INVALIDARG, batchSize == 0);
+namespace detail {
 
-    using TResult = decltype(onWork(std::declval<TItem>()));
-
-    struct BatchResult
+    template <typename TItem, typename TResult>
+    struct WorkerResult
     {
-        explicit BatchResult(TItem capturedItem) : item(std::move(capturedItem))
-        {
-        }
-
         TItem item;
         std::optional<TResult> result;
         wil::ResultException error{S_OK};
         bool hasError{false};
     };
 
-    for (size_t batchStart = 0; batchStart < items.size(); batchStart += batchSize)
+    // Holds all state for one thread pool worker. Owned via shared_ptr so the memory
+    // remains valid if ForEachAsync unwinds while a work item is still running.
+    template <typename TItem, typename TWork, typename TResult>
+    struct SharedWorker
     {
-        const size_t batchEnd = std::min(batchStart + batchSize, items.size());
+        WorkerResult<TItem, TResult> workerResult;
+        TWork* onWork{nullptr};
+        HANDLE cancelHandle{nullptr};
+        wil::unique_event done;
+        wil::unique_threadpool_work work;
+    };
 
-        std::vector<std::future<BatchResult>> futures;
-        futures.reserve(batchEnd - batchStart);
+    // Manages a fixed pool of SharedWorkers and a shared cancellation event.
+    template <typename TItem, typename TWork, typename TSuccess, typename TError>
+    struct WorkerPool
+    {
+        NON_COPYABLE(WorkerPool);
+        NON_MOVABLE(WorkerPool);
 
-        for (size_t i = batchStart; i < batchEnd; ++i)
+        using TResult = decltype(std::declval<TWork>()(std::declval<TItem>(), std::declval<HANDLE>()));
+        using TSharedWorker = SharedWorker<TItem, TWork, TResult>;
+
+        std::vector<std::shared_ptr<TSharedWorker>> workers;
+        std::vector<HANDLE> doneHandles;
+        wil::unique_event cancelEvent;
+        std::chrono::milliseconds timeout;
+        DWORD cancelDrainMs{};
+
+        WorkerPool(size_t poolSize, TWork& onWork, std::chrono::milliseconds timeout_, std::chrono::milliseconds cancelDrainTimeout) :
+            timeout(timeout_), cancelDrainMs(static_cast<DWORD>(cancelDrainTimeout.count()))
         {
-            const auto& item = items[i];
-            futures.push_back(std::async(std::launch::async, [&onWork, item]() -> BatchResult {
-                BatchResult result{item};
-                try
-                {
-                    result.result = onWork(item);
-                }
-                catch (const wil::ResultException& ex)
-                {
-                    result.hasError = true;
-                    result.error = ex;
-                }
-                return result;
-            }));
+            cancelEvent.create(wil::EventOptions::ManualReset);
+
+            workers.reserve(poolSize);
+            doneHandles.reserve(poolSize);
+
+            for (size_t i = 0; i < poolSize; ++i)
+            {
+                auto worker = std::make_shared<TSharedWorker>();
+                worker->done.create(wil::EventOptions::ManualReset);
+                worker->onWork = &onWork;
+                worker->cancelHandle = cancelEvent.get();
+
+                // Work item is created once per worker and reused for each dispatched item.
+                worker->work.reset(::CreateThreadpoolWork(ThreadPoolCallback, worker.get(), nullptr));
+                THROW_LAST_ERROR_IF(!worker->work);
+
+                doneHandles.push_back(worker->done.get());
+                workers.push_back(std::move(worker));
+            }
         }
 
-        for (auto& future : futures)
+        void Launch(size_t workerIndex, const TItem& item)
         {
-            auto batchResult = future.get();
+            auto& worker = workers[workerIndex];
+            worker->workerResult = WorkerResult<TItem, TResult>{};
+            worker->workerResult.item = item;
+            worker->done.ResetEvent();
+            ::SubmitThreadpoolWork(worker->work.get());
+        }
 
-            if (batchResult.hasError)
+        void Drain(size_t workerIndex, TSuccess& onSuccess, TError& onError)
+        {
+            auto& worker = workers[workerIndex];
+
+            // Ensure the callback has fully returned before reading results.
+            ::WaitForThreadpoolWorkCallbacks(worker->work.get(), FALSE);
+            if (worker->workerResult.hasError)
             {
-                onError(batchResult.item, batchResult.error);
+                onError(worker->workerResult.item, worker->workerResult.error);
             }
-            else if (batchResult.result.has_value())
+            else if (worker->workerResult.result.has_value())
             {
-                onSuccess(*batchResult.result);
+                onSuccess(*worker->workerResult.result);
             }
         }
+
+        // Signals cancellation, waits up to cancelDrainMs for workers to exit, then throws ERROR_TIMEOUT.
+        // Workers that do not exit within cancelDrainMs are abandoned - they retain shared_ptr ownership
+        // of their state. onWork implementations must check the cancel event at natural checkpoints and
+        // exit promptly.
+        //
+        // Note: TerminateThread() is not used - it skips C++ destructors, leaves user-mode locks
+        // permanently held (causing deadlocks), and corrupts COM apartment state.
+        [[noreturn]] void CancelAndThrow(size_t remainingItems)
+        {
+            cancelEvent.SetEvent();
+
+            ::WaitForMultipleObjects(static_cast<DWORD>(doneHandles.size()), doneHandles.data(), TRUE, cancelDrainMs);
+
+            THROW_HR_MSG(
+                HRESULT_FROM_WIN32(ERROR_TIMEOUT),
+                "ForEachAsync: worker exceeded timeout of %lld ms (%zu items remaining).",
+                static_cast<long long>(timeout.count()),
+                remainingItems);
+        }
+
+        // Thread pool callback - invoked on a pool thread for each submitted work item.
+        static void CALLBACK ThreadPoolCallback(PTP_CALLBACK_INSTANCE, void* context, PTP_WORK) noexcept
+        {
+            auto& worker = *static_cast<TSharedWorker*>(context);
+
+            try
+            {
+                worker.workerResult.result = (*worker.onWork)(worker.workerResult.item, worker.cancelHandle);
+            }
+            catch (const wil::ResultException& ex)
+            {
+                worker.workerResult.hasError = true;
+                worker.workerResult.error = ex;
+            }
+
+            worker.done.SetEvent();
+        }
+    };
+
+} // namespace detail
+
+// Invokes onWork for each element in items concurrently using the Windows thread pool,
+// with concurrency bounded to poolSize. Results are delivered serially to onSuccess.
+// Errors are delivered serially to onError.
+//
+// onWork receives a HANDLE to a cancellation event and should check it at natural
+// checkpoints using WaitForSingleObject(cancel, 0), returning early if it is set.
+// On timeout, the event is signalled and ForEachAsync waits up to cancelDrainTimeout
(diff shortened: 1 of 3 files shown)


src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,91 +9,228 @@ Module Name:
...
-                catch (const wil::ResultException& ex)
-                {
-                    result.hasError = true;
-                    result.error = ex;
-                }
-                return result;
-            }));
+            cancelEvent.SetEvent();
+
+            ::WaitForMultipleObjects(static_cast<DWORD>(doneHandles.size()), doneHandles.data(), TRUE, cancelDrainMs);
```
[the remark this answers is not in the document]
dkbennett: I believe I addressed this in the latest commit.

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,91 +9,228 @@ Module Name:
...
+                doneHandles.push_back(worker->done.get());
+                workers.push_back(std::move(worker));
+            }
+        }
+
+        void Launch(size_t workerIndex, const TItem& item)
+        {
+            auto& worker = workers[workerIndex];
+            worker->workerResult = WorkerResult<TItem, TResult>{};
+            worker->workerResult.item = item;
```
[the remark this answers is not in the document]
dkbennett: Fixed by adding constructor.

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,91 +9,228 @@ Module Name:
...
         }
+
+        THROW_LAST_ERROR_IF(waitResult == WAIT_FAILED);
+        const size_t workerIndex = waitResult - WAIT_OBJECT_0;
+        pool.Drain(workerIndex, onSuccess, onError);
+        pool.Launch(workerIndex, items[nextItem++]);
+    }
+
+    // Wait for all in-flight workers to finish and collect their final results.
+    const DWORD finalWait = ::WaitForMultipleObjects(static_cast<DWORD>(workerCount), pool.doneHandles.data(), TRUE, timeoutMs);
```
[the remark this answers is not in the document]
dkbennett: This is actually probably fine and a good thing. The purpose of the timeout is a safety net. If one particular worker is hung, the others keep going, and if its still hung at the end + after the final wait, it will have had ample time to complete its work and a a throw on a timeout is appropriate.  Its not intended to strictly limit the working bounds of each worker, but to collectively prevent it from hanging indefinitely.  If one thread does go longer than it should but then later finishes then this is a good thing - it prevented a timeout failure without really blocking anything meaningful. It's added resilience since the purpose is a safety net. So I don't think this is a bad thing or worth the extra engineering to strictly enforce the timeout.

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -56,44 +43,235 @@ void ForEachAsync(const std::vector<TItem>& items, TWork onWork, TSuccess onSucc
...
+    // Keep the pool full - as each worker completes, drain its result and immediately
+    // assign it the next pending item. WaitForMultipleObjects(FALSE) wakes on the first
+    // completion, so no worker idles while work remains.
+    while (nextItem < items.size())
+    {
+        const DWORD waitResult = ::WaitForMultipleObjects(static_cast<DWORD>(workerCount), pool.doneHandles.data(), FALSE, pool.timeoutMs);
+
+        if (waitResult == WAIT_TIMEOUT)
+        {
+            pool.CancelAndThrow(items.size() - nextItem);
```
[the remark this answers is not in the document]
dkbennett: This is OK. The purpose is a safety net and prevent unbounded execution and hangs, if one worker gets a bit more time than the others but eventually succeeds this becomes added resilience.

src/windows/wslc/core/AsyncExecution.h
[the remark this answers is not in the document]
dkbennett: I've updated the comment in ForEachAsync to indicate that the timeout is a safety net not a strictly enforced limit.

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -56,44 +43,235 @@ void ForEachAsync(const std::vector<TItem>& items, TWork onWork, TSuccess onSucc
...
-                onError(batchResult.item, batchResult.error);
+                worker.workerResult.result = worker.context->onWork(worker.workerResult.item, worker.context->cancelEvent.get());
             }
-            else if (batchResult.result.has_value())
+            catch (const wil::ResultException& ex)
             {
-                onSuccess(*batchResult.result);
+                worker.workerResult.hasError = true;
+                worker.workerResult.error = ex;
             }
```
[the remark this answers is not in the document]
dkbennett: Good catch, added an additional catch for all other exceptions.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -566,14 +566,21 @@ void ShowContainerStats(CLIExecutionContext& context)
...
-    // wall time proportional to ceil(N / batchSize) rather than N.
+    // wall time proportional to ceil(N / poolSize) rather than N.
     nlohmann::json statsJson = nlohmann::json::array();
     wsl::windows::wslc::ForEachAsync<std::wstring>(
         containers,
         // Work to be done for each container ID on a separate thread.
-        [&session](const std::wstring& containerId) {
+        // cancelHandle is signalled if the overall operation times out, check it before
+        // the blocking Stats call so we exit cooperatively without waiting a full ~1s sample.
+        [&session, userSpecifiedContainers](const std::wstring& containerId, HANDLE cancelHandle) {
```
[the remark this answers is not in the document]
dkbennett: Your suggestion drops userSpecifiedContainers, but I have fixed it by changing session to by value.

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -56,44 +43,235 @@ void ForEachAsync(const std::vector<TItem>& items, TWork onWork, TSuccess onSucc
...
+    };
+
+    // Holds per-worker state. Each Launch heap-allocates a shared_ptr<SharedWorker> as the
+    // thread pool callback context, giving the callback shared ownership and ensuring this
+    // memory is not freed while a callback is still running.
+    template <typename TWork, typename TItem, typename TResult>
+    struct SharedWorker
     {
-        const size_t batchEnd = std::min(batchStart + batchSize, items.size());
+        WorkerResult<TItem, TResult> workerResult;
```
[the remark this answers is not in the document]
dkbennett: fixed in latest commit

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -56,44 +43,244 @@ void ForEachAsync(const std::vector<TItem>& items, TWork onWork, TSuccess onSucc
...
-                catch (const wil::ResultException& ex)
-                {
-                    result.hasError = true;
-                    result.error = ex;
-                }
-                return result;
-            }));
+            auto& worker = workers[workerIndex];
+            worker->workerResult = WorkerResult<TItem, TResult>{};
+            worker->workerResult.item = item;
```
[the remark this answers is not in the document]
dkbennett: No longer relevant due to latest update.

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,91 +9,286 @@ Module Name:
...
+        if (waitResult == WAIT_TIMEOUT)
+        {
+            pool.CancelAndThrow(items.size() - nextItem);
         }
+
+        THROW_LAST_ERROR_IF(waitResult == WAIT_FAILED);
+
+        const size_t workerIndex = waitResult - WAIT_OBJECT_0;
+        pool.Drain(workerIndex, onSuccess, onError);
+        pool.Launch(workerIndex, items[nextItem++]);
```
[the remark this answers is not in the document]
dkbennett: Added a cancelOnError in scope_exit to handle this which will drain in flight, and unstarted workers are pre-set to done so they dont' block this cancel.

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,91 +9,286 @@ Module Name:
...
+                doneHandles.push_back(worker->done.get());
+                workers.push_back(std::move(worker));
+            }
+        }
+
+        void Launch(size_t workerIndex, const TItem& item)
+        {
+            auto& worker = workers[workerIndex];
+            worker->workerResult = WorkerResult<TItem, TResult>{};
+            worker->workerResult.item = item;
```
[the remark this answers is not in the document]
dkbennett: Fixed

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
 
-        std::vector<std::future<BatchResult>> futures;
-        futures.reserve(batchEnd - batchStart);
+        // Signals cancellation and waits up to cancelDrainMs for all in-flight workers to exit.
+        // Unstarted worker slots have their done events pre-signaled so they never block.
+        // Called noexcept from the scope_exit error guard in ForEachAsync.
+        void CancelAndDrainInFlight() noexcept
+        {
+            context->cancelEvent.SetEvent();
+            ::WaitForMultipleObjects(static_cast<DWORD>(doneHandles.size()), doneHandles.data(), TRUE, cancelDrainMs);
```
OneBlue, line 173: I don't recommend having a "cancel timeout" here. If this timeout ever hits, this will leak threads which will most likely to undefined behavior / crashes

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
+
+        std::vector<std::shared_ptr<TSharedWorker>> workers;
+        std::vector<HANDLE> doneHandles;
+        std::shared_ptr<TSharedContext> context;
+        std::chrono::milliseconds timeout;
+        DWORD timeoutMs{};
+        DWORD cancelDrainMs{};
+
+        WorkerPool(size_t workerCount, TWork onWork, std::chrono::milliseconds timeout_, std::chrono::milliseconds cancelDrainTimeout) :
+            context(std::make_shared<TSharedContext>(std::move(onWork))),
```
OneBlue, line 95: Assuming that we don't leak threads, `context` doesn't need to be a pointer here (could just be a regular class field)

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
+
+        using TResult = decltype(std::declval<TWork>()(std::declval<TItem>(), std::declval<HANDLE>()));
+        using TSharedWorker = SharedWorker<TWork, TItem, TResult>;
+        using TSharedContext = SharedContext<TWork>;
+
+        std::vector<std::shared_ptr<TSharedWorker>> workers;
+        std::vector<HANDLE> doneHandles;
+        std::shared_ptr<TSharedContext> context;
+        std::chrono::milliseconds timeout;
+        DWORD timeoutMs{};
```
OneBlue, line 91: `timeoutMs` is unused

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
     }
+
+    const size_t workerCount = std::min(poolSize, items.size());
+
+    detail::WorkerPool<TItem, TWork, TSuccess, TError> pool{workerCount, std::move(onWork), timeout, cancelDrainTimeout};
+
+    // On any exception from Launch, Drain, or the user callbacks (onSuccess/onError),
+    // signal cancellation and wait for in-flight workers before rethrowing. This guarantees
+    // no background thread pool callbacks outlive the ForEachAsync call.
+    auto cancelOnError = wil::scope_exit([&pool] { pool.CancelAndDrainInFlight(); });
```
OneBlue, line 285: I recommend doing this in the WorkerPool's destuctor so there's a strong guarantee of no leaked threads

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
+        std::shared_ptr<TSharedContext> context;
+        std::chrono::milliseconds timeout;
+        DWORD timeoutMs{};
+        DWORD cancelDrainMs{};
+
+        WorkerPool(size_t workerCount, TWork onWork, std::chrono::milliseconds timeout_, std::chrono::milliseconds cancelDrainTimeout) :
+            context(std::make_shared<TSharedContext>(std::move(onWork))),
+            timeout(timeout_),
+            timeoutMs(timeout_ == std::chrono::milliseconds::max() ? INFINITE : static_cast<DWORD>(timeout_.count())),
+            cancelDrainMs(static_cast<DWORD>(cancelDrainTimeout.count()))
```
OneBlue, line 98: Given that we have a `Launch()`  method, I'd recommend only creating the workers there. That way this threadpool can dynamically resize itself as needed (which will allow us to reuse it in other places)

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
-                catch (const wil::ResultException& ex)
-                {
-                    result.hasError = true;
-                    result.error = ex;
-                }
-                return result;
-            }));
+            context->cancelEvent.SetEvent();
+
+            ::WaitForMultipleObjects(static_cast<DWORD>(doneHandles.size()), doneHandles.data(), TRUE, cancelDrainMs);
```
OneBlue, line 187: nit: We should check the return code of `WaitForMultipleObjects()` here

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
-                onError(batchResult.item, batchResult.error);
+                worker.workerResult.hasError = true;
+                worker.workerResult.error = ex;
             }
-            else if (batchResult.result.has_value())
+            catch (...)
             {
-                onSuccess(*batchResult.result);
+                worker.workerResult.hasError = true;
+                worker.workerResult.error = wil::ResultException{wil::ResultFromCaughtException()};
```
OneBlue, line 216: This catch should cover both catch() cases

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
+        WorkerResult() = default;
+
+        explicit WorkerResult(const TItem& item_) : item(item_)
         {
         }
 
-        TItem item;
+        std::optional<TItem> item;
         std::optional<TResult> result;
         wil::ResultException error{S_OK};
```
OneBlue, line 42: To simplify things, I would recommend storing the error as a `std::exception_ptr` (null if no error was thrown). 

This will have the benefit of allowing us to rethrow non-wil exceptions easily

OneBlue: (and we can get rid of `hasError`)

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
+            worker->workerResult.item.emplace(item);
+            worker->workerResult.result.reset();
+            worker->workerResult.error = wil::ResultException{S_OK};
+            worker->workerResult.hasError = false;
+
+            // Heap-allocate a shared_ptr as the callback context. The callback takes ownership,
+            // keeping the worker and its SharedContext alive for the full duration of the callback
+            // regardless of WorkerPool lifetime. Re-create the work item each launch so the
+            // context pointer is fresh.
+            auto* ctx = new std::shared_ptr<TSharedWorker>(worker);
```
OneBlue, line 138: Could we merge SharedContext and SharedWorker into one structure that the threadpool owns, and pass a pointer to that structure directly as the context ? This would simplify things a lot, and get rid of the `shared_ptr` pointer. 

If we store them as an std::list, the pointers never get invalidated

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
+    // callback, including after WorkerPool is destroyed on the timeout path. Owned via
+    // shared_ptr and referenced by every SharedWorker.
+    template <typename TWork>
+    struct SharedContext
+    {
+        NON_COPYABLE(SharedContext);
+        NON_MOVABLE(SharedContext);
+
+        TWork onWork;
+        wil::unique_event cancelEvent;
```
OneBlue, line 56: I don't think this cancel event should be in the threadpool structure, since the threadpool logic doesn't actually look at it. 

Callers can capture a cancel event in their work callback in they need to

src/windows/wslc/core/AsyncExecution.h
```diff
@@ -9,92 +9,322 @@ Module Name:
...
-        const size_t batchEnd = std::min(batchStart + batchSize, items.size());
+        NON_COPYABLE(WorkerPool);
+        NON_MOVABLE(WorkerPool);
+
+        using TResult = decltype(std::declval<TWork>()(std::declval<TItem>(), std::declval<HANDLE>()));
+        using TSharedWorker = SharedWorker<TWork, TItem, TResult>;
+        using TSharedContext = SharedContext<TWork>;
+
+        std::vector<std::shared_ptr<TSharedWorker>> workers;
+        std::vector<HANDLE> doneHandles;
```
OneBlue, line 88: If we remove the "timeout on cancel" logic, we can get rid of those

OneBlue (commented):
I think a generic threadpool that we can use across the codebase has a lot of value, thank you for doing this ! 

I think we need to design it a bit more defensively though (runway threads will unfortunately cause us a lot of pain, so we should avoid "timing out" on cancellation)

I also recommend checking out `src\windows\common\WslCoreMessageQueue.h`, which uses a threadpool as well. I think a good target would be to move that class to a generic threadpool, something like: 

```
template <TResult>
struct WorkItem
{
    TResult result;
    std::exception_ptr error;
    void Wait();
    void Cancel();    
};

class ThreadPool
{
    ThreadPool(min-threads, max-threads);

    template <TFunction>
    WorkItem SubmitWork(Function &&Work);
}
```

Should cover both usecases

dkbennett:
Converting back to draft to do some rework per @OneBlue comments. There's not huge urgency here so can take a bit of time and ensure it is the best it can be.

PR #40687, opened 2026-06-01
Fix incorrect wslpath translatation of \\wsl.localhost when the current distro is a subset of the target distro

OneBlue:

This change fixes an incorrect `wslpath` translation behavior that can lead to errors like: 

```
$wsl -d ubuntu  --cd \\wsl.localhost\ubuntu-24.04\
<3>WSL (838 - Relay) ERROR: CreateProcessCommon:788: chdir(-24.04) failed 2
```

This change solves the issue by correctly validating the `\\wsl.localhost\\` path targets the current distro, instead of just matching on a string prefix. 

Issue originally found by @alexgrest.

## PR Checklist


test/linux/unit_tests/wslpath.c
```diff
@@ -1,417 +1,427 @@
-/*++
```
benhillis, line 1: Hmm looks like line endings got messed up here.

benhillis (dismissed):
Looks like the right fix, looked at wslpath.c unit test with "ignore whitespace" and the new variation looks good.

test/linux/unit_tests/wslpath.c
OneBlue: Fixed in latest iteration

benhillis (approved):

PR #40694, opened 2026-06-03
Add support for saving multiple images in a single tar

OneBlue:

This change adds support for saving multiple images in a single tar both at the API & CLI level

## PR Checklist


src/windows/service/inc/wslc.idl
```diff
@@ -729,6 +731,7 @@ interface IWSLCSession : IUnknown
     HRESULT LoadImage([in] WSLCHandle ImageHandle, [in, unique] IProgressCallback* ProgressCallback, [in] ULONGLONG ContentLength);
     HRESULT ImportImage([in] WSLCHandle ImageHandle, [in] LPCSTR ImageName, [in, unique] IProgressCallback* ProgressCallback, [in] ULONGLONG ContentLength);
     HRESULT SaveImage([in] WSLCHandle OutputHandle, [in] LPCSTR ImageNameOrID, [in, unique] IProgressCallback * ProgressCallback, [in, unique, system_handle(sh_event)] HANDLE CancelEvent);
```
OneBlue, line 743: I might get rid of that method in a future change

benhillis:
### Test failure root cause

The failing `test wslc tests - rs_prerelease` job is the new **`WSLCTests::SaveImages`** unit test, in the *"one of the images not found"* sub-case. It fails on this assertion (`test/windows/WSLCTests.cpp`):

```cpp
ValidateCOMErrorMessage(L"No such image: not-found");
```

**Why:** `ValidateCOMErrorMessage` does an **exact** comparison, and the expected string is wrong.

`SaveImages` streams from Docker's `GET /images/get`, which surfaces the daemon's raw export error. For a missing image the daemon resolves all names *before* streaming, returns a non‑200, and the message is **`reference does not exist`** (moby `refstore.ErrDoesNotExist`) — not `No such image: ...`.

This is already established in the codebase for the save path:
- Existing single-image test (`WSLCTests.cpp`): `SaveImage(..., "hello-wld:latest")` → `ValidateCOMErrorMessage(L"reference does not exist")`.
- E2E `WSLCE2E_Image_Save_ImageNotFound`: `Stderr = L"reference does not exist\r\nError code: E_FAIL\r\n"`.

`SaveImages` uses the same `ExportImage` backend, so it returns the same message. The `No such image: ...` wording only applies to delete/tag/inspect, which go through a different path mapping to `WSLC_E_IMAGE_NOT_FOUND`.

The other assertions in that block are fine (the call does fail, and the file stays 0 bytes since the error is returned before streaming), and the new E2E `..._InvalidImage` test passes because it only checks exit code and non-empty stderr — so only this exact-match line breaks.

**Fix:**
```diff
-            ValidateCOMErrorMessage(L"No such image: not-found");
+            ValidateCOMErrorMessage(L"reference does not exist");
```

benhillis:
> ### Test failure root cause
> The failing `test wslc tests - rs_prerelease` job is the new **`WSLCTests::SaveImages`** unit test, in the _"one of the images not found"_ sub-case. It fails on this assertion (`test/windows/WSLCTests.cpp`):
> 
> ```c++
> ValidateCOMErrorMessage(L"No such image: not-found");
> ```
> 
> **Why:** `ValidateCOMErrorMessage` does an **exact** comparison, and the expected string is wrong.
> 
> `SaveImages` streams from Docker's `GET /images/get`, which surfaces the daemon's raw export error. For a missing image the daemon resolves all names _before_ streaming, returns a non‑200, and the message is **`reference does not exist`** (moby `refstore.ErrDoesNotExist`) — not `No such image: ...`.
> 
> This is already established in the codebase for the save path:
> 
> * Existing single-image test (`WSLCTests.cpp`): `SaveImage(..., "hello-wld:latest")` → `ValidateCOMErrorMessage(L"reference does not exist")`.
> * E2E `WSLCE2E_Image_Save_ImageNotFound`: `Stderr = L"reference does not exist\r\nError code: E_FAIL\r\n"`.
> 
> `SaveImages` uses the same `ExportImage` backend, so it returns the same message. The `No such image: ...` wording only applies to delete/tag/inspect, which go through a different path mapping to `WSLC_E_IMAGE_NOT_FOUND`.
> 
> The other assertions in that block are fine (the call does fail, and the file stays 0 bytes since the error is returned before streaming), and the new E2E `..._InvalidImage` test passes because it only checks exit code and non-empty stderr — so only this exact-match line breaks.
> 
> **Fix:**
> 
> ```diff
> -            ValidateCOMErrorMessage(L"No such image: not-found");
> +            ValidateCOMErrorMessage(L"reference does not exist");
> ```

Testing out an agent that monitors pull requests for test failures and tries to debug them.

OneBlue:
> **Fix:**
> 
> ```diff
> -            ValidateCOMErrorMessage(L"No such image: not-found");
> +            ValidateCOMErrorMessage(L"reference does not exist");
> ```

That's not true. The root cause of the failure was that the ListImage() & DeleteImage() test cases both exit with the `alpine:latest` image deleted, which caused SaveImage() to fail. 

Fixed in latest iteration

src/windows/service/inc/wslc.idl
benhillis: if we're going to do it we need to do it soon.

benhillis (approved):

Issue #40662, opened 2026-05-28
cannot use `wsl --update` to update WSL, nor use `wsl --install` to install a Linux distribution


TheWhiteDog9487:
### Windows Version

Microsoft Windows [版本 10.0.26200.8457]

### WSL Version

2.7.3.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.114.1-1

### Repro Steps

PowerShell 7.6.2
Loading personal and system profiles took 560ms.
PS C:\Users\TheWhiteDog9487> ([Security.Principal.WindowsPrincipal][Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)
False
PS C:\Users\TheWhiteDog9487> sudo pwsh
PowerShell 7.6.2
Loading personal and system profiles took 568ms.
PS C:\Users\TheWhiteDog9487> ([Security.Principal.WindowsPrincipal][Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)
True
PS C:\Users\TheWhiteDog9487> wsl --update
正在检查更新。
正在更新适用于 Linux 的 Windows 子系统: 2.7.3。
拒绝访问。
错误代码: Wsl/UpdatePackage/E_ACCESSDENIED
PS C:\Users\TheWhiteDog9487> wsl --install archlinux
正在下载: Arch Linux
拒绝访问。
错误代码: Wsl/InstallDistro/E_ACCESSDENIED
PS C:\Users\TheWhiteDog9487> wsl --install archlinux --web-download
正在下载: Arch Linux
拒绝访问。
错误代码: Wsl/InstallDistro/E_ACCESSDENIED
PS C:\Users\TheWhiteDog9487>

### Expected Behavior

Running `wsl --update`, waiting for the download to finish, WSL successfully updated itself.  
Running `wsl --install archlinux`, waiting for the download to complete, successfully installed the Arch Linux WSL version.

### Actual Behavior

Whether updating WSL itself or downloading and installing a Linux distribution, an "Access Denied E_ACCESSDENIED" error appears immediately after the download begins, and the download and installation process is aborted straight away after the error occurs.
The issue exists on both my laptop and desktop.
For now, I can only update WSL by downloading MSIXBUNDLE or MSI format installation packages from the releases page of this repository — those work fine.
[...]

PR #40708, opened 2026-06-04
Fix E_ACCESSDENIED during download if temp is hidden

chemwolf6922:

The `winrt::Windows::Storage::StorageFolder::GetFolderFromPathAsync` function does not accept folders with hidden or system attributes. To solve that, this PR avoids the GetFolderFromPathAsync call and opens the file stream directly.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/40662

## Validation Steps Performed

Add test UnitTests::UnitTests::DownloadToHiddenSystemTempFolder


test/windows/UnitTests.cpp
```diff
@@ -5779,6 +5779,31 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
+    TEST_METHOD(ModernDistroInstall)
+    {
+        ModernDistroInstallImpl();
+    }
+
+    TEST_METHOD(ModernDistroInstallWithHiddenTempFolder)
+    {
+        // Avoid contaminating the real temp folder.
+        const auto testTempFolder = std::filesystem::temp_directory_path() / L"wsl-install-test";
+        std::filesystem::create_directories(testTempFolder);
```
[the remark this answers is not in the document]
chemwolf6922: The tests are not designed for parallel runs.

src/windows/common/wslutil.cpp
```diff
@@ -368,8 +368,17 @@ std::wstring wsl::windows::common::wslutil::DownloadFileImpl(
...
-        winrt::Windows::Storage::StorageFolder::GetFolderFromPathAsync(std::filesystem::temp_directory_path().wstring()).get();
+    // GetFolderFromPathAsync does not accept hidden or system path.
+    // To avoid the temp path having those attributes (mainly hidden),
+    // we need to create a subfolder without those attributes.
+    auto downloadFolderPath = std::filesystem::temp_directory_path() / L"wsl-downloads";
+    std::filesystem::create_directories(downloadFolderPath);
+
+    const auto attributes = GetFileAttributesW(downloadFolderPath.c_str());
+    THROW_LAST_ERROR_IF(attributes == INVALID_FILE_ATTRIBUTES);
+    THROW_IF_WIN32_BOOL_FALSE(SetFileAttributesW(downloadFolderPath.c_str(), attributes & ~(FILE_ATTRIBUTE_HIDDEN | FILE_ATTRIBUTE_SYSTEM)));
```
benhillis: minor: you can skip the call to SetFileAttributes if neither flag is set.

benhillis (commented):
Looks good maybe just drop the set when it is not needed.

src/windows/common/wslutil.cpp
```diff
@@ -368,8 +368,20 @@ std::wstring wsl::windows::common::wslutil::DownloadFileImpl(
...
+    // we need to create a subfolder without those attributes.
+    auto downloadFolderPath = std::filesystem::temp_directory_path() / L"wsl-downloads";
+    std::filesystem::create_directories(downloadFolderPath);
+
+    const auto attributes = GetFileAttributesW(downloadFolderPath.c_str());
+    THROW_LAST_ERROR_IF(attributes == INVALID_FILE_ATTRIBUTES);
+    if (attributes & (FILE_ATTRIBUTE_HIDDEN | FILE_ATTRIBUTE_SYSTEM))
+    {
+        THROW_IF_WIN32_BOOL_FALSE(SetFileAttributesW(downloadFolderPath.c_str(), attributes & ~(FILE_ATTRIBUTE_HIDDEN | FILE_ATTRIBUTE_SYSTEM)));
+    }
```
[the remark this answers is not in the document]
chemwolf6922: What's the gain in attacking that?

OneBlue (commented):
Thank you for doing this ! 

This is a really weird winrt behavior, but looks like there's no way to disable the attribute check, so let's just  work around it. 

The current implementation works, but in its current form, we'll always leak the `"wsl-downloads` folder, and adding logic to delete it would be a bit painful. 

Instead I would recommend constructing the path ourselves and then using [CreateRandomAccessStreamOnFile ](https://learn.microsoft.com/en-us/windows/win32/api/shcore/nf-shcore-createrandomaccessstreamonfile), which should allow us to open the file directly without the folder attribute issue.

OneBlue:
> Instead I would recommend constructing the path ourselves and then using [CreateRandomAccessStreamOnFile ](https://learn.microsoft.com/en-us/windows/win32/api/shcore/nf-shcore-createrandomaccessstreamonfile), which should allow us to open the file directly without the folder attribute issue.

Or even better, we could directly use [this](https://learn.microsoft.com/en-us/uwp/api/windows.storage.streams.filerandomaccessstream.openasync?view=winrt-28000#windows-storage-streams-filerandomaccessstream-openasync(system-string-windows-storage-fileaccessmode-windows-storage-storageopenoptions-windows-storage-streams-fileopendisposition)), that way we can set `openDisposition` and create the file ourselves

test/windows/Common.cpp
```diff
@@ -2474,14 +2474,32 @@ void Trim(std::wstring& string)
     std::erase_if(string, [](auto c) { return !isalnum(c); });
 }
 
-ScopedEnvVariable::ScopedEnvVariable(const std::wstring& Name, const std::wstring& Value) : m_name(Name)
+ScopedEnvVariable::ScopedEnvVariable(const std::wstring& Name, const std::wstring& Value, bool restore) : m_name(Name)
 {
+    if (restore)
```
OneBlue: nit: I think it be OK for this behavior to be the default, so we can probably remove `restore` here

test/windows/UnitTests.cpp
```diff
@@ -5852,6 +5852,31 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
     }
 
+    TEST_METHOD(ModernDistroInstall)
+    {
+        ModernDistroInstallImpl();
+    }
+
+    TEST_METHOD(ModernDistroInstallWithHiddenTempFolder)
+    {
+        // Avoid contaminating the real temp folder.
```
OneBlue: Given that the `ModernDistroInstall` test case is pretty long, I would recommend just validating that `DownloadFileImpl` works correctly when the temporary folder is hidden + system are, since that's what we want to validate here

OneBlue (commented):
LGTM, couple minor comments

src/windows/common/wslutil.cpp
```diff
@@ -368,14 +368,45 @@ std::wstring wsl::windows::common::wslutil::DownloadFileImpl(
...
+        try
+        {
+            outputStream = winrt::Windows::Storage::Streams::FileRandomAccessStream::OpenAsync(
+                               filePath,
+                               winrt::Windows::Storage::FileAccessMode::ReadWrite,
+                               winrt::Windows::Storage::StorageOpenOptions::None,
+                               winrt::Windows::Storage::Streams::FileOpenDisposition::CreateNew)
+                               .get();
+        }
+        catch (const winrt::hresult_error& e)
```
OneBlue: I recommend doing: 

```
        catch (...)
        {
            if (wil::ResultFromCaughtException() == HRESULT_FROM_WIN32(ERROR_FILE_EXISTS))
            {
            }
         }
```

test/windows/UnitTests.cpp
OneBlue: I also recommend adding test coverage for the case where the file already exists

src/windows/common/wslutil.cpp
```diff
@@ -368,14 +368,45 @@ std::wstring wsl::windows::common::wslutil::DownloadFileImpl(
...
-    const auto downloadFolder =
-        winrt::Windows::Storage::StorageFolder::GetFolderFromPathAsync(std::filesystem::temp_directory_path().wstring()).get();
-
-    const auto file =
-        downloadFolder.CreateFileAsync(Filename, winrt::Windows::Storage::CreationCollisionOption::GenerateUniqueName).get();
-    auto deleteFileOnFailure = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&] { file.DeleteAsync().get(); });
+    // GetFodlerFromPathAsync won't work if the folder is hidden or system.
+    auto downloadFolderPath = std::filesystem::temp_directory_path();
+    auto filenameStem = std::filesystem::path(Filename).stem().wstring();
+    auto filenameExtension = std::filesystem::path(Filename).extension().wstring();
```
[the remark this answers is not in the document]
chemwolf6922: Don't fight the user.

test/windows/UnitTests.cpp
```diff
@@ -7448,5 +7448,70 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
+        const auto testTempFolder = std::filesystem::temp_directory_path() / L"wsl-download-test";
+        std::filesystem::create_directories(testTempFolder);
+        auto cleanupTempFolder = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&] {
+            std::error_code error;
+            std::filesystem::remove_all(testTempFolder, error);
+        });
+
+        const auto originalAttributes = GetFileAttributesW(testTempFolder.c_str());
+        VERIFY_IS_TRUE(originalAttributes != INVALID_FILE_ATTRIBUTES);
+        VERIFY_IS_TRUE(SetFileAttributesW(testTempFolder.c_str(), originalAttributes | FILE_ATTRIBUTE_HIDDEN | FILE_ATTRIBUTE_SYSTEM));
```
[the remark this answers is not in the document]
chemwolf6922: That's not the true temp folder.

OneBlue (approved):

PR #40710, opened 2026-06-04
CLI: Add VT Support library and update tests/callbacks to use it

dkbennett:
This is a foundational addition of VT support classes, wrappers, and constants to make working with VT sequences consistent and simpler without having to deal with the actual sequences and restoration. It removes the old ChangeTerminalMode.h and rolls that into the new VTSupport.h and cpp. This is added under windows/common so it can be utilized by more than just the CLI, but the unit tests are in with the CLI.

This should also help alleviate some other pulls where VT sequences are being defined in multiple locations, and also serves as a foundation for enhanced VT support in the CLI.

## PR Checklist

* Add VT support classes to windows common
* Add unit tests for the VT classes and sequences.
* Update tests to use the new common VT support code.
* Update the BuildImageCallback and ImageProgressCallback to use the VT support classes instead of directly defined sequences.

## Validation Steps Performed
* Ran E2E tests, the Interactive TTY, Image Build and Pull tests to ensure they work as expected.
* Ran the new unit tests.
* Manual testing of the image build and image pull commands to verify the callback output is as expected.


src/windows/common/VTSupport.h
```diff
@@ -0,0 +1,530 @@
...
+            ConstructedSequence Extended(const Color& color);
+        }
+
+        ConstructedSequence Hyperlink(const std::string& text, const std::string& ref);
+
+        // Constructs a single SGR sequence with one or more semicolon-separated parameters.
+        // e.g. Sgr({1, 31}) produces "\x1b[1;31m" — bold + red in one sequence.
+        // Prefer named constants for single-parameter sequences; use this only when
+        // a multi-parameter form is required to match specific terminal output exactly.
+        ConstructedSequence Sgr(std::initializer_list<int> params);
```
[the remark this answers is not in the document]
dkbennett: removed this

src/windows/common/VTSupport.cpp
```diff
@@ -0,0 +1,355 @@
...
+    // the result part.  Reads up to s_bufferSize bytes non-blockingly; if the buffer
+    // fills completely the suffix may not be present, in which case an empty string
+    // is returned — the same outcome as any other parse failure.  Any bytes after
+    // the suffix (e.g. queued user input) are ignored.
+    std::string ExtractSequence(std::istream& inStream, std::string_view prefix, std::string_view suffix)
+    {
+        // Force discovery of available input
+        std::ignore = inStream.peek();
+
+        static constexpr std::streamsize s_bufferSize = 1024;
```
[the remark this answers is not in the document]
dkbennett: This is a theoretical concern but isn't reachable in practice in this code. I've updated the comment to remove the misleading "non-blockingly".

src/windows/common/VTSupport.h
```diff
@@ -0,0 +1,592 @@
...
+// the original cursor info on destruction. No-op if the handle is not a console.
+class ChangeTerminalMode
+{
+public:
+    NON_COPYABLE(ChangeTerminalMode);
+    NON_MOVABLE(ChangeTerminalMode);
+
+    ChangeTerminalMode(HANDLE console, bool cursorVisible) : m_console(console)
+    {
+        if (!wsl::windows::common::wslutil::IsConsoleHandle(console))
```
OneBlue: nit: Since we have a .cpp file, I would have a preference for moving the implementation of these methods to the .cpp, so we don't need to re-compile them every time they're included

src/windows/common/VTSupport.h
```diff
@@ -0,0 +1,592 @@
...
+        Paused,
+        Error
+    };
+
+    ConstructedSequence Construct(State state, std::optional<uint32_t> percentage = std::nullopt);
+} // namespace Progress
+
+// operator<< for stream output.
+// Widens the narrow sequence bytes (all ASCII) into a wide string for wostream output.
+inline std::wostream& operator<<(std::wostream& o, const Sequence& s)
```
OneBlue: I'm guessing the objective of those is to use std::wcout, but for our usecase I would recommend to keep using the `WriteConsoleW()` mecanism since it's significantly faster.

src/windows/wslc/services/ImageProgressCallback.cpp
```diff
@@ -19,18 +19,25 @@ Module Name:
 
 namespace wsl::windows::wslc::services {
 using namespace wsl::shared;
+using namespace wsl::windows::common::vt;
 
-auto ImageProgressCallback::MoveToLine(SHORT line)
+void ImageProgressCallback::WriteTerminal(std::wstring_view content) const
```
OneBlue, line 24: Given that we're already defining this method in BuildImageCallback, I'd recommend factoring this method and moving it out to `wslutil` .

At some point, I think WSL should also move to it, but let's do that in a future change

src/windows/common/VTSupport.h
OneBlue: I'm guessing that this was originally designed to use `std::wcout` to actually write to the console, but in wslc's case, I would recommend not using `std::wcout`, but directly calling into WriteTerminal.

Maybe we could simplify a bit by directly using `Get()` when we print the sequence to the terminal (and we could also use this in the tests).

That would allow us to get rid of operator<< + string / wstring methods (our version of std::format() automatically handles string <-> wstring conversions)

OneBlue (commented):
LGTM, I think there's a few things that we simplify, but this design looks great !

src/windows/common/VTSupport.h
dkbennett: I'm making an alteration to the calling code now that will support both, preferring WriteConsoleW where valid and that's handled at the higher that uses the sequences. I think its good to have the proper templates convert the sequences so they can be used as strings anywhere you'd use a string so we don't have to deal with implementation details of a sequence and it just works. The goal is that people can use VT sequences without having to know or care about the underlying sequence itself and to be able to construct it into strings naturally.

src/windows/wslc/services/ImageProgressCallback.cpp
dkbennett: I think there will probably be some refactoring to do. If we get the reporter stuff functional it is likely applicable to everything in the codebase beyond the cli and also be moved to windows common. I also expect to rewrite these handlers so everything in the CLI uses the same output mechanisms, but it isn't required yet. It seems like something we can do in the summer for general code tidiness and efficiency.

src/windows/common/VTSupport.h
dkbennett: It will write to output stream that is not wcout, and use WriteConsoleW() to flush the stream. We will absolutely use WriteConsoleW() for actual output. These operators are for ease of use and compat when working with sequences consistently in whatever way we may need to construct strings or streams with the sequences.

src/windows/common/VTSupport.h
dkbennett: I moved all eligible definitions to the .cpp to keep the header as simple as possible.

src/windows/common/VTSupport.h
dkbennett: I've updated the downstream code to not require stream output and removed the stream operator overloads.

dkbennett: Updated to remove stream operators.

src/windows/common/VTSupport.cpp
```diff
@@ -0,0 +1,551 @@
...
+{
+    return std::string{lhs.Get()} + rhs;
+}
+
+std::string operator+(const char* lhs, const Sequence& rhs)
+{
+    return lhs + std::string{rhs.Get()};
+}
+
+std::wstring operator+(const Sequence& lhs, const std::wstring& rhs)
```
OneBlue: I think we should decide to make all of our VT logic either work with wide strings, or narrow strings, so we don't need to implement both in each operator / util method

OneBlue (dismissed):
LGTM, one minor comment

OneBlue:
Sorry I'm not sure how I misclicked there. Reopening.

src/windows/common/VTSupport.cpp
```diff
@@ -0,0 +1,537 @@
...
+    DWORD current;
+    if (!GetConsoleMode(console, &current))
+    {
+        LOG_LAST_ERROR_IF(GetLastError() != ERROR_INVALID_HANDLE);
+        return;
+    }
+
+    if (mode == Mode::Input)
+    {
+        const DWORD newMode = (current & ~(ENABLE_LINE_INPUT | ENABLE_ECHO_INPUT)) | ENABLE_EXTENDED_FLAGS | ENABLE_VIRTUAL_TERMINAL_INPUT;
```
OneBlue, line 106: Sorry for not noticing this earlier, we should probably factor this with `wsl::windows::common::ConsoleState`. 

This duplication existed this change, so not a blocker, but looks like those two classes do very similar things

OneBlue (dismissed):

src/windows/common/VTSupport.cpp
dkbennett: There is partial overlap but enough differences that merging them is risky as part of this PR and will probably bloat it considerably, but definitely worth a re-examine in the near future.

OneBlue (approved):

PR #40713, opened 2026-06-04
Fix wslc image pull command when invoked in network logon session

yao-msft:
In network logon session, wincred store is not available. Treat it as "credential not exist" for anonymous pull operation to work. For write and delete operations, keep it as failed.

## PR Checklist

## Validation Steps Performed
Validated manually with 2 VMs setup, one invoking other via ssh, issue repros before this change and issue fixed after this change. It's quite complex to write an e2e test (start up 2 vm, setup ssh service and ssh client, create net login, etc) for it so I did not add test with this change.


OneBlue (approved):
LGTM ! 

As a small nit I think it would be a good idea to log a warning if we get ERROR_NO_SUCH_LOGON_SESSION back, since that could lead to some surprising behaviors for users

PR #40717, opened 2026-06-05
Fix MoveDistribution E_ACCESSDENIED when setVhdOwner fails under impersonation

### LINKED ISSUE #40716: `--move` fails to update Registry
### Windows Version

Microsoft Windows [Version 10.0.26220.8544]

### WSL Version

2.7.7.0

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.18.26.1-1

### Distro Version

Ubuntu 24.04

### Other Software

_No response_

### Repro Steps

Move the distro to a new location:
```sh
wsl --manage ubuntu-wp --move 'C:/wsl/ubuntu-wp'
```

### Expected Behavior

The move operation succeeds and the distro is moved.

### Actual Behavior

The move operation fails with the following message:
```sh
Access is denied.
Error code: Wsl/Service/MoveDistro/E_ACCESSDENIED
```

The distro's VHD is moved successfully, but the operation fails and the `BasePath` in Registry still points to the old location. Manually updating the `BasePath` makes the distro work from the new location.

### Diagnostic Logs

_No response_

benhillis:
## Summary

Fixes #40716 — `wsl --manage <distro> --move <path>` fails with `E_ACCESSDENIED` when the VHD owner becomes `BUILTIN\Administrators` after a cross-volume move.

## PR Checklist

- [x] Fixes a bug
- [x] Includes tests

## Detailed Description

After a cross-volume `MoveFileEx`, the new VHD file's owner may be set to `BUILTIN\Administrators` (Windows behavior for elevated callers). The `setVhdOwner` lambda was opening the moved file with `WRITE_OWNER` under user impersonation, which fails with `E_ACCESSDENIED` because the impersonated token doesn't have ownership-based access to the file.

The `wil::run_as_self()` call that grants SYSTEM privileges was placed **after** `CreateFileW` — too late.

**Fix:** Move `wil::run_as_self()` and `AcquirePrivilege(SE_RESTORE_NAME)` before `CreateFileW` so the file handle is opened as SYSTEM. `FILE_FLAG_OPEN_REPARSE_POINT` is retained to prevent symlink following.

Previously, this failure left the VHD physically moved but the registry `BasePath` still pointing at the old location, leaving the distro unusable.

## Validation Steps

1. Full build: `cmake . && cmake --build . -- -m`
2. Run new test: `bin\x64\debug\test.bat /name:*MoveVhdWithAdminOwner*` — **PASSED**
3. Test imports a distro, sets VHD owner to Administrators, then moves as non-elevated user. Verifies move succeeds, ownership is preserved, and distro still launches.

### DIFF
--- a/src/windows/service/exe/LxssUserSession.cpp
+++ b/src/windows/service/exe/LxssUserSession.cpp
@@ -952,13 +952,17 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
     THROW_IF_WIN32_BOOL_FALSE(MoveFileEx(distro.VhdFilePath.c_str(), newVhdPath.c_str(), MOVEFILE_COPY_ALLOWED | MOVEFILE_WRITE_THROUGH));
 
     // Restore the original VHD owner on the moved file.
+    // Run as self (SYSTEM) for both the file open and the SetSecurityInfo call,
+    // because after a cross-volume MoveFileEx the new file's owner may be
+    // BUILTIN\Administrators and the impersonated user token may lack WRITE_OWNER.
     auto setVhdOwner = [&originalOwner](const std::filesystem::path& vhdPath) {
+        auto runAsSelf = wil::run_as_self();
+        auto privileges = wsl::windows::common::security::AcquirePrivilege(SE_RESTORE_NAME);
+
         wil::unique_hfile vhdHandle(CreateFileW(
             vhdPath.c_str(), WRITE_OWNER, FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr, OPEN_EXISTING, FILE_FLAG_OPEN_REPARSE_POINT, nullptr));
         THROW_LAST_ERROR_IF(!vhdHandle);
 
-        auto runAsSelf = wil::run_as_self();
-        auto privileges = wsl::windows::common::security::AcquirePrivilege(SE_RESTORE_NAME);
         THROW_IF_WIN32_ERROR(
             ::SetSecurityInfo(vhdHandle.get(), SE_FILE_OBJECT, OWNER_SECURITY_INFORMATION, originalOwner, nullptr, nullptr, nullptr));
     };
--- a/test/windows/UnitTests.cpp
+++ b/test/windows/UnitTests.cpp
@@ -3048,6 +3048,79 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
         }
     }
 
+    WSL2_TEST_METHOD(MoveVhdWithAdminOwner)
+    {
+        // Regression test for #40716: if the VHD's owner is BUILTIN\Administrators
+        // (as happens after a cross-volume MoveFileEx from an elevated context),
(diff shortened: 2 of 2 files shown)


test/windows/UnitTests.cpp
```diff
@@ -3048,6 +3048,79 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
...
+        // Move to first folder so we know where the VHD is.
+        WslShutdown();
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(std::format(L"--manage {} --move {}", name, firstFolder)), 0L);
+
+        auto vhdPath = std::format(L"{}\\ext4.vhdx", firstFolder);
+        VERIFY_IS_TRUE(std::filesystem::exists(vhdPath));
+
+        // Simulate cross-volume MoveFileEx side-effect: change VHD owner to BUILTIN\Administrators.
+        {
+            BYTE adminsSidBuffer[SECURITY_MAX_SID_SIZE];
```
OneBlue, line 3078: nit: This can be simplified by using:  `wsl::windows::common::security::CreateSid` (and factored since we do the same thing below)

src/windows/service/exe/LxssUserSession.cpp
```diff
@@ -952,13 +952,17 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
     THROW_IF_WIN32_BOOL_FALSE(MoveFileEx(distro.VhdFilePath.c_str(), newVhdPath.c_str(), MOVEFILE_COPY_ALLOWED | MOVEFILE_WRITE_THROUGH));
 
     // Restore the original VHD owner on the moved file.
+    // Run as self (SYSTEM) for both the file open and the SetSecurityInfo call,
+    // because after a cross-volume MoveFileEx the new file's owner may be
+    // BUILTIN\Administrators and the impersonated user token may lack WRITE_OWNER.
     auto setVhdOwner = [&originalOwner](const std::filesystem::path& vhdPath) {
+        auto runAsSelf = wil::run_as_self();
```
OneBlue, line 959: Unfortunately I don't think we can do this because this could be exploited to change the owner of a file. For instance, a user could do: 

```
wsl --move <distro> <folder>
```

And while the move is running, immediately after `MoveFileEx()` returns: 
```
del <folder>\ext4.vhdx
mklink <folder>\ext4.vhdx <target-file>
```

Because we set the owner to whatever `ext4.vhdx`, if a user can get the timing right, this could be exploited to change the owner of <target-file> to whoever the original owner of the vhdx is, which the user controls

OneBlue (changes requested):

src/windows/service/exe/LxssUserSession.cpp
OneBlue: The only way I can think of working around this would be to disallow copies when we call `MoveFileEx()` and if the can't be moved because the target is in a different location, copy it manually and set the file permissions when we create the handle, which we can do while impersonating

OneBlue (approved):
Scratch my previous review, looks like the `FILE_FLAG_OPEN_REPARSE_POINT` flag already protects us from that

PR #40720, opened 2026-06-05
Add --network option to container create/run (excluding host mode)

AmelBawa-msft:

Adds `--network` support to `container run` and `container create`, letting users connect containers to one or more networks.  
`--network host` is explicitly rejected with the error: `host mode networking is not supported`, and this message is localized.

## PR Checklist

- Added validation so if any `--network` value is `host`, command execution fails with:
  - `host mode networking is not supported`
- Localized the host-mode rejection error by replacing the hardcoded exception text with a localization resource string.
- Updated existing E2E tests for `container create` and `container run` host-mode cases to expect failure.
- Added an E2E test to verify host-mode is rejected when multiple `--network` options are provided and one of them is `host`.
- Updated CLI unit tests `CreateCommand_ParseNetworkSingleValue_SetsNetwork` and `RunCommand_ParseNetworkSingleValue_SetsNetwork` to use a supported single network value (`net1`) now that `host` is rejected.
- Ran clang-format on modified files.

## Validation Steps Performed

- Ran `clang-format -i` on changed files.
- Verified formatting with `clang-format --dry-run --style=file` on changed files.
- Ran `python3 tools/devops/validate-localization.py`.
- Ran final validation tooling: code review returned no comments; CodeQL was skipped as trivial on the latest test-only follow-up change.
- Checked workflow/job status for the branch (no failed jobs found in the current run).
- Attempted environment validation commands available in this sandbox; full Windows build/test execution was not available in this environment.


test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
```diff
@@ -768,10 +770,61 @@ class WSLCE2EContainerCreateTests
...
+        result.Verify({.Stderr = L"", .ExitCode = 0});
+
+        const auto inspect = InspectContainer(WslcContainerName);
+        VERIFY_ARE_EQUAL(std::string("bridge"), inspect.HostConfig.NetworkMode);
+    }
+
+    WSLC_TEST_METHOD(WSLCE2E_Container_Create_Network_HostMode)
+    {
+        auto result =
+            RunWslc(std::format(L"container create --name {} --network host {} true", WslcContainerName, DebianImage.NameAndTag()));
```
OneBlue, line 785: FYI @beena352 : I think the change to strings added support for `--network host`, but I don't think we should expose it, since our current implementation of host mode is likely to confuse users. 

@AmelBawa-msft: For now, could fail to create a container in host mode with something like `host mode networking is not supported` ?

OneBlue (changes requested):

test/windows/wslc/e2e/WSLCE2EContainerCreateTests.cpp
AmelBawa-msft: @copilot check if one of the network option values is host, then throw an exception saying, "host mode networking is not supported". Update existing E2E tests correspondingly and add a test for checking that host throws an exception. Do not change anything else and run clang format to make sure the code formatting is correct.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -101,6 +101,11 @@ void Argument::Validate(const ArgMap& execArgs) const
             {
                 throw ArgumentException(Localization::WSLCCLI_NetworkEmptyError(m_name));
             }
+
+            if (IsEqual(value, L"host", true))
+            {
+                throw ArgumentException(L"host mode networking is not supported");
```
OneBlue: This string should be localized

AmelBawa-msft: @copilot localize the exception message. Don't change anything else.

test/windows/wslc/WSLCCLIExecutionUnitTests.cpp
```diff
@@ -345,6 +345,106 @@ class WSLCCLIExecutionUnitTests
...
+        CLIExecutionContext context;
+        command.ParseArguments(invocation, context.Args);
+
+        VERIFY_THROWS_SPECIFIC(
+            command.ValidateArguments(context.Args), wsl::windows::wslc::ArgumentException, [](const auto&) { return true; });
+    }
+
+    TEST_METHOD(CreateCommand_ParseNetworkSingleValue_SetsNetwork)
+    {
+        auto invocation = CreateInvocationFromCommandLine(L"wslc --network host ubuntu sh");
```
AmelBawa-msft: @copilot this will now fail because `host` is not supported after the latest changes. Update "CreateCommand_ParseNetworkSingleValue_SetsNetwork" and "RunCommand_ParseNetworkSingleValue_SetsNetwork" tests. Don't make other changes.

OneBlue (approved):

PR #40722, opened 2026-06-05
Fix various issues around initial tty sizes + add test coverage

OneBlue:

This change fixes various codepaths were the initial tty size of a command wouldn't be correctly set. 

To change this properly, the tty size is now passed when a container is started, not created. This is required to properly support the `wslc create`, followed by `wslc start -a` case, since the tty size might be different between the create and start call. 

## PR Checklist

### DIFF
--- a/test/windows/wslc/e2e/WSLCE2EContainerAttachTests.cpp
+++ b/test/windows/wslc/e2e/WSLCE2EContainerAttachTests.cpp
@@ -57,15 +57,13 @@ class WSLCE2EContainerAttachTests
         result.Verify({.Stderr = L"", .ExitCode = 0});
         auto containerId = result.GetStdoutOneLine();
 
-        const auto& expectedAttachPrompt = VT::BuildContainerAttachPrompt(prompt);
         const auto& expectedPrompt = VT::BuildContainerPrompt(prompt);
 
         auto session = RunWslcInteractive(std::format(L"container attach {}", containerId));
         VERIFY_IS_TRUE(session.IsRunning(), L"Container session should be running");
 
-        // The container attach prompt appears twice.
-        session.ExpectStdout(expectedAttachPrompt);
-        session.ExpectStdout(expectedAttachPrompt);
+        // Ignore resize-repaint messages. Those are emitted when the the tty initial size is set, which can happen before or after we start running commands.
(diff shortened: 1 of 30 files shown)


test/windows/wslc/e2e/WSLCE2EContainerAttachTests.cpp
```diff
@@ -57,15 +57,13 @@ class WSLCE2EContainerAttachTests
...
-        const auto& expectedAttachPrompt = VT::BuildContainerAttachPrompt(prompt);
         const auto& expectedPrompt = VT::BuildContainerPrompt(prompt);
 
         auto session = RunWslcInteractive(std::format(L"container attach {}", containerId));
         VERIFY_IS_TRUE(session.IsRunning(), L"Container session should be running");
 
-        // The container attach prompt appears twice.
-        session.ExpectStdout(expectedAttachPrompt);
-        session.ExpectStdout(expectedAttachPrompt);
+        // Ignore resize-repaint messages. Those are emitted when the the tty initial size is set, which can happen before or after we start running commands.
```
dkbennett, line 65: This solves my mystery of why the attach prompt is sent twice!

test/windows/wslc/e2e/WSLCExecutor.cpp
```diff
@@ -313,12 +356,34 @@ WSLCInteractiveSession::~WSLCInteractiveSession()
 
 void WSLCInteractiveSession::ExpectStdout(const std::string& expected)
 {
+    if (m_ignoreSequence.has_value())
```
dkbennett, line 359: OK I see, the resize sequence can appear at any time in the console output, as it is resized, so those sequences are ignored so we can skip to the parts that matter. So setting ignore sequence on the prompt allows us to skip one or more duplicate prompts that may appear in the event of resizing happening.

I would expect a future improvement might be to make ignoreSequence a vector of sequences to ignore and if we find a match for any one of them we consume and move on, but as we only need one right now, or can update it later in the test as needed, this is good for current needs.

dkbennett (approved):
I don't expect this will cause any problems with VT related things in the pipeline, as its just a swap from test-defined sequences to common library, the actual tests themselves arent' really changed so this "_should just work_" (famous last words!). 

Looks like some nice improvements to interactive sessions to support pseudoconsoles and ignoring sequences which we can expand upon further as needed in the future!

PR #40725, opened 2026-06-05
Fix relay loops spinning at 100% CPU on synchronous EOF

### LINKED ISSUE #40651: wsl.exe relay loops in `src/windows/common/relay.cpp` miss `bytesRead == 0` EOF check on synchronous-success path; cause `wsl.exe` to spin at 100% on AFD sockets
### Windows Version

Microsoft Windows [Version 10.0.26100.8390]

### WSL Version

2.7.3.0

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.114.1-1

### Distro Version

Ubuntu 24.04

### Other Software

Powershell 7.6.3

### Repro Steps

Start powershell
start wsl
keep working on that window multiple hours

### Expected Behavior

CPU usage correlated with cpu usage inside WSL

### Actual Behavior

CPU usage in WSL ~0%, Windows wsl exe cpu usage ~10-15% (2 threads at 100%) 

### Diagnostic Logs

 long-lived `wsl.exe` relay process (the one bridging stdio / interop between
a Windows console and the WSL VM) eventually enters a state where two of its
worker threads spin in a tight `ReadFile` loop on `\Device\Afd` socket
handles. From that point on the process consumes ~70% CPU per spinning thread
(~140% combined, ~6–10% of total system CPU on a 16-thread box) indefinitely,
even when WSL itself is idle.

Source review of `src/windows/common/relay.cpp` at release tag 2.7.3 shows
that **four** distinct `ReadFile` call sites all share the same defect: on
the synchronous-success branch (i.e. `ReadFile` returns `TRUE` instead of
failing with `ERROR_IO_PENDING`), none of them check `lpNumberOfBytesRead`
for zero. Per Microsoft's own Winsock documentation, a successful zero-byte
read on a stream socket is **the** canonical peer-FIN / EOF signal — it
does not surface as an error code. The result is that once the WSL VM side
of an hvsock connection closes, the affected relay loop re-issues the same
read against the EOF'd handle forever.

## Environment

| Component | Version |
|---|---|
| Windows | 11 Enterprise, 10.0.26100.8390 |
| WSL | 2.7.3.0 (also observed on 2.6.2.0) |
| WSLg | 1.0.66 |
| WSL kernel | 6.6.114.1-1 (Microsoft) |
| Distro | Ubuntu (default WSL2 install) |

## Symptom

- One specific `wsl.exe` process (the per-console relay, not `wslservice` or
  `vmmemWSL`) shows constant 6–10% CPU on the Windows host.
- Inside the WSL VM, all processes are idle (`top` shows ~0% CPU).
- The process is not stuck — interactive use through it still works — it is
  simply burning CPU in the background continuously.
- Behavior persists for the entire lifetime of the affected `wsl.exe` (observed
  in one case for over 4 hours of accumulated CPU time across two threads).
- The condition does **not** clear on its own and only ends when the affected
  `wsl.exe` is killed (which terminates the corresponding Windows console
  session).

### What we observe

Both spinning threads share these characteristics, sampled repeatedly seconds
apart:

- Win32 thread start address at the same offset inside `wsl.exe` (the entry
  point shared by both Interop and IO Relay worker threads).
- Instruction pointer pinned at `ntdll!ZwReadFile+0x14` — the `syscall`
  instruction inside the system service stub.
- The handle being read is of type `\Device\Afd` (Winsock kernel object —
  the user-mode endpoint of a Hyper-V socket bridging to the WSL VM).
- Disassembly around the call site shows a `ReadFile` call (via IAT thunk)
  whose `BOOL` return value is checked, but whose `lpNumberOfBytesRead` is
  not, on the synchronous-success path. The error-path branches do test
  `ERROR_BROKEN_PIPE` (109) and `ERROR_HANDLE_EOF` (38), but the
  success-with-zero-bytes case appears to flow back into the read loop /
  wait state without an EOF break.

| # | Function | Sync-TRUE EOF handling |
|---|---|---|
| 1 | `wsl::windows::common::relay::InterruptableRead` | Returns `bytesRead` to caller; depends on the caller to interpret 0 as EOF. |
| 2 | `BidirectionalRelay` — left side | Sets `leftReadPending = true` and relies on the later `GetOverlappedResult` branch (which *does* check `leftBytesRead == 0`) to detect EOF. |
| 3 | `BidirectionalRelay` — right side | Same pattern as the left side. |
| 4 | `ScopedMultiRelay::Run` | Calls `Write(gsl::make_span(data, BytesRead))` (a no-op when `BytesRead == 0`), sets state to `Standby`, and the outer loop re-issues `ReadFile` on the same handle. EOF is only recognized via `ERROR_HANDLE_EOF` / `ERROR_BROKEN_PIPE` on the failure branch. |

The synchronous-success-with-zero-bytes case is exactly what a Winsock
graceful FIN produces on an AFD socket handle (cited above). Under that
condition:

- For `ScopedMultiRelay::Run` the loop immediately re-enters `ReadFile` on
  the EOF'd handle, which returns `TRUE` + zero bytes again, producing a
  pure CPU-bound spin.
- For `BidirectionalRelay` the path is more subtle: the sync-success branch
  goes to `WaitForMultipleObjects`. If the overlapped event is set
  appropriately on sync completion, `GetOverlappedResult` will pick up the
  zero-byte read and tear the handle down. If the event is *not* set on
  sync completion, or if there is any state where the wait keeps returning
  immediately without `GetOverlappedResult` advancing, the same EOF is
  consumed repeatedly and the result is a spin.
- For `InterruptableRead` the spin only manifests if its caller fails to
  treat a returned `bytesRead == 0` as EOF (some callers in the codebase
  do; if any do not, that path spins).

The two-threads-per-process spinning pattern we observe is consistent with
any of these — most directly with `BidirectionalRelay` (one per direction)
or `ScopedMultiRelay::Run` running two relays in parallel. We cannot pin
the spin to one specific function without the `wsl.exe` 

## Workaround
Open new powershell and run wsl, then close "original" powershell with cpu problem.

benhillis:
## Summary

Fixes wsl.exe relay threads spinning at 100% CPU after the peer (VM-side hvsocket) closes gracefully.

## Problem

When `ReadFile` completes synchronously with 0 bytes (the canonical EOF signal on stream sockets and pipes), three relay code paths failed to recognize this as EOF:

- `BidirectionalRelay` — used by `wslrelay.exe` for localhost forwarding and socket relay
- `ScopedMultiRelay::Run` — used for set-version stderr relay

Instead of breaking out of the loop, they re-issued `ReadFile` on the EOF'd handle, which returned TRUE + 0 bytes again immediately, producing a pure CPU-bound spin.

## Fix

Add `bytesRead == 0` checks on the synchronous-success path in:
- `BidirectionalRelay` (both left and right read sites)
- `ScopedMultiRelay::Run` (both the sync-completion and overlapped-completion paths)

## Validation

Added `RelayEofDetection` unit test that creates pipe pairs, closes the write end (producing synchronous EOF), and verifies each relay function terminates within 5 seconds rather than spinning.

Fixes #40651

### DIFF
--- a/src/windows/common/relay.cpp
+++ b/src/windows/common/relay.cpp
@@ -327,9 +327,22 @@ void wsl::windows::common::relay::BidirectionalRelay(_In_ HANDLE LeftHandle, _In
             if (!ReadFile(LeftHandle, leftReadSpan.data(), gsl::narrow_cast<DWORD>(leftReadSpan.size()), &leftBytesRead, &leftOverlapped))
             {
                 THROW_LAST_ERROR_IF(GetLastError() != ERROR_IO_PENDING);
+                leftReadPending = true;
             }
+            else if (leftBytesRead == 0)
+            {
+                LeftHandle = nullptr;
+                if (WI_IsFlagSet(Flags, RelayFlags::RightIsSocket))
+                {
+                    LOG_LAST_ERROR_IF(shutdown(reinterpret_cast<SOCKET>(RightHandle), SD_SEND) == SOCKET_ERROR);
+                }
 
-            leftReadPending = true;
+                continue;
+            }
+            else
+            {
+                leftReadPending = true;
+            }
         }
 
         DWORD rightBytesRead = 0;
@@ -338,9 +351,22 @@ void wsl::windows::common::relay::BidirectionalRelay(_In_ HANDLE LeftHandle, _In
             if (!ReadFile(RightHandle, rightReadSpan.data(), gsl::narrow_cast<DWORD>(rightReadSpan.size()), &rightBytesRead, &rightOverlapped))
             {
                 THROW_LAST_ERROR_IF(GetLastError() != ERROR_IO_PENDING);
+                rightReadPending = true;
             }
+            else if (rightBytesRead == 0)
+            {
+                RightHandle = nullptr;
+                if (WI_IsFlagSet(Flags, RelayFlags::LeftIsSocket))
+                {
+                    LOG_LAST_ERROR_IF(shutdown(reinterpret_cast<SOCKET>(LeftHandle), SD_SEND) == SOCKET_ERROR);
+                }
 
-            rightReadPending = true;
+                continue;
+            }
+            else
+            {
+                rightReadPending = true;
+            }
         }
 
         const DWORD waitResult = WaitForMultipleObjects(RTL_NUMBER_OF(waitObjects), waitObjects, FALSE, INFINITE);
@@ -917,6 +943,11 @@ try
 
                         THROW_LAST_ERROR_IF(lastError != ERROR_IO_PENDING);
                     }
+                    else if (Transferred == 0)
+                    {
+                        e.State = Eof;
+                        continue;
+                    }
 
                     // IO is available.
                     Write(i, gsl::make_span(e.Buffer.data(), Transferred));
@@ -938,6 +969,12 @@ try
                 DWORD BytesRead{};
                 if (ReadFile(e.Handle, e.Buffer.data(), static_cast<DWORD>(e.Buffer.size()), &BytesRead, &e.Overlapped))
                 {
+                    if (BytesRead == 0)
+                    {
+                        e.State = Eof;
+                        continue;
+                    }
+
                     // IO is available.
                     Write(i, gsl::make_span(e.Buffer.data(), BytesRead));
 
--- a/src/windows/common/relay.cpp
+++ b/src/windows/common/relay.cpp
@@ -324,12 +324,30 @@ void wsl::windows::common::relay::BidirectionalRelay(_In_ HANDLE LeftHandle, _In
         DWORD leftBytesRead = 0;
         if (!leftReadPending && LeftHandle)
         {
+            bool leftEof = false;
             if (!ReadFile(LeftHandle, leftReadSpan.data(), gsl::narrow_cast<DWORD>(leftReadSpan.size()), &leftBytesRead, &leftOverlapped))
             {
-                THROW_LAST_ERROR_IF(GetLastError() != ERROR_IO_PENDING);
-                leftReadPending = true;
+                const auto lastError = GetLastError();
+                if ((lastError == ERROR_HANDLE_EOF) || (lastError == ERROR_BROKEN_PIPE))
+                {
+                    leftEof = true;
+                }
+                else
+                {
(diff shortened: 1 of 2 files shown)


src/windows/common/relay.cpp
```diff
@@ -918,6 +944,12 @@ try
                         THROW_LAST_ERROR_IF(lastError != ERROR_IO_PENDING);
                     }
 
+                    if (Transferred == 0)
+                    {
+                        e.State = Eof;
+                        continue;
+                    }
```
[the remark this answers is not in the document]
OneBlue: I don't think this is true ? `Transferred ` is only checked when `GetOverlappedResult()` returns success

OneBlue (dismissed):
LGTM. I think at some point we should move those to use RelayHandle and factor all the IO scheduling in one place, but for now let's merge this

benhillis:
> LGTM. I think at some point we should move those to use RelayHandle and factor all the IO scheduling in one place, but for now let's merge this

yep I considered doing that as part of this change, but we can do it as a follow-up.

src/windows/common/relay.cpp
```diff
@@ -324,23 +324,77 @@ void wsl::windows::common::relay::BidirectionalRelay(_In_ HANDLE LeftHandle, _In
...
+                    leftReadPending = true;
+                }
+            }
+            else if (leftBytesRead == 0)
+            {
+                leftEof = true;
+            }
+            else
+            {
+                leftReadPending = true;
```
OneBlue, line 347: Shouldn't we set this to false here ? That block would only be hit if ReadFile() completes synchronously, right ?

OneBlue: This might make a good case for moving to MultiHandleWait for this relay (it has the capability to shutdown sockets: [example](https://github.com/microsoft/WSL/blob/d1451ca74ee4a59c9afefacf449ea50541125eb4/src/windows/wslcsession/WSLCContainer.cpp#L710))

benhillis:
Closing as redundant: master's relay.cpp already handles synchronous (0-byte) EOF in both affected paths after #40814 ("Use overlapped IO when reading from the console"), which rewrote BidirectionalRelay to overlapped I/O (nulls the handle on a 0-byte read) and ScopedMultiRelay::Run to break on a 0-byte InterruptableRead. The 100% CPU spin can no longer occur, so this fix is no longer needed.

PR #40754, opened 2026-06-09
Fix WSLc telemetry to fire critical events

craigloewen-msft:
Changed WSLc events to be the exact same as the "ExecCritical" event. Right now it is showing up as "Measure+Critical" this change should make it show up just as "Critical"

### DIFF
--- a/src/windows/service/exe/WSLCSessionManager.cpp
+++ b/src/windows/service/exe/WSLCSessionManager.cpp
@@ -322,9 +322,9 @@ void WSLCSessionManagerImpl::CreateSession(
     });
 
     // This telemetry event is used to keep track of session creation performance (via CreationTimeMs) and failure reasons (via Result).
-    WSL_LOG_TELEMETRY(
+    WSL_LOG(
         "WSLCCreateSession",
-        PDT_ProductAndServicePerformance,
+        TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage),
         TraceLoggingKeyword(MICROSOFT_KEYWORD_CRITICAL_DATA),
         TraceLoggingValue(resolvedDisplayName.c_str(), "Name"),
         TraceLoggingValue(stopWatch.ElapsedMilliseconds(), "CreationTimeMs"),
--- a/src/windows/service/exe/WSLCSessionManager.cpp
+++ b/src/windows/service/exe/WSLCSessionManager.cpp
@@ -327,6 +327,7 @@ void WSLCSessionManagerImpl::CreateSession(
         TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage),
         TraceLoggingKeyword(MICROSOFT_KEYWORD_CRITICAL_DATA),
         TraceLoggingValue(resolvedDisplayName.c_str(), "Name"),
+        TraceLoggingValue(WSL_PACKAGE_VERSION, "wslVersion"),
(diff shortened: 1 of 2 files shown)


OneBlue (dismissed):

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -322,9 +322,9 @@ void WSLCSessionManagerImpl::CreateSession(
     });
 
     // This telemetry event is used to keep track of session creation performance (via CreationTimeMs) and failure reasons (via Result).
```
ptrivedi, line 324: This comment does not align with what is being done below. We are changing the tag from perf to usage. Also, WSL_LOG will not have wslVersion injected. Is this the intention?

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1699,9 +1699,9 @@ try
     auto result = wil::ResultFromException([&]() { CreateContainerImpl(containerOptions, Container); });
 
     // This telemetry event is used to keep track of the container creation failure rate and surface unexpected errors.
-    WSL_LOG_TELEMETRY(
+    WSL_LOG(
         "WSLCCreateContainer",
-        PDT_ProductAndServicePerformance,
+        TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage),
```
ptrivedi, line 1704: If there was an attempt to create a container, regardless of whether the attempt succeeds, would we want to update the usage metric?

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -322,9 +322,9 @@ void WSLCSessionManagerImpl::CreateSession(
...
-    WSL_LOG_TELEMETRY(
+    WSL_LOG(
         "WSLCCreateSession",
-        PDT_ProductAndServicePerformance,
+        TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage),
         TraceLoggingKeyword(MICROSOFT_KEYWORD_CRITICAL_DATA),
         TraceLoggingValue(resolvedDisplayName.c_str(), "Name"),
```
ptrivedi, line 329: I see now why you changed to WSL_LOG... WSL_LOG_TELEMETRY does not allow for CRITICAL_DATA only. In this case, you should inject the version logging. Something like:


 WSL_LOG(
     "WSLCCreateSession",
     TraceLoggingValue(WSL_PACKAGE_VERSION, "wslVersion"),    
     TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage), 
     TraceLoggingKeyword(MICROSOFT_KEYWORD_CRITICAL_DATA),
...
...
...

src/windows/service/exe/WSLCSessionManager.cpp
craigloewen-msft: I made it match exactly what we do for 'ExecCritical' which shows up correctly as "Critical" so as far as I can tell this would line up as a critical event.

src/windows/wslcsession/WSLCSession.cpp
craigloewen-msft: The "Result" value should give whether it started up or not, so yes this should give us the right data

src/windows/service/exe/WSLCSessionManager.cpp
craigloewen-msft: Great catch! Updated this

OneBlue (approved):

PR #40767, opened 2026-06-10
Use event instead of termination callback

kvega005:

Reworks how WSLC session termination is surfaced to clients. The previous design used a push-style `ITerminationCallback` that could only be subscribed at session-creation time (via `WSLCSessionSettings`), so it was unavailable when *opening* an existing session. This replaces it with a pull model that mirrors the process-exit-event pattern: a one-off "terminated" event plus a cached termination reason that any holder of an `IWSLCSession` can query — whether the session was created or opened.

## PR Checklist

- [x] **Communication:** I've discussed this with core contributors already.
- [x] **Localization:** All end user facing strings can be localized (no new user-facing strings)
- [ ] **Documentation updated:**

**Two distinct lifecycle events on the session:**
- `m_sessionTerminatingEvent` (existing) — signaled at the *start* of teardown to cancel in-flight operations.
- `m_sessionTerminatedEvent` (new) — signaled once teardown is *complete* ("the session is done"), via either an explicit `Terminate()` or an unexpected VM exit. This is the event handed to clients.

**Runtime (service / per-user session):**
- `IWSLCSession` gains `GetTerminationEvent` (returns a duplicated, caller-owned handle to the terminated event that stays valid after the session is released) and `GetTerminationReason` (returns the cached reason + details; fails with `ERROR_INVALID_STATE` until the session has actually terminated, so callers can distinguish "still running" from "terminated for an unknown reason").
- `HcsVirtualMachine` (service) computes the termination reason from the HCS exit status in `OnExit` and caches it; `IWSLCVirtualMachine` gains `GetTerminationReason`.
- `WSLCSession::Terminate()` captures the reason once: it defaults to `Shutdown` for an explicit terminate (VM still alive = we shut it down), and pulls the real crash/shutdown reason from the VM when the VM exited on its own. Capture happens under the exclusive lock before the VM reference is released, and the terminated event is signaled last so any observer sees a fully torn-down session.
- `m_terminated` bool removed — `m_sessionTerminatedEvent.is_signaled()` is now the single source of truth for `GetState`/`GetTerminationReason`.

**SDK (`wslcsdk`):**
- Removed `WslcSetSessionSettingsTerminationCallback` and the `ITerminationCallback`/`TerminationCallback` plumbing (files, CMake, `.def`, WiX proxy/stub registration).
- Added `WslcGetSessionTerminationEvent` and `WslcGetSessionTerminationReason`.
- `WslcSessionOptionsInternal` shrank (removed callback + context fields); opaque `WSLC_SESSION_OPTIONS_SIZE` updated accordingly.

**IDL:** removed the `ITerminationCallback` interface and the `TerminationCallback` field from `WSLCSessionSettings`; added the new methods to `IWSLCSession` and `IWSLCVirtualMachine`.

## Validation Steps Performed

- Updated runtime tests: `WSLCTests.cpp::TerminationEvent` waits on `IWSLCSession::GetTerminationEvent`, verifies `GetTerminationReason` returns `ERROR_INVALID_STATE` before termination and `Shutdown` after an explicit `Terminate()`.
- Updated SDK tests: `WslcSdkTests.cpp::TerminationEventViaTerminate` and `TerminationEventViaRelease` wait on the SDK-returned event (including verifying the handle stays valid after the session is released) and check the reason via `WslcGetSessionTerminationReason`.


src/windows/WslcSDK/winrt/Session.cpp
```diff
@@ -45,19 +45,35 @@ Session::Session(winrt::Microsoft::WSL::Containers::SessionSettings const& setti
     }
 }
 
+Session::~Session()
+{
+    // Stop watching the termination handle and drain any in-flight callback before the rest of the
+    // session is torn down. Releasing the session handle below can signal the termination event, so the
+    // wait must be cancelled first to avoid the callback running against a half-destructed object.
+    if (m_terminationWait)
```
OneBlue: I don't think we explicitly need to do this since `m_terminationWait` will do that for us and it always gets destroyed before the session gets released

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2958,10 +2970,43 @@ HRESULT WSLCSession::GetState(_Out_ WSLCSessionState* State)
...
 
+HRESULT WSLCSession::GetTerminationEvent(_Out_ HANDLE* Event)
+try
+{
+    RETURN_HR_IF(E_POINTER, Event == nullptr);
+
+    *Event = nullptr;
+
+    // Duplicate the "terminated" event. The caller owns the returned handle, which stays valid even after the session is released.
+    *Event = wsl::windows::common::wslutil::DuplicateHandle(m_sessionTerminatedEvent.get());
```
OneBlue: I recommend limiting the handle's permission to "SYNCHRONIZE" here, so a rogue caller can't signal the termination event

src/windows/service/inc/wslc.idl
```diff
@@ -531,6 +521,11 @@ interface IWSLCVirtualMachine : IUnknown
 
     // Returns an event that is signaled when the VM exits (graceful or forced).
     HRESULT GetTerminationEvent([out, system_handle(sh_event)] HANDLE* Event);
+
+    // Returns the cached termination reason and details. The values are only meaningful
+    // after the termination event has been signaled; before that the reason is
+    // WSLCVirtualMachineTerminationReasonUnknown and Details is an empty string.
+    HRESULT GetTerminationReason([out] WSLCVirtualMachineTerminationReason* Reason, [out] LPWSTR* Details);
```
OneBlue, line 532: Long term I think we should parse the `Details` into fields so the caller doesn't have to deal with HCS's json format. 

But I think this is fine for now, since this is already a major improvement on the previous design

OneBlue (commented):
Looks like there are some build errors, but overall design LGTM

src/windows/service/inc/wslc.idl
kvega005: I started making the change, but it ends up adding a bunch of new structs and touches a lot of files, so, I think will do this in a follow up PR.  I'm confident I can get it in before private preview.

src/windows/WslcSDK/winrt/Session.cpp
```diff
@@ -300,11 +304,15 @@ WslcSession Session::ToHandle()
...
 {
     try
     {
         auto session = static_cast<Session*>(context);
+
+        WslcSessionTerminationReason reason = WSLC_SESSION_TERMINATION_REASON_UNKNOWN;
+        LOG_IF_FAILED(WslcGetSessionTerminationReason(session->m_session.get(), &reason));
+
         session->m_terminatedEvent(static_cast<SessionTerminationReason>(reason));
     }
```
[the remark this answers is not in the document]
OneBlue: This is not true because Session's destructor will destroy m_terminationWait before releasing the session

OneBlue (approved):

PR #40776, opened 2026-06-11
diagnostics: improve collect-wsl-logs for analysis (summary.json, README, profile info, WSL/guest state)

benhillis:
## Summary

A few improvements to `collect-wsl-logs.ps1` to make collected log archives easier to analyze (by a human or an agent), motivated by a real investigation where the captured logs lacked the context needed to diagnose the reported error.

## Changes

**1. Record the capture profile (`collection-info.txt`)**
Previously there was no record of which WPR profile a capture used, so it had to be inferred from the trace provider mix, which is easy to get wrong. For example a `-LogProfile networking` capture only contains VfpExt/HNS events and none of the WSL core trace providers, which can send an investigation down the wrong path. We now write a `collection-info.txt` recording the selected `LogProfile`, the mapped WPRP profile and file, the `Dump`/`RestartWslReproMode` switches, and the collection timestamp. This is capture provenance that cannot be derived from the rest of the archive.

**2. Readable `wsl.exe` output**
Sets `WSL_UTF8` and the console output encoding so any `wsl.exe` output captured by the script (e.g. the networking-profile guest network state) is saved to the log files readably instead of as UTF-16.

**3. Drop empty dump files**
When `MiniDumpWriteDump` fails it left a 0-byte `.dmp` behind, so archives could contain misleading empty dumps. The file is now removed on failure.

## Testing

- Script parses cleanly (`[Parser]::ParseFile`).
- Verified `collection-info.txt` renders as expected with real data, and that the logic is safe under `Set-StrictMode -Version Latest`.


diagnostics/collect-wsl-logs.ps1
```diff
@@ -369,10 +435,129 @@ if ($Dump)
...
+    }
+    networkingMode      = $networkingMode
+    wslConfigPresent    = [bool](Test-Path $wslconfig)
+    distributions       = $distributions
+    dumps               = $dumps
+}
+$summary | ConvertTo-Json -Depth 5 | Out-File -FilePath "$folder/summary.json" -Encoding utf8
+
+# Human/agent-readable index of the archive contents.
+@"
```
OneBlue: If we want an agent to read those, I think it would make sense for the prompt to be stored somewhere else. It feels a bit redundant to have it in every log archive.

OneBlue: As for the rest of values, it should be relatively easy for an agent to deduct them from the content of the archive (I'm not sure that it's worth duplicating the logic to dump the distros from the registry, and try to parse .wslconfig here)

diagnostics/collect-wsl-logs.ps1
```diff
@@ -302,6 +364,10 @@ finally
     wpr.exe -stop $folder/logs.etl 2>&1 >> $wprOutputLog
 }
 
+# Collect guest-side diagnostics after the repro (dmesg retains history, so this
+# captures kernel messages emitted during the repro).
+Collect-LinuxDiagnostics -Folder $folder
```
OneBlue: I don't think we should unconditionally call wsl.exe in the log collection script because this will hang forever if the service is deadlock / if the VM is in a bad state

diagnostics/collect-wsl-logs.ps1
benhillis: I agree with your second comment, but disagree with the first. I think a basic summary / readme is helpful.

diagnostics/collect-wsl-logs.ps1
benhillis: fair.

diagnostics/collect-wsl-logs.ps1
```diff
@@ -369,10 +388,96 @@ if ($Dump)
...
+    restartWslReproMode = [bool]$RestartWslReproMode
+    wslVersion          = (Get-Prop $appx "Version")
+    windows             = [ordered]@{
+        build          = "$(Get-Prop $winCV 'CurrentBuild').$(Get-Prop $winCV 'UBR')"
+        displayVersion = Get-Prop $winCV "DisplayVersion"
+        edition        = Get-Prop $winCV "EditionID"
+    }
+    wslConfigPresent    = [bool](Test-Path $wslconfig)
+    dumps               = $dumps
+}
```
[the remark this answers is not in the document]
benhillis: Updated the PR description (option b): summary.json is intentionally minimal, and wsl-info.txt / linux_diagnostics.log were dropped along with their wsl.exe calls per earlier review feedback. The description now matches what the script produces.

OneBlue (approved):

PR #40777, opened 2026-06-11
Rework the WSL2 guest memory reduction thread around explicit reclaim helpers

benhillis:
First in a series of incremental changes reworking the WSL2 guest memory reduction thread. Each change is submitted one PR at a time; the next will follow once this merges. This PR does not change the default reclaim mode.

### What this does
Replaces the ring-buffer idle detector and user-CPU-only sampling in the mini-init memory reduction thread with a clearer, helper-based design:

- Samples **aggregate non-idle CPU time** (user, system, irq, softirq, steal) so kernel-bound work keeps the VM out of the idle state, instead of looking at user time alone.
- `ReadProcFile` reads a full procfs snapshot into a caller buffer (close-on-exec, partial-read safe); `GetReclaimableCacheBytes` reads the cache counters through it. `GetFreeMemoryBytes` uses `sysinfo()` for the free-page total.
- **Gradual** reclaims cold page cache (cgroup `memory.reclaim`) above a fixed floor while CPU-idle, with a hysteresis trigger threshold so it does not churn near the floor. Each interval reclaims **at most a fixed step** (256 MB) so the cache bleeds down over several intervals rather than being stripped to the floor in a single pass. This keeps gradual gentle, a brief idle pause does not evict a whole working set, and keeps it meaningfully distinct from DropCache.
- **DropCache** stays gated on sustained CPU idle, drops once, and re-drops only after the reclaimable cache grows meaningfully.
- **Compaction** is gated on free-memory growth so it runs only when there are newly-freed pages worth coalescing.

`RequestCgroupReclaim` performs the `memory.reclaim` write best-effort: it treats the kernel's expected `EAGAIN` (some, but not all, pages evicted) as success **without logging**, and never throws, so a transient write error cannot tear down the long-lived reduction thread.

### Series (submitted sequentially)
1. **this PR** - rework around explicit reclaim helpers, with gradual bleeding in capped steps
2. drive gradual reclaim by memory pressure (PSI)
3. adaptive working-set floor via refaults
4. make gradual the default

### DIFF
--- a/src/linux/init/util.cpp
+++ b/src/linux/init/util.cpp
@@ -3925,8 +3925,14 @@ Return Value:
         return false;
     }
 
-    const long long Cache = GetReclaimableCacheBytes();
-    if (Cache >= 0 && (!State.ReclaimedThisIdlePeriod || Cache > State.CacheAtLastDrop + c_cacheGrowthRearmBytes))
+    bool DropCache = !State.ReclaimedThisIdlePeriod;
+    if (!DropCache)
+    {
+        const long long Cache = GetReclaimableCacheBytes();
+        DropCache = Cache >= 0 && Cache > State.CacheAtLastDrop + c_cacheGrowthRearmBytes;
+    }
+
+    if (DropCache)
     {
         // Best-effort; WriteToFile logs on failure.
         if (WriteToFile("/proc/sys/vm/drop_caches", "1\n") == 0)
--- a/src/linux/init/util.cpp
+++ b/src/linux/init/util.cpp
@@ -3571,6 +3571,11 @@ Return Value:
 
     if (Content.empty())
     {
+        if (LogErrors)
+        {
+            LOG_ERROR("read({}) returned empty content", Path);
+        }
+
         return std::nullopt;
     }
 
@@ -3657,10 +3662,10 @@ Routine Description:
     bytes) by parsing /proc/meminfo.
 
     It deliberately counts only memory that cache reclaim can actually return to
-    the host: file-backed page cache (Active(file) + Inactive(file)) plus
-    reclaimable slab (SReclaimable). Anonymous memory is excluded because neither
-    drop_caches nor cgroup reclaim of clean cache can free it, so it must not
-    drive the reclaim trigger.
+    the host: known file-backed page cache counters (Active(file) and
+    Inactive(file)) plus reclaimable slab (SReclaimable), when present. Anonymous
+    memory is excluded because neither drop_caches nor cgroup reclaim of clean
+    cache can free it, so it must not drive the reclaim trigger.
 
 Arguments:
 
--- a/src/linux/init/util.cpp
+++ b/src/linux/init/util.cpp
@@ -3535,7 +3535,7 @@ Return Value:
 --*/
 
 {
-    wil::unique_fd Fd{open(Path, O_RDONLY | O_CLOEXEC)};
+    wil::unique_fd Fd{TEMP_FAILURE_RETRY(open(Path, O_RDONLY | O_CLOEXEC))};
     if (!Fd)
     {
         if (LogErrors)
@@ -3802,7 +3802,7 @@ Return Value:
 {
     const std::string Request = std::to_string(Bytes);
 
-    wil::unique_fd Fd{open(RECLAIM_PATH, O_WRONLY | O_CLOEXEC)};
+    wil::unique_fd Fd{TEMP_FAILURE_RETRY(open(RECLAIM_PATH, O_WRONLY | O_CLOEXEC))};
     if (!Fd)
     {
         LOG_ERROR("open({}) failed {}", RECLAIM_PATH, errno);
--- a/src/linux/init/util.cpp
+++ b/src/linux/init/util.cpp
@@ -3505,71 +3505,153 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
 
 #define RECLAIM_PATH CGROUP_MOUNTPOINT "/memory.reclaim"
 
-static long long int GetUserCpuTime()
+static bool ReadProcFile(const char* Path, char* Buffer, size_t Size, bool LogErrors = true)
 
 /*++
 
 Routine Description:
 
-    This routine parses /proc/stat to query a summary of all user CPU time.
+    This routine reads the full contents of a procfs file into a caller-supplied
+    NUL-terminated buffer. Because procfs content is generated on read and a
+    single read() may return only a partial snapshot, this loops until EOF (or
+    the buffer fills), which keeps later-appearing fields (for example the
+    workingset counters deep in /proc/vmstat) from being truncated away.
 
 Arguments:
 
-    None.
+    Path - Supplies the procfs path to read.
+
+    Buffer - Supplies the buffer to fill; always NUL-terminated on success.
+
+    Size - Supplies the size of Buffer in bytes (must be at least 1).
+
+    LogErrors - Supplies whether open/read failures are logged. Callers for
+        which absence is normal (for example PSI) pass false.
 
 Return Value:
 
-    The current user CPU counter for all cores.
+    true on success (Buffer holds the content), false on failure.
 
 --*/
 
 {
-    wil::unique_fd Fd{open("/proc/stat", O_RDONLY)};
+    if (Size == 0)
+    {
+        return false;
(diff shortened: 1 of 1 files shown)


src/linux/init/util.cpp
```diff
@@ -3650,89 +4006,42 @@ try
...
-                    std::this_thread::sleep_for(std::chrono::seconds(1));
-                    Stop = GetUserCpuTime();
-                    THROW_LAST_ERROR_IF(Stop == -1);
-                    if ((Stop - Start) < IdleThreshold)
-                    {
-                        THROW_LAST_ERROR_IF(WriteToFile("/proc/sys/vm/compact_memory", "1\n") < 0);
-                    }
+                    State.PreviousBusy = Busy;
+                    State.PreviousIdle = Idle;
+                    State.HavePreviousSample = true;
```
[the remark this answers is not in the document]
benhillis: Fixed in ad76aeab: FreeAtLastCompaction is now seeded with the free-memory reading at the first CPU sample, so the first compaction tick measures growth from startup instead of from zero.

src/linux/init/util.cpp
```diff
@@ -3505,98 +3505,466 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
...
 
-try
 {
     struct sysinfo Info = {};
-    THROW_LAST_ERROR_IF(sysinfo(&Info) < 0);
-    return (Info.totalram - Info.freeram) * Info.mem_unit;
+    if (sysinfo(&Info) < 0)
+    {
+        LOG_ERROR("sysinfo failed {}", errno);
+        return -1;
```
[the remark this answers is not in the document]
benhillis: Updated the PR description to state that GetFreeMemoryBytes uses sysinfo() for the free-page total (only GetReclaimableCacheBytes goes through ReadProcFile). sysinfo() is the right fit here, so no code change.

src/linux/init/util.cpp
```diff
@@ -3505,98 +3505,466 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
...
+
+// DropCache: drop only after this many consecutive idle intervals, then re-drop once the reclaimable
+// cache grows by at least the re-arm threshold.
+constexpr int c_dropCacheIdleIntervals = 30; // 5 minutes
+constexpr long long c_cacheGrowthRearmBytes = 256ll * 1024 * 1024;
+
+// Gradual: reclaimable cache below this floor is always retained; only the excess above it (beyond a
+// hysteresis margin) is reclaimed.
+constexpr long long c_floorBaseBytes = 128ll * 1024 * 1024;
+constexpr long long c_gradualHysteresisBytes = 128ll * 1024 * 1024;
```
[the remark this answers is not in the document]
benhillis: Clarified in ad76aeab: the comment now states the hysteresis value is a trigger threshold (not a retained margin) that keeps reclaim from churning near the floor.

src/linux/init/util.cpp
```diff
@@ -3505,98 +3506,476 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
...
+        if (Result == 0)
+        {
+            break;
+        }
+
+        Total += static_cast<size_t>(Result);
+    }
+
+    Buffer[Total] = '\0';
+    return Total > 0;
```
OneBlue: This will return success if the buffer is too small to read the full file. We might want to either throw an exception, or rewrite that method so it returns something like an std::string an re-allocates as needed

src/linux/init/util.cpp
```diff
@@ -3650,89 +4027,54 @@ try
...
+                    //
+
+                    const long long Free = GetFreeMemoryBytes();
+                    if (Free >= 0)
                     {
-                        THROW_LAST_ERROR_IF(WriteToFile("/proc/sys/vm/compact_memory", "1\n") < 0);
+                        State.FreeAtLastCompaction = Free;
                     }
+
+                    continue;
```
ptrivedi, line 4091: nit: if GetFreeMemoryBytes() returns -1 here then State.FreeAtLastCompaction will stay 0, so the first real RunCompactionTick will see  Free > 0 + c_compactFreeGrowthBytes and trigger one spurious compaction. It's harmless, but you could avoid it by skipping the  continue/retrying the seed on the next tick until a valid return.

ptrivedi:
This reworks a bunch of parsing and policy logic -- thread loop is hard to test, but can the pure parsing helpers like ReadAggregateCpuTimes() and GetReclaimableCacheBytes() be candidates that can be reached via test/linux/unit_tests/?

ptrivedi:
Curiosity questions around testing:
* Were you able confirm the fallback when memory.reclaim is not present ( access(RECLAIM_PATH, W_OK) < 0 ) actually switches to DropCache?
• Any before/after numbers on guest memory returned to the host, or observed behavior under sustained load vs. idle (to confirm kernel-bound work correctly keeps the VM out of the idle state)?
• Was EAGAIN from memory.reclaim observed in practice at all?

src/linux/init/util.cpp
```diff
@@ -3505,98 +3506,517 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
...
+    // Format: "cpu  user nice system idle iowait irq softirq steal guest guest_nice".
+    // The guest fields are already accounted for in user/nice and are ignored here.
     //
 
-    Buffer[Result] = '\0';
-    char* Sp1;
-    char* Info = strtok_r(Buffer, " \n", &Sp1);
-    if (Info == nullptr)
+    unsigned long long Fields[8] = {};
+    const int Parsed = sscanf(
```
ptrivedi, line 3624: are you using sscanf for the slightly better perf? otherwise, using std::stringstream is much more robust, type-safe and flexible. Over the years, linux has added fields to the cpu line, so using a rigid sscanf format is less favorable

If it is really raw perf that's desired, one could read directly from file stream using ifstream... but that has its issues if file structure changes

src/linux/init/util.cpp
```diff
@@ -3505,98 +3506,517 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
...
+    long long TotalKb = 0;
+    bool FoundCounter = false;
+    char* Save = nullptr;
+    for (char* Line = strtok_r(Content->data(), "\n", &Save); Line != nullptr; Line = strtok_r(nullptr, "\n", &Save))
     {
-        LOG_ERROR("/proc/stat first line missing cpu counter");
+        const char* Value = nullptr;
+        if (strncmp(Line, "Active(file):", 13) == 0)
+        {
+            Value = Line + 13;
```
ptrivedi, line 3696: In /proc/meminfo, the value fields have spaces in the front. hopefully strtoll is smart enough to handle that.

src/linux/init/util.cpp
```diff
@@ -3505,98 +3506,517 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
...
+    }
+
+    if (!FoundCounter)
+    {
+        LOG_ERROR("failed to find reclaimable cache counters in /proc/meminfo");
         return -1;
     }
 
-    return strtoll(Info, nullptr, 10);
+    return TotalKb * 1024;
```
ptrivedi, line 3720: literal suffix 'LL' is safer to use for 1024

src/linux/init/util.cpp
```diff
@@ -3505,98 +3506,517 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
...
+        {
+            Value = Line + 13;
+        }
+
+        if (Value != nullptr)
+        {
+            FoundCounter = true;
+            TotalKb += strtoll(Value, nullptr, 10);
+        }
+    }
```
ptrivedi, line 3712: Robustness comment: If /proc/meminfo is somehow truncated and only contains SReclaimable:, FoundCounter will be true, and the function will return successfully. It is much safer to ensure you actually found Active(file) and Inactive(file) instead of relying on a single loose boolean.

benhillis:
Closing this in favor of #41096, which supersedes this approach to the WSL2 guest memory reclaim rework. Thanks @ptrivedi for the thorough review here, especially the robustness callouts on the `/proc/meminfo` parsing -- those are carried forward. Please take a look at #41096 when you get a chance.

PR #40778, opened 2026-06-11
Drive gradual reclaim by memory pressure instead of CPU idle

benhillis:
**Stacked PR 2 of 4** — builds on #40777. Diff shown is cumulative (helpers rework + this change).

### What this does
A CPU-bound workload can sit on gigabytes of cold page cache that a CPU-idle check would never reclaim. This drives **Gradual** reclaim by *memory pressure* instead of CPU idleness:

- Reads the PSI `some avg10` memory pressure from `/proc/pressure/memory`.
- Reclaims cold cache toward the fixed floor whenever pressure is low, **even while the VM is busy**, backing off once the workload starts stalling on memory.
- A busy interval reclaims at most a bounded step (`c_gradualStepBusyBytes`) so a large backlog is drained gently; an idle interval drains the full excess.
- When PSI is unavailable (kernel built without `CONFIG_PSI`), gradual reclaim falls back to gating on CPU idle.

### Stack
1. #40777 — rework around explicit reclaim helpers
2. **this PR** — drive gradual reclaim by memory pressure (PSI)
3. adaptive working-set floor via refaults
4. make gradual the default

### DIFF
--- a/src/linux/init/util.cpp
+++ b/src/linux/init/util.cpp
@@ -24,6 +24,7 @@ Abstract:
 #include <ctype.h>
 #include <optional>
 #include <fstream>
+#include <algorithm>
 #include <iostream>
 #include <sstream>
 #include <regex>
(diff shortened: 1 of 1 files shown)


benhillis:
Closing for now. These changes are being submitted one PR at a time; this will be reopened against master once the preceding PR in the series has merged, so it shows only its own incremental change. The branch remains pushed.

PR #40779, opened 2026-06-11
Adapt the gradual reclaim floor to the working set using refaults

benhillis:
**Stacked PR 3 of 4** — builds on #40777 and the PSI change. Diff shown is cumulative.

### What this does
Reclaiming toward a single fixed floor either gives back too little on large VMs or evicts pages a larger working set immediately faults back in. This makes the floor **adaptive and self-regulating**:

- Tracks file refaults (`/proc/vmstat workingset_refault*`) as a signal that reclaim is cutting into the working set. When refaults spike (or PSI pressure rises into the backoff band), the floor is raised to protect what the workload is actually using and reclaim stops.
- After sustained calm, the floor decays back toward the base so a shrunken working set is eventually re-probed downward.
- Scales the floor's upper bound to a fraction of guest RAM so large working sets on large VMs can be fully protected, falling back to a fixed cap when total RAM is unavailable.
- The PSI-unavailable path keeps the same refault brake and floor decay so behavior degrades gracefully on kernels without `CONFIG_PSI`.

### Stack
1. #40777 — rework around explicit reclaim helpers
2. drive gradual reclaim by memory pressure (PSI)
3. **this PR** — adaptive working-set floor via refaults
4. make gradual the default

### DIFF
--- a/src/linux/init/util.cpp
+++ b/src/linux/init/util.cpp
@@ -24,6 +24,7 @@ Abstract:
 #include <ctype.h>
 #include <optional>
 #include <fstream>
+#include <algorithm>
 #include <iostream>
 #include <sstream>
 #include <regex>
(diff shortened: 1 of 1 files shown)


benhillis:
Closing for now. These changes are being submitted one PR at a time; this will be reopened against master once the preceding PR in the series has merged, so it shows only its own incremental change. The branch remains pushed.

PR #40770, opened 2026-06-11
wslc: introduce IWSLCVirtualMachineFactory for session VM creation

benhillis:
## Summary of the Pull Request

Introduces `IWSLCVirtualMachineFactory` as the abstraction the per-user `wslcsession.exe` uses to create its VM, replacing the previous model where the SYSTEM service created the VM up front and handed the pre-created `IWSLCVirtualMachine` to the session.

This is **behavior-preserving**: the session still creates exactly one VM eagerly during `Initialize`. The change simply funnels VM creation through a factory so the session owns *when* the VM is created. This is groundwork for a follow-up change that lets a session tear down its VM when idle and recreate it on demand.

## PR Checklist

- [ ] **Closes:** N/A (refactor / groundwork)
- [x] **Communication:** Foundational refactor split out of a larger idle-VM feature branch
- [x] **Tests:** No new tests; behavior-preserving. Existing WSLC container E2E suite passes.
- [x] **Localization:** No new end-user-facing strings
- [x] **Dev docs:** N/A

## Detailed Description of the Pull Request / Additional comments

- `wslc.idl`: add `IWSLCVirtualMachineFactory` (single method `CreateVirtualMachine`); change `IWSLCSession::Initialize` and `IWSLCSessionFactory::CreateSession` to take `IWSLCVirtualMachineFactory*` instead of `IWSLCVirtualMachine*`.
- `HcsVirtualMachine.{h,cpp}`: implement `WSLCVirtualMachineFactory`, a `RuntimeClass<ClassicCom, IWSLCVirtualMachineFactory, IFastRundown>` that owns a deep copy of the VM settings (including a duplicated dmesg sink) and constructs a fresh `HcsVirtualMachine` per call.
- `WSLCSessionManager.cpp`: the SYSTEM service now creates the factory (`Make<WSLCVirtualMachineFactory>(Settings)`) and passes it through `CreateSession`.
- `WSLCSession.cpp` / `WSLCSessionFactory.cpp`: the session creates the VM via `VmFactory->CreateVirtualMachine()` during `Initialize`, then proceeds exactly as before.
- `msipackage/package.wix.in`: register the new interface IID for COM marshaling (proxy/stub), matching every other cross-process WSLC interface. Omitting this fails session `Initialize` with `REGDB_E_INTERFACENOTREG` (0x80040155).

## Validation Steps Performed

- Full `cmake --build . -- -m` (x64 Debug) clean.
[...]

PR #40366, opened 2026-04-30
Add WSLC (WSL Containers) feature

benhillis:
WSLC is a container runtime built on the Windows Subsystem for Linux, enabling Windows applications to create and manage Linux containers through a native Windows API surface.

## Key components

- **wslc.exe**: CLI for managing containers, images, volumes, and networks (build, run, stop, inspect, push/pull from registries)
- **wslcsession.exe**: Per-user Windows service hosting container lifecycle, storage management, and networking
- **WSLC SDK**: C++ and C# client libraries with NuGet packaging for programmatic container management
- **Container networking**: port forwarding, DNS tunneling, virtio networking, and HCN integration
- **Storage**: VHD-backed volumes, virtiofs file sharing, overlayfs layers
- **GPU passthrough** and device host proxy support

## Stats

- 384 files changed, 73,620 insertions, 2,751 deletions
- Squash merge of `feature/wsl-for-apps` (665 commits)

PR #40781, opened 2026-06-11
wslc: idle-terminate per-user session VMs when inactive

### LINKED ISSUE #40961: vmmemwslc-cli remains running after all containers have exited
### Windows Version

Microsoft Windows [Version 10.0.26200.8655]

### WSL Version

2.9.3.0

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.18.35.2

### Distro Version

Debian

### Other Software

_No response_

### Repro Steps

1. Run a temporary Alpine container:

    ```powershell
    wslc run --rm -it alpine:latest
    ```

2. Exit the container (for example, by running exit or pressing Ctrl+D).

3. Verify that there are no running containers:

    ```powershell
    wslc ps
    ```
    
    Output:
    ```powershell
    CONTAINER ID   NAME   IMAGE   CREATED   STATUS   PORTS
    ```

4. Verify that there are no existing containers:

    ```powershell
    wslc container list
    ```
    
    Output:
    ```powershell
    CONTAINER ID   NAME   IMAGE   CREATED   STATUS   PORTS
    ```

5. Open Task Manager (or run `tasklist`) and observe that `vmmemwslc-cli-<username>` is still running.

### Expected Behavior

Once the container exits and there are no running or existing containers, the `wslc` VM should become idle and shut down automatically after the configured idle timeout. The `vmmemwslc-cli-<username>` process should terminate and release its memory.

### Actual Behavior

Although there are no running or existing containers, `vmmemwslc-cli-<username>` continues running indefinitely and keeps consuming memory (around 500 MB in my case). It does not exit even after waiting longer than the configured VM idle timeout.

### Diagnostic Logs

_No response_

benhillis:
Fixes #40961

## Summary

Idle-terminates a per-user WSLC session's backing VM when it has been inactive, freeing memory while the session object (and its persistent storage) lives on. The VM is transparently recreated on the next operation.

Builds on #40770 (IWSLCVirtualMachineFactory).

## Behavior

- Only sessions with persistent storage (StoragePath set) idle-terminate.
- An idle worker thread tears the VM down after a grace period (currently 30s) once there is no in-flight activity and no active container lock.
- In-flight work holds an activity reference so the VM cannot be torn down mid-operation:
  - VmLease wraps CLI/container operations.
  - BeginContainerOperation hands clients an activity token (IFastRundown so a client crash reclaims it promptly).
  - Long-lived root-namespace processes (e.g. plugin hosts) created via CreateRootNamespaceProcess hold a keep-alive token for their lifetime.
- Activity bookkeeping (count + wake event) lives in a shared IdleState held via shared_ptr, decoupled from the session's lifetime, so a held token suppresses idle teardown **without** extending the session object's lifetime (preserving the explicit-reset-invalidates-held-processes invariant from #40366).

## VM plugin hooks

Idle-terminate breaks the old "session == running VM" assumption (the VM is lazily created, idle-terminated, and transparently recreated while the session lives on), so plugins gain VM-level hooks that fire on every VM (re)start/teardown, alongside the existing per-session `OnSessionCreated`/`OnSessionStopping` hooks:

- `WSLPluginAPI_OnWslcVmStarted` / `WSLPluginAPI_OnWslcVmStopping` (WslPluginApi.h) with `WslcVmStarted`/`WslcVmStopping` fields in `WSLPluginHooksV1`.
- `IWSLCPluginNotifier::OnVmStarted()` / `OnVmStopping()` (new methods, no new IID — service and per-user session ship together).
- `OnWslcVmStarted` fires **after** releasing `m_lock` in `EnsureVmRunning` (the sole caller holds an activity reference across the gap, so idle teardown can't race the VM down; firing under the lock could deadlock on a reentrant `WSLCCreateProcess`).
- `OnWslcVmStopping` fires **under** the lock in `TearDownVmLockHeld`, with fire-once pairing via `std::atomic<bool> m_vmStartNotified` so a failed bring-up emits no spurious stopping.

## Testing

- New `WSLCE2EVmIdleTests` E2E suite (5 tests) including `WSLCE2E_VmIdle_RootProcessKeepsVmAlive`.
- New `PluginTests::WslcVmRestart` asserting the hook sequence: Session created -> VM started -> VM stopping -> VM started -> VM stopping -> Session stopping.
- `WSLCTests::CreateRootNamespaceProcess` still passes.
- Full x64 Debug build clean.

## Notes / follow-ups (deferred)

- Grace period is a hardcoded constexpr; making it injectable would enable deterministic race tests.
- No crash-path (client dies holding token) automated coverage yet.

> [!NOTE]
> Draft for early review.


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -317,67 +323,495 @@ try
...
+}
+
+void WSLCSession::TearDownVmLockHeld()
+{
+    std::lock_guard containersLock(m_containersLock);
+    std::lock_guard networksLock(m_networksLock);
+
+    m_containers.clear();
+    m_volumes.reset();
+    m_networks.clear();
```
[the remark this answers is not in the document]
benhillis: Addressed in ea2254cf. The idle worker now keeps the VM alive while a client still holds a container or process proxy, so teardown no longer disconnects outstanding proxies:

- Containers: `HasActiveContainerLockHeld` also returns true while a container wrapper is externally referenced (COM refcount above the single internal `m_comWrapper` reference), and `WSLCContainer::Release()` wakes the idle worker when the last client proxy is released so the VM is still reclaimed promptly.
- Exec/root processes: the returned `WSLCProcess` wrapper carries a keep-alive activity token for its client-held lifetime.

Covered by the new `WSLCE2E_VmIdle_HeldContainerProxyKeepsVmAlive` test.

src/windows/service/inc/wslc.idl
```diff
@@ -788,6 +794,19 @@ interface IWSLCSession : IUnknown
...
+    // Returns a one-off event that is signaled when the session terminates, whether due to an
+    // explicit Terminate() call or an unexpected VM exit. The returned handle is owned by the
+    // caller and remains valid (and observes the signaled state) even after the session is released.
+    HRESULT GetTerminationEvent([out, system_handle(sh_event)] HANDLE* Event);
+
+    // Returns the cached termination reason and details. The values are only meaningful after the
+    // termination event has been signaled; before that the reason is
+    // WSLCVirtualMachineTerminationReasonUnknown and Details is an empty string.
+    HRESULT GetTerminationReason([out] WSLCVirtualMachineTerminationReason* Reason, [out] LPWSTR* Details);
+
```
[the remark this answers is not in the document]
benhillis: Fixed in b8700443. Updated the IDL comment to match the implementation: GetTerminationReason returns HRESULT_FROM_WIN32(ERROR_INVALID_STATE) until the termination event is signaled, rather than returning Unknown/empty.

benhillis (commented):
Reviewed VM-related comments - all have been addressed:

**Comment on WSLCSession.h:84**: Already correct - lines 77-78 say "IWSLCVirtualMachineFactory" and "lazily on first use"

**Comment on WSLCSession.cpp:571**: Fixed by AddRef/Release activity tracking - idle worker checks ActivityCount (line 779), and container proxies increment it on AddRef 1→2 transition. VM will not idle-terminate while clients hold container proxies. See lines 672-674 comment.

**Comment on WSLCSession.cpp:380**: Already has exception handling - IdleWorker() is wrapped in CATCH_LOG() at lines 375-379

**Comment on Session.cpp:65**: Already correct - wil::unique_threadpool_wait (line 56) calls WaitForThreadpoolWaitCallbacks in destructor automatically

src/windows/wslcsession/WSLCExecutionContext.h
```diff
@@ -27,15 +27,27 @@ class WSLCExecutionContext : public wsl::windows::common::COMServiceExecutionCon
 protected:
     bool CollectUserWarning(const std::wstring& warning) override
     {
-        if (m_warningCallback != nullptr)
+        IWarningCallback* callback = m_warningCallback;
+
+        // When the operation carries no explicit callback, fall back to the callback supplied when
+        // the session was created/entered. This routes warnings emitted outside a callback-bearing
```
OneBlue: I don't think we should do this. This can lead to multiple threads calling the warning callback at the same time and warnings being printed that aren't related to the command that the user is currently running

OneBlue: If there's no warning callback from the current operation, I think just logging instead (and writing to the event log) is OK

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -342,80 +375,488 @@ try
...
+
+    if (UserSid != nullptr)
+    {
+        const auto length = GetLengthSid(UserSid);
+        const auto* bytes = reinterpret_cast<const BYTE*>(UserSid);
+        m_userSid.assign(bytes, bytes + length);
+    }
+    else
+    {
+        m_userSid.clear();
```
OneBlue: I don't think UserSid should ever be null (this is also checked in other places)

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -42,8 +42,41 @@ constexpr auto c_storageVhdFilename = L"storage.vhdx";
 constexpr DWORD c_processTerminateTimeoutMs = 30 * 1000;
 constexpr DWORD c_processKillTimeoutMs = 10 * 1000;
 
+// Grace period to keep an otherwise-idle VM running before tearing it down. This avoids
+// thrashing the VM (repeated teardown/recreate) when containers are created and destroyed,
+// or operations issued, in quick succession. The clock restarts whenever the VM is observed
+// to be non-idle, so a full grace period of continuous idleness is required before teardown.
+constexpr auto c_vmIdleGracePeriod = std::chrono::seconds(30);
```
OneBlue, line 51: This should probably be a yaml setting

src/windows/wslcsession/WSLCProcessControl.cpp
```diff
@@ -101,7 +101,15 @@ void DockerContainerProcessControl::OnContainerReleased() noexcept
     // Signal the exit event to prevent callers from being blocked on it.
     if (!m_exitEvent.is_signaled())
     {
-        m_exitedCode = 128 + WSLCSignalSIGKILL;
+        // If the container already produced a real exit code (recorded by SetExitCode but not yet
+        // signaled — e.g. an --rm container whose init-exit signal is deferred to the Destroy
+        // event), preserve it. Only synthesize SIGKILL when the container is released without ever
+        // having produced an exit code (an abrupt teardown of a still-running container).
+        if (!m_exitedCode.has_value())
```
OneBlue, line 108: Is it possible for us to reach this block while `m_exitedCode` is already set ? We check `m_exitEvent` under the lock before entering

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -342,80 +375,488 @@ try
...
+        CATCH_LOG();
+    }
+
+    m_dockerdProcess.reset();
+    m_containerdProcess.reset();
+    m_virtualMachine.reset();
+
+    // Destroy the relay unless we're on its own thread (~IORelay joins the thread, which would
+    // deadlock). On unexpected-VM-exit path (runs on relay thread), leave it for ~WSLCSession.
+    if (!m_ioRelay || !m_ioRelay->IsRelayThread())
```
OneBlue: Is it ever possible to reach this from the IO relay thread ? Is there a code path where the IO relay thread acquires a VM lease ?

OneBlue (commented):
From a first read, I think this is looking good. One thing that might help make this a bit easier would be to fully separate the "VM state, related fields and timing logic" out of `WSLCSession`, with something like: 

```
class WSLCSessionRuntime
{
    WSLCVirtualMachine m_vm;
    IORelay m_ioRelay;
    [..]
}
```

That class can handle the ref counting, timeout and such, and then whenever the session needs to access the VM, it can do something like: 

```
struct LockedRuntime
{
    WSLCVirtualMachine& vm;
    IORelay& relay;
    [...]
};

```
Which then allows do something like:

auto runtime = AcquireRuntime(...);

runtime->Vm.Operation(...)
```

AcquireRuntime starts the VM as needed and handles timeouts

File: src/windows/wslcsession/WSLCSession.cpp (hunk 7 above)
Comment (benhillis) [reply src/windows/wslcsession/WSLCSession.cpp]: This prevents ~IORelay  from joining its own thread during teardown.

File: src/windows/wslcsession/WSLCProcessControl.cpp (hunk 6 above)
Comment (benhillis) [reply src/windows/wslcsession/WSLCProcessControl.cpp:108]: m_exitedCode  can already be set on entry (the deferred  --rm  case sets the code via     SetExitCode  without signaling)

File: src/windows/wslcsession/WSLCContainer.cpp [hunk 8]
```diff
@@ -528,48 +528,51 @@ WSLCPortMapping ContainerPortMapping::Serialize() const
...
     m_labels(std::move(labels)),
     m_comWrapper(wil::MakeOrThrow<WSLCContainer>(wslcSession, std::move(onDeleted))),
-    m_dockerClient(DockerClient),
-    m_eventTracker(EventTracker),
-    m_ioRelay(Relay),
-    m_containerEvents(EventTracker.RegisterContainerStateUpdates(
+    m_dockerClient(runtime.Docker()),
+    m_eventTracker(runtime.Events()),
+    m_ioRelay(*runtime.Relay()),
+    m_containerEvents(runtime.Events().RegisterContainerStateUpdates(
```
OneBlue, line 561: `m_containerEvents` keeps a reference to the `DockerEventTracker` it was created with, but if that object goes away, it will operate on a dangling pointer.

The easiest way I can think of to solve this would be to keep the same event tracker object between different VM's, and have logic to "restart" it when a new VM starts (other wise, all containers & processes would need to re-register with the new event tracker)

src/windows/wslcsession/WSLCSessionRuntime.cpp
```diff
@@ -0,0 +1,739 @@
...
+}
+
+void WSLCSessionRuntime::TearDownVmLockHeld(bool CaptureTerminationReason)
+{
+    // Notify plugins that the VM is about to stop, paired with a prior OnVmStarted. Skipped for a
+    // VM that never finished starting (m_vmStartNotified false, e.g. bring-up failure). Fired under
+    // the lock, so the handler must not call back into the session.
+    if (m_vmStartNotified.exchange(false) && m_hooks.OnVmStopping)
+    {
+        m_hooks.OnVmStopping();
```
OneBlue: I think there's a potential deadlock here. If a plugin receives a "VM stopping" notification and runs an operation that can try to acquire a VM lease, we could get stuck on `m_lock` in `VmLease`'s constructor

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -343,81 +377,176 @@ try
...
-    // Launch dockerd with external containerd socket
-    StartDockerd();
+    hooks.TearDownSessionState = [this]() {
+        std::lock_guard containersLock(m_containersLock);
+        std::lock_guard networksLock(m_networksLock);
 
-    // Wait for dockerd to be ready before starting the event tracker.
-    THROW_WIN32_IF_MSG(
-        ERROR_TIMEOUT, !m_dockerdReadyEvent.wait(Settings->BootTimeoutMs), "Timed out waiting for dockerd to start");
+        m_containers.clear();
```
OneBlue: I don't think we can do this unfortunately, because this will break any COM references that callers might have to containers

src/windows/wslcsession/WSLCSessionRuntime.cpp
```diff
@@ -0,0 +1,739 @@
...
+    {
+        return;
+    }
+
+    // Acquire an exclusive lock to ensure that no operation is running.
+    WI_VERIFY(sessionLock);
+
+    // Tear down the VM (if running) and all VM-scoped state, capturing the termination reason.
+    // This mirrors the soft teardown used for idle shutdown, but here it is permanent.
+    TearDownVmLockHeld(/* CaptureTerminationReason */ true);
```
OneBlue, line 832: The locking is confusing me a bit here, because we're calling `TearDownVmLockHeld`, but we don't actually hold `m_lock`, only the session lock.

I think the locking still works because `StopVmLockHeld` uses an std::atomic to transition the VM to stopped, but I think we should make the locking a bit more clear here

src/windows/wslcsession/WSLCSessionRuntime.h
```diff
@@ -0,0 +1,225 @@
...
+        // the handler must not call back into the session. Paired with a prior OnVmStarted.
+        std::function<void()> OnVmStopping;
+    };
+
+    struct SessionContext
+    {
+        ULONG Id{};
+        std::wstring DisplayName;
+        const std::atomic<bool>* Terminating{};
+        wil::unique_event* SessionTerminatingEvent{};
```
OneBlue: nit: We could use `wil::shared_event` to avoid having to deal with pointers here

test/windows/testplugin/Plugin.cpp
```diff
@@ -539,6 +539,44 @@ HRESULT OnWslcImageDeleted(const WSLCSessionInformation* Session, LPCSTR ImageId
...
+CATCH_RETURN();
+
+HRESULT OnWslcVmStopping(const WSLCSessionInformation* Session)
+{
+    if (g_testType != PluginTestType::WslcVmRestart)
+    {
+        return S_OK;
+    }
+
+    g_logfile << "WSLC VM stopping, session=" << Session->SessionId << std::endl;
```
OneBlue, line 591: I recommend launching a process inside the VM & removing a mount here, to validate that we handle these calls correctly

benhillis:
Closing this in favor of #41077, which supersedes this approach with the same idle-terminate behavior plus a WSLCSessionRuntime encapsulation cleanup and two additional concurrency fixes found during review (notify-lock race between start/stop hooks, and a StopRequested disposition being clobbered before a spontaneous-exit check). Thanks @OneBlue for reviewing here -- please take a look at #41077 when you get a chance.

PR #40782, opened 2026-06-11
Fix disk-attach restore TOCTOU by impersonating the mounting user on VHD restore

benhillis:
## Summary

Closes a time-of-check/time-of-use (TOCTOU) gap in disk attach where the **restore** path re-resolved a user-controllable VHD path as **SYSTEM** after the VM was recreated.

## Background

For a live `wsl --mount --vhd`, a junction/symlink swap is already harmless: the VM access grant runs while **impersonating the user** (`GrantVmAccess`), and the SYSTEM-side `AddVhd` only succeeds on a file the VM was granted access to. A swap therefore yields `ACCESS_DENIED`, not disclosure.

The real gap was **disk restore**: when the utility VM is torn down on idle and later recreated, `LxssUserSession::_LoadDiskMount` re-attached persisted VHDs with **no user token** (as SYSTEM), re-resolving the user-supplied path and reopening the swap window.

## Fix

Disk-mount state is stored under the user's SID in a **volatile (per-boot)** registry key, so the disk being restored was mounted earlier in the same boot by the **same user**, and that user's token is already available at VM-create time. So restore simply passes the user token for VHDs and lets the existing impersonated grant close the window:

- **VHD restore** now runs under the mounting user's identity — a privileged operation can only ever touch a file that user can already reach.
- **Pass-through** (raw block device) restore stays SYSTEM: it is elevation-gated and `\\.\PhysicalDriveN` paths have no reparse-point surface, so there is nothing to swap.

This replaces the earlier handle-pinning / reparse-point-rejection approach (per review feedback), which also regressed legitimate symlinked VHDs.

## Tests

- `MountVhdThroughSymlinkSucceeds` — a symlinked VHD mounts (regression guard).
- `MountVhdThroughSymlinkSurvivesVmTimeout` — a symlinked VHD re-attaches after a VM idle-timeout restore.

### DIFF
--- a/test/windows/MountTests.cpp
+++ b/test/windows/MountTests.cpp
@@ -460,6 +460,72 @@ class MountTests
         VERIFY_ARE_EQUAL(LxsstuLaunchWsl(L"--unmount " + absolutePath.wstring()), (DWORD)0);
     }
 
+    // A VHD whose path is a symbolic link is a legitimate, supported scenario: the link is
+    // followed and the real VHD is attached. Access is granted while impersonating the user,
+    // so the user can only ever attach a file they can already reach; there is no need to
+    // reject reparse points in the path.
+    WSL2_TEST_METHOD(MountVhdThroughSymlinkSucceeds)
+    {
+        SKIP_UNSUPPORTED_ARM64_MOUNT_TEST();
+
+        const auto symlink = std::filesystem::absolute(L"TestVhdSymlink.vhd");
+        DeleteFileW(symlink.c_str());
+
+        const auto absoluteTarget = std::filesystem::absolute(TEST_MOUNT_VHD);
+
+        // Create a file symbolic link pointing at the real VHD.
+        if (!CreateSymbolicLinkW(symlink.c_str(), absoluteTarget.c_str(), SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE))
+        {
+            // Creating symlinks requires Developer Mode or SeCreateSymbolicLinkPrivilege; skip if unavailable.
+            LogSkipped("Unable to create a symbolic link (error %d); skipping", GetLastError());
+            return;
+        }
+
+        auto cleanup = wil::scope_exit([&]() { DeleteFileW(symlink.c_str()); });
+
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(L"--mount " + symlink.wstring() + L" --vhd --bare"), (DWORD)0);
+
+        const auto disk = GetBlockDeviceInWsl();
+        VERIFY_IS_TRUE(IsBlockDevicePresent(disk));
+
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(L"--unmount " + symlink.wstring()), (DWORD)0);
+    }
+
+    // A symlinked VHD must still be restored after the VM is torn down on idle. Restore runs
+    // under the mounting user's identity (the disk-mount state is stored per-SID for the
+    // current boot), so the same access check applies and the symlinked VHD re-attaches.
+    WSL2_TEST_METHOD(MountVhdThroughSymlinkSurvivesVmTimeout)
+    {
+        SKIP_UNSUPPORTED_ARM64_MOUNT_TEST();
+
+        const auto symlink = std::filesystem::absolute(L"TestVhdSymlinkRestore.vhd");
+        DeleteFileW(symlink.c_str());
+
+        const auto absoluteTarget = std::filesystem::absolute(TEST_MOUNT_VHD);
+
+        if (!CreateSymbolicLinkW(symlink.c_str(), absoluteTarget.c_str(), SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE))
+        {
+            LogSkipped("Unable to create a symbolic link (error %d); skipping", GetLastError());
+            return;
+        }
+
+        auto cleanup = wil::scope_exit([&]() { DeleteFileW(symlink.c_str()); });
+
+        WslKeepAlive keepAlive;
+
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(L"--mount " + symlink.wstring() + L" --vhd --bare"), (DWORD)0);
+
+        const auto disk = GetBlockDeviceInWsl();
+        VERIFY_IS_TRUE(IsBlockDevicePresent(disk));
+
+        WaitForVmTimeout(keepAlive);
+
+        // Recreating the VM restores the persisted disk mount; the symlinked VHD must re-attach.
+        VERIFY_IS_TRUE(IsBlockDevicePresent(disk));
+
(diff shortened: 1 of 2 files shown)


benhillis:
Thanks for the review. Addressed in 627b2b9e:

- **Persisted path (SaveDiskState/_LoadDiskMount)**: Good catch on the intent. _LoadDiskMount does re-run ResolveDiskPath (reparse-reject + canonicalize) at restore, but to match the stated goal and add defense in depth, SaveDiskState now persists DiskState::ResolvedPath (canonical, reparse-free) instead of the raw user string. Registry entries are keyed by LUN, not by path, so no string-equality matching needed updating.
- **Test comment**: Reworded — the unmount passes the junction path and the service canonicalizes internally to match the attached-disk entry.
- **pinHandle C4101 / `/WX`**: This is a false positive. `wil::unique_hfile` has a non-trivial destructor, so MSVC does not emit C4101 (the variable is "used" for its RAII lifetime, exactly like the existing `auto cleanup = wil::scope_exit(...)` and `auto runAsUser = wil::impersonate_token(...)` guards in this file). A full `/WX` build of the solution compiles cleanly; left as-is for consistency with those existing guard patterns.

benhillis:
Fixed the wsl2 CI failures (commit 7a60b63b):

**Persisted path / comment 2 follow-up:** Persisting the canonical resolved path broke the persisted-path contract that the mount-state tests assert (the registry `Disk` value must equal the mount path that clients look up). The restore path in `_LoadDiskMount` already re-runs `ResolveDiskPath` (reparse rejection + canonicalize + pin), so the SYSTEM-side restore is re-validated regardless of what string is persisted. Reverted `SaveDiskState` to persist the original map key; security is unaffected.

**Error-code expectations:** Resolving the disk by handle now surfaces a missing file/path from the `CreateFileW` in `ResolveDiskPath` rather than from the HCS layer, so two `ErrorMessages` expectations changed (`/MountDisk/HCS/ERROR_*` -> `/MountDisk/ERROR_*`) for `--mount DoesNotExist` and the broken-distro VHD attach. Same human-readable message; the error is just caught one layer earlier.

Verified locally (Version=2): `ErrorMessages`, `TestBareMountVhd`, `MountVhdThroughSymlinkIsRejected`, `MountVhdThroughDirectoryJunctionSucceeds` all pass.

benhillis:
Addressed the latest review (commit a91041fe):

1. **Other `/HCS` expectations (good catch):** `PolicyTest::MountPolicyAllowed` asserts the same `--mount DoesNotExist` scenario, so dropped its `/HCS` segment too. I also checked `UnitTests::CorruptedVhd` (`--import-in-place` with an open handle): it correctly **keeps** `/HCS/ERROR_SHARING_VIOLATION` because the pin-handle open succeeds there and the sharing violation still surfaces from the HCS `AddVhd`. Both verified locally (Version=2): pass.

2. **Header comment scope:** Clarified that `ResolveDiskPath` rejects a reparse point only on the **final** path component; intermediate junctions are followed and canonicalized via the opened handle (which is what `MountVhdThroughDirectoryJunctionSucceeds` exercises).

3. **`pinHandle` C4101/`/WX`:** This is a false positive — `wil::unique_hfile` has a non-trivial destructor, so MSVC does not emit C4101 (the full `/WX` build passes: 0 warnings). I've annotated it `[[maybe_unused]]` anyway to document the lifetime-only intent and stop the recurring flag.

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -1825,6 +1849,95 @@ void WslCoreVm::GrantVmWorkerProcessAccessToDisk(_In_ PCWSTR Disk, _In_opt_ HAND
...
+
+    auto resolve = [&]() -> ResolvedDisk {
+        // Open the target without traversing a reparse point in the final path
+        // component. Opening the handle while impersonating the caller proves the
+        // caller can reach this exact object.
+        wil::unique_hfile handle{CreateFileW(
+            Disk, desiredAccess, shareMode, nullptr, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS | FILE_FLAG_OPEN_REPARSE_POINT, nullptr)};
+        THROW_LAST_ERROR_IF(!handle);
+
+        // Reject a final component that is a reparse point (junction or symlink).
```
OneBlue: Reading through this, it's looking like this will prevent users from mounting VHDs if the file itself is a reparse point. 

I'm not sure that we actually need to do this though. Reading `WslCoreVm::AttachDiskLockHeld()`, we always impersonate the user when we grant access to the VM, so we might be able to skip this entirely and only canonicalize for block devices

benhillis: I don't think that's totally true. There's also the path that attempts to restore the VHD mounts from the registry.

OneBlue: True, but we also impersonate the user on that path since we end up calling `WslCoreVm::AttachDiskLockHeld()`, which impersonates while calling `wsl::windows::common::hcs::GrantVmAccess()`

OneBlue: Looks like the latest iteration takes that into account

OneBlue: Ah and we didn't actually impersonate. I see

test/windows/MountTests.cpp
```diff
@@ -460,6 +460,72 @@ class MountTests
...
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(L"--mount " + symlink.wstring() + L" --vhd --bare"), (DWORD)0);
+
+        const auto disk = GetBlockDeviceInWsl();
+        VERIFY_IS_TRUE(IsBlockDevicePresent(disk));
+
+        WaitForVmTimeout(keepAlive);
+
+        // Recreating the VM restores the persisted disk mount; the symlinked VHD must re-attach.
+        VERIFY_IS_TRUE(IsBlockDevicePresent(disk));
+
```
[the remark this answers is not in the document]
OneBlue: This is true

test/windows/MountTests.cpp
```diff
@@ -460,6 +460,72 @@ class MountTests
...
+        SKIP_UNSUPPORTED_ARM64_MOUNT_TEST();
+
+        const auto symlink = std::filesystem::absolute(L"TestVhdSymlinkRestore.vhd");
+        DeleteFileW(symlink.c_str());
+
+        const auto absoluteTarget = std::filesystem::absolute(TEST_MOUNT_VHD);
+
+        if (!CreateSymbolicLinkW(symlink.c_str(), absoluteTarget.c_str(), SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE))
+        {
+            LogSkipped("Unable to create a symbolic link (error %d); skipping", GetLastError());
```
OneBlue: Do we expect this to ever happen ? I think if it did it would be OK to fail the test

OneBlue (dismissed):
LGTM. Minor test comment

test/windows/MountTests.cpp
benhillis: agreed

OneBlue (approved):

PR #40785, opened 2026-06-12
wslc trust Windows host Trusted Store certs

yao-msft:
Enumerate Windows Trusted Root Store certs, export as a cert bundle and put in the vm /etc/ssl/certs

## PR Checklist

## Validation Steps Performed
Manually validated a custom cert signed registry fails when doing wslc push (unknown authority). After putting the cert in Trusted Root store, wslc push (with fresh session) succeeds. Also added test to validate the same behavior.


test/windows/wslc/e2e/WSLCE2ETlsRegistryTests.cpp
```diff
@@ -0,0 +1,262 @@
...
+        explicit TrustedRootCertificate(const CERT_CONTEXT& Cert)
+        {
+            m_store.reset(CertOpenStore(CERT_STORE_PROV_SYSTEM_W, 0, NULL, CERT_SYSTEM_STORE_LOCAL_MACHINE, L"ROOT"));
+            THROW_LAST_ERROR_IF_NULL(m_store);
+
+            PCCERT_CONTEXT added{};
+            THROW_IF_WIN32_BOOL_FALSE(CertAddEncodedCertificateToStore(
+                m_store.get(), X509_ASN_ENCODING, Cert.pbCertEncoded, Cert.cbCertEncoded, CERT_STORE_ADD_REPLACE_EXISTING, &added));
+            m_added.reset(added);
+        }
```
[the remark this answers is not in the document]
yao-msft: adding to CurrentUser will trigger "protected root" confirmation popup, and it'll block pipeline since no one will click it in ci

test/windows/wslc/e2e/WSLCE2ETlsRegistryTests.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    // Wraps DER bytes in a PEM block with the given label (e.g. "CERTIFICATE", "PRIVATE KEY").
+    std::string ToPem(const BYTE* Data, DWORD Size, const std::string& Label)
+    {
+        DWORD base64Size = 0;
+        THROW_IF_WIN32_BOOL_FALSE(CryptBinaryToStringA(Data, Size, CRYPT_STRING_BASE64, nullptr, &base64Size));
+        std::string base64(base64Size, '\0');
+        THROW_IF_WIN32_BOOL_FALSE(CryptBinaryToStringA(Data, Size, CRYPT_STRING_BASE64, base64.data(), &base64Size));
+        base64.resize(base64Size);
+        return std::format("-----BEGIN {}-----\n{}-----END {}-----\n", Label, base64, Label);
```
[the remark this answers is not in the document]
yao-msft: Read here https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptbinarytostringa
base64Size returned in write operation is 1 less than query, the code logic is correct

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -595,6 +559,47 @@ void WSLCSession::StartDockerd()
...
+    ServiceProcessLauncher launcher("/bin/sh", {"/bin/sh", "-c", script}, {}, WSLCProcessFlagsStdin);
+    auto process = launcher.Launch(*m_virtualMachine);
+
+    // Stream the PEM bundle.
+    {
+        auto io = CreateIOContext();
+        io.AddHandle(
+            std::make_unique<io::WriteHandle>(io::HandleWrapper{process.GetStdHandle(WSLCFDStdin)}, std::vector<char>{pem.begin(), pem.end()}),
+            MultiHandleWait::CancelOnCompleted);
+        io.Run(std::chrono::milliseconds(c_processTerminateTimeoutMs));
```
OneBlue: This timeout is for an unrelated operation (terminating dockerd / containerd). 

For an operation like this, I think it would be OK to run this with let's say a one minute timeout to be safe.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -595,6 +559,47 @@ void WSLCSession::StartDockerd()
...
+    }
+
+    const auto result = process.WaitAndCaptureOutput(c_processTerminateTimeoutMs);
```
OneBlue: If we expect the process to write on stdout / stderr, capturing those only after stdin is fully written can lead to a pipe deadlock. Instead of doing those sequentially, I recommend [something like this](https://github.com/microsoft/WSL/blob/aef5b0ff335db49323331d94f12b1449e862c1f3/test/windows/WSLCTests.cpp#L3727) 

(we can manually add a handle to stop on session termination as well, to be safe)

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -595,6 +559,47 @@ void WSLCSession::StartDockerd()
...
+
+    WSL_LOG(
+        "InstalledTrustedRootCertificates",
+        TraceLoggingValue(c_certPath, "Path"),
+        TraceLoggingValue(static_cast<uint64_t>(pem.size()), "BundleBytes"));
+}
+catch (...)
+{
+    // Best-effort: failing to install the host's trusted roots must not prevent the session from starting.
+    LOG_CAUGHT_EXCEPTION_MSG("Failed to install trusted root certificates into the VM");
```
OneBlue, line 638: We should display a warning if we hit this

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
```diff
@@ -515,19 +518,59 @@ std::pair<RunningWSLCContainer, std::string> StartLocalRegistry(IWSLCSession& se
...
 
-    auto address = std::format("127.0.0.1:{}", port);
-    auto url = std::format(L"http://{}/v2/", wsl::shared::string::MultiByteToWide(address));
+    if (useTls)
+    {
+        auto inspect = container.Inspect();
+        THROW_HR_IF(E_UNEXPECTED, inspect.NetworkSettings.Networks.empty());
+        auto address = std::format("{}:{}", inspect.NetworkSettings.Networks.begin()->second.IPAddress, port);
+
+        // Wait until the registry is serving by probing its own loopback from inside the container.
```
OneBlue: Oh I have a fix for this issue in #40145. Let me extract it into its own PR so we can merge this separately

src/windows/wslcsession/WindowsCertStore.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include "WindowsCertStore.h"
+#include <optional>
+#include <set>
+#include <wincrypt.h>
+
+namespace {
+
+// Converts a single DER-encoded certificate into a PEM block.
+// Returns nullopt if the certificate cannot be encoded.
+std::optional<std::string> TryEncodeCertificateAsPem(const CERT_CONTEXT& Cert)
```
OneBlue, line 25: nit: I recommend throwing an exception if we fail to encode the certificate, that way we can display a warning at the end with something like: 

```
wsl: failed to import <count> certificates
```

src/windows/wslcsession/WindowsCertStore.cpp
```diff
@@ -0,0 +1,106 @@
...
+    std::string pem(pemSize, '\0');
+    if (!CryptBinaryToStringA(Cert.pbCertEncoded, Cert.cbCertEncoded, CRYPT_STRING_BASE64HEADER, pem.data(), &pemSize))
+    {
+        LOG_LAST_ERROR_MSG("CryptBinaryToStringA (encode) failed for a root certificate; skipping it");
+        return std::nullopt;
+    }
+
+    // The pemSize after the actual write call does not include terminating null,
+    // and is 1 less than the value returned by the earlier query call.
+    pem.resize(pemSize);
```
OneBlue, line 43: In that case, could we allocate the string to `pemSize -1` directly ?

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -595,6 +559,47 @@ void WSLCSession::StartDockerd()
...
+    if (pem.empty())
+    {
+        WSL_LOG("InstallTrustedRootCertificatesSkipped");
+        return;
+    }
+
+    // dockerd and containerd read the certificates found in /etc/ssl/certs into
+    // their default system certificate pool.
+    constexpr auto c_certPath = "/etc/ssl/certs/wsl-windows-roots.pem";
+    const auto script = std::format("set -e; mkdir -p /etc/ssl/certs && cat > {}", c_certPath);
```
OneBlue: nit: I think it would be OK to assume that `/etc/ssl/certs` exist here and just simplify to: 

`cat > '{}'`

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -595,6 +559,47 @@ void WSLCSession::StartDockerd()
...
+
+    ServiceProcessLauncher launcher("/bin/sh", {"/bin/sh", "-c", script}, {}, WSLCProcessFlagsStdin);
```
OneBlue: nit: I also recommend passing `--norc` here to be safe

OneBlue (commented):
Change looks great ! Couple minor comments

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
OneBlue: Here's the PR: https://github.com/microsoft/WSL/pull/40797

With this, you should be able to remove the retry logic

src/windows/wslcsession/WindowsCertStore.cpp
yao-msft: I have not tried locally but the api doc says we have to allocate the size including nul. I don't understand why though..

OneBlue: That's OK because std::string always allocates an extra byte for the '\0'. We do this in other places as well

src/windows/wslcsession/WindowsCertStore.cpp
yao-msft: throwing will stop following certs to be included, I'll add a count if nulopt is returned

OneBlue: Yeah I meant throwing, and catching +logging it in the parent function

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -598,6 +598,7 @@ catch (...)
 {
     // Best-effort: failing to install the host's trusted roots must not prevent the session from starting.
     LOG_CAUGHT_EXCEPTION_MSG("Failed to install trusted root certificates into the VM");
+    EMIT_USER_WARNING(Localization::MessageWslcInstallCertsFailed());
```
OneBlue: nit: I recommend including the error code in the warning if we ever hit this (we can use `wsl::windows::common::wslutil::GetErrorString()` to get it)

OneBlue (dismissed):
LGTM. One minor comment, but feel free to address in a followup PR

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -555,6 +561,44 @@ void WSLCSession::StartDockerd()
...
+        WSL_LOG("InstallTrustedRootCertificatesSkipped");
+        return;
+    }
+
+    // dockerd and containerd read the certificates found in /etc/ssl/certs into
+    // their default system certificate pool.
+    constexpr auto c_certPath = "/etc/ssl/certs/wsl-windows-roots.pem";
+    const auto script = std::format("cat > '{}'", c_certPath);
+
+    ServiceProcessLauncher launcher("/bin/sh", {"/bin/sh", "--norc", "-c", script}, {}, WSLCProcessFlagsStdin);
```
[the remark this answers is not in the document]
OneBlue: It's supported with our version of sh, so this is OK

test/windows/wslc/e2e/WSLCE2EHelpers.cpp
OneBlue: That change has been merged so we can remove this retry logic

OneBlue (approved):

PR #40788, opened 2026-06-12
Fix wslc create network failure without driver specified

chemwolf6922:

The WSLCSession::CreateNetwork expects Options->Driver to be populated. However, when the driver is not specified in the cli. It's skipped and default to NULL. This PR populates it with the default "bridge" driver when not specified.

## PR Checklist

## Validation Steps Performed

Enable test:
WSLCE2ETests::WSLCE2ENetworkCreateTests::WSLCE2E_Network_Create_DefaultDriver_Success


beena352 (dismissed):

src/windows/wslc/tasks/NetworkTasks.cpp
```diff
@@ -86,10 +86,7 @@ void CreateNetwork(CLIExecutionContext& context)
         options.Labels.push_back(validation::ParseLabel(label));
     }
 
-    if (context.Args.Contains(ArgType::Driver))
-    {
-        options.Driver = WideToMultiByte(context.Args.Get<ArgType::Driver>());
-    }
+    options.Driver = context.Args.Contains(ArgType::Driver) ? WideToMultiByte(context.Args.Get<ArgType::Driver>()) : "bridge";
```
AmelBawa-msft: Should we do from the CLI side or the runtime side? CC: @OneBlue 

Looking at volume implementation, the default value for driver is set from the runtime side:

Client side _(optional value, no default)_:
https://github.com/microsoft/WSL/blob/aef5b0ff335db49323331d94f12b1449e862c1f3/src/windows/wslc/services/VolumeService.cpp#L28-L31

Runtime side _(sets a default value)_:
https://github.com/microsoft/WSL/blob/aef5b0ff335db49323331d94f12b1449e862c1f3/src/windows/wslcsession/WSLCVolumes.cpp#L103-L104

dkbennett: If the service sets a default then the CLI should not be setting its own default.

dkbennett (changes requested):
I don't think it is correct for the CLI to be setting a default if the service sets one.  The service should handle that and the CLI should not be setting its own default.

dkbennett:
To add some additional context:
* If the driver needs to be specified, then it should be a required argument
* If the driver is only required in combination with another argument in the command, then we have ValidateArgumentsInternal() for that where such custom validation can be checked and an appropriate error issued.
* If no driver specified is valid and the service sets a default, then the CLI should do nothing and let the service validate it.
* If the driver specification is an Enum or has a limited set of valid options, then the CLI can validate that argument with a validator against the Enum map or similar (see how the Signal argument is processed for a complex example).

src/windows/wslc/tasks/NetworkTasks.cpp
OneBlue: Agreed with this, this change should be done in the service

chemwolf6922:
@OneBlue Do we want to allow "" being treated as default?

OneBlue:
> @OneBlue Do we want to allow "" being treated as default?

No I think we should consider null being the default, or "" should be rejected

OneBlue (approved):

PR #40789, opened 2026-06-12
CLI: Add global options, move --session to global options, add environment variable support

dkbennett:
This updates the CLI to add support for Global Options, and allowing Environment Variables to be used to set certain Arguments.

The default global option added is `--session <name>`, which promotes the existing per-command `--session` flag to a global option. It is now parsed in the root options-only pass before the subcommand is resolved, removed from each command''s `GetArguments()`, and read by `SessionTasks::CreateSession` from `context.GlobalArgs` instead of `context.Args`.

The environment variable added is for setting NoColor, with the unofficial standard `NO_COLOR` env var support.

The design is such that we reuse the ArgumentDefinitions for defining global options and environment variable arguments like any other argument. This is intended for just the Root, but it is possible for individual commands to also have environment variables for their options, should the need arise.

Global options can be accessed off the context via `GlobalArgs`.
Environment variable-populated args will also be added to GlobalArgs, even if they are not explicitly in the global args (for example, NoColor is not specified on command line and is only set via environment variable, but it is still accessed as if it were a GlobalArg).

Many tests added for the parser changes, new command line parsing, environment variable handling as well as new parser and command line cases. The parser test suite''s `Globals` set is intentionally decoupled from `RootCommand::GetGlobalArguments()` and uses a synthetic `{Quiet (flag), Session (value)}` set so parser coverage stays stable as production globals evolve. All `--session` usages in E2E tests and helpers have been reordered to come before the subcommand.

Intentionally not included:
* Help display of available global options and environment variables - this will be included in the larger help polish.

Previously listed as intentionally excluded but now included in this PR:
* Moving `--session` to global options (now done; `Argument::Create(ArgType::Session)` removed from all 34 per-command sites and added to `RootCommand::GetGlobalArguments()`).

Previously included but now removed:
* The placeholder `-D, --debug` global option and `WSLC_CLI_DEBUG` env binding were removed. They were only ever stubs (Debug had no behavior) and the real Reporter-driven debug flag will be wired up separately when it actually does something.

## PR Checklist

- [x] **Communication:** I''ve discussed this with core contributors already. If work hasn''t been agreed, this work might be rejected

See Copilot''s first comment.

## Validation Steps Performed
* Unit tests pass
* Manual verification of the command line looks correct


src/windows/wslc/core/Main.cpp
```diff
@@ -92,24 +114,17 @@ try
...
         LOG_CAUGHT_EXCEPTION();
-
-        // Using WSL shared utility to get the HRESULT from the caught exception.
-        // CLIExecutionContext is a derived class of wsl::windows::common::ExecutionContext.
         result = wil::ResultFromCaughtException();
 
-        // If the user pressed Ctrl-C, acknowledge the cancellation and exit.
         if (context.CancelEvent && context.CancelEvent.is_signaled())
         {
             fwprintf(stderr, L"\nCancelled.\n");
```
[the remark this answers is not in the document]
dkbennett: This loc string was added in error, I removed it.

src/windows/wslc/core/Main.cpp
```diff
@@ -81,6 +81,28 @@ try
...
         Invocation invocation{std::move(args)};
+
+        // Globals scan: consume only the global options we recognize at the
+        // front of the invocation. Anything else (subcommands, unknown options,
+        // --help, --version, malformed tokens) is left in place for the regular
+        // pipeline to parse and report against the right command.
+        auto cliGlobals = rootCommand->GetGlobalArguments();
+        auto envOnly = rootCommand->GetEnvArguments();
+        auto envDefs = cliGlobals;
+        envDefs.insert(envDefs.end(), envOnly.begin(), envOnly.end());
```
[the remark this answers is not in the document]
dkbennett: Fixed this a different way.

src/windows/wslc/core/EnvironmentOptions.cpp
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        std::wstring value(len - 1, L'\0');
+        DWORD got = ::GetEnvironmentVariableW(name, value.data(), len);
+        if (got == 0 || got >= len)
+        {
+            return std::nullopt;
+        }
+        return value;
+    }
```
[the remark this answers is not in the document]
dkbennett: Fixed this a different way and made presence-only env variables declarable.

src/windows/wslc/core/EnvironmentOptions.cpp
```diff
@@ -0,0 +1,122 @@
...
+        const wchar_t* Name;
+        ArgType Type;
+    };
+
+    // Many-to-one allowed: alternate spellings for the same ArgType (e.g. NO_COLOR and WSLC_CLI_NO_COLOR).
+    constexpr EnvBinding c_bindings[] = {
+        {L"WSLC_CLI_DEBUG", ArgType::Debug},
+        {L"WSLC_CLI_NO_COLOR", ArgType::NoColor},
+        {L"NO_COLOR", ArgType::NoColor},
+    };
```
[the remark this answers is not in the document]
dkbennett: removed the duplicate, that was leftover from a previous refactor.

src/windows/wslc/core/EnvironmentOptions.cpp
```diff
@@ -0,0 +1,154 @@
...
+            }
+            if (any)
+            {
+                target.Add(arg.Type(), true);
+            }
+            break;
+        }
+        case Kind::Value:
+        {
+            // TODO: no defined policy for multiple value bindings; first wins.
```
AmelBawa-msft: Just to clarify, does this mean if an option can be provided several times? like --opt "A" --opt"B" etc...

src/windows/wslc/core/EnvironmentOptions.h
```diff
@@ -0,0 +1,56 @@
...
+
+// Many-to-one allowed: multiple env var names may bind to one ArgType
+// (e.g. NO_COLOR and WSLC_CLI_NO_COLOR both map to ArgType::NoColor).
+// The vendor-specific WSLC_CLI_NO_COLOR keeps truthy-gating so users can
+// explicitly opt back in with WSLC_CLI_NO_COLOR=0; the spec-defined
+// NO_COLOR is presence-only as required by https://no-color.org.
+constexpr EnvBinding c_envBindings[] = {
+    {L"WSLC_CLI_DEBUG", ArgType::Debug},
+    {L"WSLC_CLI_NO_COLOR", ArgType::NoColor},
+    {L"NO_COLOR", ArgType::NoColor, true},
```
AmelBawa-msft: Love it!

AmelBawa-msft (dismissed):

src/windows/wslc/core/EnvironmentOptions.cpp
dkbennett: This is only for environment variables - it means if you have a many to one mapping, similar to the NoColor setting (where we have the informal standard NO_COLOR and the client-specific WSLC_CLI_NO_COLOR) what happens if you have multiple environment variables specified with different values?  In the case of Flags, we OR them so if any are true, the variable is set.  In the case of Value kind args, this policy is intentionally undefined and left as a "TODO" because we dont have anything that falls into this category yet. We don't have any value kind environment variables yet, and no situations currently where we'd support two different environment variables to handle it, so the behavior is left undefined, so at present the first one wins as default behavior.

Separately, related to your question, for flag arguments in the command line, this PR does enable multiple *flag* variables specified on the command line, and those duplicates are ignored. Previously we'd throw an argument error for this sort of thing. Value arguments like in your example will still be an error unless that value argument is allowed to have multiples. Example would be "wslc run --rm --rm ..." Before this change, that would be an argument error on the second '--rm' flag, but with this PR that will be silently ignored. This is also how Docker and other CLIs work, where duplicate flags are no-ops.

src/windows/wslc/core/EnvironmentOptions.h
dkbennett: I'm happy to follow the NO_COLOR informal standard, and to make sure we follow it, we load environment variables first before doing any command line parsing, so any errors will not be in color accidentally prior to parsing the color variable.  Its set up so we can set color disabled before we do anything at all that prints in color.

AmelBawa-msft (dismissed):

src/windows/wslc/core/EnvironmentOptions.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    EnvironmentOptions.cpp
+
+--*/
+#include "precomp.h"
+#include "EnvironmentOptions.h"
```
OneBlue, line 11: This feels a bit overkill for our usecase. Given the small number of env variables that we'll have, I think we could simplify this to a method that runs before the command line parser does, and do something like: 

```
void ApplyEnvVariables()
{
    if (IsTruthy("NO_COLOR")
    {
        // Set global no-color flag
    }
}
```

For reference, [this is how we handle this in wsl.exe](https://github.com/microsoft/WSL/blob/08b28b7cb951f5095fc90b8de75d8ec5762e988a/src/windows/common/WslClient.cpp#L1869)

src/windows/wslc/core/EnvironmentOptions.h
```diff
@@ -0,0 +1,56 @@
...
+};
+
+// Many-to-one allowed: multiple env var names may bind to one ArgType
+// (e.g. NO_COLOR and WSLC_CLI_NO_COLOR both map to ArgType::NoColor).
+// The vendor-specific WSLC_CLI_NO_COLOR keeps truthy-gating so users can
+// explicitly opt back in with WSLC_CLI_NO_COLOR=0; the spec-defined
+// NO_COLOR is presence-only as required by https://no-color.org.
+constexpr EnvBinding c_envBindings[] = {
+    {L"WSLC_CLI_DEBUG", ArgType::Debug},
+    {L"WSLC_CLI_NO_COLOR", ArgType::NoColor},
```
OneBlue: Unless we have a usecase for it, I think just having "NO_COLOR" for now would be best to keep things simple

OneBlue (commented):
Thank you for doing this ! 

I think the support "global" argument is going to make our life a lot easier (especially for --session) but I think we can simplify the env variable side. I think we'll realistically never need to look at more than one or two env variables, so we can probably simplify a bit there

src/windows/wslc/core/EnvironmentOptions.h
dkbennett: I think it's highly likely we'd add WSLC_CLI_SESSION environment variable to handle the developer convenience use case, and I like debug being in there so it is hit early and can cover any debug messages we want in the parser and for our own convenience of not having to type it every time.

Most of Docker's global options also have environment variable support in the same way, so if we're keeping similar functionality to docker we'll likely want to do this too.

src/windows/wslc/core/EnvironmentOptions.cpp
dkbennett: NO_COLOR is a bit different in that it doesn't use truthy values, it uses presence only so even if you say "NO_COLOR=false" that's still considered NO_COLOR enabled, otherwise we dont' follow the no-color standard.  That's why we have the truthy vs non-truthy values.

One simplification might be that we treat all boolean environment variables as presence and then we dont need the truthy distinction at all, and we can add it later when we have a strong case for needing truthy. That's a straightforward simplification. We could make the debug setting operate by presence as well, I dont really see a strong need for it to be truthy, though I think truthy makes sense - if we set debug = false then I would expect that to turn it off, but presence doesn't work that way. If we add truthy later to it though that might break some folks currently using it as presence. So for future compatibility Truthy might be a smart way to go unless the var explcitly needs presence like NO_COLOR.

There's also the matching of most global variables in Docker having environment variable settings available. Config, host, context, api version, default platform, etc, and a few that dont' have global options but are still recognized, like NO_COLOR.

src/windows/wslc/core/EnvironmentOptions.h
```diff
@@ -0,0 +1,44 @@
...
+{
+    const wchar_t* Name;
+    ArgType Type;
+};
+
+// Many-to-one allowed: multiple env var names may bind to one ArgType.
+constexpr EnvBinding c_envBindings[] = {
+    {L"WSLC_CLI_DEBUG", ArgType::Debug},
+    {L"NO_COLOR", ArgType::NoColor},
+};
```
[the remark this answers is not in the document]
dkbennett: I removed this to simplify. Updated description.

src/windows/wslc/arguments/ArgumentParser.cpp
```diff
@@ -281,8 +343,10 @@ ParseArgumentsStateMachine::State ParseArgumentsStateMachine::ProcessAliasArgume
         return {};
     }
 
-    // Boolean flag - add it and process any adjoined flags
-    m_executionArgs.Add(firstArg->Type(), true);
+    // Boolean flag - add it and process any adjoined flags. Once we have added a
+    // flag to m_executionArgs for this token, stopOnUnknown no longer applies for
+    // mid-chain unknowns; the token has already been claimed.
+    AddFlag(firstArg->Type());
```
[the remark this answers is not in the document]
dkbennett: Determined this is simply an unsupported scenario and documented as such. Ther'e sonly one command where you could even validly do global + non-global aliased flags together, and that is the root command. It is strange to do this and isn't a supported scenario.

src/windows/wslc/core/Main.cpp
```diff
@@ -72,6 +69,19 @@ try
 
     std::unique_ptr<Command> command = std::make_unique<RootCommand>();
 
+    // Stable handle to the root command; `command` is moved during subcommand resolution.
+    const Command* const rootCommand = command.get();
```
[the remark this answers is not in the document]
dkbennett: Fixed this by removing this handle...it isn't needed.

src/windows/wslc/core/EnvironmentOptions.h
dkbennett: Kept debug since it will be used shortly (along with NO_COLOR), but got rid of the second way to specify NO_COLOR and simplified all flag variables to be presence-only. We can change that in the future if we come across one that needs to not be. This also reduced test coverage and cut about half the environment code out so it's only about 70 lines now, still data-driven and tied neatly into GlobalArgs.

src/windows/wslc/core/EnvironmentOptions.cpp
dkbennett: Removed the 2nd no color specifier, so we just use NO_COLOR and got rid of truthy detection and all the other code surrounding that. It's just a read for presence or value and placing into GlobalArg as if it were a flag or value argument on command line.

test/windows/wslc/WSLCCLIEnvironmentOptionsUnitTests.cpp
```diff
@@ -0,0 +1,176 @@
...
+    {
+        ClearBoundEnvVars();
+        return true;
+    }
+
+    TEST_METHOD_CLEANUP(TestMethodCleanup)
+    {
+        ClearBoundEnvVars();
+        return true;
+    }
```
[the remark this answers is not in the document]
dkbennett: Fixed.

src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -43,6 +43,7 @@ _(Command,        "command",             NO_ALIAS,          Kind::Positional,  L
 _(ContainerId,    "container-id",        NO_ALIAS,          Kind::Positional,  Localization::WSLCCLI_ContainerIdArgDescription()) \
 _(Cpus,           "cpus",                NO_ALIAS,          Kind::Value,       Localization::WSLCCLI_CpusArgDescription()) \
 _(Force,          "force",               L"f",              Kind::Flag,        Localization::WSLCCLI_ForceArgDescription()) \
+_(Debug,          "debug",               L"D",              Kind::Flag,        Localization::WSLCCLI_DebugArgDescription()) \
```
OneBlue: I'm not sure how much we should add --debug given that we don't have much to print when this is set (all the logging is coming from the service in our usecase).

src/windows/wslc/core/CLIExecutionContext.cpp
```diff
@@ -0,0 +1,24 @@
...
+    WI_ASSERT(!CancelEvent);
+    CancelEvent.create(wil::EventOptions::ManualReset);
+    return CancelEvent.get();
+}
+
+// This method should be idempotent.
+void CLIExecutionContext::ApplyGlobalOptions()
+{
+    // TODO: Add per-global side effects here as features land.
+}
```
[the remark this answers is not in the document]
dkbennett: Yes, this is fine, the thing that will consume it is in another PR.

test/windows/wslc/WSLCCLIEnvironmentOptionsUnitTests.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+Module Name:
+
+    WSLCCLIEnvironmentOptionsUnitTests.cpp
+
+Abstract:
+
+    Unit tests for ApplyEnvironmentOptions. Every binding in c_envBindings is
+    presence-only: defining the variable sets the option regardless of value;
+    unsetting it is the only opt-out. NO_COLOR follows https://no-color.org.
```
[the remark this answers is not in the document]
dkbennett: This is why I hate adding any comments ever.

dkbennett:
Test failure looks unrelated, rate limit hit on a request.

src/windows/wslc/core/CLIExecutionContext.cpp
```diff
@@ -0,0 +1,24 @@
...
+{
+    WI_ASSERT(!CancelEvent);
+    CancelEvent.create(wil::EventOptions::ManualReset);
+    return CancelEvent.get();
+}
+
+// This method should be idempotent.
+void CLIExecutionContext::ApplyGlobalOptions()
+{
+    // TODO: Add per-global side effects here as features land.
```
OneBlue, line 21: Is this still needed ? It's looking like the nocolor env variable isn't wired here ?

src/windows/wslc/core/EnvironmentOptions.cpp
```diff
@@ -0,0 +1,79 @@
...
+    {
+        std::wstring value;
+        const HRESULT hr = wil::GetEnvironmentVariableW(name, value);
+        if (FAILED(hr))
+        {
+            return std::nullopt;
+        }
+        return value;
+    }
+    catch (...)
```
OneBlue, line 28: I don't recommend having a catch-all here since it could potentially hide issues. 

Probably also a good idea to throw `hr` is anything other than success or `ERROR_ENVVAR_NOT_FOUND`

(we can CATCH_LOG() in ApplyEnvironmentOptions())

test/windows/wslc/WSLCCLIEnvironmentOptionsUnitTests.cpp
```diff
@@ -0,0 +1,160 @@
...
+class WSLCCLIEnvironmentOptionsUnitTests
+{
+    WSLC_TEST_CLASS(WSLCCLIEnvironmentOptionsUnitTests)
+
+    // Tests touch process-wide env state. Capture pre-existing values in setup
+    // and restore them in cleanup so the suite is hermetic and doesn't clobber
+    // values the test host (or CI) may have set.
+    TEST_METHOD_SETUP(TestMethodSetup)
+    {
+        m_savedNoColor = CaptureEnv(L"NO_COLOR");
```
OneBlue, line 42: nit: I recommend using `ScopedEnvVariable` for this 

(To make this even more convenient, we could also add a `Set()` method so it can be set directly from that class)

OneBlue (approved):
LGTM, minor comments

src/windows/wslc/core/CLIExecutionContext.cpp
dkbennett: It will be as soon as I complete this PR.

src/windows/wslc/arguments/ArgumentDefinitions.h
dkbennett: There is no more Debug arg in the latest version.

src/windows/wslc/core/EnvironmentOptions.cpp
dkbennett: I'll fix this in a subsequent targeted PR along with the test

test/windows/wslc/WSLCCLIEnvironmentOptionsUnitTests.cpp
dkbennett: I'll fix this in a targeted PR so it doesnt do another round trip and can use the no-color arg.

PR #40794, opened 2026-06-12
Fix input validation gaps in driver arguments and network aliases

beena352:
1. --driver "" or --driver "   " was silently accepted by the CLI parser, passing an empty/whitespace driver name to the service layer. Added ArgType::Driver validation to reject these inputs, consistent with the existing ArgType::Network and ArgType::WorkDir checks.

2. The network-alias  in ContainerService::CreateInternal did not check for primary.empty(), allowing an empty-string network name (from a direct SDK/COM caller) to bypass the user-defined network requirement.

## PR Checklist

ArgumentValidation.cpp : Added ArgType::Driver case to Argument::Validate() matching the ArgType::WorkDir pattern. Rewrote ParseDriverOption to match `ParseLabel`'s structure and empty-key rejection.

ContainerService.cpp : Added primary.empty() to the alias guard condition. Defense-in-depth: the CLI already rejects empty --network values, but SDK/COM callers bypass CLI validation.

Resources.resw : One new localization strings: WSLCCLI_DriverEmptyError (with {FixedPlaceholder} for the arg name)

### DIFF
--- a/src/windows/wslc/arguments/ArgumentValidation.cpp
+++ b/src/windows/wslc/arguments/ArgumentValidation.cpp
@@ -424,20 +424,13 @@ std::pair<std::string, std::string> ParseLabel(const std::wstring& value)
 
 std::pair<std::string, std::string> ParseDriverOption(const std::wstring& value)
 {
-    std::pair<std::string, std::string> result{};
     auto pos = value.find('=');
     if (pos == std::wstring::npos)
     {
-        result.first = WideToMultiByte(value);
-    }
-    else
-    {
-        result.first = WideToMultiByte(value.substr(0, pos));
-        result.second = WideToMultiByte(value.substr(pos + 1));
+        return {WideToMultiByte(value), std::string{}};
     }
 
-    THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::WSLCCLI_DriverOptionKeyEmptyError(), result.first.empty());
-    return result;
+    return {WideToMultiByte(value.substr(0, pos)), WideToMultiByte(value.substr(pos + 1))};
 }
 
 } // namespace wsl::windows::wslc::validation
--- a/src/windows/wslc/arguments/ArgumentValidation.cpp
+++ b/src/windows/wslc/arguments/ArgumentValidation.cpp
@@ -106,6 +106,17 @@ void Argument::Validate(const ArgMap& execArgs) const
         break;
     }
 
+    case ArgType::Driver:
+    {
+        const auto& value = execArgs.Get<ArgType::Driver>();
+        if (value.empty() ||
+            std::all_of(value.begin(), value.end(), [](wchar_t c) { return std::iswspace(static_cast<wint_t>(c)); }))
+        {
+            throw ArgumentException(Localization::WSLCCLI_DriverEmptyError(m_name));
+        }
+        break;
+    }
+
     case ArgType::Network:
     {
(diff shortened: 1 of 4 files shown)


AmelBawa-msft (commented):
I’m not sure empty keys have a meaningful use case. Docker does allow them, which is why I originally enabled the behavior, but I’m fine tightening this unless we have a specific need to support them?

```sh
root@ [ / ]# docker network create --opt "=value" --opt "a=A" test
fd5591b5....
root@ [ / ]# docker network inspect test
[
    {
        "Name": "test",
        ...
        "Options": {
            "": "value",
            "a": "A"
        },
        ...
    }
]
```

beena352:
> I’m not sure empty keys have a meaningful use case. Docker does allow them, which is why I originally enabled the behavior, but I’m fine tightening this unless we have a specific need to support them?
> 
> ```shell
> root@ [ / ]# docker network create --opt "=value" --opt "a=A" test
> fd5591b5....
> root@ [ / ]# docker network inspect test
> [
>     {
>         "Name": "test",
>         ...
>         "Options": {
>             "": "value",
>             "a": "A"
>         },
>         ...
>     }
> ]
> ```

Good catch, Docker does allow empty keys for driver options. Reverted the ParseDriverOption change and restored the original behavior + test cases. Kept the --driver empty/whitespace validation and the primary.empty() alias guard since those are unrelated.

AmelBawa-msft:
>  Kept the --driver empty/whitespace validation and the primary.empty() alias guard since those are unrelated.

Should we rely on the runtime to validate the driver value here instead? Seems like the runtime today allows an empty volume driver (uses default) but rejects an empty network (I commented on that in this PR #40788). Docker allows empty for both values, but again, not sure what the exact requirements are here, as long as we are consistent 😊✨

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -106,6 +106,17 @@ void Argument::Validate(const ArgMap& execArgs) const
         break;
     }
 
+    case ArgType::Driver:
+    {
+        const auto& value = execArgs.Get<ArgType::Driver>();
+        if (value.empty() ||
```
dkbennett, line 112: I think a standard validator for checking an argument is not empty or whitespace would be useful  since we have now used it in multiple places. We have this for validating integers and such, so it makes sense to me that we'd have this as a standard validator that would likely be applicable for many value arguments in the future. A simple boolean result would allow us to still have separate exception messages for each argument but share the same empty/whitespace check.

AmelBawa-msft: I guess, I wonder if for those cases we should just fall back to using default (from runtime side). Again, I suspect this is a nit, but would be usefult for something like:

```
wslc network create --driver "$WSLC_DRIVER"
```

where if $WSLC_DRIVER is defined, then it is used, and if not the runtime would fallback to the default one. Again, not sure if those are valid use cases, but thinking out loud ✨

dkbennett: With the environment variable support pr we could add argument-specific environment variables such as this if we want to do that so the user can set their own default driver and not specify it, but I would think that's some nice convenience polish for GA and not something urgent.

With regards to falling bak on default, if the user enters emptystring or whitepsace for an argument value that is a big red flag to me that they made an argument CLI error, not that they intend to use a default value. I think that should be treated as an error and not a signal from the user to use the default. If they wanted to use the default they wouldn't specify anything at all.

PR #40798, opened 2026-06-13
CLI: Add Reporter for structured CLI output with VT color and level routing

dkbennett:

Adds a Reporter class for structured, level-aware console output in wslc. Reporter provides a `std::format`-style API that routes Output to stdout and Info/Warning/Error to stderr, with automatic SGR color wrapping (yellow for warnings, red for errors) and VT sequence stripping when the destination is redirected or color is disabled.

Reporter is backed by OutputChannel, a byte-sink class that owns VT enablement (RAII) on console handles and falls back to `fwprintf` for redirected output. Each write is a single kernel call (WriteConsoleW or fwprintf) so concurrent threads cannot interleave.

Main.cpp error handling and cancellation output now use Reporter. Other wslc commands can be converted in follow-up PRs.

## PR Checklist

**New files:**
- `OutputChannel.h/.cpp` — Byte sink: WriteConsoleW for consoles, fwprintf for redirected output. Owns `EnableVirtualTerminal` RAII. Provides `GetConsoleWidth()` (minus one column as an autowrap guard).
- `Reporter.h/.cpp` — `std::format`-style write API with `Output`/`Info`/`Warn`/`Error` convenience methods. VT `Sequence` arguments are automatically stripped when VT is off; color `Sequence` arguments are additionally stripped when color is disabled. Level-based SGR prefix/reset wrapping for warnings (yellow) and errors (red).
- `WSLCCLIReporterUnitTests.cpp` — 20 unit tests covering OutputChannel writes, Reporter formatting, Sequence stripping, color/no-color behavior, level routing, VT state queries, and console width.
- `WSLCCLITestHelpers.h` — `CapturePipe` (RAII pipe pair for capturing FILE* output) and `CaptureReporter`/`SplitCaptureReporter` test helpers.

**Modified files:**
- `CLIExecutionContext.h` — Holds `Reporter` as a value member. `NON_MOVABLE` (Reporter owns non-movable OutputChannels).
- `Main.cpp` — Error and cancellation output routed through `context.Reporter`. Removed standalone `EnableVirtualTerminal` RAII (now owned by OutputChannel).

## Validation Steps Performed
- 128/128 WSLCCLI unit tests pass (20 Reporter + 108 existing)
- Manual validation of colored output, redirected output, cancellation, and error formatting
- Multiple automated code review passes


src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,148 @@
...
+    if (WI_AreAllFlagsClear(m_enabledLevels, level))
+    {
+        return OutputWriter(*m_out, false); // enabled=false suppresses all output
+    }
+
+    const bool vtEnabled = m_out->IsVTEnabled();
+    const bool colorEnabled = vtEnabled && !m_noColor;
+    const bool routeToStderr = !vtEnabled && (level == Level::Error);
+    OutputChannel& target = routeToStderr ? *m_err : *m_out;
+    OutputWriter result{target, true, vtEnabled, colorEnabled};
```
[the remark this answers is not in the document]
dkbennett: This is a great catch, I''ve updated to set stderr to VT enabled, and all diagnostic input always goes to stderr (Debug, Warn, and Error) and only Info goes to stdout, so output redirection on stdout is always clean of errors and warnings and other diagnostic info, while they are all contained to stderr.

src/windows/wslc/core/OutputWriter.cpp
```diff
@@ -0,0 +1,268 @@
...
+    other.m_colorWritten = false;
+}
+
+OutputWriter& OutputWriter::operator=(OutputWriter&& other) noexcept
+{
+    if (this != &other)
+    {
+        Flush();
+
+        WI_ASSERT(&m_out == &other.m_out);
```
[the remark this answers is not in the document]
dkbennett: Fixed by removing this. The move-assignment operator is not needed.

src/windows/wslc/core/OutputWriter.h
```diff
@@ -0,0 +1,155 @@
...
+#include <filesystem>
+#include <memory>
+#include <sstream>
+#include <string>
+#include <type_traits>
+
+using namespace wsl::windows::common::vt;
+
+namespace wsl::windows::wslc {
+
```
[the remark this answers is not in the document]
dkbennett: Whoops, fixed.

src/windows/wslc/core/Reporter.h
```diff
@@ -0,0 +1,130 @@
...
+    };
+
+    // stdout/stderr, VT enabled when available.
+    Reporter();
+
+    // Both output and error go to outFile. Used for test capture.
+    Reporter(FILE* outFile);
+
+    // outFile receives non-error output; errFile receives error output when VT is unavailable.
+    Reporter(FILE* outFile, FILE* errFile);
```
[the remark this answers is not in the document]
dkbennett: Good catch, I fixed this and other comments to be more consistent.

src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,153 @@
...
+               WI_IsFlagSet(mode, ENABLE_VIRTUAL_TERMINAL_PROCESSING);
+    }
+
+    std::optional<size_t> GetConsoleWidth()
+    {
+        CONSOLE_SCREEN_BUFFER_INFO info{};
+        if (GetConsoleScreenBufferInfo(GetStdHandle(STD_OUTPUT_HANDLE), &info))
+        {
+            return static_cast<size_t>(info.dwSize.X);
+        }
```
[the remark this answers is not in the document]
dkbennett: I removed this method, it was leftover from previous iterations and is not needed and not really appropriate for this class.

src/windows/wslc/core/OutputWriter.cpp
```diff
@@ -0,0 +1,243 @@
...
+    if (m_consoleHandle != INVALID_HANDLE_VALUE)
+    {
+        DWORD written = 0;
+        WriteConsoleW(m_consoleHandle, text.data(), static_cast<DWORD>(text.size()), &written, nullptr);
+    }
+    else
+    {
+        // No newline appended; the buffer already contains all desired formatting.
+        fwprintf(m_file, L"%.*ls", static_cast<int>(text.size()), text.data());
+    }
```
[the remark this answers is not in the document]
dkbennett: good spot, fixed almost as suggested, removed the if guard since that can only be false if a constructor passed nullptr in, so I added WI_ASSERT in the constructors instead.

src/windows/wslc/core/Reporter.h
```diff
@@ -0,0 +1,108 @@
...
+    Reporter();
+
+    // Test constructor: both Info and diagnostics go to outFile (single-pipe capture).
+    // VT/color disabled.
+    Reporter(FILE* outFile);
+
+    // Info writes to outFile; Debug/Warning/Error write to errFile.
+    // Each channel's VT/color state is derived independently from its FILE*,
+    // so a redirected errFile stays plain even when outFile is a TTY.
+    Reporter(FILE* outFile, FILE* errFile);
```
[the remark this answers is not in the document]
dkbennett: Comments fixed to be more accurate here and in a few others.

src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,178 @@
...
+    if (m_out)
+    {
+        CloseOutputWriter();
+    }
+}
+
+bool Reporter::IsVTEnabled() const
+{
+    return m_out && m_out->IsVTEnabled();
+}
```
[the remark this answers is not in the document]
dkbennett: This is unnecessary method, I cut it.

src/windows/wslc/core/OutputWriter.cpp
```diff
@@ -0,0 +1,247 @@
...
+    {
+        return;
+    }
+
+    if (m_consoleHandle != INVALID_HANDLE_VALUE)
+    {
+        DWORD written = 0;
+        WriteConsoleW(m_consoleHandle, text.data(), static_cast<DWORD>(text.size()), &written, nullptr);
+    }
+    else
```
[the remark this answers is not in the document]
dkbennett: Added logging to writeconsole and the fwprintf and fflush calls.

src/windows/wslc/core/OutputWriter.h
```diff
@@ -0,0 +1,171 @@
...
+            m_buffer += t;
+        }
+        else if constexpr (std::is_same_v<T, bool>)
+        {
+            m_buffer.append(t ? L"true" : L"false");
+        }
+        else if constexpr (std::is_integral_v<T> || std::is_floating_point_v<T>)
+        {
+            m_buffer.append(std::to_wstring(t));
+        }
```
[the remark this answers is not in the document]
dkbennett: Fixed.

src/windows/wslc/core/Main.cpp
```diff
@@ -43,8 +43,9 @@ try
     auto coInit = wil::CoInitializeEx(COINIT_MULTITHREADED);
     wslutil::CoInitializeSecurity();
 
-    // The execution context must be declared after COM is initialized because it stores internal
-    // COM references.
+    wsl::windows::common::vt::EnableVirtualTerminal vtModeOut{GetStdHandle(STD_OUTPUT_HANDLE)};
+    wsl::windows::common::vt::EnableVirtualTerminal vtModeErr{GetStdHandle(STD_ERROR_HANDLE)};
```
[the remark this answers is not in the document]
dkbennett: Fixed

src/windows/wslc/core/Reporter.h
```diff
@@ -0,0 +1,107 @@
...
+{
+    enum class Level : uint32_t
+    {
+        None = 0x0,
+        Debug = 0x1,
+        Info = 0x2,
+        Warning = 0x4,
+        Error = 0x8,
+        Standard = Info | Warning | Error,
+        All = Debug | Info | Warning | Error,
```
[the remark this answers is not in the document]
dkbennett: This does in fact compile, and your proposed alteration is horrific.

src/windows/wslc/core/Reporter.h
```diff
@@ -0,0 +1,107 @@
...
+    // VT/color is enabled per-handle based on the current console mode.
+    Reporter();
+
+    // Test constructor: both Info and diagnostics go to outFile (single-pipe capture).
+    // VT/color disabled.
+    Reporter(FILE* outFile);
+
+    // Info writes to outFile; Debug/Warning/Error write to errFile.
+    // Each channel's VT/color state is derived independently from its FILE*,
+    // so a redirected errFile stays plain even when outFile is a TTY.
```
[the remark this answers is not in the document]
dkbennett: Congratulations, your incessant nitpicking of comments has made me - a lifelong comment enjoyer - not want to ever comment anything again. I will let the code speak for itself from now on and only document something really cryptic or for reference.

Thanks to this, I've gone and stripped most comments out of this PR or made them very terse. Your move.

src/windows/wslc/core/OutputWriter.h
```diff
@@ -0,0 +1,177 @@
...
+
+    // Silences all further writes. Called by Reporter::CloseOutputWriter(true).
+    void Disable();
+
+private:
+    HANDLE m_consoleHandle = INVALID_HANDLE_VALUE; // valid on console path only
+    FILE* m_file = nullptr;                        // non-null on FILE* path only
+    std::atomic_bool m_enabled = true;
+    bool m_VTEnabled;
+};
```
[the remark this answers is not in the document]
dkbennett: Fixed.

src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,183 @@
...
+        result.AddFormat(Format::Default);
+        break;
+    case Level::Warning:
+        result.AddFormat(Format::Fg::BrightYellow);
+        break;
+    case Level::Error:
+        result.AddFormat(Format::Fg::BrightRed);
+        break;
+    default:
+        THROW_HR(E_UNEXPECTED);
```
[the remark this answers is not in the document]
dkbennett: No, I think unexpected is correct since it is an enum created a compile time. It is a programming error if we reach this point, which is actually unexpected. Invalid arg would imply the user put in a bad argument, which is not the case here.

localization/strings/en-US/Resources.resw
```diff
@@ -2188,6 +2188,9 @@ Usage:
   <data name="WSLCCLI_AvailableArguments" xml:space="preserve">
     <value>The following arguments are available:</value>
   </data>
+  <data name="WSLCCLI_OperationCancelled" xml:space="preserve">
+    <value>Operation cancelled.</value>
```
OneBlue: nit: We could use the system error string for ERROR_CANCELLED for this

src/windows/wslc/core/OutputWriter.cpp
```diff
@@ -0,0 +1,273 @@
...
+    {
+        if (fwprintf(m_file, L"%.*ls", static_cast<int>(text.size()), text.data()) < 0)
+        {
+            const int err = errno;
+            LOG_HR_MSG(HRESULT_FROM_WIN32(ERROR_WRITE_FAULT), "fwprintf to redirected output failed (errno=%d)", err);
+        }
+
+        // Redirected stdout/stderr are fully buffered by the CRT; flush so
+        // std::endl/std::flush reach the OS as the writer's contract promises.
+        if (fflush(m_file) != 0)
```
OneBlue: I don't recommend manually flushing here. If the output is redirected, it shouldn't need flushing on every write since that implies that this stream isn't used interactively

src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+}
+
+bool Reporter::IsVTEnabled(Level level) const
+{
+    THROW_HR_IF(E_INVALIDARG, !IsSingleLevel(level));
+    return (level == Level::Output ? m_out : m_err)->IsVTEnabled();
+}
+
+std::optional<int> Reporter::GetConsoleWidth(Level level) const
```
OneBlue, line 58: This looks unused (both at the reporter & OutputChannel level)

src/windows/wslc/core/OutputWriter.cpp
```diff
@@ -0,0 +1,273 @@
...
+    {
+        return std::nullopt;
+    }
+
+    // Visible width is (Right - Left + 1); reserve one column as an autowrap guard so
+    // callers can write the full returned width without risking an unwanted line break.
+    return std::max(0, static_cast<int>(info.srWindow.Right) - static_cast<int>(info.srWindow.Left));
+}
+
+void OutputChannel::Disable()
```
OneBlue: What's the usecase for this ?

src/windows/wslc/core/OutputWriter.h
```diff
@@ -0,0 +1,168 @@
...
+    bool m_enabled;
+    bool m_vtEnabled;
+    bool m_colorEnabled;
+    bool m_written = false;
+    bool m_flushed = false;
+    bool m_colorWritten = false;
+
+    // 1 = level format fires on first text write. Set to 2 after a caller
+    // Sequence so the level format follows one write later.
+    size_t m_formatDelay = 1;
```
OneBlue: What's the usecase for this ?

src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    // Output to stdout; diagnostics to stderr. Per-channel VT decides SGR emission.
+    OutputChannel& target = (level == Level::Output) ? *m_out : *m_err;
+    const bool vtEnabled = target.IsVTEnabled();
+    const bool colorEnabled = vtEnabled && !m_noColor;
+    OutputWriter result{target, true, vtEnabled, colorEnabled};
+
+    switch (level)
+    {
+    case Level::Debug:
```
OneBlue: I'm not sure that we need that many levels. Right now we only have regular, warnings and error. I think it would be OK to keep as is for now

src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,196 @@
...
+        CloseOutputWriter();
+    }
+}
+
+bool Reporter::IsColorEnabled() const
+{
+    return !m_noColor;
+}
+
+void Reporter::SetNoColor(bool noColor)
```
OneBlue: nit: I recommend renaming to `SetColorEnabled` to be consistent with the getter

src/windows/wslc/core/Reporter.h
```diff
@@ -0,0 +1,94 @@
...
+
+    // Safe-write width in columns of the destination console for the given level, or
+    // std::nullopt when the destination is redirected. Callers that perform width-based
+    // truncation should skip it when this returns std::nullopt.
+    std::optional<int> GetConsoleWidth(Level level) const;
+
+private:
+    Reporter(std::shared_ptr<OutputChannel> outChannel, std::shared_ptr<OutputChannel> errChannel);
+
+    std::shared_ptr<OutputChannel> m_out;
```
OneBlue: Given that we're not subclassing OutputChannel, I don't think it needs to be a pointer here

src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+void Reporter::CloseOutputWriter(bool forceDisable)
+{
+    if (!m_out)
+    {
+        return;
+    }
+
+    if (forceDisable)
+    {
```
OneBlue: This branch looks unused

src/windows/wslc/core/OutputWriter.h
```diff
@@ -0,0 +1,168 @@
...
+    OutputWriter(OutputWriter&& other) noexcept;
+    OutputWriter& operator=(OutputWriter&&) = delete;
+
+    ~OutputWriter();
+
+    void AddFormat(const Sequence& sequence);
+    void ClearFormat();
+
+    template <typename T>
+    OutputWriter& operator<<(const T& t)
```
OneBlue: Given that our codebase uses std::format for things like these, I would recommend using std::format here as well (the main advantage would be that we don't have to manually implementing every type that we don't want to print, the std::formatters that we have already deal with that)

src/windows/wslc/core/OutputWriter.h
```diff
@@ -0,0 +1,168 @@
...
+            static_assert(!sizeof(T*), "OutputWriter has no overload for this type. Convert narrow strings to wide text.");
+        }
+    }
+
+    OutputChannel& m_out;
+    bool m_enabled;
+    bool m_vtEnabled;
+    bool m_colorEnabled;
+    bool m_written = false;
+    bool m_flushed = false;
```
OneBlue: Overall I don't think we need to manually take care of flushing. Either we have an interactive console, and we'll use WriteConsoleW, which doesn't need flushing, or we're printing to a redirected output, which doesn't need to be flushed

src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,196 @@
...
+} // namespace
+
+Reporter::Reporter() :
+    Reporter(
+        std::make_shared<OutputChannel>(GetStdHandle(STD_OUTPUT_HANDLE), stdout, QueryVTEnabled(GetStdHandle(STD_OUTPUT_HANDLE))),
+        std::make_shared<OutputChannel>(GetStdHandle(STD_ERROR_HANDLE), stderr, QueryVTEnabled(GetStdHandle(STD_ERROR_HANDLE))))
+{
+}
+
+Reporter::Reporter(FILE* outFile) :
```
OneBlue: I think it would be easier for the caller to pass the `OutputChannel` here. Then we'd only need one constructor

src/windows/wslc/core/Main.cpp
```diff
@@ -43,8 +43,9 @@ try
     auto coInit = wil::CoInitializeEx(COINIT_MULTITHREADED);
     wslutil::CoInitializeSecurity();
 
-    // The execution context must be declared after COM is initialized because it stores internal
-    // COM references.
+    [[maybe_unused]] wsl::windows::common::vt::EnableVirtualTerminal vtModeOut{GetStdHandle(STD_OUTPUT_HANDLE)};
```
OneBlue: We're acquiring the stderr & stderr handles in many places. Could we do it just once and reuse the same handles ?

src/windows/wslc/core/OutputWriter.h
```diff
@@ -0,0 +1,168 @@
...
+#include <type_traits>
+
+namespace wsl::windows::wslc {
+
+using namespace wsl::windows::common::vt;
+
+// Long-lived write destination used exclusively by Reporter.
+// Console handle: WriteString() calls WriteConsoleW() directly.
+// Redirected/FILE*: WriteString() calls fwprintf() without appending a newline.
+struct OutputChannel
```
OneBlue: Looking through this, we have Reporter -> OutputWriter-> OutputChannel  -> <terminal>, which is quite a bit of levels of indirections. 

If we switch to std::format as our main mecanism, I think we could skip the OutputWriter level

src/windows/wslc/core/Reporter.cpp
dkbennett: It is at present unused, but the help / tableoutput refactor will use it, I just wanted to keep all the reporter changes contained to the reporter PR so it doesn't need to be changed later. I can refactor it if you want, but it will definitely be used with the help (tableoutput will use a reporter directly and get the width for its printing calculations and determining color output, etc).

src/windows/wslc/core/OutputWriter.h
dkbennett: This is how we know to add a reset at the end in a chain of formatting.

src/windows/wslc/core/OutputWriter.h
dkbennett: You can use a std format into it with one line.

Writer() << std::format(L"{}whatever", Format::Bright);

Its the same result, and that's what a lot of the code will do when converted.

src/windows/wslc/core/OutputWriter.h
dkbennett: Its not a flush call in the filesystem sense, its a flush to when we do the actual write to WriteConsoleW or ffprintf (based on where we are writing). It builds the string and then makes a call (which is the flush), adding formatting as needed. The m_flushed is whether we need to make a write call or not, since an OutputWriter could end up making several in its lifetime. It could be long-lived for a function builder or a one-shot. Either way this is control over when the write happens.

src/windows/wslc/core/OutputWriter.h
OneBlue: What I was recommending was to not use "operator <<" as the mechanism to print here, but instead do something like:

```
Writer.Write(std::format(...)")
```

That way, all the types that we have an std::formatter for today will immediately be supported without us having to explicitly add overload for them

src/windows/wslc/core/Reporter.cpp
dkbennett: There's multiple constructors for testing purposes, the no-argument constructor is the one production uses in a single place (in the CLIExecutionContext.h) which sets stdout and stdin as the two channels.

src/windows/wslc/core/Main.cpp
dkbennett: I'll try moving the virtual terminal mode into reporter so it is the only thing holding them and then these lines probably aren't needed.

src/windows/wslc/core/OutputWriter.h
dkbennett: Per discussion, going to rework reporter, probably reducing it to reporter + outputchannel and use std::format as suggested here.

src/windows/wslc/core/OutputChannel.cpp
```diff
@@ -0,0 +1,87 @@
...
+    if (m_file == nullptr)
+    {
+        return;
+    }
+
+    if (fwprintf(m_file, L"%.*ls", static_cast<int>(text.size()), text.data()) < 0)
+    {
+        const int err = errno;
+        LOG_HR_MSG(HRESULT_FROM_WIN32(ERROR_WRITE_FAULT), "fwprintf to redirected output failed (errno=%d)", err);
+    }
```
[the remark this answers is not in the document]
dkbennett: We intentionally remove the fflush to reduce kernel translations. It is only applicable (typically) on stdout as stderr has no buffering, and for a CLI this should not be noticeable as the close will flush the buffer as the program exits.  We could enable the flush on stdout only but seeing as how no other code in this codebase does that, we'll leave it alone for now.

src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+std::optional<int> Reporter::GetConsoleWidth(Level level) const
+{
+    return ChannelFor(level).GetConsoleWidth();
+}
+
+bool Reporter::IsLevelEnabled(Level level) const noexcept
+{
+    return WI_AreAllFlagsSet(m_enabledLevels, level);
+}
```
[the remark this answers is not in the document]
dkbennett: no more level none or bitmask

src/windows/wslc/core/Main.cpp
dkbennett: Fixed to move the vt into the OutputChannel, which holds it for the lifetime. This is all wrapped neatly into Reporter now and no longer needs to be in main.

src/windows/wslc/core/Reporter.cpp
dkbennett: Trimmed down the constructors to two:   One is used in production and the other is used for testing. Did that for outputchannel and reporter.

src/windows/wslc/core/OutputWriter.h
dkbennett: Completely reworked this so no longer applicable.

src/windows/wslc/core/OutputWriter.h
dkbennett: Updated to be std::format based now.

src/windows/wslc/core/OutputWriter.h
dkbennett: Removed it, we handle the wrapping more simply now.

src/windows/wslc/core/OutputWriter.cpp
dkbennett: Removed it. This was only practical for toggling level sets and that was overkill for this. The only level we would conceivably want to disable is debug, and since it isn't used by anything, I simply removed it. Should we wnat that later, it's easy to add it in and we can add a debug specific enable setting for the level to keep things simple.

src/windows/wslc/core/Reporter.cpp
dkbennett: ConsoleWidth is still unused currently, but since the OutputChannel has the handle to the writer we need this level of of indirection so callers of reporter can get that information for output formatting (namely the table output). Tableoutput will absolutely use this.

src/windows/wslc/core/OutputWriter.cpp
dkbennett: I removed the flush entirely. It potentially has value on stdout, as it is buffered if redirected (stderr is typically not). To keep it simple and not incur a potentially significant kernel call overhead every write, I removed it. We can revisit if we see things out of order on redirection but seeing as how no other parts of this codebase do that it's probably fine.

localization/strings/en-US/Resources.resw
dkbennett: I removed this and used Localization::MessageErrorCode for ERROR_CANCELLED similar to the other errors nearby.

src/windows/wslc/core/Reporter.h
```diff
@@ -0,0 +1,165 @@
...
+        if (!vtEnabled)
+        {
+            return {};
+        }
+        if (!colorEnabled && sequence.IsColor())
+        {
+            return {};
+        }
+        return sequence.Get();
+    }
```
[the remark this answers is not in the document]
dkbennett: The temporaries passed to Output()/Write() are bound to the function parameter pack as rvalue references, so their lifetime extends to the end of the caller's full expression (the semicolon). The wstring_view in the tuple borrows from those still-alive temporaries, and vformat consumes them before the caller's expression ends. No dangling occurs here. It's not working by luck; it's guaranteed by the standard.

To ensure this is correct, I ran a stress test which uses ConstructedSequence cursor mentioned 10,000 times. 10,000 iterations, 0 failures. Then I ran it with random sequences so the data changed, and did it 100,000 times. All passed. If that's luck I need to book a trip to Vegas.

src/windows/wslc/core/Reporter.cpp
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    switch (level)
+    {
+    case Level::Warning:
+        return Format::Fg::BrightYellow.Get();
+    case Level::Error:
+        return Format::Fg::BrightRed.Get();
+    default:
+        return Format::Default.Get();
```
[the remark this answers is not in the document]
dkbennett: This is intentional so previous color sequences that may be in effect are cleared before the write and any sequences used in the output are cleared at the end of it.  However, it would be an efficiency and save an allocation to not write the resets for stdout and if sequences are added by a caller to stdout they would need to put their own reset in, and that doesn't seem unreasonable. If previous writes are handled correctly there should not be color bleed to deal with, as they should have been wrapped with a reset, so this should be safe to change.

test/windows/wslc/WSLCCLIReporterUnitTests.cpp
```diff
@@ -0,0 +1,306 @@
...
+        VERIFY_IS_FALSE(cap.reporter.GetConsoleWidth(Reporter::Level::Info).has_value());
+        VERIFY_IS_FALSE(cap.reporter.GetConsoleWidth(Reporter::Level::Warning).has_value());
+        VERIFY_IS_FALSE(cap.reporter.GetConsoleWidth(Reporter::Level::Error).has_value());
+    }
+    TEST_METHOD(Reporter_VTStatePreservedAfterInteractiveConsoleStateUnwind)
+    {
+        HANDLE stdOut = GetStdHandle(STD_OUTPUT_HANDLE);
+        DWORD mode = 0;
+        VERIFY_WIN32_BOOL_SUCCEEDED(GetConsoleMode(stdOut, &mode));
+
```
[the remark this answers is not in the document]
dkbennett: This was in there, and was intentionally removed so we fail loudly if this behavior regresses and it isn't hidden inside a skipped test. te.exe is a console process so it should be valid and passed when running locally. I'm going to wait for the CI to complete and if it succeeds I will leave it as-is so we know if a change breaks our test environment assumptions or this behavior regresses.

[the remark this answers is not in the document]
dkbennett: The CI test passed, so not putting in a skip. If this breaks it means our test environment assumptions changed, or the functionality regressed, both of which is good to cause a test failure on.

test/windows/wslc/WSLCCLIReporterUnitTests.cpp
```diff
@@ -0,0 +1,306 @@
...
+        VERIFY_ARE_EQUAL(modeBeforeConsoleState, modeAfterUnwind);
+        VERIFY_IS_TRUE(reporter.IsVTEnabled(Reporter::Level::Output));
+    }
+
+    TEST_METHOD(Reporter_VTStatePreservedAfterNonInteractiveConsoleStateUnwind)
+    {
+        HANDLE stdOut = GetStdHandle(STD_OUTPUT_HANDLE);
+        DWORD mode = 0;
+        VERIFY_WIN32_BOOL_SUCCEEDED(GetConsoleMode(stdOut, &mode));
+
```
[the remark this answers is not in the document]
dkbennett: As above, CI failed so removing test.

test/windows/wslc/WSLCCLIReporterUnitTests.cpp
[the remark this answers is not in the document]
dkbennett: Wait no the test did fail, going to remove them then, there's no reason to have it if its just going to be skipped.

src/windows/wslc/core/Reporter.h
```diff
@@ -0,0 +1,165 @@
...
+    Reporter(FILE* outFile, bool outVtEnabled, FILE* errFile, bool errVtEnabled);
+
+    NON_COPYABLE(Reporter);
+    NON_MOVABLE(Reporter);
+
+    ~Reporter() = default;
+
+    // std::format-style write API.
+    template <typename... Args>
+    void Write(Level level, std::wformat_string<Args...> fmt, Args&&... args)
```
OneBlue, line 79: nit: I recommend switching those to something like: 

```
void Write(Level level, const std::wstring_view& content, std::optional<common::vt::Sequence>& sequence = {}); 
``` 

That way callers can do: 

```
Write(std::format("..."));
```

And optionally pass a sequence to change the format if they'd like to (with the expectation that the format will be reset automatically once Write() returns) 

This would have the benefit of simplifying this implementation, and would prevent issues where there happens to be a "{}" in the caller's string, but the caller doesn't actually want it to be formatted

test/windows/wslc/WSLCCLITestHelpers.h
```diff
@@ -62,6 +72,104 @@ inline void LogComment(const std::wstring& message)
...
+// RAII pipe pair for capturing FILE* output in tests.
+// file() is passed to OutputChannel/Reporter; captured() drains the read end after flush.
+struct CapturePipe
+{
+    CapturePipe()
+    {
+        // ReadPipeOverlapped=true so PartialHandleRead's InterruptableRead can be
+        // interrupted by m_exitEvent during teardown if fclose hasn't run yet.
+        auto [r, w] = wsl::windows::common::wslutil::OpenAnonymousPipe(0, true, false);
+        const int writeFd = _open_osfhandle(reinterpret_cast<intptr_t>(w.get()), _O_WRONLY | _O_TEXT);
```
OneBlue: I recommend using `FileFromHandle()` for this

OneBlue: (And storing as a wil::unique_file to make cleanup easier)

OneBlue (commented):
Change looks good, couple minor comments

test/windows/wslc/WSLCCLITestHelpers.h
dkbennett: Updated the test helper to use these, and that did make the cleanup easier, thanks!

src/windows/wslc/core/Reporter.h
dkbennett: The Write format string is not just for sequences; it delegates to std::format directly (with compile-time format string validation), and the typed Sequence args let us selectively strip VT sequences based on terminal capabilities. I just added a test that verifies this and protects against any regression. See Reporter_Write_MixesSequencesWithStandardFormatArgs in the unit tests — it verifies output is identical to std::format across all three possible combinations of (No VT, VT but no color, and Color and VT). This test compares the result with std::format with the same format string and non-sequence args to establish equivalency.

So it can be used as if it were std::format, which saves the caller the trouble of needing to write std::format within the call itself and it Just Works as long as they are using the proper Sequence or ConstructedSequence for the VT codes.

OneBlue: Yeah that was my understanding. What I meant is that we can have the same behavior with an optional sequence argument and that it would be preferable for format to be reset automatically at the end of a print (if the caller specified one) so callers don't need to manually reset them. 

That would also allow us to remove quite a bit of SFINAE code here, and move the implementations to the .cpp file. 

Regarding std::format(), I personally think it's better for calling to explicitly call std::format(), that way it's very clear for the reader that this is a format string, not a raw string. 

This is not blocking so feel free to merge this and we can address this later

OneBlue (approved):
Approving to unblock, one outstanding comment on how we can simplify / explicit the std::format logic, but this is not blocking

Issue #40790, opened 2026-06-12
`virtiofs=true`: automount aborts entirely when any drive cannot be shared


codydunne:
### Windows Version

Microsoft Windows [Version 10.0.26200.8655]

### WSL Version

2.7.8.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.18.33.1-1

### Distro Version

Ubuntu 26.0.4

### Other Software

**Google Drive for Desktop**: Host has fixed drives `C:`, `D:`, `F:` plus two Google Drive letters `G:`, `H:`. I belive these drives are backed by a network-redirector/filter driver and cannot be exported by the virtiofs server. Their presence seems to trigger the cascade below, but I haven't tested without it.

### Repro Steps

_Likely the same as #40773 but I've emailed logs in case this helps._

1. `%USERPROFILE%\.wslconfig`:

   ```ini
   [wsl2]
   virtiofs=true
   ```

3. `wsl --shutdown`
4. `wsl --install Ubuntu --name ubuntu-clean` (**path errors on boot, described below**)
5. _`.\collect-wsl-logs.ps1 -LogProfile storage`_
6. `wsl -d ubuntu-clean` (**path errors**)
7. _Stop `collect-wsl-logs.ps1` and email archive_
8. Observe `ls /mnt`, `dmesg`, then attempt the documented fallbacks (`mount -t drvfs`, `/etc/fstab` `drvfs` entries).

### Expected Behavior

- Shareable drives (`C:`, `D:`, `F:`) are mounted under `/mnt/<letter>` via virtiofs.
- Unshareable drives (`G:`, `H:`) are skipped gracefully.
- `mount -t drvfs C: /mnt/c` works with virtiofs even if automount was disabled.
- `/etc/fstab` `drvfs` entries work.
- Interop path translation (`%PATH%` appending, launch cwd, `wslpath`) works.

### Actual Behavior

**1. Symptom: Path translation errors**

Every session launch fails to translate the launch working directory and the Windows `%PATH%` entries:

```pwsh
c:\Users\USERNAME> wsl -d ubuntu-clean
wsl: Failed to translate 'c:\Users\USERNAME'
wsl: Failed to translate 'C:\Program Files\PowerShell\7'
...fails each other Windows %PATH% entry
```

This is because...

**2. Automount aborts.** `/mnt` contains only `wsl` and `wslg`, not the expected `c d f g h wsl wslg`

```bash
[...]

Issue #40773, opened 2026-06-11
virtiofs=true: automount of ALL fixed drives is silently skipped when AddVirtioFsShare fails for a single drive


moto-pu:
### Windows Version

Microsoft Windows [Version 10.0.26200.8655]

### WSL Version

2.7.8.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.18.33.1-1

### Distro Version

Ubuntu 24.04

### Other Software

Docker Desktop 4.77.0 (its `docker-desktop` distro is also affected, see below)

### Repro Steps

Machine has two fixed NTFS drives: `C:` (NVMe) and `P:` (SATA). Both automount fine via 9p when virtiofs is disabled.

1. Add `virtiofs=true` to the `[wsl2]` section of `.wslconfig`
2. `wsl --shutdown`, start the distro
3. `mount | grep /mnt`

### Expected Behavior

`/mnt/c` and `/mnt/p` are automounted (virtiofs where possible, 9p fallback otherwise), as documented for the virtiofs opt-in.

### Actual Behavior

**No fixed drive is automounted at all** — `/mnt/c` and `/mnt/p` simply don't exist as mounts. There is no warning on the terminal, nothing in `dmesg` and nothing in the boot log. Interop path translation also degrades (`wsl: Failed to translate 'C:\...'` warnings for every PATH entry when launching Windows executables from another distro).

The virtiofs device for `C:` **is** created and works — the tag is visible at kernel boot and a manual mount succeeds:

```
[    0.253334] virtiofs virtio1: discovered new tag: f8407c95-93b2-41f6-ac92-7b8deb88fd19
$ sudo mount -t virtiofs f8407c95-... /mnt/c   # works fine
```

Meanwhile the share for `P:` is never created. Requesting it manually shows the failure that I believe poisons the whole automount:

```
$ sudo mount -t drvfs 'P:\' /mnt/p
WSL (...) WARNING: Add virtiofs share for P:\ failed 22, falling back to Plan9
```

(`P:` is a plain healthy NTFS volume on a SATA SSD, `DriveType: Fixed`; the 9p fallback works fine, so per-drive fallback clearly exists for the runtime path — just not for the boot path.)

### Analysis (from the 2.7.8 sources)

[...]

PR #40806, opened 2026-06-15
Fix virtiofs share error handling

chemwolf6922:

Handle individual virtiofs share failures. So that one drive's failure does not cause all shares to fail.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/40790 https://github.com/microsoft/WSL/issues/40773

## Validation Steps Performed

Cannot repro the issue as the confirmed trigger (google drive) cannot be installed on company machine and within company network.


src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -895,7 +895,16 @@ void WslCoreVm::AddDrvFsShare(_In_ bool Admin, _In_ HANDLE UserToken)
...
+            try
+            {
+                AddVirtioFsShare(Admin, fixedDrivePath, TEXT(LX_INIT_DEFAULT_PLAN9_MOUNT_OPTIONS), UserToken);
+            }
+            catch (...)
+            {
+                const auto result = wil::ResultFromCaughtException();
+                WSL_LOG(
+                    "AddVirtioFsShareError", TraceLoggingValue(fixedDrivePath, "DrivePath"), TraceLoggingValue(result, "result"));
+            }
```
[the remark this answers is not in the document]
chemwolf6922: ?

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -895,7 +895,16 @@ void WslCoreVm::AddDrvFsShare(_In_ bool Admin, _In_ HANDLE UserToken)
             ULONG index;
             WI_VERIFY(_BitScanForward(&index, fixedDrives) != FALSE);
             const wchar_t fixedDrivePath[] = {gsl::narrow_cast<wchar_t>(L'A' + index), L':', L'\\', L'\0'};
-            AddVirtioFsShare(Admin, fixedDrivePath, TEXT(LX_INIT_DEFAULT_PLAN9_MOUNT_OPTIONS), UserToken);
+            try
+            {
+                AddVirtioFsShare(Admin, fixedDrivePath, TEXT(LX_INIT_DEFAULT_PLAN9_MOUNT_OPTIONS), UserToken);
```
benhillis, line 900: In what cases does this fail?

codydunne: I believe in my case #40790 it was the two drives that Google Drive created.

benhillis: Ok makes sense. @chemwolf6922 - this is fine but ideally, we'd debug the root cause of the failure instead of swallowing the error.

chemwolf6922: @benhillis . I cannot access emailed logs from codydunne. But the log from https://github.com/microsoft/WSL/issues/40773 shows `std::exception: weakly_canonical: Access is denied.: "G:\"`. That's the default drive letter for Google drive. I'll try to confirm with the OP if that's the case.

chemwolf6922: Confirmed, it's google drive as well.

benhillis: Ok - we own the virtiofs (wsldevicehost) layer as well, is this fixable there?

chemwolf6922: @benhillis The error is from here:
https://github.com/microsoft/WSL/blob/b33dbf0d8efd0d1bd9df5aab2c7d704a7df0985d/src/windows/service/exe/WslCoreVm.cpp#L2187
So, no, I don't think this is related to wsldevicehost.

benhillis (approved):

PR #40810, opened 2026-06-15
Use adaptive units for image and container prune reclaimed-space output

AmelBawa-msft:

Image and container `prune` hardcoded the reclaimed-space output to MB (e.g. `2048.00 MB`, or `0.00 MB` for sub-MB amounts). Volume `prune` already used the shared `FormatBytes` helper, which picks B/KB/MB/GB based on magnitude and matches `docker system prune`.

This updates image and container prune to use `FormatBytes` too, so all three commands format reclaimed space consistently and readably. The en-US string IDs for the reclaimed-space messages have been renamed so the localization pipeline regenerates fresh translations with the correct placeholder format.

## PR Checklist

- `src/windows/wslc/tasks/ImageTasks.cpp` and `ContainerTasks.cpp`: switched reclaimed-space formatting from hardcoded MB to `wsl::shared::string::FormatBytes(...)`.
- `localization/strings/en-US/Resources.resw`: renamed `WSLCCLI_ContainerPruneSpaceReclaimed` → `WSLCCLI_ContainerPruneSpaceReclaimedBytes` and `WSLCCLI_ImagePruneSpaceReclaimed` → `WSLCCLI_ImagePruneSpaceReclaimedBytes`, with the value placeholder updated from `{:.2f} MB` to `{}`. Renaming the IDs (rather than editing locale files directly) ensures the localization pipeline generates fresh translations with the correct `{}` placeholder, avoiding `std::format_error` or doubled units (e.g. "KB MB") in non-English locales at runtime.
- `test/windows/wslc/e2e/WSLCE2EContainerPruneTests.cpp`: adjusted container prune E2E assertion for the new reclaimed-space output format.

## Validation Steps Performed

- Localization validation script (`tools/devops/validate-localization.py`) passes with no errors.


localization/strings/en-US/Resources.resw
```diff
@@ -2456,8 +2456,8 @@ For privacy information about this product please visit https://aka.ms/privacy.<
     <value>Removes all stopped containers.</value>
   </data>
   <data name="WSLCCLI_ContainerPruneSpaceReclaimed" xml:space="preserve">
-    <value>Total reclaimed space: {:.2f} MB</value>
-    <comment>{FixedPlaceholder="{:.2f}"}Command line arguments, file names and string inserts should not be translated</comment>
+    <value>Total reclaimed space: {}</value>
+    <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
```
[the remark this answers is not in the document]
AmelBawa-msft: @copilot update all the resw files to align with this change, don't make other unrelated changes or format the files.

localization/strings/cs-CZ/Resources.resw
```diff
@@ -2456,8 +2456,8 @@ Informace o ochraně osobních údajů k tomuto produktu najdete na stránce htt
     <value>Odebere všechny zastavené kontejnery.</value>
   </data>
   <data name="WSLCCLI_ContainerPruneSpaceReclaimed" xml:space="preserve">
-    <value>Celkové uvolněné místo: {:.2f} MB</value>
```
OneBlue, line 2459: These files should not be edited in PR's. if we need to make a change to the format, the easiest solution would be just rename in the en-us file

AmelBawa-msft: @copilot address this comment by reverting the changes in the resw files, and instead rename the string id so the loc resw files get regenerated by the pipeline. Make the appropriate changes in code. Don't change other unrelated code. Also ensure the code is formatted with clang format.

OneBlue (approved):

PR #40811, opened 2026-06-15
Fix wslc image inspect missing RootFS, ExposedPorts, Volumes, and StopSignal

beena352:
wslc image inspect was missing several fields compared to docker image inspect: RootFS.Layers, Config.ExposedPorts, Config.Volumes, and Config.StopSignal. The Docker API response already contained them, but wslc_schema::InspectImage and ImageConfig didn't declare these fields, so ConvertInspectImage silently dropped them during the docker→wslc conversion.

## PR Checklist

- [ ] **Closes:** Link to issue #62098442

Schema changes (wslc_schema.h):

- Added EmptyObject helper struct 

- Added ExposedPorts, StopSignal, Volumes to ImageConfig

- Added ImageRootFS struct with Type and Layers

- Added RootFS field to InspectImage

- Converter changes (WSLCSession.cpp): ConvertInspectImage now maps ExposedPorts, Volumes (with has_value() guard to preserve nullopt-vs-populated semantics), StopSignal, and RootFS from the Docker response.

### DIFF
--- a/src/windows/inc/wslc_schema.h
+++ b/src/windows/inc/wslc_schema.h
@@ -18,6 +18,22 @@ Abstract:
 
 namespace wsl::windows::common::wslc_schema {
 
+struct EmptyObject
+{
+};
+
+inline void to_json(nlohmann::json& j, const EmptyObject& obj)
+{
+    UNREFERENCED_PARAMETER(obj);
+    j = nlohmann::json::object();
+}
+
+inline void from_json(const nlohmann::json& j, EmptyObject& obj)
+{
+    UNREFERENCED_PARAMETER(j);
+    UNREFERENCED_PARAMETER(obj);
+}
+
 struct InspectPortBinding
 {
     // WSLC always binds to localhost. Included for Docker API compatibility.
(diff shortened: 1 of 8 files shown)


yao-msft:
I'm not sure if things have changed recently, but I barely remember we left some of the data out intentionally since they are linux internal implementation related, i.e. Rootfs

beena352:
> I'm not sure if things have changed recently, but I barely remember we left some of the data out intentionally since they are linux internal implementation related, i.e. Rootfs

The bug (https://microsoft.visualstudio.com/OS/_workitems/edit/62098442) explicitly calls out missing "Layers" and "ExposedPorts" fields and compares the output directly against docker image inspect, which includes the full RootFS section. I can remove the RootFS section if we don't want to expose it.

dkbennett (dismissed):

dkbennett:
Approved assuming you want to doublecheck if these are good to add but implementation LGTM.

src/windows/inc/wslc_schema.h
```diff
@@ -18,6 +18,22 @@ Module Name:
 
 namespace wsl::windows::common::wslc_schema {
 
+struct EmptyObject
```
OneBlue: Instead of duplicating this from `docker_schema.h`, I think we should factor it in `JsonUtils.h`

test/windows/wslc/e2e/WSLCE2EHelpers.h
```diff
@@ -128,6 +128,24 @@ void EnsureVolumeDoesNotExist(const std::wstring& volumeName);
...
 
 void WriteTestFile(const std::filesystem::path& filePath, const std::vector<std::string>& envVariableLines);
+void WriteTestFileContent(const std::filesystem::path& filePath, const std::string& content);
+
+// Sets up a clean test directory and returns a scope_exit to remove it.
+inline auto SetupTestDirectory(const std::filesystem::path& directory)
+{
+    std::error_code ec;
+    std::filesystem::remove_all(directory, ec);
+    THROW_HR_IF_MSG(E_FAIL, ec.value() != 0 && std::filesystem::exists(directory), "%hs", ec.message().c_str());
```
OneBlue, line 138: nit: We could simplify by just calling the overload that doesn't take an `std::error_code` here (it will throw on error)

OneBlue: Well looks like this is moved from an existing implementation. Feel free to ignore / do in a separate PR

OneBlue (dismissed):
LGTM, minor comments

OneBlue (approved):

PR #40812, opened 2026-06-15
Implement wslc system session run and create the default session if it doesn't exist

OneBlue:

This change implements `wslc system session run <args>` and updates the `wslc system session` commands to create the default session on demand if it doesn't exist.

Note that running `wslc system session run` in an interactive terminal will currently hang. I'll fix this in a followup change

## PR Checklist


src/linux/init/WSLCInit.cpp
```diff
@@ -765,7 +765,7 @@ void HandleMessageImpl(
     auto EnvironmentArray = wsl::shared::string::ArrayFromSpan(Buffer, Message.EnvironmentIndex);
     auto EnvironmentPointers = wsl::shared::string::StringPointersFromArray(EnvironmentArray, true);
 
-    execve(Executable, (char* const*)(ArgumentPointers.data()), (char* const*)(EnvironmentPointers.data()));
+    execvpe(Executable, (char* const*)(ArgumentPointers.data()), (char* const*)(EnvironmentPointers.data()));
```
OneBlue, line 768: Switching this for convenience, since execvpe will use $PATH if the executable isn't an absolute path

src/windows/wslc/services/SessionService.cpp
```diff
@@ -180,6 +187,25 @@ Session SessionService::OpenSession(const std::wstring& displayName)
...
+int SessionService::Run(const std::wstring& sessionName, const std::vector<std::string>& arguments)
+{
+    WI_ASSERT(!arguments.empty());
+
+    auto session = OpenOrCreateSession(sessionName);
+
+    // Pass a default $PATH environment for convenience.
+    const std::vector<std::string> environment{"PATH=/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/sbin"};
+    wsl::windows::common::WSLCProcessLauncher launcher{arguments.front(), arguments, environment, WSLCProcessFlagsStdin};
+
```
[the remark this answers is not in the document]
OneBlue: That's true, but will fix in a followup change

src/windows/wslc/commands/SessionRunCommand.cpp
```diff
@@ -0,0 +1,48 @@
...
+using namespace wsl::shared;
+
+namespace wsl::windows::wslc {
+// Session Run Command
+std::vector<Argument> SessionRunCommand::GetArguments() const
+{
+    return {
+        Argument::Create(ArgType::Command, true),
+        Argument::Create(ArgType::ForwardArgs, std::nullopt, std::nullopt, Localization::WSLCCLI_SessionRunForwardArgsDescription()),
+        Argument::Create(ArgType::Session),
```
dkbennett, line 30: Global args for session won't break this but it'll be strange with both being available, something to think about / fix depending on which PR goes in first.

test/windows/wslc/e2e/WSLCE2EGlobalTests.cpp
```diff
@@ -492,6 +492,19 @@ class WSLCE2EGlobalTests
         }
     }
 
+    WSLC_TEST_METHOD(WSLCE2E_Session_Run)
```
dkbennett, line 495: nit: Should also test the session-is-specified case.

src/windows/wslc/services/SessionService.cpp
```diff
@@ -24,36 +24,43 @@ using namespace wsl::shared;
...
+        if (sessionName.empty())
         {
-            wslutil::PrintMessage(
-                sessionName.empty() ? Localization::MessageWslcDefaultSessionNotFound()
-                                    : Localization::MessageWslcSessionNotFound(sessionName.c_str()),
-                stderr);
-            return 1;
+            // Default session: open it if it exists, otherwise create it.
+            auto warningCallback = Microsoft::WRL::Make<WarningCallback>();
+            THROW_IF_FAILED(manager->CreateSession(nullptr, WSLCSessionFlagsNone, warningCallback.Get(), &session));
```
dkbennett, line 40: For non-session commands we have a common "CreateSession" task which will open the session if it is specified or fallback on creating/opening the default session.  For the session commands which started with a few one-offs relating to sessions we did not do that, mainly because the argtype is different when it is a positional argument vs a value argument. But as we add more session commands the open code is getting duplicated across several commands now.

There's opportunity to unify this code and unify the session open error messages, which I think is beyond the scope of this PR, but I will take a work item to reduce the duplication in the session open calls so they are all consistent, regardless of whether the session was created with a positional id or a global parameter.

test/windows/wslc/CommandLineTestCases.h
```diff
@@ -34,6 +34,15 @@ COMMAND_LINE_TEST_CASE(L"system session list --notanarg", L"list", false)
 COMMAND_LINE_TEST_CASE(L"system session list extraarg", L"list", false)
 COMMAND_LINE_TEST_CASE(L"system session shell session1", L"shell", true)
 COMMAND_LINE_TEST_CASE(L"system session shell", L"shell", true)
+COMMAND_LINE_TEST_CASE(L"system session run ls", L"run", true)
+COMMAND_LINE_TEST_CASE(L"system session run echo foo", L"run", true) // Command with trailing arguments
+COMMAND_LINE_TEST_CASE(L"system session run ls -la", L"run", true)   // Flags after the command are forwarded
+COMMAND_LINE_TEST_CASE(L"system session run --session session1 ls", L"run", true)
```
dkbennett, line 40: This is another area that might need changing after global session parameter exists. Its probably best this PR goes in before global options and I can update global options to incorporate it.

dkbennett (dismissed):
Looks good. I think there's some refactoring we can do with all of the session opening code but I dont think that is for this PR to fix. Best solution there is to get this in and then I can do a separate PR to unify the open code across all session opening/creating.

dkbennett (approved):

src/windows/wslc/services/SessionService.cpp
```diff
@@ -180,6 +187,25 @@ Session SessionService::OpenSession(const std::wstring& displayName)
...
 
+int SessionService::Run(const std::wstring& sessionName, const std::vector<std::string>& arguments)
+{
+    WI_ASSERT(!arguments.empty());
+
+    auto session = OpenOrCreateSession(sessionName);
+
+    // Pass a default $PATH environment for convenience.
+    const std::vector<std::string> environment{"PATH=/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/sbin"};
+    wsl::windows::common::WSLCProcessLauncher launcher{arguments.front(), arguments, environment, WSLCProcessFlagsStdin};
```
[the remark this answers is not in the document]
OneBlue: True, will do in a followup

PR #40821, opened 2026-06-16
fix(wslc): early boot dmesg collection on arm64

damanm24:
The wrong kernel command line arguments are passed when the `EarlyBootDmesg` feature is enabled in WSLC. This causes a hang on VM boot.


OneBlue (approved):
LGTM. Thank you for doing this !

PR #40822, opened 2026-06-16
Fix use-after-free in WSLCContainerImpl exec-process teardown

yao-msft:
Change m_processes to use weak pointers instead of raw pointers. Use lock() to pin on use.

## PR Checklist

## Validation Steps Performed
Existing tests

### DIFF
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -2090,7 +2090,6 @@ void WSLCContainerImpl::UnmapPorts()
 __requires_exclusive_lock_held(m_lock) void WSLCContainerImpl::ReleaseProcesses()
 {
     // Snapshot under the lock, then notify outside it, pinning each control via lock() first
-    // (avoids the use after free, and avoids calling back under m_processesLock).
     decltype(m_processes) processes;
     {
         std::lock_guard processesLock{m_processesLock};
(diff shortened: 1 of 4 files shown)


src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -2094,16 +2089,23 @@ void WSLCContainerImpl::UnmapPorts()
 
 __requires_exclusive_lock_held(m_lock) void WSLCContainerImpl::ReleaseProcesses()
 {
-    std::lock_guard processesLock{m_processesLock};
+    // Snapshot under the lock, then notify outside it, pinning each control via lock() first
+    // (avoids the use after free, and avoids calling back under m_processesLock).
```
OneBlue: nit: I'd recommend removing the `avoids the use after free` since once merged, that comment will refer to an issue that won't exist anymore

src/windows/wslcsession/WSLCProcessControl.cpp
```diff
@@ -118,11 +118,7 @@ DockerExecProcessControl::DockerExecProcessControl(
 
 DockerExecProcessControl::~DockerExecProcessControl()
```
OneBlue: nit: This destructor can be removed since it's not doing anything

OneBlue (dismissed):
LGTM ! It's wild that we never hit this during testing

OneBlue (approved):

PR #40823, opened 2026-06-16
 Fix race condition causing wslc build output to be truncated

OneBlue:

This change solves two issues: 

- One of wslc build's handle is incorrectly marked as 'CancelOnCompleted', which causes the IO loop to exit early and drop output
- The session, process and cancel events are not marked as 'NeedNotComplete', which causes them to block the IO, and requires 'CancelOnCompleted' in various places

## PR Checklist

### DIFF
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -560,13 +560,11 @@ ServiceRunningProcess WSLCSession::StartProcess(
 
     auto process = launcher.Launch(*m_virtualMachine);
 
-    m_ioRelay.AddHandle(
-        std::make_unique<windows::common::io::LineBasedReadHandle>(
-            process.GetStdHandle(1), [this, LogSource](const auto& data) { OnProcessLog(data, LogSource); }, false));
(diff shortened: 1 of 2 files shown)


yao-msft (approved):

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1258,9 +1253,7 @@ void WSLCSession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request,
     io.AddHandle(std::make_unique<io::RelayHandle<io::ReadHandle>>(
         common::io::HandleWrapper{userHandle.Get(), std::move(onInputComplete)}, common::io::HandleWrapper{Request.stream.native_handle()}));
```
johnstep, line 1254: Copilot recommends marking the input relay `NeedNotComplete` so the response handle is the sole completion signal. If docker fails before draining the upload, the relay may otherwise (1) block forever reading the client pipe or (2) throw on a write to the reset socket, masking the real error. Exiting on the response avoids both.

OneBlue: Good catch ! Fixed by: https://github.com/microsoft/WSL/pull/40839

johnstep (approved):

PR #40826, opened 2026-06-16
Support RFC3339 timestamps for wslc container logs --since/--until

ptrivedi:
## Summary

Adds RFC3339 date-time parsing as an alternative to Unix epoch seconds for the \--since\ and \--until\ arguments of \wsl container logs\. This aligns with the timestamp formats that other container CLIs have.

## Changes

- **ArgumentValidation.cpp/h**: Added \TryParseRfc3339()\ parser and \ValidateTimestamp()\ / \GetTimestampFromString()\ public functions. The parser handles:
  - UTC timestamps with \Z\ suffix (e.g. \2024-01-15T10:30:00Z\)
  - Timezone offsets in \+HH:MM\ format (note: \+HHMM\ without colon is not supported, consistent with RFC3339 and \std::chrono::parse\)
  - Fractional seconds (ignored for epoch conversion)
  - Validation of date/time component ranges
  - Rejection of pre-1970 dates, trailing garbage, and malformed strings
- **ContainerTasks.cpp**: Updated \ViewContainerLogs()\ to use \GetTimestampFromString()\ instead of \GetIntegerFromString<ULONGLONG>()\
- **Resources.resw**: Updated \--since\/\--until\ help text and added \WSLCCLI_InvalidTimestampArgumentError\ localization string
- **CommandLineTestCases.h**: Added 9 unit test cases covering valid RFC3339 formats, invalid dates, trailing garbage, and pre-epoch rejection
- **WSLCE2EContainerLogsTests.cpp**: Added e2e test verifying RFC3339 timestamps work with live container logs

## Testing

- All \CommandLineParsing_AllCases\ tests pass (1/1, 0 failures)
- All new timestamp test cases validated: valid formats accepted, invalid formats rejected with descriptive error


localization/strings/cs-CZ/Resources.resw
```diff
@@ -2897,13 +2827,13 @@ Při prvním spuštění vytvoří soubor se všemi nastaveními okomentovanými
     <value>Počet řádků, které se mají zobrazit od konce protokolů</value>
   </data>
   <data name="WSLCCLI_TimestampsArgDescription" xml:space="preserve">
-    <value>Zobrazit časová razítka ve výstupu protokolu</value>
+    <value>Show timestamps in log output</value>
```
OneBlue: These shouldn't be edited as part of the pull request (they're generated by our pipeline)

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -233,201 +192,267 @@ WSLCSignal GetWSLCSignalFromString(const std::wstring& input, const std::wstring
     return static_cast<WSLCSignal>(signalValue);
 }
 
-void ValidateFormatTypeFromString(const std::vector<std::wstring>& values, const std::wstring& argName)
+// Parses an RFC3339 timestamp (e.g. "2024-01-15T10:30:00Z" or "2024-01-15T10:30:00+05:30")
+// or a plain Unix epoch integer (seconds) into a ULONGLONG epoch value.
+static std::optional<ULONGLONG> TryParseRfc3339(const std::string& input)
 {
-    for (const auto& value : values)
+    // Minimum RFC3339: "YYYY-MM-DDTHH:MM:SSZ" = 20 chars
```
OneBlue: I recommend using `std::chrono::parse` to parse the RFC3339 format instead of writing a custom parser

test/windows/wslc/e2e/WSLCE2EContainerLogsTests.cpp
```diff
@@ -57,6 +57,139 @@ class WSLCE2EContainerLogsTests
...
+        result = RunWslc(std::format(L"container logs --timestamps {}", WslcContainerName));
+        result.Verify({.Stderr = L"", .ExitCode = 0});
+
+        // Timestamps should be in RFC3339 format (e.g. "2024-01-01T00:00:00.000000000Z hello")
+        // Verify output contains a timestamp-like prefix followed by the message
+        VERIFY_IS_TRUE(result.StdoutContainsSubstring(L"hello"));
+        auto lines = result.GetStdoutLines();
+        VERIFY_IS_TRUE(!lines.empty());
+        // A timestamp line should have at least a 'T' character (ISO 8601 separator)
+        VERIFY_IS_TRUE(lines[0].find(L'T') != std::wstring::npos);
```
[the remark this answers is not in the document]
OneBlue: Agreed with this. The easiest way would be to use [this method](https://github.com/microsoft/WSL/blob/5db2759f178b0be678853cbc909d5f945d5c9279/src/windows/wslcsession/WSLCContainer.cpp#L381) to parse the timestamp and check that it's valid

ptrivedi:
Addressed all review feedback in 911f1479:

**@OneBlue — std::chrono::parse**: Replaced the custom RFC3339 parser with \std::chrono::parse\ using \%FT%T%Ez\, matching the pattern from \WSLCContainer.cpp::ParseDockerTimestamp\. Added \year_month_day\ pre-validation to reject impossible dates (e.g. Feb 31) that \std::chrono::parse\ normalizes. Fractional seconds are stripped before parsing since we only need second-level precision.

**@OneBlue — Non-en-US locale files**: Reverted all non-en-US locale file changes. Only \n-US/Resources.resw\ is modified now.

**@Copilot — Invalid day-of-month (Feb 31)**: Fixed via \year_month_day::ok()\ pre-validation before parsing.

**@Copilot — Dot without digits (\.Z\)**: Fixed — fractional second stripping now requires at least one digit after the dot.

**@Copilot — Trailing garbage (\Zextra\)**: Fixed — \Z\ suffix is normalized to \+00:00\ before parsing with \%Ez\ (avoids \%Z\ greedy matching), and \stream.peek()\ rejects any trailing characters.

**@Copilot — E2E timestamp test weakness**: Strengthened to validate RFC3339 structure (YYYY-MM-DDTHH:MM:SS) at the start of the line instead of just checking for 'T'.

**@Copilot — \--until 0\ treated as unset**: This is a pre-existing design choice in the downstream service layer where \

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -233,6 +235,113 @@ WSLCSignal GetWSLCSignalFromString(const std::wstring& input, const std::wstring
...
+    // Normalize trailing 'Z'/'z' to '+00:00' so %Ez can parse it uniformly.
+    // This avoids %Z greedily consuming trailing characters as a timezone abbreviation.
+    if (!normalized.empty() && (normalized.back() == 'Z' || normalized.back() == 'z'))
+    {
+        normalized.pop_back();
+        normalized += "+00:00";
+    }
+
+    // Validate the date components from the input before parsing the full timestamp.
+    // std::chrono::parse normalizes invalid dates (e.g. Feb 31 → March) instead of rejecting them,
```
OneBlue: This surprised me so I tried it out, and I don't think this is true. Here's a test for that case and it fails as expected: https://godbolt.org/z/ddo5h1zn1

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -233,6 +235,113 @@ WSLCSignal GetWSLCSignalFromString(const std::wstring& input, const std::wstring
     return static_cast<WSLCSignal>(signalValue);
 }
 
+// Parses an RFC3339 timestamp (e.g. "2024-01-15T10:30:00Z" or "2024-01-15T10:30:00+05:30")
+// into a ULONGLONG Unix epoch value using std::chrono::parse.
+static std::optional<ULONGLONG> TryParseRfc3339(const std::string& input)
+{
+    // Strip fractional seconds (e.g. ".123456789") before parsing, since we only need
```
OneBlue: Could we parse into `std::chrono::sys_time<std::chrono::milliseconds>` and drop the milliseconds ? That should be easier than trying to parse ourselves

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -233,6 +235,113 @@ WSLCSignal GetWSLCSignalFromString(const std::wstring& input, const std::wstring
...
+        while (endOfFrac < normalized.size() && normalized[endOfFrac] >= '0' && normalized[endOfFrac] <= '9')
+        {
+            ++endOfFrac;
+        }
+
+        normalized.erase(dotPos, endOfFrac - dotPos);
+    }
+
+    // Normalize trailing 'Z'/'z' to '+00:00' so %Ez can parse it uniformly.
+    // This avoids %Z greedily consuming trailing characters as a timezone abbreviation.
```
OneBlue: Wouldn't this be OK though ? If the timezone turned out to be invalid, std::chrono::parse would fail right ?

dkbennett (commented):
I would recommend a few unit tests to really hammer the TryParseRfc3339 by itself added to WSLCCLIArgumentUnitTests.cpp. You could have a huge number of sample strings and formats valid or invalid to verify all manner of timestamp parsing. The existing command line tests added pretty much do this in the full command line which is also good so I wouldn't say this is blocking, but I think this logic is complex enough that we may want some extra coverage on it and the unit tests would execute more quickly than the full command line pipe.

dkbennett (approved):
Should fix up the Copilot comments but otherwise lgtm.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -233,6 +235,77 @@ WSLCSignal GetWSLCSignalFromString(const std::wstring& input, const std::wstring
     return static_cast<WSLCSignal>(signalValue);
 }
 
+// Parses an RFC3339 timestamp (e.g. "2024-01-15T10:30:00Z" or "2024-01-15T10:30:00+05:30")
+// into a ULONGLONG Unix epoch seconds value using std::chrono::parse.
+// Note: +HHMM (no colon) offsets are not supported; use +HH:MM format.
+static std::optional<ULONGLONG> TryParseRfc3339(const std::string& input)
```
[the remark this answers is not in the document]
ptrivedi: The PR description was a bit inaccurate but the implementation and tests are correct — +HHMM is intentionally rejected, matching RFC3339's canonical format. Will update PR description for posterity purposes. No code change needed for this.

src/windows/wslc/arguments/ArgumentValidation.cpp
ptrivedi: This is addressed in the merged code...?  it uses sys_time<nanoseconds> to handle up to 9 fractional digits, so it is better precision than millieconds

src/windows/wslc/arguments/ArgumentValidation.cpp
ptrivedi: yeah you are right, but the normalization below isn't really about validation. it's needed because %Ez won't parse the literal Z character at all... it only accepts numeric offsets. if this normalization isn't done, any timestamp that ends in Z would fail the parsing

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -233,6 +235,104 @@ WSLCSignal GetWSLCSignalFromString(const std::wstring& input, const std::wstring
...
+{
+    std::string normalized = input;
+
+    // Normalize trailing 'Z'/'z' to '+00:00' so %Ez can parse it uniformly.
+    if (!normalized.empty() && (normalized.back() == 'Z' || normalized.back() == 'z'))
+    {
+        normalized.pop_back();
+        normalized += "+00:00";
+    }
```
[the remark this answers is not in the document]
ptrivedi: Modified the PR description to address this already

src/windows/wslc/arguments/ArgumentValidation.cpp
ptrivedi: You are right that std::chrono::parse does reject invalid dates on the godbolt example. I think this pre-validation was added defensively looks like in case MSVC's implementation had the same behavior. It is likely redundant, should probably remove it, can do a follow-up PR. Not harmful, but does add unnecessary complexity.

src/windows/wslc/arguments/ArgumentValidation.cpp
OneBlue: What I meant is that if instead of using sys_time, we template it on the precision we want, like for instance: `    std::chrono::sys_time<std::chono::nanoseconds>`, then we don't need to manually parse the subsecond string

src/windows/wslc/arguments/ArgumentValidation.cpp
ptrivedi: The above posted example correclty rejects the invalid dates because you are using %Ez and not catching the offset. In that case, it should be the same as the std::chrono::parse used below. Also double-checked that if it is parsed into sys_seconds with %Ez and offset is not extracted explicitly, then it should fail the invalid dates.

However, in either of these cases, if the offset were extracted or %Ez wasn't used, the invalid dates would get normalized.

ptrivedi: Since the code below uses "
    stream >> std::chrono::parse("%FT%T%Ez", utcSeconds);"

where %Ez is used, output is parsed into sys_seconds, and offset is not extracted, we can remove this pre-validation and be okay. Can do in a follow=up

src/windows/wslc/arguments/ArgumentValidation.cpp
ptrivedi: Latest code does this now. In short, the comment was addressed, just not resolved.


    // Parse into nanosecond precision so fractional seconds (e.g. ".123456789") are consumed
    // by std::chrono::parse rather than requiring manual stripping.
    std::chrono::sys_time<std::chrono::nanoseconds> utcTime;
    std::istringstream stream(normalized);
    stream >> std::chrono::parse("%FT%T%Ez", utcTime);
    if (stream.fail())
    {
        return std::nullopt;
    }

    // Reject if there are trailing characters after the parsed timestamp
    if (stream.peek() != std::istringstream::traits_type::eof())
    {
        return std::nullopt;
    }

    auto epochSeconds = std::chrono::duration_cast<std::chrono::seconds>(utcTime.time_since_epoch()).count();
    if (epochSeconds < 0)
    {
        return std::nullopt;
    }

    return static_cast<ULONGLONG>(epochSeconds);

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -233,6 +235,113 @@ WSLCSignal GetWSLCSignalFromString(const std::wstring& input, const std::wstring
...
+    if (stream.fail())
+    {
+        return std::nullopt;
+    }
+
+    // Reject if there are trailing characters after the parsed timestamp
+    if (stream.peek() != std::istringstream::traits_type::eof())
+    {
+        return std::nullopt;
+    }
```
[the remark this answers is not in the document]
ptrivedi: PR description was updated in order to resolve this.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -686,14 +686,27 @@ void ViewContainerLogs(CLIExecutionContext& context)
...
+    if (context.Args.Contains(ArgType::Since))
+    {
+        since = validation::GetTimestampFromString(context.Args.Get<ArgType::Since>());
+    }
+
+    ULONGLONG until = 0;
+    if (context.Args.Contains(ArgType::Until))
+    {
+        until = validation::GetTimestampFromString(context.Args.Get<ArgType::Until>());
+    }
```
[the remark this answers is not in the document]
ptrivedi: This models docker behavior where until=0 is treated as no upper bound, so show all logs

test/windows/wslc/e2e/WSLCE2EContainerLogsTests.cpp
```diff
@@ -73,8 +73,13 @@ class WSLCE2EContainerLogsTests
...
         VERIFY_IS_TRUE(!lines.empty());
-        // A timestamp line should have at least a 'T' character (ISO 8601 separator)
-        VERIFY_IS_TRUE(lines[0].find(L'T') != std::wstring::npos);
+        // Validate RFC3339 structure: YYYY-MM-DDTHH:MM:SS at the start of the line
+        VERIFY_IS_TRUE(lines[0].size() >= 20);
+        VERIFY_ARE_EQUAL(lines[0][4], L'-');
+        VERIFY_ARE_EQUAL(lines[0][7], L'-');
+        VERIFY_ARE_EQUAL(lines[0][10], L'T');
+        VERIFY_ARE_EQUAL(lines[0][13], L':');
+        VERIFY_ARE_EQUAL(lines[0][16], L':');
```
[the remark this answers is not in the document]
ptrivedi: Fair. But this is e2e integration test, not a timestamp parser test. ts parsing is being thoroughly tested via a unit test.

adding digit asserts and trailing Z check is reasonable for extra hardening, can be done in a follow-up..

test/windows/wslc/e2e/WSLCE2EContainerLogsTests.cpp
```diff
@@ -91,7 +96,12 @@ class WSLCE2EContainerLogsTests
         VERIFY_IS_TRUE(result.StdoutContainsSubstring(L"world"));
         auto lines = result.GetStdoutLines();
         VERIFY_IS_TRUE(!lines.empty());
-        VERIFY_IS_TRUE(lines[0].find(L'T') != std::wstring::npos);
+        VERIFY_IS_TRUE(lines[0].size() >= 20);
+        VERIFY_ARE_EQUAL(lines[0][4], L'-');
+        VERIFY_ARE_EQUAL(lines[0][7], L'-');
+        VERIFY_ARE_EQUAL(lines[0][10], L'T');
+        VERIFY_ARE_EQUAL(lines[0][13], L':');
+        VERIFY_ARE_EQUAL(lines[0][16], L':');
```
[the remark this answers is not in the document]
ptrivedi: same answer as above

test/windows/wslc/CommandLineTestCases.h
```diff
@@ -207,6 +207,17 @@ COMMAND_LINE_TEST_CASE(L"container logs --until 1700000000 cont1", L"logs", true
 COMMAND_LINE_TEST_CASE(L"container logs --since 1700000000 --until 1700001000 cont1", L"logs", true)
 COMMAND_LINE_TEST_CASE(L"container logs --since abc cont1", L"logs", false)
 COMMAND_LINE_TEST_CASE(L"container logs --until abc cont1", L"logs", false)
+COMMAND_LINE_TEST_CASE(L"container logs --since 2024-01-15T10:30:00Z cont1", L"logs", true)
+COMMAND_LINE_TEST_CASE(L"container logs --until 2024-01-15T10:30:00Z cont1", L"logs", true)
+COMMAND_LINE_TEST_CASE(L"container logs --since 2024-01-15T10:30:00+05:30 cont1", L"logs", true)
+COMMAND_LINE_TEST_CASE(L"container logs --since 2024-01-15T10:30:00.123456789Z cont1", L"logs", true)
```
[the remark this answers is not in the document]
ptrivedi: addressed

test/windows/wslc/e2e/WSLCE2EContainerLogsTests.cpp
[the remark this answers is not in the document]
ptrivedi: good suggestion. risk is low, and as mentioned, thorough timstamp validation is done in unit tests. That said, using std::chrono::parse on extracted prefix could be an improvement that can be looked at in a follow-up

PR #40827, opened 2026-06-17
Fix wslc network create default driver handling when --driver is not specified

beena352:
his PR fixes wslc network create <name> failing with Invalid pointer (0x80004003) when --driver is not provided.

## PR Checklist

What changed:
- Made WSLCNetworkOptions.Driver nullable in IDL ([unique] LPCSTR Driver).
- Updated WSLCSession::CreateNetwork to default to "bridge" when driver is null or empty.
- Updated tests:
  - NetworkCreateInvalidDriverTest no longer treats empty driver as invalid.
  - Added NetworkCreateDefaultDriverTest for null/empty driver default behavior.

### DIFF
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -2333,7 +2333,7 @@ try
     RETURN_HR_IF_NULL(E_POINTER, Options->Name);
 
     std::string name = Options->Name;
-    std::string driver = (Options->Driver != nullptr && Options->Driver[0] != '\0') ? Options->Driver : WSLCBridgeNetworkDriver;
+    std::string driver = (Options->Driver != nullptr) ? Options->Driver : WSLCBridgeNetworkDriver;
 
     ValidateName(name.c_str(), WSLC_MAX_NETWORK_NAME_LENGTH);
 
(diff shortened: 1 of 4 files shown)


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2331,10 +2331,9 @@ try
 
     RETURN_HR_IF_NULL(E_POINTER, Options);
     RETURN_HR_IF_NULL(E_POINTER, Options->Name);
-    RETURN_HR_IF_NULL(E_POINTER, Options->Driver);
 
     std::string name = Options->Name;
-    std::string driver = Options->Driver;
+    std::string driver = (Options->Driver != nullptr && Options->Driver[0] != '\0') ? Options->Driver : WSLCBridgeNetworkDriver;
```
OneBlue: nit: I don't think we should consider empty string as a valid driver here

OneBlue (approved):
LGTM. 

There's a similar PR pending, but the unit test in this change would be good to merge either way: https://github.com/microsoft/WSL/pull/40788#pullrequestreview-4528964376

beena352:
> LGTM.
> 
> There's a similar PR pending, but the unit test in this change would be good to merge either way: [#40788 (review)](https://github.com/microsoft/WSL/pull/40788#pullrequestreview-4528964376)

Thanks! Yes, I'm aware of #40788, it was pending for a while and I have dependent bug fixes that need this change, so I went ahead with this PR.

PR #40834, opened 2026-06-17
Add --force option to wslc volume rm and network rm commands

ptrivedi:
## Summary

Add `--force` / `-f` flag to `wslc volume rm` and `wslc network rm` commands.

## What `--force` does

When `--force` is specified:

- **Not-found errors are silently suppressed.** Removing a volume or network that does not exist exits with code 0 and produces no stderr output, instead of printing an error and exiting with code 1.
- **Existing volumes/networks are removed normally.** If the resource exists, it is deleted and its name is printed to stdout, same as without `--force`.
- **In-use safety checks are not bypassed.** Attempting to remove a volume that is mounted by a running container still fails with `ERROR_SHARING_VIOLATION`, even with `--force`.

## How this tracks Docker behavior

This matches the semantics of `docker volume rm --force` and `docker network rm --force`, which suppress not-found errors so that `rm --force` is idempotent (safe to call in scripts and cleanup paths without checking existence first). Docker's `--force` on volume/network rm does **not** bypass in-use checks either — that is specific to `docker container rm --force`.

## Third-party storage plugins

In Docker, `--force` additionally allows untracking a volume when a third-party storage plugin is unreachable (the volume metadata is removed even if the plugin cannot clean up the underlying storage). WSLC does not support third-party storage plugins — volumes are either guest volumes (backed by the in-VM Docker local driver) or VHD volumes (backed by a host `.vhdx` file). Since there is no external plugin failure path, `--force` does not need to handle that scenario.

## Changes

- **ArgumentDefinitions.h**: New `VolumeForce` and `NetworkForce` arg types
- **VolumeRemoveCommand.cpp / NetworkRemoveCommand.cpp**: Register `--force` argument
- **VolumeTasks.cpp / NetworkTasks.cpp**: Read force flag; suppress not-found errors and exit code when set
- **Resources.resw**: Localization strings for the new argument descriptions
- **E2E tests**: Force + not-found, force + valid removal, force + mixed found/not-found scenarios; updated help message expectations

### DIFF
--- a/src/windows/wslc/commands/NetworkRemoveCommand.cpp
+++ b/src/windows/wslc/commands/NetworkRemoveCommand.cpp
@@ -27,6 +27,7 @@ namespace wsl::windows::wslc {
 std::vector<Argument> NetworkRemoveCommand::GetArguments() const
 {
     return {
+        Argument::Create(ArgType::NetworkForce),
         Argument::Create(ArgType::NetworkName, true, NO_LIMIT),
     };
 }
--- a/src/windows/wslc/commands/NetworkRemoveCommand.cpp
+++ b/src/windows/wslc/commands/NetworkRemoveCommand.cpp
@@ -27,8 +27,8 @@ namespace wsl::windows::wslc {
 std::vector<Argument> NetworkRemoveCommand::GetArguments() const
 {
     return {
-        Argument::Create(ArgType::NetworkForce),
         Argument::Create(ArgType::NetworkName, true, NO_LIMIT),
+        Argument::Create(ArgType::Force, std::nullopt, std::nullopt, Localization::WSLCCLI_NetworkForceArgDescription()),
     };
 }
 
--- a/src/windows/wslc/commands/VolumeRemoveCommand.cpp
+++ b/src/windows/wslc/commands/VolumeRemoveCommand.cpp
@@ -27,6 +27,7 @@ namespace wsl::windows::wslc {
 std::vector<Argument> VolumeRemoveCommand::GetArguments() const
 {
     return {
+        Argument::Create(ArgType::VolumeForce),
         Argument::Create(ArgType::VolumeName, true, NO_LIMIT),
     };
 }
--- a/src/windows/wslc/commands/VolumeRemoveCommand.cpp
+++ b/src/windows/wslc/commands/VolumeRemoveCommand.cpp
@@ -27,8 +27,8 @@ namespace wsl::windows::wslc {
 std::vector<Argument> VolumeRemoveCommand::GetArguments() const
 {
     return {
-        Argument::Create(ArgType::VolumeForce),
         Argument::Create(ArgType::VolumeName, true, NO_LIMIT),
+        Argument::Create(ArgType::Force, std::nullopt, std::nullopt, Localization::WSLCCLI_VolumeForceArgDescription()),
     };
 }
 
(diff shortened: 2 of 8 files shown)


src/windows/wslc/commands/NetworkRemoveCommand.cpp
```diff
@@ -27,6 +27,7 @@ namespace wsl::windows::wslc {
 std::vector<Argument> NetworkRemoveCommand::GetArguments() const
 {
     return {
+        Argument::Create(ArgType::NetworkForce),
```
dkbennett: We generally don't put these in alphabetical order (which is normally correct) but for help generation and processing it's more consistent with other commands to put the required arguments and positional arguments first, and then put optional arguments after that.

dkbennett: This could be: 
Argument::Create(ArgType::Force, std::nullopt, std::nullopt, Localization::WSLCCLI_NetworkForceArgDescription()),

To use the existing "Force" argument type definition but override the description.

src/windows/wslc/commands/VolumeRemoveCommand.cpp
```diff
@@ -27,6 +27,7 @@ namespace wsl::windows::wslc {
 std::vector<Argument> VolumeRemoveCommand::GetArguments() const
 {
     return {
+        Argument::Create(ArgType::VolumeForce),
```
dkbennett: Like network command, this should ideally be below the VolumeName arg.

And the argument coudl reuse the normal force like this:
Argument::Create(ArgType::Force, std::nullopt, std::nullopt, Localization::WSLCCLI_VolumeForceArgDescription()),

dkbennett (commented):
This is correct, but recommend reuse existing Force argument and a minor ordering change.

src/windows/wslc/tasks/VolumeTasks.cpp
```diff
@@ -60,7 +60,11 @@ static bool TryDeleteVolume(Session& session, const std::string& volumeName)
     {
         if (ex.GetErrorCode() == WSLC_E_VOLUME_NOT_FOUND)
         {
-            PrintMessage(Localization::MessageWslcVolumeNotFound(volumeName.c_str()), stderr);
+            if (!force)
```
OneBlue, line 63: I don't think this is the right behavior. 

If `--force` is passed, we should pass `force=true` when we call the [removal API](https://docs.docker.com/reference/api/engine/version/v1.52/#tag/Volume/operation/VolumeDelete)

(At least in the `WSLCGuestVolumeImpl` path)

OneBlue (changes requested):

src/windows/wslc/tasks/VolumeTasks.cpp
ptrivedi: force=true in the removal API is only useful when there are third party driver plugins involved. If there is no third party volume plugin support, it will do absolutely nothing. We could still pass it to the removal API if we think it is needed for completeness sake. But if you are thinking force=true is going to forcefully remove 'in-use' volumes or do anything else for local volumes, it doesn't.

OneBlue: That's a good point. It's probably not worth doing if the behavior is the same for the volume we support

ptrivedi: I think, in absence of third party volume plugin support, a big thing that force=true for volume removal is beneficial for is integration into the CI/CD pipelines... where it needs to be quiet :)

OneBlue (approved):
LGTM

PR #40837, opened 2026-06-17
Add --label support to wslc image build

johnstep:


src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -85,7 +86,7 @@ void BuildImage(CLIExecutionContext& context)
 
     auto cancelEvent = context.CreateCancelEvent();
     BuildImageCallback callback(cancelEvent, context.Args.Contains(ArgType::Verbose));
-    services::ImageService::Build(session, contextPath, tags, buildArgs, dockerfilePath, target, flags, &callback, cancelEvent);
+    services::ImageService::Build(session, contextPath, tags, buildArgs, labels, dockerfilePath, target, flags, &callback, cancelEvent);
```
AmelBawa-msft, line 93: For image build, do we need to call ParseLabel before passing them to the service API?

Example from other commands:
- https://github.com/microsoft/WSL/blob/8e3c0ad8b71714a737d165a6160a892714c2b4c8/src/windows/wslc/tasks/ContainerTasks.cpp#L584
- https://github.com/microsoft/WSL/blob/8e3c0ad8b71714a737d165a6160a892714c2b4c8/src/windows/wslc/tasks/VolumeTasks.cpp#L89
- https://github.com/microsoft/WSL/blob/8e3c0ad8b71714a737d165a6160a892714c2b4c8/src/windows/wslc/tasks/NetworkTasks.cpp#L86

johnstep: Moby will validate it of course, but it is a good call to do this, especially for consistency. Thanks!

src/windows/service/inc/wslc.idl
```diff
@@ -679,6 +679,7 @@ typedef struct _WSLCBuildImageOptions
     WSLCStringArray BuildArgs; // KEY=VALUE pairs passed as --build-arg to docker.
     LPCSTR Target;             // Target build stage name passed as --target to docker.
     WSLCBuildImageFlags Flags; // WSLCBuildImageFlags
+    WSLCStringArray Labels;    // KEY=VALUE pairs passed as --label to docker.
 } WSLCBuildImageOptions;
```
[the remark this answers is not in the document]
johnstep: This is fine because the binaries are always installed together.

OneBlue (approved):

PR #40840, opened 2026-06-18
SDK updates

JohnMcPMS:
Various updates to the SDK.

## PR Checklist

- [X **Communication:** I've discussed this with core contributors already. If work hasn't been agreed, this work might be rejected

Contains the following SDK changes:
- Adds session crash dump callback to the WinRT API surface
- Adds missed `SdkNeedsUpdate` value to component check enum
- Adds WSLC error values enum to the WinRT API surface
- Fixes SYSTEM user check in install API (OS calls require an impersonation token but providing no token causes it to use the effective token)
- Adds a fallback to install the latest WSL prerelease version from GitHub when it is not found via Windows Update
- Adds more/better trace logging to the Windows Update code path
- Allows the "update from GitHub" code path to be less invasive for the calling process

## Validation Steps Performed
Regression and new tests.  Manual validation of elevation check and fallback install changes in install API.


src/windows/common/install.cpp
```diff
@@ -333,24 +342,34 @@ void wsl::windows::common::install::MsiMessageCallback(INSTALLMESSAGE type, LPCW
...
-                ExitProcess(0);
-            }
-            return FALSE;
-        },
-        TRUE));
+    bool clearHandler = false;
+    auto cleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&] {
+        if (clearHandler)
+        {
+            SetConsoleCtrlHandler(nullptr, FALSE);
```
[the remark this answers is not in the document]
JohnMcPMS: According to the documentation, providing `nullptr` and `FALSE` should "return to normal CTRL signal handling", presumably removing all handlers. This is also what the code did previously.

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -1718,6 +1714,24 @@ try
...
         wuContext.RunUpdateFlow(true, callback);
+
+        // Because we do a forced install here, we expect an update.
+        if (wuContext.GetUpdateCount() == 0)
+        {
+            // During the preview period, the package may not be published yet, so fall back to getting it from GH.
+            // When moving to GA, change this to a hard error to indicate a service configuration issue.
+            if (callback)
+            {
+                callback(0);
```
OneBlue, line 1730: It's feeling a bit natural to force a 0 -> progress in all situations (especially given that this will also run on failures). 

Maybe we could just do 100 if `UpdatePackage` succeeds ?

src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -1676,10 +1672,10 @@ try
     THROW_HR_IF(runtimeResult, runtimeResult != REGDB_E_CLASSNOTREG && runtimeResult != WSLC_E_SDK_UPDATE_NEEDED);
 
     // Installing these components requires elevation.
-    auto token = wil::open_current_access_token();
     RETURN_HR_IF(
```
OneBlue, line 1680: What's the reason for this change ?

JohnMcPMS: `open_current_access_token` gets a real handle to the effective token and explicitly suggests to use `GetCurrentThreadEffectiveToken` (which returns a pseudohandle) if you are just going to call `get_token_information` anyway.  Since that is what we do, I switched to that per the recommendation.

`CheckTokenMembership` (in `IsTokenLocalSystem`) requires an impersonation token to function.  If you give it null, it will use the effective token per the docs:
```
If TokenHandle is NULL, CheckTokenMembership uses the impersonation token of the calling thread. If the thread is not impersonating, the function duplicates the thread's primary token to create an impersonation token.
```
If you were calling from a non-elevated thread and not impersonating, this would cause the overall install to fail with an "impersonation required" error rather than the desired "elevation required" error.

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: Any real failure in `RunUpdateFlow` will throw.  The path out that contains no updates sets progress to 100 itself.  I was resetting to 0 only in the fallback workaround and expect that this will just become an error by GA.

OneBlue: Ok makes sense. The latest iteration looks good now that only set to 100 on success

OneBlue (dismissed):

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -331,5 +344,24 @@ namespace Microsoft.WSL.Containers
         Windows.Storage.Streams.IBuffer Sha256 { get; };
         UInt64 SizeBytes { get; };
         Windows.Foundation.DateTime CreatedTimestamp { get; };
+    };
+
+    // Ensure wslcsdk.h and wslc.idl are also updated.
+    enum Errors
```
florelis:
```suggestion
    enum Error
```

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -28,7 +28,18 @@ namespace Microsoft.WSL.Containers
...
     };
 
-    delegate void SessionTerminationHandler(SessionTerminationReason reason);
+    delegate void SessionTerminationHandler(SessionTerminationReason reason);
+
+    runtimeclass SessionCrashDumpInformation
+    {
+        String DumpPath { get; };
+        String ProcessName { get; };
+        UInt64 Pid { get; };
```
florelis: We use UInt32 for `Process`. Can a Session have a larger range of PIDs?

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -28,7 +28,18 @@ namespace Microsoft.WSL.Containers
...
+        UInt32 Signal { get; };
```
florelis, line 38: Do we need to support more signals than what is in the Signal enum?

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -70,6 +81,7 @@ namespace Microsoft.WSL.Containers
         IVectorView<ImageInfo> Images { get; };
 
         event SessionTerminationHandler Terminated;
+        event SessionCrashDumpHandler ProcessCrashed;
```
florelis: Could we make the property name match the handler name more clearly? Like `SessionCrashed`

florelis (commented):
A few comments from pointing my local Copilot to our WinRT API guidelines:

src/windows/WslcSDK/winrt/wslcsdk.idl
JohnMcPMS: This is what comes back from Linux at the moment and I don't want to force it to be 32 bit without some knowledge as to why it is 64 bit right now.

JohnMcPMS: @OneBlue , the socket communication sends a 64-bit PID for the crash dumps, but the Process object (here and in the runtime) uses a 32-bit PID value.  It seems unlikely that I can should be able to run a new process, get the 32-bit PID, then have it crash and produce a 64-bit PID.  It seems like we should change the C API and this to be 32-bit PID values for crash dumps.  Thoughts?

src/windows/WslcSDK/winrt/wslcsdk.idl
JohnMcPMS: Our input signals are a small subset of the potential signal space.  I think it better to leave this open ended.

src/windows/WslcSDK/winrt/wslcsdk.idl
JohnMcPMS: Changing the type names to be `Process` related.

src/windows/WslcSDK/winrt/wslcsdk.idl
JohnMcPMS: Planning to convert the entire crash dump stack to use 32-bit PID in a followup.

src/windows/common/install.cpp
```diff
@@ -77,9 +77,9 @@ bool PromptForKeyPressWithTimeout()
     return waitResult == std::future_status::ready && future.get();
 }
 
-int UpdatePackageImpl(bool preRelease, bool repair)
+int UpdatePackageImpl(bool preRelease, bool repair, bool CallerOwnsProcess)
```
florelis: nit: PascalCase?

src/windows/WslcSDK/winrt/Session.h
```diff
@@ -38,6 +38,8 @@ struct Session : SessionT<Session>
     winrt::Windows::Foundation::Collections::IVectorView<winrt::Microsoft::WSL::Containers::ImageInfo> Images();
     winrt::event_token Terminated(winrt::Microsoft::WSL::Containers::SessionTerminationHandler const& handler);
     void Terminated(winrt::event_token const& token) noexcept;
+    winrt::event_token ProcessCrashed(winrt::Microsoft::WSL::Containers::ProcessCrashDumpHandler const& handler);
```
florelis: nit: we're still calling this three different things "process crash", "process crash dump" and "crash dump". Could we settle on just one?

florelis (dismissed):

src/windows/WslcSDK/winrt/Session.h
```diff
@@ -47,13 +49,17 @@ struct Session : SessionT<Session>
 
     // Threadpool callback that raises the Terminated event once the session's termination handle is signaled.
     static void CALLBACK OnTerminated(PTP_CALLBACK_INSTANCE instance, PVOID context, PTP_WAIT wait, TP_WAIT_RESULT waitResult) noexcept;
+    static void CALLBACK OnCrashDump(const WslcSessionCrashDumpInfo* info, PVOID context) noexcept;
 
     winrt::event<winrt::Microsoft::WSL::Containers::SessionTerminationHandler> m_terminatedEvent;
+    winrt::event<winrt::Microsoft::WSL::Containers::ProcessCrashHandler> m_crashDumpEvent;
```
OneBlue, line 55: Reading this, it's looking like this event callback is not blocking, which can cause a race since the service will start deleting crashes once the service is full. 

To work around this, I would recommend having the caller explicitly register a callback to us before creating the session, and making it blocking so the caller has the guarantee that the dump still exists until the callback returns

florelis (approved):

src/windows/WslcSDK/winrt/Session.h
JohnMcPMS: I'm not seeing that; it looks fully blocking from the runtime service back through to this.  Here is the stack on the caller side from the tests:
```
 	wsltests.dll!WslcSdkWinRtTests::ProcessCrashedEvent::__l3::<lambda_2>::operator()(winrt::Microsoft::WSL::Containers::ProcessCrashInformation info) Line 335	C++
 	wsltests.dll!winrt::impl::delegate<winrt::Microsoft::WSL::Containers::ProcessCrashHandler,`WslcSdkWinRtTests::ProcessCrashedEvent'::`3'::<lambda_2>>::Invoke(void * information) Line 2680	C++
 	wsltests.dll!winrt::Microsoft::WSL::Containers::ProcessCrashHandler::operator()(const winrt::Microsoft::WSL::Containers::ProcessCrashInformation & information) Line 4131	C++
 	wsltests.dll!winrt::impl::invoke<winrt::Microsoft::WSL::Containers::ProcessCrashHandler,winrt::Microsoft::WSL::Containers::implementation::ProcessCrashInformation>(const winrt::Microsoft::WSL::Containers::ProcessCrashHandler & delegate, const winrt::Microsoft::WSL::Containers::implementation::ProcessCrashInformation & <args_0>) Line 5895	C++
 	wsltests.dll!winrt::event<winrt::Microsoft::WSL::Containers::ProcessCrashHandler>::operator()<winrt::Microsoft::WSL::Containers::implementation::ProcessCrashInformation>(const winrt::Microsoft::WSL::Containers::implementation::ProcessCrashInformation & <args_0>) Line 6016	C++
 	wsltests.dll!winrt::Microsoft::WSL::Containers::implementation::Session::OnCrashDump(const WslcSessionCrashDumpInfo * info, void * context) Line 343	C++
>	wslcsdk.dll!CrashDumpCallback::OnCrashDump(const wchar_t * DumpPath, const char * ProcessName, unsigned long Pid, unsigned long Signal, unsigned __int64 Timestamp) Line 34	C++
```
It starts at the C SDK COM callback entrypoint and flows on this thread straight to the lambda in the test.

OneBlue: Ok nice, in that case we should be good. Thank you for double checking !

OneBlue (approved):

PR #40847, opened 2026-06-18
 Apply WinRT API review feedback to WSLC SDK projection

florelis:

 Address API design review recommendations for the  Microsoft.WSL.Containers WinRT projection. C SDK layer is unchanged; all changes are projection-only.

(Mostly done by Copilot)

## PR Checklist

 Naming:
 - Drop "Flags" suffix: DeleteContainerFlags -> DeleteContainerOption, ComponentFlags -> Component
 - Rename MemoryMB -> MemorySizeInMB, CmdLine -> CommandLine, SizeBytes/SizeInBytes -> Size

 Type shape:
 - Replace SessionFeatureFlags/ContainerFlags enums with individual boolean properties (IsGpuEnabled, IsAutoRemoveEnabled, IsPrivileged)
 - Make Component a non-flags enum; GetMissingComponents() now returns IVectorView<Component>
 - Replace VhdOptions OwnerUid/OwnerGid with a VhdOwner struct exposed via IReference<VhdOwner> Owner
 - Convert Images property to GetImages() method

 Lifetime:
 - Implement IClosable on Session, Container, and Process; final_release routes through Close()
 - Guard Container handle access with EnsureNotClosed() (RO_E_CLOSED)

 Async:
 - Add synchronous variants for PullImage, ImportImage, LoadImage, PushImage, and InstallWithDependencies, implemented directly against the C SDK per the synchronous API guidance

 Add TODO for structured Inspect() return. Update WinRT tests for all renames and shape changes.


src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -38,13 +31,13 @@ namespace Microsoft.WSL.Containers
         String StoragePath;
 
         Windows.Foundation.IReference<UInt32> CpuCount;
-        Windows.Foundation.IReference<UInt32> MemoryMB;
+        Windows.Foundation.IReference<UInt32> MemorySizeInMB;
         Windows.Foundation.IReference<Windows.Foundation.TimeSpan> Timeout;
         VhdOptions VhdRequirements;
-        SessionFeatureFlags FeatureFlags;
+        Boolean IsGpuEnabled;
```
JohnMcPMS: Feels like it should still be `EnableGpu` as this is a directive, not an inquiry.

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -150,7 +138,9 @@ namespace Microsoft.WSL.Containers
         Windows.Foundation.IReference<ContainerNetworkingMode> NetworkingMode;
         String HostName;
         String DomainName;
-        ContainerFlags Flags;
+        Boolean IsAutoRemoveEnabled;
+        Boolean IsGpuEnabled;
+        Boolean IsPrivileged;
```
JohnMcPMS: Same comment RE: naming based on being in the settings object.

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -166,14 +156,15 @@ namespace Microsoft.WSL.Containers
...
+    runtimeclass Container : Windows.Foundation.IClosable
     {
         void Start();
         void Stop(Signal signal, Windows.Foundation.TimeSpan timeout);
-        void Delete(DeleteContainerFlags flags);
+        void Delete(DeleteContainerOption flags);
 
         Process CreateProcess(ProcessSettings newProcessSettings);
 
+        // TODO: This should return a structured type instead of raw JSON.
```
JohnMcPMS: I'm not sure that I agree with this unless the JSON is prescribed to be of a specific schema from a standards body.

src/windows/WslcSDK/winrt/Container.cpp
```diff
@@ -118,9 +118,32 @@ winrt::Microsoft::WSL::Containers::ContainerState Container::State()
...
+    m_initProcess = nullptr;
+
+    // Methods called after Close() will fail due to EnsureNotClosedStarted().
+    m_container.reset();
+}
+
+void Container::final_release(std::unique_ptr<Container> self)
+{
+    // Ensure cleanup when refcount drops to zero even if Close() was not called explicitly.
+    self->Close();
```
JohnMcPMS, line 146: This seems unnecessary as the destructor would do the same shortly afterward anyway.

src/windows/WslcSDK/winrt/ImageInfo.h
```diff
@@ -24,13 +24,13 @@ struct ImageInfo : ImageInfoT<ImageInfo>
...
-    uint64_t SizeBytes();
+    uint64_t Size();
     winrt::Windows::Foundation::DateTime CreatedTimestamp();
 
 private:
     hstring m_name;
     winrt::Windows::Storage::Streams::IBuffer m_sha256;
-    uint64_t m_sizeBytes;
+    uint64_t m_size;
     winrt::Windows::Foundation::DateTime m_createdTimestamp;
```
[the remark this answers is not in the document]
JohnMcPMS: Please do this.

src/windows/WslcSDK/winrt/precomp.h
```diff
@@ -23,6 +23,9 @@ Module Name:
 #include <winrt/Windows.Foundation.Collections.h>
 #include <winrt/Windows.Storage.Streams.h>
 
+// Windows' LoadImage macro conflicts with Session::LoadImage
+#undef LoadImage
```
JohnMcPMS, line 27: I worry about how the C++ projection will fare here...

src/windows/WslcSDK/winrt/wslcsdk.idl
florelis: Is the JSON a common representation for containers elsewhere?
I was assuming it was arbitrary data that we decided on.

JohnMcPMS: https://docs.docker.com/reference/api/engine/version/v1.53/#tag/Container/operation/ContainerInspect

src/windows/WslcSDK/winrt/ContainerSettings.cpp
```diff
@@ -126,19 +126,49 @@ void ContainerSettings::DomainName(hstring const& value)
...
 
-ContainerFlags ContainerSettings::Flags()
+bool ContainerSettings::IsAutoRemoveEnabled()
 {
-    return m_flags;
+    return WI_IsFlagSet(m_containerFlags, WSLC_CONTAINER_FLAG_AUTO_REMOVE);
 }
 
-void ContainerSettings::Flags(ContainerFlags const& value)
+void ContainerSettings::IsAutoRemoveEnabled(bool value)
```
OneBlue: Is this a winrt pattern ? I'm finding it a bit confusing for a method named `IsAutoRemoveEnabled()` to actually be a setter

src/windows/WslcSDK/winrt/Session.cpp
```diff
@@ -142,6 +181,22 @@ IAsyncActionWithProgress<winrt::Microsoft::WSL::Containers::ImageProgress> Sessi
     THROW_MSG_IF_FAILED(hr, errorMessage);
 }
 
+void Session::LoadImage(hstring const& path)
+{
+    if (path.empty())
+    {
+        throw winrt::hresult_invalid_argument(L"Path cannot be empty");
```
OneBlue, line 192: I don't think we need this layer to handle this. The service will catch issues like this one

src/windows/WslcSDK/winrt/ContainerSettings.cpp
florelis: Yes, this is the WinRT pattern. It is confusing when you first see it in C++ but the projection in other languages is better.
```C++
// C++
T Property(); // Get
void Property(T); // Set
// C#
T Property { get; set; }
```

src/windows/WslcSDK/winrt/precomp.h
florelis: Yeah... I think the best would be to change the name, but I don't know what to.
This undef is private, so consumers wouldn't get it automatically, but it would still be confusing to see an error like "Service::LoadImageW() does not exist"

src/windows/WslcSDK/winrt/Session.cpp
florelis: If it is simple validation like this, I would rather avoid the round-trip to the service. For async versions I would also prefer to exit early before doing the whole dance to switch to a background thread.

I can remove it from the synchronous versions. I think the C SDK already has the validation like this.

src/windows/WslcSDK/winrt/Container.cpp
florelis: I think even if it will happen afterward, since we already have a Close(), it could be better to do that instead of relying on the destruction order.

src/windows/WslcSDK/winrt/wslcsdk.idl
```diff
@@ -178,14 +168,15 @@ namespace Microsoft.WSL.Containers
...
+    runtimeclass Container : Windows.Foundation.IClosable
     {
         void Start();
         void Stop(Signal signal, Windows.Foundation.TimeSpan timeout);
-        void Delete(DeleteContainerFlags flags);
+        void Delete(DeleteContainerOption flags);
 
         Process CreateProcess(ProcessSettings newProcessSettings);
 
+        // JSON response schema at: https://docs.docker.com/reference/api/engine/version/v1.53/#tag/Container/operation/ContainerInspect
```
JohnMcPMS, line 179: In theory it would be the OCI container schema, but it doesn't appear to be.  And I'm not sure that they have a schema themselves.  I also don't know that we want to have a link to a version specific example, in fact, I don't think we want to have a link at all unless we are committing to keeping it consistent.

OneBlue (approved):

src/windows/WslcSDK/winrt/wslcsdk.idl
florelis: That's fair. I'll remove the comment but I'll do in a separate PR so that we don't have to wait for the full build again before merging everything else.

PR #40849, opened 2026-06-18
Fix potential issues in DockerHttpClient SendRequest

yao-msft:

- Change header scan to start at newly peeked bytes instead of rescanning from 0 (heap overflow/integer underflow)
- Check max response header size (host memory exhaustion)
- Add timeout to receive bytes (indefinite stall)

## PR Checklist

## Validation Steps Performed
Existing tests

### DIFF
--- a/src/windows/wslcsession/DockerHTTPClient.cpp
+++ b/src/windows/wslcsession/DockerHTTPClient.cpp
@@ -822,7 +822,7 @@ std::pair<DockerHTTPClient::HTTPResponse, wil::unique_socket> DockerHTTPClient::
         THROW_HR_IF(E_ABORT, bytesRead == 0);
 
         // Scan only the newly peeked bytes [Offset, Offset + bytesRead)
-        size_t i{};
+        size_t i = 0;
         for (i = Offset; i < bytesRead + Offset && lineFeeds < 2; i++)
         {
             if (buffer[i] == '\n')
@@ -841,6 +841,7 @@ std::pair<DockerHTTPClient::HTTPResponse, wil::unique_socket> DockerHTTPClient::
         // Consume the scanned header bytes from the socket
         bytesRead = common::socket::Receive(
             context->stream.native_handle(), gsl::span(reinterpret_cast<gsl::byte*>(buffer.data() + Offset), toConsume), m_exitingEvent, 0, headerReceiveTimeoutMs);
+        THROW_HR_IF(E_ABORT, bytesRead == 0); // E_ABORT case after peek but before consume
         THROW_HR_IF_MSG(
             E_UNEXPECTED, static_cast<size_t>(bytesRead) != toConsume, "Short read consuming HTTP header: got %d, expected %zu", bytesRead, toConsume);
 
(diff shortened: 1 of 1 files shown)


src/windows/wslcsession/DockerHTTPClient.cpp
```diff
@@ -797,6 +797,9 @@ std::pair<DockerHTTPClient::HTTPResponse, wil::unique_socket> DockerHTTPClient::
 
     // Parse the response header
     constexpr auto bufferSize = 16 * 1024;
+    // Docker response header max size.
+    constexpr size_t maxHeaderSize = 64 * 1024;
+    constexpr DWORD headerReceiveTimeoutMs = 30 * 1000;
```
OneBlue: I don't think we should have a timeout here. This will likely break scenarios like `wslc stop -t 31 <container>`.

If the worry is that requests can get stuck, those can be unblocked when the session terminates. As an additional measure, we could also plumb the cancel handle + client process handle here to cancel the request when the caller cancels the call

src/windows/wslcsession/DockerHTTPClient.cpp
```diff
@@ -797,6 +797,9 @@ std::pair<DockerHTTPClient::HTTPResponse, wil::unique_socket> DockerHTTPClient::
 
     // Parse the response header
     constexpr auto bufferSize = 16 * 1024;
+    // Docker response header max size.
+    constexpr size_t maxHeaderSize = 64 * 1024;
```
OneBlue: I'd recommend making this a bit bigger just be safe. We had the issue in WSL and hit our buffer limit. I'd recommend making it to something like 1MB, which we should never hit, but 1MB of allocated memory is still pretty unsignificant

src/windows/wslcsession/DockerHTTPClient.cpp
```diff
@@ -807,16 +810,20 @@ std::pair<DockerHTTPClient::HTTPResponse, wil::unique_socket> DockerHTTPClient::
...
-            context->stream.native_handle(), gsl::span(reinterpret_cast<gsl::byte*>(buffer.data() + Offset), bufferSize), m_exitingEvent, MSG_PEEK);
+            context->stream.native_handle(), gsl::span(reinterpret_cast<gsl::byte*>(buffer.data() + Offset), bufferSize), m_exitingEvent, MSG_PEEK, headerReceiveTimeoutMs);
 
         THROW_HR_IF(E_ABORT, bytesRead == 0);
 
-        size_t i{};
-        for (i = 0; i < bytesRead + Offset && lineFeeds < 2; i++)
+        // Scan only the newly peeked bytes [Offset, Offset + bytesRead)
+        size_t i = 0;
+        for (i = Offset; i < bytesRead + Offset && lineFeeds < 2; i++)
```
OneBlue, line 825: Reading this, I'm realizing that incorrectly handle the case where one read returns "\n", followed by another read that returns "\n". In that case we'll incorrectly think that we received "\r\n". Probably something we should fix, but that can go in a followup change

OneBlue (changes requested):

src/windows/wslcsession/DockerHTTPClient.cpp
yao-msft: Will do as followup.

OneBlue (approved):
LGTM

PR #40856, opened 2026-06-19
CLI: Fix image import to output the image ID, allow untagged images, and support --no-trunc

dkbennett:
This PR fixes three related bugs with wslc image import, all of which align it with the equivalent Docker command.

1. Display image ID on stdout after import

Docker CLI outputs the imported image's ID to stdout. We now do the same — the server captures the ID from Docker's API response (via the status field) and returns it through a new [out] LPSTR* ImageId parameter on ImportImage. The CLI displays it truncated to 12 hex chars (matching Docker's default), with sha256: prefix stripped.

2. Allow untagged image imports (empty/null image name)

Docker allows docker import file.tar without specifying a repository:tag, creating a <none>:<none> image. We now support the same — ImageName is [in, unique] (nullable) in the IDL, and the server passes empty repo/tag to the Docker API when no name is provided.

3. Add --no-trunc support for import ID output

Added ArgType::NoTrunc to the import command's arguments. When passed, the full image ID is output instead of the 12-char truncated form.

## PR Checklist

## Validation Steps Performed
* Updated tests to verify image output trunc, and no trunc
* Enabled the previously skipped untagged image to now work and improved it a bit to validate a few more things.


test/windows/wslc/e2e/WSLCE2EImageImportTests.cpp
```diff
@@ -66,22 +67,66 @@ class WSLCE2EImageImportTests
         auto importResult = RunWslc(std::format(L"image import \"{}\" {}", SavedArchivePath.wstring(), ImportedImage.NameAndTag()));
         importResult.Verify({.Stderr = L"", .ExitCode = 0});
 
+        // Verify the import outputs a truncated image ID (12 chars + \r\n)
+        VERIFY_IS_TRUE(importResult.Stdout.has_value());
+        VERIFY_ARE_EQUAL(importResult.Stdout->size(), static_cast<size_t>(12 + 2));
+
```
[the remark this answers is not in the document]
dkbennett: OK fine I'll add a helper that has a better id validation for trunc and no trunc.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1267,6 +1287,10 @@ void WSLCSession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request,
         else if (parsed.status.has_value())
```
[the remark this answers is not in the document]
dkbennett: I'll update comment because its supposed to be in the status, and getting it from the stream might confuse it with the digest so I want to stick with status.

test/windows/wslc/e2e/WSLCE2EImageImportTests.cpp
```diff
@@ -47,13 +48,16 @@ class WSLCE2EImageImportTests
     WSLC_TEST_METHOD(WSLCE2E_Image_Import_HelpCommand)
     {
         auto result = RunWslc(L"image import --help");
-        result.Verify({.Stdout = GetHelpMessage(), .Stderr = L"", .ExitCode = 0});
+        result.Verify({.Stderr = L"", .ExitCode = 0});
+        VERIFY_IS_FALSE(result.Stdout.value().empty());
     }
```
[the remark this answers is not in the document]
dkbennett: yes this is intentional. We dont need to be validating every character of the help string, its' generated from the arguments automatically.

test/windows/wslc/e2e/WSLCE2EImageImportTests.cpp
```diff
@@ -66,22 +70,66 @@ class WSLCE2EImageImportTests
...
+
+        // Extract the returned ID, validate its format, and use it for cleanup
+        auto imageId = importResult.GetStdoutOneLine();
+        VerifyIdOutput(imageId, true);
+
+        // Verify that there is now one more untagged image
+        VERIFY_ARE_EQUAL(countUntaggedImages(), untaggedBefore + 1);
+
+        // Clean up the untagged image using the returned ID
+        auto deleteResult = RunWslc(std::format(L"image rm {}", imageId));
```
OneBlue: nit: This should be done in a scope_exit_log so the test doesn't leak the image on failure

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1267,6 +1291,10 @@ void WSLCSession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request,
...
         {
             WSL_LOG("ImageImportProgress", TraceLoggingValue(parsed.status->c_str(), "Status"));
+            if (parsed.status->starts_with("sha256:"))
+            {
+                imageId = *parsed.status;
```
OneBlue, line 1306: nit: I think it would be a good to make fail with E_UNEXPECTED if we hit this block and `imageId` is not empty

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1176,40 +1176,63 @@ try
...
+        *ImageId = wil::make_unique_ansistring<wil::unique_cotaskmem_ansistring>(imageId.c_str()).release();
+    }
 
-    OnImageCreated(ImageName);
     return S_OK;
 }
 CATCH_RETURN();
 
-void WSLCSession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request, const WSLCHandle ImageHandle)
+std::string WSLCSession::ImportImageImpl(DockerHTTPClient::HTTPRequestContext& Request, const WSLCHandle ImageHandle)
```
OneBlue: nit: If we expect that the image id might not always be returned, I recommend making this an std::optional to make it obvious to the caller that it might be empty

OneBlue (commented):
LGTM, minor comments

src/windows/wslcsession/WSLCSession.cpp
dkbennett: Added

src/windows/wslcsession/WSLCSession.cpp
dkbennett: It can indeed be empty, this code path is used a few places and not all of them result in an image status, so changed it to optional and callers expecting it check for the value.

test/windows/wslc/e2e/WSLCE2EImageImportTests.cpp
dkbennett: Fixed, also added a dangling image check for untagged and added to image import tests as extra resiliency.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1176,40 +1176,61 @@ try
...
-    RETURN_HR_IF_NULL(E_POINTER, ImageName);
-    RETURN_HR_IF(E_INVALIDARG, strlen(ImageName) > WSLC_MAX_IMAGE_NAME_LENGTH);
+    RETURN_HR_IF_NULL(E_POINTER, ImageId);
+    *ImageId = nullptr;
+
+    std::string repo;
+    std::string tag;
 
-    auto [repo, tagOrDigest] = wslutil::ParseImage(ImageName);
+    if (ImageName != nullptr && strlen(ImageName) > 0)
```
OneBlue: nit: I think we should reject an empty image name here

OneBlue (dismissed):
LGTM. Minor comment

src/windows/wslcsession/WSLCSession.cpp
dkbennett: removed the strlen so we delegate to ParseImage

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1187,40 +1187,61 @@ try
...
+    std::string repo;
+    std::string tag;
 
-    auto [repo, tagOrDigest] = wslutil::ParseImage(ImageName);
+    if (ImageName != nullptr)
+    {
+        RETURN_HR_IF(E_INVALIDARG, strlen(ImageName) > WSLC_MAX_IMAGE_NAME_LENGTH);
 
-    THROW_HR_IF_MSG(E_INVALIDARG, !tagOrDigest.has_value(), "Expected tag for image import: %hs", ImageName);
+        auto [parsedRepo, tagOrDigest] = wslutil::ParseImage(ImageName);
```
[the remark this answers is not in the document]
dkbennett: No we specifically want emptystring to be invalid, they can just not pass the argument if they don't want an image name

OneBlue (approved):

PR #40857, opened 2026-06-19
Fix Lost sub directory structure when mapping to a VHD volume

kvega005:

Fixes VHD-backed named volumes showing only an empty `lost+found` directory instead of the container image's contents on first use.

When a VHD volume is formatted with `mkfs.ext4`, ext4 always creates a `lost+found` directory at the filesystem root. Docker only seeds a named volume with the image's contents (copy-up) when the volume appears empty, so the stray `lost+found` made VHD volumes look non-empty and silently suppressed seeding — guest (Docker-native) volumes were unaffected. This change removes the `lost+found` directory after formatting so VHD volumes seed just like guest volumes.

## PR Checklist

- [ ] **Closes:** #xxx
- [x] **Communication:** Discussed with WSLC contributors
- [x] **Tests:** Added an e2e test; existing volume tests pass
- [x] **Localization:** New warning string added to `en-US/Resources.resw`
- [ ] **Dev docs:** N/A

Root cause: ext4's `lost+found` makes a freshly formatted VHD volume non-empty, so Docker skips copy-up and the image's files never get seeded into the volume.

Changes:
- **`WSLCVhdVolume`**: After formatting + mounting, `RemoveLostFoundDirectory()` strips `lost+found` so Docker seeds the volume. The deletion is gated: it only runs when `lost+found` is the **sole** entry in the volume root, so a volume that already holds data is never touched. Because `rmdir` only removes empty directories, a `lost+found` that captured recovered data is left intact and a localized warning is emitted.
- Applied on both the **create** path and the **recovery (`Open`)** path, so volumes created before this fix get migrated (seeded) on next reattach.

Notes:
- WSLC never runs `fsck`/`e2fsck`, so `lost+found` does not reappear once removed.

## Validation Steps Performed

- Added a test `NamedVolumesVhdSeedsImageData`: creates a VHD volume, mounts it over `/etc` (guaranteed-populated by the Debian test image), and verifies the image's `/etc` content (`passwd`) is seeded while `lost+found` is absent.
- Manually reproduced the original bug (only `lost+found` visible) and confirmed the fix seeds image contents.


src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -536,6 +536,40 @@ void WSLCVirtualMachine::Ext4Format(const std::string& Device, std::optional<uin
...
+
+    THROW_HR_IF_MSG(E_FAIL, result.Code != 0, "%hs", launcher.FormatResult(result).c_str());
+
+    auto stdOut = result.Output.find(1);
+    if (stdOut == result.Output.end())
+    {
+        return {};
+    }
+
+    return wsl::shared::string::Split(stdOut->second, '\n');
```
OneBlue: Unfortunately it's possible to create files named '\n' in linux, which would break this. 

If we want to enumerate files at a given path, I think the easiest solution would be to create a new message type. Something like (WSLC_LISTDIR), and use `wsl::shared::string::ArrayFromSpan()`  to read the list on the Windows side (similar for what we do for the the environment in WSLC_EXEC for instance)

OneBlue (commented):
Change looks good, minor comment in ListDirectory's implementation

src/windows/wslcsession/WSLCVhdVolume.cpp
```diff
@@ -151,6 +176,13 @@ std::unique_ptr<WSLCVhdVolumeImpl> WSLCVhdVolumeImpl::Create(
 
     auto mountCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { VirtualMachine.Unmount(virtualMachinePath.c_str()); });
 
+    // mkfs.ext4 always creates a lost+found directory at the filesystem root,
```
benhillis, line 179: Would it be another option to not create the lost and found directory? maybe this is configurable via an option to mkfs?

OneBlue: Sadly it's not an option

benhillis:
As an alternative, could we use a subdirectory in the ext4 filesystem for the data so docker doesn't see the lost+found?

kvega005:
> As an alternative, could we use a subdirectory in the ext4 filesystem for the data so docker doesn't see the lost+found?

I brought that up to @OneBlue before I sent out the PR. It came down to, in his words, "whether we expect to ever be able to mount the VHD to another tool when the container is not running." It might be a bit of strange behavior for users.
Regardless, he said the `lost+found` folder is only useful to `fsck`, which we never run. So we do not really need to try to keep it.

OneBlue (approved):
LGTM

PR #40858, opened 2026-06-19
[codex] Fix VS Code extension install command

aj-arts:
## Summary

Fix the VS Code Remote WSL extension installation sample command in the English resource file by changing the single-dash command form to the correct `code --install-extension ms-vscode-remote.remote-wsl` syntax.

## Impact

Users reading the VS Code integration instructions see a valid `code` CLI command.

## Validation

- `xmllint --noout localization/strings/en-US/Resources.resw`
- Confirmed the final diff is limited to `localization/strings/en-US/Resources.resw` with 1 insertion and 1 deletion.


craigloewen-msft (dismissed):

OneBlue (approved):

localization/strings/en-US/Resources.resw
```diff
@@ -1776,7 +1776,7 @@ Linux binaries running in WSL can automatically use your GPU in Windows to speed
   <data name="Settings_OOBEVSCodeIntegration_VSCodeInstallSample.Text" xml:space="preserve">
     <value>After you’ve installed VS Code, you can install the Remote WSL extension from Windows Terminal:
 
-'code –install-extension ms-vscode-remote.remote-wsl'</value>
+'code --install-extension ms-vscode-remote.remote-wsl'</value>
```
OneBlue, line 1779: Now that we have two dashes, we should add a locked string for this. 

Running `python tools\devops\validate-localization.py  --fix` should take care of this

OneBlue (dismissed):

aj-arts:
Addressed this by running `python tools\devops\validate-localization.py --fix`, which added the locked string comment for `--install-extension`.

I also verified locally with `uv run python tools/devops/validate-localization.py`. Could you please rerun `/azp run` when you get a chance?

craigloewen-msft (approved):

localization/strings/en-US/Resources.resw
aj-arts: Added a locked string by running the above command.

benhillis (approved):

PR #40859, opened 2026-06-19
Support arbitrary network driver options in wslc network create

beena352:
Adds typed CLI flags (--internal, --subnet, --gateway) to wslc network create, replacing the old behavior where these were passed as generic driver options. Arbitrary driver options are forwarded verbatim to Docker via --opt key=value, matching Docker's pass-through behavior. 

## PR Checklist

Problem: Network creation options like subnet, gateway, and internal were passed as generic driver opts, with no discoverability. Arbitrary Docker driver options were silently dropped.
Changes:
- wslc.idl: Added BOOL Internal, LPCSTR Subnet, LPCSTR Gateway to WSLCNetworkOptions
- ArgumentDefinitions.h: Added --internal, --subnet, --gateway CLI arguments
- NetworkTasks.cpp / NetworkService.cpp: Map typed flags to IDL struct fields
- WSLCSession.cpp:
- Validates --gateway requires --subnet
- Forwards all --opt driver options verbatim to Docker via request.Options 
- docker_schema.h / wslc_schema.h: Added Options field to CreateNetwork and Network structs


localization/strings/en-US/Resources.resw
```diff
@@ -2331,8 +2327,8 @@ For privacy information about this product please visit https://aka.ms/privacy.<
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
   <data name="MessageWslcInvalidNetworkDriverOption" xml:space="preserve">
-    <value>Unsupported network driver option: '{}'</value>
-    <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
+    <value>Network driver option '{}' is case-sensitive. Use the exact casing: Internal, Subnet, or Gateway.</value>
```
OneBlue: How does docker the docker API react if it's passed these options with the wrong casing ? Does it just ignore them ?

beena352: Yeah, Docker just silently accepts them. I disabled our validation temporarily and tested with --opt internal=true --opt subnet=172.55.0.0/16 --opt gateway=172.55.0.1 - all lowercase. Our InspectNetwork is a direct HTTP GET to Docker's /networks/{id} endpoint (deserialized 1:1), and it shows:
Internal: false (typed field never got set)
Subnet: 172.18.0.0/16, Gateway: 172.18.0.1 (Docker picked its own defaults)
Options: {"internal": "true", "subnet": "172.55.0.0/16", "gateway": "172.55.0.1"} (stored but doing nothing)
So from Docker's side, the lowercase keys just end up as inert entries in the Options map - the network gets created but is silently broken. That's why we reject the case mismatch up front.
image

OneBlue: Ok in that case I think we should match their behavior. From a UX perspective I think the best way to this would be to implement the CLI options for these values that docker has (`--internal`, `gateway` , ...) and not to any specific casing check in the service

src/windows/wslc/commands/NetworkCreateCommand.cpp
```diff
@@ -31,6 +31,9 @@ std::vector<Argument> NetworkCreateCommand::GetArguments() const
         Argument::Create(ArgType::Driver, std::nullopt, std::nullopt, Localization::WSLCCLI_NetworkDriverOptionDescription()),
         Argument::Create(ArgType::Options, false, NO_LIMIT),
         Argument::Create(ArgType::Label, false, NO_LIMIT, Localization::WSLCCLI_NetworkLabelArgDescription()),
+        Argument::Create(ArgType::Internal),
```
dkbennett, line 35: nit: for non-positional / forward args prefer to keep these in alphabetical order

dkbennett (dismissed):
LGTM, minor non-blocking comment

dkbennett (approved):

PR #40866, opened 2026-06-22
Fix stuck systemctl poweroff could block terminate

chemwolf6922:

The `systemctl poweroff` call is async after the operation is enqueued per https://man7.org/linux/man-pages/man1/systemctl.1.html. However, in some rare cases it could block. And that will block the `wsl --terminate` call. For example: https://github.com/microsoft/WSL/pull/40433#issuecomment-4745744177
This PR moves the calling logic to a separate thread, so the fallback timeout is always started in the systemd path.

## PR Checklist

## Validation Steps Performed

No new tests are added as the original failure could not be easily reproduced.

### DIFF
--- a/src/linux/init/init.cpp
+++ b/src/linux/init/init.cpp
@@ -2753,21 +2753,14 @@ try
     {
         THROW_LAST_ERROR_IF(UtilSetSignalHandlers(g_SavedSignalActions, false) < 0);
 
+        //
+        // systemctl poweroff is normally async but can block in rare cases.
+        // Run it in a thread so a stuck invocation can't prevent the fallback timeout below.
+        //
         try
         {
             std::thread([]() {
-                //
-                // systemctl poweroff is async after the operation is successfully enqueued.
-                // However, in some rare cases, systemctl poweroff can block.
-                // Run it in a separate thread to avoid blocking the timeout.
-                //
-                int rc = UtilExecCommandLine("systemctl poweroff", nullptr);
-                if (rc < 0)
-                {
-                    LOG_ERROR("systemctl poweroff failed {}, calling reboot(RB_POWER_OFF)", rc);
-                    reboot(RB_POWER_OFF);
-                    FATAL_ERROR("reboot(RB_POWER_OFF) failed {}", errno);
-                }
+                UtilExecCommandLine("systemctl poweroff", nullptr);
             }).detach();
         }
         CATCH_LOG();
--- a/src/linux/init/init.cpp
+++ b/src/linux/init/init.cpp
@@ -2753,11 +2753,26 @@ try
     {
         THROW_LAST_ERROR_IF(UtilSetSignalHandlers(g_SavedSignalActions, false) < 0);
 
-        if (UtilExecCommandLine("systemctl poweroff", nullptr) == 0)
+        try
         {
-            std::this_thread::sleep_for(std::chrono::milliseconds(Config.BootInitTimeout));
-            LOG_ERROR("systemctl poweroff did not terminate the instance in {} ms, calling reboot(RB_POWER_OFF)", Config.BootInitTimeout);
+            std::thread([]() {
+                //
+                // systemctl poweroff is async after the operation is successfully enqueued.
+                // However, in some rare cases, systemctl poweroff can block.
+                // Run it in a separate thread to avoid blocking the timeout.
+                //
+                if (UtilExecCommandLine("systemctl poweroff", nullptr) < 0)
+                {
+                    LOG_ERROR("systemctl poweroff failed {}, calling reboot(RB_POWER_OFF)", errno);
+                    reboot(RB_POWER_OFF);
+                    FATAL_ERROR("reboot(RB_POWER_OFF) failed {}", errno);
+                }
+            }).detach();
         }
+        CATCH_LOG();
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(Config.BootInitTimeout));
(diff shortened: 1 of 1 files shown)


src/linux/init/init.cpp
```diff
@@ -2753,11 +2753,27 @@ try
     {
         THROW_LAST_ERROR_IF(UtilSetSignalHandlers(g_SavedSignalActions, false) < 0);
 
-        if (UtilExecCommandLine("systemctl poweroff", nullptr) == 0)
+        try
         {
-            std::this_thread::sleep_for(std::chrono::milliseconds(Config.BootInitTimeout));
-            LOG_ERROR("systemctl poweroff did not terminate the instance in {} ms, calling reboot(RB_POWER_OFF)", Config.BootInitTimeout);
+            std::thread([]() {
```
benhillis: Instead of a local `std::thread(...).detach()`, use the existing `UtilCreateChildProcess` helper. This will simplify the code.

```cpp
    if (Config.BootInit && !Config.BootStartWriteSocket)
    {
        THROW_LAST_ERROR_IF(UtilSetSignalHandlers(g_SavedSignalActions, false) < 0);

        //
        // systemctl poweroff is normally async but can block in rare cases. Run it in a
        // child process so a stuck invocation can't prevent the fallback timeout below.
        //
        UtilCreateChildProcess("poweroff", []() {
            if (UtilExecCommandLine("systemctl poweroff", nullptr) < 0)
            {
                LOG_ERROR("systemctl poweroff failed, calling reboot(RB_POWER_OFF)");
                reboot(RB_POWER_OFF);
            }
        });

        std::this_thread::sleep_for(std::chrono::milliseconds(Config.BootInitTimeout));
        LOG_ERROR("systemctl poweroff did not terminate the instance in {} ms, calling reboot(RB_POWER_OFF)", Config.BootInitTimeout);
    }
```

src/linux/init/init.cpp
```diff
@@ -2753,11 +2753,27 @@ try
...
+                //
+                // systemctl poweroff is async after the operation is successfully enqueued.
+                // However, in some rare cases, systemctl poweroff can block.
+                // Run it in a separate thread to avoid blocking the timeout.
+                //
+                int rc = UtilExecCommandLine("systemctl poweroff", nullptr);
+                if (rc < 0)
+                {
+                    LOG_ERROR("systemctl poweroff failed {}, calling reboot(RB_POWER_OFF)", rc);
```
benhillis: Minor: `rc` is always `-1` here (`UtilExecCommandLine` only returns 0/-1), so it adds no info, and the helper already logs the underlying error internally.

```suggestion
                    LOG_ERROR("systemctl poweroff failed, calling reboot(RB_POWER_OFF)");
```

src/linux/init/init.cpp
```diff
@@ -2753,11 +2753,27 @@ try
...
+                    reboot(RB_POWER_OFF);
+                    FATAL_ERROR("reboot(RB_POWER_OFF) failed {}", errno);
```
benhillis: This `reboot`/`FATAL_ERROR` duplicates the one at the end of the function. If you keep the thread, drop the in-thread `FATAL_ERROR` and let it fall through on failure, so the function's trailing `reboot(RB_POWER_OFF)`/`FATAL_ERROR` stays the single fallback:

```suggestion
                    reboot(RB_POWER_OFF);
```

src/linux/init/init.cpp
benhillis: Or remove this LOG_ERROR entirely.

src/linux/init/init.cpp
chemwolf6922: Do we need a new process for this?

src/linux/init/init.cpp
```diff
@@ -2753,11 +2753,20 @@ try
...
-            std::this_thread::sleep_for(std::chrono::milliseconds(Config.BootInitTimeout));
-            LOG_ERROR("systemctl poweroff did not terminate the instance in {} ms, calling reboot(RB_POWER_OFF)", Config.BootInitTimeout);
+            std::thread([]() {
+                UtilExecCommandLine("systemctl poweroff", nullptr);
+            }).detach();
         }
+        CATCH_LOG();
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(Config.BootInitTimeout));
+        LOG_ERROR("systemctl poweroff did not terminate the instance in {} ms, calling reboot(RB_POWER_OFF)", Config.BootInitTimeout);
```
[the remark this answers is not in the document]
chemwolf6922: That's exactly how the timeout got stuck before.

src/linux/init/init.cpp
```diff
@@ -2753,11 +2753,18 @@ try
...
+        // systemctl poweroff is normally async but can block in rare cases.
+        // Run it in a thread so a stuck invocation can't prevent the fallback timeout below.
+        //
+        try
         {
-            std::this_thread::sleep_for(std::chrono::milliseconds(Config.BootInitTimeout));
-            LOG_ERROR("systemctl poweroff did not terminate the instance in {} ms, calling reboot(RB_POWER_OFF)", Config.BootInitTimeout);
+            std::thread([]() { UtilExecCommandLine("systemctl poweroff", nullptr); }).detach();
         }
+        CATCH_LOG();
```
OneBlue: This try / catch doesn't actually protect us from exceptions being thrown in the thread. I think that if std::thread itself throws, we probably shouldn't catch it

chemwolf6922: Yes, it's only meant for the thread constructor. Will remove if we don't want to catch that.

OneBlue (approved):

PR #40871, opened 2026-06-22
Fix WSLC localhost port relay wedge on repeated download (reverse) transfers

kvega005:

The WSLC localhost port relay would permanently wedge after a single download-direction
transfer through a published container port. Reproduced reliably with two consecutive
`iperf3 -c 127.0.0.1 -p 5201 -t 10 -R`  (from the host) runs: the first succeeds, the second hangs ~30s
and then fails with "Connection reset by peer", and the relay stays broken until the
session is terminated (`wslc system session terminate`).

Root cause: the guest **`PortRelay`** process that owns the vsock accept loop runs
`RunLocalHostRelay` directly but never sets `signal(SIGPIPE, SIG_IGN)`. On a download
(`-R`), a relay worker bulk-writes server→host bytes into the vsock. When the host end is
reset at the end of the transfer, the next vsock write raises `SIGPIPE`, whose default
disposition **terminates the entire `PortRelay` process**, tearing down the accept loop
and its vsock `LISTEN` socket. The next connection then has no listener and blocks until
`HVSOCKET_CONNECT_TIMEOUT` (30s) before being reset.

The fix ignores `SIGPIPE` in the `PortRelay` process before entering the relay loop, so the
failure surfaces as an `EPIPE` return from the write (already handled gracefully by the relay
loop, which just exits that worker thread) instead of killing the process. This mirrors the
existing WSL2 path (`StartLocalhostRelay` in `src/linux/init/localhost.cpp`), which has always
ignored `SIGPIPE` for the same reason.

## PR Checklist

- [x] **Documentation updated:** N/A


src/linux/init/WSLCInit.cpp
```diff
@@ -781,6 +781,12 @@ void HandleMessageImpl(
     Transaction.SendResultMessage<uint32_t>(SocketAddress.svm_port);
     Channel.Close();
     UtilSetThreadName("PortRelay");
+
+    // If the host end of a relay socket is reset, a write will raise SIGPIPE. Ignore it so
+    // the failure surfaces as an EPIPE return value (handled by the relay loop) instead of
+    // terminating this forked PortRelay process and tearing down the vsock accept listener.
+    THROW_LAST_ERROR_IF(signal(SIGPIPE, SIG_IGN) == SIG_ERR);
```
benhillis, line 788: This is worth fixing but this code is likely to be removed soon right @oneblue?

benhillis:
copilot: review

src/linux/init/WSLCInit.cpp
OneBlue: I think we'll remove it at some point, but given how new the virtionet port mapping code is, I think this will be worth keeping a bit longer

OneBlue (approved):

PR #40885, opened 2026-06-23
Implement a new yaml setting for users to configure the default port binding address

OneBlue:

This change adds one new yaml setting: `session.defaultBindingAddress`, which control the default container port binding address for the CLI session. These are used when `-p` is used without an explicit address (defaults to localhost).

This change also solves an inspect bug where the binding address was always "127.0.0.1"

## PR Checklist

### DIFF
--- a/src/windows/common/WSLCUserSettings.cpp
+++ b/src/windows/common/WSLCUserSettings.cpp
@@ -48,10 +48,8 @@ static constexpr std::string_view s_DefaultSettingsTemplate =
     "  # maxStorageSize: default\n"
     "\n"
     "  # Default host address that published ports bind to when 'container run -p' is\n"
-    "  # used without an explicit address (default: 127.0.0.1 for IPv4, ::1 for IPv6)\n"
-    "  # defaultBindingAddress:\n"
-    "  #   ipv4: default\n"
-    "  #   ipv6: default\n"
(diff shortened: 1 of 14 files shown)


benhillis (dismissed):

src/windows/common/WSLCUserSettings.cpp
```diff
@@ -47,6 +47,10 @@ static constexpr std::string_view s_DefaultSettingsTemplate =
     "  # Maximum disk image size (e.g. 500GB default: 1TB)\n"
     "  # maxStorageSize: default\n"
     "\n"
+    "  # Default host address that published ports bind to when 'container run -p' is\n"
+    "  # used without an explicit address (default: 127.0.0.1)\n"
+    "  # defaultBindingAddress: default\n"
```
yao-msft, line 52: nit: "defaultBindingAddress: default" seems a bit redundant

yao-msft (approved):

benhillis (approved):

src/windows/common/WSLCUserSettings.cpp
OneBlue: It kinda does, however both `defaults` mean something different here (the default value for a container binding address), and the default value for that setting. I'm open to other suggestions though

PR #40890, opened 2026-06-24
Add hybrid CRT link for Windows binaries

JohnMcPMS:
Configures hybrid static linkage for Windows binaries in the build.  See https://aka.ms/hybridcrt for details on this model.

## PR Checklist

The Windows binaries are all already statically linking the runtime.  Switching to the hybrid model will cut non-trivial size from every binary, without adding any new dependency requirements.

I also understand if we want to wait on taking this until after a significant release so it may have more time to stabilize.

## Validation Steps Performed
Built debug locally, ran an arbitrary test.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -328,6 +328,15 @@ if (${TARGET_PLATFORM} STREQUAL "x64")
     set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} /CETCOMPAT")
 endif()
 
+# Hybrid CRT: static-link the C++ runtime and STL (/MT|/MTd from CMAKE_MSVC_RUNTIME_LIBRARY),
+# but dynamic-link the Universal CRT (ucrt.dll) which is an OS component on Windows 10+.
+# This eliminates the VCRUNTIME/MSVCP redistributable dependency while avoiding the size
+# cost of statically linking the UCRT. See https://aka.ms/hybridcrt for details.
+set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "${CMAKE_SHARED_LINKER_FLAGS_RELEASE} /NODEFAULTLIB:libucrt.lib /defaultlib:ucrt.lib")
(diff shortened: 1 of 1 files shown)


OneBlue (dismissed):
I think this is a great change, and we should definitely merge it. 

Agreed that we should wait after the initial release though. One thing that we should validate before merging this is whether this causes us any issues on WS 2022 (specifically on the ServerCore SKU), since we've had runtime dependencies issues in this SKU in the past.

If this runs without issues on WS 2022 ServerCore, I think we should merge it (after the 2.9 release).

Blocking for now, but will unblock after 2.9 is out.

JohnMcPMS:
There is certainly a reduction in size when compared to another build from the same time, but it isn't as much as I would have hoped for.

CMakeLists.txt
```diff
@@ -328,6 +328,17 @@ if (${TARGET_PLATFORM} STREQUAL "x64")
     set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} /CETCOMPAT")
 endif()
 
+# Hybrid CRT: static-link the C++ runtime and STL (/MT|/MTd from CMAKE_MSVC_RUNTIME_LIBRARY),
+# but dynamic-link the Universal CRT (ucrt.dll) which is an OS component on Windows 10+.
+# This eliminates the VCRUNTIME/MSVCP redistributable dependency while avoiding the size
+# cost of statically linking the UCRT. See https://aka.ms/hybridcrt for details.
+foreach(config RELEASE RELWITHDEBINFO MINSIZEREL)
```
OneBlue: nit: We don't support RELWITHDEBINFO and MINSIZEREL, so we can probably remove the loop here.

OneBlue: Actually we can probably just write to `CMAKE_SHARED_LINKER_FLAGS` and `CMAKE_EXE_LINKER_FLAGS` here

OneBlue: Also, should we apply this to `STATIC_LIBRARY_FLAGS` as well ?

OneBlue (approved):
LGTM, please make sure that this runs on the ServerCore SKU of Windows Server 2019 before merging. 

Just validating that a simple container starts should be sufficient

JohnMcPMS:
> LGTM, please make sure that this runs on the ServerCore SKU of Windows Server 2019 before merging.
> 
> Just validating that a simple container starts should be sufficient

Validated both install and image pull on Windows ServerCore 2022.

PR #40894, opened 2026-06-24
Add test to repeatedly use session storage

JohnMcPMS:
Adds a test that repeatedly creates a session using the same storage.

## PR Checklist

This test is intended to detect any lingering resources being held by a session after the SDK terminate/release calls.  If the session cannot be started immediately afterward, something is being held open longer than it should or not properly synchronized against during the terminate and release of the session handle.

## Validation Steps Performed
The new test.

### DIFF
--- a/test/windows/WslcSdkTests.cpp
+++ b/test/windows/WslcSdkTests.cpp
@@ -2765,6 +2765,11 @@ class WslcSdkTests
         // without errors.
         std::filesystem::path sessionStorage = m_storagePath / "wslc-resource-reuse-storage";
 
+        auto cleanupStorage = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
+            std::error_code ec;
+            std::filesystem::remove_all(sessionStorage, ec);
+        });
+
         WslcSessionSettings sessionSettings;
         VERIFY_SUCCEEDED(WslcInitSessionSettings(L"wslc-resource-reuse", sessionStorage.c_str(), &sessionSettings));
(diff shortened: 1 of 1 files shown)


test/windows/WslcSdkTests.cpp
```diff
@@ -2756,6 +2756,53 @@ class WslcSdkTests
...
+            VERIFY_SUCCEEDED(WslcLoadSessionImageFromFile(session.get(), GetTestImagePath(c_imageName).c_str(), nullptr, nullptr));
+
+            // Run a container using the image's default entrypoint and capture its output.
+            auto output = RunContainerAndCapture(session.get(), c_imageName, {});
+            VERIFY_IS_TRUE(output.stdoutOutput.find("Hello from Docker!") != std::string::npos);
+
+            // Terminate the session, then release it. WslcReleaseSession must return only after all
+            // resources are freed so that the next iteration can reopen the same storage without error.
+            VERIFY_SUCCEEDED(WslcTerminateSession(session.get()));
+            VERIFY_SUCCEEDED(WslcReleaseSession(session.get()));
```
OneBlue, line 2806: nit: the manual release shouldn't be needed since `UniqueSession` does that

JohnMcPMS: Since the test is explicitly about ensuring proper session termination timing, I thought it better to require success for these (the session RAII does not enforce that).

OneBlue: That's fair. We could also add checks to `CloseSession()` to fix this, but not blocking for this PR

OneBlue (approved):

PR #40895, opened 2026-06-24
wslc: Grant VM access to a caller-supplied root VHD override as the user

benhillis:
## Summary

The WSLC boot path attached the root VHD to the VM without granting VM access. For a caller-supplied `RootVhdOverride`, that meant the VM worker process opened the file with no check that the user actually has access to it.

This grants VM access while impersonating the user (mirroring the WSL2 boot path in `WslCoreVm::GenerateConfigJson`). The grant requires the user to have access to the target, so an override can only ever point at a VHD the user can already reach. The grant is tracked via `DiskInfo::AccessGranted` and revoked on teardown.

The install-dir default root/modules VHDs are already reachable by the VM worker process and are left unchanged.

### DIFF
--- a/src/windows/service/exe/HcsVirtualMachine.cpp
+++ b/src/windows/service/exe/HcsVirtualMachine.cpp
@@ -259,7 +259,7 @@ HcsVirtualMachine::HcsVirtualMachine(_In_ const WSLCSessionSettings* Settings)
 
     // Setup boot VHDs
     hcs::Scsi scsiController{};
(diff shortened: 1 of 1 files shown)


OneBlue (approved):
LGTM. We can probably start considering getting this of the "root VHD override" setting relatively soon

benhillis:
> LGTM. We can probably start considering getting this of the "root VHD override" setting relatively soon

Agreed, I considered doing that now but might as well plug this hole first.

PR #40908, opened 2026-06-25
WSLC SDK install fallback uses repair

JohnMcPMS:
Use repair semantics in the temporary WSLC SDK install API fallback to GitHub.

## PR Checklist

Switch to repair semantics for this fallback path to prevent the SDK's version from being used as the current WSL install version:
https://github.com/microsoft/WSL/blob/f0ba469c5ce99424d57c0084e08421c3d4499cb0/src/windows/common/install.cpp#L89

Also passes along the "we don't own the process" flag to `DownloadFile` to prevent progress output.

## Validation Steps Performed
Manually confirmed that the latest GH release is installed when no DCAT is available.

### DIFF
--- a/src/windows/WslcSDK/wslcsdk.cpp
+++ b/src/windows/WslcSDK/wslcsdk.cpp
@@ -1731,7 +1731,8 @@ try
             {
                 callback(0);
             }
-            wsl::windows::common::install::UpdatePackage(true, false, false);
+            // Use pre-release builds and repair semantics (required since this function uses the SDK binary version as a filter).
+            wsl::windows::common::install::UpdatePackage(true, true, false);
             if (callback)
             {
                 callback(100);
--- a/src/windows/common/install.cpp
+++ b/src/windows/common/install.cpp
@@ -101,7 +101,7 @@ int UpdatePackageImpl(bool preRelease, bool repair, bool callerOwnsProcess)
     }
 
     const bool msiInstall = wsl::shared::string::EndsWith<wchar_t>(release.name, L".msi");
-    const auto downloadPath = DownloadFile(release.url, release.name);
+    const auto downloadPath = DownloadFile(release.url, release.name, callerOwnsProcess);
     if (msiInstall)
     {
         auto logFile = std::filesystem::temp_directory_path() / L"wsl-install-logs.txt";
--- a/src/windows/common/wslutil.cpp
+++ b/src/windows/common/wslutil.cpp
@@ -348,15 +348,23 @@ GUID wsl::windows::common::wslutil::CreateV5Uuid(const GUID& namespaceGuid, cons
     return EndianSwap(newGuid);
 }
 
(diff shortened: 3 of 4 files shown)


src/windows/WslcSDK/wslcsdk.cpp
```diff
@@ -1731,7 +1731,8 @@ try
             {
                 callback(0);
             }
-            wsl::windows::common::install::UpdatePackage(true, false, false);
+            // Use pre-release builds and repair semantics (required since this function uses the SDK binary version as a filter).
+            wsl::windows::common::install::UpdatePackage(true, true, false);
```
OneBlue, line 1735: Could we add test coverage for this ? We could do something similar to [this test](https://github.com/microsoft/WSL/blob/f0ba469c5ce99424d57c0084e08421c3d4499cb0/test/windows/InstallerTests.cpp#L697) and validate that "a package" is installed (either via DCAT, or injected via the GH URL override).

OneBlue (dismissed):

src/windows/WslcSDK/wslcsdk.cpp
JohnMcPMS: For DCAT, we should be able to expect that short of publishing a version incongruent with its actual build number, no available package is newer than anywhere a test is running.  However, the change here to force repair means that the code will attempt to install whatever latest version is presented to it.  I suppose that an option for the test is to:
1. expect no DCAT updates, but at least it ran the search
2. return a url to the local server that doesn't exist
3. have the test expect the install attempt that fails with a 404-type error

OneBlue: We could also return a local server URL that actually serves package. If we uninstall it before running the test, we can validate that it was reinstalled once complete

JohnMcPMS: Although on top of all of that, the API isn't easily testable as we can't request what should be installed.  If it can create a session manager, it won't attempt to update.  Honestly that seems like a righteous change to the API, as currently you can:
- Detect missing components (`WslcGetMissingComponents`)
- Install missing components (`WslcInstallWithDependencies`)
- Determine the installed WSL version (`WslcGetVersion`)

But you cannot attempt to update WSL after you have ensured that it is present but found that the installed version is too low for your scenario (3.0.0 adds support for feature X that you need, but 2.9.2 is installed).

Adding a `WslcComponentFlags` parameter to `WslcInstallWithDependencies` would allow that. The caller would decide what should be attempted, not us with an effective repeat of `WslcGetMissingComponents`. I think it would be wise then to also add an options parameter to provide the repair semantics (and potentially more later), as right now we force a DCAT update by overwriting the version.

On top of allowing all of that, it makes it much easier to test as we can directly request that code path be exercised rather than forcing a failed attempt to create a session manager.

JohnMcPMS: Followup PR for no-update test and API changes.

OneBlue (approved):

PR #40913, opened 2026-06-25
Collect WSLg logs and crash dumps in collect-wsl-logs.ps1

benhillis:
## Summary

Adds WSLg (graphical and audio application) log collection to `collect-wsl-logs.ps1`, so a standard WSL log collection also captures what's needed to diagnose WSLg issues, and documents how to read those logs.

## Changes

- **`diagnostics/collect-wsl-logs.ps1`**: collect WSLg logs into a `wslg/` folder of the archive:
  - `weston.log`, `pulseaudio.log`, `wlog.log`, `stderr.log`, `versions.txt` from `/mnt/wslg`.
  - Files are copied with `wsl.exe --system --user root`, which reaches `/mnt/wslg` even when the default distro is WSL1 or isn't running, and can read root-only logs like `pulseaudio.log`. The destination is passed to `sh` as `$1`, so paths containing a single quote are handled safely.
  - Collection runs in a background job with a 20s timeout, so a wedged WSL service can't hang log collection (it warns and skips on timeout).
  - WSLg crash dumps (`/mnt/wslg/dumps` and host `%TEMP%\wsl-crashes`, e.g. `core.weston`) are collected only when `-Dump` is passed, since users may not expect dumps to be published by default.
- **`.github/copilot/wslg-logs.md`** (new): a log-analysis guide covering the WSLg architecture, the file layout, and common log signatures (system-distro cycling, `glamor` to software-rendering fallback, RDP audio sink, FontMonitor), with references to the [microsoft/wslg](https://github.com/microsoft/wslg) repo.
- **`.github/copilot-instructions.md`**: Log Analysis Tools now points to the WSLg guide and notes WSLg code lives in microsoft/wslg.
- **`CONTRIBUTING.md`**: note that the same script covers graphical-app (WSLg) issues.
- **`doc/docs/debugging.md`**: new "WSLg logs" section with the file list, dump locations, and microsoft/wslg references.

## Testing

- Ran the full `collect-wsl-logs.ps1` end to end and confirmed the `wslg/` folder is present and populated in the resulting archive.
- Verified `--system --user root` reads the root-owned `0600` `pulseaudio.log` and copies all logs to the host folder.
- Verified the timeout path skips cleanly without hanging, and that `-Dump` gates crash-dump collection.
- `collect-wsl-logs.ps1` parses; `mkdocs build` succeeds.

### DIFF
--- a/diagnostics/collect-wsl-logs.ps1
+++ b/diagnostics/collect-wsl-logs.ps1
@@ -345,25 +345,43 @@ if ($LogProfile -eq "networking")
 $wslgFolder = "$folder/wslg"
 mkdir -p $wslgFolder | Out-Null
 
-# Resolve the destination as a WSL path so files (including binary crash dumps) can be copied inside WSL.
-$wslgFolderFull = (Resolve-Path $wslgFolder).Path
-$wslgFolderWsl = $null
-try { $wslgFolderWsl = (& wsl.exe -e wslpath -u "$wslgFolderFull" 2>$null).Trim() } catch {}
+# Run in a job with a timeout so a wedged WSL service can't hang collection. --system reaches
+# /mnt/wslg as the wslg user even when the default distro is WSL1 or isn't running.
+$wslgJob = Start-Job -ScriptBlock {
+    param($DestFull, $CollectDumps)
 
-if (-not [string]::IsNullOrWhiteSpace($wslgFolderWsl))
+    $destWsl = & wsl.exe --system -e wslpath -u "$DestFull" 2>$null
+    if ([string]::IsNullOrWhiteSpace($destWsl)) { return }
+    $destWsl = $destWsl.Trim()
+
+    # Destination is passed as $1 so paths containing a single quote are handled safely.
+    & wsl.exe --system -e sh -c 'cp /mnt/wslg/pulseaudio.log /mnt/wslg/weston.log /mnt/wslg/wlog.log /mnt/wslg/stderr.log /mnt/wslg/versions.txt "$1/" 2>/dev/null; exit 0' sh "$destWsl"
+
+    if ($CollectDumps)
+    {
+        & wsl.exe --system -e sh -c '[ -d /mnt/wslg/dumps ] && cp -r /mnt/wslg/dumps "$1/dumps"; exit 0' sh "$destWsl"
+    }
+} -ArgumentList (Resolve-Path $wslgFolder).Path, ([bool]$Dump)
+
+if (Wait-Job $wslgJob -Timeout 60)
 {
-    # Run as the super user (uid=0) so root-only logs such as pulseaudio.log are readable.
-    $wslgSuperUser = & wsl.exe -- id -nu 0
-    # pulseaudio.log, weston.log and stderr.log plus the legacy /mnt/wslg/dumps crash dumps.
-    # The destination is passed as $1 so paths containing a single quote are handled safely.
-    & wsl.exe -u $wslgSuperUser -e sh -c 'cp /mnt/wslg/pulseaudio.log /mnt/wslg/weston.log /mnt/wslg/wlog.log /mnt/wslg/stderr.log /mnt/wslg/versions.txt "$1/" 2>/dev/null; [ -d /mnt/wslg/dumps ] && cp -r /mnt/wslg/dumps "$1/dumps"; exit 0' sh "$wslgFolderWsl" 2>&1 | Out-Null
+    Receive-Job $wslgJob | Out-Null
 }
-
-# Newer builds write WSLg crash dumps (e.g. core.weston) to %TEMP%\wsl-crashes on the host.
-$wslCrashes = "$env:TEMP\wsl-crashes"
-if (Test-Path $wslCrashes)
+else
 {
-    Copy-Item $wslCrashes "$wslgFolder/wsl-crashes" -Recurse -ErrorAction Ignore
+    Write-Host -ForegroundColor Yellow "WSLg log collection timed out and was skipped."
+    Stop-Job $wslgJob
+}
+Remove-Job $wslgJob -Force
+
+# Crash dumps are only collected with -Dump, since users may not expect dumps to be published by default.
+if ($Dump)
+{
+    $wslCrashes = "$env:TEMP\wsl-crashes"
+    if (Test-Path $wslCrashes)
+    {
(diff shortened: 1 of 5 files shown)


diagnostics/collect-wsl-logs.ps1
```diff
@@ -341,6 +341,31 @@ if ($LogProfile -eq "networking")
...
+
+# Resolve the destination as a WSL path so files (including binary crash dumps) can be copied inside WSL.
+$wslgFolderFull = (Resolve-Path $wslgFolder).Path
+$wslgFolderWsl = $null
+try { $wslgFolderWsl = (& wsl.exe -e wslpath -u "$wslgFolderFull" 2>$null).Trim() } catch {}
+
+if (-not [string]::IsNullOrWhiteSpace($wslgFolderWsl))
+{
+    # Run as the super user (uid=0) so root-only logs such as pulseaudio.log are readable.
+    $wslgSuperUser = & wsl.exe -- id -nu 0
```
benhillis: `wsl.exe --user root` will also work

diagnostics/collect-wsl-logs.ps1
```diff
@@ -341,6 +341,31 @@ if ($LogProfile -eq "networking")
...
 }
 
+# Collect WSLg logs (https://github.com/microsoft/wslg)
+$wslgFolder = "$folder/wslg"
+mkdir -p $wslgFolder | Out-Null
+
+# Resolve the destination as a WSL path so files (including binary crash dumps) can be copied inside WSL.
+$wslgFolderFull = (Resolve-Path $wslgFolder).Path
+$wslgFolderWsl = $null
+try { $wslgFolderWsl = (& wsl.exe -e wslpath -u "$wslgFolderFull" 2>$null).Trim() } catch {}
```
OneBlue: I don't think we should unconditionally do this since will hang is the service is in a bad state. We should either have a timeout, or an opt-in switch to collect wslg logs

diagnostics/collect-wsl-logs.ps1
OneBlue: I'd also recommend --system so this doesn't fail if the default distro is a wsl1 distro

diagnostics/collect-wsl-logs.ps1
```diff
@@ -341,6 +341,31 @@ if ($LogProfile -eq "networking")
...
+{
+    # Run as the super user (uid=0) so root-only logs such as pulseaudio.log are readable.
+    $wslgSuperUser = & wsl.exe -- id -nu 0
+    # pulseaudio.log, weston.log and stderr.log plus the legacy /mnt/wslg/dumps crash dumps.
+    # The destination is passed as $1 so paths containing a single quote are handled safely.
+    & wsl.exe -u $wslgSuperUser -e sh -c 'cp /mnt/wslg/pulseaudio.log /mnt/wslg/weston.log /mnt/wslg/wlog.log /mnt/wslg/stderr.log /mnt/wslg/versions.txt "$1/" 2>/dev/null; [ -d /mnt/wslg/dumps ] && cp -r /mnt/wslg/dumps "$1/dumps"; exit 0' sh "$wslgFolderWsl" 2>&1 | Out-Null
+}
+
+# Newer builds write WSLg crash dumps (e.g. core.weston) to %TEMP%\wsl-crashes on the host.
+$wslCrashes = "$env:TEMP\wsl-crashes"
```
OneBlue: I recommend only doing this if `-Dump` is passed (since users might not expect dumps to be published by default)

diagnostics/collect-wsl-logs.ps1
benhillis: Done in 9c66ebea. The WSLg collection now runs in a background job with a 60s timeout (Wait-Job -Timeout), so a wedged service can no longer hang collection; on timeout it logs a warning and skips. Kept it default-on (matching the original goal) rather than opt-in.

benhillis: added a timeout

diagnostics/collect-wsl-logs.ps1
```diff
@@ -341,6 +341,49 @@ if ($LogProfile -eq "networking")
...
+$wslgFolder = "$folder/wslg"
+mkdir -p $wslgFolder | Out-Null
+
+# Run in a job with a timeout so a wedged WSL service can't hang collection. --system --user root
+# reaches /mnt/wslg even when the default distro is WSL1 or isn't running, and can read root-only logs.
+$wslgJob = Start-Job -ScriptBlock {
+    param($DestFull, $CollectDumps)
+
+    $destWsl = & wsl.exe --system --user root -e wslpath -u "$DestFull" 2>$null
+    if ([string]::IsNullOrWhiteSpace($destWsl)) { return }
```
OneBlue, line 354: nit: We should probably swap and do the Trim() before the null check

diagnostics/collect-wsl-logs.ps1
```diff
@@ -341,6 +341,49 @@ if ($LogProfile -eq "networking")
...
+    $destWsl = $destWsl.Trim()
+
+    # Destination is passed as $1 so paths containing a single quote are handled safely.
+    & wsl.exe --system --user root -e sh -c 'cp /mnt/wslg/pulseaudio.log /mnt/wslg/weston.log /mnt/wslg/wlog.log /mnt/wslg/stderr.log /mnt/wslg/versions.txt "$1/" 2>/dev/null; exit 0' sh "$destWsl"
```
OneBlue, line 359: I'm a bit confused by this, what does the `sh "$destWsl"` do at the end ?

OneBlue: Is this an edit error maybe ?

diagnostics/collect-wsl-logs.ps1
```diff
@@ -341,6 +341,49 @@ if ($LogProfile -eq "networking")
...
+
+    if ($CollectDumps)
+    {
+        & wsl.exe --system --user root -e sh -c '[ -d /mnt/wslg/dumps ] && cp -r /mnt/wslg/dumps "$1/dumps"; exit 0' sh "$destWsl"
+    }
+} -ArgumentList (Resolve-Path $wslgFolder).Path, ([bool]$Dump)
+
+if (Wait-Job $wslgJob -Timeout 60)
```
OneBlue: 60 seconds feels like a lot for this, I'd recommend something like maybe 20 seconds

OneBlue (approved):

PR #40916, opened 2026-06-25
Notice change from build: 150742739

OneBlue:


beena352:
Copilot's observations are valid but NOTICE.txt is auto-generated by the notice@0 pipeline task, hand-edits here would be overwritten on the next regen. The line-wrapping and separator spacing come from the upstream manual NOTICE-mesa.txt in WSLg (and from how notice@0 concatenates additionaldata). If we want to fix wrapping it belongs in the WSLg manual file, not in this auto-PR. Resolving these threads as not actionable.

beena352 (approved):

PR #40921, opened 2026-06-25
CLI: Fix hang with exec/run -i with self-exiting command and open stdin

dkbennett:
`wslc exec -i <container> echo hello`  didn't return until the client's stdin was closed, while  docker exec -i   returns as soon as the process exits. The hang is entirely client‑side.

When wslc's stdin is a pipe (not a console), the stdin relay runs on a worker thread blocked in a synchronous   ReadFile . That read can't be interrupted by the relay's exit event. So when the remote process exits and teardown  tries to  join()  the worker, the join blocks until stdin is closed externally — the hang.

**The fix**

 A shared teardown helper,  InterruptAndJoinInputThread  (used by both the TTY and non‑TTY relay paths), now signals  the exit event and then calls  CancelSynchronousIo  on the worker (retried until it exits) before joining — which  aborts the otherwise‑uninterruptible read. A matching one‑line change in  relay.cpp  treats the resulting   ERROR_OPERATION_ABORTED  as EOF so the relay loop ends cleanly. Termination is guaranteed: every state the worker  can block in is broken by either the cancel or the already‑set exit event.

Two regression tests are also added which exercise this hang and I confirmed these tests fail without the fix, and pass with the fix.

## PR Checklist

## Validation Steps Performed
**Tests**

 Two regression tests run a self‑exiting command ( echo hello ) and assert the process exits before stdin is closed  — one for  exec -i , one for  exec -it , since the two paths have separate teardown logic that could regress  independently. Both detect the symptom (process must exit within the timeout), so any teardown hang fails the test.

These tests fail without the fix and pass with the fix.

### DIFF
--- a/src/windows/common/relay.cpp
+++ b/src/windows/common/relay.cpp
@@ -133,7 +133,7 @@ wsl::windows::common::relay::InterruptableRead(
     if (!ReadFile(InputHandle, Buffer.data(), gsl::narrow_cast<DWORD>(Buffer.size()), &bytesRead, Overlapped))
     {
         auto lastError = GetLastError();
-        if ((lastError == ERROR_HANDLE_EOF) || (lastError == ERROR_BROKEN_PIPE))
+        if ((lastError == ERROR_HANDLE_EOF) || (lastError == ERROR_BROKEN_PIPE) || (lastError == ERROR_OPERATION_ABORTED))
         {
             return 0;
         }
--- a/src/windows/wslc/services/ConsoleService.cpp
+++ b/src/windows/wslc/services/ConsoleService.cpp
@@ -39,7 +39,7 @@ namespace {
             return;
         }
 
-        FAIL_FAST_IF(!exitEvent);
+        WI_ASSERT(exitEvent);
         exitEvent.SetEvent();
 
         // Overlapped IO will get terminated by SetEvent(). Synchronous IO will not, so we need to cancel it.
--- a/src/windows/wslc/services/ConsoleService.cpp
+++ b/src/windows/wslc/services/ConsoleService.cpp
@@ -24,6 +24,49 @@ using wsl::windows::common::io::ReadConsoleHandle;
 using wsl::windows::common::io::ReadHandle;
 using wsl::windows::common::io::RelayHandle;
 
+namespace {
+
+    // Interrupts and joins the stdin-relay worker thread at teardown.
+    //
+    // A worker thread is only created when stdin is not a character device (any non-FILE_TYPE_CHAR handle,
+    // e.g. a redirected pipe); the console (FILE_TYPE_CHAR) and detach paths create none, so this no-ops
+    // (not joinable) for them. When that handle is a non-overlapped (synchronous) pipe the worker blocks
+    // in a ReadFile() that neither the exit event nor CancelIoEx() can interrupt, so once the relay's
+    // scope-exit runs (normally after io.Run() returns) the join() below would hang until stdin is closed
+    // -- the bug this guards against. The exit event is signaled first (enough for handles that support
+    // overlapped IO), then CancelSynchronousIo() aborts the synchronous read the event cannot reach.
+    void InterruptAndJoinInputThread(std::thread& inputThread, wil::unique_event& exitEvent)
+    {
+        if (!inputThread.joinable())
+        {
+            return;
+        }
+
+        if (exitEvent)
+        {
+            exitEvent.SetEvent();
+        }
+
+        // CancelSynchronousIo() only aborts I/O in flight on the target thread at the instant it is called,
+        // so a single call can miss the worker while it is between read iterations. Cancel immediately to
+        // unblock a worker already parked in ReadFile, then retry every 50ms until the thread exits;
+        // WaitForSingleObject is both the termination test and the backoff. ERROR_NOT_FOUND means nothing
+        // was in flight to cancel, which is expected, so only unexpected failures are logged.
+        const auto threadHandle = static_cast<HANDLE>(inputThread.native_handle());
+        do
+        {
+            if (!CancelSynchronousIo(threadHandle))
+            {
+                const auto lastError = GetLastError();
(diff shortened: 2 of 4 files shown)


src/windows/wslc/services/ConsoleService.cpp
```diff
@@ -24,6 +24,49 @@ using wsl::windows::common::io::ReadConsoleHandle;
...
+    // -- the bug this guards against. The exit event is signaled first (enough for handles that support
+    // overlapped IO), then CancelSynchronousIo() aborts the synchronous read the event cannot reach.
+    void InterruptAndJoinInputThread(std::thread& inputThread, wil::unique_event& exitEvent)
+    {
+        if (!inputThread.joinable())
+        {
+            return;
+        }
+
+        if (exitEvent)
```
OneBlue: I believe `exitEvent` should always be set if we hit this block

src/windows/common/relay.cpp
```diff
@@ -133,7 +133,7 @@ wsl::windows::common::relay::InterruptableRead(
     if (!ReadFile(InputHandle, Buffer.data(), gsl::narrow_cast<DWORD>(Buffer.size()), &bytesRead, Overlapped))
     {
         auto lastError = GetLastError();
-        if ((lastError == ERROR_HANDLE_EOF) || (lastError == ERROR_BROKEN_PIPE))
+        if ((lastError == ERROR_HANDLE_EOF) || (lastError == ERROR_BROKEN_PIPE) || (lastError == ERROR_OPERATION_ABORTED))
```
OneBlue, line 136: Hmm I'm not sure that we should catch this error all the time here since that might have adverse effects in other places. Instead I recommend using an error handler for the stdin relay like we do [here](https://github.com/microsoft/WSL/blob/4f052df246a063ad7190c6367327beab2a845905/src/windows/wslcsession/WSLCSession.cpp#L962-L968) so we can ignore the error only in the place we need

src/windows/wslc/services/ConsoleService.cpp
```diff
@@ -24,6 +24,49 @@ using wsl::windows::common::io::ReadConsoleHandle;
...
+        // unblock a worker already parked in ReadFile, then retry every 50ms until the thread exits;
+        // WaitForSingleObject is both the termination test and the backoff. ERROR_NOT_FOUND means nothing
+        // was in flight to cancel, which is expected, so only unexpected failures are logged.
+        const auto threadHandle = static_cast<HANDLE>(inputThread.native_handle());
+        do
+        {
+            if (!CancelSynchronousIo(threadHandle))
+            {
+                const auto lastError = GetLastError();
+                LOG_HR_IF(HRESULT_FROM_WIN32(lastError), lastError != ERROR_NOT_FOUND);
```
OneBlue: We should throw if we get anything other than `ERROR_NOT_FOUND` here

src/windows/wslc/services/ConsoleService.cpp
```diff
@@ -24,6 +24,49 @@ using wsl::windows::common::io::ReadConsoleHandle;
...
+            }
+        } while (WaitForSingleObject(threadHandle, 50) == WAIT_TIMEOUT);
```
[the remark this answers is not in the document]
OneBlue: Agreed with this

src/windows/common/relay.cpp
OneBlue: Ah wait this is synchronous IO.  In that this this is probably OK

src/windows/wslc/services/ConsoleService.cpp
dkbennett: It should at present yes, but it could be not set by some change in the future. I could remove the if but I would want to replace it with a THROW_HR_IF so we catch any changes in the future where that is not the case. Edit: Or a failfast might be more appropriate since it is a programming error we want to catch quickly.

OneBlue: For something that's expected to be here I think an assert would be appropriate. If this ever changes in the future, the assert will fire and it should be easy to diagnose

dkbennett: Changed to assert

src/windows/wslc/services/ConsoleService.cpp
```diff
@@ -24,6 +24,49 @@ using wsl::windows::common::io::ReadConsoleHandle;
...
+
+            waitResult = WaitForSingleObject(threadHandle, 50);
+        }
+
+        // The loop only breaks on a non-timeout result; anything but WAIT_OBJECT_0 (e.g. WAIT_FAILED) is an
+        // error -- surface it instead of join()ing on a thread we can't confirm has exited.
+        THROW_LAST_ERROR_IF(waitResult != WAIT_OBJECT_0);
+
+        inputThread.join();
+    }
```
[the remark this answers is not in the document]
dkbennett: Only real thing to do here is change to a failfast but its going to crash either way. It will not "mask the original failure" because the scope exit log will log the failure first, so this is really not a "high" priority issue, it's just a discussion about whether we want a failfast crash or an abort() crash w/ a log message.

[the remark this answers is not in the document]
dkbennett: Changed to failfasts.

src/windows/wslc/services/ConsoleService.cpp
```diff
@@ -24,6 +24,50 @@ using wsl::windows::common::io::ReadConsoleHandle;
...
+                // ERROR_NOT_FOUND means nothing to cancel; any other error is a corrupt handle that shouldn't happen.
+                const auto cancelError = GetLastError();
+                if (cancelError != ERROR_NOT_FOUND)
+                {
+                    FAIL_FAST_WIN32(cancelError);
+                }
+            }
+
+            waitResult = WaitForSingleObject(threadHandle, 50);
+        }
```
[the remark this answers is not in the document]
dkbennett: This will eventually terminate, I do not want random failfast noise because some runs took a bit longer than others.

OneBlue (approved):

PR #40922, opened 2026-06-26
Added WSLc API docs 

craigloewen-msft:
This PR adds functionality for the API docs to `wsl.dev`. 

It looks like a lot of files but really there are only 5 files which have code changes: 
- .github/workflows/documentation.yml
- doc/mkdocs.yml
- doc/requirements.txt

And here are the summary of the major changes:
- Added API docs content


.github/workflows/documentation.yml
```diff
@@ -3,6 +3,7 @@ on:
   workflow_dispatch:
```
OneBlue, line 3: Doing some initial testing, it's looking the architecture graph doesn't render anymore, this is what I see: 

Image

.github/workflows/documentation.yml
```diff
@@ -21,19 +22,22 @@ jobs:
...
 
       - name: Install packages
-        run: pip install mkdocs-mermaid2-plugin mkdocs --break-system-packages
+        run: pip install -r doc/requirements.txt --break-system-packages
         shell: bash
 
-      - name: Build documentation
-        run:  mkdocs build -f doc/mkdocs.yml
+      - name: Build versioned documentation
+        run: python doc/build_versioned_docs.py --output public
```
OneBlue: I don't love the idea of having a custom script to build the documentation. I just tried it locally and `mkdocs serve` doesn't seem to serve the wslc content (unless I missed something), which makes testing local changes harder

Ideally i'd prefer to keep this under a standard mkdocs build, so this is easy to maintain.

We also probably shouldn't rebuild the entire documentation for every single tag, if we wanted to have multiple versions, I think having something that says "This method was introduced in version X" would be sufficient. 

If we absolutely want people to be able to read the doc from older versions, the md files will still be in the repo

OneBlue (commented):
Looking through the doc itself, I think we should try to put a bit more information into what each value / field / argument does. For instance for `WslcStdIOCallback`, we should explain the synchronization guarantees, that releasing the process waits for all callbacks to complete, that the IO is synchronous, etc. 

I think a good template for this would be [https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa](CreateProcess), where the documentation explain what each argument does, how they interact, and potential gotchas

.github/workflows/documentation.yml
craigloewen-msft: Yes I can pull it out.

The only call out will be then we will only have 1 'live view of the docs' that only shows the latest. 

IF someone is on a much older version and wants to go see the docs (Especially LLMs who might search and look for this) then they would have to go navigate through our GitHub markdown files which will be much harder. 

Given that if you'd still like to remove it I can, just let me know :)

.github/workflows/documentation.yml
craigloewen-msft: Fixed! Good spot

OneBlue (approved):
Change LGTM from a build perspective. As a followup I think we should look into adding more details for what each argument / field / flag does and which return values methods return.

Note that once merged, this will be published to wsl.dev

PR #40929, opened 2026-06-26
fix: replace bare except with except Exception in Python scripts

srpatcha:
## Summary

Replace bare `except:` clauses with `except Exception:` in two Python scripts to comply with PEP 8 (E722).

## Changes

**`distributions/validate-modern.py`** (line 164):
```diff
-    except:
+    except Exception:
         if debug:
             import traceback
```

**`tools/devops/create-change.py`** (line 53):
```diff
-    except:
+    except Exception:
         if debug:
             import pdb
```

## Rationale

Both locations are debug/pdb handlers that re-raise the exception. Using `except Exception:` is sufficient for pdb post-mortem and avoids accidentally catching `SystemExit`, `KeyboardInterrupt`, and `GeneratorExit`, which should propagate normally.

### DIFF
--- a/distributions/validate-modern.py
+++ b/distributions/validate-modern.py
@@ -161,7 +161,7 @@ def main(manifest: str, tar: str, compare_with_branch: str, repo_path: str, arm6
 
             sys.exit(1 if errors else 0)
 
-    except:
+    except Exception:
         if debug:
             import traceback
             traceback.print_exc()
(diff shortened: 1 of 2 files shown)


OneBlue:
Thank you for doing this but given that this change is conflicted, shows CRLF mismatches and doesn't functionally change much, I don't think it's worth it. (--pdb is only used when debugging the script, we specifically want to catch everything there).

PR #40930, opened 2026-06-26
Added samples to the API

craigloewen-msft:
Added 4 basic samples for the WSLC API that will be referenced in the WSL docs.


doc/samples/WSLC-CustomContainer/Program.cs
```diff
@@ -0,0 +1,133 @@
...
+    // ---- Session ----
+    Console.Error.WriteLine("[wslc] Creating session...");
+    var sessionSettings = new SessionSettings("WSLCCustomContainer", sessionPath)
+    {
+        CpuCount = 2,
+        MemorySizeInMB = 2048,
+        VhdRequirements = new VhdOptions(string.Empty, 4UL * 1024 * 1024 * 1024, VhdType.Dynamic),
+    };
+
+    session = new Session(sessionSettings);
```
OneBlue: nit: `session` is IDisposable so we can use `using` here

OneBlue: (Same for containers, processes, and such)

doc/samples/WSLC-HelloWorld/helloworld.c
```diff
@@ -0,0 +1,159 @@
...
+    fwprintf(stderr, L"\n");
+}
+
+// Forward container stdout/stderr straight to the Windows console.
+static void CALLBACK OnStdIO(WslcProcessIOHandle ioHandle, const BYTE* data, uint32_t dataSize, PVOID context)
+{
+    HANDLE output = (ioHandle == WSLC_PROCESS_IO_HANDLE_STDOUT) ? GetStdHandle(STD_OUTPUT_HANDLE) : GetStdHandle(STD_ERROR_HANDLE);
+    DWORD written = 0;
+    (void)context;
+    WriteFile(output, data, dataSize, &written, NULL);
```
OneBlue: Given that this is text, I recommend using `fprintf` to write on stderr / stdout (we can use `%.*s` to write non-null terminated strings)

doc/samples/WSLC-HelloWorld/helloworld.c
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// Record the exit code and wake up wmain.
+static void CALLBACK OnProcessExit(INT32 exitCode, PVOID context)
+{
+    (void)context;
+    g_exitCode = exitCode;
+    SetEvent(g_exitEvent);
```
OneBlue: We should check the exit code of SetEvent here

doc/samples/WSLC-HelloWorld/helloworld.c
```diff
@@ -0,0 +1,159 @@
...
+    WslcProcessSettings initProcess;
+    WslcProcessSettings execProcess;
+    WslcProcessCallbacks callbacks;
+    WslcPullImageOptions pullOptions;
+    wchar_t storagePath[MAX_PATH];
+
+    PCSTR initArgv[2] = { "/bin/sleep", "60" };
+    PCSTR echoArgv[2] = { "/bin/echo", "Hello, World from a WSL container!" };
+
+    CoInitializeEx(NULL, COINIT_MULTITHREADED);
```
OneBlue: We should check the error code are here as well

doc/samples/WSLC-HelloWorld/helloworld.c
```diff
@@ -0,0 +1,159 @@
...
+    ZeroMemory(&callbacks, sizeof(callbacks));
+    callbacks.onStdOut = OnStdIO;
+    callbacks.onStdErr = OnStdIO;
+    callbacks.onExit = OnProcessExit;
+    WslcSetProcessSettingsCallbacks(&execProcess, &callbacks, NULL);
+
+    hr = WslcCreateContainerProcess(container, &execProcess, &process, &error);
+    if (FAILED(hr)) { PrintError(L"Run echo", hr, error); goto cleanup; }
+
+    WaitForSingleObject(g_exitEvent, 30000);
```
OneBlue: We should display an error if WaitForSingleObject times out here

doc/samples/WSLC-Neofetch/neofetch.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+namespace
+{
+    constexpr std::wstring_view c_imageName = L"anrginit/ubuntu-neofetch:1.0";
+
+    // Forward a chunk of container stdout/stderr straight to the Windows console.
+    void WriteToConsole(DWORD stdHandle, array_view<uint8_t const> data)
+    {
+        DWORD written = 0;
+        WriteFile(GetStdHandle(stdHandle), data.data(), static_cast<DWORD>(data.size()), &written, nullptr);
```
OneBlue: I also recommend using fprintf here

doc/samples/WSLC-Neofetch/neofetch.cpp
```diff
@@ -0,0 +1,125 @@
...
+        Process process = container.CreateProcess(processSettings);
+
+        handle exitEvent{ CreateEvent(nullptr, TRUE, FALSE, nullptr) };
+        int32_t exitCode = -1;
+
+        process.OutputReceived([](array_view<uint8_t const> data) { WriteToConsole(STD_OUTPUT_HANDLE, data); });
+        process.ErrorReceived([](array_view<uint8_t const> data) { WriteToConsole(STD_ERROR_HANDLE, data); });
+        process.Exited([&](int32_t code) {
+            exitCode = code;
+            SetEvent(exitEvent.get());
```
OneBlue: Error check for SetEvent(), createEvent() and WaitForSingleObject()

doc/samples/WSLC-NextCloud/Program.cs
```diff
@@ -0,0 +1,147 @@
...
+    Console.Error.WriteLine("[wslc] Creating session...");
+    var sessionSettings = new SessionSettings("WSLCNextCloud", sessionPath)
+    {
+        CpuCount = 4,
+        MemorySizeInMB = 4096,
+        // Nextcloud image is ~1.5 GB; use a 10 GB dynamic VHD.
+        VhdRequirements = new VhdOptions(string.Empty, 10UL * 1024 * 1024 * 1024, VhdType.Dynamic),
+    };
+
+    session = new Session(sessionSettings);
```
OneBlue: using for IDisposable classes

OneBlue (commented):
Change looks good overall, couple minor comments 

I also recommend add a pipeline to validate that all of these actually build (we can use a GH pipeline to make this easy, and only trigger it when `doc/samples` is edited

craigloewen-msft:
Fixed up requests - should be good to merge now I think!

OneBlue (approved):
LGTM. I think it would be a good idea to have a github action pipeline to build the sample, just to make sure that they don't get broken in future changes, but this can be done in a future change

PR #40933, opened 2026-06-27
Add a script to validate & fix mismatches CRLF in the repo

OneBlue:

This change adds a script that blocks merging a change with mismatches CRLF in code files. clang-format does some of it, but it doesn't cover all files. 

The files can be fixed via `python  tools/devops/validate-line-endings.py . --fix`

This will prevent needless git conflicts and copilot getting confused 

## PR Checklist


florelis:
Wouldn't it be easier to add a `.gitattributes` and renormalize everything?

OneBlue:
> Wouldn't it be easier to add a `.gitattributes` and renormalize everything?

I think a .gitattributes would be a good idea, although my understanding is that this would not actually enforce that the line endings are correct so I'd still want the pipeline to enforce it.

dkbennett (approved):

PR #40947, opened 2026-06-29
Don't append the build prefix when a line is being continued from a previous log entry

OneBlue:
## PR Checklist

This change solves an issue causing the build prefix to be inserted in a middle of a line of a build status is split across two logs entries (which can lead to test failures)


dkbennett (approved):
May want to add an E2E test later for the session logging output validation to ensure it's happening and in the expected format.

Issue #40945, opened 2026-06-29
wslc pull (and WSL autoProxy) fail behind a PAC/Zscaler proxy: http_proxy is set but HTTPS_PROXY is left empty, so HTTPS registry pulls bypass the proxy and time out


ket000:
### Windows Version

Microsoft Windows [Version 10.0.26200.8655]

### WSL Version

2.9.3.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Distro Version

Oracle Linux 9

### Other Software

On a machine whose only proxy configuration is a PAC script (AutoConfigURL), wslc pull of any HTTPS registry image times out:

wslc pull container-registry.oracle.com/os/oraclelinux:9
Get "https://container-registry.oracle.com/v2/": net/http: request canceled
while waiting for connection (Client.Timeout exceeded while awaiting headers)
context deadline exceeded
Error code: E_FAIL

curl -x http://<proxy>:443 https://container-registry.oracle.com/v2/   →  HTTP 401  (expected auth challenge = connectivity OK)

Root cause (from source, 2.9.3 tree)

In src/windows/service/exe/LxssUserSession.cpp, http_proxy/HTTP_PROXY are always injected, but https_proxy/HTTPS_PROXY are only injected when SecureProxy is non-empty:

http_proxy / HTTP_PROXY = proxySettings.Proxy            // ALWAYS set
if (!proxySettings.SecureProxy.empty()) {                // CONDITIONAL
    https_proxy / HTTPS_PROXY = proxySettings.SecureProxy
}

In src/windows/service/exe/LxssHttpProxy.cpp:
Proxy       = WideToMultiByte(WINHTTP_PROXY_SETTINGS_EX.pcwszProxy);
SecureProxy = WideToMultiByte(WINHTTP_PROXY_SETTINGS_EX.pcwszSecureProxy);

When proxy config comes from a PAC, WinHTTP returns a populated pcwszProxy but an empty pcwszSecureProxy. Result: WSL injects only http_proxy and leaves HTTPS_PROXY empty. Since registry pulls are HTTPS, the request goes direct to :443, has no route out of the corporate network, and times out.

Two distinct problems
autoProxy / PAC handling (general WSL): with a PAC-only config, HTTPS_PROXY is never set. Reproducible in the default distro — autoProxy injects http_proxy=http://<proxy>:443 but HTTPS_PROXY is empty.
[...]

PR #40950, opened 2026-06-30
Fix https proxy not set when pac is used

chemwolf6922:

PAC is not directly usable in WSL or in most headless Linux environments.
The WinHttpGetProxySettingsEx call only populates the pcwszProxy segment when PAC is enabled. And WSL only sets http_proxy but leaves https_proxy empty.
This PR sets both http_proxy and https_proxy to the default PAC query result as best effort.

## PR Checklist

- [ ] **Closes:** 

## Validation Steps Performed

New unit tests:
NetworkTests::NetworkTests::NatHttpProxyPac
NetworkTests::MirroredTests::HttpProxyPac

### DIFF
--- a/test/windows/NetworkTests.cpp
+++ b/test/windows/NetworkTests.cpp
@@ -1044,10 +1044,10 @@ class NetworkTests
 
     static void ClearHttpProxySettings(bool userScope)
     {
-        auto command = L"Set-WinhttpProxy -SettingScope Machine -Proxy \\\"\\\"";
+        auto command = L"Set-WinhttpProxy -SettingScope Machine -Proxy \\\"\\\" -AutoconfigUrl \\\"\\\"";
         if (userScope)
         {
-            command = L"Set-WinhttpProxy -SettingScope User -Proxy \\\"\\\"";
+            command = L"Set-WinhttpProxy -SettingScope User -Proxy \\\"\\\" -AutoconfigUrl \\\"\\\"";
         }
         LxsstuLaunchPowershellAndCaptureOutput(command);
     }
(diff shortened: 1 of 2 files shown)


test/windows/NetworkTests.cpp
```diff
@@ -1096,7 +1096,9 @@ class NetworkTests
     static constexpr auto c_httpProxyIpV4 = L"http://198.168.1.128:8888";
     static constexpr auto c_httpProxyIpV6 = L"http://[2001::1]:8888";
     static constexpr auto c_httpProxyBypassString = L"test";
-    static constexpr auto c_httpProxyPACurl = L"testpac.pac";
+    static constexpr auto c_pacServerPrefix = L"http://127.0.0.1:12399/";
+    static constexpr auto c_pacUrl = L"http://127.0.0.1:12399/wslproxy.pac";
+    static constexpr auto c_pacScript = LR"(function FindProxyForURL(url, host) { return \"PROXY test.com:8888\"; })";
```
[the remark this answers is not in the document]
chemwolf6922: Double escape is required since there is a powershell script layer.

test/windows/NetworkTests.cpp
```diff
@@ -1270,6 +1272,27 @@ class NetworkTests
...
+        SetHttpProxySettings(L"", L"", c_pacUrl, userScope);
+
+        // The update race condition is more likely to trigger for PAC as there is an additional http round trip.
+        wsl::shared::retry::RetryWithTimeout<void>(
+            [&]() {
+                auto [out, _] = LxsstuLaunchWslAndCaptureOutput(std::wstring(L"echo -n $") + c_httpProxyLower);
+                THROW_HR_IF(E_FAIL, out != c_httpProxyString);
+            },
+            std::chrono::seconds(1),
+            std::chrono::seconds(10));
```
OneBlue: nit: I recommend making the retry window longer (up to 2 minutes) so the tests don't fail if the machine is under pressure

test/windows/NetworkTests.cpp
OneBlue: nit: I recommend using `c_httpProxyString` to make this a bit easier to follow

OneBlue (commented):
Change LGTM, although I recommend making sure that #40945  doesn't get closed because that issue is about wslc, which doesn't support proxy right now

OneBlue (approved):

PR #40952, opened 2026-06-30
Localization change from build: 151090747

OneBlue:


benhillis:
@OneBlue - looks like touchdown is doing the wrong line endings, should we just ignore resw files?

OneBlue:
> @OneBlue - looks like touchdown is doing the wrong line endings, should we just ignore resw files?

Ah that's not fun. Let me see if it's using just LF, or mixing the line endings

OneBlue:
Well the script is actually right to flag this: 

image


There are broken CRLFs in the strings themselves :(

PR #40956, opened 2026-06-30
Add configurable storage path for the default wslc session

### LINKED ISSUE #40953: Make storage path for the default CLI session configurable
When C drive is small and D drive is used as a data drive, the space with adding more wslc containers runs out quickly due to growth of storage.vhd

A storage path for the default CLI session should be configurable (e.g. CLI command, settings.yaml, or registry).

Deleting the vhd

**Additional context**

benhillis:
## Summary

The default wslc CLI session always created its `storage.vhdx` under `%LOCALAPPDATA%\wslc\sessions\<session>`, which is a pain when the system drive is small. This adds a `session.storagePath` setting so users can redirect the default session's storage to another location (e.g. a larger data drive).

```yaml
session:
  storagePath: D:\wslc
```

Resolves to `D:\wslc\wslc\sessions\<session>\storage.vhdx`, mirroring the existing layout. When the key is absent or set to `default`, behavior is unchanged (`%LOCALAPPDATA%`). Named (custom) sessions are unaffected, they already accept an explicit storage path.

Fixes #40953

## Details

- `WSLCUserSettings.h` / `.cpp`: new `SessionStoragePath` setting (`session.storagePath`), validated as a non-empty absolute path (relative/empty rejected, falls back to the default), plus a documented entry in the settings-file template.
- `WSLCSessionManager.cpp`: `SessionSettings::Default` uses the configured base when set, otherwise `%LOCALAPPDATA%`. The downstream VHD plumbing was already path-driven, so no further wiring was needed.

## Testing

- New unit tests in `WSLCCLISettingsUnitTests.cpp` cover absent, absolute, `default`, and relative-rejected cases, plus the all-known-keys no-warning check.
- Builds clean under `/WX`: `common.lib`, `wslservice.exe`, `wslc.exe`, `wsltests.dll`.
- `clang-format` clean on all changed files.

Note: the validation only checks that the path is absolute; if the path is not writable by the user, session creation surfaces the underlying error at VHD-attach time (same as the existing custom-session path).


dkbennett (dismissed):
LGTM. Do we have any runtime validation for the storage path both existing and having correct permissions to be accessible to the user, or does that surface as a session error when we attempt to use it?

OneBlue (commented):
Do we want to expose this as a configuration key ? One downside of doing this is that changing this path will essentially drop everything that the user has. 

One alternative we considered was to tell users to symlink their storage folder instead. This has the benefit of explicitly requiring user action, and so the user can also move their storage while they're doing it

benhillis:
> Do we want to expose this as a configuration key ? One downside of doing this is that changing this path will essentially drop everything that the user has.
> 
> One alternative we considered was to tell users to symlink their storage folder instead. This has the benefit of explicitly requiring user action, and so the user can also move their storage while they're doing it

Definitely one downside of this approach. I'm not a huge fan of the symlink approach. This one needs a little more UX thought but I think we really need to come up with a solution here.

src/windows/common/WSLCUserSettings.cpp
```diff
@@ -147,6 +151,16 @@ namespace details {
         return value;
     }
 
+    WSLC_VALIDATE_SETTING(SessionStoragePath)
+    {
+        if (value.empty() || !std::filesystem::path(MultiByteToWide(value)).is_absolute())
```
OneBlue: nit: `MultiByteToWide()` not needed here (same in other places)

OneBlue (dismissed):

OneBlue:
> Definitely one downside of this approach. I'm not a huge fan of the symlink approach. This one needs a little more UX thought but I think we really need to come up with a solution here.

I discussed this @dkbennett a bit and there's definitely downsides to both. I think this implementation is OK for now. If users modify the settings I think the storage being reset is a reasonable UX. We can refine later if needed

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -86,12 +88,32 @@ struct SessionSettings
     static std::unique_ptr<SessionSettings> Default(HANDLE UserToken, const std::wstring& ResolvedName)
     {
         auto userSettings = LoadUserSettings(UserToken);
-        auto localAppData = wsl::windows::common::filesystem::GetLocalAppDataPath(UserToken);
 
-        auto storagePath = (localAppData / wsl::windows::wslc::DefaultStorageSubPath / ResolvedName).wstring();
+        auto configuredStorageBase = userSettings.Get<settings::Setting::SessionStoragePath>();
```
benhillis, line 90: This all seems overly complicated.

dkbennett: Detecting the orphan might be a bit much, could simplify with a bit more documentation on the setting warning users there, so if they change it they know to check the default and handle it however they want. Maybe stash the orphan detection and revisit if customers express confusion or issues not knowing about the orphan, or (in a later release) do that detection after an upgrade or something where we might naturally do some cleanup and migration.

benhillis: Yeah that's not a bad call. Or maybe just a warning when we are creating a VHD in the non default location that, which is basically what this is doing except it's also checking for the default location too which bloats the check.

localization/strings/en-US/Resources.resw
```diff
@@ -2417,6 +2417,10 @@ For privacy information about this product please visit https://aka.ms/privacy.<
     <value>Cannot use '{}' as session storage because it is not a directory</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
+  <data name="MessageWslcSessionStorageCustomLocation" xml:space="preserve">
+    <value>Creating WSLC session storage at '{}' as configured by session.storagePath. If you change or remove this setting, data stored at this location will no longer be used by the default session.</value>
```
OneBlue: I think the warning is good idea, but the current implementation will warn pretty much all the time as long as the default VHD path exists. 

If we want to have this warning, I think it should also say something like `Delete /path/to/vhd to suppress this warning`

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -244,6 +256,12 @@ void WSLCSessionManagerImpl::CreateSession(
     {
         defaultSettings = SessionSettings::Default(callerToken.get(), resolvedDisplayName);
         Settings = &defaultSettings->Settings;
+
+        if (WarningCallback != nullptr && defaultSettings->WarnCustomStorageLocation)
+        {
+            LOG_IF_FAILED(WarningCallback->OnWarning(
```
OneBlue: Hmm I'm guessing we're doing this because EMIT_USER_WARNING isn't plumbed here. 

Unfortunately this also means that the COM user callback cancellation won't run, and so if the client call hangs, the service will hang as well. 

Maybe the easiest solution around this would be to emit the warning from `wslcsession.exe` ?

src/windows/service/exe/WSLCSessionManager.cpp
```diff
@@ -97,17 +95,11 @@ struct SessionSettings
...
-            auto runAsUser = wil::impersonate_token(UserToken);
-            std::error_code ec;
-            result->WarnCustomStorageLocation = !std::filesystem::exists(storageDir / wsl::windows::wslc::DefaultStorageVhdName, ec);
-        }
+        // wslcsession emits the custom-location warning when it actually creates the VHD, so the notice
+        // fires once at creation without a service-side callback that could stall CreateSession.
+        const auto storageFlags = customConfigured ? WSLCSessionStorageFlagsWarnCustomLocation : WSLCSessionStorageFlagsNone;
 
-        return result;
+        return std::unique_ptr<SessionSettings>(new SessionSettings(std::wstring(ResolvedName), storageDir.wstring(), storageFlags, userSettings));
```
OneBlue, line 102: nit: We could use `std::make_unique` to simplify this a bit

OneBlue (approved):

PR #40971, opened 2026-07-01
Wslc events

kvega005:
Adds a docker events style event stream to the WSLC session. A new `IWSLCEventStream` COM interface and `IWSLCSession::GetEvents` let callers subscribe to container lifecycle events (`create`, `start`, `kill`, `stop`, `destroy`), filtered by time window and key/value filters, pulled one JSON event at a time. Events are backed by a new bounded in-memory ring (`EventStore`) fed from Docker's own event stream.

## PR Checklist


src/windows/wslcsession/EventStore.cpp
```diff
@@ -0,0 +1,254 @@
...
+    // Eviction while parked wakes us too, so the caller reports the gap on its next pass.
+    const auto ready = [&] {
+        return m_terminating || SequenceNumber < m_firstSequenceNumber || GetLockHeld(SequenceNumber).has_value();
+    };
+
+    if (Until.has_value())
+    {
+        // Until is Unix seconds and system_clock shares that epoch, so it doubles as the wait deadline.
+        const std::chrono::sys_seconds deadline{std::chrono::seconds{static_cast<int64_t>(Until.value())}};
+        if (!m_updated.wait_until(Lock, deadline, ready))
```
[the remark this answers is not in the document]
kvega005: The worst outcome from a wrapped (negative) deadline: wait_until sees a deadline in the past, the ready predicate is evaluated once, and if nothing's buffered it returns false → Get returns nullopt → the stream reports "finished" immediately. The caller just gets an empty stream. No hang, no corruption, no security impact.

src/windows/wslcsession/EventStore.cpp
```diff
@@ -0,0 +1,254 @@
...
+}
+
+bool EventStore::WaitForEvent(std::unique_lock<std::mutex>& Lock, uint64_t SequenceNumber, std::optional<uint64_t> Until)
+{
+    // Ready once the reader's event is buffered, its slot is evicted, or the session terminates.
+    // Eviction while parked wakes us too, so the caller reports the gap on its next pass.
+    const auto ready = [&] {
+        return m_terminating || SequenceNumber < m_firstSequenceNumber || GetLockHeld(SequenceNumber).has_value();
+    };
+
```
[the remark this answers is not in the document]
kvega005: Addressed

localization/strings/en-US/Resources.resw
```diff
@@ -2291,6 +2291,10 @@ For privacy information about this product please visit https://aka.ms/privacy.<
     <value>Invalid name: '{}'</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
+  <data name = "MessageWslcEventsInvalidTimeWindow" xml:space = "preserve" >
+    <value>`since` time ({}) cannot be after `until` time ({})</value>
+    <comment>{FixedPlaceholder="{}"}{Locked="`since`"}{Locked="`until`"}Command line arguments, file names and string inserts should not be translated</comment>
+  </data>
```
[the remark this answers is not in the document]
kvega005: Addressed

src/windows/wslcsession/EventStore.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if (m_events.size() > c_eventRingCapacity)
+    {
+        m_events.pop_front();
+        ++m_firstSequenceNumber;
+    }
+
+    m_updated.notify_all();
+}
+
+void EventStore::Record(std::string Type, std::string Action, const std::string& ActorId, std::optional<uint64_t> TimeSeconds) noexcept
```
OneBlue: nit: We could probably make the strings rvalue references to avoid copies here since the caller will most likely be able to move them all

src/windows/wslcsession/EventStore.cpp
```diff
@@ -0,0 +1,252 @@
...
+    const uint64_t index = SequenceNumber - m_firstSequenceNumber;
+    if (index >= m_events.size())
+    {
+        return std::nullopt;
+    }
+
+    return m_events[index];
+}
+
+bool EventStore::WaitForEvent(std::unique_lock<std::mutex>& Lock, uint64_t SequenceNumber, std::optional<uint64_t> Until)
```
OneBlue, line 128: Not necessary for this change, but ideally we should also wait for the caller's process, otherwise if a user runs `wslc events` and then ctrl-c, a thread on the service side could get stuck here until an event is emitted to unblock it

OneBlue: Sadly condition_variables don't easily let us wait for multiple events so we might need to make our own for this

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -880,8 +882,20 @@ void WSLCContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCod
     // Disconnect(), so in-flight COM callers can drain from COMImplClass::m_callers.
     unique_com_disconnect comWrapper;
 
-    if (event == ContainerEvent::Stop)
+    // Record lifecycle events here, driven by Docker's own events, so the store gets Docker's
+    // timestamps and the same ordering Docker observed.
+    if (event == ContainerEvent::Start)
```
OneBlue: Ideally I think we should report these events after we've updated our internal state, otherwise a caller could see the `container stop` event and let's say try to call `container.Start()` again but if we haven't internally transitioned to the stopped state, their call would race with ours. 

My intuition there would be to emit those events in the `Transition()` method, but we're currently missing the event time when we start the container (should be easy to fix though)

OneBlue: Unfortunately, I think we're going to need to redesign the container state machine a bit (mostly because I recently realized that if the caller makes a stop call with an infinite timeout, any Kill() call will get stuck)

OneBlue (dismissed):
LGTM (looks like the pipeline failed for unrelated issues though).

OneBlue (dismissed):

src/windows/wslcsession/EventStore.cpp
```diff
@@ -0,0 +1,254 @@
...
+        // it.
+        if (SequenceNumber.value() < m_firstSequenceNumber)
+        {
+            continue;
+        }
+
+        const auto event = GetLockHeld(SequenceNumber.value()).value();
+
+        // An event past the until-bound closes the window: the stream is finished.
+        if (Until.has_value() && event.time > Until.value())
```
ptrivedi: the filtering here is exclusive `event.time > Until`, but the wait deadline is the instant the second begins. The deadline should represent the end of inclusive UntilTime second, not its beginning. 

thinking out loud, let's say clietn starts a stream at UntilTime=x. Then it drains the existing events and waits using wait_until and times out when the second begsins... so at x:00:00.000. But there could be an event received at x:00:00.400 which should get included but is lost due to stream ending. It could cause some confusion when trying to follow the order of events e.g. there may be events for a container, but no create event

src/windows/wslcsession/EventStore.cpp
```diff
@@ -0,0 +1,254 @@
...
+        {
```
ptrivedi, line 203: another thing: are the events in the ring chrono-ordered? looks like the events are added to the back in arrival/insertion order, not necessarly chronological order by event.time. could there be events that get docker's timestamp after the fact? that could cause the stream to terminate

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -2541,6 +2579,7 @@ std::shared_ptr<WSLCContainerImpl> WSLCContainerImpl::Create(
     }
 
     auto mergedLabels = StripInternalLabels(std::move(inspectData.Config.Labels));
+    const auto createdAt = wsl::windows::common::timestamp::Rfc3339ToEpoch(inspectData.Created);
```
ptrivedi, line 2546: In reference to the comment around whether the events in the ring are truly ordered... looks like here the create event would get docker's Created timestamp, so could be backdated

src/windows/wslcsession/EventStore.cpp
ptrivedi, line 203: also this check happens before event filters, so a filtered event can cause termination

kvega005: Good catch. Nondecreasing event time is intended to be an invariant here, since the until early-exit relies on it. For events from Docker’s /events stream, Docker includes the timestamp in each event and WSLC processes and records them synchronously in stream order.

While checking this, though, I noticed one hole: WSLC synthesizes the create event after WaitForObjectCreated, using the timestamp from inspect. Concurrent creates could therefore be appended out of timestamp order. I’ll fix that path and add a WI_ASSERT in EventStore::Append under the lock to make the ordering invariant explicit.

src/windows/wslcsession/EventStore.cpp
kvega005: I checked Moby’s behavior. Docker treats an integer Unix until=x as the exact instant x.000000000, not the end of second x. Its live handler times out at that instant, and buffered events are filtered using TimeNano <= until, so an event at x.400 is intentionally excluded.

I agree there is an inconsistency here, though: WSLC keeps only whole-second event.time, so an x.400 event becomes time=x and can pass the backlog comparison even though the live stream closes at x.000. To match Docker, I think we should preserve nanosecond precision for filtering rather than extend the deadline by one second. 

I had planned to do that in a follow-up PR because at the time, there was no framework for Unix timestamps in the WSLC code base. But it was added recently, so I can include the Unix timestamps as part of this PR.

src/windows/wslcsession/EventStore.cpp
ptrivedi: Awesome, let me know when done, I can approve. LGTM once addressed

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2374,6 +2391,77 @@ void WSLCSession::CreateContainerImpl(const WSLCContainerOptions* containerOptio
...
+}
+
+void WSLCSession::WaitForPendingCreateCompletion(const std::shared_ptr<PendingContainerCreate>& PendingCreate)
+{
+    auto io = CreateIOContext();
+    io.AddHandle(std::make_unique<io::EventHandle>(PendingCreate->Completed.get()));
+    io.Run({});
+
+    WI_ASSERT(PendingCreate->Completed.is_signaled());
+}
```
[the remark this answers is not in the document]
kvega005: The timeout is not necessary because when we commit the pending create operation, we expect to receive the docker create event eventually. The wait can be cancelled when the calling process exits.

src/windows/wslcsession/EventStore.cpp
```diff
@@ -0,0 +1,256 @@
...
+    {
+        // A reader that has fallen behind the ring missed events to eviction: reset it so the
+        // next call starts fresh at the oldest buffered event, and report the gap.
+        if (SequenceNumber.value() < m_firstSequenceNumber)
+        {
+            SequenceNumber = std::nullopt;
+            THROW_HR(WSLC_E_EVENTS_LOST);
+        }
+
+        if (!WaitForEvent(lock, SequenceNumber.value(), Until))
```
ptrivedi, line 178: does GetNext() need to be serialized with a per-stream mutex or something? I am thinking concurrent GetNext() calls may end up corrupting the stream cursor? m_lock here will help preventing the cursor from being mutated simultaneously, but if two calls end up passing the same curosr value to WaitForEvent above, there could be a race and one of the streams could fail and cursot would be inconsistent

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2353,13 +2358,24 @@ void WSLCSession::CreateContainerImpl(const WSLCContainerOptions* containerOptio
...
             m_pluginNotifier.get(),
             m_networks,
-            std::bind(&WSLCSession::OnContainerDeleted, this, std::placeholders::_1));
+            std::bind(&WSLCSession::OnContainerDeleted, this, std::placeholders::_1),
+            m_eventStore);
 
-        // Key the map by Docker's container ID, which is set in the WSLCContainerImpl constructor and stable for its lifetime.
-        auto [it, inserted] = m_containers.emplace(container->ID(), std::move(container));
-        WI_ASSERT(inserted);
+        auto pendingCreate = StartPendingCreate(container);
```
ptrivedi, line 2362: Hmm, so the m_pendingCreate will only get cleared when the docker create event is processed.. but if that fails for whatever reason, it will wedge all the later creates? this shoudl be time-bound

src/windows/wslcsession/EventStore.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+bool EventStore::WaitForEvent(std::unique_lock<std::mutex>& Lock, uint64_t SequenceNumber, std::optional<std::chrono::sys_seconds> Until)
+{
+    // Ready once the reader's event is buffered, its slot is evicted, or the session terminates.
+    // Eviction while parked wakes us too, so the caller reports the gap on its next pass.
+    const auto ready = [&] { return m_terminating || SequenceNumber < m_firstSequenceNumber + m_events.size(); };
+
+    if (Until.has_value())
+    {
+        if (!m_updated.wait_until(Lock, Until.value(), ready))
```
ptrivedi, line 142: pretty large positive seconds could still overflow when wait_until() converts sys_seconds to the windows system_clock?

src/windows/wslcsession/EventStore.cpp
```diff
@@ -0,0 +1,256 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#include "precomp.h"
+#include "EventStore.h"
+#include "WSLCSession.h"
+#include "WSLCExecutionContext.h"
+#include <chrono>
+#include "wslc_schema.h"
```
ptrivedi: nit: doesn't seem like this include would be needed here.

src/windows/WslcSDK/wslcsdk.h
```diff
@@ -44,6 +44,8 @@ EXTERN_C_START
 #define WSLC_E_VOLUME_NOT_AVAILABLE MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 14)       /* 0x8004060E */
 #define WSLC_E_SESSION_NOT_FOUND MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 15)          /* 0x8004060F */
 #define WSLC_E_VM_NOT_RUNNING MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 16)             /* 0x80040610 */
+#define WSLC_E_EVENTS_LOST MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 17)                /* 0x80040611 */
+#define WSLC_E_EVENT_STREAM_FINISHED MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 18)      /* 0x80040612 */
```
ptrivedi, line 48: nit: should double-check these are present in error codes documentation and in g_commonErrors

test/windows/WSLCTests.cpp
```diff
@@ -6857,6 +6858,162 @@ class WSLCTests
...
+        });
+        auto threadCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { readerThread.join(); });
+
+        auto future = getNextResult.get_future();
+
+        VERIFY_SUCCEEDED(m_defaultSession->Terminate());
+        auto restore = ResetTestSession();
+
+        // Termination wakes the parked reader; it must finish quickly and report E_ABORT.
+        VERIFY_ARE_EQUAL(std::future_status::ready, future.wait_for(10s));
```
ptrivedi: this 10s timeout does not bound the test, it will still go ahead and call future.get() and hang?

test/windows/WSLCTests.cpp
```diff
@@ -6857,6 +6858,162 @@ class WSLCTests
...
+        VERIFY_ARE_EQUAL(E_ABORT, future.get());
+    }
+
```
ptrivedi, line 7148: mssing tests for testing multiple values for the same filter key use

test/windows/WSLCTests.cpp
```diff
@@ -6857,6 +6858,162 @@ class WSLCTests
...
+        auto future = getNextResult.get_future();
+
+        VERIFY_SUCCEEDED(m_defaultSession->Terminate());
+        auto restore = ResetTestSession();
+
+        // Termination wakes the parked reader; it must finish quickly and report E_ABORT.
+        VERIFY_ARE_EQUAL(std::future_status::ready, future.wait_for(10s));
+        VERIFY_ARE_EQUAL(E_ABORT, future.get());
+    }
```
ptrivedi, line 7147: we should likely have a test for what happens when a stream falls behind, given the store retains only 256 events

ptrivedi (approved):

PR #40972, opened 2026-07-01
Support Docker-style network aliases in wslc

### LINKED ISSUE #40970: Support network aliases in wslc container run/create --network arguments
`wslc container run/create --network` currently supports selecting a network, but does not provide a way to specify per-network aliases. This makes it harder to run containerized workloads that expect Docker-style network-scoped DNS aliases, especially when a container needs different aliases on different user-defined networks.

Extend `wslc container run` and `wslc container create` so `--network` accepts Docker-style advanced network syntax, including per-network aliases, for example:

```powershell
wslc container run --network name=frontend,alias=web --network name=backend,alias=api,alias=db debian:latest ...

N/A

**Additional context**
See https://docs.docker.com/reference/cli/docker/container/run/#network for the full set of advanced network options. Docker supports additional properties beyond just `alias`, but `alias` is arguably the most important option.

### PR DESCRIPTION
**Pull request by danegsta**

This adds Docker-style per-network alias support to `wslc container run` and `wslc container create` so aliases can be scoped directly in each `--network` value. This makes multi-network container creation less ambiguous while preserving the existing `--network-alias` behavior for single-network cases.

## PR Checklist

- [x] **Closes:** #40970
- [ ] **Communication:** N/A - opening for maintainer review
- [ ] **Dev docs:** N/A

Fixes #40970

The CLI now accepts Docker-style network values such as `--network name=my-net,alias=api,alias=backend`, including values where `alias` appears before `name`. Parsed aliases are carried through the container task model into the service and launcher layers, then passed as endpoint settings for both the primary and additional network attachments.

Alias validation remains scoped to user-defined networks: aliases are rejected for built-in `bridge`, `host`, `none`, and `container:` networking modes, and ambiguous `--network-alias` usage with multiple networks is still rejected. The lower-level endpoint setting parser only accepts the `Aliases` key so unsupported endpoint settings fail explicitly.

## Validation Steps Performed

- Ran `cmake --fresh .` after setting up the Windows build environment.
- Ran the Windows build successfully before local dev-loop deployment configuration was added.
- Ran focused build after parser/localization changes: `cmake --build . --target wslc wsltests -- -m`.
- Ran elevated network alias test subset: `bin\x64\Debug\test.bat /name:*NetworkAlias*`.
- Smoke tested with a real deployed `wslc.exe`: created user-defined networks, started containers with per-network aliases, confirmed `inspect` output included the aliases, and verified DNS alias resolution from peer containers on the same network.

### CODE REVIEW

File: localization/strings/en-US/Resources.resw [hunk 1]
```diff
@@ -2884,6 +2884,14 @@ On first run, creates the file with all settings commented out at their defaults
...
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
+  <data name="WSLCCLI_NetworkDuplicateNameError" xml:space="preserve">
+    <value>Invalid {} value: network name can only be specified once</value>
+    <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
+  </data>
+  <data name="WSLCCLI_NetworkUnsupportedOptionError" xml:space="preserve">
+    <value>Invalid {} value: unsupported network option '{}'</value>
+    <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
+  </data>
```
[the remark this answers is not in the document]
danegsta: Good catch. I accepted this and added a second `{FixedPlaceholder="{}"}` to match the two string inserts in `WSLCCLI_NetworkUnsupportedOptionError`; localization validation passes.

test/windows/wslc/WSLCCLIExecutionUnitTests.cpp
```diff
@@ -429,8 +471,62 @@ class WSLCCLIExecutionUnitTests
...
+        VERIFY_ARE_EQUAL(std::string("net1"), options.Networks[0].Name);
+        VERIFY_ARE_EQUAL(2u, options.Networks[0].Aliases.size());
+        VERIFY_ARE_EQUAL(std::string("a"), options.Networks[0].Aliases[0]);
+        VERIFY_ARE_EQUAL(std::string("b"), options.Networks[0].Aliases[1]);
+        VERIFY_ARE_EQUAL(std::string("net2"), options.Networks[1].Name);
+        VERIFY_ARE_EQUAL(1u, options.Networks[1].Aliases.size());
+        VERIFY_ARE_EQUAL(std::string("c"), options.Networks[1].Aliases[0]);
+    }
+
+    TEST_METHOD(CreateCommand_ParseDockerNetworkAliasesWithNameAfterAlias_SetsPerNetworkAliases)
```
beena352, line 508: Could we also cover the  DuplicateNetworkName  and  UnsupportedOption  branches? Right now, there's no test for --network name=a,name=b or  --network aliases=x. One for --network alias=x with no name=  at all would also be nice since it lands on  EmptyNetworkName  through a different code path than the plain empty-string case.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -116,13 +116,25 @@ void Argument::Validate(const ArgMap& execArgs) const
     {
         for (const auto& value : execArgs.GetAll<ArgType::Network>())
         {
-            if (value.empty() ||
-                std::all_of(value.begin(), value.end(), [](wchar_t c) { return std::iswspace(static_cast<wint_t>(c)); }))
+            const auto parsed = models::ParseNetworkArgument(value);
```
AmelBawa-msft: Super nit: Consider moving this change into a Validate* method to align with the other block options, especially since the new change is larger

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -357,8 +369,8 @@ FormatType GetFormatTypeFromString(const std::wstring& input, const std::wstring
     }
     else
     {
-        throw ArgumentException(std::format(
-            L"Invalid {} value: {} is not a recognized format type. Supported format types are: json, table.", argName, input));
+        throw ArgumentException(
+            std::format(L"Invalid {} value: {} is not a recognized format type. Supported format types are: json, table.", argName, input));
```
AmelBawa-msft: This may be a different version of clang format attempting to format 🤔

src/windows/wslc/services/ContainerModel.h
```diff
@@ -16,17 +16,156 @@ Module Name:
...
+#include <cwctype>
+#include <optional>
 #include <string>
+#include <string_view>
+#include <vector>
 
 namespace wsl::windows::wslc::models {
 
+namespace details {
+    inline std::wstring ToWideString(std::wstring_view value)
```
AmelBawa-msft: This may be a duplicate of: `MultiByteToWide`?

src/windows/wslc/services/ContainerModel.h
```diff
@@ -16,17 +16,156 @@ Module Name:
...
+    std::wstring ErrorValue;
+};
+
+struct ContainerNetwork
+{
+    std::string Name;
+    std::vector<std::string> Aliases;
+};
+
+inline ParsedNetworkArgument ParseNetworkArgument(std::wstring_view value)
```
beena352: nit: any reason  ParseNetworkArgument  lives here instead of  arguments/ArgumentValidation.cpp ?  ParseLabel ,  ParseFilter , and  ParseDriverOption  all live there.

src/windows/wslc/services/ContainerModel.h
```diff
@@ -16,17 +16,156 @@ Module Name:
...
 
 namespace wsl::windows::wslc::models {
 
+namespace details {
+    inline std::wstring ToWideString(std::wstring_view value)
+    {
+        return {value.data(), value.size()};
+    }
+
+    inline std::vector<std::wstring_view> SplitPreserveEmpty(std::wstring_view value, wchar_t delimiter)
```
AmelBawa-msft: I wonder if we should leverage existing method here, and if not possible, promote this one next to the existing ones:

https://github.com/microsoft/WSL/blob/6c7f97fcb9dd38ba6035a742022c2dde5a1ffe75/src/windows/common/wslutil.h#L344-L345

src/windows/wslc/services/ContainerModel.h
AmelBawa-msft: We "recently" moved some parsing methods to ArgumentValidation.h/cpp. This may be a similar candidate as well

src/windows/wslc/services/ContainerModel.h
```diff
@@ -16,17 +16,156 @@ Module Name:
...
+    if (value.find(L'=') != std::wstring_view::npos)
+    {
+        parseOptions(value, true);
+    }
+    else
+    {
+        result.Name = details::ToWideString(value);
+    }
+
+    if (result.Error == NetworkArgumentParseError::None)
```
AmelBawa-msft: Super nit: We can move/decouple this validation logic to the Validate* method I commented about above

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -42,6 +42,11 @@ static void SetContainerArguments(WSLCProcessOptions& options, std::vector<const
     options.CommandLine = {.Values = argsStorage.data(), .Count = static_cast<ULONG>(argsStorage.size())};
 }
 
+static bool SupportsNetworkAliases(std::string_view network)
+{
+    return network != "bridge" && network != "host" && network != "none" && !network.starts_with("container:");
```
AmelBawa-msft, line 47: Why `container:` prefix is specified here, is it reserved? Nit: Consider adding a comment

src/windows/wslc/services/ContainerModel.h
danegsta: Agreed. I moved ParseNetworkArgument and its parsing helpers into ArgumentValidation.h/cpp next to the existing ParseLabel, ParseFilter, and ParseDriverOption helpers.

danegsta: Agreed. The network argument parser now lives in ArgumentValidation so it is no longer tied to ContainerModel and can be reused by other argument handling code.

src/windows/wslc/services/ContainerModel.h
danegsta: Agreed. I moved the network-specific validation into ValidateNetwork and have Argument::Validate delegate to that centralized validation path.

src/windows/wslc/services/ContainerService.cpp
danegsta: This was preserved from the previous code (just moved into a helper so it could be re-used). The `container:` syntax matches advanced Docker networking where a container can be configured with the network config of another container (instead of being attached to specific networks). I'll add a comment explaining what it's for since it's definitely one of the less well-known networking features.

src/windows/wslc/services/ContainerModel.h
danegsta: I switched the code from using this new helper to simply calling std::wstring{value} since it does the same thing.

src/windows/wslc/services/ContainerModel.h
danegsta: Ultimately it was just an unnecessary complication over calling std::wstring{value}.

src/windows/wslc/arguments/ArgumentValidation.cpp
danegsta: Not sure if it was clang format or drivee by LLM, but the format change should be reverted now.

src/windows/wslc/arguments/ArgumentValidation.cpp
danegsta: Done, there's a proper Validate method for the network argument now.

test/windows/wslc/WSLCCLIExecutionUnitTests.cpp
danegsta: Good call. I added targeted coverage for duplicate network names, unsupported `aliases=`, and `alias=` without `name=`, with assertions on the specific validation messages so the intended branches are exercised.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -294,9 +294,13 @@ std::map<std::string, EndpointConfig> ResolveEndpoints(
...
 
-        // Per-endpoint Settings are reserved for future use (IPAddress, Aliases, ...). No keys
-        // are interpreted today, so reject any non-empty payload rather than silently dropping it.
-        THROW_HR_WITH_USER_ERROR_IF(E_NOTIMPL, Localization::MessageWslcEndpointSettingsNotSupported(name), connections[i].SettingsCount > 0);
+        auto config = ResolveEndpointConfig(connections[i].Settings, connections[i].SettingsCount, name);
+        THROW_HR_WITH_USER_ERROR_IF(
+            E_INVALIDARG,
+            Localization::MessageWslcAliasRequiresUserDefinedNetwork(),
+            config.Aliases.has_value() && name == "bridge");
+
```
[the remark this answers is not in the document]
danegsta: Good catch. I applied this by sharing the alias eligibility check between primary and additional endpoints, so additional bridge/host/none/container: aliases now fail with the user-defined-network error before network lookup; I also added coverage for those cases.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -294,9 +294,13 @@ std::map<std::string, EndpointConfig> ResolveEndpoints(
...
         }
 
-        // Per-endpoint Settings are reserved for future use (IPAddress, Aliases, ...). No keys
-        // are interpreted today, so reject any non-empty payload rather than silently dropping it.
-        THROW_HR_WITH_USER_ERROR_IF(E_NOTIMPL, Localization::MessageWslcEndpointSettingsNotSupported(name), connections[i].SettingsCount > 0);
+        auto config = ResolveEndpointConfig(connections[i].Settings, connections[i].SettingsCount, name);
+        THROW_HR_WITH_USER_ERROR_IF(
+            E_INVALIDARG,
+            Localization::MessageWslcAliasRequiresUserDefinedNetwork(),
+            config.Aliases.has_value() && name == "bridge");
```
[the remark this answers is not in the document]
danegsta: This is covered by the latest update. Additional endpoints now use the shared `NetworkSupportsAliases` check, which rejects aliases for `bridge` and for modes where `NetworkModeAllocatesVmPorts` is false before network lookup.

beena352:
LGTM! Thanks for addressing the feedback. @OneBlue could you take a look?

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -166,6 +154,34 @@ static const std::unordered_map<std::wstring, WSLCSignal> SignalMap = {
     {L"SIGPOLL", WSLCSignalSIGPOLL}, {L"SIGPWR", WSLCSignalSIGPWR},     {L"SIGSYS", WSLCSignalSIGSYS},
 };
 
+namespace {
+    std::vector<std::wstring_view> SplitPreserveEmpty(std::wstring_view value, wchar_t delimiter)
```
OneBlue: nit: I recommend moving to `stringshared.h` since this is a generic helper (same for IsEmptyOrWhitespace)

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -192,6 +208,35 @@ void ValidateFilter(const std::vector<std::wstring>& values)
     }
 }
 
+void ValidateNetwork(const std::vector<std::wstring>& values, const std::wstring& argName)
+{
+    for (const auto& value : values)
+    {
+        const auto parsed = ParseNetworkArgument(value);
+        if (parsed.Error == NetworkArgumentParseError::EmptyNetworkName)
```
OneBlue: To simplify this, I recommend throwing the exception directly in `ParseNetworkArgument`, since the enum values are already used here

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -494,6 +539,84 @@ std::tuple<std::string, int64_t, int64_t> ParseUlimit(const std::wstring& input,
...
+            if (separator == std::wstring_view::npos)
+            {
+                result.Error = NetworkArgumentParseError::UnsupportedOption;
+                result.ErrorValue = std::wstring{part};
+                return;
+            }
+
+            const auto key = part.substr(0, separator);
+            const auto optionValue = part.substr(separator + 1);
+            if (key == L"name")
```
OneBlue: Here I would recommend parsing generically and making `ParsedNetworkArgument` store all values in an 
`std::vector<std::pair<std::string, std::string>>` (except name, which I think makes sense to special case here). 

That way, all driver options will be supported without needing CLI changes

OneBlue (commented):
Overall structure LGTM. Couple minor comments

src/windows/wslc/arguments/ArgumentValidation.cpp
danegsta: Moved both to stringshared.h

src/windows/wslc/arguments/ArgumentValidation.cpp
danegsta: Done, the exception are now thrown driectly from the parser for parsing related errors

src/windows/wslc/arguments/ArgumentValidation.cpp
danegsta: I think that even with a vector of key/value pairs this method would end up looking effectively the same. The main thing is that the options being parsed here are CLI specific options matching the network options from https://docs.docker.com/reference/cli/docker/container/run/#network. They don't share the same name as the underlying driver options (and there's a dedicated option type specifically for arbitrary driver options). If you want I can update the scope of this PR to cover more than just the alias option such that `--network` could support the full set of top level options and arbitrary network driver options.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -553,7 +553,10 @@ void SetContainerOptionsFromArgs(CLIExecutionContext& context)
         options.Networks.reserve(options.Networks.size() + networks.size());
         for (const auto& value : networks)
         {
-            options.Networks.emplace_back(WideToMultiByte(value));
+            auto parsed = validation::ParseNetworkArgument(value);
```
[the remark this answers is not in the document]
danegsta: Accepted. I passed the network argument name through SetContainerOptionsFromArgs so fallback parsing emits the same localized argument name, and added a focused unit test for the no-prior-validation path.

src/windows/wslc/arguments/ArgumentValidation.cpp
OneBlue: That's a fair point. I think the current implementation is OK for now in this case. We can always expand later

danegsta: Okay, I'll mark this as resolved for now then.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -494,6 +494,70 @@ std::tuple<std::string, int64_t, int64_t> ParseUlimit(const std::wstring& input,
...
+            {
+                if (parsedName)
+                {
+                    throw ArgumentException(Localization::WSLCCLI_NetworkDuplicateNameError(argName));
+                }
+
+                parsedName = true;
+                result.Name = WideToMultiByte(std::wstring{optionValue});
+            }
```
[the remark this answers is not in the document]
danegsta: Agreed. I now validate `name=` values as wide strings before converting them, so empty or whitespace-only advanced network names fail with the normal network-name error.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -494,6 +494,70 @@ std::tuple<std::string, int64_t, int64_t> ParseUlimit(const std::wstring& input,
...
+    }
+    else
+    {
+        result.Name = WideToMultiByte(std::wstring{value});
+    }
+
+    if (IsEmptyOrWhitespace(std::string_view{result.Name}))
+    {
+        throw ArgumentException(Localization::WSLCCLI_NetworkEmptyError(argName));
+    }
```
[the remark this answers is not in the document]
danegsta: Agreed. The parser now validates the raw `std::wstring_view` before UTF-8 conversion and only keeps a post-conversion empty-name sanity check; I also added coverage for a Unicode whitespace-only name.

beena352 (dismissed):

beena352 (dismissed):

OneBlue (approved):
LGTM, thank you !

beena352 (approved):

PR #41005, opened 2026-07-06
Add support for wslc container restart command

beena352:
Adds wslc container restart which stops and then starts a container, matching Docker CLI semantics. Supports `--signal` and `--time` options (defaults: SIGTERM, 5s timeout).

## PR Checklist

**Behavior:**
- Running container: sends signal, waits for timeout, then starts.
- Stopped/Exited container: skips stop phase, starts directly.
- Created (never-started) container: starts directly.
- `--rm` container: refused with `ERROR_INVALID_STATE` — Stop would trigger auto-delete via `OnStopped`, destroying the container.
- Not-found container: returns `WSLC_E_CONTAINER_NOT_FOUND`.

**Implementation path:** `ContainerRestartCommand` → `ContainerTasks::RestartContainers` → `ContainerService::Restart` → `IWSLCContainer::Restart` → `WSLCContainerImpl::Restart`.

**N.B. Lock design:** `WSLCContainerImpl::Restart` snapshots state/flags under a shared lock, then releases before calling `Stop` and `Start` (which each take the lock exclusively). This is intentional - `m_lock` is a non-recursive `wil::srwlock`. If the container self-exits between the snapshot and Stop, Stop no-ops on the Exited state.

**Follow-up:** To achieve full Docker parity for `--rm` containers (Docker allows restart on `--rm` by suppressing auto-remove during the restart window), a follow-up PR will add a member flag `m_restartInProgress` guarded by `wil::scope_exit` and checked in `OnStopped`. This closes the race where a natural container exit between Stop and Start would trigger auto-delete. Shipping the `--rm` guard (refuse) now is the safe default; the member-flag approach needs its own review and dedicated tests for the self-exit race.

## Validation Steps Performed
**Unit tests** (`test/windows/WSLCTests.cpp` — `ContainerRestart`):
- Restart running container → stops then starts, state returns to Running
- Restart exited container → starts, state returns to Running
- Restart created container → starts, state returns to Running
- Restart `--rm` container → refused with `ERROR_INVALID_STATE` and exact localized message
- Restart on already-deleted COM object → fails with RPC_E_DISCONNECTED (object disconnected after delete)

**E2E tests** (`test/windows/wslc/e2e/WSLCE2EContainerRestartTests.cpp`):
- `--help` output
- Restart running container (by ID)
- Restart stopped container
- Restart by name
- Container not found
- Only targeted container restarted (multi-container)
- `--rm` container refused with exact error + HRESULT
- Invalid signal (0, 32)
- Invalid timeout (non-numeric)

**Manual validation:**

- wslc container run -d --name test debian:latest sleep infinity
- wslc container restart test -t 0
- wslc container ls # confirms running
- wslc container stop test -t 0
- wslc container restart test -t 0
- wslc container ls # confirms running again

### DIFF
--- a/src/windows/wslc/commands/ContainerRestartCommand.cpp
+++ b/src/windows/wslc/commands/ContainerRestartCommand.cpp
@@ -1,0 +1,54 @@
+/*++
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    ContainerRestartCommand.cpp
+
+Abstract:
+
+    Implementation of command execution logic.
+
+--*/
+
+#include "ContainerCommand.h"
+#include "CLIExecutionContext.h"
+#include "ContainerTasks.h"
+#include "SessionTasks.h"
+#include "Task.h"
+
+using namespace wsl::windows::wslc::execution;
+using namespace wsl::windows::wslc::task;
+using namespace wsl::shared;
+
+namespace wsl::windows::wslc {
+// Container Restart Command
+std::vector<Argument> ContainerRestartCommand::GetArguments() const
+{
+    return {
+        Argument::Create(ArgType::ContainerId, std::nullopt, NO_LIMIT),
+        Argument::Create(ArgType::Signal),
+        Argument::Create(ArgType::Time),
+    };
+}
+
+std::wstring ContainerRestartCommand::ShortDescription() const
+{
+    return Localization::WSLCCLI_ContainerRestartDesc();
+}
+
+std::wstring ContainerRestartCommand::LongDescription() const
+{
+    return Localization::WSLCCLI_ContainerRestartLongDesc();
+}
+
+// clang-format off
+void ContainerRestartCommand::ExecuteInternal(CLIExecutionContext& context) const
+{
+    context
+        << ResolveSession
+        << RestartContainers;
+}
+// clang-format on
+} // namespace wsl::windows::wslc
--- a/src/windows/wslc/services/ContainerService.cpp
+++ b/src/windows/wslc/services/ContainerService.cpp
@@ -492,6 +492,14 @@ void ContainerService::Stop(Session& session, const std::string& id, StopContain
     THROW_IF_FAILED_EXCEPT(container->Stop(options.Signal, options.Timeout), WSLC_E_CONTAINER_NOT_RUNNING);
 }
 
+void ContainerService::Restart(Session& session, const std::string& id, StopContainerOptions options)
+{
+    wil::com_ptr<IWSLCContainer> container;
+    THROW_IF_FAILED(session.Get()->OpenContainer(id.c_str(), &container));
+    auto warningCallback = Microsoft::WRL::Make<WarningCallback>();
+    THROW_IF_FAILED(container->Restart(options.Signal, options.Timeout, warningCallback.Get()));
+}
+
 void ContainerService::Kill(Session& session, const std::string& id, WSLCSignal signal)
 {
     wil::com_ptr<IWSLCContainer> container;
--- a/src/windows/wslcsession/WSLCContainer.cpp
+++ b/src/windows/wslcsession/WSLCContainer.cpp
@@ -2524,6 +2524,40 @@ void WSLCContainerImpl::DisconnectFromNetwork(LPCSTR NetworkName)
         TraceLoggingValue(NetworkName, "NetworkName"));
 }
 
+void WSLCContainerImpl::Restart(WSLCSignal Signal, LONG TimeoutSeconds)
+{
+    // N.B. Snapshot under a brief shared lock and release before composing Stop/Start.
+    // m_lock is a non-recursive srwlock; Stop and Start each take it exclusively.
+    WSLCContainerState snapshotState{};
+    WSLCContainerFlags snapshotFlags{};
+    {
+        auto lock = m_lock.lock_shared();
+        snapshotState = m_state;
+        snapshotFlags = m_containerFlags;
+    }
+
+    // Refuse --rm before stopping: Stop -> OnStopped would delete the container.
+    THROW_HR_WITH_USER_ERROR_IF(
+        HRESULT_FROM_WIN32(ERROR_INVALID_STATE),
+        Localization::MessageWslcContainerRestartAutoRemove(m_id),
+        WI_IsFlagSet(snapshotFlags, WSLCContainerFlagsRm));
+
+    THROW_HR_IF_MSG(
+        HRESULT_FROM_WIN32(ERROR_INVALID_STATE),
+        snapshotState != WslcContainerStateRunning && snapshotState != WslcContainerStateExited && snapshotState != WslcContainerStateCreated,
+        "Cannot restart container '%hs', state %i",
+        m_id.c_str(),
+        snapshotState);
+
+    if (snapshotState == WslcContainerStateRunning)
+    {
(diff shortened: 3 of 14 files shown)


src/windows/wslc/commands/ContainerRestartCommand.cpp
```diff
@@ -0,0 +1,54 @@
...
+using namespace wsl::windows::wslc::execution;
+using namespace wsl::windows::wslc::task;
+using namespace wsl::shared;
+
+namespace wsl::windows::wslc {
+// Container Restart Command
+std::vector<Argument> ContainerRestartCommand::GetArguments() const
+{
+    return {
+        Argument::Create(ArgType::ContainerId, std::nullopt, NO_LIMIT),
```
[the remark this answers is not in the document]
beena352: Intentionally matches ContainerStopCommand's pattern, restart is the lifecycle sibling of stop (same -s/-t flags, same semantics). If we change restart to require the arg, we should also change stop for consistency, which is a separate discussion.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -2524,6 +2524,40 @@ void WSLCContainerImpl::DisconnectFromNetwork(LPCSTR NetworkName)
...
+    WSLCContainerState snapshotState{};
+    WSLCContainerFlags snapshotFlags{};
+    {
+        auto lock = m_lock.lock_shared();
+        snapshotState = m_state;
+        snapshotFlags = m_containerFlags;
+    }
+
+    // Refuse --rm before stopping: Stop -> OnStopped would delete the container.
+    THROW_HR_WITH_USER_ERROR_IF(
```
OneBlue, line 2540: I don't think this is the behavior we want: Docker supports restarting containers when --rm is set.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -2524,6 +2524,40 @@ void WSLCContainerImpl::DisconnectFromNetwork(LPCSTR NetworkName)
...
+    // m_lock is a non-recursive srwlock; Stop and Start each take it exclusively.
+    WSLCContainerState snapshotState{};
+    WSLCContainerFlags snapshotFlags{};
+    {
+        auto lock = m_lock.lock_shared();
+        snapshotState = m_state;
+        snapshotFlags = m_containerFlags;
+    }
+
+    // Refuse --rm before stopping: Stop -> OnStopped would delete the container.
```
OneBlue, line 2539: Locking wise, I don't think that we can drop the lock here unfortunately, otherwise another stop / start operation could arrive before we complete, which would leak to weird states

OneBlue (commented):
Unfortunately I'm not sure that Restart() can be easily implemented with our existing locking story. 

We have a work item to rethink the container state machine & locking logic to allow concurrent Stop() & Kill() calls, I think we're unfortunately going to have for that change before we can implement restart

beena352:
> Unfortunately I'm not sure that Restart() can be easily implemented with our existing locking story.
> 
> We have a work item to rethink the container state machine & locking logic to allow concurrent Stop() & Kill() calls, I think we're unfortunately going to have for that change before we can implement restart

Marking this PR as Draft and blocked by Deliverable 62987921 (“[WSL] [Containers] Rethink the WSLCContainer state machine”).
I’ll resume/rework restart once that state-machine/locking change lands, including Docker-parity  --rm  behavior.

test/windows/wslc/e2e/WSLCE2EContainerRestartTests.cpp
```diff
@@ -0,0 +1,224 @@
...
+        // Container is unaffected by the parse failure.
+        VerifyContainerIsListed(containerId, L"running");
+    }
+
+private:
+    const std::wstring WslcContainerName = L"wslc-test-container";
+    const std::wstring WslcContainerName2 = L"wslc-test-container-2";
+    const TestImage& DebianImage = DebianTestImage();
+
+    std::wstring GetHelpMessage() const
```
dkbennett, line 185: All of the help test validation has changed since this PR was created. It was changed for the better - you don't need to have precise help output validation anymore, so some bits can be cut from this PR.

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -492,6 +492,14 @@ void ContainerService::Stop(Session& session, const std::string& id, StopContain
     THROW_IF_FAILED_EXCEPT(container->Stop(options.Signal, options.Timeout), WSLC_E_CONTAINER_NOT_RUNNING);
 }
 
+void ContainerService::Restart(Session& session, const std::string& id, StopContainerOptions options)
```
dkbennett, line 495: minor nit: The ContainerService seems to have lost its alphabetical function ordering, would recommend moving the function (and header) to be as best in alphabetical as you can make it, at least as close as you can make it to other stuff, like before a cluster of S's or what similar.

src/windows/wslc/commands/ContainerRestartCommand.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+    return Localization::WSLCCLI_ContainerRestartDesc();
+}
+
+std::wstring ContainerRestartCommand::LongDescription() const
+{
+    return Localization::WSLCCLI_ContainerRestartLongDesc();
+}
+
+// clang-format off
```
dkbennett, line 46: nit: we try to avoid the clang-format off now. Other commands will put empty comments as line breaks in the context chain in ExecuteInternal instead.

See: ContainerStatsCommand.cpp as an example.   

We just do this for readability of the chain and this is a nice compromise for having that readability while keeping the clang formatter enabled.

src/windows/wslc/commands/ContainerRestartCommand.cpp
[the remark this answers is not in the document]
dkbennett: Agree we should change those too, this one should probably have required arg true for container-id. I can put out a quick PR to change those to required so they are consistent with Docker.

test/windows/wslc/e2e/WSLCE2EContainerRestartTests.cpp
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    std::wstring GetDescription() const
+    {
+        return Localization::WSLCCLI_ContainerRestartLongDesc() + L"\r\n\r\n";
+    }
+
+    std::wstring GetUsage() const
+    {
+        return L"Usage: wslc container restart [<options>] [<container-id>]\r\n\r\n";
```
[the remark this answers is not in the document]
dkbennett: This will be obsolete in the latest master changes, as we removed this help validation from the tests with the help overhaul.

dkbennett (commented):
Command implementation looks good, some minor things. Main thing to draw to attention is the change in how help is validated across the CLI from the recent help change. All other commands have been updated for it, so this one could just mimic what other container commands are now doing with regards to help tests.

src/windows/wslc/commands/ContainerRestartCommand.cpp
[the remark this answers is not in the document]
dkbennett: PR for fixing stop: https://github.com/microsoft/WSL/pull/41108

So you can change this one too.

kvega005:
This should wait until #41140 is complete.

beena352:
Closing this as I have opened a new PR: https://github.com/microsoft/WSL/pull/41435

PR #41009, opened 2026-07-06
Fix flaky E2E network list test assertion

beena352:
Add missing network cleanup for DebianImage.Name in WSLCE2EInspectTests to prevent leaked networks from causing flaky count assertions in WSLCE2E_Network_List_JsonFormat.

## PR Checklist

The `WSLCE2E_Network_List_JsonFormat` test intermittently fails with `Values (2, 3)` because `WSLCE2E_Inspect_Image_PriorityOverNetwork` creates a network named "debian" (DebianImage.Name) that isn't covered
by InspectTests' MethodSetup/ClassCleanup.
When the test process terminates abnormally (or scope_exit doesn't run), this network persists in Docker. On the next session start,
WSLCSession::RecoverExistingNetworks() re-imports any docker network with the wslc managed label, causing network list to return 3 instead of 2.

Fix: Add EnsureNetworkDoesNotExist(DebianImage.Name) to InspectTests'
MethodSetup and ClassCleanup, matching the existing pattern for
WslcNetworkName.

## Validation Steps Performed
Verified the test passes locally with the fix applied

### DIFF
--- a/test/windows/wslc/e2e/WSLCE2EInspectTests.cpp
+++ b/test/windows/wslc/e2e/WSLCE2EInspectTests.cpp
@@ -36,6 +36,7 @@ class WSLCE2EInspectTests
         EnsureContainerDoesNotExist(WslcContainerName);
         EnsureContainerDoesNotExist(DebianImage.Name);
         EnsureNetworkDoesNotExist(WslcNetworkName);
+        EnsureNetworkDoesNotExist(DebianImage.Name);
         EnsureImageIsDeleted(DebianImage);
         return true;
     }
@@ -45,6 +46,7 @@ class WSLCE2EInspectTests
         EnsureContainerDoesNotExist(WslcContainerName);
         EnsureContainerDoesNotExist(DebianImage.Name);
         EnsureNetworkDoesNotExist(WslcNetworkName);
+        EnsureNetworkDoesNotExist(DebianImage.Name);
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


OneBlue (commented):
Do we know where the leak is coming from ? Ideally we should fix the leak so this doesn't potentially cause other failures

test/windows/wslc/e2e/WSLCE2EInspectTests.cpp
```diff
@@ -36,6 +36,7 @@ class WSLCE2EInspectTests
         EnsureContainerDoesNotExist(WslcContainerName);
         EnsureContainerDoesNotExist(DebianImage.Name);
         EnsureNetworkDoesNotExist(WslcNetworkName);
+        EnsureNetworkDoesNotExist(DebianImage.Name);
```
OneBlue, line 39: This doesn't look right. We're looking for a network but passing in an image name

OneBlue (changes requested):

test/windows/wslc/e2e/WSLCE2EInspectTests.cpp
OneBlue: Ok I see what's happening from the description. Sorry for misunderstanding, I got a bit confused reading this !

OneBlue (approved):

PR #41026, opened 2026-07-08
Fix: string conversion pointer advancement, DNS send loop infinite loop, interface stack traversal, and thread-safety annotation errors

diaoyunxi:
## Overview

This PR fixes 4 defects found in the WSL codebase across shared string utilities, Linux DNS tunneling server, and Windows networking support code.

### 1. Fix string conversion pointer advancement bug in `stringshared.h` (Critical)

**Issue**: The `MultiByteToWide` and `WideToMultiByte` functions on Linux use `mbsrtowcs`/`wcsrtombs` for string conversion. These functions advance the source pointer (second parameter) after the call. After the first pass (length calculation), the source pointer has been advanced to the end of the string. The second pass (actual conversion) then operates on an empty string, resulting in an empty conversion result.

**Impact**: All non-empty string wide-character conversions on the Linux side return empty strings.

**Fix**: Use a temporary pointer variable `src` for the length calculation to avoid modifying the original pointer. Reset `src` to the original string before the second conversion pass. Also reset `mbstate_t` to support stateful encodings, and check the return value for conversion errors.

### 2. Fix infinite loop risk in DNS UDP/TCP send loops in `DnsServer.cpp` (Critical)

**Issue**: The send loops in `HandleUdpDnsResponse` and `HandleTcpDnsResponse` do not exit when `sendto`/`write` returns 0, causing an infinite loop.

**Impact**: The DNS server could hang permanently.

**Fix**: Add a `bytesSent == 0` guard. For UDP, use `return` to make the failure explicit. For TCP, remove the broken connection from `m_tcpConnectionContexts` to avoid subsequent writes to a dead file descriptor.

### 3. Fix interface stack traversal infinite loop in `WslCoreNetworkingSupport.cpp` (Critical)

**Issue**: The BFS traversal of the network interface stack does not maintain a visited set. If circular references exist in the interface stack table, this leads to an infinite loop and CPU exhaustion.

**Fix**: Add `std::unordered_set<IF_INDEX> visitedInterfaces` to track visited interfaces and skip already-visited ones before enqueueing.

### 4. Fix thread-safety annotation errors in `DnsServer.h`

**Issue**: `m_currentTcpConnectionId` and `m_tcpConnectionContexts` are TCP-related members protected by `m_tcpLock`, but their `_Guarded_by_` annotations incorrectly reference `m_udpLock`.

**Impact**: Static analysis tools produce false positives; maintainers may misunderstand lock protection scope.

**Fix**: Change `_Guarded_by_(m_udpLock)` to `_Guarded_by_(m_tcpLock)`.

## Files Changed

- `src/shared/inc/stringshared.h` — String conversion function fix
- `src/linux/init/DnsServer.cpp` — DNS send loop guard
- `src/linux/init/DnsServer.h` — Thread-safety annotation correction
- `src/windows/common/WslCoreNetworkingSupport.cpp` — Interface stack traversal safety

### DIFF
--- a/src/linux/init/DnsServer.cpp
+++ b/src/linux/init/DnsServer.cpp
@@ -131,6 +131,12 @@ try
     {
         int bytesSent = Syscall(
             sendto, m_udpSocket.get(), dnsBuffer.data() + totalBytesSent, bufferSize - totalBytesSent, 0, reinterpret_cast<sockaddr*>(&remoteAddr), sizeof(remoteAddr));
+        // Guard against infinite loop: if sendto returns 0 or a negative value, break out
+        if (bytesSent <= 0)
+        {
+            GNS_LOG_ERROR("UDP DNS sendto failed or returned 0, aborting send loop. bytesSent: {}", bytesSent);
+            break;
+        }
         totalBytesSent += bytesSent;
     }
 }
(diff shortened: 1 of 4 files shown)


OneBlue (commented):
Thank you for doing this, would you mind translating the PR description and title to English ?

PR #41043, opened 2026-07-09
Add WSLC VM start/stop plugin hooks

benhillis:
## Summary

The WSLC plugin API only exposed session-level hooks (`OnSessionCreated` / `OnSessionStopping`), which assumed session == running VM. With idle termination (#40781) the VM is created lazily, torn down when idle, and transparently recreated while the session persists, so plugins had no notification of the actual VM lifecycle.

This adds VM-level hooks that fire on every VM (re)start and teardown, alongside the once-per-session hooks:

- `OnWslcVmStarted` and `OnWslcVmStopping` in `WslPluginApi.h` (`WSLPluginHooksV1.WslcVmStarted` / `WslcVmStopping`)
- `IWSLCPluginNotifier::OnVmStarted` / `OnVmStopping` (service and per-user session ship and rebuild together, so no new IID)
- Dispatch through `PluginManager` and the runtime hooks in `WSLCSessionRuntime`

## Behavior

- Both hooks are best-effort: errors are logged and ignored, so a transparent idle restart never surfaces plugin errors on unrelated user commands.
- `OnWslcVmStarted` fires after releasing the runtime exclusive lock, so a plugin may reentrantly call back into the session (for example `WSLCCreateProcess`) without deadlocking. The triggering `VmLease` holds an activity reference across the gap, so idle teardown cannot race the VM down.
- `OnWslcVmStopping` fires under the lock during teardown, gated on a fire-once flag so a failed bring-up emits no spurious stopping.

## Tests

- New `PluginTests::WslcVmRestart` drives first start, forced idle teardown via `TriggerIdleTermination`, and lazy restart, asserting the full hook sequence and proving the reentrant `WSLCCreateProcess` from `OnWslcVmStarted` is deadlock-free.
- Existing WSLC plugin tests are unaffected (VM hook logging is gated to the new test type). All 5 WSLC plugin tests pass locally.

## Note

Based on top of #40781 (idle-terminate WSLC session VMs), which is not yet merged. The diff below includes that PR until it merges; this branch should be rebased once #40781 lands.


benhillis:
Closing in favor of folding these VM plugin-hook changes directly into #40781, since they depend on that PR's lazy-VM lifecycle (WSLCSessionRuntime + TriggerIdleTermination) and can't be cleanly separated.

PR #41057, opened 2026-07-11
Start the plan9 server on WSL2 even when automount and mountFsTab are…

### LINKED ISSUE #41056: Plan9 file server is never started on WSL2 when both automount.enabled and automount.mountFsTab are false, making \\wsl.localhost inaccessible
### Windows Version

Microsoft Windows [Version 10.0.19045.6466]

### WSL Version

2.6.3.0 (the root-cause analysis below is based on current master, where the logic is unchanged)

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.6.87.2-1

### Distro Version

Ubuntu 24.04.2 LTS

### Other Software

None

### Repro Steps

1. In a WSL2 distro, set `/etc/wsl.conf` to:
```ini
   [automount]
   enabled = false
   mountFsTab = false
```
 
2. Run `wsl --shutdown` to apply the change.
3. Start the distro again and try to access `\\wsl.localhost\<distro>` (or `\\wsl$\<distro>`) from Windows.

### Expected Behavior

The Linux filesystem is accessible from Windows. Per the documentation, `automount.enabled` and `automount.mountFsTab` only control automatic mounting of Windows drives and the processing of `/etc/fstab`. No coupling between these options and the 9p file server is documented.

### Actual Behavior

`\\wsl.localhost\<distro>` and `\\wsl$\<distro>` are inaccessible:
 
```
PS C:\Users\Administrator> ls \\wsl.localhost\Ubuntu\home
ls : Cannot find path '\\wsl.localhost\Ubuntu\home' because it does not exist.
    + CategoryInfo          : ObjectNotFound: (\\wsl.localhost\Ubuntu\home:String) [Get-ChildItem], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.GetChildItemCommand
```
 
(Original console output was localized; this is the equivalent English message.)
 
Inside the distro:
 
- `ps -ef | grep -i plan9` shows no plan9 process.
- `dmesg` contains no plan9-related errors — init *skips* starting the server rather than failing to start it (see root cause below).
Setting either of the two options back to `true` makes the plan9 process appear and restores `\\wsl.localhost` access.
 
### Root cause
 
In `ConfigInitializeInstance()` (`src/linux/init/config.cpp`), the call to `StartPlan9Server()` is guarded by:
 
```cpp
unsigned int Plan9Port = LX_INIT_UTILITY_VM_INVALID_PORT;
if ((WI_IsFlagClear(Config.FeatureFlags.value(), LxInitFeatureDisable9pServer)) && (Config.Plan9Enabled) &&
    (Config.AutoMount || Config.MountFsTab))
{
    std::tie(Plan9Port, Config.Plan9ControlChannel) = StartPlan9Server(Plan9SocketPath, Config);
}
```
 
When both `AutoMount` and `MountFsTab` are `false`, the branch is skipped entirely, and `Plan9Port` is sent back to wslservice as `LX_INIT_UTILITY_VM_INVALID_PORT` in the `LX_INIT_CONFIGURATION_INFORMATION_RESPONSE`, so there is no server for the Windows side (p9rdr.sys via wslservice.exe) to connect to.
 
The comment above this block explains the rationale for the guard: the plan9 server needs a DrvFs mount for its socket file, so either fstab or automount must be enabled for that mount to be available. **That rationale only holds for WSL1.** On WSL2 (utility VM mode), `StartPlan9Server()` (`src/linux/init/plan9.cpp`) takes the vsock branch and never touches the socket path at all:
 
```cpp
if (UtilIsUtilityVm())
{
    sockaddr_vm address;
    server.reset(UtilBindVsockAnyPort(&address, (SOCK_STREAM | SOCK_NONBLOCK)));
    ...
    result = address.svm_port;
}
else
{
    // Translate the socket path (store a copy for unlinking on shutdown).
    translatedSocketPath = TranslatePath(const_cast<char*>(socketWindowsPath));
    // Create the server socket.
    server = CreateUnixServerSocket(translatedSocketPath.c_str());
}
```
 
So on WSL2 the server has no dependency on any DrvFs mount, and the guard is a WSL1-era constraint that unintentionally disables the 9p server on WSL2. Note also that a dedicated setting already exists for intentionally disabling the server (`Config.Plan9Enabled`), so whether the server runs shouldn't be implicitly decided by two mount-related options — a side effect that isn't documented anywhere.
 
### Proposed fix
 
Exempt WSL2 from the mount requirement, keeping WSL1 behavior unchanged:
 
```cpp
(UtilIsUtilityVm() || Config.AutoMount || Config.MountFsTab)
```
 
and update the comment to note that the DrvFs dependency only applies to WSL1. This is backward compatible: WSL1 behavior is unchanged, WSL2 is fixed.
 
Happy to submit a PR if this approach looks right.
 
### Workaround
 
For anyone hitting this: keep `mountFsTab = true` together with `enabled = false`. As long as `/etc/fstab` contains no drvfs entries, `mount -a` is effectively a no-op, so no Windows drives get mounted, but the plan9 server starts normally.

### Diagnostic Logs

An ETW trace was recorded with `diagnostics/collect-wsl-logs.ps1` across a cold start of the distro (`wsl --shutdown` → start) followed by a failing `\\wsl.localhost` access. Since the repro above takes under a minute, I'm including only the relevant excerpt here for privacy; **the full trace is available on request**.
 
Excerpt from `logs.etl`, decoded with `tracerpt logs.etl -o logs.csv -of CSV -y` (provider `WslCorePort`). The hosts-file content embedded in the message is redacted and lines are re-wrapped for readability; otherwise unmodified:
 
```
SentMessage (wslservice → init):
  Header = {MessageType = LxInitMessageInitialize MessageSize = 591 SequenceNumber = 1}
  HostnameOffset = DESKTOP-DTB60UU  DomainnameOffset = localdomain
  WindowsHostsOffset = <redacted>
  Plan9SocketOffset =  TimezoneOffset = Asia/Taipei
  DrvFsVolumesBitmap = 12  DrvFsDefaultOwner = 1000  FeatureFlags = 0  DrvfsMount = 2
 
ReceivedMessage (init → wslservice):
  Header = {MessageType = LxInitMessageInitializeResponse MessageSize = 61 SequenceNumber = 2}
  Plan9Port = 4294967295  DefaultUid = 1000  InteropPort = 3162867010
  SystemdEnabled = 0  PidNamespace = 4026532221  FlavorIndex = ubuntu  VersionIndex = 24.04
```
 
`Plan9Port = 4294967295` is `0xFFFFFFFF`, i.e. `LX_INIT_UTILITY_VM_INVALID_PORT`. With `FeatureFlags = 0` (`LxInitFeatureDisable9pServer` **not** set) and the 9p server left enabled (default) in `/etc/wsl.conf`, the only clause of the guard that can have skipped `StartPlan9Server()` is `(Config.AutoMount || Config.MountFsTab)` — consistent with the analysis above. Note also the empty `Plan9SocketOffset`, confirming that WSL2 does not use a socket path (it uses vsock).
 
(The trace also shows an invalid `Plan9Port` for the system distro, `FlavorIndex = mariner` — that one is expected: unlike the Ubuntu distro, its `LxInitMessageInitialize` carries non-zero `FeatureFlags = 20`, and the system distro is never exposed via `\\wsl$`.)

Lsseeker:
Fixes #41056

On WSL2, when `/etc/wsl.conf` has both `automount.enabled = false` and
`automount.mountFsTab = false`, init never starts the plan9 file server, so
`\\wsl.localhost\<distro>` / `\\wsl$\<distro>` are inaccessible.

The guard in `ConfigInitializeInstance()` requires
`(Config.AutoMount || Config.MountFsTab)` because the server "requires a DrvFs
mount for the socket file". That rationale only applies to WSL1: on WSL2,
`StartPlan9Server()` binds an hvsocket (`UtilBindVsockAnyPort`) and never uses
the socket path — the initialize message carries an empty `Plan9SocketOffset`
(see the trace excerpt in #41056).

This change exempts WSL2 (`UtilIsUtilityVm()`) from the mount requirement:

- WSL1 behavior is unchanged.
- WSL2 with this configuration now takes exactly the same code path as the
  default configuration (where automount is enabled and the server starts).
- Anyone intentionally relying on this option combination to disable the 9p
  server can use the dedicated setting (`Config.Plan9Enabled`) instead.

**Testing:** verified by code inspection and the ETW trace attached to #41056
(`LxInitMessageInitializeResponse` carries
`Plan9Port = LX_INIT_UTILITY_VM_INVALID_PORT` while `FeatureFlags = 0`). I
don't currently have a local build environment, so this has not been build- or
runtime-tested on my side — happy to add a regression test if you can point me
at the right harness.

### DIFF
--- a/src/linux/init/config.cpp
+++ b/src/linux/init/config.cpp
@@ -845,21 +845,16 @@ try
     }
 
     //
-//
--    // Run the Plan 9 server. This requires a DrvFs mount for the socket file,
--    // so either fstab or automount must be enabled to have a chance for the
--    // mount to be available.
-+    // Run the Plan 9 server. On WSL1 this requires a DrvFs mount for the socket
-+    // file, so either fstab or automount must be enabled to have a chance for
-+    // the mount to be available. WSL2 serves over an hvsocket and has no such
-+    // dependency.
-     //
-     // N.B. Failure to start the server is non-fatal.
-     //
-     unsigned int Plan9Port = LX_INIT_UTILITY_VM_INVALID_PORT;
-     if ((WI_IsFlagClear(Config.FeatureFlags.value(), LxInitFeatureDisable9pServer)) && (Config.Plan9Enabled) &&
--        (Config.AutoMount || Config.MountFsTab))
-+        (UtilIsUtilityVm() || Config.AutoMount || Config.MountFsTab))
+    // Run the Plan 9 server. On WSL1 this requires a DrvFs mount for the socket
+    // file, so either fstab or automount must be enabled to have a chance for
+    // the mount to be available. WSL2 serves over an hvsocket and has no such
+    // dependency.
+    //
+    // N.B. Failure to start the server is non-fatal.
+    //
+    unsigned int Plan9Port = LX_INIT_UTILITY_VM_INVALID_PORT;
+    if ((WI_IsFlagClear(Config.FeatureFlags.value(), LxInitFeatureDisable9pServer)) && (Config.Plan9Enabled) &&
+        (UtilIsUtilityVm() || Config.AutoMount || Config.MountFsTab))
     {
         std::tie(Plan9Port, Config.Plan9ControlChannel) = StartPlan9Server(Plan9SocketPath, Config);
     }
--- a/src/linux/init/config.cpp
+++ b/src/linux/init/config.cpp
@@ -845,16 +845,21 @@ try
     }
 
     //
-    // Run the Plan 9 server. This requires a DrvFs mount for the socket file,
-    // so either fstab or automount must be enabled to have a chance for the
-    // mount to be available.
-    //
-    // N.B. Failure to start the server is non-fatal.
-    //
-
-    unsigned int Plan9Port = LX_INIT_UTILITY_VM_INVALID_PORT;
-    if ((WI_IsFlagClear(Config.FeatureFlags.value(), LxInitFeatureDisable9pServer)) && (Config.Plan9Enabled) &&
-        (Config.AutoMount || Config.MountFsTab))
+//
+-    // Run the Plan 9 server. This requires a DrvFs mount for the socket file,
+-    // so either fstab or automount must be enabled to have a chance for the
+-    // mount to be available.
++    // Run the Plan 9 server. On WSL1 this requires a DrvFs mount for the socket
++    // file, so either fstab or automount must be enabled to have a chance for
++    // the mount to be available. WSL2 serves over an hvsocket and has no such
++    // dependency.
+     //
+     // N.B. Failure to start the server is non-fatal.
+     //
+     unsigned int Plan9Port = LX_INIT_UTILITY_VM_INVALID_PORT;
+     if ((WI_IsFlagClear(Config.FeatureFlags.value(), LxInitFeatureDisable9pServer)) && (Config.Plan9Enabled) &&
+-        (Config.AutoMount || Config.MountFsTab))
++        (UtilIsUtilityVm() || Config.AutoMount || Config.MountFsTab))
     {
         std::tie(Plan9Port, Config.Plan9ControlChannel) = StartPlan9Server(Plan9SocketPath, Config);
     }


Lsseeker:

benhillis (approved):
Thanks for making this change, this looks correct to me.

benhillis (approved):

PR #41070, opened 2026-07-13
Add --ip, --link, --link-local-ip, --driver-opt, --network-alias flag…

beena352:
Adds endpoint configuration flags to `wslc network connect` and fixes the connect/disconnect mode gate to reject `container:*` in addition to `host` and `none`.

## PR Checklist

### New flags on `wslc network connect`
`--network-alias`, `--ip`, `--link`, `--link-local-ip`, `--driver-opt`. All flags are repeatable except `--ip` (one static IP per network). Validation includes: unknown setting key rejection, malformed IP rejection, empty alias/link rejection, malformed/blank/duplicate driver-opt rejection. Empty driver-opt values (`key=`) are accepted intentionally to match Docker CLI behavior.

The create path (`wslc container create --endpoint-settings`) also picks up the same parser, so `IPAddress`, `Links`, `LinkLocalIPs`, and `DriverOpts` now flow through to the Docker create request for the primary network (previously only `Aliases` was forwarded).

### Bug fix: container:* mode gate
Previously the connect/disconnect gate only checked for `"host"` and `"none"` modes. Containers using `container:<name>` (shared network namespace) were incorrectly allowed through. Now uses `NetworkModeAllocatesVmPorts()` which covers all three cases. Added new localized error message `MessageWslcNetworkModeNoAdditionalNetworks`.

Test changes

- Replaced stale `NetworkAliasCreateTest` block (expected `E_NOTIMPL` for `IPAddress`) with `BogusKey` / `E_INVALIDARG` unknown-key test.
- Added positive create-path roundtrip tests: combo (Aliases + IPAddress + LinkLocalIPs + DriverOpts) and Links (2 containers on user-defined bridge).
- Added `container:<name>` mode rejection test (substring match - mode resolves to container ID internally).
- Added connect-while-stopped positive test (Docker allows attach in exited state).
- Added driver-opt edge-case tests: `key=value=extra` preserves everything after first `=`; `key=` (empty value) accepted per Docker CLI parity.
- Fixed E2E `WSLCE2E_Network_Connect_NoneMode_Rejected` pattern to match new error message.
- Added E2E tests for `--driver-opt` and `--link` flag roundtrips via `wslc network connect` + inspect.

## Validation Steps Performed

1. Built and ran all `WSLCTests::NetworkAliasCreateTest` sub-blocks - unknown key rejection, positive combo roundtrip, and Links roundtrip all pass.
2. Built and ran `WSLCTests::ContainerNetworkEndpointSettingsUnknownKeyTest` and `NetworkConnectEndpointSettingsTest` - all sub-blocks pass.
3. Built and ran E2E tests: `WSLCE2E_Network_Connect_DriverOpt_RoundTrips`, `WSLCE2E_Network_Connect_Link_RoundTrips`.
4. Verified `wslc network connect --help` renders all new flags with correct padding and descriptions.


localization/strings/fr-FR/Resources.resw
```diff
@@ -2383,10 +2379,25 @@ Pour plus d’informations sur la confidentialité de ce produit, veuillez consu
...
+  <data name="MessageWslcIpAddressSingleValue" xml:space="preserve">
+    <value>Only one IP address may be specified for network '{}'.</value>
     <comment>{FixedPlaceholder="{}"}Command line arguments, file names and string inserts should not be translated</comment>
   </data>
+  <data name="MessageWslcLinkEmpty" xml:space="preserve">
+    <value>Network link cannot be empty.</value>
+  </data>
   <data name="MessageWslcAliasEmpty" xml:space="preserve">
     <value>L’alias réseau ne peut pas être vide.</value>
   </data>
```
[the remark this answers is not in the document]
beena352: Correct, I'll remove it.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -245,22 +244,83 @@ EndpointConfig ResolvePrimaryEndpointConfig(const KeyValuePair* settings, ULONG
...
+        std::map<std::string, std::string> driverOpts;
+        for (const auto& entry : it->second)
+        {
+            const auto separator = entry.find('=');
+            THROW_HR_WITH_USER_ERROR_IF(
+                E_INVALIDARG, Localization::MessageWslcDriverOptInvalid(entry), separator == std::string::npos || separator == 0);
+
+            auto key = entry.substr(0, separator);
+            auto value = entry.substr(separator + 1);
+            THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessageWslcDriverOptInvalid(entry), isBlank(key));
```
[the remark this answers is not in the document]
beena352: Intentional - Docker CLI allows empty driver-opt values ( key= )

src/windows/wslc/commands/NetworkConnectCommand.cpp
```diff
@@ -29,6 +29,11 @@ std::vector<Argument> NetworkConnectCommand::GetArguments() const
     return {
         Argument::Create(ArgType::NetworkName, true),
         Argument::Create(ArgType::ContainerId, true),
+        Argument::Create(ArgType::NetworkAlias, false, NO_LIMIT),
```
dkbennett: These should be alphabetical order after positional arguments (this affects how they will be ordered in the help output).

So NetworkName & ContainerId are correct as-is since they are positional, but then it should be DriverOpt, IpAddress, etc. A little bizarre and I should probably add sorting to the help output so it doesn't matter, but nice to be consistent either way.

src/windows/wslc/tasks/NetworkTasks.cpp
```diff
@@ -234,9 +234,36 @@ void ConnectNetwork(CLIExecutionContext& context)
     WI_ASSERT(context.Args.Contains(ArgType::NetworkName));
     WI_ASSERT(context.Args.Contains(ArgType::ContainerId));
 
-    const auto networkName = WideToMultiByte(context.Args.Get<ArgType::NetworkName>());
-    const auto containerId = WideToMultiByte(context.Args.Get<ArgType::ContainerId>());
-    NetworkService::Connect(context.Data.Get<Data::Session>(), networkName, containerId);
+    models::ConnectNetworkOptions options{};
```
dkbennett, line 239: We could be more compositional here with the tasks: e.g. have a new task that just sets network options, which is then consumed by Connect. So if we have other network commands that also need to set options we avoid duplicate code.

A good example would be SetContainerOptionsFromArgs, where we have a task which just sets container options based on the args present and adds it to context data, which then other container tasks consume as needed. I think this should be similar for more modularity in future network commands and code reuse.

dkbennett (commented):
Couple minor refactor-ish things, nice tests and overall work looks good!

src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -51,6 +51,7 @@ _(DNSOption,      "dns-option",          NO_ALIAS,          Kind::Value,       L
 _(DNSSearch,      "dns-search",          NO_ALIAS,          Kind::Value,       Localization::WSLCCLI_DNSSearchArgDescription()) \
 _(Domainname,     "domainname",          NO_ALIAS,          Kind::Value,       Localization::WSLCCLI_DomainnameArgDescription()) \
 _(Driver,         "driver",              L"d",              Kind::Value,       Localization::WSLCCLI_DriverOptionDescription()) \
+_(DriverOpt,      "driver-opt",          NO_ALIAS,          Kind::Value,       Localization::WSLCCLI_DriverOptArgDescription()) \
```
AmelBawa-msft, line 54: Nit: Is `DriverOpt` and `Options` duplicates? If just the description differs, I believe there is a way to override the description in the command definition.

localization/strings/en-US/Resources.resw
AmelBawa-msft: This can probably be removed

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -245,22 +244,83 @@ EndpointConfig ResolvePrimaryEndpointConfig(const KeyValuePair* settings, ULONG
...
+        }
+        config.Links = std::move(it->second);
+    }
+
+    if (auto it = parsed.find("LinkLocalIPs"); it != parsed.end())
+    {
+        for (const auto& address : it->second)
+        {
+            in_addr parsedAddress{};
+            ParseIpv4Address(address.c_str(), parsedAddress);
```
AmelBawa-msft, line 297: Do we currently only support IPv4?

dkbennett: I have tasks for ipv6 so I believe that is the case at present.

src/windows/wslc/arguments/ArgumentDefinitions.h
beena352: These target different Docker API fields,  --opt  sets driver options on network/volume create (IPAM/driver config), while  --driver-opt  sets per-endpoint driver options on network connect (EndpointConfig.DriverOpts). Different semantics, different API paths.

dkbennett (approved):

benhillis (approved):

PR #41071, opened 2026-07-13
Bump Linux SDK to 1.23.0 and guard char->wide formatters to Windows

benhillis:
## Summary

The new Linux SDK (1.23.0) ships libc++ 21, which now provides *deleted* `std::formatter` specializations for `char` -> `wchar_t` cross-encoding formatting per C++23 [format.formatter.spec]. Our own char->wide specializations in `stringshared.h` collide with those when building the Linux components, causing redefinition errors.

These specializations are only used by the Windows components (MSVC leaves the primary formatter template deleted-but-undefined, so our specializations are legal there and are relied upon). This change restricts them to `#ifdef WIN32`, which resolves the collision on Linux while leaving Windows behavior unchanged.

## Changes

- Bump `Microsoft.WSL.LinuxSdk` 1.20.0 -> 1.23.0
- Guard the char->wchar_t `std::formatter` specializations behind `WIN32`

### DIFF
--- a/src/shared/inc/stringshared.h
+++ b/src/shared/inc/stringshared.h
@@ -931,6 +931,11 @@ struct std::formatter<std::source_location, wchar_t>
     }
 };
 
+// char -> wchar_t formatting is only used by the Windows components. libc++ (used to
+// build the Linux components) now provides these as deleted specializations per C++23
+// [format.formatter.spec], which would collide, so restrict them to Windows.
+#ifdef WIN32
+
 template <>
 struct std::formatter<char*, wchar_t>
 {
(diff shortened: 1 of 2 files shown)


src/shared/inc/stringshared.h
```diff
@@ -931,6 +931,11 @@ struct std::formatter<std::source_location, wchar_t>
     }
 };
 
+// char -> wchar_t formatting is only used by the Windows components. libc++ (used to
```
OneBlue, line 934: I'm sad to learn about this, but I guess this will be a problem for later

OneBlue (approved):

PR #41072, opened 2026-07-13
Conformant libc++21 formatter refactor (boundary widening) [WIP]

benhillis:
**Draft / WIP.** Follow-up to #41071 (which ships the minimal `#ifdef WIN32` guard as the shipping fix).

This branch explores the fully **conformant** alternative: instead of keeping the `formatter<char*, wchar_t>` / `formatter<std::string, wchar_t>` specializations (deleted in C++20/23 [format.formatter.spec] and shipped as deleted specializations in libc++ 20+), it removes them and widens narrow arguments explicitly.

### Approach
The key insight is a single boundary helper rather than per-call-site churn:

- `wsl::shared::string::LocalizationArg()` (stringshared.h) widens narrow string args (`const char*`, `char[N]`, `std::string`, `std::string_view`) to `wchar_t` on Windows, and is an identity forward on Linux (where messages are already narrow).
- The Localization generator (`generateLocalizationHeader.ps1`) routes every `Message*()` argument through it, so `std::vformat(message, ARGS(LocalizationArg(arg)...))`.

This fixes **all product code** (common, wslc, wslcsession, service.exe) through one generator change.

### Remaining work
- Direct `std::format(L"...", narrowArg)` sites that do not go through `Localization` (primarily `test/windows` unit tests) still rely on the removed specializations and need conversion.
- Decide whether the conformant path is worth the test-side churn versus the shipped guard in #41071.

### DIFF
--- a/packages.config
+++ b/packages.config
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="CommunityToolkit.Mvvm" version="8.4.0" />
   <package id="CommunityToolkit.WinUI.Animations" version="8.2.250402" />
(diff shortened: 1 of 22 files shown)


benhillis:
Closing in favor of #41071. The fully-conformant approach (removing the char->wide formatter specializations and widening at every boundary) touches ~22 files, because MSVC and libc++ 21 both delete those specializations, so every direct wide std::format narrow call site must change. The minimal Windows-guarded formatter in #41071 is the better tradeoff.

PR #41077, opened 2026-07-14
wslc: idle-terminate inactive per-user session VMs

benhillis:
Idle-terminates per-user WSLC session VMs after a period of inactivity, so an idle container session releases its utility VM instead of holding memory indefinitely. Includes the supporting refactor that makes the VM lifecycle safe to reason about.

## What this does
- Adds an idle timer to each per-user WSLC session VM. When no container operation has run for the configured grace period, the VM is torn down; the next operation transparently restarts it.
- Container operations take an activity lease (`Session::BeginContainerOperation`) so the VM cannot be idle-terminated mid-operation.
- Adds `OnVmStarted` / `OnVmStopping` plugin hooks, paired correctly across start/stop and VM-restart.

## VM lifecycle hardening
- Extracts VM state management into a dedicated `WSLCSessionRuntime` state machine (start / stop / teardown / spontaneous-exit), replacing ad-hoc state spread across `WSLCSession`.
- Fixes several concurrency issues found during review:
  - reentrant deadlock in start/stop notifications,
  - idle-teardown vs. activity race,
  - stopping-only hook pairing,
  - two VM-lifecycle races found via multi-model review (dead-VM re-pair in the idle-timer abort path; terminating-gate ordering in `EnsureVmRunning`).
- Skips guest volume unmount when the VM has already exited.

## Encapsulation cleanup
- Narrows the `WSLCSession` / `WSLCSessionRuntime` seam: replaces ~9 accessors that handed out mutable internals with intent-revealing methods (`TryLockExclusive`, `SetContainerdProcess`, `Set/Reset/Signal/IsDockerdReady`, `SetSwapVhdPath`), and drops dead accessors.
- Marks the per-operation activity leases `[[maybe_unused]]` to document that they are held for their scope lifetime.

## Testing
Full build (x64 + arm64) + package, then locally: idle-termination (5/5), WslcVmRestart, VmKillTerminatesSession, StuckVmTermination, reentrant-mount plugin hook, and PluginTests::WslcSuccess all pass.


dkbennett (commented):
This looks like a lot of goodness!  Would like to get it in soon so we have a lot of usage time to find any corner cases in the lifetime management that might come about. 

This seems especially useful for third party developers utilizing the WSLC API to use containers in their apps, dont want those sessions piling up long after the app is done running. Though that could still happen if they add in a long-running container that could persist after the app is done. That might be intentional but perhaps we can enhance this later with some additional session termination behavior to try to prevent long-running third party sessions that aren't actually necessary but might be held open by a container they created.

src/windows/service/exe/WSLCSessionManager.h
```diff
@@ -129,6 +129,16 @@ class WSLCSessionManagerImpl
             wil::com_ptr<IWSLCSession> lockedSession;
             if (FAILED_LOG(entry.Ref->OpenSession(&lockedSession)))
             {
+                // Session is gone. Notifying plugins fires OnWslcSessionStopping, which is a plugin
+                // callback and must not nest inside another plugin callback (ExecutionContext forbids
+                // it). If a plugin called back into WSLC from within a notification (e.g. creating a
+                // process from OnWslcVmStopping), defer this lazy notify + prune to a later, non-
+                // reentrant pass rather than firing OnWslcSessionStopping on top of the current one.
+                if (InPluginNotificationContext())
```
OneBlue: This is feels bit dangerous because the execution context is originally designed as a mechanism to report errors to users and some code paths might not have one at all. 

If we want to behave differently when the call comes from a plugin notifications, I would recommend passing in a flag to determine when that's the case

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -3459,26 +3564,47 @@ CATCH_RETURN();
...
     {
+        // A kept-alive wrapper's cached state already matches dockerd's (reconciled in OnVmTornDown).
+        // Leave it (and its COM references) in place, just re-register its ports on the restarted VM.
+        if (auto existing = m_containers.find(dockerContainer.Id); existing != m_containers.end())
+        {
+            // Isolate a port re-reservation conflict to this container so it can't fail the lazy start
+            // for every client: log, warn and continue, mirroring the Open() failure path below.
+            try
+            {
+                // Honor --rm for a survivor forced to Exited in OnVmTornDown (delete deferred): remove
```
OneBlue: Is this actually possible ? If the container was still running, then we wouldn't terminate the VM right ?

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -3459,26 +3564,47 @@ CATCH_RETURN();
...
+                {
+                    m_containers.erase(existing);
+                    continue;
+                }
+
+                existing->second->RecoverPorts(dockerContainer);
+            }
+            catch (...)
+            {
+                LOG_CAUGHT_EXCEPTION_MSG("Failed to recover ports for container: %hs", dockerContainer.Id.c_str());
```
OneBlue: nit: The error message references ports, but the exception could come from something else

src/windows/wslcsession/WSLCSessionRuntime.cpp
```diff
@@ -0,0 +1,848 @@
...
+    {
+        m_runtime->EnsureVmRunning();
+
+        m_lock = m_runtime->m_lock.lock_shared();
+
+        if (m_runtime->m_vmState.load() == VmState::Running)
+        {
+            break;
+        }
+
```
OneBlue, line 834: nit: This might be worth an ETL even, just to help us debug if we ever get stuck in that loop

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -614,11 +732,11 @@ try
         TraceLoggingValue(entry.c_str(), "Content"),
         TraceLoggingValue(m_displayName.c_str(), "Name"));
 
-    if (!m_dockerdReadyEvent.is_signaled())
+    if (!m_runtime.IsDockerdReady())
```
OneBlue: nit: We should probably move this entire method to the runtime, this would simplify things a bit

src/windows/wslcsession/WSLCSessionRuntime.cpp
```diff
@@ -0,0 +1,848 @@
...
+    {
+        m_ioRelay->Stop();
+    }
+
+    {
+        std::lock_guard allocatedPortsLock(m_allocatedPortsLock);
+        m_allocatedPorts.clear();
+    }
+
+    // Not reset: session-scoped, subscriptions outlive the VM. Its stream handle dies with the IO relay
```
OneBlue: Is this comment outdated ?

OneBlue (commented):
Change looks good overall (just minor comments). 

Test coverage that would be good to add would be: 

- Create a container with a mapped port, but don't start it 
- Make the VM time out
- Launch the container and validate that the port mapping works as expected
- Stop the container
- Make the VM time out again
- Launch & validate the port mapping again

Something similar with volumes would also be a good idea

src/windows/wslcsession/WSLCSessionRuntime.cpp
```diff
@@ -665,6 +685,22 @@ bool WSLCSessionRuntime::TriggerIdleTerminationForTest()
...
+                return;
+            }
+
+            if (m_idleState->ActivityCount() != 0)
+            {
+                auto stopRequested = VmExitDisposition::StopRequested;
+                m_vmExitDisposition.compare_exchange_strong(stopRequested, VmExitDisposition::Active);
+
+                lock.reset();
+                NotifyVmStarted();
```
OneBlue: If I'm reading this correctly, if someone acquired a VM lease while we dropped the lack, we "cancel" the shutdown, but we already called the VM stop notification for plugins, so calling `NotifyVmStarted` could lead to some weird behavior since the plugins will actually run on the same VM as before.

To work around this, do we have a way to set a `VM stopping` state, that would prevent anyone from acquiring a new VM lease until the stop is complete ?

test/windows/WSLCTests.cpp
```diff
@@ -11723,114 +12166,159 @@ class WSLCTests
...
 
-    // A running --rm container that survives a forced idle teardown is dropped to Exited without the
-    // auto-remove delete (dockerd is gone by then). On the next VM restart it must be auto-removed
-    // rather than leaking as a permanent Exited container.
-    WSLC_TEST_METHOD(TriggerIdleTerminationRemovesExitedAutoRemoveContainer)
+    // A created or stopped container has no activity hold. Its port mapping and bind mount must remain
+    // usable after each idle teardown and lazy VM restart.
+    WSLC_TEST_METHOD(TriggerIdleTerminationRecoversStoppedContainerResources)
     {
         SKIP_TEST_SERVER();
```
OneBlue: nit: I don't think we want `SKIP_TEST_SERVER()` here. The WSLC tests are disabled on server at the moment either way

benhillis: Removed in 2ebee52a -- dropped all three that this change added.

src/windows/service/inc/wslc.idl
```diff
@@ -656,7 +682,14 @@ interface IWSLCSession : IUnknown
...
     // Create a process at the VM level. This is meant for debugging.
-    HRESULT CreateRootNamespaceProcess([in, ref] LPCSTR Executable, [in, ref] const WSLCProcessOptions* Options, [in] ULONG TtyRows, [in] ULONG TtyColumns, [out] IWSLCProcess** Process, [out] int* Errno);
+    // 'FromVmLifecycleCallback' is set by the service for a call issued from within a VM lifecycle
+    // plugin callback (see IWSLCPluginNotifier::OnVmStopping): such a call is served by a VM that is
+    // already stopping, instead of waiting for the teardown it is blocking. The session cannot verify
+    // this -- the callback runs in the service, in another process -- so the flag is advisory. Setting
+    // it outside that context is self-defeating rather than dangerous: it opts the caller into a VM
+    // that is about to be torn down, so its work is destroyed and only that caller's session is
+    // affected.
+    HRESULT CreateRootNamespaceProcess([in, ref] LPCSTR Executable, [in, ref] const WSLCProcessOptions* Options, [in] ULONG TtyRows, [in] ULONG TtyColumns, [in] BOOL FromVmLifecycleCallback, [out] IWSLCProcess** Process, [out] int* Errno);
```
OneBlue: nit: I'd rename to `AcquireVmLease` or similar to describe what this flag does instead of when it's set by the caller

src/windows/service/exe/PluginManager.cpp
```diff
@@ -127,7 +182,7 @@ try
     RETURN_HR_IF(E_POINTER, WindowsPath == nullptr || Mountpoint == nullptr);
 
     auto session = ResolveWslcSession(Session);
-    auto result = session->MountWindowsFolder(WindowsPath, Mountpoint, ReadOnly);
+    auto result = session->MountWindowsFolder(WindowsPath, Mountpoint, ReadOnly, FromVmLifecycleCallback(Session));
```
OneBlue: I think we should reject the caller if a plugin tries to mount a folder / execute a process while the VM is not running (regardless of whether the VM is terminating or not). 

The easiest way I could think of doing that would be to have flags in that call to determine if the caller should obtain a VM lease or not, and in the case of plugins, that flag should always be set to false

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -682,6 +690,31 @@ void WSLCContainerImpl::Initialize()
     m_comWrapper->Initialize(weak_from_this());
 }
 
+WSLCVirtualMachine& WSLCContainerImpl::Vm() const
```
OneBlue: nit: I think we can probably remove these methods, they don't feel like they add much value

src/windows/inc/WslPluginApi.h
```diff
@@ -156,19 +156,19 @@ typedef HRESULT (*WSLPluginAPI_OnWslcVmStarted)(const struct WSLCSessionInformat
...
-// keep the VM alive either; they block until the teardown completes and are then served by the next
-// VM, reported by a new OnWslcVmStarted. If the session itself is terminating there is no next VM and
-// those calls fail once the teardown completes.
-//
-// The callback must therefore not block waiting on another thread that is calling into this session:
-// that thread is waiting for the teardown, and the teardown is waiting for this callback to return.
-// Last-minute work in the VM must be done synchronously, on the callback's own thread.
+// for, for example -- dies with it. Calls made by other threads while this callback is running are
+// served by the same stopping VM, on the same terms; once the teardown starts they fail with
+// WSLC_E_VM_NOT_RUNNING rather than waiting for or creating another VM.
```
OneBlue, line 167: nit: If we're talking about this HRESULT here, we should also including it in this file, and a comment in its IDL declaration that says that it's part of the plugin API contract

src/windows/wslcsession/WSLCExecutionContext.h
```diff
@@ -27,15 +27,16 @@ class WSLCExecutionContext : public wsl::windows::common::COMServiceExecutionCon
 protected:
     bool CollectUserWarning(const std::wstring& warning) override
     {
-        if (m_warningCallback != nullptr)
+        IWarningCallback* callback = m_warningCallback;
```
OneBlue: What's the reason for this change ?

test/windows/testplugin/Plugin.cpp
```diff
@@ -697,6 +710,37 @@ try
         }
 
         g_logfile << "WSLC VM stopping leaked process: " << (SUCCEEDED(hr) ? "ok" : "failed") << std::endl;
+
+        // A call from a thread this plugin owns is served on the same terms as the callback itself:
+        // by the VM that is stopping, not by a future one and not by a new one. It must not block on
+        // the teardown, which is exactly the deadlock the old "wait for the next VM" behaviour risked.
```
OneBlue: nit: this comment refers to an old behavior that won't existing anymore once merged

OneBlue (dismissed):
LGTM. Thank you for doing this !

OneBlue (approved):

PR #41083, opened 2026-07-15
Log real elapsed time (and hangs) in SlowOperation telemetry

yeelam-gordon:
## Why

Today''s `SlowOperation` telemetry (from #40269) fires **once at 10 s and stops**, with no duration — so every slow phase looks identical. We can''t tell an 11 s blip from a 5-minute stall; a true hang (operation never returns, destructor never runs) is **invisible**; and the backend just accumulates a pile of "crossed 10 s" events that carry no insight and nothing to act on.

With the real elapsed time we can **bucket durations on a dashboard** (e.g. 10–30 s / 30–60 s / 1–5 min / hung) and actually see where startup time goes and which phase is regressing — instead of only knowing "something was slow." #40269''s own review flagged this as the follow-up: *"we know it was slow, but not how slow… logging elapsed at scope exit would be most useful."*

## What (only `SlowOperationWatcher`; no call sites touched)

At most one event, now carrying the real duration:

- **< 10 s** → nothing (unchanged fast path).
- **≥ 10 s** → one `timedOut=false` event with the true `elapsedMs`.
- **still running at `MaxDuration` (15 min)** → one `timedOut=true` event, then stop.

`timedOut` is a **finished-vs-hung** flag, deliberately **not** success-vs-failure: the watcher is a pure duration timer with no visibility into the operation''s result (a scope that exits by throwing is still `timedOut=false`), and WSL failures often don''t throw, so any inferred success/failure would be unreliable. `elapsedMs` + `timedOut` are **added** fields (backward compatible); new ctor params are defaulted (all 12 callers unchanged); only `WSL_LOG_TELEMETRY` is emitted; construction is best-effort and fully `noexcept`, and the timer is drained before teardown → no UAF, and a telemetry failure can never crash or slow the watched operation.

## Tests (new `SlowOperationWatcherUnitTests.cpp`, TAEF — 6/6 passing)

- `FastPathEmitsNothing` / `DefaultConstructionFastPathEmitsNothing` — under threshold stays silent.
- `SlowCompletionEmitsOneRecord` — one `timedOut=false` with real elapsed.
- `ResetAfterThresholdEmitsOnce` — `Reset()` + destructor emit exactly once.
- `ResetBeforeThresholdEmitsNothing` — early disarm stays silent, even if the scope runs long after `Reset()`.
- `HangEmitsOnceAtMaxThenStops` — one `timedOut=true` at the cap, no second event on later exit.

### DIFF
--- a/src/windows/common/SlowOperationWatcher.cpp
+++ b/src/windows/common/SlowOperationWatcher.cpp
@@ -8,12 +8,23 @@ Module Name:
 
 Abstract:
 
-    See header for contract. A single-shot threadpool timer is armed for SlowThreshold
-    in the constructor. If it fires, the callback emits one `SlowOperation` telemetry
-    event with the phase name and captured std::source_location. The timer is owned by
-    wil::unique_threadpool_timer, whose destroyer cancels pending callbacks and blocks
-    for any in-flight callback before closing, so OnTimerFired cannot dereference
-    `*this` after destruction.
+    See header for contract. A single-shot threadpool timer is armed for MaxDuration in
+    the constructor. Emission is at most one event, guarded by m_reported (atomic):
+
+      - If the operation finishes first (destructor or Reset()), Finish() cancels+drains
+        the timer and, if the elapsed time is at least SlowThreshold, emits one
+        timedOut=false event with the real elapsed time. Under-threshold => silent.
+      - If the operation is still running at MaxDuration, the timer fires once and emits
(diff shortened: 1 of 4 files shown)


yeelam-gordon:
Closing in favor of a fresh PR: the change was stabilized through a full Copilot review loop on my fork (converged, all threads resolved) and the telemetry field was renamed completed -> timedOut. This PR had accumulated stale/superseded review threads from earlier rounds; reopening clean for easier review.

Issue #40984, opened 2026-07-02
[WSL 2.9.3] Docker fails to bind ports with "address already in use" when networkingMode=mirrored is enabled


gsdukbh:
### Windows Version

Microsoft Windows [version 10.0.26200.8246]

### WSL Version

2.9.3.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

Linux version 6.18.35.2-microsoft-standard-WSL2 

### Distro Version

Ubuntu 22.04

### Other Software

* Docker Desktop or Docker Engine running inside WSL
* Docker Compose

### Repro Steps

1. Upgrade WSL to version **2.9.3** (pre-release).
2. Enable mirrored networking by adding the following to `%USERPROFILE%\.wslconfig`:
   ```ini
   [wsl2]
   networkingMode=mirrored
 ```
1 .Restart WSL completely (wsl --shutdown).
2. Attempt to start a Docker container that exposes a port to the host (e.g., 0.0.0.0:8003:8003). I was trying to run Immich via docker-compose up -d.

### Expected Behavior

The Docker container should start successfully, bind to the port, and the service should be accessible via both localhost:8003 and the Windows host's LAN IP address (e.g., http://192.168.1.39:8003/

### Actual Behavior

Docker fails to start the container and throws the following endpoint connectivity error:
> Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint immich_server (...): failed to bind host port 0.0.0.0:8003/tcp: address already in use

Troubleshooting Performed (To rule out user errors)
I have thoroughly checked both the Windows host and the WSL environment to ensure the port is truly available. The issue persists despite confirming the following:

1. No WinNAT Port Reservation: Checked netsh interface ipv4 show excludedportrange protocol=tcp in Windows. Port 8003 is NOT in the excluded range. Restarting WinNAT (net stop winnat / net start winnat) did not help.
2. No Windows Processes Using the Port: Ran netstat -ano | findstr :8003 on Windows. No output.
3. No WSL/Docker Zombie Processes: Ran docker ps -a and sudo ss -tulpn | grep 8003 inside WSL. No leftover containers or docker-proxy processes are holding the port.
4. No PortProxy Conflicts: Verified that there are no conflicting netsh interface portproxy rules active.
[...]

PR #41085, opened 2026-07-15
Cap guest usage of host ephemeral range to half of the ports in mirrored mode

FetoiuCatalin:
**Summary of the issue and fix**
https://github.com/microsoft/WSL/pull/40597/changes introduced an app-compat issue where scenarios such as Docker can happen to use ports from the host ephemeral range, which are now denied. To preserve app-compat and also prevent the container from exhausting the host ephemeral port range, add a cap for the number of ports the container can use, set to half of the size of the host ephemeral port range

**Testing:**
- New automated test that sets the host ephemeral range to a small size (255 ports), then attempts binding to all the ports in the range and confirming that only half (127) can bind

- Existing automated NetworkTests 

- [ ] **Closes:**  #40984

### DIFF
--- a/src/windows/service/exe/WslCoreGuestNetworkService.cpp
+++ b/src/windows/service/exe/WslCoreGuestNetworkService.cpp
@@ -64,6 +64,9 @@ void wsl::core::networking::GuestNetworkService::CreateGuestNetworkService(
         TraceLoggingValue(m_hostUdpEphemeralPortRange.first, "udpStartPort"),
         TraceLoggingValue(m_hostUdpEphemeralPortRange.second, "udpEndPort"));
 
+    m_hostTcpEphemeralPortCap = ComputeHostEphemeralPortCap(IPPROTO_TCP);
+    m_hostUdpEphemeralPortCap = ComputeHostEphemeralPortCap(IPPROTO_UDP);
+
     hns::GuestNetworkService request{};
     request.VirtualMachineId = VmId;
     request.MirrorHostNetworking = true;
@@ -126,6 +129,11 @@ std::pair<uint16_t, uint16_t> wsl::core::networking::GuestNetworkService::Alloca
     THROW_IF_FAILED(m_allocatePortRange.value()(m_service.get(), c_ephemeralPortRangeSize, &m_reservedPortRange, &port));
 
     WI_ASSERT(m_reservedPortRange.endingPort - m_reservedPortRange.startingPort == c_ephemeralPortRangeSize);
+
+    // Count the overlap of the guest's reserved ephemeral range with the host ephemeral range
+    // and seed the in-use counters accordingly.
+    m_hostTcpEphemeralPortsInUse = ComputeHostEphemeralOverlap(IPPROTO_TCP);
+    m_hostUdpEphemeralPortsInUse = ComputeHostEphemeralOverlap(IPPROTO_UDP);
 
     // setting the port to zero as we do not expect any bind requests to be sent to wslcore for ports in this range
     m_reservedPorts.emplace(std::make_pair(HCN_PORT_PROTOCOL_TCP, static_cast<uint16_t>(0)), HcnPortReservation{port, 1});
@@ -225,6 +233,32 @@ bool wsl::core::networking::GuestNetworkService::IsPortInGuestEphemeralRange(uin
     return PortNumber >= m_reservedPortRange.startingPort && PortNumber <= m_reservedPortRange.endingPort;
 }
 
+uint32_t wsl::core::networking::GuestNetworkService::ComputeHostEphemeralPortCap(int Protocol) const noexcept
+{
+    const auto& range = (Protocol == IPPROTO_UDP) ? m_hostUdpEphemeralPortRange : m_hostTcpEphemeralPortRange;
+    if (range.second < range.first)
+    {
+        return 0;
+    }
+
+    // Cap the guest at (at most) half of the host ephemeral range so it can't exhaust the host's ports.
+    return (static_cast<uint32_t>(range.second) - range.first + 1) / 2;
+}
+
+uint32_t wsl::core::networking::GuestNetworkService::ComputeHostEphemeralOverlap(int Protocol) const noexcept
+{
+    const auto& range = (Protocol == IPPROTO_UDP) ? m_hostUdpEphemeralPortRange : m_hostTcpEphemeralPortRange;
+    if (range.second < range.first)
+    {
+        return 0;
+    }
+
+    // Number of guest reserved ports that fall within the host ephemeral range.
+    const uint32_t overlapStart = std::max<uint32_t>(m_reservedPortRange.startingPort, range.first);
+    const uint32_t overlapEnd = std::min<uint32_t>(m_reservedPortRange.endingPort, range.second);
+    return (overlapStart <= overlapEnd) ? (overlapEnd - overlapStart + 1) : 0;
+}
+
 int wsl::core::networking::GuestNetworkService::OnPortAllocationRequest(const SOCKADDR_INET& Address, _In_ int Protocol, _In_ bool Allocate) noexcept
 try
 {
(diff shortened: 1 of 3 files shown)


src/windows/service/exe/WslCoreGuestNetworkService.cpp
```diff
@@ -225,6 +233,32 @@ bool wsl::core::networking::GuestNetworkService::IsPortInGuestEphemeralRange(uin
     return PortNumber >= m_reservedPortRange.startingPort && PortNumber <= m_reservedPortRange.endingPort;
 }
 
+uint32_t wsl::core::networking::GuestNetworkService::ComputeHostEphemeralPortCap(int Protocol) const noexcept
+{
+    const auto& range = (Protocol == IPPROTO_UDP) ? m_hostUdpEphemeralPortRange : m_hostTcpEphemeralPortRange;
+    if (range.second < range.first)
+    {
+        return 0;
```
keith-horton: can we ASSERT this an have tests on CHK builds?
this should never happen - it's a terrible code bug when it does

src/windows/service/exe/WslCoreGuestNetworkService.cpp
```diff
@@ -225,6 +233,32 @@ bool wsl::core::networking::GuestNetworkService::IsPortInGuestEphemeralRange(uin
...
+uint32_t wsl::core::networking::GuestNetworkService::ComputeHostEphemeralPortCap(int Protocol) const noexcept
+{
+    const auto& range = (Protocol == IPPROTO_UDP) ? m_hostUdpEphemeralPortRange : m_hostTcpEphemeralPortRange;
+    if (range.second < range.first)
+    {
+        return 0;
+    }
+
+    // Cap the guest at (at most) half of the host ephemeral range so it can't exhaust the host's ports.
+    return (static_cast<uint32_t>(range.second) - range.first + 1) / 2;
```
keith-horton: why does this need a static_cast<uint32_t>? what type is range.second?

src/windows/service/exe/WslCoreGuestNetworkService.cpp
```diff
@@ -347,6 +391,16 @@ try
...
             m_reservedPorts.erase(it);
+
+            if (isHostEphemeralPort)
+            {
+                auto& portsInUse = (Protocol == IPPROTO_UDP) ? m_hostUdpEphemeralPortsInUse : m_hostTcpEphemeralPortsInUse;
+                if (portsInUse > 0)
+                {
+                    portsInUse--;
+                }
+            }
```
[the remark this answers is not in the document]
keith-horton: please strongly consider this change

test/windows/NetworkTests.cpp
```diff
@@ -4283,66 +4283,6 @@ class MirroredTests
         VERIFY_IS_TRUE(canBindUdp);
```
keith-horton, line 4283: we should add tests for this behavior

src/windows/service/exe/WslCoreGuestNetworkService.cpp
```diff
@@ -347,6 +391,16 @@ try
         {
             result = m_releasePort.value()(it->second.Handle);
             m_reservedPorts.erase(it);
+
+            if (isHostEphemeralPort)
+            {
+                auto& portsInUse = (Protocol == IPPROTO_UDP) ? m_hostUdpEphemeralPortsInUse : m_hostTcpEphemeralPortsInUse;
+                if (portsInUse > 0)
```
keith-horton: when would this be == 0 if isHostEphemeralPort is true?
should we add an assert here too and go through testing?

test/windows/NetworkTests.cpp
keith-horton: definitely need to validate that we can hit our limit, then they fail, then once sockets are closed and it falls below our limit, we can allocate again. then repeat that pattern (hit limit -> fail, fall below limit->succeed, repeat) in the same test.

keith-horton (changes requested):

test/windows/NetworkTests.cpp
FetoiuCatalin: I was hesitant to do this as even for the smallest host range (which seems to be enforced to 255), we would need to bind 128 times and I was not sure if that's too heavy for an automated test

I can write it and see if it takes too long to run. I can set the host range temporarily to 255 and then revert it to the original host range

keith-horton: calling bind 128 times is not taxing on anything - it should be just fine

src/windows/service/exe/WslCoreGuestNetworkService.cpp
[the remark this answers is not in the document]
FetoiuCatalin: fixed

src/windows/service/exe/WslCoreGuestNetworkService.cpp
FetoiuCatalin: updated

src/windows/service/exe/WslCoreGuestNetworkService.cpp
```diff
@@ -290,6 +306,8 @@ try
         return 0;
     }
 
+    const bool isHostEphemeralPort = IsPortInHostEphemeralRange(PortNumber, Protocol);
```
keith-horton, line 306: this should be moved in the 'if (Allocate)' block - it's only needed if we take that branch

src/windows/service/exe/WslCoreGuestNetworkService.cpp
```diff
@@ -307,6 +325,24 @@ try
             return 0;
         }
 
+        // New reservation for a port in the host ephemeral range: enforce the cap.
+        if (isHostEphemeralPort)
+        {
+            const auto cap = (Protocol == IPPROTO_UDP) ? m_hostUdpEphemeralPortCap : m_hostTcpEphemeralPortCap;
+            auto& portsInUse = (Protocol == IPPROTO_UDP) ? m_hostUdpEphemeralPortsInUse : m_hostTcpEphemeralPortsInUse;
```
keith-horton: should be const auto, not auto&.
it's a uint16_t. and we don't want to have this function modify that value.

src/windows/service/exe/WslCoreGuestNetworkService.cpp
```diff
@@ -324,6 +360,11 @@ try
         if (SUCCEEDED(result))
         {
             m_reservedPorts.emplace(std::make_pair(HnsProtocol, PortNumber), HcnPortReservation{port, 1});
+
+            if (isHostEphemeralPort)
+            {
+                ((Protocol == IPPROTO_UDP) ? m_hostUdpEphemeralPortsInUse : m_hostTcpEphemeralPortsInUse)++;
```
keith-horton: can we just write out the if statement? compacting this with a tertiary block isn't really useful

keith-horton: or do what you do below, take a ref to one of the two, then increment the ref

keith-horton (changes requested):

src/windows/service/exe/WslCoreGuestNetworkService.cpp
FetoiuCatalin: it's also used on the else branch of if (Allocate), I queried it once outside both

src/windows/service/exe/WslCoreGuestNetworkService.cpp
```diff
@@ -127,6 +134,11 @@ std::pair<uint16_t, uint16_t> wsl::core::networking::GuestNetworkService::Alloca
 
     WI_ASSERT(m_reservedPortRange.endingPort - m_reservedPortRange.startingPort == c_ephemeralPortRangeSize);
 
+    // Count the overlap of the guest's reserved ephemeral range with the host ephemeral range
+    // and seed the in-use counters accordingly.
+    m_hostTcpEphemeralPortsInUse = ComputeHostEphemeralOverlap(IPPROTO_TCP);
```
keith-horton, line 136: this is adding another variable that we now need to keep up to date, where we already have m_hostUdpEphemeralPortRange and m_reservedPortRange.

can we not add this to maintain yet more state when we can just calculate it when we need it with the existing variables we are tracking?

keith-horton (changes requested):

src/windows/service/exe/WslCoreGuestNetworkService.cpp
FetoiuCatalin: I agree saving the cap in a variable is not really needed, will remove it

FetoiuCatalin: removed

keith-horton (approved):

benhillis (approved):

PR #41096, opened 2026-07-16
Improve WSL2 guest memory reclaim

benhillis:
## What this changes

This replaces the existing memory reduction loop with a shared idle detector and simpler reclaim policy.

- CPU idle detection uses aggregate busy time instead of user time only.
- Free pages are compacted when the VM enters idle and after every reclaim operation.
- Cache reclaim waits for two minutes of sustained idle, protecting normal pauses between commands.
- Reclaim and compaction CPU are excluded from the next idle sample so the thread does not reset its own grace period.

### Modes

| Mode | Behavior after two minutes idle |
|---|---|
| Disabled | No memory reduction thread |
| DropCache | Run `drop_caches` once for the idle period |
| Gradual | Reclaim cold cache toward a 128 MB floor using RAM-scaled 256 MB to 1 GB steps |

Gradual uses `memory.reclaim` with `swappiness=0`, which prevents proactive reclaim from swapping anonymous memory. If `memory.reclaim` is unavailable, it falls back to DropCache.

The default remains DropCache.

## Why

The current implementation waits approximately ten minutes before dropping cache and only measures user CPU time. This leaves `vmmemWSL` elevated long after real workloads finish and can treat kernel or I/O work as idle.

The new policy returns memory sooner while preserving cache during short user pauses.

## DropCache results

Measured with two clean 16-way Linux kernel builds, two read-heavy source archive runs, and three warm-cache reuse runs per configuration.

| Metric | Existing DropCache | New DropCache |
|---|---:|---:|
| Kernel build: 50% working-set reduction | 591s | 177s |
| Source archive: 50% working-set reduction | 542s | 142s |
| Warm reread after 60 seconds idle | 0.233s | 0.220s |
| Kernel working set after 15 minutes | 2.14 GiB | 2.12 GiB |
| Source archive working set after 15 minutes | 1.99 GiB | 1.98 GiB |

The new DropCache policy reaches the existing eventual footprint roughly seven minutes sooner without affecting the 60-second warm-cache case.

## Gradual results

A focused test using the same two-minute grace period showed:

- warm rereads after 60 seconds remained at 0.22-0.23 seconds
- 1.83 GiB of reclaimable cache remained intact through 120 seconds
- reclaim began at approximately 129 seconds
- cache fell to 0.14 GiB by 140 seconds
- `vmmemWSL` fell from 3.23 GiB to 1.63 GiB by 150 seconds

Gradual reaches a lower sustained footprint because it continues reclaiming regrown cache toward the floor. DropCache remains the default for compatibility and Gradual remains available for users who prioritize minimum memory usage.

## Validation

- `cmake --build . --config Debug -- -m`
- built and deployed the generated MSI
- verified `memory.reclaim` `swappiness=0` support on the WSL 6.18 kernel
- collected host `vmmemWSL`, guest memory, reclaim, compaction, workload timing, and warm-cache reuse data

### DIFF
--- a/src/linux/init/util.cpp
+++ b/src/linux/init/util.cpp
@@ -3505,71 +3505,102 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
 
 #define RECLAIM_PATH CGROUP_MOUNTPOINT "/memory.reclaim"
 
-static long long int GetUserCpuTime()
+static bool ReadCpuBusyIdle(unsigned long long& Busy, unsigned long long& Idle)
 
 /*++
 
 Routine Description:
 
-    This routine parses /proc/stat to query a summary of all user CPU time.
+    This routine parses the aggregate "cpu" line of /proc/stat and splits the cumulative jiffies into
+    busy and idle buckets. Idle time is idle + iowait; everything else (user, nice, system, irq,
+    softirq, steal) counts as busy, so kernel-bound work keeps the VM out of the idle state rather than
+    looking at user time alone.
 
 Arguments:
 
-    None.
+    Busy - Receives the cumulative busy jiffies across all cores.
+
+    Idle - Receives the cumulative idle jiffies (idle + iowait) across all cores.
 
 Return Value:
 
-    The current user CPU counter for all cores.
+    true on success, false on failure.
 
 --*/
 
 {
-    wil::unique_fd Fd{open("/proc/stat", O_RDONLY)};
-    if (!Fd)
+    wil::unique_fd fd{TEMP_FAILURE_RETRY(open("/proc/stat", O_RDONLY | O_CLOEXEC))};
+    if (!fd)
     {
-        LOG_ERROR("open failed {}", errno);
-        return -1;
+        LOG_ERROR("open(/proc/stat) failed {}", errno);
+        return false;
     }
 
-    char Buffer[32];
-    int Result = TEMP_FAILURE_RETRY(read(Fd.get(), Buffer, (sizeof(Buffer) - 1)));
-    if (Result <= 0)
+    char buffer[256];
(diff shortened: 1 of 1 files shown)


dkbennett (approved):
Looks good, recommend adding unit tests for CpuIdleTracker. It would need to be pulled out of the anonymous namespace but could be unit tested for confidence it is correctly tracking these states as expected.

src/linux/init/util.cpp
```diff
@@ -3509,98 +3505,284 @@ int ProcessCreateProcessMessage(wsl::shared::Transaction& Transaction, gsl::span
...
-    char* Sp1;
-    char* Info = strtok_r(Buffer, " \n", &Sp1);
-    if (Info == nullptr)
+    if (strncmp(buffer, "cpu ", 4) != 0)
     {
         LOG_ERROR("/proc/stat first line missing cpu label");
+        return false;
+    }
+
+    unsigned long long fields[8] = {};
```
OneBlue, line 3619: nit: This might be a good candidate for using std::regex (can be done in a followup though)

src/linux/init/util.cpp
```diff
@@ -3614,185 +3796,180 @@ Return Value:
...
+                            bytes = reclaimStepBytes;
                         }
+
+                        reclaimed = RequestReclaim(bytes);
                     }
-                    else
+                }
+                else if (!droppedThisIdlePeriod)
+                {
+                    if (WriteToFile("/proc/sys/vm/drop_caches", "1\n") == 0)
```
OneBlue, line 3934: Should we switch this to 3 while we're at it to drop metadata as well ?

OneBlue (approved):
LGTM, minor suggestions

benhillis:
Good calls, will address as a follow-up.

PR #41106, opened 2026-07-17
Add --secret flag to image build command

ggarzia-MSFT:
This pull request adds support for Docker-style `--secret` arguments to the image build command, enabling secure injection of secrets (from environment variables or files) into the build process. It introduces robust parsing and validation for secret specifications, updates the internal data structures and APIs to handle secrets, and ensures safe handling and cleanup of resources during the build. The most important changes are grouped below.

### Feature: Add Docker-style `--secret` support to image builds

* Added a new argument type (`Secret`) to the CLI and command definitions, along with user-facing descriptions and error messages for invalid secret specs. (`src/windows/wslc/arguments/ArgumentDefinitions.h` [[1]](diffhunk://#diff-7b1922e0cc232767dccf72d426b028a26ab41b02a7839ce9f3c35e252b4393e1R108) `localization/strings/en-US/Resources.resw` [[2]](diffhunk://#diff-c64d306b3b9e1059cb6e28f3b1a3552de31a1a94490601541c4e43547f4ee8efR2330-R2333) [[3]](diffhunk://#diff-c64d306b3b9e1059cb6e28f3b1a3552de31a1a94490601541c4e43547f4ee8efR2875-R2878)
* Implemented parsing and validation logic for `--secret` specs, supporting secrets from environment variables or files, and matching Docker's semantics and error handling. (`src/windows/wslc/tasks/ImageTasks.cpp` [[1]](diffhunk://#diff-b955712d2e30de6e791156f73a09b2e0dbfc22c14e010e120ffd8865c56d9d16R75-R227) [[2]](diffhunk://#diff-b955712d2e30de6e791156f73a09b2e0dbfc22c14e010e120ffd8865c56d9d16R262-R270)

### Internal API and Data Structure Updates

* Introduced new `BuildSecret` and `WSLCBuildSecret` structs, and updated the image build options and service interfaces to pass secrets through all layers. (`src/windows/service/inc/wslc.idl` [[1]](diffhunk://#diff-f354d8f5022677f418046fbe5c50ab602814c13ed9d0041e3d238954a153b757R172-R187) [[2]](diffhunk://#diff-f354d8f5022677f418046fbe5c50ab602814c13ed9d0041e3d238954a153b757R543) `src/windows/wslc/services/ImageService.h` [[3]](diffhunk://#diff-623ff8d4e3354ac821c3f2a27278d2ffede80cdf05dcf8db2eaa3e9ae6403934R21-R27) [[4]](diffhunk://#diff-623ff8d4e3354ac821c3f2a27278d2ffede80cdf05dcf8db2eaa3e9ae6403934R37) `src/windows/wslc/services/ImageService.cpp` [[5]](diffhunk://#diff-02bd98b2ee17399f9362d63153b45508809c4508078bcb2207b466d3171babe1R123) [[6]](diffhunk://#diff-02bd98b2ee17399f9362d63153b45508809c4508078bcb2207b466d3171babe1R176-R192) [[7]](diffhunk://#diff-02bd98b2ee17399f9362d63153b45508809c4508078bcb2207b466d3171babe1R204)
* Updated the image build command and tasks to collect, parse, and forward secrets to the service layer. (`src/windows/wslc/commands/ImageBuildCommand.cpp` [[1]](diffhunk://#diff-d83b8938c1848772336f481cc4f552d3d29ac1c4b31a4240359797d561910bb3R37) `src/windows/wslc/tasks/ImageTasks.cpp` [[2]](diffhunk://#diff-b955712d2e30de6e791156f73a09b2e0dbfc22c14e010e120ffd8865c56d9d16L129-R292)

### Resource Management and Robustness

* Refactored VM folder mounting logic to safely track and unmount all mounted paths on completion or error, preventing resource leaks. (`src/windows/wslcsession/WSLCSession.cpp` [src/windows/wslcsession/WSLCSession.cppL910-R934](diffhunk://#diff-174eed357efbb4e742cc7d6ccc9397607e1a46efe916e3a225103a8b97ffd9daL910-R934))
* Added validation to ensure secret arrays are not null when their count is nonzero, preventing invalid argument errors. (`src/windows/wslcsession/WSLCSession.cpp` [src/windows/wslcsession/WSLCSession.cppR883](diffhunk://#diff-174eed357efbb4e742cc7d6ccc9397607e1a46efe916e3a225103a8b97ffd9daR883))
## PR Checklist


src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -71,6 +72,110 @@ namespace {
 
 } // namespace
 
+static services::BuildSecret ParseSecretSpec(const std::wstring& spec)
```
benhillis: The parse and validate operations are very similar, can we avoid having two versions of this?

ggarzia-MSFT: I had split it into two functions so that all of the input validation occurs in the ArgumentValidation.cpp file. I can undo that if it is better to have the parsing and validation in the same place.

dkbennett: IMO they should be the same function - if you can parse the spec successfully without throwing then that is the validation, you just discard the result. Then call the same function in the task to get the data. This is what we do for Signal and other arguments. There should be one source of truth for correct spec parsing.

ggarzia-MSFT: Where should that function live?

ggarzia-MSFT: Ok, should be fixed now

dkbennett: If it's beefy it could live in the service code and just get called and retrieved by validation, but my preference is that argument validation / parsing is mostly in the argument validation code area. You can also add a lot of test cases in the command line to test various combinations that are valid/invalid easily, or make a specific test for it that hammers the validation code.

ggarzia-MSFT: I agree that it makes the most sense to leave in the argument validation area. I will add a lot of Unit tests to test different combinations

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -201,6 +213,183 @@ void ValidateWSLCSignalFromString(const std::vector<std::wstring>& values, const
...
+            .Id = std::move(id),
+            .Value = std::move(value),
+        };
+    }
+
+    // Docker parity: a referenced environment variable that is unset (or set but empty) yields an
+    // empty secret value rather than an error. GetEnvironmentVariableW returns 0 for an undefined
+    // variable; for a defined one it returns the buffer size needed including the null terminator.
+    std::wstring value;
+    DWORD size = GetEnvironmentVariableW(envName.c_str(), nullptr, 0);
```
dkbennett: We have Environment variable related code in core EnvironmentOptions.cpp, see "ReadEnv" that will use the WIL version and returns a std::optional<std::wstring>.

This is at present an anonymous function in the file, but we could either add a passthrough for it or update it to be accessible from elsewhere in the CLI.

It might be more appropriate if this is a wsl::windows::common method (feel free to refactor if you decide this is a good idea)

That will clean up this part of the code and use a consistent code path.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -201,6 +213,183 @@ void ValidateWSLCSignalFromString(const std::vector<std::wstring>& values, const
     }
 }
 
+services::BuildSecret ParseSecretSpec(const std::wstring& spec)
```
dkbennett: this is 150+ lines of code that could probably be a separate file here.  Maybe we could put similar "spec parsing" code together in the same file since they are similar and probably have some significant reusable code patterns. That might be good for a follow-up though. For now this is fine, but I'm seeing a good refactoring opportunity here.

test/windows/wslc/e2e/WSLCE2EImageBuildTests.cpp
```diff
@@ -261,6 +277,432 @@ class WSLCE2EImageBuildTests
...
+        VERIFY_IS_TRUE(buildResult.Stderr->find(L"source file not found or not a regular file") != std::wstring::npos);
+    }
+
+    WSLC_TEST_METHOD(WSLCE2E_Image_Build_Secret_EnvAndSrc_EnvWins_Success)
+    {
+        // Docker parity: when both 'env=' and 'src=' are given, the environment variable wins and
+        // the file path is ignored (no error).
+        constexpr auto envName = L"WSLC_E2E_ENV_WINS_VALUE";
+        constexpr auto envValue = L"env-wins-content-55555";
+        THROW_IF_WIN32_BOOL_FALSE(SetEnvironmentVariableW(envName, envValue));
```
dkbennett: Check out ScopedEnvVariable for using environment variables in tests.  Same feedback for other environment variable uses in tests.

dkbennett (commented):
Minor nit, looks good, love all the test coverage!

dkbennett (dismissed):

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -950,6 +969,89 @@ try
         buildArgs.push_back(Options->Labels.Values[i]);
     }
 
+    // Materialize each secret into a root-only tmpfs file inside the VM and hand docker a file
```
OneBlue: Depending on the secrets themselves, they could pretty big and potentially exhaust the memory on the `/run` mount. 

One thing we could do instead would be to create a temporary folder, and then mount each file via virtiofs inside the guest. This will be a bit more work, but we'll be safe from any memory exhaustion

OneBlue: We could even mount all of this in a new tmpfs, and unmount it once done, that way nothing can leak behind

ggarzia-MSFT: I did some testing on this, and secrets are limited to 500KB by the docker engine. Is this still an issue with that in mind? Or are we looking for a way to circumvent this limit for wslc?

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -950,6 +969,89 @@ try
...
+    // Each build gets its own random subdirectory so concurrent builds (which only hold a shared lock)
+    // never read or delete one another's secrets; the scope_exit removes just this build's directory.
+    std::string secretDir;
+    if (Options->Secrets.Count > 0)
+    {
+        GUID dirId{};
+        THROW_IF_FAILED(CoCreateGuid(&dirId));
+        secretDir = std::format("/run/wslc-secrets/{}", wsl::shared::string::GuidToString<char>(dirId));
+    }
+    auto removeSecrets = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
```
dkbennett: Just a note that I was curious about the failure cases where we might leave a secret on the VM which another process could hypothetically read. Since the secret is placed in fmpfs, which is RAM-backed and removed on VM teardown, it will get cleaned up regardless. In addition, if this process were to crash the VM would get indirectly torn down when the compute system is terminated as a result of the crash, which would free the tmpfs data and remove the secrets. During teardown it doesn't look like it would be accessible so this cleanup code appears to be sufficient.

In normal execution, this scope exit will remove the file. In fail-fast situation the scope exit would be skipped but the VM would ultimately get torn down and the fmpfs data deleted. Either way the secret is removed and never hits the disk, so this appears to be sufficiently resilient cleanup.

src/windows/wslc/arguments/SpecParsing.cpp
```diff
@@ -0,0 +1,639 @@
...
+        THROW_HR_WITH_USER_ERROR_IF(
+            E_INVALIDARG,
+            Localization::MessageWslcSecretInvalidSpec(spec, std::format(L"unable to determine size of source file: {}", absPath.wstring())),
+            size < 0);
+
+        std::vector<BYTE> value(static_cast<size_t>(size));
+        if (size > 0)
+        {
+            file.seekg(0);
+            file.read(reinterpret_cast<char*>(value.data()), size);
```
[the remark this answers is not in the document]
dkbennett: This is reasonable to add just to protect against really unexpected situations like someone pointing to a VHD file or something, however, I'd recommend making the upper bound check an order of magnitude higher than what BuildKit limits. That protects us against the 'enormous memory' situation but still allows us to not require any changes if BuildKit allows a higher cap in the future. If something changes where BuildKit suddenly allows 1 MB secrets we aren't broken, unless they really crank up the cap.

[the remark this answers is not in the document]
ggarzia-MSFT: I agree that there should be a check but making it larger than docker's limit seems a bit counterintuitive. I understand wanting to future proof the check, but I think that a client side check should match docker engine's check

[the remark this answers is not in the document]
ggarzia-MSFT: Mayber there is a way to figure out the limit rather than hardcoding it?

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -951,6 +1093,78 @@ try
...
+        // created once and reused by every build in this session.
+        const auto& secretRoot = EnsureSecretRoot();
+
+        GUID dirId{};
+        THROW_IF_FAILED(CoCreateGuid(&dirId));
+        secretHostDir = secretRoot / wsl::shared::string::GuidToString<wchar_t>(dirId, wsl::shared::string::GuidToStringFlags::None);
+        std::filesystem::create_directory(secretHostDir);
+
+        std::vector<std::pair<std::string, std::string>> secretRefs; // (id, host filename) pairs
+        for (ULONG i = 0; i < Options->Secrets.Count; i++)
```
OneBlue, line 1003: Instead of manually copying the secrets to a given folder, what I would recommend would be to individually mount each secret into the guest, so we never need to copy them. 

Onside the guest, we can create a folder, and then mount the parent folder of each secret and give the fully assembled path to docker. 

Copying the secret on a different Windows path would carry the secret of moving a potentially encrypted file to a non-encrypted location (if the user account is using EFS for instance)

OneBlue: Sorry if my previous suggestion was unclear

ggarzia-MSFT: No worries, thanks for catching that. Should be fixed now

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -951,6 +1093,117 @@ try
...
+
+            if (secret.SourcePath != nullptr && secret.SourcePath[0] != L'\0')
+            {
+                // File secret: mount the file's parent directory read-only and reference the file in
+                // place - the bytes are never copied. Mounting the whole directory (not just the file) is
+                // inherent to virtiofs sharing a directory tree; sibling files are exposed to this user's
+                // own build VM read-only for the build's duration only.
+                std::filesystem::path sourcePath(secret.SourcePath);
+                auto parent = sourcePath.parent_path();
+                auto fileNameUtf8 = wsl::shared::string::WideToMultiByte(sourcePath.filename().wstring());
```
OneBlue: nit: This can be simplified to: 

```
auto fileNameUtf8 = sourcePath.filename().string();
```

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -951,6 +1093,117 @@ try
...
+                auto it = fileSecretDirMounts.find(parent);
+                if (it == fileSecretDirMounts.end())
+                {
+                    it = fileSecretDirMounts.emplace(parent, mountInVm(parent.c_str(), TRUE)).first;
+                }
+                secretArgs.emplace_back(secret.Id, std::format("{}/{}", it->second, fileNameUtf8));
+            }
+            else
+            {
+                // Env/in-memory secret: materialize the bytes into a host file mounted read-only.
```
OneBlue: Instead of writing files on the host, could we use env variables and use an env based secret via the command line we invoke `docker build` with ? That would allow us to completely remove the logic to write to secret files, and will also allow us to skip all the logic to cleanup the files later

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -908,12 +1006,56 @@ try
...
+        {
+            // Best-effort but not silent: a failed unmount can leave a share mounted and makes the
+            // secret-directory cleanup (removeSecrets) more likely to fail, so log it. Never throw here.
+            LOG_IF_FAILED(m_virtualMachine->UnmountWindowsFolder(path.c_str()));
+        }
+    });
+    auto mountInVm = [&](LPCWSTR windowsPath, BOOL readOnly) -> std::string {
+        GUID id{};
+        THROW_IF_FAILED(CoCreateGuid(&id));
+        auto vmPath = std::format("/mnt/{}", wsl::shared::string::GuidToString<char>(id));
```
OneBlue: I recommend mounting the secret dir under `/run`. That way it's not actually written to disk, and if let's say the session crashes, we won't need to clean anything up on the next boot

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -951,6 +1093,117 @@ try
...
+            // ',' or '=' a malicious caller could use to inject extra options.
+            RETURN_HR_IF(E_INVALIDARG, std::string_view(secret.Id).find_first_of(",=") != std::string_view::npos);
+
+            if (secret.SourcePath != nullptr && secret.SourcePath[0] != L'\0')
+            {
+                // File secret: mount the file's parent directory read-only and reference the file in
+                // place - the bytes are never copied. Mounting the whole directory (not just the file) is
+                // inherent to virtiofs sharing a directory tree; sibling files are exposed to this user's
+                // own build VM read-only for the build's duration only.
+                std::filesystem::path sourcePath(secret.SourcePath);
```
OneBlue, line 1023: We should fail if `secret.SourcePath` is not an absolute path (since the client & server aren't guaranteed to have the same current directory)

test/windows/wslc/e2e/WSLCE2EImageBuildTests.cpp
```diff
@@ -261,6 +276,586 @@ class WSLCE2EImageBuildTests
...
+
+        auto dockerfilePath = testRoot / L"Dockerfile";
+        WriteTestFileContent(dockerfilePath, "FROM debian:latest\n");
+
+        auto missingFile = testRoot / L"does-not-exist.txt";
+        auto buildResult = RunWslc(std::format(
+            L"build \"{}\" -f \"{}\" --secret id=x,src=\"{}\"", contextDir.wstring(), dockerfilePath.wstring(), missingFile.wstring()));
+        VERIFY_ARE_EQUAL(1u, buildResult.ExitCode.value_or(0u));
+        VERIFY_IS_TRUE(buildResult.Stderr.has_value());
+        VERIFY_IS_TRUE(buildResult.Stderr->find(L"source file not found or not a regular file") != std::wstring::npos);
```
OneBlue: I recommend validating the exact error message to make sure that we lock down the exact error behavior we want (applies to other tests as well)

src/windows/wslc/arguments/SpecParsing.cpp
```diff
@@ -0,0 +1,631 @@
...
+        auto absPath = std::filesystem::weakly_canonical(std::filesystem::absolute(srcPath), ec);
+        if (ec.value() != 0 || !std::filesystem::is_regular_file(absPath, ec))
+        {
+            throw ArgumentException(Localization::MessageWslcSecretInvalidSpec(
+                spec, std::format(L"source file not found or not a regular file: {}", absPath.wstring())));
+        }
+
+        // Metadata-only size guard (no bytes are read): keep parity with BuildKit's per-secret cap so an
+        // oversize file is rejected early with a friendly error instead of failing later in the daemon.
+        const auto fileSize = std::filesystem::file_size(absPath, ec);
```
OneBlue: Now that we mount the files in the guest, I would recommend removing this check from the client side (mostly because there's a race between the time we check the file size and the time buildkit does)

src/windows/wslc/arguments/SpecParsing.cpp
```diff
@@ -0,0 +1,631 @@
...
+
+    if (!srcPath.empty())
+    {
+        std::error_code ec;
+        // Resolve symlinks (and normalize '..') so we mount the directory that actually holds the
+        // secret's bytes rather than a link node's directory.
+        auto absPath = std::filesystem::weakly_canonical(std::filesystem::absolute(srcPath), ec);
+        if (ec.value() != 0 || !std::filesystem::is_regular_file(absPath, ec))
+        {
+            throw ArgumentException(Localization::MessageWslcSecretInvalidSpec(
```
OneBlue: I recommend letting this fail in the service if the file can't be mounted (we also probably want to support symlinks by resolving their canonical paths in the service for situations like these)

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -951,13 +975,107 @@ try
...
+        {
+            const auto& secret = Options->Secrets.Values[i];
+            RETURN_HR_IF_NULL(E_INVALIDARG, secret.Id);
+            RETURN_HR_IF(E_INVALIDARG, secret.Id[0] == '\0');
+            RETURN_HR_IF(E_INVALIDARG, secret.Id[0] == '-');
+            // Id is interpolated into docker's comma/'='-delimited --secret spec below, so reject any
+            // ',' or '=' a malicious caller could use to inject extra options.
+            RETURN_HR_IF(E_INVALIDARG, std::string_view(secret.Id).find_first_of(",=") != std::string_view::npos);
+
+            if (secret.SourcePath != nullptr && secret.SourcePath[0] != L'\0')
```
OneBlue: nit: I think it would be OK to fail if `SourcePath` was empty here

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -951,13 +975,107 @@ try
...
+
+        for (ULONG i = 0; i < Options->Secrets.Count; i++)
+        {
+            const auto& secret = Options->Secrets.Values[i];
+            RETURN_HR_IF_NULL(E_INVALIDARG, secret.Id);
+            RETURN_HR_IF(E_INVALIDARG, secret.Id[0] == '\0');
+            RETURN_HR_IF(E_INVALIDARG, secret.Id[0] == '-');
+            // Id is interpolated into docker's comma/'='-delimited --secret spec below, so reject any
+            // ',' or '=' a malicious caller could use to inject extra options.
+            RETURN_HR_IF(E_INVALIDARG, std::string_view(secret.Id).find_first_of(",=") != std::string_view::npos);
```
OneBlue: nit: we should log the secret id in case we ever hit this to make debugging easier

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -951,13 +975,107 @@ try
...
+                    value = std::string_view(reinterpret_cast<const char*>(secret.Value), secret.ValueSize);
+                }
+                // An environment variable value cannot contain a NUL; reject rather than silently
+                // truncate the secret.
+                RETURN_HR_IF(E_INVALIDARG, value.find('\0') != std::string_view::npos);
+
+                // Random variable name (not derived from the id) so nothing about the id influences it
+                // and it cannot collide with a real environment variable. GUID dashes become '_' to keep
+                // it a conventional identifier.
+                GUID varId{};
```
OneBlue: nit: We could simplify things a bit by using `i` here and use something like `WSL_SECRET_{i}`

src/windows/wslc/arguments/SpecParsing.cpp
OneBlue: We can deal with symlinks in a future change though, for now I'd say that just letting the mount go through and letting buildkit reject it if the file isn't readable is OK

OneBlue (commented):
LGTM, minor comments

src/windows/wslc/arguments/SpecParsing.cpp
ggarzia-MSFT: I agree, I think this was leftover from when we read each file into memory

src/windows/wslc/arguments/SpecParsing.cpp
```diff
@@ -161,27 +155,15 @@ services::BuildSecret ParseSecretSpec(const std::wstring& spec)
...
     {
         std::error_code ec;
-        // Resolve symlinks (and normalize '..') so we mount the directory that actually holds the
-        // secret's bytes rather than a link node's directory.
+        // Normalize to an absolute path (the service requires one to mount the file's directory) but do
+        // not verify the file exists or is a regular file here: that would be a TOCTOU race with the
+        // build, and the file may only be reachable from the service's context. Let the service/BuildKit
+        // reject an unmountable or unreadable file instead. weakly_canonical also collapses '..' and
+        // resolves symlinks for the portion of the path that exists; it succeeds for a missing file.
         auto absPath = std::filesystem::weakly_canonical(std::filesystem::absolute(srcPath), ec);
```
OneBlue: I think we should fail if `weakly_canonical` fails. We also don't don't need to explicitely call `absolute` here since `weakly_canonical` does that for us

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1048,13 +1053,7 @@ try
...
 
-                // Random variable name (not derived from the id) so nothing about the id influences it
-                // and it cannot collide with a real environment variable. GUID dashes become '_' to keep
-                // it a conventional identifier.
-                GUID varId{};
-                THROW_IF_FAILED(CoCreateGuid(&varId));
-                auto varName = std::format(
-                    "WSLC_SECRET_{}", wsl::shared::string::GuidToString<char>(varId, wsl::shared::string::GuidToStringFlags::None));
+                auto varName = std::format("WSLC_SECRET_{}", std::to_string(i));
                 std::ranges::replace(varName, '-', '_');
```
OneBlue: We can remove now that we don't have a GUID anymore

test/windows/wslc/e2e/WSLCE2EImageBuildTests.cpp
```diff
@@ -496,12 +496,14 @@ class WSLCE2EImageBuildTests
         auto dockerfilePath = testRoot / L"Dockerfile";
         WriteTestFileContent(dockerfilePath, "FROM debian:latest\n");
 
+        // The client no longer validates that the source file exists; the path is forwarded and the
```
OneBlue: nit: This comment makes sense in the PR, but once merged it will be a bit confusing. I recommend rephrasing to something like `Expect the build to fail if the secret file is missing`

Also what's the error output like in situations like these ?

test/windows/wslc/e2e/WSLCE2EImageBuildTests.cpp
```diff
@@ -756,36 +758,10 @@ class WSLCE2EImageBuildTests
 
     WSLC_TEST_METHOD(WSLCE2E_Image_Build_Secret_MaxSizeFile_Success)
     {
-        // Exactly BuildKit's per-secret cap (500 KiB == 512000 bytes) must still succeed.
+        // A 500 KiB secret still mounts and builds; the client no longer caps secret file size.
```
OneBlue: Same for this comment, I think the previous version was a bit better

test/windows/wslc/e2e/WSLCE2EImageBuildTests.cpp
```diff
@@ -756,36 +758,10 @@ class WSLCE2EImageBuildTests
...
         RunSizedFileSecretSuccess(BuiltImageSecretMaxSize, L"wslc-e2e-build-secret-max-size", c_maxSecretSize);
     }
 
-    WSLC_TEST_METHOD(WSLCE2E_Image_Build_Secret_OversizeFile_Fails)
-    {
-        // One byte over BuildKit's cap (500 KiB + 1) must be rejected client-side from the file's
```
OneBlue: Shouldn't this test still apply ? I'd expect the build to still fail with a buildkit error here

test/windows/wslc/e2e/WSLCE2EImageBuildTests.cpp
ggarzia-MSFT: The error output is: Invalid --secret value 'id=x,src=dne.txt': source file not found or not a regular file: C:\Users\gavingarzia\repos\WSL3\dne.txt

OneBlue: Ok that's pretty good, in that case let's add a check to validate that this error is in the output ?

dkbennett (dismissed):

dkbennett (approved):

OneBlue (approved):

PR #41109, opened 2026-07-17
Added keep alive setting

craigloewen-msft:

Added a setting to WSL settings to keep the WSL VM alive at all times. 

## PR Checklist

Very simple change.
Adds a 'InstanceIdleTimeout' setting and a "WSL keep VM alive" setting to WSL settings.

Did this in WSL settings only, no major code changes. 

## Validation Steps Performed

Validated settings loaded correctly.


src/windows/wslsettings/Views/Settings/OptionalFeaturesPage.xaml
```diff
@@ -55,6 +58,20 @@
                         </ctControls:SettingsCard>
                     </ctControls:SettingsExpander.Items>
                 </ctControls:SettingsExpander>
+                <ctControls:SettingsExpander x:Name="InstanceIdleTimeoutExpander" x:Uid="Settings_InstanceIdleTimeout" IsEnabled="{x:Bind ViewModel.InstanceIdleTimeoutEnabled, Mode=OneWay}">
+                    <TextBlock Style="{StaticResource TextBlockSettingStyle}" Text="{x:Bind ViewModel.InstanceIdleTimeout, Mode=OneWay, Converter={StaticResource MillisecondsStringConverter}}"/>
```
benhillis, line 62: When the toggle is on, `InstanceIdleTimeout` is `-1`, and `MillisecondsStringConverter` only special-cases `ulong 0`, so this collapsed summary ends up reading literally `-1 ms`. Same thing shows up in the apply-changes dialog, since `SettingsApplyHelper.FormatValue` runs the new entry through `Settings_MillisecondsStringFormat` - so toggling `Keep WSL VM alive` produces a confirmation line like `Distribution idle timeout: -1 ms` for a field the user never touched.

Could we have the converter return null (or something like `Never`) for negative values, and map the toggle to `Settings_KeepVMAlive/Header` in the apply summary?

src/windows/wslsettings/ViewModels/Settings/OptionalFeaturesViewModel.cs
```diff
@@ -72,6 +77,32 @@ public bool IsOnSparseVHD
...
 
+    // Keeping WSL alive is driven by the distribution idle timeout: a negative value tells the
+    // service to never idle-terminate a distribution, and because the VM is only considered idle
+    // once every distribution has stopped, this keeps the WSL VM running as well.
+    public bool IsOnKeepVMAlive
+    {
+        get { return _instanceIdleTimeout!.Int32Value < 0; }
+        set
+        {
+            Set(ref _instanceIdleTimeout!, value ? -1 : _defaultInstanceIdleTimeout, nameof(IsOnKeepVMAlive));
```
benhillis: Toggling off always snaps back to the built-in default, so someone who had `instanceIdleTimeout = 5000` and flips this on and then off again silently loses their 5000. Worth stashing the previous positive value in a field and restoring that, falling back to `_defaultInstanceIdleTimeout` only when there wasn't one.

localization/strings/en-US/Resources.resw
```diff
@@ -1985,6 +1985,18 @@ You can also access more VS Code Remote options through the command palette with
   <data name="Settings_VMIdleTimeout.Description" xml:space="preserve">
     <value>The number of milliseconds that a VM is idle, before it is shut down.</value>
   </data>
+  <data name="Settings_KeepVMAlive.Header" xml:space="preserve">
+    <value>Keep WSL VM alive</value>
+  </data>
+  <data name="Settings_KeepVMAlive.Description" xml:space="preserve">
+    <value>Keep the WSL virtual machine, and the distributions running inside it, from being automatically shut down after they have been idle.</value>
```
benhillis: Two small wording things here. This is backed by `general.instanceIdleTimeout`, which also applies to WSL1 instances (see the `LxssInstance` path in `LxssUserSession.cpp`), so `Keep WSL VM alive` undersells what it actually does. It also doesn't start anything - it only stops an already-running VM from being shut down, and `wsl --shutdown` still works. Might be worth reflecting both in the description so nobody expects the VM to come up on boot.

benhillis (commented):
Nice, the mechanism checks out - `_VmIsIdle()` only returns true once no running instance has a valid client id, so pinning `instanceIdleTimeout` to a negative value does keep the VM up. A few things I hit while reading through it.

src/windows/wslsettings/Views/Settings/OptionalFeaturesPage.xaml
craigloewen-msft: I actually prefer it showing `-1` as it is clearly communicating what it's setting. I think changing it to show 'Never' while making it more readable, makes it less clear on what actual settings are changing. 
So I'd prefer to stick with displaying it as `-1` as that is what gets truly set in the file.

OneBlue (approved):

PR #41118, opened 2026-07-20
fixed title error visibility

### LINKED ISSUE #41112: Freshly installed on a brand new Windows 11, just opens and closes, doesn't even let me see an error message
### Windows Version

11

### WSL Version

don't know because I can't run it

- [ ] WSL 2
- [ ] WSL 1

### Kernel Version

_No response_

### Distro Version

_No response_

### Other Software

_No response_

### Repro Steps

I just installed WSL by running from a PowerShell: 
```
wsl --install
```

To reproduce:
1. Click the Start button
2. Type "WSL" into the search box
3. run "WSL", the only result that shows up

### Expected Behavior

should work, and if there is a problem, should show an error message

### Actual Behavior

A window that seems to be some sort of terminal shows up for a fraction of a second and immediately closes.
I guess some error message is shown inside of that terminal but it's impossible to read it.

That's so f****** stupid. At least make sure that if there's an error a proper error message is shown somewhere that I can read it.

### Diagnostic Logs

_No response_

shuaiyuanxx:
The Start menu launches the MSIX `wsl.exe`, which forwards the command to the installed MSI `wsl.exe`. The inner process prints the actual error, but only the outer MSIX process retains the App activation context and owns the console window. Previously, the outer process returned immediately when the inner process failed.

The outer process now detects an App-activated launch and prompts before exiting when the inner process returns WSL's client failure code (`-1`). Linux command exit codes are not affected.

## PR Checklist
- [x] **Closes:** Closes #41112

The MSIX `wsl.exe` owns the Start menu console, while the forwarded MSI `wsl.exe` prints the actual startup error. This change keeps the outer process open when the inner process returns WSL's client failure code (`-1`), without affecting Linux exit codes or  terminal launches.

## Validation Steps Performed
 - Clean-built `wsl.exe` and the signed glue MSIX.
 - Triggered a deterministic startup failure through the real Start menu entry.
 - Verified the official build closed immediately, while the patched build kept the error window open.
 - Verified normal terminal invocations and exit codes were unchanged.

### DIFF
--- a/src/windows/common/WslClient.cpp
+++ b/src/windows/common/WslClient.cpp
@@ -1525,7 +1525,14 @@ int WslMain(_In_ std::wstring_view commandLine)
     // Call the MSI package if we're in an MSIX context
     if (wsl::windows::common::wslutil::IsRunningInMsix())
     {
-        return wsl::windows::common::install::CallMsiPackage();
+        const bool launchedViaAppActivation = winrt::Windows::ApplicationModel::AppInstance::GetActivatedEventArgs() != nullptr;
+        const auto exitCode = wsl::windows::common::install::CallMsiPackage();
+        if (launchedViaAppActivation && exitCode == -1)
+        {
+            g_promptBeforeExit = true;
+        }
+
+        return exitCode;
     }
 
     // Use exit code -1 so invokers of wsl.exe can distinguish between a Linux


OneBlue (approved):
I think this is a good change, although if we happen to run a command line like `wsl --install --prompt-before-exit` in that context, we would follow that command line to the inner process, and therefore end up with two prompts. 

I don't think that's a huge deal though, and probably not worth the additional complexity to try to parse the command line at that stage

benhillis:
Thanks for making this change, merging.

PR #41121, opened 2026-07-20
CLI: Add quiet support to pull and align with Docker

dkbennett:

Add a `--quiet` / `-q` option to `wslc pull` with output behavior that matches `docker pull` exactly. In quiet mode, layer/progress output is suppressed and only the final canonical image reference is printed. As part of matching Docker's client-side behavior, `wslc pull` now also prints `Using default tag: latest` for name-only references and always prints the fully-qualified canonical reference as the final line.

This is an entirely client-side change (no IDL/service changes); the behavior was verified line-by-line against Docker's `cli/command/image/pull.go`.

image

## PR Checklist

`wslc pull` now mirrors the Docker CLI's `runPull` flow:

- **`--quiet` / `-q`** — suppresses progress output by passing a null progress callback to `ImageService::Pull` (the COM `IProgressCallback*` is already `[in, unique]`/nullable). Warnings are unaffected because the warning callback is built internally from the `Reporter`. Help text is "Suppress verbose output", matching Docker.
- **`Using default tag: latest`** — for a name-only reference (no tag or digest), the tag defaults to `latest` and, unless quiet, the client reports this on stdout before contacting the registry (exactly as Docker does).
- **Canonical reference always printed** — the resolved fully-qualified reference is always printed as the final line, in both quiet and non-quiet modes (e.g. `docker.io/library/ubuntu:latest`). Previously `wslc pull` did not print this line; this closes a Docker-parity gap.

A new helper, `wsl::windows::common::wslutil::GetCanonicalImageReference`, mirrors the client-side normalization in `github.com/distribution/reference` (`ParseNormalizedNamed` adds the `docker.io` default domain and `library/` prefix; `TagNameOnly` defaults a name-only reference to `:latest`) and produces the reference string printed by `docker pull`, preserving a tag and digest together when both are present.

**Example behavior (matches Docker):**

- `wslc pull ubuntu:latest -q` prints `docker.io/library/ubuntu:latest` (single line, no progress)
- `wslc pull alpine:latest` prints progress on stdout, then `docker.io/library/alpine:latest`
- `wslc pull ubuntu` prints `Using default tag: latest` first, then progress + canonical reference
- `wslc pull ubuntu:22.04@sha256:...` prints `docker.io/library/ubuntu:22.04@sha256:...` (both tag and digest preserved)

## Validation Steps Performed
* New tests ran and pass
* Existing pull tests pass
* Manual verification from deployed local build

### DIFF
--- a/src/windows/common/wslutil.cpp
+++ b/src/windows/common/wslutil.cpp
@@ -1437,15 +1437,23 @@ std::string wsl::windows::common::wslutil::GetCanonicalImageReference(const std:
 {
     // Mirror the Docker CLI's client-side reference normalization so the final line matches `docker pull` exactly.
     // See github.com/distribution/reference (normalize.go, reference.go) and github.com/docker/cli
-    // (cli/command/image/pull.go). A digest takes precedence over a tag here, consistent with ParseImage.
-    EnumReferenceFormat format = EnumReferenceFormatNone;
-    auto [repo, tagOrDigest] = ParseImage(input, &format);
-    auto [domain, path] = NormalizeRepo(repo);
+    // (cli/command/image/pull.go). Unlike ParseImage -- which collapses to a single tag-or-digest field with digest
+    // precedence -- Docker's canonical string keeps both a tag and a digest when both are present, so compose the
+    // reference directly from the parsed name, tag and digest groups.
+    static const auto regex = BuildImageReferenceRegex();
+    std::smatch match;
+    if (!std::regex_match(input, match, regex))
+    {
+        THROW_HR_WITH_USER_ERROR(E_INVALIDARG, wsl::shared::Localization::MessageWslcInvalidImage(input.c_str()));
+    }
 
-    // A digest is joined with '@', a tag with ':'; a name-only reference defaults to the ":latest" tag.
-    const std::string_view separator = (format == EnumReferenceFormatDigest) ? "@" : ":";
-    const std::string_view reference = (format == EnumReferenceFormatNone) ? "latest" : std::string_view{tagOrDigest.value()};
-    return std::format("{}/{}{}{}", domain, path, separator, reference);
+    auto [domain, path] = NormalizeRepo(match[1].str());
+
+    // A tag joins with ':' and a digest with '@'. A name-only reference (no tag and no digest) defaults to ":latest";
+    // a digest-only reference is not name-only, so it keeps no tag (matching Docker's TagNameOnly).
+    const std::string tag = match[2].matched ? std::format(":{}", match[2].str()) : (match[3].matched ? "" : ":latest");
+    const std::string digest = match[3].matched ? std::format("@{}", match[3].str()) : "";
+    return std::format("{}/{}{}{}", domain, path, tag, digest);
 }
 
 void wsl::windows::common::wslutil::PrintSystemError(_In_ HRESULT result, _Inout_ FILE* const stream)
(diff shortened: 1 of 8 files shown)


src/windows/common/wslutil.cpp
```diff
@@ -1433,6 +1433,29 @@ std::pair<std::string, std::optional<std::string>> wsl::windows::common::wslutil
...
+        THROW_HR_WITH_USER_ERROR(E_INVALIDARG, wsl::shared::Localization::MessageWslcInvalidImage(input.c_str()));
+    }
+
+    auto [domain, path] = NormalizeRepo(match[1].str());
+
+    // A tag joins with ':' and a digest with '@'. A name-only reference (no tag and no digest) defaults to ":latest";
+    // a digest-only reference is not name-only, so it keeps no tag (matching Docker's TagNameOnly).
+    const std::string tag = match[2].matched ? std::format(":{}", match[2].str()) : (match[3].matched ? "" : ":latest");
+    const std::string digest = match[3].matched ? std::format("@{}", match[3].str()) : "";
+    return std::format("{}/{}{}{}", domain, path, tag, digest);
```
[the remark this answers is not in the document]
dkbennett: My AI analysis disagrees with this AI's analysis

Docker does not lowercase the domain; it preserves it exactly and has tests which verify this.

The repository path does not lowercase - it rejects uppercase.

WSLC already mirrors these exactly, and the BuildImageReferenceRegex also covers this by allowing uppercase in the domain but disallowing it from the repository path. This will fail the regex and throw (which Docker also does).

I am making a change as a result of this feedback though - an additional test which verifies that the mixed-case domain is preserved and intentional to protect against future regression.

src/windows/common/wslutil.cpp
```diff
@@ -1433,6 +1433,29 @@ std::pair<std::string, std::optional<std::string>> wsl::windows::common::wslutil
     return {repo.str(), std::move(tagOrDigest)};
 }
 
+std::string wsl::windows::common::wslutil::GetCanonicalImageReference(const std::string& input)
+{
+    // Mirror the Docker CLI's client-side reference normalization so the final line matches `docker pull` exactly.
```
OneBlue, line 1438: I think this is a good idea, but I don't love forking the ParseImage() logic here. Maybe we could do something like this: 

```
struct ImageReference
{
    string repo;
    optional<string> tag;
    optional<string> digest
};


ImageReference ParseImage(const std::string& input);
```

That way `GetCanonicalImageReference` could just call `ParseImage()` ?

src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -223,10 +223,26 @@ void PullImage(CLIExecutionContext& context)
...
+    EnumReferenceFormat format = EnumReferenceFormatNone;
+    ParseImage(image, &format);
+    if (!quiet && format == EnumReferenceFormatNone)
+    {
+        context.Reporter.Output(L"{}\n", Localization::WSLCCLI_PullUsingDefaultTag(L"latest"));
+    }
 
+    // Match `docker pull`: in quiet mode, suppress progress output by passing no progress callback. Warnings are
+    // unaffected because the warning callback is built internally by ImageService::Pull from the Reporter.
     ImageProgressCallback callback(context.Reporter, Reporter::Level::Output);
```
OneBlue, line 240: nit: This could be simplified a bit via: 

```
std::optional<ImageProgressCallback> callback;

if (!quiet)
{
    callback.emplace(context.Reporter, Reporter::Level::Output);
}
```

This also skips the callback instantiation when we don't need it

OneBlue (commented):

OneBlue (approved):
Feel free to merge now and resolve the comments in a followup

src/windows/common/wslutil.cpp
dkbennett: I like this suggestion but it has a larger footprint so I will do this refactor in a separate PR as recommended.

PR #41125, opened 2026-07-20
Track implicit autobind ports from listen() in Consomme networking

### LINKED ISSUE #41117: [Consomme] TCP listener using implicit listen() autobind is unreachable via 127.0.0.1
### Windows Version

Microsoft Windows [Version 10.0.26200.8894]

### WSL Version

2.9.4.0

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.18.35.2-microsoft-standard-WSL2

### Distro Version

_No response_

### Other Software

fakeroot-tcp 1.38.1-1 (AUR)
Python 3.14.6 (used for the minimal reproduction)
paru 2.1.0

Relevant `.wslconfig` settings:

```ini
[wsl2]
memory=14GB
kernelCommandLine="WSL_DEBUG=hvsocket"
networkingMode=Consomme
virtio=true
virtiofs=true
ipv6=true
firewall=false
autoProxy=false
```

### Repro Steps

1. Configure WSL to use `networkingMode=Consomme` and run `wsl --shutdown`.
2. Start a WSL 2 distribution using the official Microsoft kernel.
3. Run this minimal reproduction inside WSL:

```bash
python - <<'PY'
import socket

server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
server.listen()
address = server.getsockname()
print(f"listener={address}")

client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
client.settimeout(2)
client.connect(("127.0.0.1", address[1]))
print("connected")
PY
```

`listen()` performs the standard Linux implicit autobind because no explicit `bind()` call was made. The resulting listener is visible as `0.0.0.0:<ephemeral-port>` in `ss -ltn`.

The real-world reproducer is `fakeroot-tcp`. Its `faked` daemon uses the same `socket(); listen(); getsockname()` sequence, and `libfakeroot` then connects to the assigned port through `127.0.0.1`. A stateful fakeroot operation fails:

```text
libfakeroot: connect: Connection refused
```

> An explicit-port control works though: starting `faked --port 46321` binds `127.0.0.1:46321`

Many wsl distros are distributed with fakeroot-tcp over fakeroot to also support WSL1, this breaks them. In my use case, for example, I can't install AUR packages because libfakeroot fails when using `fakeroot-tcp` with `Consomme` mode.

### Expected Behavior

The connection to the implicitly assigned listening port on `127.0.0.1` should succeed, as it does on native Linux and for a listener registered using an explicit `bind()` call. `fakeroot-tcp` should be able to communicate with its local `faked` daemon.

### Actual Behavior

The minimal reproduction raises:

```text
ConnectionRefusedError: [Errno 111] Connection refused
```

It seems like Consomme's virtio adapter only handles explicit `bind()` calls, and does not capture those implicit binds from `listen()`. To fix this, it probably needs to handle `listen()` calls as well

### Diagnostic Logs

[WslLogs-2026-07-20_11-48-58.tar.gz](https://github.com/user-attachments/files/30175253/WslLogs-2026-07-20_11-48-58.tar.gz)

benhillis:
## Summary

WSL2's Consomme networking mode only forwards TCP/UDP ports that were explicitly `bind()`'d in the guest. If an application calls `listen()` without first calling `bind()` (relying on the kernel's implicit autobind to an ephemeral port on the wildcard address), the port tracker never observes the bind, so the port is never registered for host forwarding and the socket is unreachable via `127.0.0.1:<port>` from Windows.

Fixes #41117

## Root cause

The seccomp filter installed by `RegisterSeccompHook()` in `src/linux/init/main.cpp` only ever trapped `bind()`. `listen()` was never trapped, so a `socket()` + `listen()` sequence with no prior `bind()` call was invisible to `GnsPortTracker`.

## What this changes

- Extends the seccomp BPF filter to also trap `listen()` across all supported architectures (native 64-bit `__NR_listen`, x86 compat via `socketcall(SYS_LISTEN, ...)`, and 32-bit ARM EABI via a new `ARMV7_NR_listen` constant).
- `GnsPortTracker::GetCallInfo` dispatches `listen()` calls through a new `ParseListen` path, reusing the existing `DeferredPortLookup`/`ResolvePortZeroBind` mechanism already used for explicit `bind(port=0)`.
- To avoid adding latency to the overwhelmingly common `bind()` + `listen()` sequence (where the port is already known before `listen()` runs), `ParseListen` first checks `getsockname()` synchronously and only falls back to the deferred/polling resolution path for genuine implicit-autobind-via-`listen()` calls (no prior `bind()`).

## Testing

- Added `ListenWithoutBindIsTracked` to both `NetworkTests::ConsommeTests` and `NetworkTests::MirroredTests`, mirroring the existing `PortZeroBindIsTracked` coverage but exercising `listen()` with no preceding `bind()`.
- Verified with a real before/after run against a full local build+deploy:
  - Pre-fix: `ConsommeTests::ListenWithoutBindIsTracked` **fails** (host can bind the port — Consomme never tracked it).
  - Post-fix: passes.
  - `MirroredTests::ListenWithoutBindIsTracked` passes both before and after — mirrored mode already has an independent port+protocol refresh mechanism (`OnRefreshAllocatedPorts`, added in #40287) that covers this case, so this change doesn't affect mirrored mode's behavior.
- Also validated the raw BPF filter bytes directly against the running kernel via a standalone seccomp-unotify harness, confirming the old filter never traps `listen()` and the new filter does (`__NR_listen`, syscall nr 50 on x86_64).

### DIFF
--- a/test/windows/NetworkTests.cpp
+++ b/test/windows/NetworkTests.cpp
@@ -2130,14 +2130,15 @@ class NetworkTests
 
         // Perl one-liner: socket() + listen() with no bind(), print the kernel-assigned
         // port via getsockname(), then accept() (blocking) to keep the socket alive.
-        const std::wstring wslCmd = L"perl -MSocket -e '"
-                                     L"$|=1;"
-                                     L"socket(S,AF_INET,SOCK_STREAM,0) or die;"
-                                     L"listen(S,5) or die;"
-                                     L"my $port=(sockaddr_in(getsockname(S)))[0];"
-                                     L"print \"PORT=$port\\n\";"
-                                     L"accept(C,S);"
-                                     L"'";
+        const std::wstring wslCmd =
+            L"perl -MSocket -e '"
+            L"$|=1;"
+            L"socket(S,AF_INET,SOCK_STREAM,0) or die;"
+            L"listen(S,5) or die;"
+            L"my $port=(sockaddr_in(getsockname(S)))[0];"
+            L"print \"PORT=$port\\n\";"
+            L"accept(C,S);"
+            L"'";
         auto cmd = LxssGenerateWslCommandLine(wslCmd.data());
 
         auto process = LxsstuStartProcess(cmd.data(), nullptr, stdOutWrite.get(), nullptr);
--- a/test/windows/NetworkTests.cpp
+++ b/test/windows/NetworkTests.cpp
@@ -2120,6 +2120,124 @@ class NetworkTests
         return {std::move(process), assignedPort};
     }
 
+    // Create a TCP listening socket in the guest via listen() WITHOUT ever calling bind() first
+    // (implicit autobind to an ephemeral port on the wildcard address, e.g. INADDR_ANY:0).
+    // This exercises the seccomp listen() trap added for the implicit-autobind port tracking fix,
+    // as opposed to BindGuestPortZero() which exercises the pre-existing explicit bind(0) path.
+    static std::tuple<unique_kill_process, uint16_t> BindGuestPortViaListenOnly()
+    {
+        auto [stdOutRead, stdOutWrite] = CreateSubprocessPipe(false, true);
+
+        // Perl one-liner: socket() + listen() with no bind(), print the kernel-assigned
+        // port via getsockname(), then accept() (blocking) to keep the socket alive.
+        const std::wstring wslCmd = L"perl -MSocket -e '"
+                                     L"$|=1;"
+                                     L"socket(S,AF_INET,SOCK_STREAM,0) or die;"
+                                     L"listen(S,5) or die;"
+                                     L"my $port=(sockaddr_in(getsockname(S)))[0];"
+                                     L"print \"PORT=$port\\n\";"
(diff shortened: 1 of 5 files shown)


test/windows/NetworkTests.cpp
```diff
@@ -2120,6 +2120,125 @@ class NetworkTests
...
+            LogInfo("output %hs", output.c_str());
+            std::string_view outputView(output.data(), writeOffset);
+            auto pos = outputView.find("PORT=");
+            if (pos != std::string_view::npos)
+            {
+                auto portStr = outputView.substr(pos + 5);
+                auto end = portStr.find_first_not_of("0123456789");
+                if (end != std::string_view::npos)
+                {
+                    portStr = portStr.substr(0, end);
```
OneBlue: Doing this can create a race. If the perl process writes "PORT=ABC" and read returned "PORT=AB", we could end up with an incomplete port number. 

We could fix this by reading until `\n`

OneBlue (approved):

PR #41129, opened 2026-07-21
Use single virtiofs device for all shares

damanm24:
With virtio-fs enabled, each drvfs share was backed by its own virtio-fs device. Each device consumes an MMIO aperture, of which there are a fixed amount available for a VM. When a VM runs out of them subsequent calls to mount shares will hang. 

To work around this limit, all drvfs shares are backed by a single aggregate virtio-fs device per server (one for the non-admin drvfs, and another for the admin). The aggregate device exposes a synthetic root with one child per share keyed by a GUID subname. Adding a share now adds a child to the existing device instead of creating a new device. In the guest the aggregate device is mounted once at a hidden location (`/run/wsl/virtiofs-mounts/drvfsa`) and each share is a bind-mount of `<device>/<child-name>` -> target.

Key Changes:
Wsl Service:

- `WslCoreVm::AddVirtioFsShare` now creates the aggregate device on first use (`VirtiofsShareKind_Aggregate`) and calls `AddVirtiofsChild(device, childName, options, path)` for every share
- New `GuestDeviceManager::AddVirtiofsChild` / `DeviceHostProxy::AddVirtiofsChild` + `GetVirtiofsDevice`, forwarding to the device-host IWslVirtiofsDevice::AddChild.

Protocol changes:
- `LX_INIT_ADD_VIRTIOFS_SHARE_RESPONSE_MESSAGE` gains `ChildNameOffset` so the guest learns which child to bind.

Guest init changes:

- `MountVirtioFs` mounts the aggregate root once, then bind-mounts <VIRTIOFS_MOUNT_DIR>/<tag>/<childName> → target. Falls back to Plan 9 if the share-add or the bind fails (so shares on volumes the server can't fully back still mount).
- `QueryVirtiofsMountSource` takes a Root argument to resolve an aggregate child back to its Windows path; config.cpp mount enumeration/remount handles the aggregate (non-/ root) case and skips the internal hidden mounts.
- `UtilFindMount` excludes internal mounts under VIRTIOFS_MOUNT_DIR from reverse (Windows→Linux) translation, so wslpath resolves to the user-facing `/mnt/c` instead of the internal `/run/wsl/virtiofs-mounts/... path`.


src/linux/init/drvfs.cpp
```diff
@@ -44,6 +47,173 @@ int MountFilesystem(const char* FsType, const char* Source, const char* Target,
...
+    const auto rootError = rootResult < 0 ? errno : 0;
+
+    LOG_INFO(
+        "VirtioFsMountState stage={} namespaceInode={} namespaceError={} rootTarget={} rootDevice={} rootInode={} rootError={} "
+        "childSource={} target={}",
+        Stage,
+        namespaceResult == 0 ? namespaceStat.st_ino : 0,
+        namespaceError,
+        RootTarget,
+        rootResult == 0 ? rootStat.st_dev : 0,
```
benhillis: won't rootStat already be zeroed from above?

benhillis: same for namespaceStat

benhillis: I assume this print is just for debugging.

src/linux/init/drvfs.cpp
```diff
@@ -44,6 +47,173 @@ int MountFilesystem(const char* FsType, const char* Source, const char* Target,
...
+    auto bindFlags = parsed.MountFlags;
+    if ((bindFlags & (MS_NOATIME | MS_STRICTATIME)) == 0)
+    {
+        bindFlags |= MS_RELATIME;
+    }
+
+    if (mount(nullptr, Target, nullptr, MS_BIND | MS_REMOUNT | bindFlags, parsed.StringOptions.c_str()) < 0)
+    {
+        const auto savedError = errno;
+        umount2(Target, MNT_DETACH);
```
benhillis, line 170: Why do we have to unmount first?

src/windows/common/DeviceHostProxy.cpp
```diff
@@ -151,8 +151,6 @@ GUID DeviceHostProxy::AddVirtiofsDevice(
             .rootPath = rootPath.get(),
             .kind = Kind,
             .shmemSizeMb = ShmemSizeMb,
-            // To workaround memory aperture limitations, limit virtiofs devices to one queue.
-            .queueCount = 1,
```
benhillis, line 155: I think we should probably still set the queueCount, otherwise it could potentially use a very large number.

benhillis: Maybe @Brian-Perkins has a suggestion on a good number?

test/linux/unit_tests/drvfs.c
```diff
@@ -1740,10 +1740,10 @@ Return Value:
     // directory entries are cached.
     //
     // N.B. This is not the case with Plan 9 because Linux doesn't know the
-    //      file system is case-insensitive.
```
benhillis, line 1743: Odd, why did we need to update the unit test assumptions for this change?

Brian-Perkins:
I suggest we make this a config option and set it to enabled by default, because (1) there may be issues that require disabling it; (2) longer term we are probably moving back to individual devices so it would probably be good to maintain and test that code. Agreed though that it is a pain to have to test both variations.

test/linux/unit_tests/drvfs.c
damanm24: My openvmm change fixed an issue with virtiofs mounts over an FAT share. Previously, virtiofs over FAT would fallback to using Plan9, and this test cases would be skipped. That's not the case anymore.

src/linux/init/drvfs.cpp
damanm24: Yes this is a debugging artifact. Removed

src/linux/init/drvfs.cpp
damanm24: We have two mounts here. The first creates the bind mount and the second one applies flags that are supposed to be specific to that share (such as readonly, etc; if we applied the flags at the first mount call then the flags get applied to everything under the aggregate share). This unmount call is for the case, where the first bind mount succeeds, but the subsequent mount fails. If we didn't unmount here, then the original virtiofs bind mount would persist.

src/linux/init/drvfs.cpp
```diff
@@ -44,6 +47,144 @@ int MountFilesystem(const char* FsType, const char* Source, const char* Target,
...
+        !wsl::shared::string::ToGuid(ChildName) || SubPath.empty() || SubPath.front() != '/')
+    {
+        errno = EINVAL;
+        return -1;
+    }
+
+    const auto rootTarget = std::format("{}/{}", VIRTIOFS_MOUNT_DIR, Tag);
+    const auto childSource = std::format("{}/{}{}", rootTarget, ChildName, SubPath == "/" ? std::string_view{} : SubPath);
+    {
+        std::lock_guard<std::mutex> lock(g_virtiofsDeviceMutex);
```
benhillis, line 109: Why is this lock needed?

src/linux/init/drvfs.cpp
```diff
@@ -625,22 +772,45 @@ try
...
+    {
+        const auto childError = errno;
+        LOG_WARNING("Mounting virtiofs child for {} failed {}, falling back to Plan9", Source, childError);
+
+        // Clear any mount-failure exit code so the Plan9 attempt isn't
+        // pre-judged as a failure.
+        if (ExitCode != nullptr)
+        {
+            *ExitCode = 0;
+        }
```
benhillis: Is this needed? Doesn't MountPlan9 set this anyway?

src/linux/init/drvfs.cpp
```diff
@@ -731,7 +913,16 @@ try
     // Validate the tag is a GUID.
     //
 
-    const auto Guid = wsl::shared::string::ToGuid(Tag);
+    std::string_view mappingName{Tag};
+    if (Root != nullptr)
+    {
+        if (const auto mountRoot = ParseAggregateVirtioFsMountRoot(Tag, Root))
```
benhillis, line 912: I'm worried mountRoot will go out of scope making mappingName (which is a non-owning string_view) point to stale memory. Probably better to just use a std::string for mappignName.

benhillis: Maybe this is fine since ParseAggregateVirtioFsMountRoot just sets a bunch of string views...

benhillis:
It does not look like WSLC was updated to use the new logic.

benhillis:
> It does not look like WSLC was updated to use the new logic.

That can be done as a follow-up if you like.

src/linux/init/drvfs.cpp
damanm24: Since MountVirtioFsChild can be called concurrently, it's guarding against a potential race between two threads that both observe that the root share has not been mounted yet, and they both try to mount the same virtiofs device at the same location.

src/linux/init/drvfs.cpp
damanm24: Yep removed

test/linux/unit_tests/lxtmount.c
```diff
@@ -630,14 +649,20 @@ Return Value:
...
+    // all shares report the same device number. A device-based lookup is
+    // therefore ambiguous and can return an arbitrary share's mount, which makes
+    // callers such as the drvfs parent-id checks nondeterministic.
     //
 
     FileSystem = NULL;
     Table = NULL;
     LxtCheckErrnoZeroSuccess(stat(Path, &Stat));
-    LxtCheckResult(MountFindMount(MOUNT_PROC_MOUNTINFO, NULL, Stat.st_dev, &Table, &FileSystem, MNT_ITER_BACKWARD));
+    LxtCheckResult(MountFindMount(MOUNT_PROC_MOUNTINFO, Path, 0, &Table, &FileSystem, MNT_ITER_BACKWARD));
```
benhillis: `mnt_table_find_target()` only matches an exact mount target, but `Path` explicitly does not have to be a mount point. In that case this still falls back to the device-number lookup, which remains ambiguous when aggregate virtio-fs shares all use the same device.

Can we parse the table here and use `mnt_table_find_mountpoint(Table, Path, MNT_ITER_BACKWARD)` instead? It walks up to the nearest containing mount without relying on `st_dev`, so the device fallback can be removed.

benhillis (approved):
Looks good, thanks for doing this!

PR #40466, opened 2026-05-08
Implement WSLC container group policies

benhillis:
Implement WSLC container group policies

Adds enforcement for two group policies whose ADMX templates landed in #40422:

- AllowWSLContainer (DWORD): master switch for wslc.exe and the WSLCSessionManager COM API. Enforced inside WSLCSessionManagerFactory so CoCreateInstance returns WSL_E_CONTAINER_DISABLED directly when the policy is off; every caller (wslc.exe, WslcSDK, plugins) fails fast through one code path rather than getting a manager whose every method errors out individually.

- WSLContainerCLIRegistryAllowlist: ADMX `<list valuePrefix=\"AllowedRegistry\">` policy stored as a sub-key whose REG_SZ value data is each allowed registry hostname. Enforced in PullImage and PushImage via a shared helper. BuildImage is currently rejected outright when an allowlist is configured. This will be addressed in a follow-up change.

Adds two new HRESULTs:
- WSL_E_CONTAINER_DISABLED (0x33)
- WSL_E_REGISTRY_BLOCKED_BY_POLICY (0x34)

Adds PolicyTests covering disabled-state, CLI surfacing, allowlist denial, build rejection, and pure-function unit tests for the allowlist evaluator.

PR #41133, opened 2026-07-21
wslc: switch image build from docker build to docker buildx build

beena352:
Change the command wslc runs inside the container VM for wslc image build from docker build to docker buildx build. Zero behavior change, same flags, same context mount, same rawjson output parsed by the existing BuildKitSolveStatus parser.
This is a preparatory PR. It unblocks a follow-up that will enforce the WSLContainerRegistryAllowlist group policy on image builds. Classic docker build cannot honor a BuildKit source policy; docker buildx build can. Splitting the invocation swap from the policy logic keeps both PRs small and gives a clean bisect point.

## PR Checklist

Why we're doing this
The WSLContainerRegistryAllowlist group policy (added in #40466) lists which registries wslc may talk to. PullImage and PushImage enforce it per-call, but BuildImage couldn't, a Dockerfile's FROMs resolve inside dockerd via ARGs, COPY --from=, --build-context, etc., which we can't reliably parse client-side. So #40466 took the strict-secure default: whenever the allowlist is configured, every wslc image build is refused outright. That leaves enterprise customers with the policy enabled unable to build images at all, even from an allowed registry.
The follow-up PR fixes that by pushing enforcement into BuildKit itself, via its per-invocation source-policy JSON (evaluated after ARG substitution, before any registry request). The catch: classic docker build in moby v25.0.7 silently drops the SourcePolicy field, no CLI flag or env var fills it in. docker buildx build with the default docker driver speaks BuildKit gRPC directly, so SourcePolicy (via EXPERIMENTAL_BUILDKIT_SOURCE_POLICY) travels intact.
This PR makes wslc image build go through buildx so the follow-up can enforce the policy. That's it.

Zero-behavior-change verification
Argv shape: buildx accepts every flag BuildImage passes (--progress=rawjson, -f -, -t, --build-arg, --label, --target, --no-cache, --pull) plus the mounted context path - unchanged.
Rawjson stream: all 14 fields the C++ parser destructures (vertexes, statuses, logs, digest, name, started, completed, error, id, vertex, current, total, data, stream) parse identically.
VM availability: buildx already ships at /usr/libexec/docker/cli-plugins/docker-buildx (BuildKit v0.12.5) - no new dependency.

### DIFF
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -914,7 +914,7 @@ try
     auto unmountFolder =
         wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { m_virtualMachine->UnmountWindowsFolder(mountPath.c_str()); });
 
-    std::vector<std::string> buildArgs{"/usr/bin/docker", "buildx", "build", "--progress=rawjson"};
+    std::vector<std::string> buildArgs{"/usr/bin/docker", "buildx", "build", "--builder", "default", "--progress=rawjson"};
     if (WI_IsFlagSet(Options->Flags, WSLCBuildImageFlagsNoCache))
     {
         buildArgs.push_back("--no-cache");
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -914,7 +914,7 @@ try
     auto unmountFolder =
         wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { m_virtualMachine->UnmountWindowsFolder(mountPath.c_str()); });
 
-    std::vector<std::string> buildArgs{"/usr/bin/docker", "build", "--progress=rawjson"};
+    std::vector<std::string> buildArgs{"/usr/bin/docker", "buildx", "build", "--progress=rawjson"};
     if (WI_IsFlagSet(Options->Flags, WSLCBuildImageFlagsNoCache))
     {
         buildArgs.push_back("--no-cache");
--- a/test/windows/WSLCTests.cpp
+++ b/test/windows/WSLCTests.cpp
@@ -1639,38 +1639,6 @@ class WSLCTests
         VERIFY_IS_TRUE(result.Output[1].find("Hello from a WSL container!") != std::string::npos);
     }
 
-    WSLC_TEST_METHOD(BuildImageIgnoresSelectedBuildxBuilder)
-    {
-        constexpr auto imageTag = "wslc-test-build-selected-builder:latest";
-        constexpr auto builderName = "wslc-test-unreachable-remote";
-        auto contextDir = std::filesystem::current_path() / "build-context-selected-builder";
-        std::filesystem::create_directories(contextDir);
-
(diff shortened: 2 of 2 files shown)


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -914,7 +914,7 @@ try
     auto unmountFolder =
         wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { m_virtualMachine->UnmountWindowsFolder(mountPath.c_str()); });
 
-    std::vector<std::string> buildArgs{"/usr/bin/docker", "build", "--progress=rawjson"};
+    std::vector<std::string> buildArgs{"/usr/bin/docker", "buildx", "build", "--progress=rawjson"};
```
benhillis: `docker build` and direct `docker buildx build` are not behaviorally equivalent when a non-default Buildx builder has been selected. Docker CLI v25 explicitly injects `BUILDX_BUILDER=<current context>` while forwarding `docker build` to Buildx so it "always create[s] a local docker image" even after `docker buildx ... --use` selected a container builder: https://github.com/docker/cli/blob/v25.0.7/cmd/docker/builder.go#L105-L116. Invoking `buildx build` directly bypasses that override. Docker documents that only the `docker` driver automatically loads images; other drivers retain a successful result only in build cache unless `--load` is specified: https://docs.docker.com/build/builders/drivers/#loading-to-local-image-store. Because this path neither pins the Docker-context builder nor passes `--load`, the command can return success without putting the tagged image in WSLC's local image store. Please pin the intended builder or explicitly load the result.

benhillis (changes requested):
The direct Buildx invocation loses Docker CLI forwarding behavior that guarantees the result is loaded into the local image store.

src/windows/wslcsession/WSLCSession.cpp
beena352: Good catch, invoking buildx build directly bypasses the BUILDX_BUILDER injection at cmd/docker/builder.go:105, so any buildx create --use state in the VM could silently break auto-load.
Pinned --builder default on argv (mirrors what the CLI bridge does with BUILDX_BUILDER, but stays visible in the BuildImageStart trace log for future debugging). Chose that over --load because --load implies "we might be on a container driver but force-export back", pinning is the more targeted intent.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -914,7 +914,7 @@ try
     auto unmountFolder =
         wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() { m_virtualMachine->UnmountWindowsFolder(mountPath.c_str()); });
 
-    std::vector<std::string> buildArgs{"/usr/bin/docker", "build", "--progress=rawjson"};
+    std::vector<std::string> buildArgs{"/usr/bin/docker", "buildx", "build", "--builder", "default", "--progress=rawjson"};
```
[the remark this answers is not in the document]
beena352: --load is redundant because --builder default already forces the docker driver, and the stale comment will be removed in PR 2.

benhillis (approved):

johnstep (approved):

PR #41140, opened 2026-07-22
Container lifecycle transitions

kvega005:
`WSLCContainerImpl` lifecycle operations (Start, Stop, Delete) now coordinate through a shared `StateTransition` object published under `m_lock`, replacing the per-operation `m_stopNotification`/`m_destroyEvent` handshake serialzied by `m_stopLock`; a second lifecycle request is no longer blocked behind an in-flight `Stop` that is waiting indefinitely (`WSLC_STOP_TIMEOUT_NONE`), and the fixed 60-second waits in `Stop()` and `Delete()` are removed. `DockerEventTracker` now invokes callbacks outside its own lock, fixing a lock inversion between `WSLCContainerImpl::Exec` and event delivery that wedges the entire Docker event stream, and an iterator invalidation when a container stop unregisters its exec callbacks mid-dispatch.

## PR Checklist

**Problem.** Container lifecycle state was driven by two standalone events and a mutex. `Stop()` acquired `m_stopLock` then `m_lock`, issued the Docker request, and waited up to `60s` on `m_stopNotification.Event`. `OnEvent(Stop)` acquired `m_stopLock` with `std::try_to_lock` and returned early when a `Stop` was already in flight, deferring cleanup to that caller. Ownership of stop cleanup was therefore decided by a try-lock race, and because `m_stopLock` is held for the entire duration of the wait, any second lifecycle request was blocked for as long as the first `Stop` waited — unbounded for `WSLC_STOP_TIMEOUT_NONE`. `Delete()` and the auto-remove path waited on `m_destroyEvent` under the same fixed `60s` cap, and `OnEvent(Destroy)` asserted `!m_destroyEvent.is_signaled()`. `OnEvent` was not `noexcept`, and `ContainerEvent::Start` was not handled at all, so a container started outside WSLC went unrecorded.

`DockerEventTracker::OnContainerEvent` held `m_lock` across `e.Callback(...)`, which produced two defects. First, a lock inversion: `WSLCContainerImpl::Exec` holds the container `m_lock` shared while constructing `DockerExecProcessControl`, whose constructor calls `RegisterExecStateUpdates` and takes the tracker lock; the event thread took those in the opposite order, holding the tracker lock while calling `WSLCContainerImpl::OnEvent`, which takes the container `m_lock` exclusive. Because the tracker is per-session, that deadlock stops delivery of all subsequent container and volume events. Second, iterator invalidation: the container's `die` callback reaches `ReleaseProcesses()`, which unregisters every live exec registration and erases from the same `m_containerCallbacks` vector the dispatch loop was iterating; the tracker's `std::recursive_mutex` allowed the erase to proceed and invalidate the loop's cached end iterator.

**Change.** Lifecycle requests now publish a `StateTransition` and wait on it:

```
struct StateTransition
{
    const TransitionKind Kind;               // Start | Stop | Delete
    ContainerEvent ExpectedEvent;
    wil::unique_event Completed{wil::EventOptions::ManualReset};
    std::exception_ptr Exception;
    unique_com_disconnect Wrapper;           // access under WSLCContainerImpl::m_lock
};
```
`StartTransition` publishes into `_Guarded_by_(m_lock) std::shared_ptr<StateTransition> m_transition`; `OnEvent` matches the arriving event against `ExpectedEvent` and calls `CompleteTransition`. `WaitForConflictingTransitionToComplete` serializes conflicting operations while allowing a caller of the same `TransitionKind` to attach to the in-flight transition rather than queue behind a mutex, which is what unblocks a concurrent `Kill` or shorter-timeout `Stop`. `WaitForTransitionCompletion` waits through `WSLCSession::CreateIOContext()` and an `EventHandle`, so the wait is bounded by session termination instead of a hard-coded `60s`. Failures raised on the event thread are captured into `StateTransition::Exception` and rethrown by `AttachToTransition` on the requesting thread, so `OnEvent` is now `noexcept`.

A second reader/writer lock, `m_lifecycleLock`, is held shared by lifecycle requests until their transition is published and exclusive by event delivery, closing the window between issuing the Docker request and publishing the transition. `StateTransition::Wrapper` carries the deferred `unique_com_disconnect` so `Disconnect()` runs on the COM caller after it leaves `OnEvent`'s critical section rather than on the event thread, where it would block draining in-flight COM callers. `OnStopped` now takes the exit code and, for `WSLCContainerFlagsRm`, upgrades the active transition's `ExpectedEvent` to `ContainerEvent::Destroy` so auto-remove completes under the same transition instead of a second wait. `StopNotification`, `m_destroyEvent`, and `m_stopLock` are removed, and `DeleteExclusiveLockHeld` becomes `RequestDeleteExclusiveLockHeld` since resource release now happens on the `Destroy` event.

**Event tracker locking notes.** `OnContainerEvent` and `OnVolumeEvent` copy the matching registrations into a local vector under `m_lock`, release it, and invoke through `InvokeCallbacks`. Registrations are stored as `std::shared_ptr<ContainerCallback>` / `std::shared_ptr<VolumeCallback>` so a snapshot entry stays alive if it is unregistered mid-dispatch. Both derive from a new `CallbackRegistration` base holding a `std::recursive_mutex InvokeLock` and a `_Guarded_by_(InvokeLock) bool Unregistered`; the dispatch loop holds `InvokeLock` while the callback runs and skips unregistered entries, and `UnregisterCallback` erases under `m_lock`, releases it, then acquires `InvokeLock` to block until any in-flight invocation completes. That preserves the previous guarantee that a callback never runs after its `EventTrackingReference` is reset — required because `DockerExecProcessControl` declares `m_eventTrackingReference` last and so unregisters before its other members are destroyed. `InvokeLock` is recursive so a running callback can unregister itself. With no callback running under `m_lock`, it is now a leaf lock and is downgraded from `std::recursive_mutex` to `std::mutex`. `ContainerEvent::Restart` is added for the Docker `restart` action.

**Lifecycle / state-machine impact.** `Start()` publishes a `Start` transition and commits `WslcContainerStateRunning` only when the Docker `start` event arrives; an unmatched `start` is logged as `UnexpectedContainerStart` instead of being ignored. `Stop()` skips creating a transition once `WslcContainerStateExited` is already observed, and does not create one when the signal is not `SIGKILL` and cannot be expected to terminate the container. `Delete()` publishes a `Delete` transition expecting `ContainerEvent::Destroy`, which is also where `ReleaseResources()` and the COM disconnect now run, so a `Stop` with auto-remove and an explicit `Delete` converge on the same path.

## Validation Steps Performed
- `WSLCTests::ConcurrentContainerStopAndKill` — guards against a second lifecycle request being blocked behind `m_stopLock` while an indefinite `Stop(WSLC_STOP_TIMEOUT_NONE)` is waiting; the `Kill` must reach Docker and return while the `Stop` is still outstanding.
- `WSLCTests::ConcurrentContainerStopTimeoutOverride` — guards against a shorter `Stop(..., 0)` being unable to override an in-flight indefinite `Stop`; both calls must complete.
- `WSLCTests::ConcurrentContainerStopAndStart` — asserts `Start` issued during an in-flight `Stop` fails with `WSLC_E_CONTAINER_IS_RUNNING` rather than racing the transition.
- `WSLCTests::ConcurrentContainerStopAndForceDelete` — asserts a force `Delete` during an in-flight `Stop` completes and both callers observe `WslcContainerStateDeleted`.
- `WSLCTests::ForceDeleteAutoRemoveContainer` — covers force-delete of a `WSLCContainerFlagsRm` container, where the stop transition is upgraded to expect `Destroy`.
- `WSLCTests::ExecContainerStopManyExecs` — guards against the mid-dispatch iterator invalidation. Uses eight concurrent execs, because the pre-existing `ExecContainerDelete` uses a single exec and leaves nothing after the erased entry for the dispatch loop to walk into.
- `WSLCTests::ExecContainerEventStress` — guards against the `Exec`/event-delivery lock inversion, and against a registration being destroyed while its own callback is in flight. Four threads run short-lived execs and release each process without waiting, so unregistration overlaps `exec_die` dispatch. A regression manifests as a hang rather than an assertion failure, since the deadlock is service-side and blocks the COM call itself.
- The `ConcurrentContainer*` tests use a `LaunchContainerWithBlockingStopHandler` helper whose init process traps the stop signal and blocks on stdin, so the in-flight `Stop` window is deterministic rather than timing-dependent.
- `PluginTests.cpp` expected log output adds `WSLC Container stopping, session=*, id=*`, reflecting that `OnContainerStopping` is now raised from `OnStopped`.


src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -995,28 +1034,31 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill)
...
+    auto transition = m_transition;
+
+    // A Stop while expecting Start should not occur normally: Docker emits start before die, and the event stream processes
+    //  them serially. It would indicate external manipulation. Ignoring it avoids applying an old exit code to the newly
+    //  staged init process.
+    if (transition && (transition->ExpectedEvent == ContainerEvent::Start))
+    {
+        WSL_LOG("UnexpectedContainerExit", TraceLoggingValue(m_id.c_str(), "Id"), TraceLoggingValue(exitCode, "ExitCode"));
+        return;
+    }
```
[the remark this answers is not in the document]
kvega005: This is inaccurate. If `docker.StartContainer()` succeeds, we are guaranteed that a Start event will arrive. If a "die" event arrives first, it must be from an older iteration of the container and thus we can ignore and continue to wait for start.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -873,54 +876,86 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
-        // (and any post-delete cleanup) rather than the prior Running/Exited state.
+        // Signal init exit after the state transition and resource cleanup so awaiters observe Deleted.
         SignalInitProcessExit();
+
+        if (transition)
+        {
+            WI_ASSERT(transition->ExpectedEvent == ContainerEvent::Destroy);
+            transition->Wrapper = std::move(comWrapper);
+            CompleteTransition(transition);
+        }
```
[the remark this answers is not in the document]
kvega005: Transitions can only be committed if docker accepts the request. the assertion is valid, because if docker accepts a start/stop request the container must not have been deleted. therefore, `OnEvent` will observe start/stop event and resolve the transition before observing `Destroy`.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -873,54 +876,85 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
+    auto lock = m_lock.lock_exclusive();
+    auto transition = m_transition;
 
-        // If Stop() is already in flight, it will wake when the stop event is signaled and take care of cleanup.
-        if (!stopGuard.owns_lock())
+    if (event == ContainerEvent::Start)
+    {
+        // Only WSLC should start the container, so if we receive a start event, it must be expected by a transition.
+        // Otherwise the container was started externally. Log if the container was started externally.
+        if (transition && transition->ExpectedEvent == ContainerEvent::Start)
```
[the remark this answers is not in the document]
kvega005: Restart is not yet supported.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1031,46 +1073,70 @@ __requires_exclusive_lock_held(m_lock) unique_com_disconnect WSLCContainerImpl::
...
+
+    transition = std::make_shared<StateTransition>(ContainerEvent::Destroy);
+    WI_ASSERT(!m_transition);
+    m_transition = transition;
+
+    lock.reset();
+    WaitForTransition(transition);
+
+    lock = m_lock.lock_exclusive();
 }
```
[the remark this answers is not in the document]
kvega005: Explained in comment.

src/windows/wslcsession/DockerEventTracker.cpp
```diff
@@ -153,7 +153,11 @@ void DockerEventTracker::OnEvent(const std::string_view& event)
 void DockerEventTracker::OnContainerEvent(const nlohmann::json& parsed, const std::string& action, std::uint64_t eventTime)
 {
     static std::map<std::string, ContainerEvent> events{
-        {"start", ContainerEvent::Start}, {"die", ContainerEvent::Stop}, {"destroy", ContainerEvent::Destroy}, {"exec_die", ContainerEvent::ExecDied}};
+        {"start", ContainerEvent::Start},
+        {"die", ContainerEvent::Stop},
+        {"destroy", ContainerEvent::Destroy},
+        {"exec_die", ContainerEvent::ExecDied},
+        {"restart", ContainerEvent::Restart}};
```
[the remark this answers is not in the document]
kvega005: This is ok, restart is not yet supported.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -873,54 +876,86 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
     cleanup.release();
 }
 
-void WSLCContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCode, std::uint64_t eventTime)
+void WSLCContainerImpl::WaitForTransition(const std::shared_ptr<StateTransition>& transition) const
 {
-    // We must release m_lock and m_stopLock before the wrapper's destructor calls
-    // Disconnect(), so in-flight COM callers can drain from COMImplClass::m_callers.
-    unique_com_disconnect comWrapper;
+    THROW_HR_IF_MSG(
```
OneBlue: I think it would also be a good idea to plumb the caller's process here. The easiest way to do that would be to use `WSLCSession::CreateIOContext()` to setup the wait. 

That way if a caller does `wslc stop <id>` and then ctrl-C, and the stop is stuck, the service thread doing the wait will be released

test/windows/PluginTests.cpp
```diff
@@ -721,6 +721,7 @@ class PluginTests
             WSLC Image created, session=*, id=sha256:*, name=wslc-registry:latest
             WSLC Container started, session=*, id=*, name=*, image=wslc-registry:latest, state=running
             WSLC Image created, session=*, id=sha256:*, name=127.0.0.1:5000/debian:latest
+            WSLC Container stopping, session=*, id=*
```
OneBlue, line 724: Interesting, I guess there was a race before and we didn't record the transition correctly before the session terminated ?

test/windows/WSLCTests.cpp
```diff
@@ -6826,6 +6826,94 @@ class WSLCTests
...
+        container.SetDeleteOnClose(false);
+
+        VERIFY_ARE_EQUAL(container.State(), WslcContainerStateDeleted);
+        VERIFY_ARE_EQUAL(container.Get().Delete(WSLCDeleteFlagsForce), RPC_E_DISCONNECTED);
+
+        wil::com_ptr<IWSLCContainer> openedContainer;
+        VERIFY_ARE_EQUAL(m_defaultSession->OpenContainer(id.c_str(), &openedContainer), WSLC_E_CONTAINER_NOT_FOUND);
+        VERIFY_ARE_EQUAL(m_defaultSession->OpenContainer(name.c_str(), &openedContainer), WSLC_E_CONTAINER_NOT_FOUND);
+    }
+
```
OneBlue, line 7068: I recommend also test coverage for the "stuck stop" case, where we do something like: 

wslc run -it debian:latest sleep 9999999

then in thread A: wslc stop --timeout -1  <id> # Should get stuck forever

And in thread B: wslc stop --timeout 0 <id> # Should actually kill the container, and both threads should successfully exit

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -873,54 +876,86 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
         }
     }
 
+    transition = std::make_shared<StateTransition>(ContainerEvent::Start);
```
OneBlue: We're doing this (creating a new `StateTransition` + WI_ASSERT() +  assigning to `m_transition` in 3 places so I think this should be worth extracting to a `StartTransition()` method (which would return the shared_ptr)

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -873,150 +874,226 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
 
-        // Signal init exit after the state transition so awaiters observe state=Deleted
-        // (and any post-delete cleanup) rather than the prior Running/Exited state.
-        SignalInitProcessExit();
-    }
+void WSLCContainerImpl::WaitForTransitionCompletion(const std::shared_ptr<StateTransition>& transition) const
+{
+    THROW_HR_IF_MSG(
+        E_UNEXPECTED,
+        !m_wslcSession.WaitForEventOrSessionTerminating(transition->Completed.get(), 60s),
```
OneBlue: I'm not sure that we want a timeout here. On the Stop() path, if an "infinite stop" is running, and another stop arrives, we shouldn't time out when waiting for the state change

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -873,150 +874,226 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
+        auto lock = m_lock.lock_exclusive();
+        WaitForConflictingTransitionToComplete(lock, TransitionKind::Stop);
+
+        if (m_transition)
         {
-            THROW_DOCKER_USER_ERROR_MSG(e, "Failed to %hs container '%hs'", Kill ? "kill" : "stop", m_id.c_str());
+            WI_ASSERT(m_transition->Kind == TransitionKind::Stop);
+            transition = m_transition;
+            lock.reset();
+            AttachToTransition(transition);
```
OneBlue: Unfortunately I don't think we can just attach to an existing stop in progress in this case. 

If we have: 

```
wslc stop <container> -t -1
```

Followed by: 

```
wslc stop <container> -t 0
```

The second Stop() call should also a stop API call to docker as well

test/windows/PluginTests.cpp
kvega005: Not really a race. More like, deterministically we didn't record the stop event before the session terminated because we didn't wait for the stop event to arrive. Now we do.

OneBlue (changes requested):
Unfortunately there seems to be two threading issues with the current state of this change: 

```
1. Deadlock between  Exec()  and event delivery
 WSLCContainer.cpp:957  now acquires  m_lock  for every event, including  ExecDied . Concurrently,  Exec()  holds  m_lock  from line 1390 while registering its callback, which acquires  DockerEventTracker::m_lock  at  DockerEventTracker.cpp:271 .
Interleaving:
• Event thread holds tracker lock at line 194, then waits for container lock.
•  Exec()  holds container lock, then waits for tracker lock.
This deadlocks the global event tracker, preventing all subsequent lifecycle events. Avoid taking container lifecycle locks for  ExecDied / Restart , or preferably invoke callbacks outside the tracker lock.

2. Stop handling invalidates the callback vector being iterated
The tracker iterates  m_containerCallbacks  at  DockerEventTracker.cpp:196 . The container callback now calls  OnStopped()  ( WSLCContainer.cpp:979 ), which calls  ReleaseProcesses()  at line 1130. Each exec control unregisters its callback, erasing elements from that same vector.
Because the tracker uses a recursive mutex, erasure succeeds during iteration, invalidating the cached iterator/end and causing undefined behavior—potentially skipped callbacks, use-after-free, or a crash. Previously, normal in-flight Stop cleanup happened on the Stop caller rather than inside this callback.
Snapshot matching callbacks before invocation, or otherwise defer callback removal until iteration completes.
```

I think those shouldn't be too hard to solve though (we probably need to rethink callback registration in the event tracker so those can be done outside the lock)

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1192,150 +1193,247 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
+            }
+            else
+            {
+                WSL_LOG("UnexpectedContainerStart", TraceLoggingValue(m_id.c_str(), "Id"));
             }
         }
-        else
+        else if (event == ContainerEvent::Stop)
+        {
+            WI_ASSERT(exitCode.has_value());
```
ptrivedi, line 1285: I think we should let the Stop cleanup continue even when there is no exitCode present? we could remove the assert here and then make a change to OnStopped() so that it will check and record existCode only when it is present

I am afraid that if we just return when the exitCode is missing, a Stop() call could wait indefinitely. We should only skip SetExitCode() when exitCode is not there, but we should let the cleanup carry on

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1192,150 +1193,247 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
-        if (Kill || e.StatusCode() != 304)
+        auto lifecycleLock = m_lifecycleLock.lock_shared();
+        auto lock = m_lock.lock_exclusive();
+        WaitForConflictingTransitionToComplete(lock, lifecycleLock, TransitionKind::Stop);
+
+        transition = m_transition;
+        WI_ASSERT(!transition || transition->Kind == TransitionKind::Stop);
+
+        // There can be an active stop transition post observing the exited state for cases where additional work needs to be done
+        // after the contaier stopped: e.g. auto remove, restart, etc. Therefore, if there is an active stop transition, we still
```
ptrivedi:
```suggestion
        // after the container stopped: e.g. auto remove, restart, etc. Therefore, if there is an active stop transition, we still
```

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1192,150 +1193,247 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
+                SignalArg = Signal;
+            }
+
+            ValidateStopTimeout(TimeoutSeconds, true);
+
+            // Don't wait for the container to stop if we're not sending SIGKILL, since it may not stop the container.
+            // N.B. If the signal was SIGTERM for instance, we'll receive the stop notification via OnEvent().
+            bool waitForStop = !Kill || (SignalArg.value_or(WSLCSignalSIGKILL) == WSLCSignalSIGKILL);
+
+            lock.reset();
```
beena352, line 1365: Trying to understand: We release  m_lock  here but keep  lifecycleLock  (at 1334) held shared, so with  WSLC_STOP_TIMEOUT_NONE  we hold it shared until the container actually exits.
If an exec exits during that window,  OnEvent  takes  m_lifecycleLock  exclusive, it does that for every event before dispatch, and the container callback registers with no exec ID, so it also matches  exec_die . While  OnEvent  is waiting on that exclusive acquire, would a second  Stop(..., 0)  block trying to take the lock shared? That second stop is what the first one is waiting on, so I want to make sure this can't end up in a circular wait.
The test ConcurrentContainerStopTimeoutOverride  wouldn't catch this because no exec is running there, so nothing queues for the exclusive lock while the first stop is blocked.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1192,150 +1193,247 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
+    // Either owner may disconnect the COM wrapper, so both must outlive m_lock.
+    unique_com_disconnect comWrapper;
+    std::shared_ptr<StateTransition> transition;
 
-    try
     {
-        if (Kill)
-        {
-            m_runtime.Docker().SignalContainer(m_id, SignalArg);
+        auto lifecycleLock = m_lifecycleLock.lock_exclusive();
```
ptrivedi, line 1264: so, the exclusive lifestyle lock is taken by OnEvent here before determining if there's any work needed. In the meantime, a longish stop operation can hold that lock shared as well, and say an exec that tries to exit while that stop is running, could block the single docker event-reading thread. Won't that delay lifecycle events for every container in the session?

Shoudl we route the exec_die events to the matching exec callback and also drop the unsupported restart events before they reach here?

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1192,150 +1193,247 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
+    m_transition = transition;
+    return transition;
+}
 
-        // Signal init exit after the state transition so awaiters observe state=Deleted
-        // (and any post-delete cleanup) rather than the prior Running/Exited state.
-        SignalInitProcessExit();
-    }
+void WSLCContainerImpl::WaitForTransitionCompletion(const std::shared_ptr<StateTransition>& transition) const
+{
```
ptrivedi, line 1223: would be useful if we could test that terminating the waiting client returns E_ABORT but doesn't invalidate the published transition... that another caller or event can still finish it safely

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1350,26 +1448,51 @@ __requires_exclusive_lock_held(m_lock) unique_com_disconnect WSLCContainerImpl::
...
-    // For the Rm path, defer init-exit signaling to OnEvent(Destroy) so callers waiting
-    // on init exit observe destroy-side cleanup first.
-    if (WI_IsFlagClear(m_containerFlags, WSLCContainerFlagsRm))
+    SignalInitProcessExit();
+
+    if (transition)
     {
-        SignalInitProcessExit();
+        CompleteTransition(transition, std::move(transitionException));
     }
```
ptrivedi, line 1501: this all seems asycnchronous here, could one of the operations being slow over here delay events for other unrelated containers? should we consider queueing this cleanup so that the transition can finish asynchronously?

src/windows/wslcsession/WSLCContainer.h
```diff
@@ -151,17 +151,52 @@ class WSLCContainerImpl : public std::enable_shared_from_this<WSLCContainerImpl>
...
+    {
+        StateTransition(TransitionKind kind, ContainerEvent expectedEvent) : Kind(kind), ExpectedEvent(expectedEvent)
+        {
+        }
+
+        const TransitionKind Kind;
+        ContainerEvent ExpectedEvent;
+        wil::unique_event Completed{wil::EventOptions::ManualReset};
+        std::exception_ptr Exception;
+        // Access under WSLCContainerImpl::m_lock.
```
ptrivedi, line 171: I think the comment above applies to Exception as well right? Also to ExpectedEvent? Would be good to update it so it's clear to other devs

src/windows/wslcsession/WSLCContainer.cpp
OneBlue: Yeah sadly I think that we can't hold `m_lifecycleLock` while we do the Stop() API call to docker, since it can block

OneBlue: Looking `DockerEventTracker::OnContainerEvent()` though, I do think that it would be a good idea to exclude exec event updates to "container level" callbacks, the current condition will indeed let those go through: 

```
        if (e.ContainerId == containerId && (!e.ExecId.has_value() || e.ExecId == execId))
```

I think we could just switch to comparsing the optionals, and that should solve that:

```
        if (e.ContainerId == containerId && e.ExecId == execId)

```

src/windows/wslcsession/WSLCContainer.cpp
OneBlue: `exitCode` being absent there would be a symptom of a deeper problem, since we'd essentially be a in broken state whatever we do. 

If we wanted to be ultra safe here, we have a LOG_HR + default to something like `255`

OneBlue (changes requested):

src/windows/wslcsession/WSLCContainer.cpp
kvega005: I think `Stop()` doesn't actually need the lifecycle lock across the docker request. The lock only exists to keep `OnEvent` out until a transition is published. Checking state under `m_lock` afterwards covers the same window.

kvega005: As far as excluding exec, I think it's not super critical at the moment and I can leave for a follow-up. I want to make sure we don't actually need to for container events before I add that code.

src/windows/wslcsession/WSLCContainer.h
kvega005: Updated. They aren't all the same though, so I documented the actual discipline rather than extending the one comment: ExpectedEvent and Wrapper are guarded by m_lock (Wrapper specifically because multiple callers can attach to one transition and would otherwise both move it out), while Exception is published by Completed — written before the event is signaled and only read after WaitForTransitionCompletion, which is why AttachToTransition reads it outside the lock.

src/windows/wslcsession/WSLCContainer.cpp
kvega005: Yes, OnEvent runs synchronously on the shared Docker event stream, so a slow handler can delay events for unrelated containers. That serialization and the stop cleanup on this path both predate this PR.

This PR does not materially increase that delay though. The transition-specific work here is local bookkeeping and event signaling (`SignalInitProcessExit()` / `CompleteTransition()`), with no additional I/O. Queueing the broader cleanup could improve isolation, but that would be a separate architectural change requiring us to preserve per-container event ordering and object lifetime. I’d prefer to handle that separately.

src/windows/wslcsession/WSLCContainer.cpp
kvega005: I think we can do this with an E2E test, using the CLI. I'll look into it.

src/windows/wslcsession/WSLCContainer.cpp
kvega005: You are right. There was a real issue here.

`Stop` now releases `m_lifecycleLock` together with `m_lock` before issuing the Docker request, and never re-takes it — only `m_lock` is re-acquired afterwards to publish the transition, and `AttachToTransition()` runs with neither held. So with `WSLC_STOP_TIMEOUT_NONE` the shared lock is held only for the bounded state check up front, the pending `OnEvent` writer drains promptly, and the second `Stop` isn't stuck behind it.

OneBlue: Unfortunately, releasing the lock causes another potential hang in Stop(), here's a sample sequence of operation that could lead to that (copilot generated): 

```
m_lifecycleLock  lets the container stop and restart before Stop publishes its transition, causing an indefinite wait
 WSLCContainer.cpp:1375-1376 ,  WSLCContainer.cpp:1407-1414 
Removing the lifecycle lock across the blocking Docker call fixes the previously reported event-stream deadlock, but checking only  m_state  afterwards is insufficient because state can complete an entire cycle while the Stop thread is descheduled:
1. Stop releases both locks and calls Docker.
2.  die  is processed; state becomes  Exited , with no transition.
3. A concurrent Start observes  Exited , starts the container, receives  start , and commits  Running .
4. The original Stop returns from Docker and reacquires  m_lock .
5. It sees  Running  and no transition, creates a new Stop transition, and waits for a second  die  event even though it never stopped this new incarnation.
```

OneBlue: The best way I can think of to solve this would be to keep track of the "generation" of container that we're operating on (protected by m_lock) , so when stop re-acquires `m_lock`, it can compare the current generation with the one it started operating on, and just exit if it's higher.

OneBlue (changes requested):

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1181,150 +1182,260 @@ void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessSt
...
+                    transition.reset();
+                }
+            }
         }
     }
 
-    // Wait for the stop event to get the Docker timestamp.
-    std::optional<std::int64_t> stopTimestamp;
-    if (m_wslcSession.WaitForEventOrSessionTerminating(m_stopNotification.Event.get(), 60s))
+    if (transition)
```
OneBlue, line 1423: If we have thread A doing something like: 

wslc stop <id> 

And in parallel thread B does: 

wslc kill -s SIGUSR1 <id>, 

thread will have `waitForStop = false`, but will have `transition` set to the pending stop transition, so when we reach this block, thread B will attach to thread A's transition, which is not what we want. 

The easiest fix I can think of would be switching to: 

```
if (waitForStop)
{
    lock = m_lock.lock_exclusive();
    [...]
}
else
{
    transition.reset();
}
```

It might also be a good idea to add a test case covering that specific potential hang. We could have a purposefully hung stop, followed by a kill with a dummy signal like USR1 in a separate thread, to validate that the second thread doesn't get stuck

OneBlue (approved):
LGTM, thank you for doing this. This change will give us a much stronger foundation than the previous implementation and will unblock things like parallel stops & wslc restart

PR #41143, opened 2026-07-22
Remove narrow-to-wide formatter specializations

benhillis:
## Summary

Follow-up to #41071. Remove the custom narrow-to-wide `std::formatter` specializations entirely and make cross-encoding explicit at the appropriate boundary.

Generated localization helpers now adapt `char*`, character arrays, `std::basic_string<char>`, and `std::basic_string_view<char>` arguments through `wsl::shared::string::MultiByteToWide`. Adapted values are materialized as local variables before `make_wformat_args` so their storage remains valid through `vformat`.

Direct wide `std::format`, Reporter, and user-error paths convert narrow values explicitly.

## Validation

- Clean full x64 Debug build
- `WSLCTests::InvalidNames`
- `WSLCVolumeMountUnitTests::VolumeMount_IsValidNamedVolumeName_InvalidNames`
- `UnitTests::WslInfo`

### DIFF
--- a/test/windows/wslc/e2e/WSLCE2EHelpers.cpp
+++ b/test/windows/wslc/e2e/WSLCE2EHelpers.cpp
@@ -97,7 +97,11 @@ namespace {
             std::filesystem::remove_all(storagePath, error);
             if (error)
             {
-                Log::Error(std::format(L"Failed to cleanup storage path {}: {}", storagePath.wstring(), error.message()).c_str());
+                Log::Error(std::format(
+                               L"Failed to cleanup storage path {}: {}",
+                               storagePath.wstring(),
+                               wsl::shared::string::MultiByteToWide(error.message()))
+                               .c_str());
             }
         }
     }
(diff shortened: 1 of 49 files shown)


test/windows/wslc/e2e/WSLCE2EHelpers.cpp
```diff
@@ -367,8 +371,8 @@ void EnsureImageContainersAreDeleted(const TestImage& image)
         auto nameAndTag = wsl::shared::string::WideToMultiByte(image.NameAndTag());
         if (container.Image.find(nameAndTag) != std::string::npos)
         {
-            auto result = RunWslc(std::format(L"container remove --force {}", container.Id));
-            result.Verify({.Stdout = std::format(L"{}\r\n", container.Id), .Stderr = L"", .ExitCode = 0});
+            auto result = RunWslc(std::format(L"container remove --force {}", wsl::shared::string::MultiByteToWide(container.Id)));
```
dkbennett: Nit: maybe ask copilot to convert all these long namespaces to have a using wsl::shared::string::MultiByteToWide and reduce some namespace clutter.

dkbennett (commented):
Its unfortunate that we have to remove this specifier; I was relying on it in the Reporter class to keep the string use simple and naturally handle wide and narrow strings. I might update the signature of the Reporter in a followup to regain this functionality but this looks fine for me. Just a minor nit about the namespace clutter where MultiByteToWide was added.

Would be nice if we could align around wide strings wherever possible and maybe we wouldn't need so many MultiByteToWide calls in the client code.

PR #41151, opened 2026-07-23
virtiofs: Use aggregate share feature in WSLC

damanm24:
This is a follow-up to PR: #41129 which removes the need to limit the amount of virtiofs shares in WSLC. It does so by mounting all virtiofs shares under a single, unified root share which limits the memory footprint of virtio devices. See the original PR for more details.

### DIFF
--- a/src/shared/inc/lxinitshared.h
+++ b/src/shared/inc/lxinitshared.h
@@ -1642,6 +1642,7 @@ struct WSLC_MOUNT
     unsigned int TypeIndex{};
     unsigned int OptionsIndex{};
     unsigned int Flags{};
+    unsigned int ChildNameIndex{};
 
     enum MountType : uint8_t
     {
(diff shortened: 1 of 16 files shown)


src/shared/inc/lxinitshared.h
```diff
@@ -1654,7 +1655,7 @@ struct WSLC_MOUNT
 
     char Buffer[];
 
-    PRETTY_PRINT(FIELD(Header), STRING_FIELD(SourceIndex), STRING_FIELD(DestinationIndex), STRING_FIELD(TypeIndex), STRING_FIELD(OptionsIndex));
+    PRETTY_PRINT(FIELD(Header), STRING_FIELD(SourceIndex), STRING_FIELD(DestinationIndex), STRING_FIELD(TypeIndex), STRING_FIELD(OptionsIndex), STRING_FIELD(ChildNameIndex));
```
benhillis, line 1678: The Mount message is pretty generic, and ChildName is specific to VirtioFs. I'm not sure if I have a better solution though.

damanm24: @OneBlue do you have an opinion here? The ChildNameIndex ends up being used to construct the source argument in the `mount` syscall. 

Some of the options we discussed offline are:
1. introducing a virtiofs specific message for WSLC_MOUNTS
2. Having sourceIndex be a concatenated string of <device-tag>/<child-guid>
3. keeping this as is

OneBlue: Yeah the mount message was originally designed as a generic mount, so adding another field here feels a bit off. 

One solution I can think off would be to inject a mount flag in the mount options, something like `childname=<name>`, and then update `MountParseFlags()` to allow the caller to extract a specific mount option so `MountVirtioFsChild()` can get the child value from there 

If that's too complicated maybe we can subclass that message into a `WSLC_MOUNT_VIRTIOFS` message that adds an additional field

damanm24: Discussed offline. Going with forking the mount message.

benhillis (approved):

benhillis:
Looks good to me!

PR #41157, opened 2026-07-23
Add --output flag to wslc build

ggarzia-MSFT:
This pull request introduces support for the `--output` option to the image build command, allowing users to specify a build output destination using the Docker Buildx exporter spec (e.g., `type=local,dest=path` or `type=tar,dest=out.tar`). The changes include parsing and validation logic for the new option, updates to the service interface and flags, and user-facing descriptions and error messages.

**Support for `--output` option and Buildx exporter spec:**

* Added parsing, validation, and formatting functions for the Buildx `--output` spec in `SpecParsing.cpp` and `SpecParsing.h`, including handling for CSV quoting, supported exporter types, and destination resolution. Also added helpers to determine if the output streams to the client or is a directory exporter. [[1]](diffhunk://#diff-0b75e608446791966cc2f262c6de7dbb8210668a15877301d4107502a1907599R194-R497) [[2]](diffhunk://#diff-689916116384a710c025948a0c211815525d8328bd70e2d8cb8d36db598cc519L26-R27) [[3]](diffhunk://#diff-689916116384a710c025948a0c211815525d8328bd70e2d8cb8d36db598cc519R47-R62)
* Updated the `ImageBuildCommand` to accept the new `--output` argument with an appropriate user-facing description. [[1]](diffhunk://#diff-d83b8938c1848772336f481cc4f552d3d29ac1c4b31a4240359797d561910bb3R37) [[2]](diffhunk://#diff-c64d306b3b9e1059cb6e28f3b1a3552de31a1a94490601541c4e43547f4ee8efR2898-R2901)

**Service and API updates:**

* Extended `WSLCBuildImageOptions` and related IDL definitions to include `Output` and `OutputHandle` fields, and added a new flag `WSLCBuildImageFlagsOutputIsDirectory` to indicate directory exporters. [[1]](diffhunk://#diff-f354d8f5022677f418046fbe5c50ab602814c13ed9d0041e3d238954a153b757R552-R553) [[2]](diffhunk://#diff-31454ae55bf107b64eb0765c40ca90aef82b4ba8b50d94edbf1cb2f7b0ed0f9aR223-R226)
* Modified the `ImageService::Build` method signature to accept the new `output` parameter and included necessary header imports. [[1]](diffhunk://#diff-02bd98b2ee17399f9362d63153b45508809c4508078bcb2207b466d3171babe1R17-R18) [[2]](diffhunk://#diff-02bd98b2ee17399f9362d63153b45508809c4508078bcb2207b466d3171babe1R126)

**User experience improvements:**

* Added localized error messages and help text for invalid or missing `--output` values in `Resources.resw`. [[1]](diffhunk://#diff-c64d306b3b9e1059cb6e28f3b1a3552de31a1a94490601541c4e43547f4ee8efR2333-R2336) [[2]](diffhunk://#diff-c64d306b3b9e1059cb6e28f3b1a3552de31a1a94490601541c4e43547f4ee8efR2898-R2901)

## PR Checklist

## Validation Steps Performed

### DIFF
--- a/src/shared/inc/stringshared.h
+++ b/src/shared/inc/stringshared.h
@@ -69,7 +69,7 @@ inline bool EndsWith(const std::basic_string<T>& String, const std::basic_string
 // locale-independent (no Turkish-'I' surprises) and has no signed-char UB, which is what you want when
 // normalizing ASCII protocol tokens such as buildx CSV keys.
 template <class T>
-inline std::basic_string<T> AsciiToLower(const std::basic_string_view<T> &String)
+inline std::basic_string<T> AsciiToLower(const std::basic_string_view<T>& String)
 {
     std::basic_string<T> Result(String);
     for (auto& Ch : Result)
@@ -87,7 +87,7 @@ inline std::basic_string<T> AsciiToLower(const std::basic_string_view<T> &String
 // the ASCII subset of Go's strings.TrimSpace. Returns a view into the input, so the input must outlive
 // the result. The stdlib has no trim, so this centralizes the find_first/last_not_of idiom.
 template <class T>
-inline std::basic_string_view<T> TrimAscii(const std::basic_string_view<T> &String)
+inline std::basic_string_view<T> TrimAscii(const std::basic_string_view<T>& String)
 {
     constexpr T Whitespace[] = {
         static_cast<T>(' '), static_cast<T>('\t'), static_cast<T>('\r'), static_cast<T>('\n'), static_cast<T>('\v'), static_cast<T>('\f'), static_cast<T>('\0')};
--- a/src/shared/inc/stringshared.h
+++ b/src/shared/inc/stringshared.h
@@ -65,6 +65,43 @@ inline bool EndsWith(const std::basic_string<T>& String, const std::basic_string
     return std::equal(Suffix.rbegin(), Suffix.rend(), String.rbegin());
 }
 
+// Lowercases ASCII 'A'-'Z' only, leaving every other code unit untouched. Unlike std::tolower this is
+// locale-independent (no Turkish-'I' surprises) and has no signed-char UB, which is what you want when
+// normalizing ASCII protocol tokens such as buildx CSV keys.
+template <class T>
+inline std::basic_string<T> AsciiToLower(std::basic_string_view<T> String)
+{
+    std::basic_string<T> Result(String);
+    for (auto& Ch : Result)
+    {
+        if (Ch >= static_cast<T>('A') && Ch <= static_cast<T>('Z'))
+        {
+            Ch = static_cast<T>(Ch - static_cast<T>('A') + static_cast<T>('a'));
+        }
+    }
+
+    return Result;
+}
+
+// Trims leading and trailing ASCII whitespace (space, tab, CR, LF, vertical tab, form feed), matching
+// the ASCII subset of Go's strings.TrimSpace. Returns a view into the input, so the input must outlive
+// the result. The stdlib has no trim, so this centralizes the find_first/last_not_of idiom.
+template <class T>
+inline std::basic_string_view<T> TrimAscii(std::basic_string_view<T> String)
+{
+    constexpr T Whitespace[] = {
+        static_cast<T>(' '), static_cast<T>('\t'), static_cast<T>('\r'), static_cast<T>('\n'), static_cast<T>('\v'), static_cast<T>('\f'), static_cast<T>('\0')};
+
+    const auto First = String.find_first_not_of(Whitespace);
+    if (First == std::basic_string_view<T>::npos)
+    {
+        return {};
+    }
+
+    const auto Last = String.find_last_not_of(Whitespace);
+    return String.substr(First, Last - First + 1);
+}
+
 template <class T, class TInput>
 inline std::basic_string<T> Join(const std::vector<TInput>& Input, T Separator)
 {
--- a/src/shared/inc/stringshared.h
+++ b/src/shared/inc/stringshared.h
@@ -69,7 +69,7 @@ inline bool EndsWith(const std::basic_string<T>& String, const std::basic_string
 // locale-independent (no Turkish-'I' surprises) and has no signed-char UB, which is what you want when
 // normalizing ASCII protocol tokens such as buildx CSV keys.
 template <class T>
-inline std::basic_string<T> AsciiToLower(std::basic_string_view<T> String)
+inline std::basic_string<T> AsciiToLower(const std::basic_string_view<T> &String)
 {
     std::basic_string<T> Result(String);
     for (auto& Ch : Result)
@@ -87,7 +87,7 @@ inline std::basic_string<T> AsciiToLower(std::basic_string_view<T> String)
 // the ASCII subset of Go's strings.TrimSpace. Returns a view into the input, so the input must outlive
 // the result. The stdlib has no trim, so this centralizes the find_first/last_not_of idiom.
 template <class T>
-inline std::basic_string_view<T> TrimAscii(std::basic_string_view<T> String)
+inline std::basic_string_view<T> TrimAscii(const std::basic_string_view<T> &String)
 {
     constexpr T Whitespace[] = {
         static_cast<T>(' '), static_cast<T>('\t'), static_cast<T>('\r'), static_cast<T>('\n'), static_cast<T>('\v'), static_cast<T>('\f'), static_cast<T>('\0')};
--- a/src/windows/service/inc/wslc.idl
+++ b/src/windows/service/inc/wslc.idl
@@ -549,8 +549,8 @@ typedef struct _WSLCBuildImageOptions
     WSLCBuildImageFlags Flags; // WSLCBuildImageFlags
     WSLCStringArray Labels;    // KEY=VALUE pairs passed as --label to docker.
     WSLCBuildSecretArray Secrets; // --secret entries; the server writes each secret's bytes to a host file exposed to the VM read-only over virtiofs and emits the corresponding id=...,src=... spec.
-    [unique, string] LPCSTR Output; // buildx exporter spec passed as --output to docker build (e.g. type=local, type=tar). The client omits dest= for exporters with a client destination; the server rewrites it to a VM temp path (streamed back over OutputHandle) or into the OutputMountPath mount (single-file exporters with a real destination).
-    WSLCHandle OutputHandle; // When Type != WSLCHandleTypeUnknown, the server streams the build exporter output to this handle after a successful build: a raw stream for dest=- (stdout), or a tar stream of the export directory for type=local (indicated by WSLCBuildImageFlagsOutputIsDirectory).
+    [unique, string] LPCSTR Output; // buildx exporter spec passed as --output to docker build (e.g. type=tar, type=oci). The client omits dest= for exporters with a client destination; the server rewrites it to a VM temp path (streamed back over OutputHandle) or into the OutputMountPath mount (single-file exporters with a real destination). Directory exporters (type=local, or oci/docker with tar=false) are not supported.
+    WSLCHandle OutputHandle; // When Type != WSLCHandleTypeUnknown, the server streams the single-file exporter output (a tar/oci/docker tarball, or dest=- stdout) to this handle after a successful build.
     [unique, string] LPCWSTR OutputMountPath; // When set, the server mounts this Windows directory read-write into the VM and points the exporter's dest at it (plus OutputMountFile). Used for single-file exporters with a real destination; mutually exclusive with OutputHandle.
     [unique, string] LPCWSTR OutputMountFile; // Leaf filename within OutputMountPath that a single-file exporter writes to.
 } WSLCBuildImageOptions;
--- a/src/windows/service/inc/wslc.idl
+++ b/src/windows/service/inc/wslc.idl
@@ -549,8 +549,10 @@ typedef struct _WSLCBuildImageOptions
     WSLCBuildImageFlags Flags; // WSLCBuildImageFlags
     WSLCStringArray Labels;    // KEY=VALUE pairs passed as --label to docker.
     WSLCBuildSecretArray Secrets; // --secret entries; the server writes each secret's bytes to a host file exposed to the VM read-only over virtiofs and emits the corresponding id=...,src=... spec.
-    [unique, string] LPCSTR Output; // buildx exporter spec passed as --output to docker build (e.g. type=local, type=tar). When OutputHandle is set the client omits dest= and the server rewrites it to a VM temp path, then streams the result back over OutputHandle.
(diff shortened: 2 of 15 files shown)


dkbennett (commented):
This factoring is fine per additional review.

src/windows/service/inc/wslc.idl
```diff
@@ -526,6 +542,9 @@ typedef struct _WSLCBuildImageOptions
     LPCSTR Target;             // Target build stage name passed as --target to docker.
     WSLCBuildImageFlags Flags; // WSLCBuildImageFlags
     WSLCStringArray Labels;    // KEY=VALUE pairs passed as --label to docker.
+    WSLCBuildSecretArray Secrets; // --secret entries; the server writes each secret's bytes to a root-only tmpfs file in the VM and emits the corresponding id=...,src=... spec.
+    [unique, string] LPCSTR Output; // buildx exporter spec passed as --output to docker build (e.g. type=local, type=tar). When OutputHandle is set the client omits dest= and the server rewrites it to a VM temp path, then streams the result back over OutputHandle.
```
dkbennett: These inline comments are exceptionally long, should break them at at least 120 chars.

src/shared/inc/stringshared.h
```diff
@@ -65,6 +65,43 @@ inline bool EndsWith(const std::basic_string<T>& String, const std::basic_string
     return std::equal(Suffix.rbegin(), Suffix.rend(), String.rbegin());
 }
 
+// Lowercases ASCII 'A'-'Z' only, leaving every other code unit untouched. Unlike std::tolower this is
+// locale-independent (no Turkish-'I' surprises) and has no signed-char UB, which is what you want when
+// normalizing ASCII protocol tokens such as buildx CSV keys.
+template <class T>
+inline std::basic_string<T> AsciiToLower(std::basic_string_view<T> String)
```
OneBlue: nit: `String` should be a const reference here to avoid a copy. Same for below

src/windows/wslc/services/ImageService.cpp
```diff
@@ -191,6 +194,149 @@ void ImageService::Build(
...
+                // Refuse to dump the binary exporter stream onto an interactive console (matching docker);
+                // stdout must be redirected to a file or pipe. Otherwise ToCOMInputHandle would fail the
+                // marshal with a cryptic ERROR_NOT_SUPPORTED for the console handle. A redirected character
+                // device such as NUL is not a console, so IsConsoleHandle (which also checks
+                // GetConsoleMode) still lets those through.
+                THROW_HR_WITH_USER_ERROR_IF(
+                    E_INVALIDARG,
+                    Localization::MessageWslcOutputInvalidSpec(
+                        validation::FormatOutputSpec(spec),
+                        L"refusing to write build output to the console; redirect stdout to a file or pipe (for example "
```
OneBlue: This error message should be localized. Also I don't recommend suggesting a `>` redirection, because we don't know for sure that the caller is using a shell

src/windows/wslc/services/ImageService.cpp
```diff
@@ -191,6 +194,149 @@ void ImageService::Build(
...
+                        L"'> out.tar') or pass 'dest=' with a file path"),
+                    IsConsoleHandle(outputHandle));
+            }
+            else
+            {
+                // Create the destination's parent directory (buildx creates missing parents) and stream
+                // into a sibling temp file that is renamed over the destination only on success (below).
+                auto destPath = std::filesystem::absolute(spec.Dest);
```
OneBlue, line 242: Instead of having to write to a file and then extract, could we give use a pipe and pass it directly to tar to simplify this logic ? We can always delete the directory on failure

src/windows/wslc/services/ImageService.cpp
```diff
@@ -191,6 +194,149 @@ void ImageService::Build(
...
+            outputStr = wsl::windows::common::string::WideToMultiByte(validation::FormatOutputSpec(vmSpec));
+        }
+        else if (isDirExporter)
+        {
+            std::error_code dirError;
+            std::filesystem::create_directories(std::filesystem::absolute(spec.Dest), dirError);
+            THROW_HR_IF_MSG(HRESULT_FROM_WIN32(dirError.value()), !!dirError, "Failed to create directory: %ls", spec.Dest.c_str());
+
+            // Strip a trailing separator so the CRT does not parse '\"' as an escaped quote.
+            auto destDir = std::filesystem::absolute(spec.Dest).wstring();
```
OneBlue: Instead of doing this, would a `weakly_canonical` path cover both cases ?

src/windows/wslc/services/ImageService.cpp
```diff
@@ -191,6 +194,149 @@ void ImageService::Build(
...
+
+            // The server picks the VM-side dest, so forward the spec without the client's dest.
+            BuildOutput vmSpec = spec;
+            vmSpec.Dest.clear();
+            outputStr = wsl::windows::common::string::WideToMultiByte(validation::FormatOutputSpec(vmSpec));
+        }
+        else if (isDirExporter)
+        {
+            std::error_code dirError;
+            std::filesystem::create_directories(std::filesystem::absolute(spec.Dest), dirError);
```
OneBlue: I think it would be OK to use the overload that directly throws an exception here (so we don't have to handle it manually)

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1340,6 +1396,52 @@ try
...
+    // the temp file; for directory exporters (type=local, or oci/docker with tar=false) tar the export
+    // directory so the client can extract it into the destination directory. The streamer's stdout is
+    // relayed to the client OutputHandle.
+    if (streamOutput)
+    {
+        auto userHandle = OpenUserHandle(Options->OutputHandle);
+
+        std::vector<std::string> streamArgs;
+        if (outputIsDirectory)
+        {
```
OneBlue: Instead of doing this, could we mount the target folder via a virtiofs mount, and have `docker buildx` directly write in the right place ?

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1340,6 +1396,52 @@ try
...
+            streamArgs = {"/usr/bin/tar", "-C", outputDestPath, "-cf", "-", "."};
+        }
+        else
+        {
+            streamArgs = {"/bin/cat", outputDestPath};
```
OneBlue: Is there a way to stream this directly as part of the docker buildx process so we don't have to do this in a second step ? We could potentially start that process with an additional fd (pointing to a docker) and then telling it to write to let's say `/proc/self/fd/X`. 

Ideally it would be nice to completely skip the temp file logic

src/windows/wslc/services/ImageService.cpp
ggarzia-MSFT: This code is for when we are outputting as a tarfile. We write to a temp file and then copy it to the output file so that it remains a single operation that will not overwrite an existing file if interrupted. 
If the output is expected as a directory structure then we do stream it directly to tar -xf as shown by the following code near line 315 in ImageService.cpp.
```
auto tarCmd = std::format(L"tar.exe -xf - -C \"{}\"", destDir);
wsl::windows::common::SubProcess tarProcess(nullptr, tarCmd.c_str());
tarProcess.SetStdHandles(pipeRead.get(), nullptr, nullptr);
extractProcess = tarProcess.Start();
```

ggarzia-MSFT: Let me know if I am misunderstanding your comment

src/windows/wslc/services/ImageService.cpp
ggarzia-MSFT: weakly_canonical would resolve C: to C:\, which would result in one of the quotes being escaped in the tar command. We can use weakly_canonical to replace the while loop, but we would still need to check to see if destDir is a root drive, and then add a dot at the end similar to what we are currently doing

src/windows/wslc/services/ImageService.cpp
OneBlue: Oh Ok, I misread this and thought we were extracting it later. I do think that it would be nicer to directly write to the destination here. Having logic to write to a temporary file first makes things a bit more complex, and I think a user would expect an existing file to be overwritten either way in a situation like that one

src/windows/wslcsession/WSLCSession.cpp
ggarzia-MSFT: This addresses your comments above as well

src/windows/wslcsession/WSLCSession.cpp
ggarzia-MSFT: Could I use the same logic as the previous comment and mount the parent directory of the output dest and then write the tarfile there? The only disadvantage I see is possibly overwriting an existing tar file with a truncated file if it fails halfway through the write.

src/windows/wslc/services/ImageService.cpp
ggarzia-MSFT: This is no longer relevant after making this change: https://github.com/microsoft/WSL/pull/41157#discussion_r3686411156

test/windows/wslc/WSLCCLIOutputParserUnitTests.cpp
```diff
@@ -0,0 +1,532 @@
...
+            VERIFY_IS_TRUE(message.find(expectedReasonSubstr) != std::wstring::npos);
+        }
+    }
+
+    // --- Valid: shorthand (single token, no key=value pairs) ---
+
+    TEST_METHOD(Output_Shorthand_LocalDirectory)
+    {
+        // A bare path is shorthand for exporting the final stage's filesystem to that directory.
+        VerifyValid(L"./out", L"local", L"./out");
```
dkbennett: These are a lot of individual tests for one validation, perhaps combine most of these into a small set of related tests?

dkbennett (commented):
Looks good overall, few small items

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -45,6 +45,75 @@ constexpr DWORD c_processKillTimeoutMs = 10 * 1000;
...
+            const auto value = field.substr(eq + 1);
+            if (key == "type")
+            {
+                type = toLower(value);
+            }
+            else if (key == "tar")
+            {
+                // Go's ParseBool: 1/t/true and 0/f/false (case-insensitive).
+                const auto boolValue = toLower(value);
+                if (boolValue == "0" || boolValue == "f" || boolValue == "false")
```
OneBlue: nit: I recommend using `wsl::shared::string`::ParseBool` for this

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -45,6 +45,75 @@ constexpr DWORD c_processKillTimeoutMs = 10 * 1000;
...
+        }
+        pos = comma + 1;
+    }
+
+    if (type == "local")
+    {
+        return true;
+    }
+
+    return (type == "oci" || type == "docker") && tar.has_value() && !tar.value();
```
OneBlue: nit: This could be simplified to `(type == "oci" || type == "docker") && !tar.value_or(true);

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -45,6 +45,75 @@ constexpr DWORD c_processKillTimeoutMs = 10 * 1000;
 
 namespace {
 
+// Minimal parse of a docker buildx --output spec to detect directory exporters (type=local, or
```
OneBlue: Could we factor the parser logic between the server & client side ? It feels a bit odd to have two parsers for the same spec

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -953,6 +1023,85 @@ try
...
+    RETURN_HR_IF(E_INVALIDARG, (streamOutput || mountOutput) && (Options->Output == nullptr || Options->Output[0] == '\0'));
+    RETURN_HR_IF(E_INVALIDARG, streamOutput && mountOutput);
+    // Directory exporters cannot be materialized faithfully on a Windows destination and are unsupported.
+    // The client rejects them while parsing --output; reject them here too for direct COM callers.
+    RETURN_HR_IF(E_INVALIDARG, Options->Output != nullptr && IsDirectoryExporterSpec(Options->Output));
+    std::string outputDestPath;
+    auto removeOutput = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
+        if (!outputDestPath.empty())
+        {
+            ServiceProcessLauncher cleanup("/bin/sh", {"/bin/sh", "--norc", "-c", std::format("rm -rf '{}'", outputDestPath)}, {}, WSLCProcessFlagsNone);
```
OneBlue: In recommend appending a `--` before outputDestPath there to be safe

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -953,6 +1023,85 @@ try
...
+    });
+
+    if (Options->Output != nullptr && Options->Output[0] != '\0')
+    {
+        std::string outputSpec = Options->Output;
+        if (streamOutput)
+        {
+            GUID id{};
+            THROW_IF_FAILED(CoCreateGuid(&id));
+            constexpr auto c_outputBaseDir = "/var/lib/docker/tmp/wslc-build-output";
```
OneBlue: Could we make the docker process write the output on stdout directly instead of going through a file ? That would have the benefit of completely getting rid of the folder creation, cleanup, and later streaming. 

It will also reduce VM disk usage since the image would go directly through hvsockets

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1412,9 +1315,21 @@ try
     };
 
     // With --progress=rawjson, docker writes progress to stderr and the final image ID to stdout on success (empty on
-    // failure). Stdout is drained into allOutput (shown only on error) and its EOF signals build completion.
-    io.AddHandle(std::make_unique<io::ReadHandle>(
-        buildProcess.GetStdHandle(1), [&](const auto& content) { allOutput.append(content.begin(), content.end()); }));
+    // failure). Stdout's EOF signals build completion.
```
OneBlue: nit: I'd remove ` Stdout's EOF signals build completion.` since we need to wait both for the process to exit, and EOF on this handle to continue, so this comment is a bit misleading

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1412,9 +1315,21 @@ try
...
+    //
+    // For dest=- the exporter tarball is written to stdout instead, so it is relayed straight to the
+    // client handle as the build runs rather than being collected. RelayHandle is an overlapped handle,
```
OneBlue: `rather than being collected` refers to a previous iteration of this change and so won't be relevant once merged. I recommend removing this part

OneBlue (approved):

PR #41160, opened 2026-07-24
CLI: Align WSLC argument parser with Docker flag and value semantics

dkbennett:
**Purpose: a valid Docker command line, mapped to the equivalent wslc command, should produce a valid wslc command.** This PR brings the wslc CLI argument parser in line with Docker's command-line semantics so that users coming from Docker are not surprised by parse failures or different results.

Boolean flags now accept Docker-style adjoined values: `--flag`, `--flag=true`, `--flag=1`, `--flag=t`, and their false counterparts `--flag=false`, `--flag=0`, `--flag=f` (all case-insensitive), on both long names and alias forms including chains (`-ab=false`). A space-separated token after a flag is never consumed as the flag's value (`--flag value` leaves `value` as a positional), matching Docker. Adjoined values may be double-quoted (`--flag="true"`), consistent with value arguments. Repeated flags and single-value arguments are last-wins, and unlimited value arguments accumulate, all matching Docker.

Flags now store their explicit parsed value (true or false) rather than being encoded purely by presence. A new `ArgMap::GetFlag<ArgType::X>(defaultValue = false)` shorthand folds the presence check and the stored value into a single call, and all flag on/off tests in the CLI now use it. This closes a forward-looking Docker-parity gap: a flag whose behavior is on by default can be represented and disabled with `--flag=false` via `GetFlag<...>(true)`, which pure presence-encoding could not express. `Contains()` is unchanged and still used for "was it specified" checks (value arguments and mutual-exclusion validation).

Supporting changes: `ParseBool` gains an opt-in `AllowExtendedForms` parameter (default preserves today's behavior for every other caller; the CLI opts in to also accept `t`/`f`); the count limit is modeled as `enum class Limit { Single, Unlimited }` instead of an int-plus-sentinel; and a latent bug where unlimited positionals never rendered the help "..." indicator is fixed.

## PR Checklist

**Docker-style boolean flag values (`ArgumentParser.cpp`/`.h`)**
- `ApplyFlagValue` strips one pair of surrounding double quotes, then parses the token as a Docker-style boolean. Invalid tokens (for example `--flag=maybe`) raise an argument error instead of being silently ignored.
- `StripSurroundingQuotes` is shared by the flag and value adjoined-value paths so both treat `="value"` identically.
- The alias, alias-chain, and named-flag paths all route through `ApplyFlagValue`, so `-q=false`, `-qi=false`, and `--flag=false` behave the same.

**Flag value storage + `GetFlag` shorthand (`ArgumentTypes.h`, `ArgumentParser.cpp`)**
- `SetFlag(type, value)` clears any prior entry (including an env-preloaded overridable default), then stores the explicit value. Clearing first keeps a flag at a single entry and gives last-wins for repeats (`--flag --flag=false` ends up false).
- `ArgMap::GetFlag<E>(defaultValue = false)` returns the stored value when present, or `defaultValue` when absent, with a `static_assert` that `E` is a `Kind::Flag`. Pass `true` for a default-on flag.

**Consumer migration (43 sites)**
- Every behavioral flag on/off test moved from `Contains(ArgType::Flag)` to `GetFlag<ArgType::Flag>()`.
- Value/positional `Contains()` checks are untouched.

**`ParseBool` extended forms (`src/shared/inc/stringshared.h`)**
- New `AllowExtendedForms` parameter, defaulting to false so `configfile`, `CommandLine.h`, and `OptionParser` are unaffected. When true, `t`/`f` (case-insensitive) are accepted in addition to `true`/`false`/`1`/`0`. The CLI flag path opts in.

**Count limit model + latent fix**
- `enum class Limit { Single, Unlimited }` replaces the int count plus `NO_LIMIT` sentinel across `Argument`, the parser, and all commands. Single value args are last-wins; unlimited args accumulate.
- `Command.cpp` help rendering used `Limit() > 1` (always false under the old model), so unlimited positionals never showed "..."; it now uses `IsUnlimited()`.

## Validation Steps Performed
- `WSLCCLIParserUnitTests.cpp`: value-aware assertions via `GetFlag`; `Flag_FalseForms_StoreSingleFalseEntry` (present with false); `Flag_FalseOverridesPreloadedDefault`; and `Flag_GetFlagDefaultTrue_DefaultOnFlag` demonstrating the default-on pattern. `ParserTestCases.h` covers true/false/1/0/t/f, quoted, alias, and reject cases. `SimpleTests.cpp` covers `ParseBool` default vs extended modes.


src/windows/wslc/arguments/ArgumentTypes.h
```diff
@@ -37,6 +37,18 @@ enum class Kind
     Forward,
 };
 
+// How many times an argument may be supplied on a command line.
+enum class Limit
+{
+    // Accepts a single value. Repeats are last-wins (docker-style): a later occurrence
+    // overwrites the earlier one rather than being an error. For flags, "present" always
+    // means true, so a later "--flag=false" clears it.
```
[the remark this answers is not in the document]
dkbennett: Nice catch, Fixed

ggarzia-MSFT (approved):
lgtm

test/windows/wslc/WSLCCLIParserUnitTests.cpp
```diff
@@ -537,9 +532,9 @@ class WSLCCLIParserUnitTests
...
     }
 
-    // Duplicate value on the CLI (no override) still stacks so Validate can
-    // catch the Limit violation.
-    TEST_METHOD(DuplicateValueOnCli_StillStacks)
+    // Duplicate single-value arg on the CLI (no preload) is last-wins (docker-style):
+    // the final value replaces the earlier one instead of accumulating.
+    TEST_METHOD(DuplicateValueOnCli_LastWins)
     {
         auto inv = WSLCTestHelpers::CreateInvocationFromCommandLine(L"wslc --signal 9 --signal 1");
```
OneBlue, line 539: nit: If an override happens, I think outputting a warning would be a good idea

OneBlue (approved):
LGTM, minor suggestion

test/windows/wslc/WSLCCLIParserUnitTests.cpp
dkbennett: I agree that it may indicate a user mistake and a warning makes sense, however Docker does not do this and overwrites are silent.  The reason for this is partially that it is the POSIX/GNU convention and Docker inherited it from dependent packages. The functional reason is that the idea is defaults layer and later things can override. This behavior would also allow us to have a feature for configurable defaults that are later overridden by the command line, but some users may do this today via scripting.

Given that this behavior has been in docker for a decade and some consumers may rely on it, I am inclined to not put in a warning on overrides as that would violate the convention and might pose problems for users and scripts and future features that may rely on this override behavior.

I think it is worth discussing this behavior and whether we want to intentionally diverge, but that can be a follow-up PR.

PR #41164, opened 2026-07-24
Bound relay drains so a wedged guest socket cannot hang wsl.exe

yeelam-gordon:
## Why (the symptom)

`wsl.exe` install can **hang forever and never exit** (the same teardown path is also used by import / export / resize). Seen intermittently in automated test runs on heavily loaded machines. Tracked as **62543726** (duplicates 63009527, 63219697).

## Root cause

At teardown, `ScopedRelay::Sync()` does an **unconditional** `m_thread.join()`. That relay thread only exits when the Linux guest closes its stdout/stderr hvsocket. If the guest never closes it, the join — and the whole process — blocks forever.

## The fix

Add a **bounded** `Sync(timeout)` overload:

- Wait for the stream's natural EOF up to `timeout` → a slow-but-alive stream is still drained in full (**no truncation**).
- If the deadline passes, signal the relay's exit event and join → teardown is **bounded**, not infinite.

Safe because the relay verdict is captured *before* the drain, so bounding it cannot change the reported result. Exec stdout/stderr workers become `std::optional<ScopedRelay>` to share the same path. Default `c_relayDrainTimeout = 60s`.

Why bound at teardown and not per-read: a per-read timeout would truncate a slow-but-alive stream; a teardown bound only trips when the stream truly never ends.

## Testing

New `ScopedRelay` unit tests:
- wedged pipe → `Sync(timeout)` returns bounded (no hang).
- stream that finishes before the deadline → **not truncated**.

Plus a full MSI deployed to a live WSL host: install and exec (20k / 200k lines) complete cleanly, no hang.

---

@shuaiyuanxx — could you review? You own the relay / centralized-IO path. (Bug 62543726 assigned to you.)

### DIFF
--- a/src/windows/common/relay.cpp
+++ b/src/windows/common/relay.cpp
@@ -445,6 +445,27 @@ void ScopedRelay::Sync()
     }
 }
 
+void ScopedRelay::Sync(std::chrono::milliseconds Timeout)
+{
+    // Drain to natural EOF within the timeout; otherwise cancel before joining.
+    if (!m_thread.joinable())
+    {
+        return;
+    }
+
+    // Keep the wait bounded and below INFINITE.
+    const DWORD timeoutMs = Timeout.count() <= 0
+                                ? 0
+                                : static_cast<DWORD>(std::min<long long>(Timeout.count(), static_cast<long long>(INFINITE) - 1));
+
+    if (!m_completed.wait(timeoutMs))
+    {
+        m_exitEvent.SetEvent();
+    }
+
+    m_thread.join();
+}
+
 ScopedRelay::~ScopedRelay()
 {
     try
--- a/src/windows/common/relay.cpp
+++ b/src/windows/common/relay.cpp
@@ -454,9 +454,8 @@ void ScopedRelay::Sync(std::chrono::milliseconds Timeout)
     }
 
     // Keep the wait bounded and below INFINITE.
-    const DWORD timeoutMs = Timeout.count() <= 0
-                                ? 0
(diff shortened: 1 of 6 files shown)


OneBlue (changes requested):
Do we know why things get stuck in the guest in the first place ? If there's a socket that stays opened, that's what we should resolve

yeelam-gordon:
Sorry for the late follow-up. Traced this through the ETL evidence from bug 62543002.

### Why the guest socket stays open

Same failure class as the boot timeout (PR #41165). The host-side ETL shows the throw at `relay.cpp:673` after 30.44s — the host is waiting on guest exit, and the guest exceeds the 30s `KernelBootTimeout`. No host-side socket leak.

We can't yet prove whether the guest is "slow" or "stuck" — boot-stage telemetry is compiled out of release builds, so guest dmesg isn't captured before the timeout kills the boot. The instrumentation in PR #41165 + #41092 makes those events release-visible to answer this definitively.

### Why a bounded drain

This is the safety net alongside #41165 (root-cause instrumentation + timeout increase). Today, if the guest is slow to exit for any reason — the known ARM64 contention case or a future unknown path — `wsl.exe` hangs forever on `m_thread.join()`. The bounded `Sync(timeout)` caps that so `wsl.exe` degrades gracefully instead of hanging indefinitely.

OneBlue:
> We can't yet prove whether the guest is "slow" or "stuck" — boot-stage telemetry is compiled out of release builds, so guest dmesg isn't captured before the timeout kills the boot. The instrumentation in PR #41165 + #41092 makes those events release-visible to answer this definitively.

I'm a bit confused about this. What exactly is being compiled out of release builds ? Telemetry is only enabled on official release builds so I'm not following this

Issue #40783, opened 2026-06-11
WSL2 kernel 6.18.33.1 (WSL 2.7.8.0) fails to boot on AMD FX-8370 (Piledriver) — CreateVm 0x800705b4 timeout; kernel 6.6.87.2 (WSL 2.6.3.0) boots fine on same machine


mturcich24:
### Windows Version

Microsoft Windows [Version 10.0.19045.7417]

### WSL Version

2.7.8.0 (fails) — does not reproduce on 2.6.3.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.18.33.1-1 (fails to boot) 6.6.87.2-1 (boots successfully on identical hardware/OS)

### Distro Version

Debian (also reproduced with Docker Desktop's docker-desktop distro import)

### Other Software

Docker Desktop 4.77.0 (how the issue was first observed — its `wsl --import-in-place docker-desktop` call fails)
CPU: AMD FX-8370 (Piledriver, 2014 — supports AVX but NOT AVX2)
Hypervisor confirmed running (HypervisorPresent = True), virtualization enabled in firmware, Memory Integrity/HVCI not enabled (SecurityServicesRunning = {0})

### Repro Steps

1. Windows 10 22H2 (19045.7417) machine with AMD FX-8370 CPU, WSL 2.7.8.0 MSI installed (kernel 6.18.33.1-1).
2. Attempt to register/create any WSL2 distro:

   wsl --install -d Debian --no-launch
   wsl -d Debian echo test

   (or launch Docker Desktop, which runs `wsl --import-in-place docker-desktop ...`)

3. Operation hangs ~60s, then fails:

   This operation returned because the timeout period expired.
   Error code: Wsl/Service/RegisterDistro/CreateVm/0x800705b4

4. No error events in Microsoft-Windows-Hyper-V-Compute-Admin — only normal HCS start events.
5. With debugConsole=true in .wslconfig, the kernel boot log shows boot proceeding normally, then the VM connection closes immediately after "Mountpoint-cache hash table entries" (full log attached). panic=-1 on the kernel command line appears to kill the VM before any panic trace is flushed.
6. Downgrade test: uninstall the 2.7.8.0 MSI, install the 2.6.3.0 MSI (kernel 6.6.87.2-1) from GitHub releases. Same machine, same Windows build, no other changes.
7. Repeat step 2 — distro installs and runs successfully. Docker Desktop also imports and runs normally.

[...]

PR #41165, opened 2026-07-24
Mitigate and instrument the utility-VM boot-connect timeout

yeelam-gordon:
## Why (the symptom)

On heavily loaded machines, utility-VM creation intermittently fails with `.../CreateVm/0x800705b4` (ERROR_TIMEOUT), which blocks install. Seen in automated test runs. Tracked as **62543002**.

## Root cause

During VM `Initialize`, the service waits for the guest to connect back over hvsocket within `KernelBootTimeout` (default **30s**). This is a give-up **watchdog**, not a boot budget — a healthy boot connects in ~1-3s. Under host CPU contention a healthy-but-slow boot exceeds 30s and trips it. Two gaps:

1. 30s leaves too little headroom on a contended host.
2. The notify-channel accept was **uninstrumented**, so telemetry could not say which boot phase was slow.

## The fix (two small changes)

1. Raise default `KernelBootTimeout` **30s -> 90s**. Still bounds a genuinely stuck VM; just gives a slow-but-alive boot room to finish. Overridable via `wsl2.kernelBootTimeout`.
2. Wrap the notify-channel accept in its own `SlowOperationWatcher{"WaitForNotifyChannelConnect"}` (mirrors the adjacent mini_init one). **Diagnostic only — no control-flow change.**

Note: this is **not** fixed by #40519. That protects processes *inside* the guest *after* it connects; this failure is the *host* starving the VM's vCPUs *before* it connects — different phase, different contention. The real cure (a host-side CPU reservation for the booting VM) is a follow-up, gated on the telemetry above confirming CPU is the starved resource.

## Testing

- `SlowOperationWatcher` unit tests: 6/6 pass (emits real `elapsedMs`; hang backstop emits once).
- `wslservice` build-verified with both changes.
- Deterministic repro on the same accept path: setting `kernelBootTimeout=1` yields the exact `.../CreateVm/0x800705b4`.

---

@shuaiyuanxx — could you review? You own the SlowOperationWatcher / boot path. (Bug 62543002 assigned to you.)

### DIFF
--- a/src/windows/common/WslCoreConfig.h
+++ b/src/windows/common/WslCoreConfig.h
@@ -340,7 +340,7 @@ struct Config
     int InstanceIdleTimeout = (15 * 1000);
     std::filesystem::path DebugConsoleLogFile;
     std::wstring VmSwitch;
-    int KernelBootTimeout = (30 * 1000);
+    int KernelBootTimeout = (90 * 1000);
     int DistributionStartTimeout = (60 * 1000);
     int MountDeviceTimeout = (5 * 1000);
     bool EnableHostFileSystemAccess = true;
--- a/src/windows/service/exe/WslCoreVm.cpp
+++ b/src/windows/service/exe/WslCoreVm.cpp
@@ -427,7 +427,11 @@ void WslCoreVm::Initialize(const GUID& VmId, const wil::shared_handle& UserToken
     }
 
     // Accept the connection from the Linux guest for notifications.
-    m_notifyChannel = AcceptConnection(m_vmConfig.KernelBootTimeout);
+    // Instrumented so slow/hung boots are attributed to this phase vs WaitForMiniInitConnect.
+    {
+        SlowOperationWatcher slowOperation{"WaitForNotifyChannelConnect"};
+        m_notifyChannel = AcceptConnection(m_vmConfig.KernelBootTimeout);
+    }
 
     // Receive and parse the guest kernel version
     {


OneBlue (commented):
I'm surprised that we find machines that take more than 30 seconds to connect the hvsocket. 

The bug mentioned in the description points to a test failure, but in the test we increase that timeout to 120 : https://github.com/microsoft/WSL/blob/404453c940459b25b06b672060cf80fa0bdec204/test/windows/Common.cpp#L2096-L2097 

(Default is [here](https://github.com/microsoft/WSL/blob/404453c940459b25b06b672060cf80fa0bdec204/test/windows/Common.cpp#L1540))

So if we timed out in the tests, I'd expect that the issue is somewhere else. Was there anything in dmesg in that test failure ?

yeelam-gordon:
Sorry for the late follow-up. I went back to the ETL from a Jul 9 repro.

### What the ETL shows

The throw is at `relay.cpp:673` after **30.44s** — the 30s `KernelBootTimeout` fired, not the 120s test timeout. The boot was aborted at 30s; it exceeded the limit on contended ARM64 SP11 hardware.

Key facts (WSL 2.7.10, post-kernel-fix):
- Still reproducing — **not** the kernel 6.18.33.1 regression (#40783)
- ~26 failures/week across ~12 branches (ADO heat map) — intermittent, load-correlated
- Retail Surface Pro 11 — same hardware real users have

### On dmesg

Guest dmesg was **not captured** — the boot is killed at 30s before we can pull guest logs. This is the gap: with the current release build, we can't distinguish "slow but healthy" from "truly hung" because boot-stage markers (`SlowOperationWatcher`) use `WSL_LOG_DEBUG` (compiled out in release).

That's exactly what this PR's instrumentation addresses — making boot-stage telemetry release-visible so we can answer "slow vs. hung" definitively. PR #41092 promotes the `SlowOperationWatcher` events; this PR raises the timeout and wires them into the boot path.

PR #41164 adds the safety net on the teardown side.

OneBlue:
> The throw is at `relay.cpp:673` after **30.44s** — the 30s `KernelBootTimeout` fired, not the 120s test timeout. The boot was aborted at 30s; it exceeded the limit on contended ARM64 SP11 hardware.

That's definitely not expected. Are you saying that the tests don't correctly override the boot timeout ? If so, that's what we should fix

PR #41170, opened 2026-07-24
Serialize move/resize/set-sparse against in-progress conversions

benhillis:

Fixes a latent concurrency bug: `--manage --move`, `--manage --resize`, and `--manage --set-sparse` were not serialized against an in-progress `--set-version` conversion or `--export`.

## Detailed Description

`MoveDistribution`, `SetSparse`, and `ResizeDistribution` mutate a distribution's VHD but never consulted `m_lockedDistributions`. A conversion (`--set-version`) or export calls `_ConversionBegin`, which releases `m_instanceLock` while the long operation runs yet keeps the distribution registered in `m_lockedDistributions`. Because the three manage operations above only checked `m_runningInstances`/`IsVhdAttached` (neither of which reflects an in-progress conversion), they could start mid-conversion and race on the same VHD — e.g. a corrupt export, or the VHD being moved/resized out from under a running conversion.

### Fix

Add `_EnsureNotLocked(DistroGuid)` (under `m_instanceLock`) to `MoveDistribution`, `SetSparse`, and `ResizeDistribution` so they fail with `E_ILLEGAL_STATE_CHANGE` while a conversion/export/compaction is in progress — matching how the conversion paths already guard each other via `_ConversionBegin` -> `_EnsureNotLocked`.

### Test

Adds `ManageRejectedWhileLocked` (WSL2): imports a distro, starts an `--export` to a pipe that is deliberately not drained so the export blocks mid-stream and deterministically holds the `Exporting` lock, polls `wsl -l -v` until the state is `Exporting`, then asserts `--resize`, `--set-sparse`, and `--move` are all rejected with `E_ILLEGAL_STATE_CHANGE`.

## PR Checklist

- [x] Tests: added `ManageRejectedWhileLocked`
- [ ] Documentation updated: n/a (internal behavior)

### DIFF
--- a/test/windows/UnitTests.cpp
+++ b/test/windows/UnitTests.cpp
@@ -3269,6 +3269,59 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
         }
     }
 
+    // Verifies that VHD-mutating manage operations (--resize, --set-sparse, --move) are rejected while a
+    // long-running conversion/export holds the distribution lock, rather than racing with it on the VHD.
+    WSL2_TEST_METHOD(ManageRejectedWhileLocked)
+    {
+        constexpr auto name = L"manage-locked-test-distro";
+
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(std::format(L"--import {} . \"{}\" --version 2", name, g_testDistroPath)), 0L);
+        auto cleanupName =
+            wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [name]() { LxsstuLaunchWsl(std::format(L"--unregister {}", name)); });
+        WslShutdown();
+
+        // Start an export to a pipe we deliberately don't drain. Once the pipe buffer fills, the export
+        // blocks mid-stream, deterministically holding the distribution in the "Exporting" locked state.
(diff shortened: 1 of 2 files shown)


OneBlue (dismissed):

test/windows/UnitTests.cpp
```diff
@@ -3269,6 +3269,60 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
...
+            readPipe.reset();
+            if (exportThread.joinable())
+            {
+                exportThread.join();
+            }
+        });
+
+        // Wait until the service reports the distribution as Exporting (i.e. the lock is held).
+        bool locked = false;
+        for (int i = 0; i < 100 && !locked; ++i)
```
OneBlue, line 3301: nit: I recommend increasing the total timeout to 2 minutes (currently only 10 seconds) and using `retryshared.h` for this

OneBlue (approved):

test/windows/UnitTests.cpp
benhillis: Good call, I'll do that in a follow-up change since I want to get this in for the open source maintainer to rebase his branch.

PR #41186, opened 2026-07-27
Pin .resw files to CRLF via .gitattributes

benhillis:
The localization service doesn't consistently normalize line endings in the `.resw` files it generates, so some locales land as CRLF blobs while the rest are LF. Those files then show as full rewrites: in #41183, de-DE/ko-KR/zh-TW each showed ~3,570 changed lines for a 5-7 line translation delta.

`validate-line-endings.py` requires CRLF in the working tree but nothing enforced how the files are stored, so it depended on the committer's `core.autocrlf`.

Mark `*.resw` as `text eol=crlf` so they're always stored LF and checked out CRLF, and renormalize the one file that was stored as CRLF (`en-US`, line endings only).

Verified: both loc validators pass, `git status` is clean afterward, and a clone with `core.autocrlf=false` still gets CRLF for all 22 `.resw` files.

Scoped to `*.resw` only — extending to `.c/.cpp/.h/.hpp/.idl` would renormalize 132 more files and conflict with in-flight PRs.


benhillis:
This can probably be undone when the localization pipeline is fixed...

OneBlue (dismissed):

OneBlue (approved):

PR #41189, opened 2026-07-27
CLI: Cache converted CLI arguments with on-demand revalidation, and unify typed value accessors

dkbennett:
Argument validation in the WSLC CLI converts string inputs into typed values (memory sizes, durations, signals, timestamps, labels, and so on). Until now that conversion ran twice for each argument: once while parsing and validating the command line, and again during execution when a task re-parsed the raw string. Beyond the wasted work, the two conversions could drift apart, and some arguments validated their input late, in the middle of execution, so bad input failed after side effects had already started. This PR validates and converts each argument as needed and caches the typed result on `ArgMap`, so execution reuses the cached value instead of re-parsing. These are CLI-only changes: no WSLC API or runtime changes.

Arguments remain mutable until their first effective read through `GetValue` or `GetAllValues`. Mutations before that read invalidate the cached validation state, so values added during execution are validated before use. The first read freezes that argument, and later attempts to add, remove, or invalidate its data throw as programmer errors. Validate-only arguments follow the same path.

The single-source-of-truth model is backed by compile-time type safety. Each argument declares its converted type in one place (the `WSLC_ARGUMENTS` X-macro), and the cache accessors derive their types from that declaration. A read that disagrees with the declared type, or an attempt to cache an argument that has no converted type, is a compile error rather than a runtime surprise.

Singular access is unified behind `GetValue`, including flags. Missing values use the supplied default or the value type's default constructor, and the resolved default is retained for later reads. `Contains()` still checks raw presence without freezing the argument. Raw `Get`/`GetAll` access and execution-path converters are no longer visible to task and command code, so bypassing typed access and validation will not compile.

## PR Checklist

- Validate-and-cache: `Argument::Validate` converts each raw string and stores the typed value on the `ArgMap`. Execution reads the cached value instead of re-parsing.
- Freeze on first read: mutation before the first `GetValue`/`GetAllValues` read invalidates cached validation state. The read validates the final value and makes that argument immutable; later mutation or invalidation throws.
- `ArgMap` moved into its own header. `ArgumentTypes.h` keeps only the argument enums and type mappings that `ArgMap` is built on.
- New `ArgumentConvertedTypes.h` types-only header exposes the converted-type aliases and mappings without exposing the converter functions to task and command code.
- Strong type safety: each argument declares its converted type in the `WSLC_ARGUMENTS` X-macro. Cache accessors derive their types from that declaration, so wrong-type access is a compile error.
- Unified value accessors: `GetValue<E>(default = {})` handles flags and other singular arguments. An absent argument resolves and retains its typed default. `GetAllValues<E>()` handles non-flag arguments that may contain multiple values.
- Raw access is restricted: `Get`/`GetAll` are private to `ArgMap` and exposed only to validation internals. Task, command, and command-specific validation code must use `GetValue`/`GetAllValues`.
- `Label`, `Options`, and `Type` now convert during validation (`Label`/`Options` -> key/value pair, `Type` -> `InspectType`) instead of during execution.
- `ArgumentValidation.h` is no longer included by task or command files, so execution code cannot call validation converters directly.

## Validation Steps Performed
- Added coverage that every converted argument type converts and caches during validation, including on-demand validation with no prior validation pass.
- Added scenarios for invalid first reads, mutation before first read, immutability after a successful read, retained defaults, and repeated values in insertion order.
- Verified that validate-only arguments never populate the converted cache and invalid input throws during validation.
- The affected WSLCCLI argument and parser unit suites pass.


src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -264,7 +250,7 @@ void DeleteImage(CLIExecutionContext& context)
 {
     WI_ASSERT(context.Data.Contains(Data::Session));
     auto& session = context.Data.Get<Data::Session>();
-    const auto& imageIds = context.Args.GetAll<ArgType::ImageId>();
+    const auto& imageIds = context.Args.GetAllValues<ArgType::ImageId>();
```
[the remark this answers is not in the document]
dkbennett: This isn't dangling -- `GetAllValues()` returns a `vector` by value, and binding it to `const auto&` lifetime-extends the temporary, so it was well-defined. But `auto` is equally efficient and clearer for a value return, so I've switched to `auto` for consistency.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -754,30 +743,24 @@ void SetContainerOptionsFromArgs(CLIExecutionContext& context)
...
     {
-        options.Arguments.emplace_back(WideToMultiByte(context.Args.Get<ArgType::Command>()));
+        options.Arguments.emplace_back(WideToMultiByte(context.Args.GetValue<ArgType::Command>()));
     }
 
     if (context.Args.Contains(ArgType::EnvFile))
     {
-        auto const& envFiles = context.Args.GetAll<ArgType::EnvFile>();
+        auto const& envFiles = context.Args.GetAllValues<ArgType::EnvFile>();
         for (const auto& envFile : envFiles)
```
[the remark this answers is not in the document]
dkbennett: Same as above, not a dangling reference but going to change it for consistency anyway.

src/windows/wslc/arguments/ArgumentTypes.h
```diff
@@ -127,6 +143,127 @@ struct ArgMap : wsl::windows::wslc::EnumBasedVariantMap<ArgType, wsl::windows::w
...
         return Contains(E) ? Get<E>() : defaultValue;
     }
+
+    // Validated-value cache. Argument validation converts raw strings into typed values and caches
+    // them here so execution reuses them without re-parsing. The store is type-erased (std::any keyed
+    // by ArgType) to keep this base header free of the domain headers that define the converted types;
+    // access is by a compile-time ArgType whose value type is derived from the argument's ConvertedType
+    // (ArgumentConvertedTypes.h), so a wrong-type access is a compile error. A multimap preserves order
+    // and multiplicity for arguments that allow multiple values.
+    template <ArgType E>
```
ggarzia-MSFT: Consider moving this to its own header

ggarzia-MSFT (dismissed):

src/windows/wslc/arguments/ArgMap.h
```diff
@@ -0,0 +1,233 @@
...
+    // ArgTypes validated against their current raw values. Distinct from m_validated (only converted
+    // arguments populate that), so validate-only arguments are covered too. Cleared per type by
+    // InvalidateValidated on a raw Add/Remove.
+    mutable std::set<ArgType> m_validatedTypes;
+};
+
+// Only raw mutations invalidate; reads are ignored. We touch only ArgMap's mutable validation state,
+// so the base's const-ness is honored. The base subobject is at offset 0 of ArgMap, so recovering
+// the ArgMap pointer from `map` is valid.
+inline void ArgMapInvalidateValidatedCache(const void* map, ArgType type, EnumBasedVariantMapAction action)
```
OneBlue, line 344: nit: In the parent class, the description of this callback says `// A callback function that can be used for logging map actions.`, but with this, we're not logging anymore, we're actually making state changes in this callback

src/windows/wslc/arguments/ArgMap.h
```diff
@@ -0,0 +1,233 @@
...
+    // outlives the raw data. Const: touches only mutable memoized state, not the raw storage.
+    void InvalidateValidated(ArgType type) const
+    {
+        m_validated.erase(type);
+        m_validatedTypes.erase(type);
+    }
+
+    // Records `type` as validated for its current raw values so reads skip re-validation. Const:
+    // updates only mutable memoized state.
+    void MarkValidated(ArgType type) const
```
OneBlue: Not a blocker, but making this method const with mutable fields feels a bit off. 

What's the reason for making this method const ?

src/windows/wslc/arguments/ArgMap.h
```diff
@@ -0,0 +1,233 @@
...
+        if constexpr (std::is_same_v<typename details::ArgConvertedTypeMapping<E>::value_t, details::NoConversion>)
+        {
+            // Validate-only arguments have no converted cache but can still fail validation, so run
+            // it on demand before returning the raw value (covers values added post-validation).
+            EnsureValidated(E);
+            return Get<E>();
+        }
+        else
+        {
+            return GetValidated<E>();
```
OneBlue: So if this was a validated argument, we're going to return a const reference to the "converted" type. This can be a bit dangerous since `ArgMapInvalidateValidatedCache()` can invalid those, but returning a copy doesn't feel great either. 

I guess that's OK as long as the caller is careful to not call Add() / Emplace()

OneBlue: Probably a good idea to fix the const issue I mentioned earlier though, so callers can't call `InvalidateValidated()` on a const reference and potentially break everyone else's refs

OneBlue (commented):
LGTM, minor comments

src/windows/wslc/arguments/ArgMap.h
dkbennett: Good catch, I'll fix that and merge in the secrets changes.

The invalidate scenario is quite rare, and I'm not sure it even happens in real code. Could return a copy to be safe, seems like a  potentially cryptic and difficult to source bug if it ever happens.  I'll think on this one a bit. Realistically we dont have very many args on a given command so some extra copies for safety is likely not an issue vs ensured correctness.

Or perhaps a compromise would be to set a flag whenever GetValue is called indicating that we've taken out a reference to the value and there can be no further mutation of the data, and any calls to Invalidate after that would assert or throw to identify improper usage. Then we could keep the const reference and enforce correct calling pattern. This seems safe as getting the value and then changing it later and using the new value is actual data alteration that would make the command execution unstable - you have two different values for the same data - which is correct? Not allowing it to be mutated after use makes sense to me.

OneBlue: Yeah I think forbidding mutations would be the safest path. I can't think of a reason why a caller would want that anyway

dkbennett: I've updated to keep it a reference return for no copies and add freezing behavior on read to everything, including flags and add default behavior to all argument types, not just flags. 

Contains() can still be used to check for existence without freezing the arg if that is needed, but now any reads using GetValue or GetAllValues will make the argument immutable. Calls that attempt to change an immutable values will throw. Updated tests for this behavior and added a few new ones as well.

ggarzia-MSFT (dismissed):
LGTM

OneBlue (dismissed):

src/windows/wslc/core/CLIExecutionContext.cpp
```diff
@@ -19,7 +19,7 @@ HANDLE CLIExecutionContext::CreateCancelEvent()
 // This method should be idempotent.
 void CLIExecutionContext::ApplyGlobalOptions()
 {
-    if (GlobalArgs.GetFlag<ArgType::NoColor>())
+    if (GlobalArgs.GetValue<ArgType::NoColor>())
     {
         Terminal.SetNoColor(true);
     }
```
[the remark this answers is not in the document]
dkbennett: should be fine for now, but am investigating whether this needs future proofing or hardening for later additions

OneBlue (approved):

src/windows/wslc/core/CLIExecutionContext.cpp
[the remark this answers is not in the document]
dkbennett: The specific issue is that global args have two reads - once very early in Main to set no-color specifically, and then potentially again during argument parsing. This is not presently an issue because no-color is not a CLI argument, it is environment only. But were we to have a global arg that is applied at the same time from an early environment read and then also provided (and possibly overwritten) in the command line, then it would have an illegal method call throw.

This is something I will address in a followup as a small future-proofing bug. I doubt this will ever be a concern because no-color is unique and the only reason it's applied early is due to the need to disable VT color as soon as possible so any errors in CLI processing aren't colorized if it is set. There is a workaround to this - we can copy the global args into a local copy and use that for the early apply with the understanding that the CLI may later overwrite it. That is probably a safe thing to do to prevent any future global args (such as logging or similar) from tripping over this later. But I will only make that change if I have to spin another build due to merge.

PR #41200, opened 2026-07-29
Added outcome telemetry for cold start

shuaiyuanxx:
Adds `CreateInstanceOutcome` telemetry for WSL cold starts to report whether instance creation succeeded, failed, or exceeded a three-minute timeout.

## PR Checklist

Introduces `TimedOperationOutcomeReporter`, which uses a one-shot thread-pool timer and an atomic completion gate to report exactly one outcome:

  - `Success` when cold-start creation completes successfully within three minutes.
  - `Failure` when creation returns an error within three minutes.
  - `Timeout` when creation remains incomplete after three minutes.

  The timeout is telemetry-only and does not cancel or otherwise affect instance creation.

  The new event includes the distro name, WSL version, instance ID, outcome, elapsed time, timeout threshold, and HRESULT. Existing `CreateInstanceBegin` and `CreateInstanceEnd` telemetry remains unchanged.

## Validation Steps Performed
Manually tested.


craigloewen-msft (changes requested):
This operation looks like it's doing some kind of 'Timed output', do we need that?

IMO the root ask here is to "Move us from tracking WSL startup success with 'CreateInstanceBegin' and 'CreateInstanceEnd' events to instead a single source of truth telemetry event".

That new event should follow the exact same pattern / usage of WSLCCreateSession, which this new event doesn't seem to. 

Is it possible to adjust this PR to reflect that?

shuaiyuanxx:
@craigloewen-msft  Thanks for the clarification. I’ve updated the implementation to follow the `WSLCCreateSession` pattern:

- Removed `CreateInstanceBegin` and `CreateInstanceEnd`.
- Removed the timer-based timeout reporting.
- Wrapped instance creation with `wil::ResultFromException`.
- Emit a single `CreateInstanceOutcome` event after the operation completes, containing `CreationTimeMs` and the final Result.
- Propagate failures after emitting the event.

One limitation of this completion-based pattern is that a true hang—such as an HCS operation waiting indefinitely or a synchronous plugin callback that never returns—will not emit `CreateInstanceOutcome`, because execution never reaches the telemetry call. Ctrl+C or terminating the client also does not reliably cancel the service-side operation.

This matches the `WSLCCreateSession` pattern and makes the event the single source of truth for completed startup attempts. Detecting operations that never return would require separate watchdog or correlation-based telemetry. Please let me know if that should be handled independently

craigloewen-msft:
It's good to call out that limitation - but I think that's ok.

Can we keep the old events alive for now (And remove them in a future update) so we can compare and contrast?

Lastly, right now the "WSL_LOG" macro is being used and not the "WSL_LOG_TELEMETRY" macro. It's a bit confusing (Especially since we have other areas like the 'ExecCritical' where we use WSL_LOG) can we fix that as part of this PR too?

OneBlue (changes requested):
Instead of adding a new event, could we add the fields we want into `CreateInstanceEnd` instead ?

shuaiyuanxx:
@OneBlue Done, thanks for your suggestion.

craigloewen-msft (approved):

craigloewen-msft:
Looks good!

OneBlue (approved):

PR #41209, opened 2026-07-30
Fix concurrent console state restoration

### LINKED ISSUE #41201: Concurrent wsl.exe clients can restore CONIN$ mode out of order and leave VT input enabled
## Windows Version

`Microsoft Windows [Version 10.0.22000]` (Windows 11 Pro x64)

## WSL Version

`2.7.11.0` (Microsoft Store package version)

## Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

## Kernel Version

Not relevant to the observed race; it occurs in the Windows-side `wsl.exe` console-state handling. The exact kernel version was not collected while another long-running session was active.

## Distro Version

The default WSL 2 distribution was used only to run POSIX `sh` and `sleep`; no distro-specific behavior is involved.

## Other Software

- OpenAI Codex CLI `0.145.0` exposed the real-world impact in a native Windows TUI.
- The race was then reproduced independently in a disposable native Windows console without Codex.
- Related Codex symptom report: https://github.com/openai/codex/issues/28301

## Repro Steps

Open a fresh, disposable PowerShell window. The following script snapshots the original input mode, starts two overlapping `wsl.exe` clients in the same inherited console, prints the mode timeline, and restores the original mode in `finally` so that the test window is not left broken.

```powershell
Add-Type @'
using System;
using System.ComponentModel;
using System.Runtime.InteropServices;

public static class ConsoleInputMode
{
    private const int STD_INPUT_HANDLE = -10;

    [DllImport("kernel32.dll", SetLastError = true)]
    private static extern IntPtr GetStdHandle(int nStdHandle);

    [DllImport("kernel32.dll", SetLastError = true)]
    private static extern bool GetConsoleMode(IntPtr handle, out uint mode);

    [DllImport("kernel32.dll", SetLastError = true)]
    private static extern bool SetConsoleMode(IntPtr handle, uint mode);

    public static uint Read()
    {
        IntPtr handle = GetStdHandle(STD_INPUT_HANDLE);
        uint mode;
        if (!GetConsoleMode(handle, out mode))
            throw new Win32Exception(Marshal.GetLastWin32Error());
        return mode;
    }

    public static void Write(uint mode)
    {
        IntPtr handle = GetStdHandle(STD_INPUT_HANDLE);
        if (!SetConsoleMode(handle, mode))
            throw new Win32Exception(Marshal.GetLastWin32Error());
    }
}
'@

function Start-SleepingWsl([int] $Seconds) {
    $startInfo = [Diagnostics.ProcessStartInfo]::new()
    $startInfo.FileName = "$env:SystemRoot\System32\wsl.exe"
    $startInfo.Arguments = "-- sh -c `"sleep $Seconds`""
    $startInfo.UseShellExecute = $false
    [Diagnostics.Process]::Start($startInfo)
}

$initial = [ConsoleInputMode]::Read()
$a = $null
$b = $null

try {
    'Initial       : 0x{0:X}' -f $initial

    $a = Start-SleepingWsl 3

    $deadline = (Get-Date).AddSeconds(5)
    do {
        Start-Sleep -Milliseconds 10
        $duringA = [ConsoleInputMode]::Read()
    } while ($duringA -eq $initial -and (Get-Date) -lt $deadline)

    'A active      : 0x{0:X}' -f $duringA
    if ($duringA -eq $initial) {
        throw 'WSL did not change the console input mode before the timeout.'
    }

    $b = Start-SleepingWsl 7
    Start-Sleep -Milliseconds 500

    $a.WaitForExit()
    'After A exits : 0x{0:X} (A={1}, B still running={2})' -f `
        [ConsoleInputMode]::Read(), $a.ExitCode, (-not $b.HasExited)

    $b.WaitForExit()
    $final = [ConsoleInputMode]::Read()
    'After B exits : 0x{0:X} (B={1})' -f $final, $b.ExitCode
}
finally {
    if ($a -and -not $a.HasExited) { $a.WaitForExit() }
    if ($b -and -not $b.HasExited) { $b.WaitForExit() }
    [ConsoleInputMode]::Write($initial)
    'Restored      : 0x{0:X}' -f [ConsoleInputMode]::Read()
}
```

Representative output from the isolated test:

```text
Initial       : 0x1F7
A active      : 0x3D8
After A exits : 0x1F7 (A=0, B still running=True)
After B exits : 0x3D8 (B=0)
Restored      : 0x1F7
```

Control tests:

- A single `wsl.exe` invocation restores the original mode.
- 30 strictly serial `wsl.exe -- true` invocations restored correctly 30/30 times.
- One WSL process overlapping an ordinary PowerShell process restored correctly in either exit order.
- PowerShell-to-WSL execution with inherited or redirected stdout/stderr restored correctly in 15/15 tests.
- Two overlapping WSL clients restore correctly when the client that started first also exits last.
- The failure is deterministic when client A saves the original mode, client B saves A's temporary mode, A exits first, and B exits last.

## Expected Behavior

After all `wsl.exe` clients attached to a console have exited, the console input mode should be the mode that existed before the first client started. One client's temporary WSL mode should never become the final persistent mode solely because overlapping clients exited in a different order.

## Actual Behavior

The final input mode is `0x3D8`, even though both WSL processes exit successfully. This leaves `ENABLE_VIRTUAL_TERMINAL_INPUT` (`0x200`) enabled and also reflects WSL's other temporary flag changes.

For native Windows TUIs that consume Win32/crossterm key events, arrow and editing keys are then translated into VT sequences and their tails can appear as literal text: arrows become `[A`/`[B`/`[C`/`[D`, Delete becomes `[3~`, and Enter/Backspace/control-key handling can fail. The console can remain broken after WSL exits until the original input mode is restored or the console is recreated.

The behavior matches WSL 2.7.11's `ConsoleState` implementation:

- Each client snapshots the shared `CONIN$` mode and applies WSL's temporary mode: https://github.com/microsoft/WSL/blob/2.7.11/src/windows/common/ConsoleState.cpp#L65-L86
- Each client later restores its own snapshot unconditionally: https://github.com/microsoft/WSL/blob/2.7.11/src/windows/common/ConsoleState.cpp#L118-L135

For overlapping clients, the observed sequence is:

```text
A saves original mode, then applies WSL mode
B saves A's WSL mode
A exits and restores original mode
B exits and restores the WSL mode it saved
```

Possible fix directions include coordinating console-state ownership across overlapping `wsl.exe` clients, reference-counting per console input buffer, or avoiding restoration of a snapshot known to represent another live client's temporary state.

## Diagnostic Logs

No private application/session logs are required for this standalone reproduction. I can collect WSL diagnostics for the isolated test if maintainers need them, but I have intentionally not attached unrelated project data.

shuaiyuanxx:

Fixes console-state races when concurrent `wsl.exe` clients share one Windows console. The WSL interactive state now remains active until the last participating client exits, then the original input/output modes and code pages are restored.

## PR Checklist

- [x] **Closes:** Fixes #41201
- [x] **Localization:** No end-user-facing strings were added
- [x] **Dev docs:** No development documentation changes are needed
- [ ] **Documentation updated:** No user documentation changes are needed

## Details

This change coordinates clients directly with a per-console named mutex and shared file mapping:

- the first client captures the baseline and configures the console
- overlapping clients share the configured state
- only the last normally exiting client restores the baseline
- restoration remains conditional, preserving external console changes
- dead owner processes are pruned so later clients are not blocked

`RestorePolicy::Always` remains the default for WSLC and other existing callers. Coordinated behavior is limited to opted-in WSL launch paths, including WSLg, legacy bash, and debug shell.

Restoring the baseline after the sole remaining client is forcibly terminated remains outside scope because no client remains to restore process-local code pages. A later client can still acquire coordination without blocking.

## Validation

- Built Debug `common`, `wsl`, `wslservice`, `wslc`, and `wsltests`
- Verified two overlapping clients with the Debug `wsl.exe`: the console remained configured after the first client exited and returned to baseline after the final client exited
- `git diff --check` passed


src/windows/common/ConsoleState.cpp
```diff
@@ -58,6 +58,18 @@ try
...
 CATCH_LOG()
 
+std::optional<DWORD> TryGetConsoleMode(_In_ HANDLE Handle)
+{
+    DWORD mode{};
+    if (!GetConsoleMode(Handle, &mode))
+    {
+        LOG_LAST_ERROR_MSG("GetConsoleMode failed");
+        return std::nullopt;
+    }
```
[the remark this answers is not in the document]
shuaiyuanxx: Fixed in af4f116. `TryGetConsoleMode()` now suppresses expected `ERROR_PIPE_NOT_CONNECTED` and `ERROR_INVALID_HANDLE` teardown failures while continuing to log unexpected errors. I also addressed the module-path truncation and clang-format findings.

test/windows/wslc/WSLCCLIVTSupportUnitTests.cpp
```diff
@@ -33,6 +35,69 @@ static wil::unique_hfile MakeScreenBuffer()
     return handle;
 }
 
+static std::wstring ResolveWslExecutablePath()
```
dkbennett: These unit tests are for the WSLC CLI specifically, which does use ConsoleState that is shared between wsl and wslc. These tests should go in the parent UnitTests.cpp with these helper methods going in the Common.cpp/Common.h test file.

These are all good tests to add, but as this can affect both wsl.exe and wslc.exe, it should be in the parent UnitTests set.

test/windows/wslc/WSLCCLIVTSupportUnitTests.cpp
```diff
@@ -505,6 +570,228 @@ class WSLCCLIVTSupportUnitTests
...
+            return;
+        }
+
+        DWORD baseline{};
+        VERIFY_WIN32_BOOL_SUCCEEDED(GetConsoleMode(conin.get(), &baseline));
+
+        auto process = StartSleepingWslProcess(2, CREATE_NEW_CONSOLE | CREATE_UNICODE_ENVIRONMENT | EXTENDED_STARTUPINFO_PRESENT);
+        VERIFY_IS_TRUE(!!process);
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
dkbennett: Sleeps in the test should be carefully scrutinized, because this can lead to flaky tests when the test system is under heavy load (and it often is).

There's still a race condition in SetConsoleState because reading and setting the state is not atomic, and a system under heavy load with sleep timers might randomly fail due to these. This PR reduces the window of vulnerability to the race, but it can't eliminate it completely.

test/windows/wslc/WSLCCLIVTSupportUnitTests.cpp
```diff
@@ -505,6 +570,228 @@ class WSLCCLIVTSupportUnitTests
...
+        VERIFY_ARE_EQUAL(externallyAdjusted, afterFirstExit);
+
+        second.reset();
+
+        DWORD finalMode{};
+        VERIFY_WIN32_BOOL_SUCCEEDED(GetConsoleMode(conin.get(), &finalMode));
+        VERIFY_ARE_EQUAL(externallyAdjusted, finalMode);
+    }
+
+    TEST_METHOD(ConsoleState_WslProcesses_SharedConsole_OutOfOrderRestore)
```
dkbennett: The first two ConsoleState tests test ConsoleState in isolation - good unit tests. 

ConsoleState_ConcurrentInputClients_OutOfOrderRestore - good unit test, put in UnitTests.cpp
ConsoleState_ConcurrentInputClients_ExternalModeChangePreserved - good unit test, put in UnitTests.cpp

The three other "WslProcess" tests are starting WSL processes and are not unit tests of the ConsoleState in isolation, they should be moved to the "SimpleTests.cpp" as functional tests and not run as part of the unit test suite.

test/windows/wslc/WSLCCLIVTSupportUnitTests.cpp
```diff
@@ -33,6 +35,69 @@ static wil::unique_hfile MakeScreenBuffer()
     return handle;
 }
 
+static std::wstring ResolveWslExecutablePath()
+{
+    // Prefer a co-located test-built wsl.exe when present so process tests
```
OneBlue: I don't recommend doing this. This will lead to surprising results when running the tests locally. 

We should have the same behavior as the other tests: Use the installed version of wsl.exe

test/windows/wslc/WSLCCLIVTSupportUnitTests.cpp
dkbennett: Also:
Most of the new helpers re-implement things the test framework already provides.
When these move into `Common.h`/`Common.cpp`, please fold them into the existing
helpers rather than adding parallel copies:

- `WaitForProcessExit(process, timeoutMs)` is a one-line wrapper over
  `wsl::windows::common::SubProcess::GetExitCode(process, timeoutMs)`
  (`src/windows/common/SubProcess.h`). Call `GetExitCode` directly and drop the wrapper.
- `ResolveWslExecutablePath()` re-derives the wsl.exe path; its own fallback already
  calls `LxssGenerateWslCommandLine(nullptr, L"wsl.exe")` (`test/windows/Common.h:469`),
  which is the canonical way to locate the test wsl.exe. The co-located-binary probe
  looks redundant with that.
- `StartSleepingWslProcess(...)` builds a command line by hand and calls
  `LxsstuStartProcess` (`test/windows/Common.h:504`). Prefer building the args via
  `LxssGenerateWslCommandLine`, or reuse `LxsstuLaunchWsl` / `LxsstuLaunchWslAndCaptureOutput`
  (`test/windows/Common.h:445-471`) where a wait-and-capture fits.

Net: only the console-mode polling helper (`WaitForInputModeChange`) is genuinely new;
the process-launch/exit helpers should reuse the existing common ones.

src/windows/common/ConsoleState.cpp
```diff
@@ -137,29 +158,53 @@ void ConsoleState::RestoreConsoleState()
...
+            }
+
             m_SavedOutputCodePage.reset();
+            m_ConfiguredOutputCodePage.reset();
         }
 
         if (m_SavedOutputMode.has_value())
         {
-            TrySetConsoleMode(m_OutputHandle.get(), m_SavedOutputMode.value());
+            const auto currentMode = TryGetConsoleMode(m_OutputHandle.get());
```
dkbennett: I'm a little wary of this change because it is actually changing our single client behavior and might have unexpected consequences in the WSLC relayed console scenarios (attached to a container or session)

Old contract: the destructor unconditionally reapplied m_SavedInputMode. WSL was authoritative and whatever it found at teardown, it forced back to the baseline it started with.

New contract: WSL restores the baseline only if the console is still exactly the mode it configured. If anything moved it off of what it configured in the meantime no action is taken at all. I'm concerned this might leave us with a corrupted terminal whereas the unconditional re-apply would always get us back to the state we started when the CLI exits.

I'm concerned about the impact to WSLC. This issue is a wsl.exe issue that WSLC does not currently have, and it can cause a regression in a legitimate WSLC scenario:

A session causes a host CONIN/CONOUT mode or drift, and then the CLI exits normally. The CLI will not reset to baseline because it is not exactly as it was configured. I very much like the defensiveness of the unconditional restore for the CLI to prevent attach scenarios from corrupting the user's console.

I think the correct solution here is to have opt-in behavior change and tests which verify the behaviors.

dkbennett (dismissed):
`ConsoleState` is shared code, and the race this fixes (#41201) is specifically a `wsl.exe` problem: two overlapping `wsl.exe` clients sharing one inherited console and exiting out of order. WSLC's interactive paths (container attach/exec, session enter) are single foreground console owners, so they do not hit that race, but they do rely on `ConsoleState` unconditionally sanitizing the console on exit.

As written, the conditional restore changes behavior for every caller: if the console mode drifts from what a client configured, teardown now skips the restore. For WSLC that removes the guaranteed reset-on-exit and can leave a corrupted console (for example, echo or line input left disabled) after the CLI exits, with nothing else owning `CONIN$` to repair it.

Proposal: gate the new behavior behind an explicit `RestorePolicy` that defaults to the existing unconditional restore, and opt only the `wsl.exe` sites into the conditional path.

```cpp
enum class RestorePolicy
{
    // Always reapply saved state on teardown. Guarantees the console is sanitized on exit
    // even if the mode drifted mid-session. Correct for a sole console owner.
    Always,

    // Reapply only if the console still matches what this instance configured. Prevents
    // concurrent shared-console clients from restoring a stale temporary mode captured from
    // another client. Trades away the guaranteed reset.
    OnlyIfUnchanged,
};

explicit ConsoleState(RestorePolicy restorePolicy = RestorePolicy::Always);

// ... stored as:
RestorePolicy m_restorePolicy;
```

Restore branches on the policy, keeping the readback and configured tracking and just gating the check:

```cpp
if (m_SavedInputMode.has_value())
{
    const bool restore = (m_restorePolicy == RestorePolicy::Always) ||
                         !m_ConfiguredInputMode.has_value() ||
                         (TryGetConsoleMode(m_InputHandle.get()) == m_ConfiguredInputMode);
    if (restore)
    {
        TrySetConsoleMode(m_InputHandle.get(), m_SavedInputMode.value());
    }

    m_SavedInputMode.reset();
    m_ConfiguredInputMode.reset();
}
```

The same shape applies to the input code page, output mode, and output code page.

Call sites:

Opt in to the concurrency fix (wsl.exe):
- `src/windows/common/WslClient.cpp` -> `ConsoleState console{RestorePolicy::OnlyIfUnchanged};`
- `src/windows/common/svccomm.cpp` -> `ConsoleState Io{RestorePolicy::OnlyIfUnchanged};`

Unchanged, default `Always` (guaranteed reset):
- `src/windows/wslc/services/ContainerService.cpp`
- `src/windows/wslc/services/SessionService.cpp`
- `src/windows/wslc/services/ConsoleService.cpp`

Notes:
- The default is the safe behavior, so existing WSLC call sites need no edits and keep today's guarantee.
- With this default, the two `wsl.exe` sites must opt in explicitly or the fix is inert. That is intended, and it makes the behavior change visible at the call site.
- WSLC can opt individual instances into `OnlyIfUnchanged` later if a real shared-console case appears, without another shared-code behavior change.

Also the tests need to be refactored into appropriate locations (2 are actual unit tests, 3 are functional, and they are common tests, not WSLC CLI tests), with the timing tests receiving a bit of scrutiny due to the potential to be flaky.

shuaiyuanxx:
Hi @dkbennett Thanks for the detailed review. The requested changes are addressed:

- Added `RestorePolicy`, defaulting to `Always` to preserve WSLC behavior.
- Opted only WSL/WSLg process launches into `OnlyIfUnchanged`.
- Moved tests to `UnitTests.cpp` and `SimpleTests.cpp`.
- Replaced sleep-based timing with deterministic pipe synchronization.

The relevant builds and focused tests pass. Could you please take another look?

dkbennett (commented):
Removing my block since the parity concern has been resolved, I'm sure others may want to review this and the tests.

chemwolf6922:
The current change only ensures a correct final state. But does not ensure the correct order / state during the operation. For example:
Original state is 0 -> A sets it from 0 to 1 -> B sets it from 1 to 1 -> A resets it from 1 to 0 -> B is running with the wrong state -> B skips the reset -> final state is 0.

src/windows/common/ConsoleState.h
```diff
@@ -20,11 +20,23 @@ Module Name:
...
+class SvcComm;
+
+// Controls when ConsoleState attempts to restore the original console state.
+enum class RestorePolicy
+{
+    // Always restore to the original state captured by this instance.
+    Always,
+
+    // Restore only when the state still matches what this instance configured.
+    OnlyIfUnchanged,
```
chemwolf6922, line 32: Please consider rename this to Exclusive / Cooperative instead of Always / OnlyIfUnchanged to better reflect the purpose.

src/windows/common/ConsoleState.cpp
```diff
@@ -58,11 +57,129 @@ try
...
+    struct Owner
+    {
+        DWORD ProcessId;
+        ULONG References;
+        ULONGLONG CreationTime;
+    };
+
+    ConsoleValues Baseline;
+    ConsoleValues Configured;
+    Owner Owners[128];
```
chemwolf6922: This owner logic could be replaced with a named event. 
When setting the values:
```
{
    MutexLock();
    m_event = CreateEvent(..., name);
    isFirst = GetLastError() != ERROR_ALREADY_EXISTS;
    if (isFirst)
        SetTheValues();
}
```
When resetting the values:
```
{
    MutexLock();
    CloseHandle(m_event);
    event = CreateEvent(..., name);
    isLast = GetLastError() != ERROR_ALREADY_EXISTS;
    if (isLast)
        ResetTheValues();
}
```

src/windows/common/ConsoleState.cpp
```diff
@@ -147,38 +286,216 @@ void ConsoleState::SetOutputCodePageUtf8()
...
+bool ConsoleState::AcquireCoordination()
+{
+    const auto consoleId = TryGetConsoleId();
+    if (!consoleId || !m_InputHandle || !m_OutputHandle)
+    {
+        return false;
+    }
+
+    const auto name = std::format(L"Local\\WSL.ConsoleState.v1.{:X}", consoleId.value());
+    m_coordinationMutex.reset(CreateMutexW(nullptr, FALSE, (name + L".Mutex").c_str()));
```
chemwolf6922: A correct security descriptor is required for elevated and normal processes to access the same object (mutex, event, file mapping).

src/windows/common/ConsoleState.cpp
```diff
@@ -88,23 +205,37 @@ void ConsoleState::SetInteractiveMode()
         return;
     }
 
+    if ((m_restorePolicy == RestorePolicy::Cooperative) && AcquireCoordination())
```
chemwolf6922: The current logic is mixing two solutions because of this fallback. Should this just fail instead of falling back to the old partial solution?

src/windows/common/ConsoleState.h
```diff
@@ -20,11 +20,21 @@ Module Name:
 
 namespace wsl::windows::common {
 
+// Controls when ConsoleState attempts to restore the original console state.
+enum class RestorePolicy
+{
+    // Exclusive restore to the original state captured by this instance.
+    Exclusive,
+
+    // Restore only when the state still matches what this instance configured.
```
chemwolf6922: Stale comment.

src/windows/common/ConsoleState.cpp
```diff
@@ -147,38 +286,216 @@ void ConsoleState::SetOutputCodePageUtf8()
...
+    if (m_coordinationView)
+    {
+        try
+        {
+            ReleaseCoordination();
+        }
+        CATCH_LOG()
+
+        LOG_IF_WIN32_BOOL_FALSE(UnmapViewOfFile(m_coordinationView));
+    }
```
chemwolf6922: Please consider put this into the RestoreConsoleState function.

src/windows/common/ConsoleState.cpp
```diff
@@ -147,38 +286,216 @@ void ConsoleState::SetOutputCodePageUtf8()
...
+        }
+        else if (owner.ProcessId && !IsOwnerAlive(owner))
+        {
+            owner = {};
+        }
+    }
+
+    if (std::none_of(std::begin(state.Owners), std::end(state.Owners), [](const auto& owner) { return owner.ProcessId != 0; }))
+    {
+        if (TryGetConsoleMode(m_InputHandle.get()) == state.Configured.InputMode)
```
chemwolf6922: These comparisons are not required for the new solution.
They also do not perfectly serve the purpose of preserving third party set modes if the third party set modes happens to be the configured one. Please consider remove the configured values and these comparisons if the fallback path is removed.

OneBlue (changes requested):
@shuaiyuanxx: Looking at the original issue, I'm not sure that the complexity that this change brings is worth it. 

Running multiple processes under the same TTY in parallel is inherently going to lead to race conditions, since there will be a TOCTOU between reading the current terminal's state and deciding whether it should be changed or not. 

If a user wants to do something like this, the easiest solution would just be to not run multiple interactive processes in the same terminal session, and use pipes or `CREATE_NEW_CONSOLE`

PR #41211, opened 2026-07-30
Add registry allowlist support to image builds

beena352:
The WSLContainerRegistryAllowlist group policy was not enforced by wslc image build - only wslc image pull respected it, so a Dockerfile's FROM/COPY --from could pull from a registry the admin had explicitly disallowed. This PR closes that gap by generating a BuildKit source-policy document (EXPERIMENTAL_BUILDKIT_SOURCE_POLICY) that denies all image sources except the allowlisted hosts, and passes it to docker buildx build.

## PR Checklist

- The policy JSON is generated in the user-side wslcsession process during VM boot (ConfigureBuildKitPolicy in WSLCVirtualMachine::Initialize) and sent to the guest over the existing init message channel. The guest handler writes it to /run/wsl/buildkit-policy.json on tmpfs. On each wslc image build, if the policy was configured at boot, EXPERIMENTAL_BUILDKIT_SOURCE_POLICY is set in the build environment pointing at that path
- The JSON contains a DENY-all rule followed by per-allowlisted-host ALLOW rules (REGEX match, last-rule-wins), with hostnames lowercased and regex-metacharacters escaped
- Fail-closed: if the allowlist policy is configured but can't be read (e.g. registry access denied), the build is refused with WSLC_E_REGISTRY_BLOCKED_BY_POLICY rather than silently proceeding without enforcement. If the message transaction or guest write fails, the VM fails to initialize entirely
- The registry is read once at boot rather than per-build - this closes a race where a compromised user process could tamper with the policy between write and build. The staleness window is acceptable because wslc VMs are short-lived and Group Policy itself refreshes on a ~90-minute cadence

## Validation Steps Performed
- Automated: PolicyTests.cpp - RegistryAllowlistBlocksImageBuild, RegistryAllowlistAllowsImageBuild, RegistryAllowlistImageBuildIsCaseInsensitive, RegistryAllowlistBlocksImageBuildCopyFrom, RegistryAllowlistBlocksImageBuildImplicitDockerIo, RegistryAllowlistDenies, ReadRegistryAllowlistSnapshotFromPoliciesRoot_Logic. Tests now terminate the session before each build to ensure a fresh boot-time snapshot
- Manual: verified an allowed FROM builds successfully and a disallowed FROM/COPY --from is blocked with the policy-block error surfaced to the user.


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -909,6 +900,22 @@ try
 
     THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_virtualMachine);
 
+    // Declared before unmountAll: scope_exits run in reverse order, so the plan9/virtiofs share
+    // is torn down before we ask the service to delete the host folder.
+    wil::unique_cotaskmem_string policyHostPath;
+    const auto policyHr = m_virtualMachine->PrepareBuildKitSourcePolicy(&policyHostPath);
```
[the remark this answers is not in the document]
beena352: wil::unique_cotaskmem_string  overloads  operator& to return the address of its internal LPWSTR* (that's the WIL out-param pattern). No change needed.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -931,15 +938,22 @@ try
...
-    auto mountPath = mountInVm(Options->ContextPath, TRUE);
-
-    std::vector<std::string> buildArgs{"/usr/bin/docker", "buildx", "build", "--builder", "default", "--progress=rawjson"};
     // Environment for the docker process. Env/in-memory secrets are delivered as variables here so their
     // values never touch disk; kept off telemetry (only buildArgs is logged).
     std::vector<std::string> buildEnv;
+
+    if (policyHostPath)
+    {
+        auto policyMountPath = mountInVm(policyHostPath.get(), TRUE);
```
OneBlue: Instead of writing the policy file on Windows and then mounting it, could we directly writing inside the VM during boot ? That would simplify things quite a bit

src/windows/service/exe/HcsVirtualMachine.cpp
```diff
@@ -73,6 +77,70 @@ std::wstring SanitizeHostingProcessNameSuffix(std::wstring_view name)
...
+    }
+    return escaped;
+}
+
+// DENY-all first, then per-host ALLOW: BuildKit evaluates rules in order and last match wins.
+// Hosts are lowercased because BuildKit normalises identifiers before regex matching.
+// https://github.com/moby/buildkit/blob/master/docs/sourcepolicy.md
+std::string BuildBuildKitSourcePolicyJson(const std::vector<std::string>& allowedHosts)
+{
+    nlohmann::json rules = nlohmann::json::array();
```
OneBlue: @craigloewen-msft : Given that we can use buildkit policies, what do you think of putting the entire policy in the registry / well known file path ? That would both allow users to do more powerful things without having us to parse / validate the policies ourselves

beena352: My concern with pure passthrough is that push/pull already use WSLContainerRegistryAllowlist  and don't go through BuildKit, so they can't consume a BuildKit source-policy JSON. Admins would end up setting two different things for the same idea.
Could we do both? Keep the allowlist as the default, and add a file path option later (something like WSLContainerBuildKitSourcePolicyFile ) for people who need the full BuildKit policy syntax. I'd rather not put that in this PR though - happy to do it as a follow-up if we hear from users who actually need it.
Waiting on Craig's take before I change anything here.

src/shared/inc/lxinitshared.h
```diff
@@ -1926,6 +1928,21 @@ struct WSLC_GET_GUEST_CAPABILITIES
     PRETTY_PRINT(FIELD(Header));
 };
 
+struct WSLC_SET_BUILDKIT_POLICY
```
OneBlue: If we want to use a new message to write this file, I would recommend creating a generic "write file" message. Something like: 

```
struct WSLC_WRITE_FILE
{
[...]
    unsigned int PathIndex;
    int OpenFlags;
    int Permissions;
    char Buffer[];
};
```

That way we can reuse this in other places

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -440,6 +494,38 @@ void WSLCVirtualMachine::ReadGuestCapabilities()
     THROW_IF_FAILED(m_vm->ApplyGuestCapabilities(&capabilities));
 }
 
+void WSLCVirtualMachine::ConfigureBuildKitPolicy()
+{
+    const auto snapshot = wsl::windows::policies::ReadRegistryAllowlistSnapshotFromPoliciesRoot();
+    if (snapshot.State == wsl::windows::policies::RegistryAllowlistState::ReadFailed)
+    {
+        // Fail closed: `BuildImage` refuses new builds when it sees this state.
+        m_buildKitPolicyState = BuildKitPolicyState::ReadFailed;
```
OneBlue: I think it would be OK to hard fail if we couldn't read the policy. We could throw an exception when we read the registry and let it bubble up to fail the session creation entirely

src/windows/service/exe/HcsVirtualMachine.cpp
craigloewen-msft: Not totally opposed to that. I would lean to file path. 
Only challenge here would be: What if we change underlying tech in the future? 
But as long as we're not too worried about that I'd be ok with exposing a file. We can do it as experimental for now and then wait until we hear more from companies if they need specific needs.

beena352: Sounds good - I'll keep the allowlist here and open a follow-up to add an experimental WSLContainerBuildKitSourcePolicyFile for admins who need the full BuildKit policy syntax.

src/linux/init/WSLCInit.cpp
```diff
@@ -896,6 +896,33 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, wsl::shared::Transac
...
+        {
+            result = errno;
+        }
+        else if (UtilWriteBuffer(fd.get(), content) != static_cast<ssize_t>(content.size()))
+        {
+            result = errno;
+        }
+    }
+
+    Transaction.SendResultMessage<int32_t>(result);
```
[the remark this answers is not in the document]
beena352: Good catch on the caller-response angle, ProcessMessage 's outer catch swallows exceptions without sending a reply, so a malformed message would leave the host waiting on the transaction timeout. Added the bounds guard with an early EINVAL  response; kept the shape flat to avoid re-indenting the existing body.

src/windows/wslcsession/WSLCVirtualMachine.cpp
```diff
@@ -444,6 +498,52 @@ void WSLCVirtualMachine::ReadGuestCapabilities()
     THROW_IF_FAILED(m_vm->ApplyGuestCapabilities(&capabilities));
 }
 
+void WSLCVirtualMachine::ConfigureBuildKitPolicy()
+{
+    const auto snapshot = wsl::windows::policies::ReadRegistryAllowlistSnapshotFromPoliciesRoot();
+
+    // Fail closed: an unreadable policy must not silently let builds through.
+    THROW_HR_WITH_USER_ERROR_IF(
+        WSLC_E_REGISTRY_BLOCKED_BY_POLICY,
```
OneBlue: This error feels a bit confusing because it gives the impression that the policy explicitly prevents builds. Instead I would recommend throwing directly in `ReadRegistryAllowlistSnapshotFromPoliciesRoot`  with an error message saying that the policies are invalid (and if possible, details as to what was invalid)

test/windows/PolicyTests.cpp
```diff
@@ -485,38 +485,106 @@ class PolicyTest
...
+    WSLC_TEST_METHOD(RegistryAllowlistBlocksImageBuild)
     {
         auto revert = SetRegistryAllowlist({L"mcr.microsoft.com"});
 
-        // Set up a minimal build context with a one-line Dockerfile in TEMP.
-        const auto contextDir = std::filesystem::temp_directory_path() / L"wsl-policy-build-test";
+        auto [exitCode, output] = RunImageBuild(L"FROM docker.io/library/alpine:latest\n", L"wsl-policy-build-blocked");
+
+        VERIFY_ARE_NOT_EQUAL(0, exitCode);
+        if (output.find(L"docker.io") == std::wstring::npos)
```
OneBlue: I recommend validating the full error message here. We can use `VerifyPatternMatch` if the error message has variable fields

OneBlue: (Same for below)

OneBlue (commented):

src/linux/init/WSLCInit.cpp
```diff
@@ -896,6 +896,40 @@ void HandleMessageImpl(wsl::shared::SocketChannel& Channel, wsl::shared::Transac
     Transaction.SendResultMessage<int32_t>(result);
 }
 
+void HandleMessageImpl(
+    wsl::shared::SocketChannel& Channel, wsl::shared::Transaction& Transaction, const WSLC_WRITE_FILE& Message, const gsl::span<gsl::byte>& Buffer)
+{
+    const auto* path = wsl::shared::string::FromSpan(Buffer, Message.PathIndex);
```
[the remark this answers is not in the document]
benhillis: This looks worth fixing

src/windows/inc/wslpolicies.h
```diff
@@ -200,4 +200,74 @@ inline bool HasRegistryAllowlist(HKEY policiesKey)
...
+
+    return snapshot;
+}
+
+// Convenience for callers with no open policies key. Throws MessageRegistryAllowlistPolicyInvalid
+// when the policies key can't be opened.
+inline RegistryAllowlistSnapshot ReadRegistryAllowlistSnapshotFromPoliciesRoot()
+{
+    wil::unique_hkey policiesKey;
+    const auto openResult = RegOpenKeyExW(HKEY_LOCAL_MACHINE, c_registryKey, 0, KEY_READ, &policiesKey);
```
OneBlue, line 261: Here we're opening `c_registryKey` and then ReadRegistryAllowlistSnapshot() opens `c_wslContainerRegistryAllowlist` as a sub key. 

Could we open that subkey directly and pass it to ReadRegistryAllowlistSnapshot() instead ?

test/windows/PolicyTests.cpp
```diff
@@ -494,7 +490,7 @@ class PolicyTest
         auto [exitCode, output] = RunImageBuild(L"FROM docker.io/library/alpine:latest\n", L"wsl-policy-build-blocked");
 
         VERIFY_ARE_NOT_EQUAL(0, exitCode);
-        if (output.find(L"docker.io") == std::wstring::npos)
+        if (output.find(L"docker.io") == std::wstring::npos || output.find(L"denied by policy") == std::wstring::npos)
```
OneBlue, line 493: I recommend validating the full error message here (same for below)

test/windows/PolicyTests.cpp
```diff
@@ -582,4 +647,59 @@ class PolicyTest
...
-};
\ No newline at end of file
+
+    // Pure-function tests for ReadRegistryAllowlistSnapshot (used by `wslc image build` to
+    // decide between fail-open-no-policy, generate-source-policy, and fail-closed paths).
+    TEST_METHOD(ReadRegistryAllowlistSnapshot_Logic)
+    {
+        // Null policies key -> NotConfigured, no hosts.
+        {
+            const auto snapshot = ReadRegistryAllowlistSnapshot(nullptr);
```
OneBlue, line 657: nit: This code path should be impossible now so I think we can remove this test

OneBlue (approved):
LGTM, minor comments. Feel free to address in a followup

PR #41212, opened 2026-07-30
Add --iidfile to wslc image build

ggarzia-MSFT:

Adds a `--iidfile <path>` option to `wslc build`, mirroring `docker build --iidfile`. After a successful build, the resulting image ID is written to the given file. The file is created/truncated up front (so a bad path fails fast before the long build runs) and is removed if the build fails, so a failed build never leaves a stale file behind.

## PR Checklist

- **CLI argument:** New `iidfile` value argument registered in `ArgumentDefinitions.h` and wired into the build command. Its description string (`WSLCCLI_IidFileArgDescription`) is added to `Resources.resw` for localization.
- **RAII helper (`IidFile`):** New `ImageModel.{h,cpp}` introduces a `models::IidFile` class that:
  - Opens the file with `CREATE_ALWAYS` at construction so an invalid path surfaces an error *before* the build starts (matching `docker build`), throwing a localized user error on failure.
  - Writes the image ID on `Commit()`.
  - Deletes the file on destruction if `Commit()` was never called, so a failed/cancelled build leaves no file behind.
  - Is `NON_COPYABLE` / `NON_MOVABLE`.
- **Service/session plumbing:** `ImageService::Build` now returns the image ID string. A new `WSLCBuildImageFlagsReturnImageId` flag is set only when `--iidfile` is supplied, and the server (`WSLCSession`) returns the image ID via a new out-parameter on `BuildImage` (IDL updated in `wslc.idl` / `WSLCShared.idl`). `ImageTasks.cpp` orchestrates: create the `IidFile`, run the build, then `Commit()` the returned ID.

## Validation Steps Performed

- **Unit tests:** Updated `test/windows/WSLCTests.cpp` for the new flag/out-parameter and image-ID return path.
- **E2E tests:** Added coverage in `test/windows/wslc/e2e/WSLCE2EImageBuildTests.cpp` exercising `wslc build --iidfile`, verifying the file contains the built image ID and that a failed build does not leave the file behind.
- **Manual:** Ran `wslc build --iidfile <path>` and confirmed the image ID is written to the file; verified an invalid path fails before the build runs and a failed build leaves no file.

### DIFF
--- a/src/windows/wslc/services/ImageService.cpp
+++ b/src/windows/wslc/services/ImageService.cpp
@@ -123,6 +123,7 @@ void ImageService::Build(
     const std::wstring& dockerfilePath,
     const std::wstring& target,
     const std::optional<BuildOutput>& output,
+    const std::optional<std::wstring>& iidFilePath,
     WSLCBuildImageFlags flags,
     IProgressCallback* callback,
     HANDLE cancelEvent)
@@ -259,6 +260,15 @@ void ImageService::Build(
     }
 
     auto contextPathStr = absolutePath.wstring();
+
+    // Resolve the --iidfile destination against the client's working directory; the server mounts its
+    // parent directory read-write into the VM so buildx writes the image ID straight to it.
+    std::wstring iidPathStr;
+    if (iidFilePath.has_value())
+    {
+        iidPathStr = std::filesystem::weakly_canonical(std::filesystem::absolute(*iidFilePath)).wstring();
+    }
+
     WSLCBuildImageOptions options{
         .ContextPath = contextPathStr.c_str(),
         .DockerfileHandle = ToCOMInputHandle(dockerfileHandle),
@@ -272,6 +282,7 @@ void ImageService::Build(
         .OutputHandle = outputHandle != nullptr ? ToCOMInputHandle(outputHandle) : WSLCHandle{.Type = WSLCHandleTypeUnknown},
         .OutputMountPath = outputMountPath.empty() ? nullptr : outputMountPath.c_str(),
         .OutputMountFile = outputMountFile.empty() ? nullptr : outputMountFile.c_str(),
+        .IidFilePath = iidPathStr.empty() ? nullptr : iidPathStr.c_str(),
     };
 
     THROW_IF_FAILED(session.Get()->BuildImage(&options, callback, cancelEvent));
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -1014,9 +1014,9 @@ try
 
     // Reserve up front so mountInVm's push_back can never reallocate-and-throw after a successful
     // MountWindowsFolder, which would leak a mount the scope_exit hasn't recorded yet. At most the build
-    // context (1), the single-file exporter output destination (1), and one parent directory per file
-    // secret are mounted.
-    mountedPaths.reserve(static_cast<size_t>(2) + Options->Secrets.Count);
+    // context (1), the single-file exporter output destination (1), the --iidfile destination (1), and
+    // one parent directory per file secret are mounted.
+    mountedPaths.reserve(static_cast<size_t>(3) + Options->Secrets.Count);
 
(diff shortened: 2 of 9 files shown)


src/windows/wslc/services/ImageModel.h
```diff
@@ -34,4 +34,24 @@ struct PruneImagesResult
     std::vector<std::string> UntaggedImages;
     ULONGLONG SpaceReclaimed{};
 };
+
+// RAII helper for `wslc build --iidfile <path>`. Creates (or truncates) the file up front so a bad path
```
OneBlue: I would recommend skipping this check on the client side since there's a race condition between the time the client tests for access and the service does. 

For something like this I think just passing a `weakly_canonical` path to the service and letting it do all the work

ggarzia-MSFT: Good catch, I think this was leftover from before I mounted the destination into the VM and just wrote there

test/windows/wslc/e2e/WSLCE2EImageBuildTests.cpp
```diff
@@ -1267,6 +1267,59 @@ class WSLCE2EImageBuildTests
...
+        WriteTestFileContent(dockerfilePath, "FROM debian:latest\nRUN exit 7\n");
+
+        const auto iidFilePath = testRoot / L"image.id";
+
+        auto buildResult = RunWslc(std::format(
+            L"build \"{}\" -f \"{}\" --iidfile \"{}\"", contextDir.wstring(), dockerfilePath.wstring(), iidFilePath.wstring()));
+        VERIFY_ARE_EQUAL(1u, buildResult.ExitCode.value_or(0u));
+        VERIFY_IS_FALSE(std::filesystem::exists(iidFilePath), L"a failed build must not leave an iidfile behind");
+    }
+
```
OneBlue, line 1322: nit: I recommend adding test coverage for the case where the file is not writeable, and where the parent directly doesn't exist, just to make sure that handle those error paths correctly

OneBlue (dismissed):
LGTM, minor test comments. Feel free to address in a followup

dkbennett (approved):

src/windows/wslc/services/ImageService.cpp
```diff
@@ -259,6 +260,15 @@ void ImageService::Build(
     }
 
     auto contextPathStr = absolutePath.wstring();
+
+    // Resolve the --iidfile destination against the client's working directory; the server mounts its
+    // parent directory read-write into the VM so buildx writes the image ID straight to it.
+    std::wstring iidPathStr;
+    if (iidFilePath.has_value())
+    {
+        iidPathStr = std::filesystem::weakly_canonical(std::filesystem::absolute(*iidFilePath)).wstring();
```
OneBlue, line 269: nit: `std::filesystem::absolute` is not needed here. weakly_canonical paths are always absolute

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1081,6 +1081,23 @@ try
...
+    // Docker-style --iidfile. The destination's parent directory is mounted read-write into the VM so
+    // buildx writes the image ID straight to the client's --iidfile path.
+    if (Options->IidFilePath != nullptr && Options->IidFilePath[0] != L'\0')
+    {
+        std::filesystem::path iidPath(Options->IidFilePath);
+        // The client and server have different current directories, so a relative path is ambiguous.
+        THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessagePathNotAbsolute(Options->IidFilePath), !iidPath.is_absolute());
+
+        auto iidParent = iidPath.parent_path();
+        auto iidFileNameUtf8 = wsl::shared::string::WideToMultiByte(iidPath.filename().wstring());
```
OneBlue, line 1094: nit: This could be simplified to: `auto iidFileNameUtf8 = iidPath.filename().string();`

OneBlue (approved):
LGTM. Feel free to resolve the comments in a followup change

PR #41215, opened 2026-07-30
CLI: Unify interactive input and output in Reporter

dkbennett:

Unifies interactive user input with user-facing output under the WSLC CLI's `Reporter`. Previously `Reporter` only handled output, and interactive input (prompting, echo masking) was handled ad hoc inside commands like `wslc registry login`, where console handling does not belong. This moves that logic into a single, testable owner so commands can prompt consistently instead of reimplementing console handling, and `registry login` is simplified to just ask the `Reporter` for input. Prompts for input were changed from stderr to stdout intentionally (see detailed description for reasoning).

Note that `Reporter` is planned to be renamed in a follow-up PR to reflect that it now provides unified input/output management for the CLI; the current name is kept here to keep the diff focused and does not yet fully describe the expanded responsibility.

Sample output from manual testing with stderr redirected to show prompts still function as expected consistent with other CLIs:
image
(note these commands failed expectedly but the purpose was to verify the input behavior on a real console; the tests verify the command)

## PR Checklist

- New `InputChannel`: reads a line at a time from the console or a redirected file/pipe. On an interactive console it can mask echo (for password entry) and report whether input is interactive; it distinguishes an empty line from end of input.
- `Reporter` now owns an input channel alongside its output channels and exposes:
  - `IsInputInteractive()` to check whether a prompt can be shown.
  - `ReadLine(mask)` to read a single line, optionally masking echo.
  - `PromptForLine(label, mask)` to write a label and read the response in one call.
- `wslc registry login` now uses `PromptForLine` for the username and (masked) password prompts, replacing the previous inline prompting.
- Updated the interactive-login E2E assertions from stderr to stdout to match the new prompt stream.
- Unit tests and test helpers covering the input channel and the reporter's prompt/read paths.

### Prompt default output stream

Interactive login prompts now write to stdout. Previously they were emitted at Info level, which routes to stderr; this PR moves them to stdout (`Level::Output`). This is an intentional behavior change, not a regression.

For most CLIs stderr is arguably the more correct target for prompts, since it keeps stdout clean for piped or captured output. Two practical reasons favor stdout here:
1. The container CLI ecosystem prompts on stdout: Docker (`cli.Out()`), containerd/nerdctl (`cmd.OutOrStdout()`), and Apple container all prompt there, so WSLC matches the tools its users expect.
2. Redirecting stderr to a log file (`command 2> file`) is common; with prompts on stderr that redirect would send the prompt to the file and leave the user at an apparently blank line. Keeping prompts on stdout avoids that.

The interactive-login E2E assertions were updated from stderr to stdout to match.

## Validation Steps Performed
- Unit tests, including the new input tests that were added, pass.
- E2E Registry tests pass, including the ones for interactive input testing.
- Manual invocation of deployed CLI of the registry login command to verify behavior on a real console.

### DIFF
--- a/src/windows/wslc/core/InputChannel.cpp
+++ b/src/windows/wslc/core/InputChannel.cpp
@@ -1,0 +1,106 @@
+/*++
+
+Copyright (c) Microsoft. All rights reserved.
+
+Module Name:
+
+    InputChannel.cpp
+
+Abstract:
+
+    Implementation of InputChannel.
+
+--*/
+#include "precomp.h"
+#include "InputChannel.h"
+
+#include <wil/resource.h>
+
+namespace wsl::windows::wslc {
+
+InputChannel::InputChannel(HANDLE consoleHandle, FILE* readFile) : m_file(readFile)
+{
+    DWORD mode = 0;
+    if (consoleHandle != INVALID_HANDLE_VALUE && consoleHandle != nullptr && GetConsoleMode(consoleHandle, &mode))
+    {
+        m_consoleHandle = consoleHandle;
+    }
+}
+
+InputChannel::InputChannel(FILE* readFile, bool interactiveOverride) :
+    m_file(readFile), m_interactiveOverride(interactiveOverride)
+{
+}
+
+bool InputChannel::IsInteractive() const noexcept
+{
+    if (m_consoleHandle != nullptr)
+    {
+        DWORD mode = 0;
+        return GetConsoleMode(m_consoleHandle, &mode) != FALSE;
+    }
+
+    return m_interactiveOverride;
+}
+
+std::optional<std::wstring> InputChannel::ReadLine(bool mask) const
+{
+    if (m_file == nullptr)
+    {
+        return std::nullopt;
+    }
+
+    // Disable console echo while reading when masking is requested and input is a
+    // real console. Armed only after echo is actually disabled so the restore is a
+    // no-op otherwise; runs on every exit path including exceptions.
+    DWORD previousMode = 0;
+    bool echoDisabled = false;
+    auto restoreEcho = wil::scope_exit([&]() {
+        if (echoDisabled)
+        {
+            SetConsoleMode(m_consoleHandle, previousMode);
+        }
+    });
+
+    if (mask && m_consoleHandle != nullptr && GetConsoleMode(m_consoleHandle, &previousMode))
+    {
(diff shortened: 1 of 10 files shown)


src/windows/wslc/core/InputChannel.h
```diff
@@ -0,0 +1,56 @@
+/*++
+
+Copyright (c) Microsoft. All rights reserved.
+
```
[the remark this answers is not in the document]
dkbennett: That one was probably a mistake and this repo generally uses the block style copyright lines.

src/windows/wslc/core/InputChannel.cpp
```diff
@@ -0,0 +1,106 @@
...
+        if (echoDisabled)
+        {
+            LOG_IF_WIN32_BOOL_FALSE(SetConsoleMode(m_consoleHandle, previousMode));
+        }
+    });
+
+    if (mask && m_consoleHandle != nullptr && GetConsoleMode(m_consoleHandle, &previousMode))
+    {
+        // Fail rather than echo a secret if the mode cannot be changed.
+        THROW_IF_WIN32_BOOL_FALSE(SetConsoleMode(m_consoleHandle, previousMode & ~ENABLE_ECHO_INPUT));
```
ggarzia-MSFT, line 68: Do we want to echo asterisks as the user types ( `Password: ******` ) instead of silencing echo entirely? Fully suppressed echo gives no feedback that keystrokes are registering.

dkbennett: No, for the following reasons:
- **Docker/CLI parity:** Matches `docker login`, which disables terminal echo (`prompt.DisableInputEcho` -> `term.DisableEcho`) and shows nothing. It also matches `sudo`, `ssh`, and `git` credential prompts, as well as Apple's `container` (`console.disableEcho()`).
- **Security:** Asterisks reveal the password length to anyone watching the screen; a silent prompt reveals nothing. This is exactly why `sudo` displays nothing as you type.
- **Complexity:** Asterisk masking cannot use a simple line read. It requires switching the console to raw, character-by-character mode and manually handling echo, Backspace/Delete, Ctrl-U, paste, and multi-byte input. That is significantly more surface area and edge cases for no real benefit.

ggarzia-MSFT: Docker does not do this and therefore we will not as well

ggarzia-MSFT (approved):

PR #41236, opened 2026-08-03
Add pipeline stage to produce DCAT submission files

florelis:

Adds a new stage to the release pipeline for generating the JSON files we need for submission to DCAT. The maintainers of DCAT requested that the schema of these files be kept private, so the script to produce them will live in an internal repo `wsl-release-scripts`. See internal PR 16367081.

## PR Checklist

## Validation Steps Performed

Executed the release pipeline (minus all the release steps, and using non-release configuration) and confirmed that the JSON files are produced and have expected content.

### DIFF
--- a/.pipelines/wsl-build-release-onebranch.yml
+++ b/.pipelines/wsl-build-release-onebranch.yml
@@ -33,6 +33,10 @@ resources:
       type: git
       name: OneBranch.Pipelines/GovernedTemplates
       ref: refs/heads/main
+    - repository: releaseScripts
+      type: git
+      name: Microsoft.WSL/wsl-release-scripts
+      ref: refs/tags/v1
 
 extends:
   template: v2/Microsoft.Official.yml@templates
@@ -74,6 +78,8 @@ extends:
         traceLoggingConfig: $(ReleaseTraceLoggingConfig)
         vsoOrg: microsoft
         vsoProject: Microsoft.WSL
+
+    - template: dcat-stage.yml@self
 
     - template: test-stage.yml@self
       parameters:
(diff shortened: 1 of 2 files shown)


.pipelines/wsl-build-release-onebranch.yml
```diff
@@ -75,6 +79,8 @@ extends:
         vsoOrg: microsoft
         vsoProject: Microsoft.WSL
 
+    - template: dcat-stage.yml@self
```
benhillis, line 82: should we move the dcat stage after the test stage, like the existing flight stage?

florelis: We could, but we don't need it yet. This is currently only creating the artifacts we need and not doing any publishing, so it doesn't hurt if it runs before the tests are complete.

I'll add better sequencing when I add the actual publishing. I'll have to deal with sequencing/triggering between multiple rings at that point, anyway.

.pipelines/wsl-build-release-onebranch.yml
```diff
@@ -33,6 +33,10 @@ resources:
       type: git
       name: OneBranch.Pipelines/GovernedTemplates
       ref: refs/heads/main
+    - repository: releaseScripts
+      type: git
+      name: Microsoft.WSL/wsl-release-scripts
+      ref: refs/tags/v1
```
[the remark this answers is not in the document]
florelis: The tag is already intended to be versioned and not be moved. It may change for a bit while setting it up for the first time, but after that it should be stable and protected by ADO policies.

OneBlue (approved):
LGTM. There's clearly more to this, we can merge this now so we can test this part during the next release

PR #41244, opened 2026-08-04
Preserve caller impersonation in SetSparse

benhillis:
## Summary
- capture the caller token before opening distribution state
- use an explicit token impersonation scope for the SetSparse operation
- avoid nested COM impersonation scopes


benhillis:
Superseded by #41245.

PR #41245, opened 2026-08-04
Use explicit tokens for user registry access

benhillis:
This change updates per-user WSL registry access to require an explicit caller token. The registry helper now uses that token through a nesting-safe impersonation scope, and callers reuse captured tokens when subsequent operations need the same identity. This makes caller identity management explicit and avoids hidden impersonation behavior inside the registry helper.


OneBlue (approved):
LGTM. Requiring the token to be passed feels like the right design for situations like these

PR #41250, opened 2026-08-05
Coordinate console state across concurrent WSL clients

shuaiyuanxx:
## Summary

- coordinate console-state ownership across overlapping WSL clients attached to the same console
- keep the configured mode active until the last client exits
- conditionally restore the original input/output modes and code pages without overwriting external changes
- add unit and process-level coverage for out-of-order client teardown

## Context

This is a follow-up to #41209 and addresses the overlapping-client intermediate-state concern raised in the review discussion. It is intentionally stacked on commit `0cd50878` from that PR so the diff contains only the coordination layer.

## Validation

- built Debug `msixinstallerpackage` and `wsltests`
- focused concurrent-client unit test: passed
- real two-process shared-console test: passed
- all `*ConsoleState*` tests: 5/5 passed
- repeated the overlap scenario in the active Claude Code console: passed twice
- `git diff --check`: passed

Crash and `TerminateProcess` recovery remain outside this change's scope; this coordinates normal cooperative client teardown.

### DIFF
--- a/src/windows/common/ConsoleState.cpp
+++ b/src/windows/common/ConsoleState.cpp
@@ -79,7 +79,7 @@ std::optional<DWORD> TryGetConsoleMode(_In_ HANDLE Handle)
 
 namespace wsl::windows::common {
 
-ConsoleState::ConsoleState(RestorePolicy Policy) : m_restorePolicy(Policy)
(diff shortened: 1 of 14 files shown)


shuaiyuanxx:
Closing because the follow-up commit has been pushed directly to #41209.

PR #41252, opened 2026-08-05
Add timeout to create process accepts

### LINKED ISSUE #41242: Session relay's vsock accept has no timeout; orphaned relays never exit, pinning vmbus rings until new sessions fail
## Windows Version

Microsoft Windows [Version 10.0.26200.7840]

## WSL Version

2.6.3.0 (bug still present on master by inspection)

## Are you using WSL 1 or WSL 2?

WSL 2

## Kernel Version

6.6.87.2-1

## Distro Version

Ubuntu 24.04

## Other Software

Docker Desktop (WSL integration); agent-heavy workload — 5–15 concurrent sessions, frequent short `wsl.exe -e` calls

## Repro Steps

No deterministic repro — the trigger is wslservice's per-session hvsocket connects partially failing (#40650 traced that to WSAENOBUFS on the Windows side). But the defect that turns a transient failure into a permanent leak is visible in the source, and easy to observe on any long-lived session-heavy machine:

1. Run a session-heavy workload for days.
2. `ps -e -o pid,ppid,comm | awk '$3=="Relay"'` — live relays are named `Relay(<childpid>)`, stuck ones stay exactly `Relay`.
3. Watch the count grow (ours: 2 → 37 → 62 → 44 distinct per day, 131 total over 6 days), each stuck relay logging `UtilAcceptVsock:244: Waiting for abnormally long accept` every ~30 s.
4. Eventually new sessions hang (`wsl.exe -e echo ok` times out) while existing sessions and `wsl --list --running` still work — the #40650 / #40795 end state.

## Expected Behavior

Relay accept is bounded — precedent exists, the session-leader accept path passes `SESSION_LEADER_ACCEPT_TIMEOUT_MS` (30 s, `common.h`) and fails fast with `accept4 failed 110`. And a relay whose session leader died exits: there's nothing left to relay.

## Actual Behavior

In `CreateProcessUtilityVm` (`init.cpp` ~1451 on tag 2.6.3, ~1481 on master) the relay thread calls `UtilAcceptVsock(ListenSocket, SocketAddress)` with no timeout → default `Timeout = -1` (`util.h`) → the poll loop in `util.cpp` retries forever. If wslservice completes only some of the session's `LX_INIT_UTILITY_VM_CREATE_PROCESS_SOCKET_COUNT` connects, the relay blocks there indefinitely:

- never exits, even long after its session leader is dead;
- ignores SIGTERM (`SigIgn: fffffffc7ff8fefe`, per #40650) — SIGKILL only;
- keeps its accepted hv_sock channels open, and each channel pins an order-7 (512 KB physically contiguous) `vmbus_alloc_ring` allocation (#40795). Enough stuck relays fragment guest memory until new channels can't allocate rings: new sessions fail, old ones work, and per the #40650 field reports `compact_memory` doesn't recover it (rings are unmovable).

So a transient host-side connect failure compounds into a permanent guest-side leak.

Terminal stage on our machine: after ~6 days of accumulation the VM was cleanly shut down with processes still running (orderly systemd shutdown in the guest journal, no triggering event of any kind in the Windows System log) — all overnight work killed. Evidence in our #40650 comment.

## Diagnostic Logs

Host-side ETW of the wslservice retry loop + guest journal data: https://github.com/microsoft/WSL/issues/40650#issuecomment-5184806063. More on request.

## Workaround

The GC from #40650 (SIGKILL processes matching: cmdline `/init`, comm exactly `Relay`, no children, no `/run/WSL/<pid>_interop`, session leader dead, age > 10 min). One portability note: on a systemd distro orphaned relays reparent to pid 2 (the WSL `/init`), not pid 1 — the ppid check needs to accept both.

## Related

#40650 (host-side trigger, no tunable limit), #40795 (memory mechanism), #41173 (interop busy-loop; sessions over sshd reportedly immune). Filing separately because the fix target differs: this is the guest-side lifecycle bug — unbounded accept, orphans never exiting — and it stays a bug even if the host-side limit gets raised.

chemwolf6922:

When creating a process, if the Windows side fails to create all 5/6 connections, the Linux side will block forever at accept. This leaks relay processes and hv socket connections.

This PR applies a 30s timeout to the accept calls to allow the process to exit on potential failure.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/41242

## Validation Steps Performed

Manually tested by injecting a failure after the first connect call:
image
Before the fix, the leaked processes live forever. After the fix, those processes exit after 30 seconds.

### DIFF
--- a/src/linux/init/init.cpp
+++ b/src/linux/init/init.cpp
@@ -1478,7 +1478,7 @@ Return Value:
 
     for (auto& Socket : Sockets)
     {
-        Socket.reset(UtilAcceptVsock(ListenSocket, SocketAddress));
+        Socket.reset(UtilAcceptVsock(ListenSocket, SocketAddress, SESSION_LEADER_ACCEPT_TIMEOUT_MS));
         if (Socket.get() < 0)
         {
             Result = -1;


OneBlue (approved):
For a long time there's been an issue where things get stuck missing a socket in this loop, and we've specifically left this on to help root cause that issue, but sadly we never went to the bottom of it. 

At this point it's probably worth fixing this. We can always revert if we see another repro of the hanging accept in this code path

PR #41257, opened 2026-08-05
Add nested virtualization support for WSLC containers

### LINKED ISSUE #40736: Add nested virtualization support for wslc containers

Workloads that need `/dev/kvm` inside a wslc container (`qemu-kvm`, libvirt-managed VMs, Firecracker, Android emulators, and similar) cannot run today, because the wslc utility VM never asks the hypervisor for virtualization extensions.

`HcsVirtualMachine.cpp` builds the utility VM that hosts every wslc container, and sets `Processor.Count`, `EnablePerfmonPmu`, and `EnablePerfmonLbr`, but does not set `Processor.ExposeVirtualizationExtensions`. There is also no field in `WSLCSessionSettings` or `WSLCFeatureFlags` for a caller to ask for it. So even on a Windows 11 host where nested virt is otherwise straightforward, a wslc container's guest sees no `vmx` / `svm` flag and `/dev/kvm` is absent.

The equivalent opt-in already exists for WSL2 distros via `nestedVirtualization=true` in `.wslconfig`. See #40735 for a Windows 10 regression in that path.

A new `WslcFeatureFlagsNestedVirtualization` bit in the existing `WSLCFeatureFlags` enum in `src/windows/service/inc/wslc.idl`, wired through `HcsVirtualMachine.cpp` to set `vmSettings.ComputeTopology.Processor.ExposeVirtualizationExtensions = true` when the flag is set. Callers of `IWSLCSessionManager::CreateSession` opt in via `WSLCSessionSettings::FeatureFlags`.

A draft of the service-side wiring is on [sirredbeard/WSL@win10-nested-virt-and-wslc](https://github.com/sirredbeard/WSL/tree/win10-nested-virt-and-wslc). It also includes the same Windows 10 perfmon guard from #40735, since Hyper-V on Windows 10 rejects partition creation when `ExposeVirtualizationExtensions` and `EnablePerfmonPmu` are both set.

- Enable virtualization extensions by default for every wslc utility VM. Too invasive. Nested virt costs memory and partition setup time, and most container workloads do not need it. Opt-in via feature flag matches the existing pattern for `WslcFeatureFlagsGPU` and `WslcFeatureFlagsVirtioFs`.
- A separate IDL field on `WSLCSessionSettings`. Would break the existing pattern for what is conceptually a feature toggle, and the `WSLCFeatureFlags` enum already exists for this.

**Additional context**

End-to-end testing on the wslc side will need a client surface (CLI flag or programmatic API) that sets the flag through to `CreateSession`. The draft branch only changes the service side. Happy to extend it if that would be useful.

benhillis:
## Summary
- add an opt-in WSLC session feature flag for nested virtualization
- expose virtualization extensions only when HCS reports support, otherwise fail session creation with a localized error
- mount the kernel modules VHD through a dedicated guest protocol message and load the vendor-specific KVM module using guest CPUID
- pass `/dev/kvm` into containers in nested-virtualization sessions
- share nested virtualization capability detection with the existing WSL2 path

## Validation
- Ran unit tests and inspection that the config option worked as expected

Fixes #40736


src/linux/init/WSLCInit.cpp
```diff
@@ -669,6 +675,43 @@ void HandleMessageImpl(
...
+    }
+    else if (strcmp(vendor.data(), "AuthenticAMD") == 0)
+    {
+        module = "kvm_amd";
+    }
+
+    THROW_ERRNO_IF(ENOTSUP, module == nullptr);
+
+    const char* argv[] = {"/sbin/modprobe", module, nullptr};
+    THROW_LAST_ERROR_IF(UtilCreateProcessAndWait("/sbin/modprobe", argv) < 0);
```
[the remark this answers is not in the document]
benhillis: Fixed in c6f17179 by reporting an explicit EIO when modprobe fails instead of relying on errno.

src/linux/init/WSLCInit.cpp
```diff
@@ -669,6 +675,43 @@ void HandleMessageImpl(
     Transaction.Send(Response);
 }
 
+void LoadKvmModule()
```
benhillis, line 678: This isn't my favorite - still thinking about better ways to know which kvm module to load. We could just try both and ignore the failure...

src/shared/inc/lxinitshared.h
```diff
@@ -1678,6 +1679,21 @@ struct WSLC_MOUNT_VIRTIOFS
     PRETTY_PRINT(FIELD(Header), STRING_FIELD(SourceIndex), STRING_FIELD(DestinationIndex), STRING_FIELD(TypeIndex), STRING_FIELD(OptionsIndex), STRING_FIELD(ChildNameIndex));
 };
 
+struct WSLC_MOUNT_MODULES
```
benhillis, line 1682: I think splitting this out is good, especially with the upcoming change we have that switches the modules VHD layout substantially.

src/linux/init/WSLCInit.cpp
OneBlue: Do we actually need to load it ourselves ? We don't it in WSL and things seem to work without issues. Maybe qemu is smart enough to do this for us ?

benhillis: Good question, but I think we do because it is a loadable module that doesn't load by default, and users in a container will not be able to load it themselves inside their container, so it needs to be present before we launch the container.

But I might be missing an easier solution.

src/linux/init/WSLCInit.cpp
```diff
@@ -669,6 +675,43 @@ void HandleMessageImpl(
...
+    if (strcmp(vendor.data(), "GenuineIntel") == 0)
+    {
+        module = "kvm_intel";
+    }
+    else if (strcmp(vendor.data(), "AuthenticAMD") == 0)
+    {
+        module = "kvm_amd";
+    }
+
+    THROW_ERRNO_IF(ENOTSUP, module == nullptr);
```
benhillis: An error message that includes vendor might be useful.

PR #41267, opened 2026-08-06
Use unified kernel artifacts VHD layout

benhillis:
## Summary
- update to the unified kernel package and package `artifacts.vhd` instead of `modules.vhd`
- load kernel modules from the versioned `<release>/modules` directory for both standard WSL distributions and WSLC sessions
- retain support for legacy flat module VHDs in the standard WSL boot path
- update development, packaging, and test paths for the renamed VHD

Kernel headers and perf exposure are intentionally deferred to a follow-up PR.

## Testing
- full x64 Debug build
- `bin\x64\Debug\test.bat /name:*KernelModules*` (1 passed)

### DIFF
--- a/src/linux/init/main.cpp
+++ b/src/linux/init/main.cpp
@@ -3223,7 +3223,10 @@ try
         // N.B. The VHD is mounted as read-only but with a writable overlayfs layer. The modules
         //      directory must be writable for tools like depmod to work.
         //
-
+        // N.B. The artifacts VHD nests the modules under <release>/modules.
+        //      Older module-only VHDs place the modules tree at the filesystem root; fall back to that
+        //      layout when the nested modules directory is not present.
+        //
         if (EarlyConfig->KernelModulesDeviceId != UINT_MAX)
         {
             THROW_LAST_ERROR_IF(
@@ -3232,9 +3235,33 @@ try
 
             utsname UnameBuffer{};
             THROW_LAST_ERROR_IF(uname(&UnameBuffer) < 0);
+            const std::string Release{UnameBuffer.release};
 
-            std::string Target = std::format("{}/{}", KERNEL_MODULES_PATH, UnameBuffer.release);
-            THROW_LAST_ERROR_IF(UtilMountOverlayFs(Target.c_str(), KERNEL_MODULES_VHD_PATH, (MS_NOATIME | MS_NOSUID | MS_NODEV)) < 0);
+            const std::string ArtifactsBase = std::format("{}/{}", KERNEL_MODULES_VHD_PATH, Release);
+            const std::string NestedModules = ArtifactsBase + "/modules";
+
+            std::error_code Error{};
+            const bool NestedLayout = std::filesystem::is_directory(NestedModules, Error);
+            const std::string ModulesLower = NestedLayout ? NestedModules : std::string{KERNEL_MODULES_VHD_PATH};
+            const bool LegacyLayout = !NestedLayout && std::filesystem::is_regular_file(ModulesLower + "/modules.dep", Error);
+
+            //
+            // A valid artifacts VHD nests the tree under <release>/modules; a legacy module-only VHD
+            // places it at the root.
+            //
+            if (LegacyLayout)
+            {
+                LOG_WARNING(
+                    "kernel modules VHD uses the legacy flat layout; support for the legacy modules VHD format will be "
+                    "removed in a future version");
+            }
+            else if (!NestedLayout)
+            {
+                LOG_WARNING("kernel modules VHD does not contain modules for {}", Release);
+            }
+
+            std::string Target = std::format("{}/{}", KERNEL_MODULES_PATH, Release);
+            THROW_LAST_ERROR_IF(UtilMountOverlayFs(Target.c_str(), ModulesLower.c_str(), (MS_NOATIME | MS_NOSUID | MS_NODEV)) < 0);
 
             const std::string KernelModulesList = wsl::shared::string::FromSpan(Buffer, EarlyConfig->KernelModulesListOffset);
             for (const auto& Module : wsl::shared::string::Split(KernelModulesList, ','))
--- a/test/windows/UnitTests.cpp
+++ b/test/windows/UnitTests.cpp
@@ -2923,8 +2923,9 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
         // Ensure the kernel modules folder is mounted correctly.
         std::wstring command = std::format(
             L"mount | grep -iF 'none on /usr/lib/modules/{} type overlay "
-            L"(rw,nosuid,nodev,noatime,lowerdir=/modules,upperdir=/lib/modules/{}/rw/upper,workdir=/lib/modules/{}/rw/"
+            L"(rw,nosuid,nodev,noatime,lowerdir=/modules/{}/modules,upperdir=/lib/modules/{}/rw/upper,workdir=/lib/modules/{}/rw/"
             L"work,uuid=on)'",
+            kernelVersion,
             kernelVersion,
             kernelVersion,
             kernelVersion);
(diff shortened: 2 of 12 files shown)


src/linux/init/config.cpp
```diff
@@ -1125,6 +1125,94 @@ Return Value:
...
+                            {
+                                LOG_ERROR("lstat({}) failed {}", linkPath, errno);
+                                createLink = false;
+                            }
+
+                            if (createLink && (symlink(headersRoot.c_str(), linkPath.c_str()) < 0))
+                            {
+                                LOG_ERROR("symlink({}, {}) failed {}", headersRoot, linkPath, errno);
+                            }
+                        }
```
benhillis: this whole thing seems overly complicated

src/linux/init/config.cpp
```diff
@@ -170,6 +172,32 @@ class RemoveMountAndEnvironmentOnScopeExit
...
+{
+    auto tempMount = RemoveMountAndEnvironmentOnScopeExit(MountEnvironmentName);
+    const char* target = tempMount ? getenv(PathEnvironmentName) : nullptr;
+    if (target == nullptr)
+    {
+        return;
+    }
+
+    const std::string targetPath{target};
+    unsetenv(PathEnvironmentName);
```
benhillis, line 192: nit: move unsetenv first

src/linux/init/config.cpp
```diff
@@ -1664,6 +1731,25 @@ Return Value:
         {
             ConfigAppendToPath(Environment, LXSS_LIB_PATH);
         }
+
+        //
+        // Add the kernel-matched perf tools to the $PATH variable and point perf at its helper
+        // scripts since it is built with a prefix that does not match where it is mounted.
+        //
+
+        if (Config.KernelPerfPath.has_value())
```
benhillis: I'm torn on if this is needed or not.

benhillis: I guess it's ok and makes perf more usable / discoverable.

src/linux/init/config.cpp
```diff
@@ -1110,20 +1138,59 @@ Return Value:
...
+            return;
+        }
+
+        //
+        // Point /lib/modules/<release>/build at the kernel headers, replacing any entry that the
+        // distro may have created so that it can't shadow the headers matching the running kernel.
+        //
+
+        const std::string linkPath = kernelModulesPath + "/build";
+        if ((unlink(linkPath.c_str()) < 0) && (errno != ENOENT))
```
OneBlue: If this folder exists and isn't a symlink `unlink` would fail since it's not empty. We could do a recursive delete but that feels a bit dangerous. 

Maybe the easiest solution would to bind mind `linkPath` to `target` ? That way we don't need to delete anything

src/linux/init/main.cpp
```diff
@@ -3170,9 +3204,30 @@ try
...
+            const std::string Release{UnameBuffer.release};
+
+            const std::string ArtifactsBase = std::format("{}/{}", KERNEL_MODULES_VHD_PATH, Release);
+            const std::string NestedModules = ArtifactsBase + "/modules";
+
+            struct stat StatBuffer{};
+            const bool NestedLayout = (stat(NestedModules.c_str(), &StatBuffer) == 0) && S_ISDIR(StatBuffer.st_mode);
+            const std::string ModulesLower = NestedLayout ? NestedModules : std::string{KERNEL_MODULES_VHD_PATH};
+            const bool LegacyLayout =
+                !NestedLayout && (stat((ModulesLower + "/modules.dep").c_str(), &StatBuffer) == 0) && S_ISREG(StatBuffer.st_mode);
```
OneBlue: nit: this could be simplified a bit via [std::filesystem::is_regular_file](https://en.cppreference.com/cpp/filesystem/is_regular_file)

OneBlue: And `std::filesystem::is_directory`  for the above check

src/linux/init/util.cpp
```diff
@@ -1738,6 +1738,10 @@ Return Value:
 int UtilMountFile(const char* Source, const char* Destination)
 try
 {
+    struct stat sourceInfo{};
+    THROW_LAST_ERROR_IF(stat(Source, &sourceInfo) < 0);
+    THROW_ERRNO_IF(EINVAL, !S_ISREG(sourceInfo.st_mode));
```
OneBlue: Do we need this check ? I would expect the mount call to fail otherwise

test/windows/UnitTests.cpp
```diff
@@ -2980,6 +2981,87 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
...
+        // via $PATH, with PERF_EXEC_PATH pointing at its helper scripts.
+        //
+        // N.B. The cleanup is registered before the distro's file system is modified so that a
+        //      failure can't leave a perf binary behind, which would be shadowed by the bind mount
+        //      on the next boot and break subsequent runs.
+        auto perfCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
+            LxsstuLaunchWsl(L"umount /usr/bin/perf 2>/dev/null; rm -f /usr/bin/perf", nullptr, nullptr, nullptr, nullptr);
+        });
+
+        VERIFY_ARE_EQUAL(LxsstuLaunchWsl(L"umount /usr/bin/perf 2>/dev/null; rm -f /usr/bin/perf", nullptr, nullptr, nullptr, nullptr), 0u);
```
OneBlue: What's the reason for unmounting & deleting perf here ? I'd expect that we'd correctly override the distro's perf executable in the test distro

OneBlue: Actually it doesn't look like perf is in the test distro at all so we can probably skip this entirely

src/linux/init/util.cpp
```diff
@@ -1738,6 +1738,11 @@ Return Value:
 int UtilMountFile(const char* Source, const char* Destination)
 try
 {
+    //
```
benhillis: remove comment

src/linux/init/config.cpp
```diff
@@ -1110,20 +1138,73 @@ Return Value:
...
+        // Point /lib/modules/<release>/build at the kernel headers, replacing any entry that the
+        // distro may have created so that it can't shadow the headers matching the running kernel.
+        //
+        // N.B. A directory can't be replaced with a symlink (and removing it would mean a recursive
+        //      delete), so bind mount the headers over it instead.
+        //
+
+        const std::string linkPath = kernelModulesPath + "/build";
+        struct stat existing{};
+        if ((lstat(linkPath.c_str(), &existing) == 0) && S_ISDIR(existing.st_mode))
```
OneBlue: Do we still need this check ? `UtilMount` creates the folder if it doesn't exist

src/linux/init/config.cpp
```diff
@@ -1110,20 +1138,73 @@ Return Value:
...
-                tempMount.MoveMount(target);
+                LOG_ERROR("UtilMount({}, {}) failed {}", target, linkPath, errno);
             }
+
+            return;
         }
-    }
-    CATCH_LOG()
+
+        if ((unlink(linkPath.c_str()) < 0) && (errno != ENOENT))
```
OneBlue: I don't think we need this `unlink()` anymore

src/linux/init/util.cpp
benhillis: Checking...

src/linux/init/config.cpp
benhillis: I think this was to handle if there was a regular file or symlink in the expected place, let me confirm.

src/linux/init/config.cpp
benhillis: I think you're right.

src/linux/init/util.cpp
benhillis: Yes. `mount()` would fail, but only after `UtilMountFile()` may remove a destination symlink or create an empty destination file. Validating the source first avoids those side effects and enforces that this helper only mounts regular files.

src/linux/init/config.cpp
benhillis: I took another look and it is still needed. This branch handles an existing file or symlink which must be removed before creating the replacement symlink. It is only redundant when the path does not exist.

src/linux/init/config.cpp
benhillis: Yes. The check selects behavior based on the type of the existing entry, rather than only ensuring that the path exists. Existing directories are bind-mounted; files and symlinks are replaced with a symlink. UtilMount() also treats mkdir() returning EEXIST as success without verifying that the entry is a directory.

OneBlue (approved):
LGTM. 

Reading this I also realized that the whole `ModulesMountPoint` logic isn't useful anymore since we always mount the modules after chroot, but that's outside the scope of this change. 
I'll clean this up once merged

PR #41271, opened 2026-08-06
Normalize NULL char* to <null> in Linux logging

chemwolf6922:

The Linux logging uses std::format to construct the log string. And std::format expects valid C strings for char* / const char* types. There is one confirmed instance of passing NULL to the log macros in WSL:
When mounting modules.vhd and system.vhd. The options paramter is NULL. If the mount fails, it's passed directly to the log macro. Which could result in a segfault.

This PR normalizes all NULL char* / const char* to "<null>" in the log implementation to avoid passing NULL to std::format.

## PR Checklist

## Validation Steps Performed

Tested locally by adding temporary log NULL code. 
Before the fix:
```
> wsl
Operation aborted
Error code: Wsl/Service/CreateInstance/CreateVm/E_ABORT
```
After the fix:
In dmesg:
```
[    0.331626] WSL (1 - ): NULL_LOG_TEST: value=<null>
```

### DIFF
--- a/src/linux/init/common.h
+++ b/src/linux/init/common.h
@@ -45,6 +45,7 @@ Abstract:
 #include <poll.h>
 #include <utility>
 #include <format>
+#include <type_traits>
 #include <new>
 #include <thread>
 #include <vector>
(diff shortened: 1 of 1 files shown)


OneBlue (approved):
LGTM. As a nit it feels a bit off to have a potential situation where two threads will end up with the same pointer if they have null strings, but this pointer should never be written so I think this is OK

PR #41290, opened 2026-08-07
Make paths absolute before weakly_canonical and add regression tests

ggarzia-MSFT:
weakly_canonical only returns an absolute path when a leading element of the input exists: it builds its result from the root path plus the longest leading sequence it can canonicalize, so a relative path whose first element is missing is returned unchanged. Callers that pass a not-yet-created path therefore produced a relative result, which the service rejects with "Path is not absolute".

This affected `wslc build --iidfile <relative>` and `--secret src=<relative>` when the target did not exist, plus the distribution path and virtiofs share path helpers in the service.

Wrap each call in std::filesystem::absolute so the result is absolute regardless of whether the path exists.

Existing coverage could not catch this: every --iidfile e2e test passed an absolute path, and the relative --secret test used a file that already existed, which makes weakly_canonical succeed on its own. Add two tests that fail without the fix - a parser unit test using a relative src that names a missing file, and an e2e test running `build --iidfile image.id` from the context directory.

## PR Checklist


src/windows/service/exe/LxssUserSession.cpp
```diff
@@ -3847,7 +3847,7 @@ void LxssUserSessionImpl::_ValidateDistributionNameAndPathNotInUse(
 
     if (Path != nullptr)
     {
-        canonicalPath = std::filesystem::weakly_canonical(Path, error);
+        canonicalPath = std::filesystem::weakly_canonical(std::filesystem::absolute(Path, error), error);
```
OneBlue, line 3850: Passing the same reference to both calls feels a bit risky here (if the first call fails, we could path a default path to the second, which might lead to weird behavior). 

Given how often we use this, maybe the best approach would be to have a helper ? Something like: 

```
std::pair<std::filesystem::path, std::error_code GetCanonicalPathNoThrow(...);
```

And for convenience we could have a throwing version of it: 

```
std::filesystem::path GetCanonicalPath(input); 
```

That would throw a user error like "Failed to resolve path {}"

test/windows/FilesystemUnitTests.cpp
```diff
@@ -0,0 +1,193 @@
...
+    // A relative path naming a file that does not exist must still resolve to an absolute path.
+    // This is the case std::filesystem::weakly_canonical cannot handle on its own: it builds its result
+    // from the longest leading sequence of elements that exist, so a bare missing file name has nothing
+    // to canonicalize and is returned unchanged.
+    TEST_METHOD(GetCanonicalPath_RelativeMissingPathIsMadeAbsolute)
+    {
+        ScopedTempDirectory directory;
+        const auto name = UniqueMissingName(directory.Path());
+
+        ScopedCurrentDirectory scopedDirectory(directory.Path());
```
OneBlue: nit: Do we actually need to change the current directory for this test (and the below one) ? If I'm understanding this correctly, this test should work without having to do that

OneBlue: Actually I'm not following that test. We create a new temp directory, cd into it, and then we try to canonicalize its local relative path, but that path isn't validate anymore since we changed directory.

OneBlue: If the objective is to cover the case of a missing relative path, I think we can simplify this something like: 

```
std::filesystem::path nonExistingDir = "does-not-exist";
VERIFY_IS_FALSE(std::filesystem::exists(nonExistingDir));

const auto result = GetCanonicalPath(name);
VERIFY_IS_TRUE(result.is_absolute());
```

OneBlue (approved):

PR #41300, opened 2026-08-10
Localization change from build: 154382310

OneBlue:


benhillis (approved):
Stray CR fixup pushed in a4bd69a6; formatting & localization checks are green. Translation content otherwise unchanged from the build output.

craigloewen-msft (approved):

PR #41302, opened 2026-08-10
Enable Unity build for wslclib and enable CI unity builds

ggarzia-MSFT:

Extends unity builds from `wsltests` to `wslclib`, generalizes the CMake option that controls them, and enables unity builds in CI at batch size 4.

`WSL_TEST_UNITY_BATCH_SIZE` (test-only, default `2`) is renamed to `WSL_UNITY_BATCH_SIZE`, now applies to both `wslclib` and `wsltests`, and defaults to `0` (off) so local builds behave as they did before unity was introduced. CI opts in explicitly with `-DWSL_UNITY_BATCH_SIZE=4`.

Enabling unity on `wslclib` surfaced several latent name-lookup conflicts in `src/windows/wslc` that only appear when translation units are combined; those are fixed here.

## PR Checklist

No new user-facing strings. `UserConfig.cmake.sample` documents the renamed option.

**Build configuration**

- `CMakeLists.txt` — renames the option and adds `WSL_UNITY_BATCH_SIZE_DEFAULT`, set to `0`. **This changes the default from `2` to `0`**, so a plain `cmake .` no longer uses unity builds.
- `src/windows/wslc/CMakeLists.txt` — `wslclib` now honors the option (`UNITY_BUILD_MODE BATCH`).
- `.pipelines/build-job.yml` — adds `-DWSL_UNITY_BATCH_SIZE=4` to the per-platform Release configure, the step that builds `wslc` and `wsltests` on x64 and arm64.

**Unity-build conflict fixes in `src/windows/wslc`**

Correctness fixes independent of grouping, not workarounds tied to a batch size:

- `core/Reporter.h`, `services/BuildImageCallback.cpp`, `services/ImageProgressCallback.cpp` — two `.cpp` files each defined their own file-scope `constexpr int c_fallbackConsoleWidth = 79;`. Merged into one `inline constexpr` in `Reporter.h`.
- `core/TableOutput.h` — qualifies `details::WrapText`. Three distinct `namespace details` blocks exist under `wslc`, so the unqualified name is ambiguous once those headers share a TU.
- `tasks/SessionTasks.cpp` — drops a redundant `using namespace wsl::shared::string;` that overlapped with `wsl::windows::common::string` (both declare `WideToMultiByte`).
- `services/SessionModel.h` — qualifies `wil::com_ptr<::IUnknown>`. `WslcSdkWinRtTests.cpp` brings `winrt::Windows::Foundation` into scope, making the unqualified name ambiguous (C2872) when both land in one unity TU.

Each fix qualifies the use site rather than relocating a `using namespace` directive, since moving a directive changes name lookup and can silently redirect overload resolution.

**Benchmarking scripts**

`tools/benchmarking/benchmark-build.ps1` and `run-unity-batch-sweep.ps1` reproduce the batch-size measurements (N full + M incremental builds, emitting a CSV). `.gitignore` excludes the results directory.

**Known trade-off**

With `UNITY_BUILD_MODE BATCH`, group membership depends on source order, and `wslclib` sources come from `file(GLOB_RECURSE)` — so adding or renaming a file reshuffles every downstream batch and can reintroduce a conflict without touching any existing file. This happened once during development.

## Validation Steps Performed

- Built `wslc` and `wsltests` (debug, x64) with `-DWSL_UNITY_BATCH_SIZE=4` — both succeed. `wsltests` produces 21 unity TUs from 81 sources (≈ batch 4).
- Built with `-DWSL_UNITY_BATCH_SIZE=0` to confirm the default path still compiles every file separately.
- `clang-format` 19.1.5 `--dry-run -Werror` clean on all changed C++ files.
- `validate-line-endings.py` logic run across all 839 matched files — no mismatches, including the CMake files that validator now covers.


src/windows/wslc/core/Terminal.h
```diff
@@ -31,6 +31,11 @@ Module Name:
 
 namespace wsl::windows::wslc {
 
+// Fallback width for progress displays when the console width can't be queried. This
+// value already includes the autowrap guard (visible width minus one) so a wrapped line
+// can't corrupt cursor-based rendering.
+inline constexpr int c_fallbackConsoleWidth = 79;
```
dkbennett, line 37: Good refactor!

test/windows/CMakeLists.txt
```diff
@@ -23,11 +23,11 @@ add_compile_definitions(INLINE_TEST_METHOD_MARKUP)
 
 add_library(wsltests SHARED ${SOURCES} ${HEADERS})
 
-if (DEFINED WSL_TEST_UNITY_BATCH_SIZE AND WSL_TEST_UNITY_BATCH_SIZE GREATER 0)
+if (DEFINED WSL_UNITY_BATCH_SIZE AND WSL_UNITY_BATCH_SIZE GREATER 0)
     set_target_properties(wsltests PROPERTIES
         UNITY_BUILD ON
         UNITY_BUILD_MODE BATCH
-        UNITY_BUILD_BATCH_SIZE ${WSL_TEST_UNITY_BATCH_SIZE})
+        UNITY_BUILD_BATCH_SIZE ${WSL_UNITY_BATCH_SIZE})
```
dkbennett, line 30: maybe want to change this to WSLC or WSLCCLI_UNITY_BATCH_SIZE just so we can independently tune the different projects for optimal batch size. as the project changes, optimal batch sizes might change too.

.gitignore
```diff
@@ -78,4 +78,5 @@ deploy-log.txt
 test-output*.txt
 test-results.txt
 testfile.txt
-output/
\ No newline at end of file
+output/
+tools/benchmarking/benchmark-results/
```
dkbennett, line 82: I like the addition of these tools so we can do our own benchmarking and periodically check things out for changes that might affect batch size tuning in the future.

dkbennett (approved):
Some minor comments.

test/windows/CMakeLists.txt
ggarzia-MSFT: Will do this in a future PR

OneBlue (approved):

PR #41303, opened 2026-08-10
CLI: Add --ipv6 to network create

dkbennett:
Add Docker-compatible `--ipv6` support to `wslc network create`.

## PR Checklist

- Plumb `--ipv6` through the CLI, session API, and Docker network-create request.
- Report Docker's `EnableIPv6` state through `network inspect`.
- Preserve IPv6 state when recovering managed networks.
- Add an end-to-end test for outbound IPv6 connectivity.
- Share Docker session test helpers across e2e tests.

## Validation Steps Performed
- Full x64 Debug build.
- `WSLCE2ENetworkCreateTests`: 17/17 passed.
- Affected warning test passed.


src/windows/service/inc/wslc.idl
```diff
@@ -643,6 +643,7 @@ typedef struct _WSLCNetworkOptions
     [unique] LPCSTR Subnet;
     [unique] LPCSTR Gateway;
     [unique] LPCSTR IpRange;
+    BOOL EnableIPv6;
```
JohnMcPMS, line 654: This should probably be exposed to the SDK as well.

ranm-msft (approved):
Followed the flag end to end rather than just reading the CLI change: arg parsing -> `CreateNetworkOptions` -> the IDL `EnableIPv6` -> the docker `CreateNetwork` body (picked up by the type macro) -> `NetworkEntry` and `RecoverExistingNetworks`, so a recovered network keeps the same view as a freshly created one. `inspect` already deserialized `EnableIPv6`, so the e2e assertion is exercising the real path, and asserting `false` in the default-create test is a good touch.

Looks good to me.

PR #41307, opened 2026-08-10
add --progress to wslc build

ggarzia-MSFT:

Implements `wslc build --progress`, which was previously defined but commented out in the
argument table. Supports the same modes as `docker build --progress`: `auto`, `tty`, `plain`,
`quiet`, and `rawjson`. `auto` resolves to `tty` when progress output is an interactive VT
console and `plain` otherwise.

No existing behavior changes when `--progress` is omitted: the resolved default (`auto`) picks
the same renderer that was previously hard-coded.

## PR Checklist

**Argument plumbing**
- Enabled the `Progress` entry in `ArgumentDefinitions.h` and registered it on `wslc build`.
- Added `models::ProgressMode` (`ContainerModel.h`) and `GetProgressModeFromString` /
  `ValidateProgressModeFromString`. Values are validated up front by `Argument::Validate`, so an
  invalid mode fails client-side before any build starts.
- Values are lowercase-only, matching docker and the existing `--format` option.

**Rendering**
- `tty` — existing in-place renderer: a scrolling log window plus per-entry pull progress that
  updates in place using cursor movement.
- `plain` — append-only. Emits no cursor movement, erases, or color, so output is identical
  whether it goes to a console or a redirected stream.
- `quiet` — suppresses live progress but still retains log output, so a failed build replays its
  logs on error.
- `rawjson` — forwards docker's raw BuildKit progress JSON verbatim. This required a new
  `WSLCBuildImageFlagsRawJson` flag in `WSLCShared.idl` so the server bypasses its own parsing and
  formatting.

**Known intentional divergence from docker**

BuildKit's plain printer (`util/progress/progressui/printer.go`) appends throttled download status
lines (`#5 sha256:... 1.05MB / 31.4MB 0.2s`), reprinting a status only after it advances 5% or 5
seconds elapse. This PR omits pull progress entirely in `plain` instead, since without in-place
updates those lines are mostly noise. This is documented in a comment at the call site. Everything
else about plain output matches: append-only, no escape sequences.

Note for anyone revisiting this: the server's pull-progress message (`WSLCSession.cpp`) is the only
progress message emitted **without** a trailing newline, because it was written for in-place
rendering. It would need one before it could be appended.

## Validation Steps Performed

**Automated tests added**

- 14 parser cases (`ParserTestCases.h`) covering `--progress=value` and `--progress value` forms,
  all five valid modes, case-sensitivity, unrecognized values, an empty value, and a missing value
  at end of input. The `Build` argument set mirrors `ImageBuildCommand::GetArguments()` so the
  parser tests exercise the real option set.
- 5 unit tests (`WSLCCLIBuildImageCallbackUnitTests.cpp`) driving `BuildImageCallback` against a
  VT-enabled capture terminal: `plain` emits no escape sequences, `plain` output is byte-identical
  on a console and when redirected, `quiet` emits nothing, and `rawjson` is not wrapped in cursor
  control. A `tty` test asserting escape sequences **are** emitted is included as a control, so the
  plain assertions cannot pass vacuously.
- 4 E2E tests (`WSLCE2EImageBuildTests.cpp`): invalid mode is rejected client-side, `rawjson`
  contains BuildKit `"vertexes"` objects, `plain` contains no `0x1b`, and `quiet` emits no progress
  on success.

**Negative controls**

The plain-mode fix was verified by reintroducing the bug and re-running: both
`PlainOnVtConsole_EmitsNoEscapeSequences` and `PlainMatchesRedirectedOutput` failed, while the
`tty`, `quiet`, and `rawjson` tests still passed. Restoring the fix returned all 5 to passing.

This mattered: the E2E `plain` test passes with *and* without the fix, because `RunWslc` redirects
output and so never reaches the VT-console path where the redraw occurred. The unit tests are what
actually cover it.

**Manual / build validation**

- Full `cmake --build . -- -m` clean.
- `clang-format` 19.1.5 clean across all files changed on the branch.
- Test runs: 4/4 `*Progress*` E2E, 5/5 `WSLCCLIBuildImageCallbackUnitTests`,
  18/18 `WSLCCLIArgumentUnitTests`, 27/27 `WSLCCLIParserUnitTests`.
- 4/4 pre-existing `*IidFile*` E2E tests still pass, confirming the rebase onto the argument-caching
  and `Reporter`→`Terminal` refactors did not regress the neighbouring build options.

### DIFF
--- a/src/windows/wslc/arguments/SpecParsing.cpp
+++ b/src/windows/wslc/arguments/SpecParsing.cpp
@@ -682,6 +682,38 @@ models::PullPolicy GetPullPolicyFromString(const std::wstring& input, const std:
     throw ArgumentException(Localization::WSLCCLI_InvalidPullPolicyError(argName, input, supportedValues));
 }
 
+models::ProgressMode GetProgressModeFromString(const std::wstring& input, const std::wstring& argName)
+{
+    if (IsEqual(input, L"auto"))
+    {
+        return models::ProgressMode::Auto;
+    }
+    else if (IsEqual(input, L"tty"))
+    {
+        return models::ProgressMode::Tty;
+    }
+    else if (IsEqual(input, L"plain"))
+    {
+        return models::ProgressMode::Plain;
+    }
+    else if (IsEqual(input, L"quiet"))
+    {
+        return models::ProgressMode::Quiet;
+    }
+    else if (IsEqual(input, L"rawjson"))
+    {
+        return models::ProgressMode::RawJson;
+    }
+    else
+    {
+        throw ArgumentException(std::format(
+            L"Invalid {} value: {} is not a recognized progress type. Supported progress types are: auto, tty, plain, "
+            L"quiet, rawjson.",
+            argName,
+            input));
+    }
+}
+
 models::InspectType GetInspectTypeFromString(const std::wstring& input, const std::wstring& argName)
 {
     if (IsEqual(input, L"image"))
--- a/src/windows/wslc/arguments/SpecParsing.cpp
+++ b/src/windows/wslc/arguments/SpecParsing.cpp
@@ -684,34 +684,34 @@ models::PullPolicy GetPullPolicyFromString(const std::wstring& input, const std:
 
 models::ProgressMode GetProgressModeFromString(const std::wstring& input, const std::wstring& argName)
 {
-    if (IsEqual(input, L"auto"))
+    static constexpr std::pair<std::wstring_view, models::ProgressMode> c_progressModes[] = {
+        {L"auto", models::ProgressMode::Auto},
+        {L"tty", models::ProgressMode::Tty},
+        {L"plain", models::ProgressMode::Plain},
+        {L"quiet", models::ProgressMode::Quiet},
+        {L"rawjson", models::ProgressMode::RawJson},
+    };
+
+    for (const auto& [name, mode] : c_progressModes)
     {
-        return models::ProgressMode::Auto;
+        if (IsEqual(input, name))
+        {
+            return mode;
+        }
     }
-    else if (IsEqual(input, L"tty"))
+
+    std::wstring supportedValues;
+    for (const auto& progressMode : c_progressModes)
     {
-        return models::ProgressMode::Tty;
+        if (!supportedValues.empty())
+        {
+            supportedValues += L", ";
+        }
+
+        supportedValues += progressMode.first;
     }
-    else if (IsEqual(input, L"plain"))
-    {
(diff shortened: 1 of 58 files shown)


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1413,6 +1414,27 @@ try
         }
     };
 
+    // rawjson mode: forward each complete JSON object docker writes to stderr verbatim (as newline-delimited
```
OneBlue: I'm not sure that we want to implement this mode in our current state. Passing through the docker json means that we if we make engine changes, that json might break. 

My recommendation for now would be not to add support for this mode for now, and if at some point we want to implement it, we should have a parser that can re-emit the raw json with a schema that we control

OneBlue (dismissed):

OneBlue (approved):

src/windows/wslc/arguments/SpecParsing.cpp
```diff
@@ -682,6 +682,37 @@ models::PullPolicy GetPullPolicyFromString(const std::wstring& input, const std:
     throw ArgumentException(Localization::WSLCCLI_InvalidPullPolicyError(argName, input, supportedValues));
 }
 
+models::ProgressMode GetProgressModeFromString(const std::wstring& input, const std::wstring& argName)
+{
+    static constexpr std::pair<std::wstring_view, models::ProgressMode> c_progressModes[] = {
```
dkbennett, line 687: Nice, I like the use of the enum here as with the other similar conversions of string to a named type.

src/windows/wslc/arguments/SpecParsing.cpp
```diff
@@ -682,6 +682,37 @@ models::PullPolicy GetPullPolicyFromString(const std::wstring& input, const std:
...
+    }
+
+    std::wstring supportedValues;
+    for (const auto& progressMode : c_progressModes)
+    {
+        if (!supportedValues.empty())
+        {
+            supportedValues += L", ";
+        }
+
```
dkbennett, line 709: now that I realize we've done the supported values from enum pattern a few times, I wonder if this can be a generalized helper (something to do later, not important for this PR).

dkbennett (approved):
Looks good, progress routing to Info is correct and this follows all the patterns closely.

src/windows/wslc/arguments/SpecParsing.cpp
ggarzia-MSFT: I will look into this and make a separate PR if we can easily do something like this

PR #41311, opened 2026-08-11
mountutil: skip empty tokens in mount option parser

Eamon2009:
mountutil: skip empty option tokens and clean up includes

Empty tokens in comma-separated mount options (e.g. from "foo,,bar")
were being propagated into the mount data string as empty entries,
which can confuse downstream filesystem option parsers. Skip them
during parsing.

Also remove the duplicate <sys/types.h> and unused <functional>
and <dirent.h> includes.


benhillis (approved):
Looks good to me, thanks for the submission!

PR #41318, opened 2026-08-11
Add --filter support to wslc network list

beena352:
Adds --filter to wslc network list so users can filter results by label, driver, or name, the same way container list and image list already work. When no filter is passed, behavior is unchanged (cache-only, no Docker call). When a filter is passed, we query Docker with the user's filters plus the WSLC managed-network label so results stay scoped to WSLC-managed networks.

## PR Checklist

Mirrors the existing shape used by container list --filter and image list --filter.

## Validation Steps Performed
- New unit tests in WSLCTests.cpp cover label filters (key-only, key=value, AND), driver hit/miss, the explicit managed-label being idempotent, and null filter key/value returning E_POINTER
- New E2E tests in WSLCE2ENetworkListTests.cpp cover malformed value, invalid key, driver, label, name substring match, and NDJSON stdout being exactly empty on zero matches.

### DIFF
--- a/src/windows/wslc/commands/NetworkListCommand.cpp
+++ b/src/windows/wslc/commands/NetworkListCommand.cpp
@@ -27,6 +27,7 @@ namespace wsl::windows::wslc {
 std::vector<Argument> NetworkListCommand::GetArguments() const
 {
     return {
+        Argument::Create(ArgType::Filter, false, Limit::Unlimited),
         Argument::Create(ArgType::Format),
         Argument::Create(ArgType::Quiet, false, std::nullopt, Localization::WSLCCLI_NetworkListQuietArgDesc()),
     };
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -3011,15 +3011,6 @@ try
             dockerNetworks = m_runtime.Docker().ListNetworks(filters);
         }
         CATCH_AND_THROW_DOCKER_USER_ERROR("Failed to list networks");
-
-        for (const auto& dockerNetwork : dockerNetworks)
-        {
-            // Only report networks that we manage.
-            if (!m_networks.contains(dockerNetwork.Name))
-            {
-                WSL_LOG("ListedUnknownNetwork", TraceLoggingValue(dockerNetwork.Name.c_str(), "NetworkName"));
-            }
-        }
     }
 
     auto output = wil::make_unique_cotaskmem<WSLCNetworkInformation[]>(m_networks.size());
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -3047,6 +3047,7 @@ try
         auto it = m_networks.find(dockerNetwork.Name);
         if (it == m_networks.end())
         {
+            WSL_LOG("ListedUnknownNetwork", TraceLoggingValue(dockerNetwork.Name.c_str(), "NetworkName"));
             continue;
         }
 
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -2979,7 +2979,7 @@ try
 }
 CATCH_RETURN();
 
-HRESULT WSLCSession::ListNetworks(WSLCNetworkInformation** Networks, ULONG* Count)
+HRESULT WSLCSession::ListNetworks(const WSLCFilter* Filters, ULONG FiltersCount, WSLCNetworkInformation** Networks, ULONG* Count)
 try
 {
     WSLCExecutionContext context(this);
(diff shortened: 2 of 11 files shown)


src/windows/wslc/commands/NetworkListCommand.cpp
```diff
@@ -27,6 +27,7 @@ namespace wsl::windows::wslc {
 std::vector<Argument> NetworkListCommand::GetArguments() const
 {
     return {
+        Argument::Create(ArgType::Filter, false, Limit::Unlimited),
```
dkbennett, line 30: We need a new argtype for this one "NetworkFilter" that is almost a duplicate of Filter because ArgType::Filter has no alias for '-f', and docker's network list does alias '-f' to filter.

docker network ls: --filter has the -f alias.
docker network prune: --filter has no short alias (because -f means --force in this command)

The resolution is to copy the Filter ArgType and make a new one like "FilterAliased" or "ListFilter" or some other enum value that has "f" as the alias and use that one for list and use the regular Filter for prune.

This is a seam in the argument model where the idenity of an ArgType is its name, alias, kind, and conversion type and if any of those need to change it must be a new ArgType. I may rework this later to allow alias override, but its not urgent and adding another enum value is an easy workaround.

dkbennett (commented):
Looks good overall, just need a new ArgType definition that is otherwise identical but has an alias for 'f' to be consistent with Docker CLI.

Edit: I was wrong, list has -f, it is prune that is using the aliased filter incorrectly since in Docker -f is for force.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2990,22 +2990,69 @@ try
...
-    for (const auto& [name, entry] : m_networks)
+    for (const auto& dockerNetwork : dockerNetworks)
     {
-        THROW_HR_IF(E_UNEXPECTED, strcpy_s(output[index].Name, name.c_str()) != 0);
-        THROW_HR_IF(E_UNEXPECTED, strcpy_s(output[index].Id, entry.Id.c_str()) != 0);
-        THROW_HR_IF(E_UNEXPECTED, strcpy_s(output[index].Driver, entry.Driver.c_str()) != 0);
+        auto it = m_networks.find(dockerNetwork.Name);
+        if (it == m_networks.end())
+        {
+            continue;
```
OneBlue, line 3023: nit: This is probably worth an ETL event since this would be unexpected

src/windows/wslc/commands/NetworkListCommand.cpp
dkbennett: And i realized I have this backwards. It's prune that has the bug, list is fine. Prune has a bug to fix (to create one without the alias). Updating to approve this.

dkbennett (dismissed):

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2990,22 +2990,69 @@ try
     *Networks = nullptr;
     *Count = 0;
 
+    auto filters = wsl::windows::common::wslutil::ParseKeyMultiValuePairs(Filters, FiltersCount);
+
     auto lock = AcquireLease();
+
+    if (filters.empty())
+    {
+        std::lock_guard networksLock(m_networksLock);
```
OneBlue: I recommend factoring the with & without filter blocks. We could do something like this: 

```
std::lock_guard networksLock(m_networksLock);

std::optional<std::vector<docker_schema::Network> dockerNetworks> selectedNetworks;
if (!filters.empty())
{
    try
    {
        selectedNetworks = m_runtime.Docker().ListNetworks(filters);
    }
    CATCH_AND_THROW_DOCKER_USER_ERROR("Failed to list networks");
}

for (const auto& [name, entry] : m_networks)
{
    if (selectedNetworks.has_value())
    {
           // Skip if the network was not returned by the docker query
           [...]
           continue;
    }
   [...]
}

File: src/windows/wslcsession/WSLCSession.cpp [hunk 4]
```diff
@@ -2990,9 +2990,46 @@ try
...
+    if (filtered)
+    {
+        // Scope the filtered query to WSLC-managed networks.
+        filters["label"].push_back(WSLCNetworkManagedLabel);
+    }
+
     auto lock = AcquireLease();
+    if (filtered)
+    {
+        THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_runtime.HasDocker());
```
OneBlue: What's the reason for this check ? This should be impossible given that we have a runtime lease at this point

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2990,9 +2990,46 @@ try
...
+    }
+
     std::lock_guard networksLock(m_networksLock);
 
+    std::optional<std::unordered_set<std::string>> dockerNames;
```
OneBlue: To simplify things, I recommend directly storing the `std::vector<docker_schema::Network>` here, and having a simple loop iterating on `m_networks` and optionally filtering by the networks returned by docker. 

Doing this will require moving the `if (m_networks.empty())` into a `if index == 0` after that loop, but I think that's worth the simplification

src/windows/wslcsession/WSLCSession.cpp
beena352: Good catch, the lease already guarantees it. Removing.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2990,25 +2990,59 @@ try
...
+    if (filtered)
     {
-        return S_OK;
+        try
+        {
+            dockerNetworks = m_runtime.Docker().ListNetworks(filters);
+        }
+        CATCH_AND_THROW_DOCKER_USER_ERROR("Failed to list networks");
+
+        for (const auto& dockerNetwork : dockerNetworks)
```
OneBlue: Now that we check the networks in the "main" loop, we can remove that one. I don't think it's worth keeping just for logging

OneBlue (approved):

Issue #41173, opened 2026-07-25
ProcessInteropMessages spins at 100% CPU forever when an interop control socket reaches EOF


racterub:
### Windows Version

```
Microsoft Windows [Version 10.0.26200.8037]
```

### WSL Version

```
2.7.10.0
```

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

```
6.18.33.2-microsoft-standard-WSL2
```

### Distro Version

```
Ubuntu 24.04.1 LTS
```

### Other Software

Windows Terminal (the affected `wsl.exe` was the relay for a single long-lived
tab). No third-party software is implicated — the defect is in WSL's own
interop path and is reachable from any terminal.

### Related issues

**#40651** reports the same symptom — `wsl.exe` spinning at 100% because a
`ReadFile` loop has no `bytesRead == 0` branch on the synchronous-success path.
It is closed, but not resolved: the `microsoft-github-policy-service` bot
auto-closed it after seven days of author inactivity, with the note *"If you're
still experiencing this issue please re-file it as a new issue."* This report is
that re-file, and it differs in three ways that I think matter:

1. **Different file.** #40651 points at `src/windows/common/relay.cpp`. A
   maintainer [replied](https://github.com/microsoft/WSL/issues/40651#issuecomment-4561130265)
   that `ScopedMultiRelay` *"is indeed missing the read length check. But it's
   only used in set version. So it's likely not the cause of the busy loop you
   are seeing."* I agree — the loop in this report is
   `ProcessInteropMessages` in `src/windows/common/interop.cpp`, which is a
   different function with a different exit structure.
2. **The function is pinned.** The original reporter noted *"We cannot pin the
   spin to one specific function."* Here the stack, the thread name, and the
   handle all resolve to a single loop, detailed below.
3. **The requested dumps exist.** The same comment asked for *"several dumps
   seconds apart with `procdump -ma <pid>`"*. I captured three, 20 seconds
   apart, and the zero-byte read is measured in them rather than inferred.

### Repro Steps

**I do not have a deterministic repro.** The trigger is a narrow race, but the
race itself is specific and reviewable, so I've described it precisely rather
than guessing at steps.

**What produced it in practice:**

1. Open a WSL2 session in a terminal and leave it open for days
2. Use it normally — in particular, run Windows binaries from inside Linux
[...]

PR #41319, opened 2026-08-12
Fix potential spin in interop message relay

chemwolf6922:

The original logic ignores the success result of the initial `ReadFile` call. If the connection is a socket and its close is not captured by `GetOverlappedResult`. It could cause this `ReadFile` call to return success with bytes read == 0. This will put this loop into a spin.

This PR refactors the loop with `io::MultiHandleWait`.

## PR Checklist

- [x] **Closes:** https://github.com/microsoft/WSL/issues/41173

## Validation Steps Performed

This issue is triggered by a race condition. I'm not able to repro the original issue on my machine.
The normal function was tested locally.

### DIFF
--- a/src/windows/common/interop.cpp
+++ b/src/windows/common/interop.cpp
@@ -421,62 +421,56 @@ ProcessInteropMessages(_In_ HANDLE MessageHandle, _Inout_ CreateProcessResult* R
     DWORD ExitCode = 1;
     for (;;)
     {
-        DWORD BytesRead;
+        DWORD BytesRead{};
         LX_INIT_WINDOW_SIZE_CHANGED WindowSizeMessage;
         bool Success = ReadFile(MessageHandle, &WindowSizeMessage, sizeof(WindowSizeMessage), &BytesRead, &Overlapped);
         if (!Success)
         {
             const auto LastError = GetLastError();
-            if ((LastError == ERROR_BROKEN_PIPE) || (LastError == ERROR_HANDLE_EOF))
+            if ((LastError != ERROR_BROKEN_PIPE) && (LastError != ERROR_HANDLE_EOF))
             {
-                if (WI_IsFlagClear(Result->Flags, LX_INIT_CREATE_PROCESS_RESULT_FLAG_GUI_APPLICATION))
+                THROW_LAST_ERROR_IF(LastError != ERROR_IO_PENDING);
+
+                auto CancelIo = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&] {
+                    CancelIoEx(MessageHandle, &Overlapped);
+                    GetOverlappedResult(MessageHandle, &Overlapped, &BytesRead, TRUE);
+                });
+
+                const DWORD WaitStatus = WaitForMultipleObjects(RTL_NUMBER_OF(WaitHandles), WaitHandles, FALSE, INFINITE);
+                if (WaitStatus == WAIT_OBJECT_0)
                 {
-                    THROW_IF_WIN32_BOOL_FALSE(TerminateProcess(Result->Process.get(), 1));
+                    Success = GetOverlappedResult(MessageHandle, &Overlapped, &BytesRead, FALSE);
+                    CancelIo.release();
                 }
-
-                break;
-            }
-
-            THROW_LAST_ERROR_IF(LastError != ERROR_IO_PENDING);
-
-            auto CancelIo = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&] {
-                CancelIoEx(MessageHandle, &Overlapped);
-                GetOverlappedResult(MessageHandle, &Overlapped, &BytesRead, TRUE);
-            });
-
-            const DWORD WaitStatus = WaitForMultipleObjects(RTL_NUMBER_OF(WaitHandles), WaitHandles, FALSE, INFINITE);
-            if (WaitStatus == WAIT_OBJECT_0)
-            {
-                Success = GetOverlappedResult(MessageHandle, &Overlapped, &BytesRead, FALSE);
-                CancelIo.release();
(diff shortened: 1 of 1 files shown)


src/windows/common/interop.cpp
```diff
@@ -419,64 +419,68 @@ ProcessInteropMessages(_In_ HANDLE MessageHandle, _Inout_ CreateProcessResult* R
...
     DWORD ExitCode = 1;
+    DWORD Offset = 0;
+    LX_INIT_WINDOW_SIZE_CHANGED WindowSizeMessage;
     for (;;)
     {
-        DWORD BytesRead;
-        LX_INIT_WINDOW_SIZE_CHANGED WindowSizeMessage;
-        bool Success = ReadFile(MessageHandle, &WindowSizeMessage, sizeof(WindowSizeMessage), &BytesRead, &Overlapped);
+        DWORD BytesRead{};
+        bool Success = ReadFile(
```
OneBlue: nit: Since we're refactoring this, this could be a good opportunity to port this to using a `MultiHandleWait`  (via `ReadHandle` and `EventHandle`), which will deal with all the overlapped IO logic for us

src/windows/common/interop.cpp
```diff
@@ -419,64 +419,68 @@ ProcessInteropMessages(_In_ HANDLE MessageHandle, _Inout_ CreateProcessResult* R
...
+            break;
+        }
+
+        Offset += BytesRead;
+        if (Offset != sizeof(WindowSizeMessage))
+        {
+            continue;
         }
+
+        WI_ASSERT(WindowSizeMessage.Header.MessageType == LxInitMessageWindowSizeChanged);
```
OneBlue: nit: This is pre-existing code, but probably not a great idea to assert here, since a malicious client could send broken message on purpose. A THROW_HR if we find an incorrect message type feels like the right kind of error here

OneBlue (dismissed):
Change looks good, minor comments. Feel free to address in a followup

chemwolf6922:
Hi @OneBlue . I refactored the implementation with `io::MultiHandleWait` following your suggestion. Could you help take another look? Thanks.

src/windows/common/interop.cpp
```diff
@@ -408,78 +409,70 @@ std::string FormatCommandLine(gsl::span<gsl::byte> CommandLineData, USHORT Comma
...
-                    break;
+                    return;
                 }
 
-                WI_ASSERT((BytesRead == sizeof(WindowSizeMessage)) && (WindowSizeMessage.Header.MessageType == LxInitMessageWindowSizeChanged));
+                auto remaining = input;
+                while (!remaining.empty())
+                {
+                    const size_t bytesToCopy = (std::min)(remaining.size(), sizeof(pendingMessage) - pendingSize);
+                    std::copy_n(remaining.data(), bytesToCopy, reinterpret_cast<char*>(&pendingMessage) + pendingSize);
```
OneBlue: I think we can skip the copy into `pendingMessage` here. If we have enough bytes for a full message, we could just access it via: 

```
const auto& message = gslhelper::get_struct<LX_INIT_WINDOW_SIZE_CHANGED>(remaining); 
```

src/windows/common/interop.cpp
```diff
@@ -408,78 +409,70 @@ std::string FormatCommandLine(gsl::span<gsl::byte> CommandLineData, USHORT Comma
...
+                            THROW_IF_WIN32_BOOL_FALSE(TerminateProcess(Result->Process.get(), 1));
+                        }
                     }
 
-                    break;
+                    return;
                 }
 
-                WI_ASSERT((BytesRead == sizeof(WindowSizeMessage)) && (WindowSizeMessage.Header.MessageType == LxInitMessageWindowSizeChanged));
+                auto remaining = input;
```
OneBlue, line 451: `remaining` should be stored outside of this lambda, since it's possible for a read to be partial, and if that happens, we want the bytes to be stored until the next call. 

In an ideal world, we'd reuse `ReadSocketMessageHandle`, but it currently only supports sockets. Maybe we can add a flag to add support for non-socket handles, but that's outside the scope of this change

src/windows/common/interop.cpp
chemwolf6922: Will there be alignment concerns?

src/windows/common/interop.cpp
chemwolf6922: I guess this needs to combine with the copy change. 
However, `remaining` is a reference to the `ReadHandle`'s reusable internal buffer. The remainder needs to be copied out. When the next input come, we also need to stitch the remainder to the head of the new input in order to access the struct directly.
IMO this may not worth the change.

src/windows/common/interop.cpp
```diff
@@ -408,78 +409,75 @@ std::string FormatCommandLine(gsl::span<gsl::byte> CommandLineData, USHORT Comma
...
+                            THROW_IF_WIN32_BOOL_FALSE(TerminateProcess(Result->Process.get(), 1));
+                        }
                     }
 
-                    break;
+                    return;
                 }
 
-                WI_ASSERT((BytesRead == sizeof(WindowSizeMessage)) && (WindowSizeMessage.Header.MessageType == LxInitMessageWindowSizeChanged));
+                std::vector<char> stitchedInput;
```
OneBlue, line 450: nit: This could be simplified by doing: 

```
pending.insert(pending.end(), input.begin(), input.end);
```

And then only using `pending` to read messages from

OneBlue (approved):
LGTM, feel free to address comment in a followup

src/windows/common/interop.cpp
chemwolf6922: This could introduce alignment issues. For example, if only one byte is copied from input. The next struct in input will be misaligned.

PR #41326, opened 2026-08-12
Fix localhost relay hang on peer disconnect and ioctl interface name handling

Eamon2009:
## Summary
Fixed the relay worker thread hanging when a peer disconnects. The loop only checked for POLLIN, so POLLHUP and POLLERR from a closed connection were ignored, causing an infinite spin at 100% CPU. Now it handles all three events and shuts down cleanly.
Changed the relay exit condition from || to && so the thread waits for both directions to close before exiting. This prevents dropping buffered data when one side shuts down early.
Also made the interface name null-termination explicit in the ioctl handler. The code relied on struct zero-initialization, which works but is fragile. Now it copies sizeof-1 bytes and sets the last byte to '\0' directly.

also #41342


Eamon2009:

src/linux/init/localhost.cpp
```diff
@@ -517,6 +516,7 @@ int RunPortTracker(int Argc, char** Argv)
 
         auto& ifRequest = *reinterpret_cast<ifreq*>(ifreqMemory->data());
         memcpy(request.InterfaceName, ifRequest.ifr_ifrn.ifrn_name, sizeof(request.InterfaceName));
+        request.InterfaceName[sizeof(request.InterfaceName) - 1] = '\0';
```
OneBlue, line 520: nit: Instead of doing this, we could update the previous line to: 

`memcpy(request.InterfaceName, ifRequest.ifr_ifrn.ifrn_name, sizeof(request.InterfaceName) - 1);`, since the structure is zero-initialized

OneBlue (commented):
LGTM, one minor comment

src/linux/init/localhost.cpp
OneBlue: The last iteration has the memcpy change, so we can remove the explicit '\0` assignment now

PR #41333, opened 2026-08-12
Simplify VHD ownership handling during distribution moves

benhillis:

Simplifies VHD ownership handling when moving a distribution. The duplicated caller token now uses the caller's user SID as its default owner, and the move and rollback operations run under that token.

## PR Checklist

Cross-volume `MoveFileExW` operations create a new file using the impersonation token's default owner. This change sets the duplicated caller token's default owner to its user SID before moving the VHD, keeping the resulting file accessible to that user.

All filesystem operations, including rollback, use the modified caller token. This removes the need to reopen the destination and restore its owner after the move. Same-volume moves remain renames and preserve the existing owner.

## Validation Steps Performed

- `cmake --build . -- -m`
- `bin\x64\Debug\test.bat -f /name:*MoveVhdWithAdminOwner` — passed (1/1)

### DIFF
--- a/src/windows/service/exe/LxssUserSession.cpp
+++ b/src/windows/service/exe/LxssUserSession.cpp
@@ -929,77 +929,47 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
 
     RETURN_HR_IF(E_NOTIMPL, WI_IsFlagClear(distro.Flags, LXSS_DISTRO_FLAGS_VM_MODE));
 
-    // Build the final vhd path.
-    std::filesystem::path newVhdPath = Location;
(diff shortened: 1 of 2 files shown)


OneBlue (approved):
Change LGTM, one small change in behavior is that now we're reverting the VHD owner to the current token on failure, instead of the original VHD's owner. 

I could imagine some situations where that could lead to issues, especially if there are some custom ACL's, but I don't think that's necessarily a blocker. That could be fixed in a followup

benhillis:
> Change LGTM, one small change in behavior is that now we're reverting the VHD owner to the current token on failure, instead of the original VHD's owner.
> 
> I could imagine some situations where that could lead to issues, especially if there are some custom ACL's, but I don't think that's necessarily a blocker. That could be fixed in a followup

Yeah that's true, I think this is an improvement over the existing so I'll look at fixing the ACL stuff in a follow-up.

PR #41335, opened 2026-08-12
Return non vhd volumes in container inspect

OneBlue:

This change adds logic to return guest volumes in the container inspect's output. For now, the source path is set to an empty string since we don't want to expose VM paths, but this can be revisited later if needed 

## PR Checklist

### DIFF
--- a/src/windows/inc/wslc_schema.h
+++ b/src/windows/inc/wslc_schema.h
@@ -32,7 +32,7 @@ struct InspectMount
 {
     std::string Type;
     std::string Name;
-    std::optional<std::string> Source;
+    std::string Source;
     std::string Destination;
     bool ReadWrite{};
 
--- a/src/windows/inc/wslc_schema.h
+++ b/src/windows/inc/wslc_schema.h
@@ -30,13 +30,13 @@ struct InspectPortBinding
 
 struct InspectMount
 {
-    // TODO: Support different mount types (plan9/VHD) when VHD volumes are implemented.
     std::string Type;
-    std::string Source;
+    std::string Name;
+    std::optional<std::string> Source;
     std::string Destination;
     bool ReadWrite{};
 
-    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(InspectMount, Type, Source, Destination, ReadWrite);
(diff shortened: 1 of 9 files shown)


test/windows/wslc/e2e/WSLCE2EContainerRemoveTests.cpp
```diff
@@ -249,32 +243,21 @@ class WSLCE2EContainerRemoveTests
...
-        std::vector<std::wstring> added;
-        for (const auto& name : ListVolumeNames())
-        {
-            if (std::find(before.begin(), before.end(), name) == before.end())
-            {
-                added.push_back(name);
-            }
-        }
+        const auto inspect = InspectContainer(containerName);
+        const auto mount = std::ranges::find_if(inspect.Mounts, [](const auto& entry) { return entry.Type == "volume"; });
```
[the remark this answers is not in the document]
OneBlue: By design, this expects a single volume mount since that's what the tests do

src/windows/inc/wslc_schema.h
```diff
@@ -30,13 +30,13 @@ struct InspectPortBinding
 
 struct InspectMount
 {
-    // TODO: Support different mount types (plan9/VHD) when VHD volumes are implemented.
     std::string Type;
-    std::string Source;
+    std::string Name;
+    std::optional<std::string> Source;
```
dkbennett: Docker engine always defines the mount source a string, never nullable, and tmpfs serializes as empty string, so this breaks the json contract for consumers.

dkbennett (commented):
Normally I like the semantic clarity of making source optional so intent is clear, in this case it is a public schema where Source is already expected to be a string and perhaps more importantly, Docker itself uses empty string when no source applies. Nullable is cleaner conceptually but introduces a compatibility risk. If we want our json output to be consumable by tooling that expects Docker-like output, this is a change that could break that tooling.

So in the interest of Docker alignment & compat, I would recommend leaving Source as non-nullable string and leaving it empty for anonymous volumes.

src/windows/inc/wslc_schema.h
OneBlue: Good call ! Switched to an empty string

dkbennett (approved):

benhillis (approved):

PR #41337, opened 2026-08-12
CLI: Add --mount support for create and run

dkbennett:

Adds Docker-compatible `--mount` support to `wslc container run` and `wslc container create`, and consolidates all CLI mount forms into one typed mount pipeline.

`--mount`, `--volume`, and `--tmpfs` now produce `mount::Spec` values during argument validation. A single mount collection performs validation and duplicate-destination detection before `WSLCContainerLauncher::AddMount` translates each specification for the backend.

## PR Checklist

## Implementation

- Adds the transport-neutral `mount::Spec` model and Docker-compatible parsers in `src/windows/common`.
- Supports bind, named-volume, anonymous-volume, and tmpfs mounts.
- Supports Docker aliases, CSV quoting, Go-compatible boolean values, read-only mounts, default volume type, and tmpfs size/mode options.
- Allows colons in `--mount` targets because that syntax uses comma-separated key/value fields.
- Uses the shared `ParseStorageSize` implementation for tmpfs sizes.
- Preserves opaque legacy `--tmpfs` options such as `uid`, `gid`, `nr_inodes`, `nosuid`, `nodev`, and `noexec`.
- Keeps `AddVolume`, `AddNamedVolume`, and `AddTmpfs` as compatibility wrappers that delegate to `AddMount`.
- Preserves the Docker bind-source distinction:
  - `--mount type=bind` requires the source to exist.
  - `--volume host:target` creates a missing host directory.
- Preserves the distinction between structured tmpfs mounts and legacy `--tmpfs` entries in the Docker request and inspect output.
- Reports parser, unsupported-feature, and validation failures with localized errors.
- Validates option families so type-specific options cannot be used with another mount type.
- Detects normalized duplicate destinations across `--mount`, `--volume`, and `--tmpfs` using one collection.
- Adds inspect mappings for bind, named-volume, anonymous-volume, and tmpfs mounts.
- Excludes anonymous volumes from named-volume availability checks while retaining generated-volume cleanup behavior.

## Why WSLC Parses `--mount`

Docker CLI converts the textual `--mount` grammar into typed mount objects before calling Docker Engine. WSLC must do the same because its launcher and backend boundary are structured, and bind mounts require additional Windows-host and utility-VM preparation.

WSLC also needs to reject options that its current transport cannot represent faithfully. Parsing and capability validation before the backend request prevents silent semantic loss and produces consistent CLI errors.

## Validation

- Full x64 Debug build.
- `WSLCCLIMountParserUnitTests`: passed.
- `WSLCCLITmpfsParserUnitTests`: 2/2 passed.
- `WSLCVolumeMountUnitTests`: passed through the common volume parser.
- Legacy `--volume` create E2E suite: 9/9 passed.
- Legacy and structured tmpfs create E2E suite: 7/7 passed.
- Anonymous and named `--mount type=volume` E2E coverage passed.
- Cross-option duplicate-destination E2E coverage passed for `--mount`, `--volume`, and `--tmpfs`.
- Bind missing-source E2E coverage verifies that `--mount` fails without creating a directory while `--volume` succeeds and creates it.
- Tmpfs runtime coverage verifies the requested size and mode, not only mount usability.


src/windows/wslc/services/ContainerService.cpp
```diff
@@ -141,6 +143,24 @@ static wsl::windows::common::RunningWSLCContainer CreateInternal(Terminal& termi
         }
     }
 
+    for (const auto& mountSpec : options.Mounts)
+    {
+        switch (mountSpec.MountType)
+        {
+        case mount::Type::Bind:
+            containerLauncher.AddVolume(mountSpec.Source, mountSpec.Target, mountSpec.ReadOnly);
```
ptrivedi: routing --mount type=bind through AddVolume might cause the backend to create the host directory when it does not exist. Docker --mount functionality fails when there's a missing bind source. So a typo could silently create and mount an empty directory. Let's preserve the distinction from --volume. Add an e2e test confirming that the directory is not created.

Example:
With --mount, the following is expected to fail:
wslc container run --mount "type=bind,source=C:\nonexistent-dir,target=/data" alpine

With --volume, the following is expected to succeed and create the nonexistent-dir:
wslc container run --volume "C:\nonexistent-dir:/data" alpine

since in this case both are routed through AddVolume, we will lose this distinction... which docker intentionally distinguishes.

src/windows/common/MountSpecParsing.cpp
```diff
@@ -0,0 +1,685 @@
...
+void ValidateMountSpec(const Spec& mount)
+{
+    if (mount.Target.empty())
+    {
+        ThrowInvalid(L"target is required");
+    }
+
+    if (mount.Target.find(':') != std::string::npos)
+    {
+        ThrowInvalid(L"target paths containing ':' are not supported.");
```
ptrivedi: While this is correct for the "--volume", "-v", and "--tmpfs", docker does allow : in target names for --mount, because --mount uses comma-separated key-value fields. 

For example,
--mount type=bind,source=<whatever>,target=/path:mntdir
--mount type=volume,source=<whatever>,target=/path:voldir
--mount type=tmpfs,target=/path:tmpfs
are all supported.

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -221,6 +220,21 @@ void Argument::Validate(ArgMap& execArgs) const
...
+        CacheConverted<ArgType::Mount>(execArgs, m_name, [](const std::wstring& value, const std::wstring&) {
+            try
+            {
+                auto mountSpec = mount::ParseDockerMountString(value);
+                mount::ValidateMountSpec(mountSpec);
+                return mountSpec;
+            }
+            catch (const mount::ValidationException& ex)
+            {
+                throw ArgumentException(Localization::WSLCCLI_InvalidMountError(value, ex.Reason()));
```
ptrivedi, line 277: we have hardcoded english error strings in MountParsingSpec.cpp, so the text in ex.Reason() will not be translated, correct? could the parser return type and values and then the localized string can be built here?

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -141,6 +143,24 @@ static wsl::windows::common::RunningWSLCContainer CreateInternal(Terminal& termi
...
+
+        case mount::Type::Volume:
+            containerLauncher.AddNamedVolume(string::WideToMultiByte(mountSpec.Source), mountSpec.Target, mountSpec.ReadOnly);
+            break;
+
+        case mount::Type::Tmpfs:
+            containerLauncher.AddTmpfs(mountSpec.Target, mount::FormatTmpfsOptions(mountSpec));
+            break;
+        }
+    }
```
ptrivedi, line 156: this will require us to make correspoinding changes to container inspect side as well, correct? we should check if proper coverage for inspect is present with the new mount types we are supporting

src/windows/common/MountSpecParsing.cpp
```diff
@@ -0,0 +1,685 @@
...
+        if (!std::filesystem::path(mount.Source).is_absolute())
+        {
+            ThrowInvalid(L"bind source path must be absolute");
+        }
+        break;
+
+    case Type::Volume:
+        if (mount.Source.empty())
+        {
+            ThrowInvalid(L"anonymous volume mounts are not supported.");
```
ptrivedi: would be good to mention this gap in --mount help description

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -333,6 +333,47 @@ TmpfsMount TmpfsMount::Parse(const std::string& value)
     return result;
 }
 
+void ValidateUniqueMountDestinations(const ContainerOptions& options)
+{
+    try
+    {
+        mount::ValidateMountCollection(options.Mounts);
```
ptrivedi, line 265: mounts are validated here, then another set is allocated for dests and processed again. i think for --volume and --tmpfs, parsing happens here and in ContainerService. it would be good to convert all the flags into one mount collection and then validation, dupes, etc. can be done on that one collection

src/windows/common/MountSpecParsing.cpp
```diff
@@ -0,0 +1,685 @@
...
+        Supported,
+        Unsupported,
+        ValueDependent,
+    };
+
+    struct FieldDefinition
+    {
+        std::wstring_view Name;
+        Field Id;
+        Family OptionFamily;
```
ptrivedi, line 71: i don't see the OptionFamily being used. is it an anticipatory placeholder or something?

src/windows/wslc/arguments/ArgumentValidation.cpp
```diff
@@ -221,6 +220,21 @@ void Argument::Validate(ArgMap& execArgs) const
...
+            }
+        });
```
ptrivedi, line 279: also, every ValidationException here will result in an "Invalid argument". Could some of the ones like 'volume-nocopy', 'volume-opt' etc. be reported as 'not supported' by wslc?

test/windows/wslc/e2e/WSLCE2EContainerRunTests.cpp
```diff
@@ -1038,6 +1038,81 @@ class WSLCE2EContainerRunTests
...
+        WriteTestFileContent(EnvTestFile1, "WSLC Mount Bind Test");
+
+        const auto hostDirectory = EnvTestFile1.parent_path();
+        const auto fileName = EnvTestFile1.filename().wstring();
+        auto result = RunWslc(std::format(
+            L"container run --rm --mount \"type=bind,source={},target=/data,readonly\" {} cat /data/{}",
+            hostDirectory.wstring(),
+            DebianImage.NameAndTag(),
+            fileName));
+        result.Verify({.Stdout = L"WSLC Mount Bind Test", .Stderr = L"", .ExitCode = 0});
```
ptrivedi, line 1095: based on previous comments around missing source path for bind, the tests should cover that

test/windows/wslc/e2e/WSLCE2EContainerRunTests.cpp
```diff
@@ -1038,6 +1038,81 @@ class WSLCE2EContainerRunTests
...
+        result = RunWslc(std::format(
+            L"container run --rm --mount type=volume,source={},target=/data,readonly {} sh -c \"echo changed > /data/value\"",
+            WslcVolumeName,
+            DebianImage.NameAndTag()));
+        result.Verify({.Stdout = L"", .Stderr = L"sh: 1: cannot create /data/value: Read-only file system\n", .ExitCode = 2});
+
+        result = RunWslc(std::format(
+            L"container run --rm --mount type=volume,source={},target=/data {} cat /data/value", WslcVolumeName, DebianImage.NameAndTag()));
+        result.Verify({.Stdout = L"original", .Stderr = L"", .ExitCode = 0});
+    }
```
ptrivedi, line 1129: tests to validate corresponding inspect output

test/windows/wslc/e2e/WSLCE2EContainerRunTests.cpp
```diff
@@ -1038,6 +1038,81 @@ class WSLCE2EContainerRunTests
...
+
```
ptrivedi, line 1130: based on previous comments, tests around colon-containing destinations paths

test/windows/wslc/e2e/WSLCE2EContainerRunTests.cpp
```diff
@@ -1038,6 +1038,81 @@ class WSLCE2EContainerRunTests
...
     }
 
+    WSLC_TEST_METHOD(WSLCE2E_Container_Run_Mount_Tmpfs_Success)
+    {
+        auto result = RunWslc(std::format(
+            L"container run --rm --mount type=tmpfs,target=/wslc-tmpfs {} sh -c \"echo -n 'tmpfs_test' > /wslc-tmpfs/data && cat "
+            L"/wslc-tmpfs/data\"",
+            DebianImage.NameAndTag()));
+        result.Verify({.Stdout = L"tmpfs_test", .Stderr = L"", .ExitCode = 0});
+    }
```
ptrivedi, line 1082: we should add/augment e2e tests to verify that tmpfs size/mode are applied correctly, instead of only testing that the mount is usable

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -141,6 +143,24 @@ static wsl::windows::common::RunningWSLCContainer CreateInternal(Terminal& termi
         }
     }
```
ptrivedi: for when we add sdk support for this, it might make sense to not convert into the legacy AddVolume, AddNamedVolume, AddTmps calls... maybe a single typed mount structure that the launcher/backend can accept would be good. that way CLI and API can go through on reusable path

src/windows/wslc/services/ContainerService.cpp
dkbennett: I have to wire up a new service entry for HostConfig.Mount so that should resolve the issue with bind and volume, and I will add a test for ensuring no empty directory for this case.

src/windows/common/MountSpecParsing.cpp
dkbennett: Will fix and add parsing test cases for each of those.

src/windows/wslc/services/ContainerService.cpp
dkbennett: Yes all of the shared usage of existing Volume, NamedVolume, and Tmpfs calls are going to be replaced by AddMount which will have the distinction and ensure it goes into the Mounts config, along with E2E tests that verify the inspect output matches that expectation. We won't be reusing any of those for this code with my next update, but I'm noting all the differences as explicit tests to add.

test/windows/wslc/e2e/WSLCE2EContainerRunTests.cpp
dkbennett: Tests have been added for colon-containing paths, both parser and E2E.

test/windows/wslc/e2e/WSLCE2EContainerRunTests.cpp
dkbennett: E2E now tests validate inspect mount entries for bind, volume, and tmpfs, including type, source, destination, and read/write state. Colon-containing destinations are included.

test/windows/wslc/e2e/WSLCE2EContainerRunTests.cpp
dkbennett: E2E tests verify:
•  --mount type=bind  fails when the source is missing and does not create it.
• Legacy  --volume  succeeds and creates the missing source directory.

src/windows/wslc/arguments/ArgumentValidation.cpp
dkbennett: Valid but unsupported options such as  volume-nocopy ,  volume-opt ,  volume-driver , bind propagation, and unsupported mount types now throw  MountUnsupportedException  and produce the localized "uses an unsupported feature" CLI error.

Malformed or invalid values remain  MountParseException / MountValidationException  and report "Invalid argument."

src/windows/common/MountSpecParsing.cpp
dkbennett: Anonymous volumes are now supported as of a very recent PR merge, so I will update --mount to support it.

src/windows/wslc/services/ContainerService.cpp
dkbennett: `--mount type=bind` now uses `AddMount` and rejects a missing source path, while legacy `--volume` preserves its create-if-missing behavior. E2E coverage verifies both outcomes and confirms that the failed `--mount` operation does not create the source directory.

src/windows/wslc/arguments/ArgumentValidation.cpp
dkbennett: All mount parser, unsupported-feature, and validation reasons now come from localization resources. `MountException::Reason()` therefore carries localized text, and the CLI composes it with the localized invalid or unsupported `--mount` error wrapper.

src/windows/wslc/services/ContainerService.cpp
dkbennett: Container inspect now reports bind, named volume, anonymous volume, and tmpfs mounts created through `--mount`. E2E coverage verifies each mount's type, source or name, destination, and read/write state, including colon-containing destinations.

src/windows/common/MountSpecParsing.cpp
dkbennett: `OptionFamily` now drives option-family tracking directly from the field-definition table, removing duplicated per-field bookkeeping. This also fixes a validation gap where `bind-recursive=enabled` could be accepted with non-bind mount types. A regression test verifies that this mismatch is rejected.

test/windows/wslc/e2e/WSLCE2EContainerRunTests.cpp
dkbennett: The tmpfs E2E test now specifies `tmpfs-size=1MB` and `tmpfs-mode=0700`, then verifies the mounted filesystem reports a 1024 KiB capacity and mode `700` at runtime.

src/windows/wslc/services/ContainerModel.cpp
dkbennett: Going to merge these together into a single mount::Spec that covers all forms of mounts and volumes, and "AddMount" will do the right thing for them. Then we can cross-validate them all using the same language as one collection, while still retaining the important implementation distinctions.

dkbennett: Moved all mount types together into a single mount spec, unified parser and cross-validation across all mounts provided, added cross validation tests and updated existing legacy tests for volume and tmpfs as needed to account for now-localized messages.

ptrivedi (approved):

src/windows/common/MountSpecParsing.cpp
```diff
@@ -0,0 +1,690 @@
...
+    std::wstring source;
+    if (FAILED(wil::GetFullPathNameW(rawSource.c_str(), source)))
+    {
+        ThrowParse(Localization::WSLCCLI_VolumeHostPathInvalid(value, rawSource));
+    }
+
+    if (GetFileAttributesW(source.c_str()) == INVALID_FILE_ATTRIBUTES && GetLastError() == ERROR_INVALID_NAME)
+    {
+        ThrowParse(Localization::WSLCCLI_VolumeHostPathInvalid(value, rawSource));
+    }
```
OneBlue, line 514: nit: I recommend using `wsl::windows::common::filesystem::GetCanonicalPath()` since it's what we use in other places

src/windows/common/MountSpecParsing.cpp
```diff
@@ -0,0 +1,690 @@
...
+    {
+        options.emplace_back(std::format("size={}", FormatDockerTmpfsSize(mount.TmpfsSizeBytes.value())));
+    }
+
+    return wsl::shared::string::Join<char>(options, ',');
+}
+
+std::string NormalizeDestination(std::string destination)
+{
+    std::replace(destination.begin(), destination.end(), '\\', '/');
```
OneBlue, line 639: If I'm following correctly, this is the container path, and in Linux '\\' is actually valid in a path so we shouldn't translate it to `/`

src/windows/common/MountSpecParsing.cpp
```diff
@@ -0,0 +1,690 @@
...
+
+    std::vector<std::string> components;
+    size_t start = 0;
+    while (start <= destination.size())
+    {
+        const auto end = destination.find('/', start);
+        const auto component = destination.substr(start, end - start);
+        if (!component.empty() && component != ".")
+        {
+            if (component == "..")
```
OneBlue, line 649: Do we actually need to do this ? I would expect that the docker API does that for us

src/windows/common/MountSpecParsing.h
```diff
@@ -0,0 +1,125 @@
...
+#include <exception>
+#include <optional>
+#include <span>
+#include <string>
+#include <string_view>
+#include <utility>
+
+namespace wsl::windows::common::mount {
+
+inline constexpr std::string_view c_dockerCliMountGrammarVersion = "25.0.3";
```
OneBlue, line 27: This looks unused

src/windows/wslc/services/ContainerModel.cpp
```diff
@@ -318,19 +258,21 @@ std::vector<std::wstring> EnvironmentVariable::ParseFile(const std::wstring& fil
...
+void ValidateUniqueMountDestinations(const ContainerOptions& options)
 {
-    TmpfsMount result{};
-    auto colonPos = value.find(':');
-    if (colonPos == std::string::npos)
+    try
     {
-        result.m_containerPath = value;
-        return result;
+        mount::ValidateMountCollection(options.Mounts);
```
OneBlue, line 265: I don't think we need to do this in the client side, I would expect the docker API to give us an error if this happens

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -443,6 +457,18 @@ auto MountVolumes(std::vector<WSLCVolumeMount>& volumes, WSLCVirtualMachine& par
     return std::move(errorCleanup);
 }
 
+auto MountVolumesWithUserError(std::vector<WSLCVolumeMount>& volumes, WSLCVirtualMachine& parentVM)
+{
+    try
+    {
+        return MountVolumes(volumes, parentVM);
```
OneBlue, line 464: nit: Instead of having an intermediary exception, we could use `THROW_HR_WITH_USER_ERROR()`  in MountVolumes and skip the wrapper

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -531,6 +557,210 @@ std::map<std::string, std::string> StripInternalLabels(std::optional<std::map<st
...
+            WI_IsAnyFlagSet(value.Flags, ~WSLCMountSpecFlagsValid),
+            "Mount at index %lu has invalid flags: 0x%x",
+            i,
+            value.Flags);
+
+        mount::Type type;
+        switch (value.Type)
+        {
+        case WSLCMountTypeBind:
+            type = mount::Type::Bind;
```
OneBlue, line 583: nit: Could we reuse the same enum value directly to avoid having to map ?

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -531,6 +557,210 @@ std::map<std::string, std::string> StripInternalLabels(std::optional<std::map<st
...
+                if (mount.BindSource == mount::BindSourcePolicy::CreateIfMissing)
+                {
+                    continue;
+                }
+
+                std::error_code error;
+                const auto sourceExists = std::filesystem::exists(mount.Source, error);
+                if (error)
+                {
+                    throw mount::MountValidationException(Localization::MessageWslcBindSourcePathError(mount.Source, error.message()));
```
OneBlue, line 646: Same here, I would recommend throwing the user error directly to make things a bit easier to follow

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -531,6 +557,210 @@ std::map<std::string, std::string> StripInternalLabels(std::optional<std::map<st
...
+    Create,
+    Reject,
+};
+
+PreparedBindMount PrepareBindMount(const std::wstring& source, const std::string& target, bool readOnly, MissingBindSource missingSource)
+{
+    GUID volumeId;
+    THROW_IF_FAILED(CoCreateGuid(&volumeId));
+
+    auto parentVMPath = std::format("/mnt/{}", wsl::shared::string::GuidToString<char>(volumeId));
```
OneBlue, line 723: nit: This could be constructed later since we don't use it until line 749

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -531,6 +557,210 @@ std::map<std::string, std::string> StripInternalLabels(std::optional<std::map<st
...
+    std::filesystem::path hostPath = source;
+    THROW_HR_WITH_USER_ERROR_IF(E_INVALIDARG, Localization::MessagePathNotAbsolute(source), !hostPath.is_absolute());
+
+    std::wstring sourceFilename;
+    {
+        std::error_code ec;
+        hostPath = std::filesystem::canonical(hostPath, ec);
```
OneBlue, line 730: nit: I recommend using `wsl::windows::common::filesystem::GetCanonicalPath()` here (probably didn't exist when you started working on this change)

src/windows/service/inc/wslc.idl
```diff
@@ -245,6 +245,37 @@ typedef struct _WSLCTmpfsMount
...
+typedef enum _WSLCMountSpecFlags
+{
+    WSLCMountSpecFlagsNone = 0,
+    WSLCMountSpecFlagsTmpfsSize = 1,
+    WSLCMountSpecFlagsTmpfsMode = 2,
+    WSLCMountSpecFlagsCreateSourceIfMissing = 4,
+    WSLCMountSpecFlagsTmpfsOptions = 8,
+} WSLCMountSpecFlags;
+
+cpp_quote("#define WSLCMountSpecFlagsValid (WSLCMountSpecFlagsTmpfsSize | WSLCMountSpecFlagsTmpfsMode | WSLCMountSpecFlagsCreateSourceIfMissing | WSLCMountSpecFlagsTmpfsOptions)")
```
OneBlue, line 264: Do we need `WSLCMountSpecFlagsTmpfsOptions` or could we just set `TmpfsOptions` to null when unset ?

src/windows/common/WSLCContainerLauncher.cpp
```diff
@@ -229,29 +229,74 @@ void WSLCContainerLauncher::AddUlimit(const std::string& Name, std::int64_t Soft
...
+        .ReadOnly = ReadOnly,
+    });
+}
+
+void wsl::windows::common::WSLCContainerLauncher::AddMount(const mount::Spec& Mount)
+{
+    WSLCMountSpec mount{};
+    switch (Mount.MountType)
+    {
+    case mount::Type::Bind:
```
OneBlue, line 256: nit: I recommend using the same enum here as well so we don't need to map

PR #41338, opened 2026-08-12
Use alpine instead of debian for registry push/pull test payloads

ggarzia-MSFT:

Four WSLC E2E tests push an image to a local registry purely as transfer payload. They used `debian:latest` (~118 MB) where a much smaller image exercises the identical code path. This switches them to `alpine:latest` (~8.3 MB).

Measured effect on the full WSLC suite: **22.25 min → 21.11 min**, with identical pass/fail results.

| Test | File |
|---|---|
| `WSLCE2E_Image_PushPull` | `WSLCE2EPushPullTests.cpp` |
| `WSLCE2E_Image_Pull_QuietOption` | `WSLCE2EPushPullTests.cpp` |
| `WSLCE2E_Registry_LoginLogout_PushPull_AuthFlow` | `WSLCE2ERegistryTests.cpp` |
| `WSLCE2E_Image_Build_Pull_Success` | `WSLCE2EImageBuildTests.cpp` |

Test-only change; no product code is touched.

## PR Checklist

### Why

Measured against a local registry, recreating the registry container per iteration so blob deduplication does not mask the real cost:

| Operation | debian | alpine |
|---|---|---|
| Cold push | **5.73s** (5.76 / 5.71 / 5.70) | **0.58s** |
| Pull | ~0.15s | ~0.15s |

Four cold pushes × (5.73 − 0.58) ≈ **~21s**.

All of the saving is on the push side. Pulls are already nearly free because the tests delete only the registry-*tagged* copy — the source image's blobs remain local, so the pull is just a manifest fetch.

### Why alpine is safe here

None of the four tests launch a container from the pushed image, so the payload is opaque and the distro is irrelevant. The only one that executes anything (`Build_Pull_Success`) runs `CMD ["echo", "pull-ok"]`, which BusyBox provides.

There is existing precedent: `WSLCE2ETlsRegistryTests.cpp:206` already uses `AlpineTestImage()` for this exact tag→push pattern. The debian usage in these four tests traces back to the original WSLC commit (`358ab87d`) and looks like pattern propagation rather than a deliberate requirement.

### Notes

- Two hardcoded `L"debian:latest"` literals in the registry auth flow are replaced with the `testImage.NameAndTag()` helper, matching the surrounding style.
- `WSLCE2EImageBuildTests` only loaded debian in `TEST_CLASS_SETUP` (still required by the other ~46 build tests), so the modified test adds its own `EnsureImageIsLoaded(AlpineTestImage())`.
- `WSLCTests.cpp`'s `PullImageAdvanced` is deliberately **left on debian** — it retains intentional large-payload round-trip coverage.

## Validation Steps Performed

Full WSLC suite (`test.bat /select:@TestCategory='WSLC'`) run before and after, from the same base commit, against the same product build, on the same machine:

| Run | Payload | Wall clock | Result |
|---|---|---|---|
| Before (master) | debian ~118 MB | 22.25 min (1335.1s) | `Total=1322, Passed=1320, Failed=0, Skipped=2` |
| After (this branch) | alpine ~8.3 MB | 21.11 min (1266.3s) | `Total=1322, Passed=1320, Failed=0, Skipped=2` |
| **Delta** | | **−68.8s (−5.2%)** | identical |

All four modified tests pass, verified individually in the log.

Caveat on the magnitude: the isolated push measurements above only account for ~21s of the observed 68.8s. With n=1 per configuration on a ~22 minute workload, the remainder cannot be separated from run-to-run variance, so **~21s is the defensible figure** and the 68.8s should not be read as reproducible.

### DIFF
--- a/test/windows/wslc/e2e/WSLCE2EImageBuildTests.cpp
+++ b/test/windows/wslc/e2e/WSLCE2EImageBuildTests.cpp
@@ -173,10 +173,12 @@ class WSLCE2EImageBuildTests
     WSLC_TEST_METHOD(WSLCE2E_Image_Build_Pull_Success)
     {
         // A local registry acts as the private image source that --pull re-resolves the base image from.
+        EnsureImageIsLoaded(AlpineTestImage());
+
         auto session = OpenDefaultElevatedSession();
         auto [registryContainer, registryAddress] = StartLocalRegistry(*session, "", "", c_registryPort);
 
-        auto registryImage = TagImageForRegistry(DebianTestImage().NameAndTag(), string::MultiByteToWide(registryAddress));
+        auto registryImage = TagImageForRegistry(AlpineTestImage().NameAndTag(), string::MultiByteToWide(registryAddress));
         auto registryImageCleanup = DeleteImageOnExit(registryImage);
 
         RunWslcAndVerify(std::format(L"push {}", registryImage), {.Stderr = L"", .ExitCode = 0});
--- a/test/windows/wslc/e2e/WSLCE2EPushPullTests.cpp
+++ b/test/windows/wslc/e2e/WSLCE2EPushPullTests.cpp
@@ -55,8 +55,8 @@ class WSLCE2EPushPullTests
 
     WSLC_TEST_METHOD(WSLCE2E_Image_PushPull)
     {
-        const auto& debianImage = DebianTestImage();
-        EnsureImageIsLoaded(debianImage);
+        const auto& testImage = AlpineTestImage();
+        EnsureImageIsLoaded(testImage);
 
         // Start a local registry without auth.
         auto session = OpenDefaultElevatedSession();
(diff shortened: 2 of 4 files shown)


ggarzia-MSFT:
Updated Benchmarking:
Each modified test was run 25 times to account for run to run variance. The failures were caused by the tests destroying and recreating a registry container on the same port for each run and were excluded from results.
| Test | debian | alpine | delta | % faster | debian fails | alpine fails |
|---|---|---|---|---|---|---|
| pushpull | 6.76s | 2.00s | -4.76s | -70.4% | 2 | 1 |
| quiet | 6.75s | 1.96s | -4.79s | -71.0% | 0 | 1 |
| authflow | 7.66s | 2.82s | -4.84s | -63.2% | 2 | 1 |
| buildpull | 8.44s | 3.58s | -4.86s | -57.6% | 0 | 0 |
| **Total** | **29.61s** | **10.36s** | **-19.25s** | **-65.0%** | **4/100** | **3/100** |

OneBlue (approved):

PR #41342, opened 2026-08-13
Fix relay loop to track closed states independently for left and righ…

Eamon2009:

Fix relay loop to track closed states independently for left and right handles. Previously, the loop could exit or misbehave if only one handle was null. This update introduces `leftReadClosed` and `rightReadClosed` boolean flags to properly manage half-closed connections. The relay loop now correctly continues reading from an open handle and only breaks when both the left and right reads are fully closed.

## PR Checklist

### Independent close tracking
The previous relay loop did not distinguish between the left and right handles when detecting read-side closures. If one side closed its write direction (half-close) or its read handle became null, the loop would either exit prematurely or behave unpredictably, dropping data that was still buffered on the open side.

This change introduces two explicit boolean flags -`leftReadClosed` and `rightReadClosed` -that are set independently when `POLLHUP`/`POLLERR` is detected on each respective handle or when the handle becomes null. This allows the loop to continue relaying data in the direction that is still open.

### Break condition
The exit condition is now gated on **both** flags being true:

```cpp
if (leftReadClosed && rightReadClosed)
{
    break;
}
```
This ensures the relay thread stays alive until both directions have been fully drained, preventing data loss during half-closed or asynchronous teardown scenarios.


src/linux/init/localhost.cpp
```diff
@@ -516,7 +514,8 @@ int RunPortTracker(int Argc, char** Argv)
         }
 
         auto& ifRequest = *reinterpret_cast<ifreq*>(ifreqMemory->data());
-        memcpy(request.InterfaceName, ifRequest.ifr_ifrn.ifrn_name, sizeof(request.InterfaceName));
+        memcpy(request.InterfaceName, ifRequest.ifr_ifrn.ifrn_name, sizeof(request.InterfaceName) - 1);
+        request.InterfaceName[sizeof(request.InterfaceName) - 1] = '\0';
```
OneBlue, line 518: nit: This can be skipped now that we never write the last byte

Eamon2009:
Hi @OneBlue just checking in on this PR. CI is green. Please let me know if there's anything else needed from my side. Thanks for your time!

Eamon2009:
Hi @OneBlue just checking in on this PR. CI is green. Please let me know if there's anything else needed from my side. Thanks for your time!

Eamon2009:
Hi @benhillis just checking in on this PR. CI is green. Please let me know if there's anything else needed from my side. Thanks for your time!

PR #41348, opened 2026-08-13
CLI: Set defaults for signal arguments in stop/kill commands, update help documentation

dkbennett:

Provides explicit CLI defaults for `--signal`: `SIGKILL` for `container kill` and none for `container stop`. This makes the code a little cleaner for both and clarifies the stop signal behavior defaulting to "SignalNone" is intentional.

Updates the description for the Signal arguments to be more clear about the behavior and gives Stop it's own arg description with it's different default behavior.

All of this aligns to and is consistent with Docker behavior.

## PR Checklist

The defaults were being set explicitly by kill already, and stop was being set implicitly by being the default in the StopContainerOptions, so there should be no functional change here, the CLI is just being more compact and explicit about intended defaults and updating the documentation.

Updated the argument handling to get the value or provide the appropriate default so each command will always have the correct default set.

## Validation Steps Performed
- E2E Container Kill and Container Stop tests passed.

### DIFF
--- a/localization/strings/en-US/Resources.resw
+++ b/localization/strings/en-US/Resources.resw
@@ -2622,10 +2622,11 @@ Example: tar -cf - files | wslc container cp - CONTAINER:/path</value>
     <value>Display detailed information about a container.</value>
   </data>
   <data name="WSLCCLI_ContainerKillDesc" xml:space="preserve">
-    <value>Kill containers.</value>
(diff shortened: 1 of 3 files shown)


OneBlue (approved):
LGTM. The test failure is unrelated to this change. Restarted the pipeline

PR #41353, opened 2026-08-13
Cache container images during tests to prevent unnecessary loading

ggarzia-MSFT:

Adds a `TestImageRegistry` to the wslc E2E tests that tracks which test images are already
loaded in a session, so repeated setup across test classes stops reloading the same image
tarballs. Measured **43.6 s (10.1%)** faster across the 475 affected tests.

## PR Checklist

Test-only change; no product code is touched.

Previously every test class independently ran `EnsureImageIsLoaded` in setup and
`EnsureImageIsDeleted` in cleanup. Because classes share a small set of base images
(debian, alpine, hello-world, python), each cleanup deleted an image the next class
immediately reloaded, costing ~1 s per reload.

`TestImageRegistry` is a per-process singleton that seeds itself once per session from a
live `image list` query, then answers `EnsureLoaded` from memory. Shared-image deletes were
removed from class cleanups; images that a test genuinely needs gone are still deleted
explicitly.

Keeping a cache honest is the hard part, so rather than trusting call sites to report
mutations, `NoteCommand` is hooked into the three `RunWslc` primitives and observes *every*
wslc invocation. Only removal verbs matter — `image remove`/`delete`/`rm`, `image prune`,
and `system session terminate` — because commands that *add* an image can at worst leave the
cache pessimistic, which costs one extra query rather than producing a wrong answer.

Invalidation is per-image where the command names its targets, and session-wide only when
the affected set can't be known (`image prune`, session teardown, digest or port-qualified
references). This distinction matters: an earlier revision invalidated session-wide on every
removal, which destroyed the seed and gave back 13.5 s of the win.

All image handling now lives on the registry; the free `EnsureImageIsLoaded` and
`EnsureImageIsDeleted` helpers are deleted. `Delete` deliberately queries `image list`
directly instead of consulting the cache, since built and imported images are created
outside the registry and a cached negative would silently leak them.

## Validation Steps Performed

Full 475-test cacheable selection (27 classes), x64 Debug, 3 runs per arm against an
identical deployment:

| Arm | Runs | Mean | Spread |
|---|---|---|---|
| master | 428.00 / 430.96 / 429.72 | **429.56 s** | 3.0 s |
| this change | 385.55 / 384.40 / 388.04 | **386.00 s** | 3.6 s |

**43.56 s faster (10.1%)**, 475/475 passing on all six runs. Ranges are separated by a 40 s
gap with no overlap.

### DIFF
--- a/test/windows/wslc/e2e/TestImageRegistry.cpp
+++ b/test/windows/wslc/e2e/TestImageRegistry.cpp
@@ -29,92 +29,6 @@ std::wstring TestImageRegistry::FormatCommand(const std::wstring& sessionName, c
     return std::format(L"--session \"{}\" {}", sessionName, command);
 }
 
-std::wstring TestImageRegistry::ParseSessionName(const std::wstring& commandLine)
-{
-    constexpr std::wstring_view sessionFlag = L"--session ";
-    const auto flagOffset = commandLine.find(sessionFlag);
-    if (flagOffset == std::wstring::npos)
-    {
-        return {};
-    }
-
-    auto valueOffset = commandLine.find_first_not_of(L' ', flagOffset + sessionFlag.size());
-    if (valueOffset == std::wstring::npos)
-    {
-        return {};
-    }
-
-    if (commandLine[valueOffset] == L'"')
-    {
-        const auto closingQuote = commandLine.find(L'"', valueOffset + 1);
-        if (closingQuote == std::wstring::npos)
-        {
-            return {};
-        }
-
-        return commandLine.substr(valueOffset + 1, closingQuote - valueOffset - 1);
-    }
-
-    const auto end = commandLine.find(L' ', valueOffset);
-    return commandLine.substr(valueOffset, end == std::wstring::npos ? std::wstring::npos : end - valueOffset);
-}
-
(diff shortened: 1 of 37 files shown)


test/windows/wslc/e2e/TestImageRegistry.cpp
```diff
@@ -0,0 +1,245 @@
...
+        return commandLine.substr(valueOffset + 1, closingQuote - valueOffset - 1);
+    }
+
+    const auto end = commandLine.find(L' ', valueOffset);
+    return commandLine.substr(valueOffset, end == std::wstring::npos ? std::wstring::npos : end - valueOffset);
+}
+
+bool TestImageRegistry::CommandCanRemoveImages(const std::wstring& commandLine)
+{
+    constexpr std::wstring_view removalCommands[] = {
```
OneBlue: I wouldn't recommend parsing the command lines to try to figure out if an image could get removed, instead I would recommend making the tests always restore the state of the test session whenever they delete an image. 

(otherwise we might hit some tricky corner cases). Doing this would also allow us to skip the synchronization part

OneBlue (approved):

PR #41355, opened 2026-08-14
Preserve token owner during distribution move rollback

benhillis:

Preserves the source VHD's owner when rolling back a distribution move.

## PR Checklist

Follow-up to #41333. Before moving the VHD, capture its current owner SID under caller impersonation. If the operation must be rolled back, use that SID as the duplicated token's default owner before the reverse `MoveFileExW` operation.

This keeps the normal forward move owned by the caller's user SID while ensuring a cross-volume rollback recreates the VHD with its pre-move owner.

## Validation Steps Performed

- `FormatSource.ps1`
- `cmake --build . -- -m`
- `bin\x64\Debug\test.bat -f /name:*MoveVhd*` was attempted; shared module setup failed while setting the default `test_distro`, before either selected test executed.

### DIFF
--- a/src/windows/service/exe/LxssUserSession.cpp
+++ b/src/windows/service/exe/LxssUserSession.cpp
@@ -949,7 +949,14 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
     // Cross-volume MoveFileEx creates a new file using the impersonation token's
     // default owner. Normalize that owner to the caller's user SID so elevated moves
     // do not produce a VHD owned by BUILTIN\Administrators.
-    auto originalTokenOwner = wil::get_token_information<TOKEN_OWNER>(userToken.get());
+    PSID originalVhdOwner = nullptr;
+    wil::unique_hlocal originalSecurityDescriptor;
+    {
+        auto impersonate = wil::impersonate_token(userToken.get());
+        THROW_IF_WIN32_ERROR(GetNamedSecurityInfoW(
+            distro.VhdFilePath.c_str(), SE_FILE_OBJECT, OWNER_SECURITY_INFORMATION, &originalVhdOwner, nullptr, nullptr, nullptr, &originalSecurityDescriptor));
+    }
+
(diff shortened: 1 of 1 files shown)


src/windows/service/exe/LxssUserSession.cpp
```diff
@@ -949,6 +949,7 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
     // Cross-volume MoveFileEx creates a new file using the impersonation token's
     // default owner. Normalize that owner to the caller's user SID so elevated moves
     // do not produce a VHD owned by BUILTIN\Administrators.
+    auto originalTokenOwner = wil::get_token_information<TOKEN_OWNER>(userToken.get());
```
OneBlue: Sorry if I was unclear, what I meant on the original change was to look up the owner of the VHD before we move it, and to use that SID as the owner when we restore it, so if the operation fails, nothing should change

benhillis: nope, that was clear, I asked copilot to address your feedback and it went down the wrong path.

src/windows/service/exe/LxssUserSession.cpp
```diff
@@ -967,6 +968,8 @@ HRESULT LxssUserSessionImpl::MoveDistribution(_In_ LPCGUID DistroGuid, _In_ LPCW
     }
 
     auto revert = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
+        LOG_IF_WIN32_BOOL_FALSE(SetTokenInformation(userToken.get(), TokenOwner, originalTokenOwner.get(), sizeof(*originalTokenOwner)));
+
```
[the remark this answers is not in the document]
benhillis: skip

src/windows/service/exe/LxssUserSession.cpp
benhillis: Updated to capture the VHD owner SID before the move and use that SID as TokenOwner for rollback. This now preserves the pre-move owner rather than the token default owner.

OneBlue (approved):

PR #41360, opened 2026-08-14
CLI: Fix shm-size and all text-to-byte size conversion and align them all to Docker

dkbennett:

Adds common Windows storage-size parsing and formatting helpers with Docker-compatible decimal and binary unit support.

The converters are used for WSLC memory arguments, WSLC user settings, container statistics, and byte-count output. Fractional sizes such as `1.5G` are now parsed correctly, and formatting supports configurable decimal places and spacing.

This specifically fixes `--shm-size` rejecting fractional Docker-compatible values such as `1.5G`. The value is parsed as 1,610,612,736 bytes and verified against the runtime `/dev/shm` capacity.

## PR Checklist

- Adds `ParseStorageSize()` and `FormatStorageSize()` to the Windows common string helpers.
- Supports Docker-compatible decimal and binary units through `P`, including case-insensitive `K`, `KB`, and `KiB` forms, fractional values, exponent syntax, and optional spacing.
- Uses checked integer arithmetic for integral inputs so byte counts above `2^53` remain exact; floating-point parsing is limited to fractional and exponent forms.
- Supports configurable output precision and decimal or binary unit formatting.
- Uses the common parser for WSLC `--memory`, `--shm-size`, `session.memorySize`, and `session.maxStorageSize` values.
- Preserves support for leading whitespace in existing quoted WSLC settings values.
- Preserves localized CLI errors and signed `int64_t` range validation for memory arguments.
- Replaces duplicated container statistics formatting with the common formatter.
- Moves the Windows-only `FormatBytes()` helper out of the cross-platform shared header.
- Leaves the separate `wsl.exe` and `.wslconfig` legacy size parsers unchanged.
- Adds table-driven parser, formatter, rounding, invalid-input, boundary, and bidirectional round-trip tests.
- Adds 8 `StringUnitTests` with 350 cases covering Docker formats, invalid inputs, rounding, `2^53` and integer boundaries, and bidirectional round trips.
- Adds WSLC settings coverage for fractional and IEC values, leading-whitespace compatibility, sub-MiB rejection, and `uint32_t` MiB limits.

## Validation Steps Performed

- Full x64 Debug build
- Deployed the rebuilt MSI to the host
- `*StringUnitTests*`: 8 passed
- `*WSLCCLIArgumentUnitTests*`: 18 passed
- `*WSLCCLISettingsUnitTests*`: 38 passed
- `*WSLCE2ETests::*ShmSize*`: 4 passed

### DIFF
--- a/src/windows/common/string.cpp
+++ b/src/windows/common/string.cpp
@@ -13,6 +13,9 @@ Abstract:
 --*/
 
 #include "precomp.h"
+#include <charconv>
+#include <cmath>
+#include <limits>
 
 std::vector<std::string> wsl::windows::common::string::InitializeStringSet(_In_count_(BufferSize) LPCSTR Buffer, _In_ SIZE_T BufferSize)
 {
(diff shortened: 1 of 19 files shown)


src/windows/common/string.cpp
```diff
@@ -286,6 +289,103 @@ std::string wsl::windows::common::string::WideToMultiByte(_In_ std::wstring_view
...
+    }
+
+    auto narrowNumber = WideToMultiByte(number);
+    if (!narrowNumber.empty() && narrowNumber.front() == '+')
+    {
+        narrowNumber.erase(0, 1);
+    }
+
+    double value{};
+    const auto result = std::from_chars(narrowNumber.data(), narrowNumber.data() + narrowNumber.size(), value, std::chars_format::general);
```
ptrivedi: nit: this doesn't matter as much for --memory and --shm-size because precision loss will begin at very high (~9PB), so not realistic for container sizes. But it is a shared utility and a general uint64_t helper, so worth a comment. 
Using double here can silently change byte counts above 2^53. Could integer-only inputs use checked integer arithmetic, with floating-point parsing used for fractional forms?
Also would be useful to add tests for boundary tests around INT64_MAX and 2^53

src/windows/wslc/arguments/SpecParsing.cpp
```diff
@@ -814,13 +814,14 @@ models::InspectType GetInspectTypeFromString(const std::wstring& input, const st
 
 int64_t GetMemorySizeFromString(const std::wstring& input, const std::wstring& argName)
 {
-    auto parsed = wsl::shared::string::ParseMemorySize(input.c_str());
-    if (!parsed.has_value())
+    const auto bytes =
+        wsl::windows::common::string::ParseStorageSize(std::wstring_view{input}, wsl::windows::common::string::StorageSizeUnit::Binary);
```
ptrivedi, line 818: nit: only wslc arguments are using the new parser. WSLCSettings, CommandLine, configfile etc. still use the legacy ParseMemorySize, so the size syntax will be inconsistent across WSL. Out of the scope of the PR, but something to look into and/or at least document

test/windows/StringUnitTests.cpp
```diff
@@ -0,0 +1,273 @@
...
+    std::wstring Text;
+    StorageSizeUnit Unit;
+    uint32_t DecimalPlaces;
+    bool IncludeSpace;
+};
+
+void VerifyDockerStorageSize(const std::string& Input, StorageSizeUnit Unit, std::optional<uint64_t> Expected)
+{
+    const auto wideInput = wsl::shared::string::MultiByteToWide(Input);
+    VERIFY_ARE_EQUAL(ParseStorageSize(wideInput, Unit), Expected);
```
ptrivedi: nit: VERIFY_ARE_EQUAL expects (expected, actual). This is reversed. Doesn't affect test run i guess, only debugging/diags

ptrivedi (dismissed):
Few minor comments, LGTM overall

src/windows/wslc/arguments/SpecParsing.cpp
dkbennett: Yes, I am intentionally leaving .wslconfig and wsl.exe alone (and CommandLine) as they have different parsing requirements and are a separate product that can decide to use or not use the new library independently, but the settings should be converted over and consistent for WSLC, so will update those.

src/windows/common/string.cpp
dkbennett: Will fix this and use int64_t for everything non-exponent or non-fractional, with float used for fractional and exponent forms and add boundary tests.

OneBlue (approved):

PR #41369, opened 2026-08-17
wslc: match docker output format for image list

ggarzia-MSFT:

Brings `wslc image list` output to parity with `docker images` for both `--format json` and the default/table format, so existing Docker tooling and scripts can consume `wslc` output unchanged.

## PR Checklist

Output was compared field-by-field against Docker 29.6.2. Note that Docker 29.x's *default* `docker images` now renders the containerd-store tree layout (`IMAGE / ID / DISK USAGE / CONTENT SIZE / EXTRA`), so `docker images --format table` is the correct baseline for the classic 5-column contract.

**JSON format**

- `Containers` previously hardcoded `"N/A"`. It now reports a real count. Docker semantics were verified empirically: it counts **all** containers (running *and* stopped) keyed by image ID, and reports `"0"` — never `"N/A"` — for unused images. Falls back to `"N/A"` only if the container list can't be retrieved.
- Plumbed a resolved image ID from the Docker API through to the CLI so images and containers can be joined. `WSLCContainerEntry.Image` holds the reference the *caller* passed (e.g. `alpine:latest`), not a resolved ID, so it can't be matched against `ImageInformation.Id`. Added `ImageID` to `ContainerInfo` (`docker_schema.h`), `ImageId[256]` to `WSLCContainerEntry` (`wslc.idl`), and `ImageId` to `ContainerInformation`.
  - `IWSLCSession` is an internal, non-stable interface rebuilt in lockstep with its only clients, so appending a struct field is allowed here — unlike `WSLCCompat.idl` / `WslPluginApi.h`.
  - `ImageId` is deliberately **excluded** from `ContainerInformation`'s serialization macro, so `container list --format json` output is unchanged and keeps reporting the image reference to match Docker.

**Table format**

- Untagged images now render `<none>` instead of `<untagged>` in both `REPOSITORY` and `TAG`.
- Sizes use Docker's SI formatting (`120MB`, `8.44MB`, `10.1kB`) instead of `119.86 MB`.
- `--no-trunc` now preserves the `sha256:` prefix. `TruncateId()` always strips it, so the table branch now passes the raw ID through when not truncating.
- The table branch was refactored to reuse the same `ToImageOutput` helper as the JSON branch, so the two formats can no longer drift.

**Relative timestamps**

`FormatRelativeTime` is now an exact port of Go's `units.HumanDuration` (per `docker/go-units/duration.go`), which Docker uses for `CreatedSince`. Beyond wording, the bucket boundaries were wrong:

| elapsed | before | after |
|---|---|---|
| < 1s | `0 seconds ago` | `Less than a second ago` |
| 60–119s | `1 minute ago` | `About a minute ago` |
| ~1h | `1 hour ago` | `About an hour ago` |
| 24–47h | `1 day ago` | `24`–`47 hours ago` |
| 1 year | `1 year ago` | `12 months ago` |

Go stays in hours until 48h, days until 14d, weeks until ~60d, and months until ~2 years, which makes the singular day/week/month/year forms unreachable. Hours are also *rounded*, not truncated (`int(d.Hours()+0.5)`).

The logic was split into `FormatRelativeTime(timestamp)` (computes elapsed vs. now) and a pure `FormatElapsedSeconds(elapsed)` so the duration buckets are deterministically unit-testable without clock-tick flakiness.

This function is shared with `container list`, which picks up the corrected wording as well.

**Known remaining gaps (not addressed here)**

- `Digest`: nlohmann emits a literal `<none>` while Go's `encoding/json` HTML-escapes it to `\u003cnone\u003e`. The decoded values are equal; only the raw bytes differ.
- `container list` STATUS renders `exited <time>` where Docker renders `Exited (0) <time>`. Out of scope for this PR; being handled in a follow-up container-list pass.

## Validation Steps Performed

- Full build, deployed via `tools\deploy\deploy-to-host.ps1`, and compared live `wslc` output against Docker 29.6.2 side by side for both JSON and table formats, including `--no-trunc`.
- `Containers` verified end to end: created one running and one exited container from the same image and confirmed Docker reports `"2"`; confirmed `wslc` reports `"2"` for an image with two containers, `"0"` for an unused image, and `"0"` again after removal.
- Added e2e coverage in `WSLCE2EImageListTests`: container-count reporting, a numeric-`Containers` assertion in the Docker-shape test, table-vs-JSON value agreement, and `--no-trunc` algorithm-prefix retention. **19/19 pass.**
- Added `WSLCCLIRelativeTimeUnitTests` covering every `HumanDuration` bucket and boundary, hour rounding, negative/zero elapsed. **10/10 pass.**
- `WSLCE2EContainerListTests` 17/17 pass (shared relative-time change).
- `FormatSource.ps1` clean.

### DIFF
--- a/src/windows/wslc/services/ContainerModel.h
+++ b/src/windows/wslc/services/ContainerModel.h
@@ -130,6 +130,9 @@ struct ContainerInformation
     std::string Id;
     std::string Name;
     std::string Image;
+    // Id of the image the container was created from. Not serialized: `container list --format json`
+    // reports the image reference, matching docker.
+    std::string ImageId;
     WSLCContainerState State;
     ULONGLONG StateChangedAt{};
     ULONGLONG CreatedAt{};
--- a/src/windows/wslc/services/ContainerModel.h
+++ b/src/windows/wslc/services/ContainerModel.h
@@ -130,9 +130,6 @@ struct ContainerInformation
     std::string Id;
     std::string Name;
     std::string Image;
-    // Id of the image the container was created from. Not serialized: `container list --format json`
-    // reports the image reference, matching docker.
-    std::string ImageId;
     WSLCContainerState State;
     ULONGLONG StateChangedAt{};
     ULONGLONG CreatedAt{};
--- a/src/windows/wslc/services/ContainerService.cpp
+++ b/src/windows/wslc/services/ContainerService.cpp
@@ -309,49 +309,65 @@ std::wstring ContainerService::FormatRelativeTime(ULONGLONG timestamp)
         return L"";
     }
 
+    return FormatElapsedSeconds(static_cast<LONGLONG>(std::time(nullptr)) - static_cast<LONGLONG>(timestamp));
+}
+
+std::wstring ContainerService::FormatElapsedSeconds(LONGLONG elapsedSeconds)
+{
     constexpr LONGLONG SecondsPerMinute = std::chrono::duration_cast<std::chrono::seconds>(1min).count();
     constexpr LONGLONG SecondsPerHour = std::chrono::duration_cast<std::chrono::seconds>(1h).count();
-    constexpr LONGLONG SecondsPerDay = std::chrono::duration_cast<std::chrono::seconds>(24h).count();
-    constexpr LONGLONG SecondsPerWeek = SecondsPerDay * 7;
-    constexpr LONGLONG SecondsPerMonth = SecondsPerDay * 30;
-    constexpr LONGLONG SecondsPerYear = SecondsPerDay * 365;
+    constexpr LONGLONG HoursPerDay = 24;
+    constexpr LONGLONG MinutesPerHour = 60;
 
-    auto elapsed = static_cast<LONGLONG>(std::time(nullptr)) - static_cast<LONGLONG>(timestamp);
-    if (elapsed < 0)
-    {
-        elapsed = 0;
-    }
+    const auto elapsed = std::max<LONGLONG>(elapsedSeconds, 0);
 
-    auto pluralize = [](LONGLONG count, const wchar_t* singular, const wchar_t* plural) {
-        return std::format(L"{} {} ago", count, (count == 1 ? singular : plural));
-    };
-
-    if (elapsed < SecondsPerMinute)
+    if (elapsed < 1)
     {
-        return pluralize(elapsed, L"second", L"seconds");
+        return L"Less than a second ago";
     }
-    else if (elapsed < SecondsPerHour)
+    else if (elapsed == 1)
     {
-        return pluralize(elapsed / SecondsPerMinute, L"minute", L"minutes");
+        return L"1 second ago";
     }
-    else if (elapsed < SecondsPerDay)
+    else if (elapsed < SecondsPerMinute)
     {
-        return pluralize(elapsed / SecondsPerHour, L"hour", L"hours");
-    }
-    else if (elapsed < SecondsPerWeek)
-    {
-        return pluralize(elapsed / SecondsPerDay, L"day", L"days");
-    }
-    else if (elapsed < SecondsPerMonth)
-    {
-        return pluralize(elapsed / SecondsPerWeek, L"week", L"weeks");
-    }
-    else if (elapsed < SecondsPerYear)
-    {
-        return pluralize(elapsed / SecondsPerMonth, L"month", L"months");
+        return std::format(L"{} seconds ago", elapsed);
     }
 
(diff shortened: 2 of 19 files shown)


src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -69,6 +71,101 @@ namespace {
...
+            size /= 1000.0;
+            unit++;
+        }
+
+        return std::format("{:.3g}{}", size, units[unit]);
+    }
+
+    // Formats a unix timestamp the way docker does, matching Go's time.Time.String()
+    // layout "2006-01-02 15:04:05 -0700 MST" (e.g. "2026-07-12 17:00:00 -0700 PDT").
+    std::string FormatDockerTimestamp(LONGLONG timestamp)
```
dkbennett: This should move to windows common string, this is definitely something we would want to use across WSLC projects.

src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -69,6 +71,101 @@ namespace {
...
+    {
+        // Report the count as unavailable rather than failing the listing, which is what docker
+        // does when it has no count for an image.
+        LOG_CAUGHT_EXCEPTION();
+        return std::nullopt;
+    }
+
+    // Formats a byte count the way docker does: SI units with three significant digits
+    // (119856765 -> "120MB").
+    std::string FormatDockerSize(int64_t bytes)
```
dkbennett: the shm-size fix PR handles all of these conversions to docker-compatible, so this can probably be removed when that PR is merged and use the new storage size conversions.

dkbennett (commented):
Overall looks good, few comments about factoring

src/windows/wslc/tasks/ImageTasks.cpp
ggarzia-MSFT: Good catch

src/windows/wslc/services/ContainerModel.h
```diff
@@ -130,6 +130,9 @@ struct ContainerInformation
     std::string Id;
     std::string Name;
     std::string Image;
+    // Id of the image the container was created from. Not serialized: `container list --format json`
```
OneBlue: I'm a bit confused about what the "Not serialized: `container list --format json`" part means here. I think just leaving "Id of the image the container was created from" is sufficient here

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -309,49 +309,65 @@ std::wstring ContainerService::FormatRelativeTime(ULONGLONG timestamp)
...
+    constexpr LONGLONG MinutesPerHour = 60;
+
+    const auto elapsed = std::max<LONGLONG>(elapsedSeconds, 0);
 
-    auto elapsed = static_cast<LONGLONG>(std::time(nullptr)) - static_cast<LONGLONG>(timestamp);
-    if (elapsed < 0)
+    if (elapsed < 1)
     {
-        elapsed = 0;
+        return L"Less than a second ago";
```
OneBlue: nit: This should be localized

src/windows/wslc/services/ImageModel.h
```diff
@@ -28,6 +25,29 @@ struct ImageInformation
...
+    std::string Size;
+    std::string Tag;
+    std::string UniqueSize;
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(
+        ImageOutputInformation, Containers, CreatedAt, CreatedSince, Digest, ID, Repository, SharedSize, Size, Tag, UniqueSize);
+};
+
+// Value docker emits for repository, tag, and digest data it does not have.
+inline constexpr std::string_view c_dockerNone = "<none>";
```
OneBlue: nit: I think calling this just `c_none` would be sufficient. `c_dockerNone` feels a bit misleading

src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -69,6 +70,81 @@ namespace {
         Terminal& m_terminal;
     };
 
+    // Values docker reports as unavailable. wslc does not track image digests or layer sharing, so
+    // these use the same placeholders docker emits.
+    constexpr std::string_view c_notAvailable = "N/A";
+    constexpr std::string_view c_none = models::c_dockerNone;
+
+    // Counts the containers created from each image, keyed by image id. Docker includes stopped
```
OneBlue: If we want to have a count for this, I would recommend counting in the service, under the container lock. Otherwise there's a TOCTOU between the time we get the image list and the time we count the containers using them

OneBlue (approved):

PR #41375, opened 2026-08-18
Align container list format with Docker specifications

ggarzia-MSFT:

Brings `wslc container list` to output parity with `docker container list`, for both the table and `--format json`. This was the largest remaining gap found in a systematic wslc-vs-docker output comparison across 81 command pairs.

Three changes:

- **Table** — column order now matches docker (`CONTAINER ID / IMAGE / COMMAND / CREATED / STATUS / PORTS / NAMES`), adding the missing `COMMAND` column and moving `NAMES` last. `STATUS` now reports the runtime's own description (`Up 5 minutes`, `Exited (0) 2 hours ago`) instead of a locally derived string.
- **`--format json`** — emits docker's 16-key shape, with every value a pre-rendered string apart from the nested `Platform` object. This matches the convention already used by `wslc image list` and `wslc network list`.
- **`--quiet`** — truncates to docker's 12-character short ID, with `--no-trunc` for the full ID.

The service now supplies the additional data docker reports (`Command`, `Status`, `Labels`, `Networks`, `Mounts`, `LocalVolumes`), so `WSLCContainerEntry` gains those fields.

## PR Checklist

### JSON output is a breaking change

`container list --format json` previously emitted a raw-data view of the internal model:

```json
{"Id":"<64 chars>","Name":"web","Image":"alpine:3.20","State":1,"StateChangedAt":1755543210,"CreatedAt":1755543200,"Ports":[...]}
```

It now emits docker's shape:

```json
{"Command":"\"sleep 3600\"","CreatedAt":"2026-08-18 12:06:50 -0700 PDT","HealthStatus":"none","ID":"0dba0f244a97","Image":"alpine:3.20","Labels":"...","LocalVolumes":"0","Mounts":"","Names":"web","Networks":"bridge","Platform":{"architecture":"amd64","os":"linux"},"Ports":"","RunningFor":"28 seconds ago","Size":"0B","State":"running","Status":"Up 27 seconds"}
```

Notably Id → ID, Name → Names, integer timestamps → formatted strings, the Ports array → a rendered string, and IDs truncated unless --no-trunc . This mirrors the rename already made deliberately for network list, and is what makes docker-oriented tooling and Go-template-style consumers work unchanged.

## Design

 ToContainerOutput() builds a single  ContainerOutputInformation  consumed by both the table and the JSON renderer, so the two cannot drift. This is the same pattern as  ToImageOutput() /  ImageOutputInformation .

 ContainerOutputInformation  is kept separate from  ContainerInformation  on purpose: the former mirrors docker's all-string output shape, the latter mirrors the service's native types.

Container state names exist in two forms —  ContainerStateName() returns invariant English because it feeds the machine-readable JSON, while  LocalizedContainerStateName() is the display form used by the table.

## Validation Steps Performed

• Added WSLCE2E_Container_List_TableFormat_MatchesDockerColumnOrder , asserting the header row matches docker's column order.
• Added WSLCE2E_Container_List_JsonFormat_MatchesDockerShape , asserting the exact 16-key set, that every value is a string apart from Platform , and that IDs are 12 characters by default.
• Rewrote the --quiet E2E test to cover both truncated and --no-trunc  output.
• Added 10 unit tests for  FormatCommand  (quoting, escaping, 20-character shortening, code-point counting for multi-byte input) and  FormatStatus  (runtime description preferred, fallback when empty).
• Migrated the ~12 existing E2E call sites and the shared  ListAllContainers() helper to the new JSON shape.
• Expected output was validated against real docker container list output captured from a parity harness run covering 81 wslc/docker command pairs.
•  clang-format 19.1.5 clean across all files changed on the branch.


src/windows/wslc/services/ContainerService.cpp
```diff
@@ -396,37 +396,109 @@ int ContainerService::Attach(Terminal& terminal, Session& session, const std::st
...
+
+        if (codePoints > c_maxDisplayWidth)
+        {
+            wide.resize(cutoff);
+            wide += L'\u2026';
+        }
+    }
+
+    std::wstring quoted{L'"'};
+    for (const auto character : wide)
```
OneBlue: I'm guessing this is something that docker does ? If so, should we do this before we truncate ? Otherwise we could exceed `c_maxDisplayWidth`

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -396,37 +396,109 @@ int ContainerService::Attach(Terminal& terminal, Session& session, const std::st
...
+    {
+        // Count code points rather than code units so a surrogate pair is never split.
+        size_t codePoints = 0;
+        size_t index = 0;
+        size_t cutoff = 0;
+        for (; index < wide.size(); ++codePoints)
+        {
+            if (codePoints == c_maxDisplayWidth - 1)
+            {
+                cutoff = index;
```
OneBlue: Shouldn't we break here ?

src/windows/wslc/services/ContainerService.cpp
ggarzia-MSFT: Yes, docker does this. Exceeding c_maxDisplayWidth is expected behavior, docker truncates to c_maxDisplayWidth (20) and then quotes it. This results in the final string being larger than the max display width.

src/windows/wslc/services/ContainerService.cpp
ggarzia-MSFT: Good catch

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -396,37 +396,116 @@ int ContainerService::Attach(Terminal& terminal, Session& session, const std::st
...
+    {
+        // Count code points rather than code units so a surrogate pair is never split.
+        size_t codePoints = 0;
+        size_t index = 0;
+        size_t cutoff = 0;
+        for (; index < wide.size(); ++codePoints)
+        {
+            if (codePoints > c_maxDisplayWidth)
+            {
+                break;
```
OneBlue: What's the reason for putting the break in a new block, instead of breaking after we set `cutoff = index` ? 

I wonder if I'm missing something

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -114,6 +114,35 @@ nlohmann::json ComputeContainerStatsJson(const wsl::windows::common::docker_sche
...
+// Builds the representation of a container, shared by the table and json output so the two cannot
+// drift. Every value is emitted as a string apart from the platform object, and the id is truncated
+// unless --no-trunc is passed.
+ContainerOutputInformation ToContainerOutput(const ContainerInformation& container, bool truncate)
+{
+    ContainerOutputInformation entry;
+    entry.Command = WideToMultiByte(ContainerService::FormatCommand(container.Command, truncate));
+    entry.CreatedAt = FormatDockerTimestamp(static_cast<LONGLONG>(container.CreatedAt));
+    // wslc does not surface health checks in the listing, which docker reports as "none".
+    entry.HealthStatus = "none";
```
ptrivedi: can't the HealthStatus be dervied from container.Status which is available here?

ptrivedi: also, if we end up populating HealthStatus based on container.Status, we should add test cases for healthy and unhealthy state verification

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -396,37 +396,116 @@ int ContainerService::Attach(Terminal& terminal, Session& session, const std::st
...
+
+    // Quoting happens after truncation, so the result can exceed c_maxDisplayWidth. This matches docker, which truncates
+    // the command first and quotes the truncated value.
+    std::wstring quoted{L'"'};
+    for (const auto character : wide)
+    {
+        if (character == L'"' || character == L'\\')
+        {
+            quoted += L'\\';
+        }
```
ptrivedi, line 487: docker formats this field using strconv.Quote, which escapes newlines, tabs, and other control characters. Here this loop only escapes quotes and backslashes, so a command that containers newline can break the table row and will end up producing different json value. Could we try to implement equivalent escaping?

ptrivedi, line 487: once we do, we should also test \n \t

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -114,6 +114,35 @@ nlohmann::json ComputeContainerStatsJson(const wsl::windows::common::docker_sche
...
+    entry.ID = truncate ? TruncateId(container.Id) : container.Id;
+    entry.Image = container.Image;
+    entry.Labels = container.Labels;
+    entry.LocalVolumes = std::to_string(container.LocalVolumes);
+    entry.Mounts = container.Mounts;
```
ptrivedi: unless --no-trunc is specified, docker seems to truncate each mount name to a certain column width. Here we would end up always assigning the full joined string

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2562,6 +2571,47 @@ try
         THROW_HR_IF(E_UNEXPECTED, strcpy_s(output[index].Image, e->Image().c_str()) != 0);
         THROW_HR_IF(E_UNEXPECTED, strcpy_s(output[index].Name, e->Name().c_str()) != 0);
         THROW_HR_IF(E_UNEXPECTED, strcpy_s(output[index].Id, e->ID().c_str()) != 0);
+
+        // Commands and status descriptions have no bound imposed by the runtime, so truncate them
+        // rather than failing the listing.
+        CopyTruncated(output[index].Command, dockerContainer.Command);
+        CopyTruncated(output[index].Status, dockerContainer.Status);
```
ptrivedi: are these fixed-size truncations intentional? I think --no-trunc will not be able to return the complete values once it gets truncated here, same is true for the labels, networks, etc. below

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -396,37 +396,116 @@ int ContainerService::Attach(Terminal& terminal, Session& session, const std::st
...
+    constexpr size_t c_maxDisplayWidth = 20;
+
+    auto wide = wsl::shared::string::MultiByteToWide(command);
+    if (truncate)
+    {
+        // Count code points rather than code units so a surrogate pair is never split.
+        size_t codePoints = 0;
+        size_t index = 0;
+        size_t cutoff = 0;
+        for (; index < wide.size(); ++codePoints)
```
ptrivedi: will counting code points here correctly shorten wide characters? what does docker do? could we also add a test for wide chars?

ptrivedi (commented):
thank you for the changes, a few minor comments

src/windows/wslc/services/ContainerService.cpp
ggarzia-MSFT: Looking into this, I believe we currently have a divergence from docker. I will check what docker does.

src/windows/wslc/services/ContainerService.cpp
ggarzia-MSFT: Good catch, will change this

src/windows/wslc/services/ContainerService.cpp
ggarzia-MSFT: Pulled from the docker repo:
```go
func charWidth(r rune) int {
    switch width.LookupRune(r).Kind() {
    case width.EastAsianWide, width.EastAsianFullwidth:
        return 2
    case width.Neutral, width.EastAsianAmbiguous, width.EastAsianNarrow, width.EastAsianHalfwidth:
        return 1
    default:
        return 1
    }
}
```
Go has a builtin rune type that docker uses which represents a single 32bit Unicode code point. These can have different sizes in the terminal. You are correct that the current solution does not handle those characters that take up two columns in the terminal correctly. After some research I believe that I can use the ICU library to get the same behavior. If there is something else that you think would work better, please let me know.

Sorry for the confusion, I will rewrite this using the ICU lib and add a test for wide chars

src/windows/wslc/services/ContainerService.cpp
ggarzia-MSFT: This whole loop needs to be restructured, see my response to Pooja's comment above. But right now, it adds the ellipses (`u/2026`) in the last if block, so breaking after `cutoff = index` would cause that to be skipped. This needs to be done in the loop because a command of 19 code points should not have ellipses appended.

src/windows/wslc/tasks/ContainerTasks.cpp
ggarzia-MSFT: Didn't catch this, thanks

src/windows/wslc/services/ContainerService.cpp
ggarzia-MSFT: I also think there may be a way to abstract this to a single helper in common code, as I believe that docker has the same truncating logic for all cases (commands, mounts, etc)

src/windows/wslc/tasks/ContainerTasks.cpp
ggarzia-MSFT: Yes it can

src/windows/common/WSLCContainerEntry.h
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include <wil/resource.h>
+#include "wslc.h"
+
+namespace wsl::windows::common::wslc {
+
+// IWSLCSession::ListContainers allocates a string for every unbounded field on an entry, so each
+// element owns memory that releasing the array alone would miss.
+struct ContainerEntryDeleter
```
OneBlue: nit: This can be a regular function (like what we do [here](https://github.com/microsoft/WSL/blob/4cd2389160cc6db0038824cef9cd81cb433df336/src/windows/WslcSDK/winrt/Container.h#L50))

ggarzia-MSFT: That example uses `unique_any` which expects a function pointer for the deleter, while this uses `unique_any_array_ptr`, which expects a default constructible struct. However, after looking at the docs, it seems like the best pattern is to wrap the `WSLCContainerEntry` in a `unique_struct`. I will change this so it uses the `unique_struct` implementation

From the [docs](https://github.com/microsoft/wil/wiki/RAII-resource-wrappers#wilunique_any_array_ptr):

> If the elements of the array require individual destruction, create a custom unique_any or unique_struct as necessary.
>
> ```C++
> struct WIDGET_ITEM
> {
>    PWSTR name;
>    IWidget* widget;
> };
>
> void DestroyWidgetItem(WIDGET_ITEM* item)
> {
>   CoTaskMemFree(item->name);
>   if (item->widget) item->widget->Release();
> }
>
> using unique_widget_item = wil::unique_struct<WIDGET_ITEM, decltype(&::DestroyWidgetItem), ::DestroyWidgetItem>;
> wil::unique_cotaskmem_array_ptr<unique_widget_item> itemArray;
> ```

ranm-msft (commented):
Nice work on this - I went through the earlier feedback and it all looks genuinely addressed at head, not just papered over: `Ellipsis()` does real display-width truncation via ICU East Asian width, the quoting is a faithful `strconv.Quote` port (including `IsPrint` semantics and surrogate handling), `FormatMounts` respects `--no-trunc`, health is derived from the status suffix, and the service side dropped the fixed-size `strncpy_s` copies in favour of allocations with a `scope_exit` guard so a throw mid-loop no longer leaks.

One thing I think is still open:

**`--format json` can emit localized values.** `ToContainerOutput()` fills `RunningFor` from `FormatRelativeTime()`, which resolves to `Localization::WSLCCLI_RelativeTime*()`, and `Status` from `FormatStatus()`, whose empty-runtime-status fallback goes through `LocalizedContainerStateName()` plus the same relative-time helper. So the same container can serialize different `Status` / `RunningFor` values depending on UI language. That cuts against the parity goal - docker's formatter renders those strings in invariant English - and the new invariant/localized split currently only protects `State`.

Sharing one representation between the table and JSON is the right instinct, but these two fields are exactly where the two outputs need to differ. Could the JSON path use invariant relative-time and fallback-status formatters?

craigloewen-msft:
@ranm-msft I agree with these concerns. However I think we should keep this not localized for now as there could be programs who might be keying off these values, and so I'd want to remain similar to other container tools. 
When we go GA we will have a note about localization (As some of our error codes will not be localized) and I will make sure to include this behaviour in that note.

craigloewen-msft (approved):

PR #41377, opened 2026-08-19
Match docker's network list, inspect and prune output in wslc

ggarzia-MSFT:

Brings `wslc network list`, `wslc network inspect` and `wslc network prune` output in line with docker. A side-by-side parity harness run against docker showed eleven divergences in the network command; this change addresses all of them except error message wording, which is tracked separately.

## PR Checklist

**`network list`**

- Docker's predefined `bridge`, `host` and `none` networks are now listed. `ListNetworks` queries the daemon directly instead of returning only the session's managed-network map.
- Added the `SCOPE` column, and networks are ordered by name, matching docker.
- Table columns use docker's minimum column width so spacing is identical.
- `--format json` now emits docker's nine all-string keys (`CreatedAt`, `Driver`, `ID`, `IPv4`, `IPv6`, `Internal`, `Labels`, `Name`, `Scope`) via a dedicated `NetworkOutputInformation` model, following the same pattern already used for image list output.
- `CreatedAt` is rendered the way docker renders network timestamps: UTC, with the daemon's fractional-second digits preserved verbatim (`2026-08-18 23:35:57.900958925 +0000 UTC`). This is a new `FormatDockerTimestamp(std::string_view)` overload; the existing `LONGLONG` overload, which converts to local time for image and container output, is unchanged.

**`network inspect`**

- Now a live passthrough of the daemon's inspect response instead of being reconstructed from cached session state, so predefined networks are inspectable and values cannot drift.
- Emits docker's full field set: `Created`, `EnableIPv4`, `EnableIPv6`, `Attachable`, `Ingress`, `ConfigOnly`, `ConfigFrom`, `Containers`, `Status` and `IPAM.Options`.
- `Options` is emitted as an object rather than `null`, and empty `Gateway`/`IPRange` entries are omitted, both matching docker.

**`network prune`**

- Prints docker's `Deleted Networks:` header followed by bare network names and a trailing blank line, and prints nothing when no networks were removed.

**Supporting changes**

- The internal `IWSLCSession::ListNetworks` now returns JSON instead of a fixed-size struct array. The list shape includes variable-length labels, which cannot be marshaled safely inside an out-array of structs. `WSLCNetworkInformation` and its serializer are removed. This interface is internal and ships in lockstep with its only clients; the SDK-facing `WSLCCompat.idl` and the plugin API are untouched.
- `docker_schema::Network` and `docker_schema::IPAM` use hand-written deserializers that tolerate `null`, because the daemon reports some empty maps as `null` on predefined networks and the default deserializer rejects that.
- The internal `com.microsoft.wsl.network.managed` label is stripped from both list and inspect output.
- `ParseDockerTimestamp` moved out of `WSLCContainer.cpp` into `docker_schema::ParseTimestamp` and is now shared.

Deliberately out of scope: error message wording, the global JSON indent width and key ordering (shared by container, image and volume inspect), and `network prune --force` (`-f` is already the `--filter` alias and no wslc prune command prompts).

## Validation Steps Performed

- Full x64 Debug build clean; clang-format 19.1.5 clean.
- Deployed to the host and re-ran the wslc-vs-docker parity harness (81 command pairs). All network cases now match docker: `network list` (table, `--format json`, `-q`, `--no-trunc`, `-f driver=bridge`) is byte-identical apart from the daemon-assigned IDs, `network inspect` matches field for field, and `network prune` matches exactly.
- Remaining inspect differences are daemon-version artifacts, not CLI behavior: the host daemon populates `Status.IPAM` and `com.docker.network.enable_ipv4/6` options that the daemon in the WSL VM does not report. Both are passed through unchanged.
- Tests updated and added: service-level list tests reworked for the JSON API and predefined networks; new E2E coverage for predefined networks in list and inspect, the nine-key JSON shape, the UTC timestamp suffix, the new inspect fields, and the prune output format.

### DIFF
--- a/src/windows/common/string.cpp
+++ b/src/windows/common/string.cpp
@@ -426,7 +426,17 @@ std::string wsl::windows::common::string::TruncateId(_In_ std::string_view id, b
     return TruncateIdImpl(id, shortenLength);
 }
 
-std::string wsl::windows::common::string::FormatDockerTimestamp(LONGLONG timestamp)
+std::uint64_t wsl::windows::common::string::Rfc3339ToEpoch(const std::string& timestamp)
+{
+    std::chrono::sys_seconds utcSeconds;
+    std::istringstream stream(timestamp);
+    stream >> std::chrono::parse("%FT%H:%M:%S%Z", utcSeconds);
+    THROW_HR_IF_MSG(E_INVALIDARG, stream.fail(), "Failed to parse timestamp '%hs'", timestamp.c_str());
+
+    return static_cast<std::uint64_t>(utcSeconds.time_since_epoch().count());
+}
+
+std::string wsl::windows::common::string::EpochToLocalDisplayTime(LONGLONG timestamp)
 {
     const auto time =
         std::chrono::floor<std::chrono::seconds>(std::chrono::system_clock::from_time_t(static_cast<std::time_t>(timestamp)));
@@ -444,7 +454,7 @@ std::string wsl::windows::common::string::FormatDockerTimestamp(LONGLONG timesta
     }
 }
 
-std::string wsl::windows::common::string::FormatDockerTimestamp(std::string_view timestamp)
+std::string wsl::windows::common::string::Rfc3339ToUtcDisplayTime(std::string_view timestamp)
 {
     if (timestamp.empty())
     {
--- a/src/windows/common/string.cpp
+++ b/src/windows/common/string.cpp
@@ -16,6 +16,7 @@ Abstract:
 #include <charconv>
 #include <cmath>
 #include <limits>
+#include <sstream>
 
 std::vector<std::string> wsl::windows::common::string::InitializeStringSet(_In_count_(BufferSize) LPCSTR Buffer, _In_ SIZE_T BufferSize)
 {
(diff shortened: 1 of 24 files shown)


src/windows/common/string.cpp
```diff
@@ -442,3 +453,38 @@ std::string wsl::windows::common::string::FormatDockerTimestamp(LONGLONG timesta
...
+        }
+
+        fraction = parsable.substr(separator, end - separator);
+        parsable.erase(separator, end - separator);
+    }
+
+    std::chrono::sys_seconds parsed{};
+    std::istringstream stream(parsable);
+    stream >> std::chrono::parse("%FT%H:%M:%S%Z", parsed);
+    if (stream.fail())
```
OneBlue, line 483: If we fail to parse I think we should throw an exception here, otherwise this could mask errors

OneBlue: I guess the easiest way to solve this would be to call `Rfc3339ToEpoch` here

OneBlue (approved):
LGTM. Feel free to resolve the minor comments in a followup PR

src/windows/common/string.cpp
ggarzia-MSFT: Will address in follow up

Issue #12599, opened 2025-02-18
WSL crashes when distro is stored on BitLocker drive


sciencebug:
### Windows Version

Microsoft Windows [Version 10.0.26100.3194]

### WSL Version

2.4.11.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

5.15.167.4-1

### Distro Version

Ubuntu 24.04

### Other Software

BitLocker

### Repro Steps

- create a virtual disk (DiskManagement -> Create VHD) and mount it
- enable BitLocker for the mounted drive
- install any distro (e.g. ubuntu 24.04, named _alice_) from rootfs file
- choose the unlocked BitLocker drive as the target folder (WSL distro _ext4.vhdx_ will be stored there)

i take extra care to terminate the WSL distro (`wsl -t alice`) before unmounting the drive,
and only starting it again after having the drive mounted and unlocked.
still, WSL crashes...

### Expected Behavior

- distros on the BitLocker drive wont launch when the drive is unmounted or locked
- other distros work as usual
- WSL does not become unresponsive or crash

### Actual Behavior

- WSL becomes unresponsive (all distros!)
- new terminal windows remain black and empty
- after some minutes the terminal window closes
- only a windows restart fixes the problem

sciencebug:
logs have been added.

also, `wsl --shutdown` hangs without output (like everything else WSL-related),
but then after ~10mins seems to complete and fix the problem

OneBlue:
Thank you @sciencebug. Just to make sure I understand the issue, does this happen when the bitlocker drive is unlocked, or only when it's locked ?

crramirez:
Hello @sciencebug ,

I guest that any other variable is in the equation to produce those crashes. I have a drive created using Storage Spaces, it is encrypted with bitlocker and I store my vhdx file there. I work daily with this configuration and no crashes.

I also have a laptop which is fully encrypted with bitlocker and WSL works fluently.

Maybe is the installation via the rootfs. Can you elaborate more on the steps?

I think that the title is too generic, because in general distros stored on BitLocker drives do work.

Best

PR #41379, opened 2026-08-19
Fix: Reattach stale cached device after backing volume is reattached

chemwolf6922:

The current disk attach caching logic only checks the vhd path. When the backing volume is detached than reattached, the check will pass but the stale reference won't work inside the VM. 
The backing volume reattaching can happen after bitlocker locking and unlocking #12599, or vhdx detaching then reattaching.

This PR checks if the cached vhd is stale by holding a handle to the vhd file. And checks if it's still valid before reuse. If not, the vhd will be reattached.

## PR Checklist

- [x] **Closes:** #12599

## Validation Steps Performed

The proposed regression test was removed because one of its diskpart commands did not work correctly on the pipeline. 
The change was tested manually with these steps:
1. Run distro B to keep the VM running.
2. Run distro A. Which has its distro vhd inside another vhd.
3. Terminate distro A.
4. Detach the outer vhd.
5. Attach the outer vhd.
6. Run distro A. Without the fix, this will fail.


OneBlue (approved):
This is tough because there's no way to atomically check that the disk is actually available before we mount it through HCS, which will reopen the handle on its end.

So there's a race between time we check, and then time HCS actually mounts the file, but this is still an improvement in behavior, so we should merge it. Thank you for looking into this !

PR #41385, opened 2026-08-19
Bound WslCoreInstance init transactions

### LINKED ISSUE #41384: WSLService can become unresponsive when init transactions wait indefinitely under session locks
# WSLService can become unresponsive when init transactions wait indefinitely under session locks

## Summary

When the Linux guest stops responding, process launches can wait indefinitely. A concurrent shutdown can then block behind the same instance lock while holding the session lock, causing distribution enumeration, shutdown, and service termination to stop progressing.

Two request/response operations and one send operation in `WslCoreInstance` use the infinite default `SocketChannel` timeout while holding locks needed by other session operations.

## Impact

If the guest remains connected but stops responding, the affected operation can wait indefinitely while holding locks needed by shutdown and other session operations. This can prevent distribution enumeration, shutdown, and service termination from progressing.

## Code analysis

On Windows, `SocketChannel::DefaultSocketTimeout` is `INFINITE`.

The following operations omitted the existing `m_socketTimeout` value:

```cpp
sessionLeader->GetChannel().Transaction<LX_INIT_CREATE_PROCESS_UTILITY_VM>(messageSpan)
m_initChannel->GetChannel().StartTransaction()
```

The affected paths are:

- `WslCoreInstance::CreateLxProcess()` while holding `WslCoreInstance::m_lock`;
- `WslCoreInstance::Initialize()` while holding `m_lock`, with its caller also holding `LxssUserSessionImpl::m_instanceLock`;
- `WslCoreInstance::UpdateTimezone()` while holding the channel lock, with its caller holding `m_instanceLock`.

`RequestStop()` and `Stop()` also require `m_lock`. A guest that does not answer a process-creation transaction can therefore prevent the shutdown path from acquiring the lock and progressing.

The same class already uses `m_socketTimeout` for comparable init operations. The value comes from `DistributionStartTimeout`, which defaults to 60 seconds and is configurable.

## Expected behavior

A non-responsive guest may cause the individual operation to fail, but it must not leave `WSLService`, shutdown, enumeration, or unrelated session operations blocked indefinitely.

## Proposed change

Pass `m_socketTimeout` to all three transactions:

```cpp
Transaction<LX_INIT_CREATE_PROCESS_UTILITY_VM>(messageSpan, nullptr, m_socketTimeout)
StartTransaction(m_socketTimeout)
```

This applies the timeout policy already used by the surrounding code and does not introduce a new timeout value.

## Tests

The added unit tests verify that:

- the peer receives the transaction request;
- a peer that remains connected without responding produces `HRESULT_FROM_WIN32(ERROR_TIMEOUT)` within a bounded interval;
- a valid response received before the deadline succeeds normally.

These tests validate `SocketChannel::Transaction` timeout propagation and behavior. They do not constitute a complete reproduction using a real `WslCoreInstance` or Linux guest.

SylvainM98:

`SocketChannel` defaults to an infinite timeout when none is supplied. Three `WslCoreInstance` operations rely on that default while holding locks needed by shutdown or other session operations. If the guest stays connected but stops responding, those operations can wait indefinitely and prevent `WslService` from making progress.

This passes the instance's existing `m_socketTimeout` to all three operations.

## PR Checklist

- [x] **Closes:** Closes #41384
- [x] **Localization:** All end user facing strings can be localized — no user-facing strings were changed
- [x] **Dev docs:** Added/updated if needed — no developer documentation changes are required

On Windows, `DefaultSocketTimeout` is `INFINITE`, so a `SocketChannel` transaction that does not receive an explicit timeout can wait without bound.

`WslCoreInstance` already owns the applicable timeout value: `m_socketTimeout`, sourced from `DistributionStartTimeout`, which defaults to 60 seconds and is configurable through `.wslconfig` as `wsl2.distributionStartTimeout`.

The class already applies this timeout to comparable operations:

| Operation | Bounded call |
|---|---|
| Instance creation result | `ReceiveMessage(..., m_socketTimeout)` |
| DrvFs mounting | `Transaction(..., m_socketTimeout)` |
| Instance termination | `StartTransaction(m_socketTimeout)` |
| Linux process creation | `CreateLinuxProcess(..., m_socketTimeout)` |
| Session creation | `Transaction(..., m_socketTimeout)` |

The timeout was omitted from three operations:

| Function | Lock held |
|---|---|
| `CreateLxProcess()` | `WslCoreInstance::m_lock` and the session-leader channel lock |
| `Initialize()` | `WslCoreInstance::m_lock`; its caller holds `LxssUserSessionImpl::m_instanceLock` |
| `UpdateTimezone()` | The init-channel lock; its caller holds `m_instanceLock` |

A process-creation transaction can therefore hold `WslCoreInstance::m_lock` indefinitely. `RequestStop()` and `Stop()` require the same lock, while their callers hold `LxssUserSessionImpl::m_instanceLock`. This can prevent shutdown and other session operations from progressing.

If service or session teardown subsequently enters `ClearSessionsAndBlockNewInstancesLockHeld()`, it holds `g_sessionTerminationLock` while waiting for shutdown. New COM session activation can then also stop progressing.

This change passes `m_socketTimeout` to the three omitted operations. When the timeout expires, the affected operation fails, unwinds, and releases its locks. It does not change the timeout value or introduce a new policy.

This removes three unbounded waits. It does not address other independent causes of guest or service instability. Other unbounded `m_miniInitChannel` operations in `WslCoreVm` are intentionally left out because each requires a separate timeout and cancellation analysis.

## Validation Steps Performed

Extended `UnitTests::SocketChannel` with two cases:

- A connected peer that remains open without replying makes the transaction fail with `HRESULT_FROM_WIN32(ERROR_TIMEOUT)`. The elapsed time is checked to verify that the configured deadline was honored rather than the operation returning immediately.
- A peer that replies before the deadline completes normally and returns the expected payload.

Static inspection confirmed that the test helpers and assertions are available in the existing test scope.

The Windows build and tests were not run locally. Build, `clang-format`, and test validation are pending repository CI.

### DIFF
--- a/test/windows/UnitTests.cpp
+++ b/test/windows/UnitTests.cpp
@@ -7711,60 +7711,6 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
             const auto hr = wil::ResultFromException([&]() { channel.ReceiveMessage<RESULT_MESSAGE<int32_t>>(nullptr, 100); });
             VERIFY_ARE_EQUAL(hr, HRESULT_FROM_WIN32(ERROR_TIMEOUT));
         }
-
-        {
-            constexpr DWORD transactionTimeout = 200;
-            constexpr LONGLONG schedulingTolerance = 50;
-            constexpr LONGLONG maximumWait = 5000;
-            auto [client, server] = MakeSocketPair();
-            wsl::shared::SocketChannel channel{std::move(client), "client"};
-            auto transaction = channel.StartTransaction(transactionTimeout);
-
-            RESULT_MESSAGE<int32_t> request{};
-            request.Header.MessageType = RESULT_MESSAGE<int32_t>::Type;
-            request.Header.MessageSize = sizeof(request);
-            transaction.Send(request);
-
-            RESULT_MESSAGE<int32_t> receivedRequest{};
-            recvAll(server.get(), &receivedRequest, sizeof(receivedRequest));
-            VERIFY_ARE_EQUAL(receivedRequest.Header.TransactionStep, static_cast<unsigned int>(TRANSACTION_STEP::REQUEST));
-
-            const auto start = std::chrono::steady_clock::now();
-            const auto hr = wil::ResultFromException([&]() { transaction.Receive<RESULT_MESSAGE<int32_t>>(); });
-            const auto elapsed =
-                std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start).count();
-
-            VERIFY_ARE_EQUAL(hr, HRESULT_FROM_WIN32(ERROR_TIMEOUT));
-            VERIFY_IS_GREATER_THAN_OR_EQUAL(elapsed, static_cast<LONGLONG>(transactionTimeout) - schedulingTolerance);
-            VERIFY_IS_LESS_THAN(elapsed, maximumWait);
-        }
-
-        {
-            constexpr DWORD transactionTimeout = 1000;
-            constexpr int32_t expectedResult = 42;
-            auto [client, server] = MakeSocketPair();
-            wsl::shared::SocketChannel channel{std::move(client), "client"};
-            auto transaction = channel.StartTransaction(transactionTimeout);
-
-            RESULT_MESSAGE<int32_t> request{};
-            request.Header.MessageType = RESULT_MESSAGE<int32_t>::Type;
-            request.Header.MessageSize = sizeof(request);
-            transaction.Send(request);
-
-            RESULT_MESSAGE<int32_t> receivedRequest{};
-            recvAll(server.get(), &receivedRequest, sizeof(receivedRequest));
-
-            RESULT_MESSAGE<int32_t> response{};
-            response.Header.MessageType = RESULT_MESSAGE<int32_t>::Type;
-            response.Header.MessageSize = sizeof(response);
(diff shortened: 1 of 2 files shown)


SylvainM98:

test/windows/UnitTests.cpp
```diff
@@ -7711,6 +7711,60 @@ Error code: Wsl/InstallDistro/WSL_E_INVALID_JSON\r\n",
...
+            recvAll(server.get(), &receivedRequest, sizeof(receivedRequest));
+            VERIFY_ARE_EQUAL(receivedRequest.Header.TransactionStep, static_cast<unsigned int>(TRANSACTION_STEP::REQUEST));
+
+            const auto start = std::chrono::steady_clock::now();
+            const auto hr = wil::ResultFromException([&]() { transaction.Receive<RESULT_MESSAGE<int32_t>>(); });
+            const auto elapsed =
+                std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start).count();
+
+            VERIFY_ARE_EQUAL(hr, HRESULT_FROM_WIN32(ERROR_TIMEOUT));
+            VERIFY_IS_GREATER_THAN_OR_EQUAL(elapsed, static_cast<LONGLONG>(transactionTimeout) - schedulingTolerance);
```
OneBlue: These tests don't seem to exercise the code that this PR changes, although I don't recommend having timing based tests like this, since they will fail if the machine is under pressure

PR #41387, opened 2026-08-19
Avoid holding the instance lock while waiting for OOBE

### LINKED ISSUE #41386: WSL2 shutdown can deadlock while process creation waits for OOBE
# WSL2 shutdown can deadlock while process creation waits for OOBE

## Summary

`WslCoreInstance::CreateLxProcess()` waits for an in-progress OOBE while holding `WslCoreInstance::m_lock`. The shutdown path needs the same lock to stop the instance and signal its destruction event.

If another process launch arrives while OOBE is waiting for user input, distribution termination or shutdown can remain blocked until OOBE completes.

## Code analysis

`CreateLxProcess()` acquires `m_lock` before checking `m_oobeCompleteEvent`:

```cpp
std::lock_guard lock(m_lock);

if (m_oobeCompleteEvent && !m_oobeCompleteEvent.is_signaled())
{
    m_oobeCompleteEvent.wait();
}
```

`RequestStop()` and `Stop()` also acquire `m_lock`. `Stop()` signals `m_destroyingEvent`, but it cannot do so while the process-creation thread owns the lock.

The resulting dependency is:

1. A process-creation thread owns `m_lock` and waits for OOBE completion.
2. Shutdown waits for `m_lock` before it can stop the instance.
3. The process-creation wait does not observe `m_destroyingEvent`.

OOBE is intentionally allowed to wait indefinitely for user interaction. That wait should not retain the instance lock or prevent shutdown.

## Expected behavior

An OOBE flow may remain active for an arbitrary duration, but distribution termination and shutdown must still be able to stop the instance.

## Proposed change

Capture stable handles for the OOBE-complete and instance-destroying events while holding `m_lock`, release the lock, and wait for either event. After reacquiring the lock, revalidate that the instance is still running before continuing process creation.

## Proposed validation

Add a WSL2 regression test that:

- starts an OOBE command that remains active;
- starts a second process creation that waits for OOBE;
- terminates the distribution concurrently;
- verifies that termination completes before OOBE's natural completion;
- verifies that the waiting client processes exit.

This report is based on static source analysis and a deterministic regression scenario. No machine logs are attached.

SylvainM98:

Process creation can wait for a distribution's OOBE to complete while holding
`WslCoreInstance::m_lock`. Because `Stop()` requires the same lock before it can
signal `m_destroyingEvent`, terminating the distribution can deadlock until the
OOBE exits on its own.

This change releases the instance lock while waiting and waits for either OOBE
completion or instance destruction. After reacquiring the lock, it validates
that the instance is still running before continuing.

## PR Checklist

- [x] **Closes:** Closes #41386

`WslCoreInstance::CreateLxProcess()` currently acquires `m_lock` before checking
whether OOBE is still running. If it is, the function waits indefinitely on
`m_oobeCompleteEvent` without releasing the lock.

`WslCoreInstance::Stop()` acquires `m_lock` before setting
`m_destroyingEvent`. This creates the following cycle:

1. Process creation holds `m_lock` and waits for OOBE completion.
2. Distribution termination waits for `m_lock` in `Stop()`.
3. `Stop()` cannot signal `m_destroyingEvent` while process creation holds the lock.

The wait now uses duplicated event handles captured while the instance state is
protected, releases `m_lock`, and waits for either OOBE completion or instance
destruction. It then reacquires the lock and revalidates `m_initChannel` and
`m_consoleManager` before using them.

## Validation Steps Performed

Added `ModernOOBETermination`, a WSL2 regression test that:

1. Starts an OOBE command that signals readiness and remains active.
2. Starts a second process that waits for OOBE completion.
3. Terminates the distribution while that process is waiting.
4. Verifies termination completes before the OOBE command's natural exit.
5. Verifies both waiting processes exit during cleanup.

Static inspection was performed locally. The Windows build, formatting check,
and test execution are left to repository CI because the integration tests
affect machine-wide WSL state.

### DIFF
--- a/src/windows/service/exe/WslCoreInstance.cpp
+++ b/src/windows/service/exe/WslCoreInstance.cpp
@@ -180,10 +180,8 @@ void WslCoreInstance::CreateLxProcess(
     {
         EMIT_USER_WARNING(wsl::shared::Localization::MessageWaitingForOobe(m_configuration.Name.c_str()));
 
-        const wil::unique_handle oobeCompleteEvent{
-            wsl::windows::common::wslutil::DuplicateHandle(m_oobeCompleteEvent.get(), SYNCHRONIZE)};
-        const wil::unique_handle destroyingEvent{
-            wsl::windows::common::wslutil::DuplicateHandle(m_destroyingEvent.get(), SYNCHRONIZE)};
+        const auto oobeCompleteEvent = m_oobeCompleteEvent;
+        const auto destroyingEvent = m_destroyingEvent;
         const HANDLE waitHandles[] = {oobeCompleteEvent.get(), destroyingEvent.get()};
         lock.unlock();
         const DWORD waitResult = WaitForMultipleObjects(RTL_NUMBER_OF(waitHandles), waitHandles, FALSE, INFINITE);
--- a/src/windows/service/exe/WslCoreInstance.cpp
+++ b/src/windows/service/exe/WslCoreInstance.cpp
@@ -173,13 +173,24 @@ void WslCoreInstance::CreateLxProcess(
 
     // Ensure the instance is still running.
 
-    std::lock_guard lock(m_lock);
+    std::unique_lock lock(m_lock);
     THROW_HR_IF(HCS_E_TERMINATED, (!m_initChannel || !m_consoleManager));
 
     if (m_oobeCompleteEvent && !m_oobeCompleteEvent.is_signaled())
     {
         EMIT_USER_WARNING(wsl::shared::Localization::MessageWaitingForOobe(m_configuration.Name.c_str()));
-        m_oobeCompleteEvent.wait();
+
+        const wil::unique_handle oobeCompleteEvent{
+            wsl::windows::common::wslutil::DuplicateHandle(m_oobeCompleteEvent.get(), SYNCHRONIZE)};
+        const wil::unique_handle destroyingEvent{
+            wsl::windows::common::wslutil::DuplicateHandle(m_destroyingEvent.get(), SYNCHRONIZE)};
+        const HANDLE waitHandles[] = {oobeCompleteEvent.get(), destroyingEvent.get()};
+        lock.unlock();
+        const DWORD waitResult = WaitForMultipleObjects(RTL_NUMBER_OF(waitHandles), waitHandles, FALSE, INFINITE);
+        THROW_LAST_ERROR_IF(waitResult == WAIT_FAILED);
+        lock.lock();
+        THROW_HR_IF(HCS_E_TERMINATED, waitResult == WAIT_OBJECT_0 + 1 || !m_initChannel || !m_consoleManager);
+        THROW_HR_IF(E_UNEXPECTED, waitResult != WAIT_OBJECT_0);
     }
 
     // Initialize the create process message.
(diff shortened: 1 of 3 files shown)


src/windows/service/exe/WslCoreInstance.cpp
```diff
@@ -173,13 +173,24 @@ void WslCoreInstance::CreateLxProcess(
...
-    std::lock_guard lock(m_lock);
+    std::unique_lock lock(m_lock);
     THROW_HR_IF(HCS_E_TERMINATED, (!m_initChannel || !m_consoleManager));
 
     if (m_oobeCompleteEvent && !m_oobeCompleteEvent.is_signaled())
     {
         EMIT_USER_WARNING(wsl::shared::Localization::MessageWaitingForOobe(m_configuration.Name.c_str()));
-        m_oobeCompleteEvent.wait();
+
+        const wil::unique_handle oobeCompleteEvent{
```
OneBlue: nit: Instead of duplicating the handles, we could switch those to be `wil::shared_handle`, and then just make a copy in this block

test/windows/UnitTests.cpp
```diff
@@ -4747,6 +4747,57 @@ VERSION_ID="Invalid|Format"
...
+
+        wsl::windows::common::SubProcess waitingProcessBuilder(
+            nullptr, LxssGenerateWslCommandLine(L"-d " LXSS_DISTRO_NAME_TEST_L L" true").c_str());
+        const auto waitingProcess = waitingProcessBuilder.Start();
+        const DWORD waitingResult = WaitForSingleObject(waitingProcess.get(), waiterObservationTimeout);
+
+        wsl::windows::common::SubProcess terminationProcessBuilder(
+            nullptr, LxssGenerateWslCommandLine(L"--terminate " LXSS_DISTRO_NAME_TEST_L).c_str());
+        const auto terminationProcess = terminationProcessBuilder.Start();
+        const DWORD terminationResult = WaitForSingleObject(terminationProcess.get(), terminationTimeout);
```
OneBlue: nit: I would recommend just calling `terminationProcessBuilder.Run(terminationTimeout);`. This will throw an exception if the process times out, which will fail the test

OneBlue (commented):
Thank you for fixing this ! This approach looks good, couple minor comments

src/windows/service/exe/WslCoreInstance.cpp
```diff
@@ -173,13 +173,22 @@ void WslCoreInstance::CreateLxProcess(
...
     if (m_oobeCompleteEvent && !m_oobeCompleteEvent.is_signaled())
     {
         EMIT_USER_WARNING(wsl::shared::Localization::MessageWaitingForOobe(m_configuration.Name.c_str()));
-        m_oobeCompleteEvent.wait();
+
+        const auto oobeCompleteEvent = m_oobeCompleteEvent;
+        const auto destroyingEvent = m_destroyingEvent;
+        const HANDLE waitHandles[] = {oobeCompleteEvent.get(), destroyingEvent.get()};
+        lock.unlock();
+        const DWORD waitResult = WaitForMultipleObjects(RTL_NUMBER_OF(waitHandles), waitHandles, FALSE, INFINITE);
```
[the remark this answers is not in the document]
OneBlue: Hmm this is true. The easiest solution I can think of would be to only create `m_oobeCompleteEvent` if it isn't already created

PR #41395, opened 2026-08-20
Add timeout for udp dns requests to avoid resource leak

chemwolf6922:

The tunneled UDP DNS request could leak if the Windows side produces no response. 

This PR adds a 60s timeout for each request to avoid this potential leak.

## PR Checklist

## Validation Steps Performed

Tested manually:
1. Add temporary periodic size logging for m_udpRequests.
2. Enabled DNS tunneling and configured a nonexistent external-interface constraint so Windows returned no responses.
3. Sent 1,000 paced UDP DNS queries.
4. Confirmed the map reached 1,000 entries.

Before the fix: 
5. Waited over 70 seconds and confirmed the map remained above 1,000 and continued growing.

After the fix:
5. Confirmed requests were removed approximately 60 seconds after insertion.

### DIFF
--- a/src/linux/init/DnsServer.cpp
+++ b/src/linux/init/DnsServer.cpp
@@ -5,7 +5,6 @@
 #include <netinet/in.h>
 #include <sys/socket.h>
 #include "DnsServer.h"
-#include "RuntimeErrorWithSourceLocation.h"
 #include "Syscall.h"
 #include "util.h"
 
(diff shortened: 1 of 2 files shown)


OneBlue (approved):
Indeed looking through `DnsResolver::HandleDnsQueryCompletion` it's looking like there are code paths where we don't actually send an error back to the client. 

Instead of handling failures on the client side through, I think a better path would be to return an error response to the linux side so it can cleanup its resources. That way the timeout is 100% driven by Windows.

Approving since the current change is an improvement over the current behavior

PR #41400, opened 2026-08-20
Expose kernel headers and perf to distributions

benhillis:
## Summary

- mount kernel-matched headers from the unified artifacts VHD at `/usr/src/linux-headers-$(uname -r)` and point `/lib/modules/$(uname -r)/build` to them
- mount kernel-matched perf tooling at `/usr/lib/linux-tools/$(uname -r)`, add it to the default `PATH`, and configure `PERF_EXEC_PATH`
- shadow an existing distro `/usr/bin/perf` with the matching bundled binary without replacing the distro file
- add WSL2 coverage for header compilation, perf execution, stale build links, and symlinked distro perf binaries

Follow-up to #41267.

## Validation

- `cmake --build . -- -m`
- `bin\x64\debug\test.bat /name:*KernelArtifacts*`
- `mkdocs build -f doc\mkdocs.yml`
- `.\FormatSource.ps1`


test/windows/UnitTests.cpp
```diff
@@ -3003,6 +3003,100 @@ Error code: Wsl/InstallDistro/WSL_E_DISTRO_NOT_FOUND
...
+            LxsstuLaunchWsl(L"test -s /lib/modules/$(uname -r)/build/include/linux/version.h", nullptr, nullptr, nullptr, nullptr), 0u);
+
+        // Headers are usable: compile and run a tiny program that includes recent uapi headers. The
+        // identifiers below fail to compile if the headers are missing or too old (BPF_PROG_TYPE_NETFILTER
+        // added in 6.4, IORING_OP_FUTEX_WAKE added in 6.7). Their numeric values are not a stable API
+        // contract, so the program only checks that <linux/version.h> matches the running kernel.
+        VERIFY_ARE_EQUAL(
+            LxsstuLaunchWsl(
+                LR"BASH(bash -ec '
+                    d=$(mktemp -d)
```
OneBlue, line 3025: nit: I think this would be a bit easier to read if we just built in the current directory. We can just remove `t` in scope exit

OneBlue (approved):
LGTM, one minor comment

PR #41403, opened 2026-08-20
Consolidate timestamp parsing and support docker-style options

ggarzia-MSFT:

Follow-up to #41377. Consolidates all timestamp and duration parsing used by `wslc` into `wsl::windows::common::string`, fixes a parser that silently swallowed errors, and broadens the `--since` / `--until` grammar to match what users expect from an OCI container CLI.

## PR Checklist

**Silent parse failures (review feedback from #41377)**

`Rfc3339ToUtcDisplayTime` used a `TryParse`-style helper and fell back to echoing the raw input when parsing failed, so a malformed daemon timestamp surfaced as a garbled display string instead of an error. It now delegates to `Rfc3339ToEpoch`, which throws. The `TryParse` variant is deleted.

**Parser consolidation**

`SpecParsing.cpp` had grown its own copies of RFC 3339 and Go-duration parsers, one of which duplicated logic already present a few hundred lines away in the same file. Everything now lives in `src/windows/common/string.{hpp,cpp}`:

| Helper | Purpose |
| --- | --- |
| `Rfc3339ToEpoch` | Strict RFC 3339 → epoch seconds. Single source of validation. |
| `ExpandToRfc3339` | Canonicalizes partial/zone-less user input into strict RFC 3339. |
| `TryParseDuration` | Go-style duration (`10m`, `1h30m`, `1.5h`) → `std::chrono::nanoseconds`. |
| `EpochToLocalDisplayTime` | Epoch → local display string. |
| `Rfc3339ToUtcDisplayTime` | RFC 3339 → UTC display string. |

The loose CLI grammar is implemented by canonicalizing input and then handing it to the strict parser, so validation and error messages exist in exactly one place. `ServiceMain.cpp` also picks up `Rfc3339ToEpoch`, which fixes a latent bug where it used `mktime` (local time) on a UTC timestamp.

**Broader `--since` / `--until` input**

`GetTimestampFromString` (used only by `--since` / `--until`) now tries, in order: raw integer epoch → Go duration → `Rfc3339ToEpoch(ExpandToRfc3339(...))`. Newly accepted:

- Zone-less local times — `2024-01-02T15:04:05` (interpreted in the local zone)
- Date-only and partial times — `2024-01-02`, `2024-01-02T15`, `2024-01-02T15:04`
- Go durations — `10m`, `1h30m`, `1.5h` (relative to now)
- Pre-1970 timestamps

Still rejected: `10x`, `1h30`, `2024-13-01`, `2024-02-31`, and other malformed values. Note that `std::chrono::parse` silently wraps invalid calendar dates (Feb 31 → Mar 2), so `Rfc3339ToEpoch` pre-validates via `std::chrono::year_month_day::ok()`.

**Signed epoch plumbing**

Supporting pre-1970 values required `Since` / `Until` to change from `ULONGLONG` to `LONGLONG` through `wslc.idl` → `WSLCContainer` → `DockerHTTPClient` → `ContainerService` → `ContainerTasks` → `ArgumentDefinitions.h`. `Tail` stays `ULONGLONG`. This touches `wslc.idl`, which is internal and explicitly non-ABI-stable (see the comment at the top of that file); the SDK-facing `WSLCCompat.idl` and the public `WslPluginApi.h` are untouched.

**Sentinel handling**

`0001-01-01T00:00:00Z` means "unset" in daemon output but "from the beginning of time" as CLI input. A `c_unsetTimestamp` constant in `docker_schema.h` guards the daemon-side call site; the parser itself simply returns `-62135596800`.

## Validation Steps Performed

- Full `cmake --build .` from a clean tree — 0 errors across all targets.
- `FormatSource.ps1` / clang-format 19.1.5 run over all changed files.
- Unit tests in `test/windows/wslc/WSLCCLIArgumentUnitTests.cpp` extended from 3 to 7 `ValidateTimestamp_*` cases, covering the newly accepted forms (`_ValidPreEpoch`, `_ValidZoneLessLocalTime`, `_ValidPartialAndDateOnly`, `_ValidGoDuration`) and additional rejections (`10x`, `1h30`) in `_InvalidRfc3339_Rejected`. Existing epoch literals updated from `ULL` to `LL`.


src/windows/common/string.hpp
```diff
@@ -67,16 +69,29 @@ std::string WideToMultiByte(_In_ std::wstring_view Source);
 std::wstring TruncateId(_In_ std::wstring_view id, bool shortenLength = true);
 std::string TruncateId(_In_ std::string_view id, bool shortenLength = true);
 
-// Converts an RFC 3339 timestamp to seconds since the unix epoch. Only the 'Z' zone designator is
-// accepted; numeric offsets are not.
-std::uint64_t Rfc3339ToEpoch(const std::string& timestamp);
+// Expands a partial timestamp into a full RFC 3339 one. Hour-only, minute-only and date-only values
```
ggarzia-MSFT: @OneBlue we have a bunch of timestamp related code here, thoughts on putting it in its own file?

OneBlue: I think moving them all to a single file would be a good idea !

OneBlue (approved):

Issue #41282, opened 2026-08-07
wslc: support wslc system info


paulmartrencharpro:
**Is your feature request related to a problem? Please describe.**
docker info (https://docs.docker.com/reference/cli/docker/system/info/) is used by one of my company internal tool to check the workstation installation before running it.
So, I cannot replace docker desktop with wslc for this.

**Describe the solution you'd like**
Implement "wslc system info" - https://docs.docker.com/reference/cli/docker/system/info/

paulmartrencharpro:
Logs (in french):

> \> wslc system info
> Copyright (c) Microsoft Corporation. Tous droits réservés.
> Pour plus d’informations sur la confidentialité de ce produit, veuillez consulter https://aka.ms/privacy.
> 
> Commande non reconnue : 'info'
> Commandes de gestion de niveau système.
> 
> Utilisation : wslc system [<commande>] [<options>]
> 
> Les sous-commandes suivantes sont disponibles :
>   session  Gérer les sessions.
> 
> Pour en savoir plus sur une commande spécifique, passez-la à l’argument aide. [-?]
> 
> Les options suivantes sont disponibles :
>   -?,--help  Affiche l’aide sur la commande sélectionnée


Compared to docker desktop:

> \> docker info
> Client:
>  Version:    29.6.2
>  Context:    desktop-linux
>  Debug Mode: false
>  Plugins:
>   agent: Docker AI Agent Runner (Docker Inc.)
>     Version:  v1.115.0
>     Path:     C:\Users\[username]\.docker\cli-plugins\docker-agent.exe
>   ai: Docker AI Agent - Ask Gordon (Docker Inc.)
>     Version:  v1.27.0
>     Path:     C:\Users\[username]\.docker\cli-plugins\docker-ai.exe
>   buildx: Docker Buildx (Docker Inc.)
>     Version:  v0.35.0-desktop.2
>     Path:     C:\Users\[username]\.docker\cli-plugins\docker-buildx.exe
>   compose: Docker Compose (Docker Inc.)
>     Version:  v5.3.1
>     Path:     C:\Users\[username]\.docker\cli-plugins\docker-compose.exe
>   debug: Get a shell into any image or container (Docker Inc.)
>     Version:  0.0.47
>     Path:     C:\Users\[username]\.docker\cli-plugins\docker-debug.exe
>   desktop: Docker Desktop commands (Docker Inc.)
>     Version:  v0.4.3
[...]

PR #41408, opened 2026-08-21
Add wslc system info command

beena352:
Adds wslc system info (and wslc info as a root alias) that shows client version details and session manager state in both table and JSON formats. The command does not start a VM, it queries the session manager service the same way session list does.

## PR Checklist

- [X] **Closes:** Link to issue #41282 

 What it shows:
- Client block: package version, kernel, direct3d, dxcore, and Windows version, plus the settings file path. Dev builds also show the compiler version, commit hash, and build time
- Server block: session manager version and the list of active sessions (reuses the same table as session list)
Design choices:
- No VM boot. The command only calls IWSLCSessionManager::GetVersion and ListSessions, which don't take a lease. This is the same pattern as session list
- --format json builds the full document before writing, so a service failure fails the whole command. Both table and JSON modes query the service first. A stopped service fails the whole command cleanly with no partial output.
- The session table is hidden when there are no sessions (no empty header row). session list keeps showing the header - that's intentional, they serve different purposes
- The WriteSessionTable helper is shared between ListSessions and ShowSystemInfo in SessionTasks.cpp, which is why there's no separate SystemTasks file
- JSON session entries use "ID" (not "Id") to match the convention in ContainerTasks and NetworkTasks.

What this doesn't do:
- No Docker engine version or container runtime info. That would need a running VM and HTTP client plumbing-separate work
- --session is a global argument and is silently accepted. This matches how other non-session commands behave.

## Validation Steps Performed
- Ran wslc system info and wslc info with 0 and 1 active sessions - output is correct in both cases
- Ran wslc system info --format json and verified valid JSON with Client and Server sections, correct key casing ("ID", not "Id")
- Ran wslc system info --format invalid - clean error message, exit code 1, nothing on stdout
- CLI parse tests: 12 new cases for system info and info with valid/invalid args
- Structural unit tests: 5 new tests covering command name, subcommands, format argument, and registration under both system and root
- E2E tests: 5 new tests (table output, invalid format, JSON validation with key assertions, root alias equivalence, does-not-create-session check)
- Terminal unit test: 1 new test for Flush - verified it goes red when Flush is stubbed out, green with the real implementation

### DIFF
--- a/src/windows/wslc/commands/SystemCommand.cpp
+++ b/src/windows/wslc/commands/SystemCommand.cpp
@@ -12,13 +12,10 @@ Abstract:
 
 --*/
 #include "CLIExecutionContext.h"
-#include "SystemCommand.h"
 #include "SessionCommand.h"
-#include "SessionTasks.h"
-#include "Task.h"
+#include "SystemCommand.h"
 
 using namespace wsl::windows::wslc::execution;
-using namespace wsl::windows::wslc::task;
 using namespace wsl::shared;
 
 namespace wsl::windows::wslc {
@@ -48,28 +45,5 @@ std::wstring SystemCommand::LongDescription() const
 void SystemCommand::ExecuteInternal(CLIExecutionContext& context) const
 {
     OutputHelp(context.Terminal);
-}
-
-// System Info Command
-std::vector<Argument> SystemInfoCommand::GetArguments() const
-{
-    return {
-        Argument::Create(ArgType::Format),
-    };
-}
-
-std::wstring SystemInfoCommand::ShortDescription() const
-{
-    return Localization::WSLCCLI_SystemInfoDesc();
-}
-
-std::wstring SystemInfoCommand::LongDescription() const
-{
-    return Localization::WSLCCLI_SystemInfoLongDesc();
-}
-
-void SystemInfoCommand::ExecuteInternal(CLIExecutionContext& context) const
-{
-    context << ShowSystemInfo;
 }
 } // namespace wsl::windows::wslc
--- a/src/windows/wslc/commands/SystemCommand.cpp
+++ b/src/windows/wslc/commands/SystemCommand.cpp
@@ -14,14 +14,18 @@ Abstract:
 #include "CLIExecutionContext.h"
 #include "SystemCommand.h"
 #include "SessionCommand.h"
+#include "SessionTasks.h"
+#include "Task.h"
 
 using namespace wsl::windows::wslc::execution;
+using namespace wsl::windows::wslc::task;
 using namespace wsl::shared;
 
 namespace wsl::windows::wslc {
 std::vector<std::unique_ptr<Command>> SystemCommand::GetCommands() const
 {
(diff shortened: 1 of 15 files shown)


src/windows/wslc/tasks/SessionTasks.cpp
```diff
@@ -73,20 +94,93 @@ void ListSessions(CLIExecutionContext& context)
...
+        // A JSON document can't be emitted partially, so an unreachable service fails the whole command.
+        const auto managerVersionText = FormatManagerVersion(SessionService::ManagerVersion());
+        const auto sessions = SessionService::List();
+
+        nlohmann::json root;
+
+        auto& client = root["Client"];
+        client["Version"] = std::string{WSL_PACKAGE_VERSION};
+        client["KernelVersion"] = std::string{KERNEL_VERSION};
+        client["WslgVersion"] = std::string{WSLG_VERSION};
```
OneBlue: We shouldn't display wslg and msrdc versions, since they're irrelevant to wslc

OneBlue: (Same for the below bloc)

src/windows/wslc/commands/SystemCommand.cpp
```diff
@@ -45,4 +49,27 @@ void SystemCommand::ExecuteInternal(CLIExecutionContext& context) const
 {
     OutputHelp(context.Terminal);
 }
+
+// System Info Command
```
dkbennett: nit: we want to put a new command implementation in its own .cpp file, but the header definition in the root is OK. Should be SystemInfoCommand.cpp so if the command has creep it doesn't clutter up the root file, simpler merges, etc.

src/windows/wslc/tasks/SessionTasks.cpp
```diff
@@ -73,20 +94,89 @@ void ListSessions(CLIExecutionContext& context)
...
-    table.Complete();
+        if constexpr (!wsl::shared::OfficialBuild)
+        {
+            context.Terminal.Output(L"{}\n", Localization::MessageBuildInfo(_MSC_VER, COMMIT_HASH, __TIME__ " " __DATE__));
+        }
+
+        context.Terminal.Output(L"{}\n", Localization::WSLCCLI_SystemInfoSettingsFile(settingsFilePath));
+
+        // Query the service only after the client block is written so a stopped service still reports client info.
+        context.Terminal.Flush(Terminal::Level::Output);
```
dkbennett: Curious what issue was discovered that required Flush to be added and used here.

beena352: The intent was to ensure the Client block reaches the user even if the service call on the next line throws, since stdout is fully buffered when redirected, the error would print first and the client info would be lost. But looking at it again, the JSON branch handles the same situation by calling the service before any output. I'll remove it and move the service calls above the output to match the JSON branch.

dkbennett (approved):

craigloewen-msft (approved):

PR #41411, opened 2026-08-21
Enable discard for WSLc volumes

JohnMcPMS:
Enable `discard` for primary and secondary WSLc volumes.

## PR Checklist

From an internal report, VHD size would grow with repeated calls to `LoadImage` with the same large tar.  The test in this PR was initially created targeting `debian:latest`, but it wasn't large enough to exhibit the behavior.  Synthetically adding a gigabyte of data into it and saving it did.

The test shows that it isn't additional guest storage that is increasing the VHD, and with `discard` the behavior is resolved.

Enabled `discard` for WSLc volumes, just as it is enabled for WSL:
https://github.com/microsoft/WSL/blob/58c89ade0240f9a43f179d967863fb14920e66a8/src/windows/service/exe/WslCoreVm.cpp#L1261

This causes the guest filesystem to TRIM blocks that are no longer in use, allowing their reuse and preventing the VHD from growing when the guest isn't actually taking up any more space.

## Validation Steps Performed
Added a new test that loads a tar over 1 GB in size repeatedly to ensure that the VHD does not grow in size after the initial load.

### DIFF
--- a/test/windows/WSLCTests.cpp
+++ b/test/windows/WSLCTests.cpp
@@ -1419,19 +1419,80 @@ class WSLCTests
 
     // Loading the same image tar repeatedly must not permanently grow the session storage VHD.
     //
-    // Docker's /images/load handler extracts the entire tar into a temporary directory under its data
+    // Docker's /images/load handler extracts the incoming tar into a temporary directory under its data
     // root (/var/lib/docker, which is the storage VHD) before inspecting any digest, so every call
     // writes roughly one tar's worth of data to the VHD even when the layers already exist and are
     // deduplicated. That temporary directory is deleted afterwards, but the VHD is a non-sparse
     // dynamically expanding VHDX mounted without 'discard', so the freed blocks are never returned to
-    // the host. The result is a VHD that grows by about the tar size on every load.
+    // the host, and ext4 tends to satisfy the next extraction from a different region rather than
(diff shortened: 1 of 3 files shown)


test/windows/WSLCTests.cpp
```diff
@@ -1417,6 +1417,210 @@ class WSLCTests
...
+    // reusing the just-freed one. The result is a VHD that grows by about the tar size on every load.
+    //
+    // The size of the tar matters: a small tar is re-extracted into blocks the VHDX has already
+    // allocated, so the growth plateaus immediately and the bug does not reproduce. This test therefore
+    // builds a large image with incompressible layers (which is what a real from-source application
+    // image looks like once 'save' has written its uncompressed layers out) rather than reusing one of
+    // the small prebuilt test tars.
+    WSLC_TEST_METHOD(LoadImageRepeatedDoesNotGrowStorageVhd)
+    {
+        SKIP_TEST_SERVER();
```
OneBlue, line 1437: What's the reason for skipping this on server ?

OneBlue (approved):

test/windows/WSLCTests.cpp
benhillis: we already skip all WSLC tests on server don't we?

JohnMcPMS: Consistency with several other image loading tests, but not sure why they are skipped on server.

PR #41415, opened 2026-08-21
Poll for the result of operations during MSI installation

OneBlue:

This change updates the custom action logic to poll for the completion of appx operations during the MSI installations.

Calling `.get()` registers a delegate that, due to a cppwinrt lifetime issue, might not get released until `wslinstall.dll` is unloaded, causing a crash since the release will jump into unloaded code. 

## PR Checklist


benhillis (approved):
This definitely seems like a cpp / winrt bug, hopefully we can get a fix or better guidance from them.

PR #41428, opened 2026-08-24
Sign DCAT script

florelis:

The stage that prepares files for DCAT is failing because it publishes an unsigned script as a pipeline artifact.
This change signs that file.

To avoid repeating all the signing parameters, I created a template for the signing step that is reused any time we sign something.

## PR Checklist

## Validation Steps Performed

Currently running a (non official) build for validation.

### DIFF
--- a/.pipelines/build-job.yml
+++ b/.pipelines/build-job.yml
@@ -34,9 +34,6 @@ parameters:
   - name: artifactSuffix
     type: string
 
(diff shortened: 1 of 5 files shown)


florelis:
On second thought, I don't think this script is really worth the trouble of signing given that it is basically a search-and-replace, and that we probably should use a key code for internal use only.

PR #41430, opened 2026-08-24
Match Docker output for prune operations and container inspection

ggarzia-MSFT:

Aligns five `wslc` output surfaces with `docker`, so output can be consumed by tooling written against docker:

- `image prune` — emits a `Deleted Images:` header and lowercase `untagged:` / `deleted:` lines.
- `image delete` — emits `Untagged:` / `Deleted:` lines instead of succeeding silently.
- `container prune` — emits a `Deleted Containers:` header, and drops the leading blank line when nothing was pruned.
- `container inspect` — `Name` carries docker's leading slash, and `Image` reports the image digest rather than the tag.
- `network list` — labels always render as `key=value`, including when the value is empty.

Sizes in the prune reclaimed-space lines now use the same formatter as `volume prune` (`0B` rather than `0 B`).

Note the deliberate casing split between `image prune` and `image delete`: docker uses lowercase in one and capitalized in the other, so these use separate resource strings rather than sharing one.

## PR Checklist

Each change was checked against docker's own source rather than against observed output alone.

**`image prune`** — `cli/command/image/prune.go` writes a `Deleted Images:` header followed by lowercase `untagged: ` / `deleted: `. `PruneImages` emits the header when either list is non-empty and follows with a conditional blank line, matching the shape already used by `PruneVolumes`.

**`image delete`** — `cli/command/image/remove.go` uses capitalized `Untagged:` / `Deleted:`, the opposite of `image prune`. The `DeleteImage` COM call already returned the deleted-image records; the CLI discarded them. `ImageService::Delete` now returns a `std::vector<DeletedImageEntry>` preserving daemon order, and `DeleteImage` renders it. No IDL change was required.

**`container prune`** — `cli/command/container/prune.go` writes the `Deleted Containers:` header only when containers were removed, then bare IDs, a blank line, and the reclaimed-space line. Reclaimed space is only computed inside that branch, so an empty prune prints just the trailing line with no leading blank.

**`container inspect`** — moby's `daemon/inspect.go` sets `Name: ctr.Name` (stored with a leading `/`) and `Image: ctr.ImageID.String()` (the `sha256:` digest). A `QualifyContainerName()` helper sits alongside the existing `CleanContainerName()`. `Config.Image` continues to carry the tag, matching docker.

**`network list`** — `cli/command/network/formatter.go` always joins as `k + "=" + v`, so an empty value yields `k=`. The previous behavior emitted a bare key in that case. Ordering already matched, since `std::map` iterates sorted and docker sorts explicitly.

## Validation Steps Performed

Built clean and formatted with `FormatSource.ps1`. Localization validated with `tools/devops/validate-localization.py`.

Automated:

- `*Prune*` — 38/38 passed
- `*Inspect*` — 60/60 passed (two stale assertions in `WSLCTests.cpp` updated)
- `*ImageDelete*` — 10/10 passed

Manual, compared side by side against docker 29.6.2:

- `container prune` on an empty daemon prints exactly `Total reclaimed space: 0B` with no leading blank line; with a stopped container it prints the `Deleted Containers:` header, the bare ID, a blank line, then the reclaimed-space line.
- `image delete alpine:latest` prints two `Untagged:` lines (tag and digest reference) followed by `Deleted: sha256:…`, matching docker's shape.
- `network create --label bare --label kv=1` then `network list --format json` yields `"Labels":"bare=,kv=1"`.
- `container inspect` reports `Name = /demo-app`, `Image = sha256:…`, and `Config.Image = alpine:latest`.


src/windows/wslc/services/ImageService.cpp
```diff
@@ -367,6 +368,15 @@ void ImageService::Delete(wsl::windows::wslc::models::Session& session, const st
 
     wil::unique_cotaskmem_array_ptr<WSLCDeletedImageInformation> deletedImages;
     THROW_IF_FAILED(session.Get()->DeleteImage(&options, &deletedImages, deletedImages.size_address<ULONG>()));
+
+    std::vector<wsl::windows::wslc::models::DeletedImageEntry> result;
+    result.reserve(deletedImages.size());
+    for (auto ptr = deletedImages.get(), end = deletedImages.get() + deletedImages.size(); ptr != end; ++ptr)
```
OneBlue: nit: This can be iterated over via a ranged based for loop, like: 
`for (auto ptr : deletedImages)`

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -499,6 +499,17 @@ std::string CleanContainerName(const std::string& name)
...
 
+std::string QualifyContainerName(const std::string& name)
+{
+    // Docker reports inspect names with a leading '/'.
+    if (name.empty() || name[0] == '/')
+    {
+        return name;
+    }
+
+    return "/" + name;
```
OneBlue: Does docker always prefix containers with `/` ? If so, could we skip that logic entirely and just returned what docker inspect gave us ?

test/windows/PluginTests.cpp
```diff
@@ -726,7 +726,7 @@ class PluginTests
             WSLCMountFolder(relative): {}
             Test completed
             WSLC Image created, session=*, id=sha256:*, name=debian:latest
-            WSLC Container started, session=*, id=*, name=wslc-plugin-container, image=debian:latest, state=*
+            WSLC Container started, session=*, id=*, name=/wslc-plugin-container, image=debian:latest, state=*
```
OneBlue, line 729: Hmm this is a bit unfortunate because this might be a surprise to some plugins, but getting aligned with docker's schema is the right thing to do. I'll reach out to the MDE team about this

OneBlue (dismissed):

test/windows/PluginTests.cpp
ggarzia-MSFT: OK, do you want me to hold off on merging this into master until after you have talked to the MDE team?

OneBlue: No this can be merged now, I reached out to them and it's not a major concern

OneBlue (approved):

503badrr (commented):
قم بالاصلاح

PR #41435, opened 2026-08-25
Add wslc container restart command

beena352:
Adds the wslc container restart command (and its top-level wslc restart alias). This builds on the backend and COM layer added in #41140 and wires the full CLI vertical: command → task → service → COM.

## PR Checklist

The command is registered in two places, so both wslc container restart and wslc restart
work. It follows the same shape as the other container commands: the command reads the
arguments, hands them to a task, and the task calls the service, which calls into COM.
I reused what was already there instead of adding new things. It shares StopContainerOptions
with the stop command, and it reuses the stop command's --signal description, so there are no
duplicate strings. Only two new strings were added, for the command's short and long help text.
One small difference from stop: ContainerService::Restart uses plain THROW_IF_FAILED. Stop
and start use the _EXCEPT version because they have HRESULTs that just mean "nothing to do",
but restart has none of those, so anything that fails is a real failure.

Why there is a new --timeout argument-

Restart uses --timeout, but stop uses --time because that is what docker uses. I did not
want to rename the existing one and change how stop behaves, so I added ArgType::Timeout next
to ArgType::Time. They share the same number converter and the same description text, so
nothing about container stop --time changes.


src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1426,6 +1426,31 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill)
...
+        // N.B. Stop() and Start() each take m_lock, so it cannot be held across both phases.
+        m_manualRestart = wasRunning;
+    }
+
+    auto restartCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [this]() {
+        auto lock = m_lock.lock_exclusive();
+        m_manualRestart = false;
+    });
+
+    if (wasRunning)
```
OneBlue, line 1542: Unfortunately we're going to need to "lock" across the whole restart, because a separate Stop() could happen after our Stop() returns. 

For a transition like this, we could either use docker's restart API, which I think will do both atomically (but doing that will require some changes to our event handling logic, since IIRC this sends a different event), or to register that we're doing a "restart" transaction, and update the Start() & Stop() logic to stand down if there's a restart in progress

beena352: Good catch, m_manualRestart only suppresses the --rm auto-delete in OnStopped. Once the Stop() phase completes, m_transition is cleared and the container is unguarded until Start() re-enters.
I'll go with the restart transaction rather than Docker's restart API: StateTransition carries a single ExpectedEvent, and a Docker-side restart emits die then start, so that route means reworking the state machine to await a sequence. A restart kind alongside TransitionKind { Start, Stop, Delete } lets WaitForConflictingTransitionToComplete do the standing-down, and OnEvent stays as-is. m_manualRestart then goes away, OnStopped checks "restart in progress" instead.
One question: should a concurrent Stop()/Delete() wait, or fail fast? Waiting matches the other transitions, but for Delete() you may prefer an explicit error.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -1036,6 +1036,28 @@ void StopContainers(CLIExecutionContext& context)
...
+    options.Signal = context.Args.GetValue<ArgType::Signal>(WSLCSignalNone);
+
+    if (context.Args.Contains(ArgType::Time))
+    {
+        options.Timeout = context.Args.GetValue<ArgType::Time>();
+    }
+
+    for (const auto& id : containersToRestart)
+    {
+        ContainerService::Restart(context.Terminal, session, WideToMultiByte(id), options);
```
dkbennett: If one container fails to restart, this will throw here and prevent further processing of containers. 

Docker continues with all containers and joins all errors. So this should be a try-catch here  with the failure output on Error() and continue on to the next container.

Also if there's any failed containers set exit code to nonzero but proceed with all containers best-effort.

dkbennett (commented):
There's an issue where Docker's -t refers to --timeout and "--time" is deprecated and hidden.

Making this match docker will require a CLI core change to arguments to support deprecated/hidden arguments and print a deprecation warning.  I will work on that separately so for now this is fine to not support it, but you should use "--timeout" and not "--time".

I will add an argument option for a deprecation flag so the argument can be used but will be hidden from help and display a warning as Docker does.

src/windows/wslcsession/WSLCContainer.cpp
OneBlue: That's a good question, I would probably want to match docker's behavior there. We should be able to test what docker test by trigger a very slow container restart, and see what happens if a Stop() / Delete() comes while that's in progress

beena352: Tested both cases against moby:
Stop during restart, docker actually has a race here. The stop returns rc=0 but the container ends up running again in most timing windows. I made Stop() wait for the restart to finish instead, so the result is always consistent.
Delete during restart, docker doesn't guard this at all. If the delete lands between the two phases, it just wins and the restart's start fails with "marked for removal." Did the same thing here, Delete() doesn't wait for the restart transaction.
The mechanism is a shared_ptr<RestartTransaction> in m_restart, set for both phases. Start/Stop wait on it, Delete skips it. Each phase still goes through the normal StartTransition path so the state machine is untouched. m_manualRestart is gone, OnStopped checks m_restart instead.

src/windows/wslc/tasks/ContainerTasks.cpp
beena352: Fixed, extracted a ForEachContainerBestEffort helper that catches per-container failures, prints the error to stderr, sets ExitCode = 1, and continues. Applied it to restart, stop, kill, and rm since they all had the same issue.
The error printing is now a shared CLIExecutionContext::OutputError that also clears m_error after printing, without the reset, CollectErrorImpl's same-HRESULT dedup would silently drop the second container's message.

Separately, container rm --force <missing> exits 1 where docker exits 0. That predates this change so I left it alone, happy to fix in a follow-up.

src/windows/wslc/arguments/ArgumentDefinitions.h
```diff
@@ -130,7 +130,7 @@ _(Subnet,           "subnet",               NO_ALIAS,         Kind::Value,
 _(Tail,             "tail",                 L"n",             Kind::Value,      ULONGLONG,    Localization::WSLCCLI_TailArgDescription()) \
 _(Tag,              "tag",                  L"t",             Kind::Value,      NoConversion, Localization::WSLCCLI_TagArgDescription()) \
 _(Target,           "target",               NO_ALIAS,         Kind::Positional, NoConversion, Localization::WSLCCLI_TargetArgDescription()) \
-_(Time,             "time",                 L"t",             Kind::Value,      LONG,         Localization::WSLCCLI_TimeArgDescription()) \
+_(Timeout,          "timeout",              L"t",             Kind::Value,      LONG,         Localization::WSLCCLI_TimeoutArgDescription()) \
 _(TMPFS,            "tmpfs",                NO_ALIAS,         Kind::Value,      ParsedMount,  Localization::WSLCCLI_TMPFSArgDescription()) \
```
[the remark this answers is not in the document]
beena352: Fixed, ArgType::Time is restored for container stop so --time still works as it did before. container restart uses ArgType::Timeout with --timeout. -t works on both. Deprecation/hidden alias support is being handled separately

localization/strings/en-US/Resources.resw
```diff
@@ -3316,7 +3326,10 @@ On first run, creates the file with all settings commented out at their defaults
     <value>New image reference in the image-name[:tag] format</value>
   </data>
   <data name="WSLCCLI_TimeArgDescription" xml:space="preserve">
-    <value>Time in seconds to wait before executing (default 5)</value>
+    <value>Seconds to wait before killing the container (default: the container's configured stop timeout)</value>
+  </data>
+  <data name="WSLCCLI_TimeoutArgDescription" xml:space="preserve">
+    <value>Seconds to wait before killing the container (default: the container's configured stop timeout)</value>
   </data>
```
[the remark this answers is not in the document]
beena352: Fixing this.

src/windows/wslc/core/CLIExecutionContext.cpp
```diff
@@ -23,4 +26,17 @@ void CLIExecutionContext::ApplyGlobalEnvironmentOptions()
     Terminal.SetNoColor(GlobalArgs.GetValue<ArgType::NoColor>());
 }
 
+void CLIExecutionContext::OutputError(HRESULT result)
```
dkbennett: I like how this cleans up Main a bit!

Would recommend renaming it to "ReportError()" however, since "Output" is used to refer to output on stdout in the Terminal class, while errors are on stderr.  Just to keep some name consistency.

src/windows/wslc/core/CLIExecutionContext.cpp
```diff
@@ -23,4 +26,17 @@ void CLIExecutionContext::ApplyGlobalEnvironmentOptions()
...
+{
+    std::wstring message;
+    if (const auto& reported = ReportedError())
+    {
+        const auto strings = wslutil::ErrorToString(*reported);
+        message = strings.Message.empty() ? strings.Code : strings.Message;
+    }
+
+    Terminal.Error(L"{}\n", Localization::MessageErrorCode(message, wslutil::ErrorCodeToString(result)));
+    m_error.reset();
```
dkbennett, line 43: The m_error reset changes main's functionality since it does not currently do this.

I think reporting an error and resetting the error should be explicitly separate actions.  There's an argument that the m_error should be the first reported error, and this always clears the error state so we immediately lose it.

Best solution here would be to add another method on context for "ClearError()" or make m_error a collection and capture all errors and never reset them.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -129,6 +129,26 @@ nlohmann::json ComputeContainerStatsJson(const wsl::windows::common::docker_sche
 
 namespace wsl::windows::wslc::task {
 
+// Every container is attempted even if an earlier one fails; the command still exits nonzero.
```
dkbennett, line 164: This is another good optimization I like it!

One thought is that this could just be called "ForEachContainer" with a boolean or enum for behavior (continue-on-error or stop-on-error) with default being continue-on-error. 

If we do not have any stop-on-error cases right now, this could just be called "ForEachContainer" and we can add the behavior and stop-on-error semantics later when it exists.

test/windows/wslc/e2e/WSLCE2EContainerStopTests.cpp
```diff
@@ -249,9 +249,36 @@ class WSLCE2EContainerStopTests
         }
     }
 
+    WSLC_TEST_METHOD(WSLCE2E_Container_Stop_ContinuesAfterFailure)
```
dkbennett, line 252: good class of tests to add!

dkbennett (dismissed):

test/windows/WSLCTests.cpp
```diff
@@ -6856,6 +6856,139 @@ class WSLCTests
...
+
+            WaitForOutput(initProcess.GetStdHandle(1), stopSignalMarker);
+
+            // The gap between the two phases is short, so poll for it: until the container has exited, every
+            // delete is turned away by the ordinary running-container guard rather than by the restart.
+            HRESULT deleteResult = WSLC_E_CONTAINER_IS_RUNNING;
+            while (deleteResult == WSLC_E_CONTAINER_IS_RUNNING && restartFuture.wait_for(std::chrono::seconds(0)) != std::future_status::ready)
+            {
+                deleteResult = container.Get().Delete(WSLCDeleteFlagsNone);
+            }
```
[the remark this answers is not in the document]
dkbennett: We have a RetryWithTimeout method that could be used here that avoids this issue and sleeps. Recommend using that to resolve this issue.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1287,12 +1323,23 @@ void WSLCContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCod
...
+
+void WSLCContainerImpl::StopPhase(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill, bool RestartPhase)
 {
     std::shared_ptr<StateTransition> transition;
 
     {
         auto lifecycleLock = m_lifecycleLock.lock_shared();
         auto lock = m_lock.lock_exclusive();
+
+        if (!RestartPhase)
```
AmelBawa-msft: Should we skip the wait for `Kill` as well in this condition?

Since restart attempts to stop a container (which may take long), if we attempt to kill the container to accelerate the stop process, then the 'restart' is "blocked", and 'kill' is also blocked on restart to complete.

beena352: Good catch, Kill should skip the restart wait the same way Delete does. The fix is in #41454 since that's where StopPhase's restart handling lives; it'll be there before this one merges.

src/windows/service/inc/wslc.idl
```diff
@@ -574,6 +574,7 @@ interface IWSLCContainer : IUnknown
     HRESULT DisconnectFromNetwork([in] LPCSTR NetworkName);
     HRESULT UploadArchive([in] WSLCHandle TarHandle, [in, string] LPCSTR DestPath, [in] ULONGLONG ContentSize);
     HRESULT DownloadArchive([in, string] LPCSTR SrcPath, [in] WSLCHandle OutHandle);
+    HRESULT Restart([in] WSLCSignal Signal, [in] LONG TimeoutSeconds, [in, unique] IWarningCallback* WarningCallback);
```
JohnMcPMS, line 585: If there is value to this being implemented in the service (as opposed to the caller just chaining existing calls), it should probably also be exposed to the SDK.  Is that part of a future change?

beena352: Not in this change. The SDK doesn't expose everything the service supports, Kill  is the closest example, it's on IWSLCContainer  but isn't in the SDK either.
It's in the service rather than left to the caller because the stop and start have to be one operation. The  --rm  auto-delete has to be suppressed in between, and anything else touching the container has to wait for the whole restart, not just the phase in flight. Calling Stop() then Start() gets neither.
Happy to add it to the SDK if you think it should be there.

src/windows/wslc/core/CLIExecutionContext.cpp
```diff
@@ -7,6 +7,9 @@ Copyright (c) Microsoft. All rights reserved.
 #include "Argument.h"
 #include "CLIExecutionContext.h"
 
+using namespace wsl::shared;
```
dkbennett, line 10: nit: may want to put these usings under the namespaces to avoid global conflicts. We don't do this consistently in the CLI but we should start doing it for new things to avoid future problems.

dkbennett (dismissed):

src/windows/wslc/core/CLIExecutionContext.cpp
beena352: Sure, I’ll do this in a follow-up.

craigloewen-msft (dismissed):

craigloewen-msft (approved):

Issue #41394, opened 2026-08-20
Background update via MSI/Store (version 2.7.12.0) breaks WSL auto start


SOndrejcik:
### Windows Version

Microsoft Windows [Version 10.0.19045.7663]

### WSL Version

2.7.12.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [x] WSL 1

### Kernel Version

Linux version 4.4.0-19041-Microsoft (Microsoft@Microsoft.com) (gcc version 5.4.0 (GCC) ) #7663-Microsoft Sat Aug 08 10:24:00 PST 2026

### Distro Version

Debian

### Other Software

WBS script for the WSL autostart "start_invisible.vbs" called via "HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Run"

### Repro Steps

1. Have an active WSL session running background processes (managed via VBScript or Startup Registry "HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Run").
2. Windows Update / Windows Store triggers a silent background update of the WSL subsystem (specifically upgrading to version 2.7.12.0 via "wsl.msi" / AppX bundle).
3. The "RestartManager" attempts to restart "wsl.exe" and "wslhost.exe" to apply files but encounters a security/session lock.
4. The system attempts to run a startup VBScript calling "wsl.exe" right after login or during this event.
 
- the user is not notified why his WSL tools are currently offline,
- the user is not asked to reboot the PC or restart the service.

### Expected Behavior

The update installer should safely defer the update until system shutdown, or notify the user via a standard Windows Action Center GUI notification that WSL has been updated and a machine restart is required to use Linux applications again.

### Actual Behavior

The update silently bricks the WSL subsystem. All commands (`wsl -l -v`) report the distros as `Stopped` and subsequent execution attempts fail silently or crash. No user-facing warning is shown, making the application stack look broken to end-users (in our case, automated car service workstations) until a manual machine reboot is performed.

### Diagnostic Logs

Event Viewer captured the following entries during the failure:
- **MsiInstaller:** Windows Installer started transaction: `C:\Program Files\WindowsApps\MicrosoftCorporationII.WindowsSubsystemForLinux_2.7.12.0_x64__8wekyb3d8bbwe\wsl.msi`. Client Process ID: 8616.
- **RestartManager:** Application 'C:\Program Files\WSL\wsl.exe' (pid 9896) cannot be restarted - Application SID does not match Conductor SID.
[...]

Issue #41227, opened 2026-08-01
Increasing net.ipv4.ip_local_port_range beyond the default WSL range causes localhost TCP connections to hang


somewhereMaDan:
### Windows Version

Microsoft Windows [Version 10.0.26200.8875]

### WSL Version

2.6.3.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Repro Steps

Changing net.ipv4.ip_local_port_range from the WSL default causes all localhost TCP connections to hang. This affects multiple applications (Redis, Python HTTP server, curl, nc) and is not specific to a single service.

The issue is reproducible and is resolved only after running wsl --shutdown.

1. Start a fresh WSL session.

2. Verify localhost connectivity.

$ redis-cli ping

PONG

3. Verify the default ephemeral port range.

$ cat /proc/sys/net/ipv4/ip_local_port_range

44620    48715

4. Change the ephemeral port range.

sudo sysctl -w net.ipv4.ip_local_port_range="44620 65535"

(or)

sudo sysctl -w net.ipv4.ip_local_port_range="32768 60999"

5. Test localhost connectivity again.

redis-cli ping

or

python3 -m http.server 8000

curl http://127.0.0.1:8000

The attached screenshot shows:

<img width="1772" height="903" alt="Image" src="https://github.com/user-attachments/assets/1a42165e-536a-468c-a9e8-953cfe25be75" />

redis-cli ping and curl http://127.0.0.1:8000 working with the default ephemeral port range (44620 48715).
Both commands hanging after changing net.ipv4.ip_local_port_range to 10000 65535 or 32768 60999.

### Expected Behavior

Changing the ephemeral port range should not affect localhost TCP connectivity.

### Actual Behavior

redis-cli ping hangs indefinitely.
curl http://127.0.0.1:8000 hangs.
nc hangs.
ss -tanp shows connections remaining in SYN-RECV.
redis-server continues listening on 127.0.0.1:6379.
Running wsl --shutdown restores localhost connectivity.

### Diagnostic Logs

Diagnostic information

While redis-cli is hanging:

$ ss -tanp | grep 6379

LISTEN    127.0.0.1:6379
SYN-RECV  127.0.0.1:6379  127.0.0.1:xxxxx

somewhereMaDan:
[stderr.log](https://github.com/user-attachments/files/30624395/stderr.log)

chemwolf6922:
[...]

PR #41447, opened 2026-08-26
Defer MSI upgrades while WSL is active

hoobnn:

A Store-triggered update can start an MSI upgrade while WSL processes and container sessions are still running. Restart Manager may terminate those processes but fail to restore them, leaving distributions stopped after a successful background update.

This change tracks active WSL processes and WSLC sessions with a named activity object. `WslInstaller` defers the MSI upgrade while that object exists instead of replacing binaries that active sessions are using.

## PR Checklist

- [x] **Closes:** #41394

`WslActivityMarker` maintains a process-local reference count and keeps the `Global\WslActive` event object alive while at least one tracked operation exists. The marker is owned by `LxssCreateProcess` for regular WSL processes and by `SessionEntry` for WSLC sessions.

After determining that an update is needed, `WslInstaller` checks for the activity object. If WSL is active, it records `WSL is active; deferring MSI upgrade until WSL is idle.` in the install log and stops without launching `msiexec`. A later Store update or installer activation can retry after WSL becomes idle.

This is deliberately headless: background updates do not display a Restart Manager prompt or terminate active WSL sessions.

## Validation Steps Performed

- Reproduced while upgrading to WSL 2.9.9.0:
  - MSI installation completed successfully;
  - Restart Manager event 10010 reported application SID mismatches for `wsl.exe`, `wslhost.exe`, and `msrdc.exe`;
  - the running distribution did not resume automatically.
- Built `wslservice`, `wslinstaller`, `wsltests`, `msipackage`, and `msixinstallerpackage` for x64 Release.
- Added `InstallerTests::MsixUpgradeDefersWhileWslIsActive`.
- Ran the built `wslinstaller.exe` as a real LocalSystem service while a WSL `sleep` process was active and the installed-version registry value was set to `1.0.0`:
  - the installer service stopped normally;
  - the version remained `1.0.0`, proving that MSI replacement did not start;
  - the WSL process remained alive;
  - the install log contained the expected deferral message.
- Verified the installed 2.9.9.3 build still passed the #41227 mirrored-network regression and left the existing distributions healthy.

The new TAEF method builds successfully. On the validation host, the repository-wide AppX fixture failed before entering the test method because AppX deployment could not reach `PackagesInUseClosed`; the LocalSystem service validation above exercises the production deferral path directly.


hoobnn:

OneBlue:
This change is essentially a rehash of #40786, and therefore has the same issues. We have a plan to deprecated the MSIX installer logic, and so we shouldn't make changes in that area at this time.

PR #41454, opened 2026-08-26
Add container restart runtime support

beena352:
Adds container restart support to the runtime service layer. Implements the backend Restart() API that stops a container, waits for it to exit, then starts it again as an atomic operation. Prevents external Stop() and Start() calls from landing between the two phases via a mutual-exclusion guard.

## PR Checklist

This PR brings the runtime component of container restart to the service boundary (the COM interface layer). Changes:
- WSLCContainer.cpp/.h: RestartTransaction struct to mark restart phases as atomic; Restart() method that runs stop + start with the transaction held, preventing external operations from landing between phases. Includes guards for delete-during-restart and restart-during-delete races
- wslc.idl: Restart(Signal, TimeoutSeconds, WarningCallback) COM interface method
- wslcsdk.h: New error code WSLC_E_CONTAINER_DELETED for containers deleted while a restart is in flight

The CLI command layer that consumes this API is in #41435 (depends on this PR).


src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1190,11 +1221,16 @@ __requires_exclusive_lock_held(m_lock) std::shared_ptr<WSLCContainerImpl::StateT
...
 {
     auto io = m_wslcSession.CreateIOContext();
-    io.AddHandle(std::make_unique<EventHandle>(transition->Completed.get()));
+    io.AddHandle(std::make_unique<EventHandle>(Event));
     io.Run({});
+}
+
+void WSLCContainerImpl::WaitForTransitionCompletion(const std::shared_ptr<StateTransition>& transition) const
+{
+    WaitForCompletionEvent(transition->Completed.get());
```
kvega005, line 1247: Why was this changed?

The IOContext here was necessary and important. As it allowed us to cancel the wait if the calling process exits so the COM api doesnt hang.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1181,6 +1196,22 @@ void WSLCContainerImpl::WaitForConflictingTransitionToComplete(
...
+        {
+            auto restart = m_restart;
+            lock.reset();
+            lifecycleLock.reset();
+            WaitForCompletionEvent(restart->Completed.get());
+        }
+
+        lifecycleLock = m_lifecycleLock.lock_shared();
+        lock = m_lock.lock_exclusive();
+    }
```
kvega005: **Why a second synchronization primitive instead of `TransitionKind::Restart`?**

This PR adds a restart-only copy of machinery `StateTransition` already provides.

A transition is already allowed to outlive the event it was created for — `OnStopped()` re-arms `transition->ExpectedEvent = ContainerEvent::Destroy` and keeps the *same* transition alive across the Docker delete for `--rm`. A restart is that same shape: create one transition with `Kind = Restart`, `ExpectedEvent = Stop`; when the stop event lands, re-arm `ExpectedEvent` to `Start` instead of completing it.

That looks like it gives, for free:

- **Whole-restart mutual exclusion.** `m_transition` is non-null from the first phase through the last, so every existing `WaitForConflictingTransitionToComplete()` caller stands down — no second wait loop, no second manual-reset event.
- **No `RestartPhase` bool.** "Don't stand down for the restart I'm part of" is exactly what the existing `kind` parameter does, and `StopPhase()` already uses it with `TransitionKind::Stop`.
- `AttachToTransition()` already carries the completion wait, the deferred COM disconnect, and the exception rethrow across both phases.
- The `--rm` suppression in `OnStopped()` becomes `transition->Kind == TransitionKind::Restart` rather than a separate `&& !m_restart`.
- `CommitState()` no longer needs to reach out and `m_restart.reset()` on `Running` — the transition completes itself, so state commit stays a pure state commit.

**And `WSLC_E_CONTAINER_MARKED_FOR_REMOVAL` looks like it falls out of the same thing.** It exists only because `Delete()` deliberately skips `WaitForRestartToComplete()` and can move the container to `Deleted` in the gap between the two phases. If a restart is a single transition, `Delete()` uses the same conflict wait as everything else and that window closes. Is the interleave a hard requirement (docker parity), or an artifact of restart not being a transition? Adding a public HRESULT to the SDK surface is a one-way door, so it's worth being sure it's not just `ERROR_INVALID_STATE` under a new name.

Two things I can see cutting the other way — happy to be told I'm missing something:

- `StartTransition()` asserts `WI_ASSERT(!m_transition)`, so the phases would need to re-arm the in-flight restart transition rather than create a new one. But that's the same re-arm `OnStopped()` already performs.
- `StopPhase()` drops `m_lock` across the blocking Docker stop and re-checks `m_stateGeneration`. If the restart transition is published *before* the Docker request, `OnEvent()` always has a transition to land on, so that generation dance likely gets simpler, not harder.

The concern is that two independent exclusion concepts means every future lifecycle operation has to reason about both (`Start`/`Stop` wait on both, `Delete` waits on one), and the rules for which one applies live only in comments. Is there a case a `TransitionKind::Restart` genuinely can't express?

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1397,6 +1444,45 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill)
...
+        restart->Completed.SetEvent();
+    });
+
+    if (wasRunning)
+    {
+        StopPhase(Signal, TimeoutSeconds, false, true);
+    }
+
+    StartPhase(WSLCContainerStartFlagsNone, nullptr, true);
+}
```
kvega005, line 1576: I suggest instead of seperate `StopPhase`/`StartPhase`, use the `Docker.Restart` API, which already handles the "container is currently stoppoed" case and emits a `ContainerEvent:Restart` that `OnEvent()` can observe and use to complete a transition.

kvega005 (dismissed):
**Restart shouldn't tear down and re-acquire runtime resources**

Right now `Restart()` = `StopPhase()` + `StartPhase()`, and the stop half goes through `OnStopped()` → `ReleaseRuntimeResources()` → `UnmapPorts()` + `UnmountVolumes()`. The start half then re-runs `MountVolumes()` + `MapPorts()`. So a restart drops every host/VM-side reservation the container already legitimately owned and races to re-take it.

Concrete failure modes:

- `UnmapPorts()` resets `VmMapping.VmPort` for `host` mode, returning the VM port to the pool. `MapPorts()` then re-runs `TryAllocatePort()` and throws `WSAEADDRINUSE` / `MessageWslcPortInUse` if anything grabbed it in the gap. `wslc restart` failing with *"port already in use"* naming the container's **own** port is not a sensible outcome.
- Same for the host side: `VmMapping.Unmap()` releases the reservation, so any other process on the box can take the published port during the window, and `MapPort()` comes back `ERROR_ALREADY_EXISTS`/`WSAEADDRINUSE`.
- `UnmountVolumes()` + `MountVolumes()` round-trips every bind mount through the VM. If the host path was removed while the container was running, restart now fails on a mount error — or worse, silently **re-creates** the directory, because `MountVolumes()` honours `CreateSourceIfMissing` on this path too.

And because none of this is atomic, a restart that fails in the start half leaves the container `Exited` with its resources gone. The caller asked for a restart and got a stop.

The container's ports and mounts are host/VM-scoped, not run-scoped — they should simply survive the restart. `OnStopped()` should skip `ReleaseRuntimeResources()` when the stop is part of a restart, and the start half should skip `MapPorts()`/`MountVolumes()` correspondingly.

**Use Docker's `/containers/{id}/restart` instead of hand-rolling the two phases**

`POST /containers/{id}/restart?signal=&t=` takes exactly the two parameters `Restart(Signal, TimeoutSeconds)` already has, and it handles the "container is currently stopped" case itself — which removes the `wasRunning` branch, the conditional `StopPhase()`, and a good chunk of the reason `m_restart` has to exist at all.

Worth noting the plumbing for this is already here and currently dead: `DockerEventTracker.h:28` declares `ContainerEvent::Restart` and `DockerEventTracker.cpp:164` already maps Docker's `"restart"` action onto it — but `OnEvent()` only handles `Start`/`Stop`/`Destroy`, so the event is delivered and dropped. `DockerHTTPClient` just needs a `RestartContainer()` alongside `StopContainer()`/`StartContainer()`.

Docker still emits `die` → `start` → `restart` for this, so `OnStopped()` still has to know a restart is in flight and not release resources — that part is needed either way. But the payoff is that "restart" becomes one Docker call plus one expected terminal event, which is exactly the shape `StateTransition` is built for.

**Smaller things**

- **Stop timeout isn't validated when the container is stopped.** `ValidateStopTimeout()` only runs inside `StopPhase()`'s `m_state == Running` branch, and `Restart()` skips `StopPhase()` entirely when `!wasRunning`. So `Restart(SIGTERM, -5)` returns `E_INVALIDARG` on a running container and `S_OK` on a stopped one. Validate at the top of `Restart()`.
- **Plugin rejection turns a restart into a stop.** `StartPhase()` calls `OnContainerStarted()` and, on failure, stops the container and throws. Combined with `OnStopped()` having already fired `OnContainerStopping()`, a plugin-guarded container that fails the restart is left down. Is that intended, or should a rejected restart roll back?
- **The restart transaction ends before `Restart()` returns.** `CommitState()` clears `m_restart` as soon as the start phase commits `Running`, i.e. while `Restart()` is still inside `AttachToTransition()`. The `if (m_restart == restart)` guard in `restartCleanup` exists because of that window. It looks benign today, but it means the header comment "keeps the pair indivisible" isn't quite what the code does — another reason to let a single transition own the whole operation and complete itself.

src/windows/wslcsession/WSLCContainer.cpp
beena352: The IOContext is still there, it just moved up into WaitForCompletionEvent() (:1224), which still does CreateIOContext(), AddHandle(EventHandle), Run({}). WaitForTransitionCompletion() calls that now, so nothing changed about cancelling the wait when the caller exits. The upside is the restart wait uses the same helper, so it gets that behaviour too instead of me writing a second wait.

src/windows/wslcsession/WSLCContainer.cpp
beena352: Fair, and honestly this is what I said I'd do originally, see r3856416334 on #41435, where I told Pierre, I'd add a restart kind next to Start/Stop/Delete. I switched to a separate m_restart because StartTransition() asserts !m_transition (:1219) and each phase makes its own transition. You've already answered that: re-arm the transition instead of creating a new one, same as OnStopped() does for Destroy. I'll take it.

There's one thing we need to sort out first though. Delete() not waiting isn't an accident, Pierre asked for Docker parity (r3857377142) and I tested it against moby. A forced remove during a restart wins, and the restart then fails with "container is marked for removal and cannot be started."

The problem is those two don't compose. Delete() calls WaitForConflictingTransitionToComplete() with no kind (:1606), so it waits on any transition. The moment restart becomes a transition, Delete() starts standing down and the parity is gone. Keeping it means Delete() explicitly not standing down for Kind == Restart, which is narrower than m_restart, but it's still a special case, so it doesn't fully buy back the simplicity you're after.

So: keep the parity and accept one special case in Delete(), or drop the parity and get the uniform rule? I'm fine either way, but I'd rather we pick deliberately than have it change as a side effect. Same question decides whether WSLC_E_CONTAINER_MARKED_FOR_REMOVAL needs to exist at all, and if it does, I'm happy to use ERROR_INVALID_STATE with the same user message instead of a new public HRESULT.

src/windows/wslcsession/WSLCContainer.cpp
beena352: Pierre, raised this one first on #41435 (r3856328071) and we closed it there, but it's worth writing down properly.

/restart handles the sequencing fine, including the already-stopped case. The catch is it leaves us no room to do our own start work. The Docker start call is one line of StartPhase() (:1129) - everything around it gets skipped if the daemon does the restart: the named-volume health check that produces WSLC_E_VOLUME_NOT_AVAILABLE (:1109), OnContainerStarted() so plugins can see the start and reject it, and rebuilding m_initProcess/m_initProcessControl (:1095), which otherwise stay pointed at the dead process's stream. Plus detach keys and TTY sizing. Worth noting moby composes it too, daemon/restart.go::containerRestart is a stop followed by a start.

You're right that ContainerEvent::Restart is dead, declared at DockerEventTracker.h:29, mapped at .cpp:164, never handled. It's dead because a composed stop+start never emits restart; moby only logs that from containerRestart. So it can only show up if someone runs docker restart behind our back. That's worth handling, but as a "something changed outside of us" signal, not as our completion event.

beena352:
> **Restart shouldn't tear down and re-acquire runtime resources**
> 
> Right now `Restart()` = `StopPhase()` + `StartPhase()`, and the stop half goes through `OnStopped()` → `ReleaseRuntimeResources()` → `UnmapPorts()` + `UnmountVolumes()`. The start half then re-runs `MountVolumes()` + `MapPorts()`. So a restart drops every host/VM-side reservation the container already legitimately owned and races to re-take it.
> 
> Concrete failure modes:
> 
> * `UnmapPorts()` resets `VmMapping.VmPort` for `host` mode, returning the VM port to the pool. `MapPorts()` then re-runs `TryAllocatePort()` and throws `WSAEADDRINUSE` / `MessageWslcPortInUse` if anything grabbed it in the gap. `wslc restart` failing with _"port already in use"_ naming the container's **own** port is not a sensible outcome.
> * Same for the host side: `VmMapping.Unmap()` releases the reservation, so any other process on the box can take the published port during the window, and `MapPort()` comes back `ERROR_ALREADY_EXISTS`/`WSAEADDRINUSE`.
> * `UnmountVolumes()` + `MountVolumes()` round-trips every bind mount through the VM. If the host path was removed while the container was running, restart now fails on a mount error — or worse, silently **re-creates** the directory, because `MountVolumes()` honours `CreateSourceIfMissing` on this path too.
> 
> And because none of this is atomic, a restart that fails in the start half leaves the container `Exited` with its resources gone. The caller asked for a restart and got a stop.
> 
> The container's ports and mounts are host/VM-scoped, not run-scoped — they should simply survive the restart. `OnStopped()` should skip `ReleaseRuntimeResources()` when the stop is part of a restart, and the start half should skip `MapPorts()`/`MountVolumes()` correspondingly.
> 
> **Use Docker's `/containers/{id}/restart` instead of hand-rolling the two phases**
> 
> `POST /containers/{id}/restart?signal=&t=` takes exactly the two parameters `Restart(Signal, TimeoutSeconds)` already has, and it handles the "container is currently stopped" case itself — which removes the `wasRunning` branch, the conditional `StopPhase()`, and a good chunk of the reason `m_restart` has to exist at all.
> 
> Worth noting the plumbing for this is already here and currently dead: `DockerEventTracker.h:28` declares `ContainerEvent::Restart` and `DockerEventTracker.cpp:164` already maps Docker's `"restart"` action onto it — but `OnEvent()` only handles `Start`/`Stop`/`Destroy`, so the event is delivered and dropped. `DockerHTTPClient` just needs a `RestartContainer()` alongside `StopContainer()`/`StartContainer()`.
> 
> Docker still emits `die` → `start` → `restart` for this, so `OnStopped()` still has to know a restart is in flight and not release resources — that part is needed either way. But the payoff is that "restart" becomes one Docker call plus one expected terminal event, which is exactly the shape `StateTransition` is built for.
> 
> **Smaller things**
> 
> * **Stop timeout isn't validated when the container is stopped.** `ValidateStopTimeout()` only runs inside `StopPhase()`'s `m_state == Running` branch, and `Restart()` skips `StopPhase()` entirely when `!wasRunning`. So `Restart(SIGTERM, -5)` returns `E_INVALIDARG` on a running container and `S_OK` on a stopped one. Validate at the top of `Restart()`.
> * **Plugin rejection turns a restart into a stop.** `StartPhase()` calls `OnContainerStarted()` and, on failure, stops the container and throws. Combined with `OnStopped()` having already fired `OnContainerStopping()`, a plugin-guarded container that fails the restart is left down. Is that intended, or should a rejected restart roll back?
> * **The restart transaction ends before `Restart()` returns.** `CommitState()` clears `m_restart` as soon as the start phase commits `Running`, i.e. while `Restart()` is still inside `AttachToTransition()`. The `if (m_restart == restart)` guard in `restartCleanup` exists because of that window. It looks benign today, but it means the header comment "keeps the pair indivisible" isn't quite what the code does — another reason to let a single transition own the whole operation and complete itself.

You're right about the resources, that's a real bug. OnStopped() always calls ReleaseRuntimeResources() (:1512), and then the start half turns around and re-runs MountVolumes() (:1122) and MapPorts() (:1125). So a restart gives up reservations the container already owns and then races to get them back. Both failure modes are real, in host mode UnmapPorts() resets VmPort, so MapPorts() goes back through TryAllocatePort() and can throw WSAEADDRINUSE naming the container's own port (:2887), and on the host side Unmap()/MapPort() can come back ERROR_ALREADY_EXISTS in any mode. MountVolumes() hits CreateDirectoryDeepNoThrow when CreateSourceIfMissing is set (:443), so a bind source someone deleted while the container was running quietly comes back.

Worth noting moby avoids this deliberately, containerRestart wraps the stop+start in daemon.Mount(container) with a defer daemon.Unmount(container), commented "Avoid unnecessarily unmounting and then directly mounting the container when the container stops and then starts again." Same idea, they just hold it at a different layer.

The fix is to skip ReleaseRuntimeResources() when a restart is in flight and skip the matching re-acquire on the way back up. Just that one call though, ReleaseProcesses() and CommitState(Exited) still need to run.

The timeout is a real bug too. ValidateStopTimeout() lives inside the Running branch (:1376), and restart skips the stop phase entirely when the container isn't running. So a bad timeout gets rejected on a running container and accepted on a stopped one. I'll move the check to the top of Restart().

On the plugin rejection leaving the container down, Start() already behaves that way today, both go through the same StartPhase() (:1025 and :1483), so restart isn't introducing it. Happy to fix it, but I'd rather do that separately unless you feel strongly.

One thing worth pinning down: as you noted, the teardown skip is needed either way. Under /restart docker still emits die then start, so OnEvent() still sees the die and OnStopped() still releases, and there'd be no StartPhase() to re-acquire, so the container would come back up with its ports unmapped. So the resource fix is independent of the /restart question and I'll land it regardless of which way that one goes.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1440,7 +1526,7 @@ __requires_exclusive_lock_held(m_lock) void WSLCContainerImpl::OnStopped(int exi
     }
 
     // Stop with Rm must initiate Delete.
-    if (WI_IsFlagSet(m_containerFlags, WSLCContainerFlagsRm))
+    if (WI_IsFlagSet(m_containerFlags, WSLCContainerFlagsRm) && !m_restart)
```
ptrivedi, line 1626: hmm, so this will suppress automatic deletion when m_restart is set, which seems good for when a restart succeeds. But if startup phase subsequently fails, it will never perform the deferred deletion

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1397,6 +1444,45 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill)
...
+{
+    bool wasRunning{};
+    auto restart = std::make_shared<RestartTransaction>();
+
+    {
+        auto lifecycleLock = m_lifecycleLock.lock_shared();
+        auto lock = m_lock.lock_exclusive();
+        WaitForRestartToComplete(lock, lifecycleLock);
+
+        wasRunning = m_state == WslcContainerStateRunning;
```
ptrivedi, line 1504: Before reading m_state here and setting m_restart, we are not waiting for conflicting transitions. I think we need to re-evaluate both conditions whenever locks are re-acquired. Integrating the restart guard into the transition wait loop should take care of it

test/windows/WSLCTests.cpp
```diff
@@ -6856,6 +6856,140 @@ class WSLCTests
...
+            }
+            else
+            {
+                // The start phase closed the gap first, so the container was running again by the last attempt.
+                VERIFY_ARE_EQUAL(deleteResult, WSLC_E_CONTAINER_IS_RUNNING);
+                VERIFY_SUCCEEDED(restartHr);
+            }
+        }
+    }
+
```
ptrivedi, line 7113: should add test coverage for two overlapping Restart calls against the same container and verify things like:
- second restart waits for the first rather than racing it
- both calls are able to finish successfully
- container ends up in running state

this way WaitforRestartToComplete() can be exercised

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1397,6 +1444,45 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill)
...
+                m_restart.reset();
+            }
+        }
+
+        restart->Completed.SetEvent();
+    });
+
+    if (wasRunning)
+    {
+        StopPhase(Signal, TimeoutSeconds, false, true);
```
ptrivedi, line 1544: nit: annotate the booleans like 
```suggestion
        StopPhase(Signal, TimeoutSeconds, /* Kill */ false, /* RestartPhase */ true);
```

src/windows/service/inc/wslc.idl
```diff
@@ -877,3 +878,4 @@ cpp_quote("#define WSLC_E_SESSION_NOT_FOUND MAKE_HRESULT(SEVERITY_ERROR, FACILIT
 // N.B. WSLC_E_VM_NOT_RUNNING is part of the plugin API contract and is also defined in WslPluginApi.h.
 // The two definitions must stay in sync.
 cpp_quote("#define WSLC_E_VM_NOT_RUNNING MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 16) /* 0x80040610 */")
+cpp_quote("#define WSLC_E_CONTAINER_MARKED_FOR_REMOVAL MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 17) /* 0x80040611 */")
```
ptrivedi: what about docs? does error-codes.md need to be updated manually with this new code?

ptrivedi: I think it might need to get added in common/wslutil.cpp as well?

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1021,11 +1021,21 @@ void WSLCContainerImpl::Attach(LPCSTR DetachKeys, WSLCHandle* Stdin, WSLCHandle*
...
     std::shared_ptr<StateTransition> transition;
     auto lifecycleLock = m_lifecycleLock.lock_shared();
     auto lock = m_lock.lock_exclusive();
-    WaitForConflictingTransitionToComplete(lock, lifecycleLock);
+
+    WaitForConflictingTransitionToComplete(lock, lifecycleLock, std::nullopt, /* waitForRestart */ !RestartPhase);
+
+    // A Delete() that raced a restart may have already moved the container to the Deleted state.
+    THROW_HR_WITH_USER_ERROR_IF(
+        WSLC_E_CONTAINER_MARKED_FOR_REMOVAL, Localization::MessageWslcContainerMarkedForRemoval(m_id), m_state == WslcContainerStateDeleted);
```
ptrivedi: nit: we call this "marked for removal", but the state is called 'deleted'. The WSLC_E_CONTAINER_MARKED_FOR_REMOVAL would be in the public sdk header and get localized etc. would it be better to name it 'WSLC_E_CONTAINER_DELETED' and use corrsponding wording like "Container '{}' has been dleeted and cannot be started' or something like that?

test/windows/WSLCTests.cpp
```diff
@@ -6856,6 +6856,161 @@ class WSLCTests
...
+            WaitForOutput(initProcess.GetStdHandle(1), stopSignalMarker);
+
+            VERIFY_SUCCEEDED(container.Get().Stop(WSLCSignalSIGKILL, 0));
+            VERIFY_SUCCEEDED(restartResult.get_future().get());
+            VERIFY_ARE_EQUAL(container.State(), WslcContainerStateExited);
+        }
+
+        // A delete issued during a restart deliberately does not wait for it, matching docker: whichever of
+        // the delete and the restart's start phase lands first wins, and the other one fails.
```
ptrivedi, line 7029: so, if delete wins, then restart returns WSLC_E_CONTAINER_MARKED_FOR_REMOVAL
if restart reaches the running first, then delete returns WSLC_E_CONTAINER_IS_RUNNING and then restart ends up succeeding?

ptrivedi (dismissed):

test/windows/WSLCTests.cpp
beena352: Yes, that's right. Delete passes waitForRestart: false so it can always cut through the gap between the two restart phases. If delete wins that gap, the container is already marked for removal and restart's start phase fails with WSLC_E_CONTAINER_MARKED_FOR_REMOVAL. If restart wins and commits Running first, then delete without --force gets back WSLC_E_CONTAINER_IS_RUNNING (with --force it succeeds and kills the freshly-started container, which matches docker).

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1397,6 +1439,89 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill)
...
+                ReleaseRuntimeResources();
+            }
+
+            if (WI_IsFlagClear(m_containerFlags, WSLCContainerFlagsRm) || m_state != WslcContainerStateExited)
+            {
+                return;
+            }
+        }
+
+        Delete(WSLCDeleteFlagsForce | WSLCDeleteFlagsDeleteVolumes);
```
kvega005: `OnFailedRestart()` drops `m_lock` between the state check and the `Delete(WSLCDeleteFlagsForce | WSLCDeleteFlagsDeleteVolumes)` call. A `Start()` that was standing down on the restart transaction is released by `restartCleanup` immediately before this and is likely already queued on `m_lock` while `ReleaseRuntimeResources()` does its unmap/unmount RPC, so it can win the lock the moment we release it and bring the container back up. Because `Force` is set, `RequestDeleteExclusiveLockHeld()`'s running-container guard is bypassed, and we force-delete a now-running container.

test/windows/WSLCTests.cpp
```diff
@@ -6856,6 +6856,161 @@ class WSLCTests
         }
     }
 
+    WSLC_TEST_METHOD(ContainerRestart)
+    {
```
kvega005, line 6866: You should probably add a test that covers ports and volumes survive a restart.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1287,13 +1321,21 @@ void WSLCContainerImpl::OnEvent(ContainerEvent event, std::optional<int> exitCod
...
     std::shared_ptr<StateTransition> transition;
 
     {
         auto lifecycleLock = m_lifecycleLock.lock_shared();
         auto lock = m_lock.lock_exclusive();
-        WaitForConflictingTransitionToComplete(lock, lifecycleLock, TransitionKind::Stop);
+
+        // Kill is the escape hatch when a restart's stop phase is stuck, so it must not wait on the very
+        // restart it is meant to unblock.
+        WaitForConflictingTransitionToComplete(lock, lifecycleLock, TransitionKind::Stop, /* waitForRestart */ !RestartPhase && !Kill);
```
kvega005: `Kill()` during a restart is inconsistent with `Stop()` and untested. If it lands during the stop phase it merges into the same `Stop` transition, `OnStopped()` suppresses release/auto-remove because `m_restart` is set, and the start phase brings the container back — so `Kill()` returns `S_OK` with the container `Running` and an `--rm` container not removed, the opposite of `test-restart-race-stop`'s outcome. If it lands between the phases instead, the container is `Exited` with no transition, so it falls through to `WSLC_E_CONTAINER_NOT_RUNNING` — the "escape hatch" only works while the container is still up.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1397,6 +1439,89 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill)
...
+
+        StartPhase(WSLCContainerStartFlagsNone, nullptr, /* RestartPhase */ true);
+    }
+
+    failureCleanup.release();
+}
+
+// N.B. Runs after the restart transaction has been released, so the delete below is no longer
+// suppressed by OnStopped().
+void WSLCContainerImpl::OnFailedRestart() noexcept
```
kvega005: Is the `m_runtimeResourcesHeld` release in `OnFailedRestart()` reachable? `StartPhase`'s own `resourceCleanup` already releases on every throw from line 1119 onward, and of the earlier throws only `WSLC_E_VOLUME_NOT_AVAILABLE` is possible during a restart — the attach and tty paths are skipped because `Restart()` passes `WSLCContainerStartFlagsNone` and a null `StartOptions`, and the `Deleted` case early-returns on `m_state != Exited`. That leaves one narrow trigger plus a `StopPhase` Docker failure, neither of which any test can reach, so both of this function's branches are unexercised.

src/windows/wslcsession/WSLCContainer.cpp
beena352: Good catch, a Start() released by restartCleanup is already queued on m_lock while ReleaseRuntimeResources() does its unmap/unmount, so it can win the lock the moment we drop it and bring the container back up before Delete() re-acquires it. I'll mirror what OnStopped() does and issue the DeleteContainer call inline under the same lock scope so the state check and the removal are atomic.

src/windows/wslcsession/WSLCContainer.cpp
kvega005: Most of that list doesn't hold up.

Detach keys, the IO stream and TTY sizing aren't start-path work, they're attach-path work, and they already have their own APIs. `Attach()` does the same `AttachContainer` plus relay setup and only needs the container Running. `IWSLCProcess::ResizeTty` is public and the CLI already calls it independently on every window resize. They're bundled into `StartPhase()` so `run -it` can do it in one shot without racing the start, and restart isn't that. `Restart()` passes `WSLCContainerStartFlagsNone` and a null `StartOptions` anyway, so none of those three run today. Docker's `restart` isn't an attaching verb either, you re-attach manually after.

 `m_initProcess` is a pointer swap under `m_processesLock` with `io` null. It just has to land between the `die` and the `start`, and that branch has to exist regardless for the `ReleaseRuntimeResources()` skip.

 The named volume check is a precondition for mounting. If we're skipping the release there's nothing to re-verify, it's there for the cold path where the VM was torn down and the VHD may not have come back.
And the resource point cuts the other way. Skip the release while a restart is in flight and there's nothing to re-acquire, so `/restart` has fewer moving parts there, not more.

 What's left is `OnContainerStarted()`. `StartPhase()` holds `m_lifecycleLock` shared across `StartContainer()` and the plugin call, so a rejection stops the container and throws before the transition exists, the start event lands unexpected, and `CommitState(Running)` never runs. A rejected container is never observably Running. Under `/restart` both events land while we're blocked, so it commits to Running and takes the activity hold first and rejection becomes a rollback. That's a real difference and I'm fine with it being the reason, but it should be the reason rather than the list.

src/windows/wslcsession/WSLCContainer.cpp
kvega005: I agree with this!

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1514,7 +1644,10 @@ void WSLCContainerImpl::Delete(WSLCDeleteFlags Flags)
     std::shared_ptr<StateTransition> transition;
     auto lifecycleLock = m_lifecycleLock.lock_shared();
     auto lock = m_lock.lock_exclusive();
-    WaitForConflictingTransitionToComplete(lock, lifecycleLock);
+
+    // N.B. Unlike Start() and Stop(), this deliberately does not wait for an in-flight restart.
+    // A remove that lands between the two phases takes effect, and the restart's start phase fails.
+    WaitForConflictingTransitionToComplete(lock, lifecycleLock, std::nullopt, /* waitForRestart */ false);
```
kvega005: This only holds for a non-force delete. With `WSLCDeleteFlagsForce` the running-container guard in `RequestDeleteExclusiveLockHeld()` is bypassed, so the remove can land *during* the stop phase — including the entire Docker `StopContainer` window, where no transition is published. The start phase is then never reached, and `Restart()` fails with `WSLC_E_CONTAINER_NOT_RUNNING` from `StopPhase` (or a Docker 404) rather than the `WSLC_E_CONTAINER_MARKED_FOR_REMOVAL` added for this case. `test-restart-race-delete` only uses `WSLCDeleteFlagsNone`, so nothing covers it.

test/windows/WSLCTests.cpp
```diff
@@ -6856,6 +6856,161 @@ class WSLCTests
...
+            WSLCContainerLauncher launcher("debian:latest", "test-restart-race-stop", ignoreStopSignal);
+            auto container = launcher.Launch(*m_defaultSession);
+            auto initProcess = container.GetInitProcess();
+
+            std::promise<HRESULT> restartResult;
+            std::thread restartThread(
+                [&]() { restartResult.set_value(container.Get().Restart(WSLCSignalSIGTERM, stopTimeoutSeconds, nullptr)); });
+
+            auto joinThread = wil::scope_exit([&]() { restartThread.join(); });
+
```
kvega005: nit: `thread.join()` can throw. std::thread::join() can throw std::system_error, and a plain scope_exit will std::terminate if that happens while unwinding from a failed `VERIFY_*`. You should use `wil:scope_exit_log` instead.

test/windows/WSLCTests.cpp
```diff
@@ -6856,6 +6856,161 @@ class WSLCTests
...
+            WSLCContainerLauncher launcher("debian:latest", "test-restart-race-delete", ignoreStopSignal);
+            auto container = launcher.Launch(*m_defaultSession);
+            auto initProcess = container.GetInitProcess();
+
+            std::promise<HRESULT> restartResult;
+            std::thread restartThread(
+                [&]() { restartResult.set_value(container.Get().Restart(WSLCSignalSIGTERM, stopTimeoutSeconds, nullptr)); });
+
+            auto joinThread = wil::scope_exit([&]() { restartThread.join(); });
+            auto restartFuture = restartResult.get_future();
```
kvega005: Ditto about `wil:scope_exit_log`

test/windows/WSLCTests.cpp
```diff
@@ -6856,6 +6856,161 @@ class WSLCTests
...
+            WSLCContainerLauncher launcher("debian:latest", "test-restart-race-restart", ignoreStopSignal);
+            auto container = launcher.Launch(*m_defaultSession);
+            auto initProcess = container.GetInitProcess();
+
+            std::promise<HRESULT> restartResult;
+            std::thread restartThread(
+                [&]() { restartResult.set_value(container.Get().Restart(WSLCSignalSIGTERM, stopTimeoutSeconds, nullptr)); });
+
+            auto joinThread = wil::scope_exit([&]() { restartThread.join(); });
+
```
kvega005: Ditto about `wil:scope_exit_log`

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1397,6 +1439,89 @@ void WSLCContainerImpl::Stop(WSLCSignal Signal, LONG TimeoutSeconds, bool Kill)
...
+        auto lifecycleLock = m_lifecycleLock.lock_shared();
+        auto lock = m_lock.lock_exclusive();
+        WaitForConflictingTransitionToComplete(lock, lifecycleLock);
+
+        wasRunning = m_state == WslcContainerStateRunning;
+
+        // N.B. Stop() and Start() each take m_lock, so it cannot be held across both phases. m_restart
+        // stands them down until the start phase commits Running instead.
+        m_restart = restart;
+    }
```
kvega005, line 1509: Nit: `m_restart` is published here but its releaser isn't armed until `restartCleanup` two statements down. Anything throwing in that gap leaves `m_restart` set permanently, and every later `Start()`, `Stop()`, and `Restart()` on the container blocks forever in `WaitForConflictingTransitionToComplete()` on a `Completed` event nobody will ever signal. Nothing in the gap can actually throw today (lock destructors are noexcept, `scope_exit_log` construction doesn't allocate), so this is more about robustness than a bug. But its worth addressing or at least a comment describing the requirement that nothing in that gap can throw

src/windows/wslcsession/WSLCContainer.h
```diff
@@ -231,6 +251,17 @@ class WSLCContainerImpl : public std::enable_shared_from_this<WSLCContainerImpl>
...
+    // Stop() stand down for that window, and OnStopped() keeps the container's runtime resources mapped
+    // and skips the auto-delete of an --rm container. Delete() does not stand down: a remove that lands
+    // between the two phases takes effect, and the restart's start phase fails.
+    _Guarded_by_(m_lock) std::shared_ptr<RestartTransaction> m_restart;
+
+    // True between a successful StartPhase() and the release of the container's ports and mounts. A
+    // restart leaves this set across the two phases, which is what tells the start phase they are still
+    // held and must not be re-acquired.
+    _Guarded_by_(m_lock) bool m_runtimeResourcesHeld {};
+
```
kvega005: nit: unnecessary space after `m_runtimeResourcesHeld`

doc/docs/api-reference/c/error-codes.md
```diff
@@ -18,6 +18,7 @@
 #define WSLC_E_VOLUME_NOT_AVAILABLE MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 14)       /* 0x8004060E */
 #define WSLC_E_SESSION_NOT_FOUND MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 15)          /* 0x8004060F */
 #define WSLC_E_VM_NOT_RUNNING MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 16)             /* 0x80040610 */
+#define WSLC_E_CONTAINER_MARKED_FOR_REMOVAL MAKE_HRESULT(SEVERITY_ERROR, FACILITY_ITF, WSLC_E_BASE + 17) /* 0x80040611 */
```
kvega005: This block mirrors `wslcsdk.h` verbatim, but the header realigned all 16 existing comment columns to fit the longer symbol and the doc didn't, so the new line now juts out past every other one and the two copies no longer match.

test/windows/WSLCTests.cpp
```diff
@@ -6856,6 +6856,161 @@ class WSLCTests
...
+            auto joinThread = wil::scope_exit([&]() { restartThread.join(); });
+            auto restartFuture = restartResult.get_future();
+
+            WaitForOutput(initProcess.GetStdHandle(1), stopSignalMarker);
+
+            // The gap between the two phases is short, so poll for it: until the container has exited, every
+            // delete is turned away by the ordinary running-container guard rather than by the restart.
+            HRESULT deleteResult = WSLC_E_CONTAINER_IS_RUNNING;
+            while (deleteResult == WSLC_E_CONTAINER_IS_RUNNING &&
```
kvega005: `deleteResult` is seeded with `WSLC_E_CONTAINER_IS_RUNNING`, the same value the else branch asserts at `WSLCTests.cpp:6987`. If `restartFuture` is already ready on the first `wait_for`, the loop exits without calling `Delete()` once and the test still passes, having exercised no race at all. Nothing distinguishes "we tried and were correctly refused" from "we never tried."

src/windows/wslcsession/WSLCContainer.h
beena352: Fixed. clang-format puts the space back for  {} , so it's  = false  now.

src/windows/wslcsession/WSLCContainer.cpp
beena352: Fixed, StopPhase  now checks for Deleted  up front and again in the docker error handler, so a force deletes during the stop phase reports MARKED_FOR_REMOVAL  instead of NOT_RUNNING

src/windows/wslcsession/WSLCContainer.cpp
beena352: Added a comment. Can't reorder it,  restartCleanup  has to come after  failureCleanup  so it runs first

src/windows/wslcsession/WSLCContainer.cpp
beena352: Coming back to this. The blocker was the Delete() parity question above, Delete() calls WaitForConflictingTransitionToComplete() with no kind, so the moment restart becomes a transition it starts standing down and the Docker parity Pierre asked for is gone. Keeping it means Delete() explicitly not standing down for Kind == Restart, which is a special case either way.
I've kept m_restart for this push and addressed everything else you flagged.
Happy to do the TransitionKind::Restart refactor as a follow-up if you still want it, it's a structural change across both phases plus every race test, so I'd rather not land it late in this PR

src/windows/wslcsession/WSLCContainer.cpp
beena352: yes. Kill() skips the restart wait on purpose so it can cut through a stuck restart instead of blocking on it, that's the escape hatch.
The NOT_RUNNING  in the gap between the stop and start phases is real though, and it's the same root cause as your  TransitionKind::Restart  question.  m_restart  and  m_transition  being separate means there's a window where the transition is null but the restart is still in flight, so  Kill()  sees "not running". Unifying them would close that window as a side effect. Keeping  m_restart  for this PR for the reasons in the other thread, and this gap is a known consequence of that, it'd go away with the follow-up.

src/windows/wslcsession/WSLCContainer.cpp
beena352: Reachable, the volume check throws before resourceCleanup is armed, so StartPhase doesn't release anything on that path. m_runtimeResourcesHeld is already true going into a restart, so OnFailedRestart() is the only thing that releases it. Same if StopPhase itself fails, StartPhase never runs at all.
Coverage-wise: I added a restart to the VHD-recovery test, which exercises OnFailedRestart() on a failed restart, but not the release itself, that needs the volume attached to a running container, and the .vhdx is open at that point so the test can't delete it.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1020,11 +1020,20 @@ void WSLCContainerImpl::Attach(LPCSTR DetachKeys, WSLCHandle* Stdin, WSLCHandle*
 }
 
 void WSLCContainerImpl::Start(WSLCContainerStartFlags Flags, const WSLCProcessStartOptions* StartOptions)
+{
+    StartPhase(Flags, StartOptions, /* RestartPhase */ false);
```
benhillis: We don't do this style of comments on function arguments.

src/windows/wslcsession/WSLCContainer.cpp
```diff
@@ -1513,7 +1658,10 @@ void WSLCContainerImpl::Delete(WSLCDeleteFlags Flags)
     std::shared_ptr<StateTransition> transition;
     auto lifecycleLock = m_lifecycleLock.lock_shared();
     auto lock = m_lock.lock_exclusive();
-    WaitForConflictingTransitionToComplete(lock, lifecycleLock);
+
+    // N.B. Unlike Start() and Stop(), this deliberately does not wait for an in-flight restart.
+    // A remove that lands between the two phases takes effect, and the restart's start phase fails.
+    WaitForConflictingTransitionToComplete(lock, lifecycleLock, std::nullopt, false);
```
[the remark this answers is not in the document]
beena352: waitForRestart=false  only skips the  m_restart  wait; the  else if  at line 1194 still runs, and with  kind == nullopt  it reduces to a plain  m_transition  check, same behavior as master, where  Delete()  called this with the default  kind . Not a regression.
The assert concern doesn't apply: with  kind == nullopt  the loop only returns once  m_transition == nullptr , and  m_lock  is held continuously from there through  StartTransition()  at line 1667, so  WI_ASSERT(!m_transition)  can't fire.
During a restart's stop phase, both locks are released before the Docker call (lines 1380-1381) and no transition exists until after it returns (line 1429), so a force delete in that window sees  m_transition == nullptr  and proceeds immediately,as intended, and as the race test verifies.

ptrivedi (approved):

PR #41455, opened 2026-08-26
wslc: match docker prune semantics (confirmation prompt, -f aliases --force)

ggarzia-MSFT:

Aligns `wslc <object> prune` with docker: `-f` now means `--force`, and prune without `--force` shows docker's warning plus an `Are you sure you want to continue? [y/N]` confirmation. Also plumbs `container prune --filter` through to the service, which was previously parsed but discarded.

## PR Checklist

In docker, `--filter` only carries a `-f` short alias on listing commands (`docker images`, `docker ps`); on every `prune` command `-f` is bound to `--force` instead. `wslc` bound `-f` to `--filter` everywhere, so `wslc container prune -f` did the wrong thing.

- Added a `PruneFilter` argument type: identical parsing to `Filter`, but with no short alias. The prune commands now register `PruneFilter` + `Force`. Listing commands keep `Filter` with `-f`, unchanged.
  - A separate type is required because `Argument::Create` can override description/required/limit but not the alias, and `findArgumentByAlias` is first-match-wins with no collision detection.
- `Terminal::Confirm` mirrors docker's `prompt.Confirm`: renders `<message> [y/N] `, trims the answer, and accepts only a case-insensitive `y`. End of input declines rather than hanging.
- `CLIExecutionContext::ConfirmPrune` is the single choke point: returns immediately when `--force` is set, otherwise prints the per-command warning and the shared confirmation question.
- Warning text matches docker per command, including the `--all` variants for `image prune` and `volume prune`, and volume prune's filter help that documents `label=` rather than `until=`.
- `ContainerService::Prune` now marshals the parsed filters into `WSLCFilter[]` (mirroring `NetworkService::Prune`) instead of passing `nullptr, 0`.
- Prune filter parse failures report `WSLCCLI_InvalidPruneFilterError` so the message can name the correct flag spelling.

## Validation Steps Performed

All new and affected tests were run against a locally deployed build. 46/46 pass.

Unit tests:
- `WSLCCLITerminalUnitTests` — 6 new `Terminal_Confirm*` tests covering lowercase/uppercase `y`, surrounding whitespace, rejection of other answers, end-of-input declining, and verbatim rendering of messages containing format characters.
- `WSLCCLICommandUnitTests` — `PruneCommands_BindShortFToForce` asserts all four prune commands expose `--force`/`-f`, expose `--filter` with no alias, and no longer register `ArgType::Filter`; `ListCommands_KeepShortFOnFilter` guards that the four listing commands still bind `-f` to `--filter`.

E2E tests:
- `WSLCE2EContainerPruneTests` (10) — including 5 new tests: declining at end of input, accepting via piped `y`, `-f` skipping the prompt, `--filter` excluding a non-matching container, and a malformed filter being rejected before the prompt is shown.
- `WSLCE2EImagePruneTests` (7), `WSLCE2ENetworkPruneTests` (10), `WSLCE2EVolumePruneTests` (11) — existing prune invocations updated to pass `--force`, since these run with stdin bound to NUL and would otherwise decline at the prompt. The three malformed-filter tests deliberately omit `--force` to prove validation runs before the prompt.


src/windows/wslc/core/Terminal.cpp
```diff
@@ -80,4 +80,11 @@ std::wstring Terminal::PromptForLine(Level level, std::wstring_view label, bool
     return line.value_or(std::wstring{});
 }
 
+bool Terminal::Confirm(std::wstring_view message)
```
ggarzia-MSFT, line 83: @dkbennett would like your feedback on this part

src/windows/wslc/arguments/SpecParsing.h
```diff
@@ -74,6 +74,10 @@ std::pair<std::string, std::string> ParseDriverOption(const std::wstring& value)
 // Parses a --filter spec ("key=value"); the separator is required.
 std::pair<std::string, std::string> ParseFilter(const std::wstring& value);
 
+// Parses a prune --filter spec ("key=value"); the separator is required. Reports the option as
+// '--filter' only, since prune commands reserve the 'f' alias for --force.
+std::pair<std::string, std::string> ParsePruneFilter(const std::wstring& value);
```
dkbennett: Do we need a separate method for this when it uses the same parsing? The alias difference shouldn't matter for this.

src/windows/wslc/core/CLIExecutionContext.cpp
```diff
@@ -16,6 +16,16 @@ HANDLE CLIExecutionContext::CreateCancelEvent()
...
 }
 
+bool CLIExecutionContext::ConfirmPrune(std::wstring_view warning)
+{
+    if (Args.GetValue<ArgType::Force>())
+    {
+        return true;
+    }
+
+    return Terminal.Confirm(std::format(L"{}\n{}", warning, wsl::shared::Localization::WSLCCLI_PruneConfirmPrompt()));
```
dkbennett: The warning message and the confirm prompt should be separate.

Warning should be on the Warning channel, which goes to stderr and will be colored yellow as other warnings are, while the confirmation prompt should be on output channel which is on stdout and not be colored.

src/windows/wslc/core/CLIExecutionContext.h
```diff
@@ -50,6 +50,11 @@ struct CLIExecutionContext : public wsl::windows::common::ExecutionContext
 
     HANDLE CreateCancelEvent();
 
+    // Returns true when a destructive prune may proceed, either because --force was passed or
+    // because the user accepted the confirmation prompt. End of input declines, so a prune with no
+    // interactive input aborts rather than blocking.
+    bool ConfirmPrune(std::wstring_view warning);
```
dkbennett: I dont think this should be on the execution context

src/windows/wslc/core/Terminal.cpp
dkbennett: this looks good, nice consistent shortcut for confirmation prompts

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -1071,7 +1071,15 @@ void PruneContainers(CLIExecutionContext& context)
     WI_ASSERT(context.Data.Contains(Data::Session));
     auto& session = context.Data.Get<Data::Session>();
 
-    auto result = ContainerService::Prune(session);
+    if (!context.ConfirmPrune(Localization::WSLCCLI_ContainerPruneConfirm()))
```
dkbennett: This should not be on the context, this is a task level distinction.

test/windows/wslc/WSLCCLITerminalUnitTests.cpp
```diff
@@ -552,6 +552,57 @@ class WSLCCLITerminalUnitTests
         VERIFY_ARE_EQUAL(expected, result);
     }
 
+    TEST_METHOD(Terminal_ConfirmAcceptsLowercaseY)
```
dkbennett: Many of these should be combined into a single test IMO, bunch of different confirmation tests do not need to be separate tests, which should speed them up a little bit.

src/windows/wslc/arguments/SpecParsing.h
ggarzia-MSFT: Yes, the error message is different. the normal --filter error mentions the -f alias in it

dkbennett (dismissed):

localization/strings/en-US/Resources.resw
```diff
@@ -3038,8 +3038,41 @@ On first run, creates the file with all settings commented out at their defaults
     <value>Filter output based on conditions provided</value>
   </data>
   <data name="WSLCCLI_InvalidFilterError" xml:space="preserve">
-    <value>Invalid value "{}" for the '-f, --filter' option: bad format of filter (expected name=value)</value>
-    <comment>{FixedPlaceholder="{}"}{Locked="--filter'"}Command line arguments, file names and string inserts should not be translated</comment>  </data>
+    <value>Invalid value "{}" for the '--filter' option: bad format of filter (expected name=value)</value>
+    <comment>{FixedPlaceholder="{}"}{Locked="--filter'"}Command line arguments, file names and string inserts should not be translated</comment>
+  </data>
```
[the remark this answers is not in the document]
ggarzia-MSFT: will address in a followup pr once argumetn aliasing is changed

dkbennett (approved):

src/windows/wslc/core/Main.cpp
```diff
@@ -149,6 +149,11 @@ try
         context.Terminal.Error(L"{}\n", ee.Message());
         return 1;
     }
+    catch (const TerminateException&)
+    {
+        // The user declined a confirmation prompt, so the requested action is not performed.
+        return 0;
```
OneBlue, line 155: It surprised me that we return 0 on cancellation, but that's also docker's behavior so I guess we need to match it

OneBlue (approved):

src/windows/wslc/core/Main.cpp
ggarzia-MSFT: I think the logic is that the command didn't fail in any way, the user just saw the warning and chose not to proceed

PR #41457, opened 2026-08-26
wslc: add --digests to image list for docker parity

ggarzia-MSFT:

Adds `--digests` to `wslc image list` (and its `image ls` / `images` spellings), matching `docker images --digests`. The DIGEST column was previously hardcoded to `<none>` in json output and absent from the table.

## PR Checklist

`WSLCListImagesFlagsDigests` was already defined and already honored end to end — `WSLCSession::ListImages` forwards it to `Docker().ListImages(all, digests, filters)` and populates `WSLCImageInformation::Digest`. The CLI never set the flag and never read the field, so `ToImageOutput` hardcoded `Digest` to `<none>`.

- New `ArgType::Digests` (`--digests`, no short alias, matching docker where only listing commands' `--filter` carries `-f`).
- `models::ImageInformation` gains a `Digest` field, and `ImageService::List` takes a defaulted `bool digests` that sets the flag and populates it.
- Table output adds a DIGEST column **between TAG and IMAGE ID** only when `--digests` is passed, matching docker's `defaultImageTableFormatWithDigest`.
- The service reports a repo digest (`repo@sha256:...`); docker's DIGEST column shows only the digest. New `DigestFromRepoDigest` helper in `ImageModel.h` performs that reduction and is unit tested directly.
- The digest is only populated when the flag is set. This matters for json: docker gates on `needDigest(ctx)`, which is `ctx.Digest || format.Contains("{{.Digest}}")`, and `--format json` satisfies neither. So `image list --format json` still reports `<none>`, and only `--digests --format json` reports a real digest. An earlier revision of this change leaked the digest into plain json output; the tests below lock the gated behavior in.
- Column configs were hoisted into named `constexpr` values so the 5- and 6-column tables cannot drift.

## Validation Steps Performed

All new and existing tests in the touched areas were run against a locally deployed build. 25/25 pass.

Unit tests (`WSLCCLIImageDigestUnitTests`, new file, 3 tests):
- `DigestFromRepoDigest_StripsRepositoryPrefix` — plain, fully qualified, and registry-with-port repositories all reduce to the bare `sha256:` digest. The port case guards the split from being confused by a colon.
- `DigestFromRepoDigest_PassesThroughBareValues` — empty, already-bare, and `<none>` values are untouched.
- `DigestFromRepoDigest_SplitsOnFirstSeparator` — splitting happens once, so a digest is never truncated further.

Unit test (`WSLCCLICommandUnitTests`):
- `ImageListCommand_HasDigestsArgument` — both the `image list` and root-scoped `images` constructions register `--digests` as an optional flag with no short alias.

E2E tests (`WSLCE2EImageListTests`, 22 total, 5 new):
- `Digests_AddsColumnBetweenTagAndImageId` — asserts DIGEST sits after TAG and before IMAGE ID, and that the default listing does not gain the column.
- `Digests_ReportsDigestOnlyWhenRequested` — json reports `<none>` without the flag; anything reported with the flag must be a bare `sha256:` digest with no `repo@` prefix.
- `Digests_TableMatchesJson` — every digest in json output appears in the table's DIGEST column, so the two renderings cannot drift.
- `Digests_QuietStillOutputsIdsOnly` — `--quiet` wins over `--digests`, emitting bare ids with no header, as docker does.
- `Digests_ListedInHelp` — `--digests` and its localized description appear in `image list --help`.

Note on the test images: they are provisioned via `image load` from a tarball, so they carry no repo digest and correctly report `<none>` even with `--digests` — the same as docker for a tar-loaded image. The digest-reduction logic is therefore covered by the unit tests rather than by asserting a live registry digest in e2e, which would make the suite network dependent.


src/windows/wslc/tasks/ImageTasks.cpp
```diff
@@ -222,17 +222,46 @@ void ListImages(CLIExecutionContext& context)
...
 
-        // Create table — only IMAGE ID uses fixed width; other columns shrink to fit the console.
+        constexpr ColumnWidthConfig c_shrink{.Overflow = Shrink};
+
+        // Only IMAGE ID uses fixed width; other columns shrink to fit the console.
         // When --no-trunc is passed, IMAGE ID also shows full length via TruncateId().
+        constexpr ColumnWidthConfig c_imageId{.MinWidth = 12, .MaxWidth = 12, .Overflow = Shrink};
+
+        // Matches docker: DIGEST sits between TAG and IMAGE ID, and is only shown with --digests.
+        if (context.Args.GetValue<ArgType::Digests>())
```
dkbennett: Duplicate tables feels like the wrong thing, recommend adding support to tableoutput to set m_dropEmptyColumns to true and then make digests either emptystring or populated based on the flag presence. Then we have one table definition which dynamically adjusts if we are to show digests in it.

Also add a test to the tableoutput unit tests which verify this empty column behavior doesn't add any additional padding (if it doesn't already exist)

dkbennett (commented):
--quiet and --digests are valid options together and Docker CLI allows both.  --quiet means only image ids are output, but --digests means that an image with multiple digests can appear multiple times in the output (same is true for tags)

Should add tests here which verify the --quiet and --digests combination (and also images with multiple tags should appear multiple times with --quiet). 

This may be a pre-existing issue with quiet that could be addressed here.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1905,82 +1905,110 @@ try
...
+    // digests is reported with neither.
+    struct ImageRow
+    {
+        const docker_schema::Image* Source;
+        std::string Image;
+        std::string Digest;
+    };
 
-    size_t index = 0;
+    std::vector<ImageRow> rows;
```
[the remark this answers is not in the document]
ggarzia-MSFT: we do not know the end size of rows because each row can have multiple digests

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1905,82 +1905,110 @@ try
...
+        if (rows.size() == rowsBefore)
+        {
+            rows.push_back({&e, "<none>:<none>", std::string{}});
+        }
     }
 
-    WI_ASSERT(index == entries);
+    auto output = wil::make_unique_cotaskmem<WSLCImageInformation[]>(rows.size());
+
+    for (size_t index = 0; index < rows.size(); ++index)
```
OneBlue: nit: This can be simplified to: 

```
for (auto& row: output)
{

...
}

File: src/windows/wslcsession/WSLCSession.cpp [hunk 4]
```diff
@@ -1905,82 +1905,110 @@ try
...
 
-    size_t index = 0;
+    std::vector<ImageRow> rows;
     for (const auto& e : images)
     {
-        // Build a map from repo name to digest for this image
-        // RepoDigests format: "repo@sha256:digest"
-        std::map<std::string, std::string> repoToDigest;
+        // RepoDigests format: "repo@sha256:digest". A repository can be referenced by several digests.
+        // References the daemon reports for an unnamed image are not parseable, and are skipped so that
```
OneBlue: Hmm, what do those references actually look like ? Are they just hashes ?

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1905,82 +1905,110 @@ try
...
+            if (!reference.has_value())
+            {
+                continue;
+            }
 
-            // Set digest if available
-            if (!e.RepoDigests.empty())
+            auto repoName = reference->Repository.Name;
+            const auto it = digestsByRepo.find(repoName);
+            taggedRepos.insert(std::move(repoName));
```
OneBlue: nit: This could be: `taggedRepos.emplace(std::move(repoName))`

OneBlue (dismissed):

src/windows/wslcsession/WSLCSession.cpp
ggarzia-MSFT: Yes

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -1953,82 +1953,109 @@ try
...
-                THROW_HR_IF(E_UNEXPECTED, strcpy_s(output[index].Digest, e.RepoDigests[0].c_str()) != 0);
+                rows.push_back({&e, tag, std::string{}});
             }
             else
             {
-                output[index].Digest[0] = '\0';
+                for (const auto& repoDigest : it->second)
+                {
+                    rows.push_back({&e, tag, repoDigest});
+                }
```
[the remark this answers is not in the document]
ggarzia-MSFT: Yes, this is expected behavior

OneBlue (approved):

Issue #41449, opened 2026-08-26
wslservice stuck in stopping state & wsl --shutdown shows the error as OOM


chemwolf6922:
### Windows Version

Microsoft Windows [Version 10.0.26310.28089]

### WSL Version

2.9.9.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Distro Version

6.18.40.1-1

### Other Software

Not related

### Repro Steps

This is not a deterministic issue. This was found during development. Where wsl --shutdown always returns
```
Not enough memory resources are available to complete this operation.
Error code: Wsl/E_OUTOFMEMORY
```
The original trigger is not captured.

### Expected Behavior

wsl shuts down normally.

### Actual Behavior

The error has two folds:
1. The service stuck in the stopping state
The wsl VM was stopped. But the wslc VM was not. Initial analysis shows this could be a dead lock in the wslc shutdown logic.

2. The wsl --shutdown call translates the stopping state to OOM. 
https://github.com/microsoft/WSL/blob/00df65ad829adcdcbe0cb82af5179f67f6234464/src/windows/service/exe/LxssUserSessionFactory.cpp#L168-L174
This logic here translates the stopping state to S_FALSE. And the function is returned with S_FALSE and ppCreated = NULL. This may have caused COM to treat this as an OOM. There was system memory pressure during the repro.

### Diagnostic Logs

Captured wsl log and wsl service dumps were emailed.

PR #41460, opened 2026-08-27
Fix activating a stopping service treated as OOM

chemwolf6922:

The current factory function for LxssUserSession and WSLCSessionManager translates the error code `CO_E_SERVER_STOPPING` to `S_FALSE`. Which combined with `*ppCreated == NULL` causes COM to treat this as an OOM. Leading to error messages like this when activating a stopping service:
```
Not enough memory resources are available to complete this operation.
Error code: Wsl/E_OUTOFMEMORY
```

This PR removes this conversion. So, COM actually retries when the service is stopping. And returns `CO_E_SERVER_EXEC_FAILURE` if the retry times out. 

## Note

The original `S_FALSE will cause COM to retry if the service is stopping.` comment is confusing. There is no public documentation on this behavior. And after reading the Windows source code. It's actually the `CO_E_SERVER_STOPPING` code that will tell COM to retry. And COM will retry up to 4 mins. 
I'm not sure if the original code intended to retry or to not retry. This PR's current change assumes we should retry here. If the intention is to fail fast without the COM retry, a fix would be converting the error code to a failure like `HRESULT_FROM_WIN32(ERROR_SHUTDOWN_IN_PROGRESS)` instead. So it's not treated as OOM by COM and produces correct error messages on the client side.

## PR Checklist

- [x] **Closes:** #41449 partially. Why the WSLC VM got into that stuck stopping state is unclear.

## Validation Steps Performed

Tested by injecting a `CO_E_SERVICE_STOPPING` in the factory. And:
1. Manually activate the COM service.
2. Call wsl.

Before the fix:
The manual activation fails immediately, and wsl call reports `Wsl/E_OUTOFMEMORY`

After the fix:
The manual activation fails after 2-4 minutes*, and wsl call reports `Wsl/CO_E_SERVER_EXEC_FAILURE`

\* The first two run waited for 4 min. The third run waited for 2 min.

### DIFF
--- a/src/windows/service/exe/LxssUserSessionFactory.cpp
+++ b/src/windows/service/exe/LxssUserSessionFactory.cpp
@@ -167,7 +167,6 @@ HRESULT LxssUserSessionFactory::CreateInstance(_In_ IUnknown* pUnkOuter, _In_ RE
     }
     CATCH_RETURN()
 
-
     WSL_LOG("LxssUserSessionCreateInstanceEnd", TraceLoggingLevel(WINEVENT_LEVEL_VERBOSE));
 
     return S_OK;
--- a/src/windows/service/exe/LxssUserSessionFactory.cpp
+++ b/src/windows/service/exe/LxssUserSessionFactory.cpp
@@ -167,6 +167,7 @@ HRESULT LxssUserSessionFactory::CreateInstance(_In_ IUnknown* pUnkOuter, _In_ RE
     }
     CATCH_RETURN()
 
+
     WSL_LOG("LxssUserSessionCreateInstanceEnd", TraceLoggingLevel(WINEVENT_LEVEL_VERBOSE));
 
     return S_OK;
--- a/src/windows/service/exe/LxssUserSessionFactory.cpp
+++ b/src/windows/service/exe/LxssUserSessionFactory.cpp
@@ -165,13 +165,7 @@ HRESULT LxssUserSessionFactory::CreateInstance(_In_ IUnknown* pUnkOuter, _In_ RE
         const auto userSession = wil::MakeOrThrow<LxssUserSession>(instance);
         THROW_IF_FAILED(userSession.CopyTo(riid, ppCreated));
     }
-    catch (...)
-    {
-        const auto result = wil::ResultFromCaughtException();
-
-        // Note: S_FALSE will cause COM to retry if the service is stopping.
-        return result == CO_E_SERVER_STOPPING ? S_FALSE : result;
-    }
+    CATCH_RETURN()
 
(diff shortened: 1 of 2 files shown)


src/windows/service/exe/LxssUserSessionFactory.cpp
```diff
@@ -165,13 +165,7 @@ HRESULT LxssUserSessionFactory::CreateInstance(_In_ IUnknown* pUnkOuter, _In_ RE
...
         THROW_IF_FAILED(userSession.CopyTo(riid, ppCreated));
     }
-    catch (...)
-    {
-        const auto result = wil::ResultFromCaughtException();
-
-        // Note: S_FALSE will cause COM to retry if the service is stopping.
-        return result == CO_E_SERVER_STOPPING ? S_FALSE : result;
-    }
+    CATCH_RETURN()
```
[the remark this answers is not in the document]
chemwolf6922: Activating a stopping service is not something stable enough to be included in test cases.

src/windows/service/exe/LxssUserSessionFactory.cpp
```diff
@@ -165,13 +165,7 @@ HRESULT LxssUserSessionFactory::CreateInstance(_In_ IUnknown* pUnkOuter, _In_ RE
         const auto userSession = wil::MakeOrThrow<LxssUserSession>(instance);
         THROW_IF_FAILED(userSession.CopyTo(riid, ppCreated));
     }
-    catch (...)
-    {
-        const auto result = wil::ResultFromCaughtException();
-
-        // Note: S_FALSE will cause COM to retry if the service is stopping.
```
benhillis, line 172: Interesting, so from your experimentation this has just been incorrect the whole time?

chemwolf6922: Not from experimentation, it's from reading the COM implementation code. I'll try to inject a failure and test this out.

chemwolf6922: Updated the description with manual tests. The experiments match the code review conclusions.

JohnMcPMS (approved):

src/windows/service/exe/LxssUserSessionFactory.cpp
chemwolf6922: Sorry... The changes were not saved. Now it's updated.

OneBlue (commented):
Great catch ! This is an issue that's been there for a long time. Thank you for getting to the bottom of it

chemwolf6922:
Hi @OneBlue . I think this PR do need a review from the wsl maintainers group. John's approval isn't allowing me to complete it. Please help take a look. Thanks.

OneBlue (approved):

PR #41468, opened 2026-08-28
Add WSLC_SESSION env for the global --session option

chemwolf6922:

Currently to work with an application session, the users need to prepend the --session option to every command they call. This is especially complex for adding the "attach to application session containers" feature to Dev Containers.

This PR adds a WSLC_SESSION environment for the --session option. When it's present and the --session option is not, the command uses its value for the --session option.
This PR also adds the ArgumentValueSource enum to track where each argument is coming from. This helps with better error messages.

Please check internal debug experience docs for details.

## PR Checklist

## Validation Steps Performed

Add tests:
WSLCCLIEnvironmentOptionsUnitTests::WSLCCLIEnvironmentOptionsUnitTests::ApplyEnvironmentOptions_Session_SetsValue
WSLCCLIEnvironmentOptionsUnitTests::WSLCCLIEnvironmentOptionsUnitTests::ApplyEnvironmentOptions_SessionAbsent_DoesNotSetValue
WSLCCLIEnvironmentOptionsUnitTests::WSLCCLIEnvironmentOptionsUnitTests::ApplyEnvironmentOptions_SessionEmptyValue_DoesNotSetValue
WSLCCLIEnvironmentOptionsUnitTests::WSLCCLIEnvironmentOptionsUnitTests::ApplyEnvironmentOptions_CliSessionOverridesEnvironmentSession
WSLCE2ETests::WSLCE2EGlobalTests::WSLCE2E_Session_TargetingFromEnvironment_NotFound
WSLCE2ETests::WSLCE2ESessionEnterTests::WSLCE2E_SessionEnter_EnvironmentSessionReportsSource

Update tests:
WSLCCLIEnvironmentOptionsUnitTests::WSLCCLIEnvironmentOptionsUnitTests::ApplyEnvironmentOptions_TargetAlreadyContainsArg_LeavesItUntouched
WSLCE2ETests::WSLCE2EGlobalTests::WSLCE2E_Session_Targeting


src/windows/wslc/commands/RootCommand.cpp
```diff
@@ -88,6 +88,7 @@ std::vector<Argument> RootCommand::GetEnvArguments() const
 {
     return {
         Argument::Create(ArgType::NoColor),
+        Argument::Create(ArgType::Session),
```
dkbennett: This needs to be removed. Session is already a global argument and they are merged in GetGlobalsAndEnvArguments which is called in main to set the global args.

EnvArguments are environment-only arguments that do not otherwise exist in the global namespace unless they are defined in an environment variable. It is unlikely we will ever have more than NoColor, but if we do it will likely be another accessibility-related option.

test/windows/wslc/WSLCCLICommandUnitTests.cpp
```diff
@@ -210,15 +210,17 @@ class WSLCCLICommandUnitTests
         VERIFY_ARE_EQUAL(Kind::Value, globals[0].Kind());
     }
 
-    // RootCommand exposes NoColor as the sole env-eligible global option.
-    TEST_METHOD(RootCommand_EnvArguments_OnlyNoColor)
+    // RootCommand exposes only the options with supported environment bindings.
+    TEST_METHOD(RootCommand_EnvArguments_NoColorAndSession)
```
dkbennett: Also no need for this since should not add to this arg list.

dkbennett (dismissed):
I had already proposed adding this environment variable support for session, but this was explicitly rejected by the team. The primary reason is that it is easy to get into an unexpected state without more support in the CLI for handling this and making it visible to the user whenever session is being set by the environment. Also for no-color support even an empty environment variable is valid, which it is not for Session, leading to obscure errors if the environment variable is not completely removed.

At a minimum, we'd want to have a way to easily reset the environment variable (delete it) from the CLI, and possibly issue diagnostic output whenever session is being set by the environment so it is obvious to the user which session is being used. We just need more careful environment variable handling, validation, and information whenever environment variable is being used for session targeting for this to be a good user experience.

So I'm marking this as "Request Changes" but realistically this PR must unfortunately be rejected, because not having it is a design decision at this time. We may revisit this in the future as we get customer feedback, but for now session is command line specified only by-design.

chemwolf6922:
Hi @dkbennett . I made the suggested changes and converted this into a draft for now. This was discussed with @craigloewen-msft and @ptrivedi to support the Dev Containers attach scenario. But has not been discussed with the broader team yet.

dkbennett (commented):
Changes look good, the empty value boolean is a good addition. If we get a third behavior type would want to make that an enum but easy change to make later.

Good test with making sure CLI overrides the environment variable. 

One thing I would question is the error case of a non-existent session being specified in an environment variable and it may not be visible to the user.

If session is set by environment variable and that session does not exist, what do we do?

In the --session does-not-exist case, it's a clear argument error on the command line that should be obvious to the user. They specified a session that does not exist. 

In the environment variable case, it may not be obvious to the user that the session specified in the environment is invalid. The user may have set it and forgotten about it and be confused why the session doesn't exist. Or a script set it and they are not aware of it at all. Such environment variables will also span updates to the WSL install, so it seems plausible that this could be encountered by any user who sets the environment variable for a specific session or has a script or tool or app that does it.

One option here is Argument could have an additional member variable added - the value's Source.  I'm expecting this could be an enum with values such as "Default" (if the value's default setting was used) or "CommandLine" or "Environment" or "Policy".  Any call to Add a value to an argument could pass along the Source, then we could differentiate behavior based on the source. For example, when resolving the Session, we could check on error if the source is Environment, and if so display a warning or different error message to the user indicating that the environment variable for session is pointing to an invalid session. Its possible with policy support later that some settings might be policy-controlled so we could have special errors or warnings for that as well (such as not allowing a value set by Policy to be overridden by anything).

This is a bigger change to support a feature like this, but it is one I would want to make if session environment variable is supported so we have clear user messaging when the environment variable is invalid.

A "system session reset/clear/default" command may also be useful that removes any environment variables set as an easy way to get back into a working state if a user encounters this situation.

src/windows/wslc/core/EnvironmentOptions.cpp
```diff
@@ -36,13 +36,26 @@ try
...
             if (arg.Kind() == Kind::Flag)
             {
-                target.Add(arg.Type(), true);
+                target.Add(arg.Type(), true, argument::ArgumentValueSource::Environment);
             }
             else if (arg.Kind() == Kind::Value)
             {
-                target.Add(arg.Type(), std::move(*value));
+                target.Add(arg.Type(), *value, argument::ArgumentValueSource::Environment);
+            }
```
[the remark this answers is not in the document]
chemwolf6922: Value is used later in log.

chemwolf6922:
Hi @dkbennett . Thanks for the input. I added the value source to track where the paramter is coming from for better error messages. However, the session reset part is not very doable as a process normally cannot change it's parent's env.
Please help take a look. Thanks.

OneBlue (changes requested):
@craigloewen-msft: I don't think having a `WSLC_SESSION` is a good idea. This kind of environment override will lead to broken scripts if the environment has a `WSLC_SESSION` set, since scripts will expect to run in the default session if they don't specify --session. 

Could you share a bit more about the issue we're trying to solve with this ? If this is for user convenience, a shell alias can easily solve the issue.

chemwolf6922:
Closing this as we agreed on not having this.

PR #41472, opened 2026-08-28
Fix unnecessary delay in the port tracking loop

### LINKED ISSUE #41470: bind() takes ~10 ms for every socket family under WSL2, independent of networking mode
### Windows Version

Microsoft Windows [Version 10.0.26200.9106]

### WSL Version

2.7.10.0

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.18.33.2-2

### Distro Version

Ubuntu 24.04 (Noble Numbat)

### Other Software

openvpn3-linux v26-3 and v27.1 (OpenVPN core 3.11.5 / 3.11.7) — affected, but only
as a symptom. The benchmark below reproduces the underlying behaviour with nothing
installed beyond Python 3.

### Repro Steps

`bind()` takes ~10 ms on every socket family under WSL2, while `socket()` creation
is free. This is not specific to networking mode — a private (non-mirrored) network
namespace shows the same cost.

1. `.wslconfig`:

```ini
[wsl2]
networkingMode=mirrored
dnsTunneling=true
firewall=false
autoProxy=true

[experimental]
hostAddressLoopback=true
```

2. Save and run this benchmark. `socket()` is timed separately as a control:

```python
import socket, time, statistics, os
N = 200
def bench(label, make, act):
    ts = []
    for _ in range(N):
        s = make()
        t0 = time.perf_counter(); act(s); t1 = time.perf_counter()
        ts.append((t1 - t0) * 1000); s.close()
    ts.sort()
    print(f"{label:36s} median {statistics.median(ts):7.3f} ms  min {ts[0]:6.3f}  max {ts[-1]:7.3f}")

bench("AF_NETLINK bind()",
      lambda: socket.socket(socket.AF_NETLINK, socket.SOCK_RAW, socket.NETLINK_ROUTE),
      lambda s: s.bind((0, 0)))
bench("AF_NETLINK socket() only (control)",
      lambda: socket.socket(socket.AF_NETLINK, socket.SOCK_RAW, socket.NETLINK_ROUTE),
      lambda s: None)
bench("AF_UNIX bind() (abstract)",
      lambda: socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM),
      lambda s: s.bind("\0b%d_%d" % (os.getpid(), time.time_ns())))
bench("AF_INET/UDP bind()",
      lambda: socket.socket(socket.AF_INET, socket.SOCK_DGRAM),
      lambda s: s.bind(("127.0.0.1", 0)))
```

3. To separate mirrored-mode cost from baseline cost, run the same script again
   inside a private network namespace, which is not mirrored:

```bash
sudo unshare -n -- sh -c "ip link set lo up; exec python3 bench.py"
```

### Expected Behavior

`bind()` on `AF_UNIX` and `AF_NETLINK` involves no network stack coordination and
should complete in microseconds, as it does on a conventional Linux kernel.
`AF_INET` `bind()` to an ephemeral port on loopback should likewise be
inexpensive.

### Actual Behavior

```
                            host netns      private netns
AF_NETLINK bind()            10.098 ms       10.092 ms     <- unchanged
AF_UNIX bind()               10.150 ms       10.495 ms     <- unchanged
AF_INET/UDP bind()           21.141 ms       10.500 ms     <- halved
socket() only (control)       0.000 ms        0.000 ms
```

Two independent effects:

1. **A baseline ~10 ms `bind()` cost on every socket family**, identical inside a
   private namespace. Not related to mirrored networking. This is the one with
   real impact.
2. **Mirrored mode adds a second ~10 ms increment, but only to `AF_INET`**
   (21 ms vs 10.5 ms). Plausibly port coordination with the Windows host —
   `AF_UNIX` and `AF_NETLINK` have no ports and are unaffected.

`socket()` measuring ~0 ms rules out general syscall overhead.

> **Correction (see comment below).** I originally wrote that the ~10 ms figure
> looked like a scheduler tick. That is wrong — this kernel is `CONFIG_HZ=250`
> (a 4 ms tick), and 10 ms is not a multiple of 4 ms. A 400-sample histogram is
> sharply unimodal at 10–12 ms for `AF_NETLINK` (97% within 8–12 ms) and at
> 20–22 ms for `AF_INET` (90%), i.e. almost exactly 2×. That points at one and
> two instances of a fixed ~10 ms synchronous cost rather than anything
> timer-aligned. Confirmed independently of Python via `strace -T` on
> `openvpn3-service-netcfg` (C++): 901 `bind()` calls, mean 9.1 ms, max 10.8 ms.

**Real-world impact.** `openvpn3-linux` cannot establish a tunnel at all. Its
`openvpn3-service-netcfg` opens a fresh `AF_NETLINK` socket per configuration
operation; with a split-tunnel profile pushing ~900 routes (a commonplace
corporate configuration) that is ~900 × ~10 ms ≈ 9 s, against a fixed 5 s D-Bus
timeout on its `Establish()` call:

```
901 bind() calls    mean 9.1 ms   max 10.8 ms   total 8.183 s

Error calling NetCfgDevice::Establish(): ... Timeout was reached
Failed configuring TUN device (TUN_IFACE_CREATE)
```

507 of the 900 operations had completed when the caller aborted. The identical
route set installed through a single reused socket (`ip -batch`) takes ~0.4 s, so
the per-`bind()` cost is the entire difference between working and not working.
Switching to NAT mode does not help, since the baseline cost is not
mirrored-mode related. The per-socket churn is also being reported to
openvpn3-linux separately.

### Diagnostic Logs

Per-operation pattern from `strace -f -tt -T` on `openvpn3-service-netcfg`,
repeating once per route:

```
socket(AF_UNIX, SOCK_DGRAM) -> ioctl(SIOCGIFINDEX) -> close
socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE)
  -> setsockopt(SO_SNDBUF) -> setsockopt(SO_RCVBUF)
  -> bind()          <-- ~9-10 ms, dominant cost
  -> getsockname -> sendmsg -> recvmsg -> close
```

Within the 5 s `Establish()` window: 507 `bind()` calls, 4.601 s in `bind()`
alone out of 4.939 s total worker syscall time — 93% of the budget.

Possibly related, though a different mechanism: #41443 (mirrored networking drops
replies from a TUN-based VPN adapter). Same kernel and distro; that one is about
packet delivery, this one about syscall latency.

chemwolf6922:

Before this change, an additional 10ms delay was added in the port tracker loop presumably to make the main loop slower than the worker loop. So, the worker result does not get super dated. This is not reliable. And the 10ms delay also slows down every consecutive bind call.

This PR refactors the thread synchronization method. So, the timing between those two loops is more deterministic. And removes the performance penalty for all bind calls.

## PR Checklist

- [x] **Closes:** #41470

## Validation Steps Performed

Measure 100 consecutive bind calls for different AF.
| Type | Before | After |
| --- | --- | --- |
| netlink | 1030ms | 56ms |
| UDS | 1008ms | 57ms |
| UDP bind 0 | 2174ms | 1135ms |

The extra UDP bind 0 time is from the `getsockname` poll. Which is required.

### DIFF
--- a/src/linux/init/GnsPortTracker.cpp
+++ b/src/linux/init/GnsPortTracker.cpp
@@ -23,8 +23,6 @@
 
 constexpr size_t c_bind_timeout_seconds = 60;
 constexpr auto c_sock_diag_refresh_delay = std::chrono::milliseconds(500);
-constexpr auto c_sock_diag_poll_timeout = std::chrono::milliseconds(10);
-constexpr auto c_bpf_poll_timeout = std::chrono::milliseconds(500);
 
 GnsPortTracker::GnsPortTracker(
     std::shared_ptr<wsl::shared::SocketChannel> hvSocketChannel,
@@ -54,11 +52,10 @@ void GnsPortTracker::RunPortRefresh()
         // Netlink will sometimes return EBUSY. Don't fail for that
         try
         {
-            std::promise<void> resume;
-            auto result = PortRefreshResult{ListAllocatedPorts(), time(nullptr), std::bind(&std::promise<void>::set_value, &resume)};
-            m_allocatedPortsRefresh.set_value(result);
+            auto result = PortRefreshResult{ListAllocatedPorts(), time(nullptr)};
+            m_events.post(TrackerEvent{std::move(result)});
 
-            resume.get_future().wait();
+            m_portRefreshResume.get();
         }
         catch (const NetlinkTransactionError& e)
         {
@@ -74,8 +71,7 @@ void GnsPortTracker::RunPortRefresh()
 
 int GnsPortTracker::ProcessSecCompNotification(seccomp_notif* notification)
 {
-    seccomp_notif notificationCopy = *notification;
-    m_request.post(notificationCopy);
+    m_events.post(TrackerEvent{*notification});
     return m_reply.get();
 }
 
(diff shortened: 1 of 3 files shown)


src/linux/init/waitablevalue.h
```diff
@@ -44,7 +45,7 @@ struct WaitableValue
         {
             m_cv.wait(lck);
         }
-        auto return_value = m_value.value();
+        auto return_value = std::move(m_value.value());
```
keith-horton, line 48: why are we moving the value out of m_value vs. just safely taking a copy?

src/linux/init/GnsPortTracker.h
```diff
@@ -169,10 +170,10 @@ class GnsPortTracker
     std::map<PortAllocation, std::optional<time_t>> m_allocatedPorts;
     std::shared_ptr<wsl::shared::SocketChannel> m_hvSocketChannel;
     NetlinkChannel m_channel;
-    std::promise<PortRefreshResult> m_allocatedPortsRefresh;
 
-    WaitableValue<seccomp_notif> m_request;
+    WaitableValue<TrackerEvent> m_events;
```
FetoiuCatalin: minor: should this be called m_event since it can hold a single one?

FetoiuCatalin:
@OneBlue @chemwolf6922 I started reviewing but will probably finish this on Monday

src/linux/init/GnsPortTracker.cpp
```diff
@@ -88,104 +84,101 @@ void GnsPortTracker::Run()
...
-        {
-            int result = 0;
-            if (bindCall->Request.has_value())
+            auto event = m_events.get();
+            if (const auto* notification = std::get_if<seccomp_notif>(&event))
             {
-                PortAllocation& allocationRequest = bindCall->Request.value();
-                result = HandleRequest(allocationRequest);
-                if (result == 0)
+                int result = EIO;
```
FetoiuCatalin, line 112: can we add a comment that we initialize this to non-zero so that in case of any failure or exception below we deny the bind (as 0 will let it go through to the kernel which we don't want if we fail to process something)

src/linux/init/GnsPortTracker.cpp
```diff
@@ -88,104 +84,101 @@ void GnsPortTracker::Run()
...
+                    }
+                }
+                CATCH_LOG()
+            });
 
-        if (bindCall.has_value())
-        {
-            int result = 0;
-            if (bindCall->Request.has_value())
+            auto event = m_events.get();
```
FetoiuCatalin: can we add a comment that if seccomp notification and a refresh result happen to arrive at the same time m_events will serialize them (instead of overwriting), as it's not very clear without reading the implementation of WaitableValue

src/linux/init/GnsPortTracker.cpp
```diff
@@ -88,104 +84,101 @@ void GnsPortTracker::Run()
...
+                if (bindCall.Request.has_value())
                 {
-                    TrackPort(allocationRequest);
-                    GNS_LOG_INFO(
-                        "Tracking bind call: family ({}) port ({}) protocol ({})",
-                        allocationRequest.Family,
-                        allocationRequest.Port,
-                        allocationRequest.Protocol);
+                    PortAllocation& allocationRequest = bindCall.Request.value();
+                    const auto requestResult = HandleRequest(allocationRequest);
```
FetoiuCatalin: can we assign directly to result instead of using an additional variable?

src/linux/init/GnsPortTracker.cpp
```diff
@@ -88,104 +84,101 @@ void GnsPortTracker::Run()
...
-            // If this loop's iteration had a bind call, it's possible that RefreshAllocatedPort
-            // was called before the bind called was processed. Make sure that the port list
-            // is up to date (If this is called, the next block will schedule another refresh)
-            if (!bindCall.has_value())
+            else
             {
-                OnRefreshAllocatedPorts(refreshResult->Ports, refreshResult->Timestamp);
-            }
-        }
+                auto refreshResult = std::get<PortRefreshResult>(std::move(event));
```
FetoiuCatalin: can we rename this struct and the functions using it to something like "ListPortsResult" ? right now the meaning is a bit conflated between "refreshing the list of bound ports" and "refreshing the allocated ports"

FetoiuCatalin (changes requested):

src/linux/init/GnsPortTracker.h
FetoiuCatalin: actually after understanding this better, should we rename this to m_eventQueue? since it's technically a queue

src/linux/init/GnsPortTracker.cpp
FetoiuCatalin: alternatively, we could rename WaitableValue to something like DepthOneQueue

src/linux/init/GnsPortTracker.cpp
```diff
@@ -86,110 +82,110 @@ void GnsPortTracker::Run()
...
 
-        // Only look at bound ports if there's something to deallocate to avoid wasting cycles
-        if (refreshResult.has_value())
-        {
-            if (!m_allocatedPorts.empty())
-            {
-                future = m_allocatedPortsRefresh.get_future();
-                refreshResult->Resume(); // This will resume the sock_diag thread
-                refreshResult.reset();
+                ReconcileAllocatedPorts(listPortsResult.Ports, listPortsResult.Timestamp);
```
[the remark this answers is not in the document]
FetoiuCatalin: this is an interesting point, but I do not think we should give more priority to seccomp notifications vs listing the current bound ports and applying ReconcileAllocatedPorts. the goal is just to have the state be consistent after both those events

FetoiuCatalin (approved):

chemwolf6922:
Hi @OneBlue Could you please help take a look? Thanks.

keith-horton (approved):

src/linux/init/waitablevalue.h
chemwolf6922: This only carried smaller data before. Now the port listing result also uses this, which could be large. So move would be more efficient now.

OneBlue (approved):

PR #41477, opened 2026-08-28
wslc: add --size to container list for docker parity

ggarzia-MSFT:

Adds the `--size` / `-s` option to `wslc container list` (and its `ls` / `ps` aliases) for docker CLI parity.

`docker ps --size` appends a `SIZE` column showing how much data the container has written to its writable layer, plus the virtual total once the read-only image layers are included. `wslc ps` had no way to surface that, and the underlying `SizeRw` / `SizeRootFs` fields were never requested from or parsed out of the daemon.

## PR Checklist

## Validation Steps Performed

**Manual**, against a local build:

```
> wslc container list --help
  -q  --quiet     Outputs the container IDs only
  -s  --size      Display total file sizes

> wslc container list extra
Found a positional argument when none was expected: 'extra'
```

**Parser coverage** (`CommandLineTestCases.h`) — 12 new cases:

- `container list --size`, `container list -s`, `ps --size`, `container ps -s`
- `--size=true` / `--size=false` explicit forms, `--size=invalid` rejected
- `-as` and `-s -q` alias chains, `--size --no-trunc`
- `container list --size extra` rejected — `ps` takes no positional arguments

**Unit test** (`WSLCCLICommandUnitTests.cpp`):

- `ContainerListCommand_HasSizeArgumentWithDockerAlias` — asserts the registered argument is named `size`, aliased `s`, and is a `Kind::Flag`.
- The pre-existing `AllCommands_NoAmbiguousArgumentNamesOrAliases` tree walk still passes, which is the check that matters here: `ArgType::Signal` also uses `-s`, and this confirms no command registers both.

**End-to-end tests** (`WSLCE2EContainerListTests.cpp`):

- `WSLCE2E_Container_List_SizeOption` — runs a container, confirms plain `container list` has **no** `SIZE` column, then confirms `--size`, `-s`, and `ps --size` all produce the column and that the container's row contains the `(virtual …)` total.
- `WSLCE2E_Container_List_SizeOption_QuietStillOutputsIdsOnly` — `--quiet --size` still emits bare IDs with no column headers.
- `WSLCE2E_Container_List_SizeOption_ListedInHelp` — the flag and its description appear in `container list --help`.


src/windows/service/inc/WSLCShared.idl
```diff
@@ -231,9 +231,10 @@ typedef enum _WSLCListContainersFlags
 {
     WSLCListContainersFlagsNone = 0,
     WSLCListContainersFlagsAll = 1,
```
JohnMcPMS, line 233: It feels like either All should have its name changed or the sizes should just be included in it.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -55,6 +55,27 @@ std::string FormatStatsIo(uint64_t Bytes)
...
+    const auto writable = FormatHumanReadableSize(static_cast<uint64_t>(std::max<LONGLONG>(SizeRw, 0)), c_statsIoPrecision);
+    if (SizeRootFs <= 0)
+    {
+        return writable;
+    }
+
+    const auto total = FormatHumanReadableSize(static_cast<uint64_t>(SizeRootFs), c_statsIoPrecision);
+    if (format == FormatType::Json)
+    {
+        return std::format(L"{} (virtual {})", writable, total);
```
JohnMcPMS, line 73: The JSON value has this format?  It isn't a number (or two numbers)?

JohnMcPMS (approved):
I see that all is named that way by docker... still not a great name but probably not to be changed.

src/windows/wslc/tasks/ContainerTasks.cpp
ggarzia-MSFT: Yes, this is how docker formats the size

src/windows/service/inc/WSLCShared.idl
ggarzia-MSFT: All here refers to the `--all` flag

OneBlue (approved):

PR #41487, opened 2026-08-31
Fix p9 drvfs read only mount regression

### LINKED ISSUE #41475: DrvFs mount regression in WSL 2.9.9: standard mount options fail with "Operation not supported"
### Windows Version

Microsoft Windows [Version 10.0.26200.9106]

### WSL Version

2.9.9.0

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.18.40.1-microsoft-standard-WSL2

### Distro Version

Ubuntu 24.04

### Other Software

usbipd-win

### Repro Steps

See related investigation: dorssel/usbipd-win#1464.

Create the mount point if it does not already exist:

mkdir -p /var/run/usbipd-win

Run the DrvFs mount command used by usbipd-win:

mount -t drvfs -o "ro,umask=222" "C:\Program Files\usbipd-win\WSL" "/var/run/usbipd-win"

This fails:

mount: /run/usbipd-win: mount(2) system call failed: Operation not supported.
       dmesg(1) may have more information after failed mount system call.
<3>WSL (...) ERROR: UtilCreateProcessAndWait:735: /bin/mount failed with status 0x2000

Specifying the options as separate -o arguments also fails:

mount -t drvfs -o ro -o umask=222 "C:\Program Files\usbipd-win\WSL" "/var/run/usbipd-win"

Output:

mount: /run/usbipd-win: mount(2) system call failed: Operation not supported.
       dmesg(1) may have more information after failed mount system call.
<3>WSL (5018 - ) ERROR: UtilCreateProcessAndWait:735: /bin/mount failed with status 0x2000

Adding metadata does not help:

mount -t drvfs -o "ro,metadata,umask=222" "C:\Program Files\usbipd-win\WSL" "/var/run/usbipd-win"

Output:

mount: /run/usbipd-win: mount(2) system call failed: Operation not supported.
       dmesg(1) may have more information after failed mount system call.
<3>WSL (9663 - ) ERROR: UtilCreateProcessAndWait:735: /bin/mount failed with status 0x2000

Even a single standard option fails:

mount -t drvfs -o "ro" "C:\Program Files\usbipd-win\WSL" "/var/run/usbipd-win"

Output:

mount: /run/usbipd-win: mount(2) system call failed: Operation not supported.
       dmesg(1) may have more information after failed mount system call.
<3>WSL (9684 - ) ERROR: UtilCreateProcessAndWait:735: /bin/mount failed with status 0x2000

However, replacing the comma with a semicolon unexpectedly succeeds:

mount -t drvfs -o "ro;umask=222" "C:\Program Files\usbipd-win\WSL" "/var/run/usbipd-win"

The resulting mount is accessible:

ls -la /var/run/usbipd-win
total 216
dr-xr-xr-x  1 root root   4096 Aug 26 13:55 .
drwxr-xr-x 40 root root   1160 Aug 28 09:18 ..
-r--r--r--  1 root root    440 Oct  5  2025 README.md
-r-xr-xr-x  1 root root 162352 Oct  5  2025 usbip
-r-xr-xr-x  1 root root  55768 Oct  5  2025 usbip-auto-attach

There is no relevant output in dmesg after the failed mounts.

### Expected Behavior

DrvFs should accept normal Linux mount option syntax.

In particular, this should succeed:

mount -t drvfs -o "ro,umask=222" "C:\Program Files\usbipd-win\WSL" "/var/run/usbipd-win"

Comma-separated options are the normal syntax for mount -o.

At minimum, a single option should also work:

mount -t drvfs -o "ro" "C:\Program Files\usbipd-win\WSL" "/var/run/usbipd-win"

The original reporter of dorssel/usbipd-win#1464 reports that the comma-separated form worked after downgrading WSL from 2.9.8 to 2.9.4, indicating that this is a regression in recent WSL versions.

### Actual Behavior

On WSL 2.9.9.0, DrvFs mounts using normal mount -o syntax fail with:

mount(2) system call failed: Operation not supported.
<3>WSL (...) ERROR: UtilCreateProcessAndWait:735: /bin/mount failed with status 0x2000

The following forms all fail:

-o "ro,umask=222"
-o ro -o umask=222
-o "ro,metadata,umask=222"
-o "ro"

Unexpectedly, this succeeds:

-o "ro;umask=222"

A semicolon is not the standard separator for Linux mount options, so this appears to be a regression in DrvFs option handling rather than an usbipd-win issue.

This currently breaks usbipd-win's WSL integration because it uses the standard comma-separated DrvFs mount options.

### Diagnostic Logs

There is no relevant output in dmesg after reproducing the failure.

Environment:

WSL version: 2.9.9.0
Kernel version: 6.18.40.1-1
WSLg version: 1.0.79
MSRDC version: 1.2.7214
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release
Windows version: 10.0.26200.9106

chemwolf6922:

#41129 introduces a regression where the ";ro" option is passed to the host for p9 shares. However, that option is not supported by the p9 server and causes the mount to fail.

This PR removes the special handling of "ro" in the common parser. And use `MountParseFlags` to add the required virtio option.

## PR Checklist

- [x] **Closes:** #41475

## Validation Steps Performed

Add tests:
DrvFsTests::WSL2Plan9::DrvfsMountReadOnly
DrvFsTests::WSL2VirtioFs::DrvfsMountReadOnly

### DIFF
--- a/src/linux/init/drvfs.cpp
+++ b/src/linux/init/drvfs.cpp
@@ -235,7 +235,7 @@ Return Value:
     }
 }
 
-std::pair<std::string, std::string> ConvertDrvfsMountOptionsToPlan9(std::string_view Options, const wsl::linux::WslDistributionConfig& Config, bool PropagateHostReadOnlyOption)
+std::pair<std::string, std::string> ConvertDrvfsMountOptionsToPlan9(std::string_view Options, const wsl::linux::WslDistributionConfig& Config)
 
 /*++
 
@@ -249,8 +249,6 @@ Arguments:
     Options - Supplies the DrvFs mount options.
 
     Config - Supplies the distribution configuration.
-
-    PropagateHostReadOnlyOption - Supplies whether the read-only option should be propagated to the host share.
 
 Return Value:
 
@@ -266,17 +264,7 @@ Return Value:
     while (!Options.empty())
     {
         auto Option = UtilStringNextToken(Options, ",");
-        if (Option == "ro")
-        {
-            if (PropagateHostReadOnlyOption)
-            {
-                Plan9Options += ";ro";
-            }
-
-            StandardOptions += "ro,";
-        }
-        else if (
-            (Option == "metadata") || (StartsWith(Option, PLAN9_CASE_OPTION)) || (StartsWith(Option, "uid=")) ||
+        if ((Option == "metadata") || (StartsWith(Option, PLAN9_CASE_OPTION)) || (StartsWith(Option, "uid=")) ||
             (StartsWith(Option, "gid=")) || (StartsWith(Option, "umask=")) || (StartsWith(Option, "dmask=")) ||
             (StartsWith(Option, "fmask=")) || (StartsWith(Option, PLAN9_SYMLINK_ROOT_OPTION)))
         {
@@ -674,7 +662,7 @@ try
     //
 
     std::string MountOptions = "cache=mmap,";
-    auto ParsedOptions = ConvertDrvfsMountOptionsToPlan9(Options ? Options : "", Config, false);
+    auto ParsedOptions = ConvertDrvfsMountOptionsToPlan9(Options ? Options : "", Config);
     Plan9Options += ParsedOptions.first;
     MountOptions += ParsedOptions.second;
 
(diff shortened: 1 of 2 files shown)


benhillis (commented):
I think we can avoid adding `PropagateHostReadOnlyOption` and keep this conversion backend-neutral.

If `ro` falls through into `StandardOptions`, Plan9 gets the normal Linux `MS_RDONLY` flag without sending the unsupported `;ro` host option. Then `MountVirtioFs()` can derive its host-side option from the effective flags:

```cpp
auto [Plan9Options, MountOptions] =
    ConvertDrvfsMountOptionsToPlan9(Options ? Options : "", Config);

const auto ParsedOptions = mountutil::MountParseFlags(MountOptions);
if ((ParsedOptions.MountFlags & MS_RDONLY) != 0)
{
    Plan9Options += ";ro";
}
```

This also preserves normal last-option-wins behavior: `ro,rw` is writable, while `rw,ro` is read-only. With the current token-based propagation, any occurrence of `ro` makes the host share read-only even if a later `rw` overrides it.

I tried this locally on top of master and added coverage for `ro,umask=222`, `ro,rw`, and `rw,ro` in both the Plan9 and virtiofs test classes. The full build and targeted tests pass.

benhillis (approved):
Nice, thanks for taking the feedback.

PR #41494, opened 2026-09-01
wslc: add --all-tags to pull for docker parity

ggarzia-MSFT:

Adds the `--all-tags` / `-a` flag to `wslc pull` and `wslc image pull`. When specified, every tag in the repository is downloaded instead of a single tag.

## PR Checklist

### Behavior
1. A reference that already names a tag or digest is rejected. `wslc pull ubuntu:22.04 --all-tags` fails with `tag can't be used with --all-tags/-a` rather than silently ignoring one of the two.
2. The `Using default tag: latest` line is suppressed. because with `--all-tags` no default tag is applied.
3. The final line prints the repository, not a tagged reference. Docker prints `distributionRef.String()`, and `TagNameOnly` is not applied under `--all-tags`, so `docker pull -a ubuntu` ends with `docker.io/library/ubuntu`.

### Tests

- 11 command-line parser cases covering `--all-tags`, `-a`, both `pull` and `image pull` spellings, `--all-tags=true`, flag-before-positional ordering, and the negative cases `--alltags` and the case-sensitive `-A`.
- `ImagePullCommand_HasAllTagsArgumentWithDockerAlias` unit test.
- New `WSLCE2EImagePullTests.cpp` with `WSLCE2E_Image_Pull_AllTagsListedInHelp`, `WSLCE2E_Image_Pull_AllTagsRejectsTaggedReference`, and `WSLCE2E_Image_Pull_AllTagsRejectsDigestReference`.

## Validation Steps Performed

Full `cmake --build .` succeeds. `wslc image pull --help` was run against the local build and lists `-a  --all-tags  Download all tagged images in the repository`.


src/windows/service/inc/wslc.idl
```diff
@@ -709,7 +709,7 @@ interface IWSLCSession : IUnknown
     HRESULT GetTerminationReason([out] WSLCVirtualMachineTerminationReason* Reason, [out] LPWSTR* Details);
 
     // Image management.
-    HRESULT PullImage([in] LPCSTR Image, [in, unique] LPCSTR RegistryAuthenticationInformation, [in, unique] IProgressCallback* ProgressCallback, [in, unique] IWarningCallback* WarningCallback);
+    HRESULT PullImage([in] LPCSTR Image, [in, unique] LPCSTR RegistryAuthenticationInformation, [in] BOOL AllTags, [in, unique] IProgressCallback* ProgressCallback, [in, unique] IWarningCallback* WarningCallback);
```
OneBlue, line 739: Not a blocker but for things like `AllTags` I recommend using flags, that way it's easy to add more boolean switches in the future

OneBlue (dismissed):

src/windows/service/inc/wslc.idl
ggarzia-MSFT: will make a follow up PR for this and push

jahangirbokhs9:
[jahangirbokhs72@outlook.com](jahangirbokhs72@gmail.com)

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -913,6 +914,19 @@ try
 }
 CATCH_LOG()
 
+void WSLCSession::OnRepositoryImagesCreated(const wslutil::RepositoryReference& Repository) noexcept
+try
+{
+    // An --all-tags pull names a repository, so the images it created are enumerated rather than
+    // derived from the requested reference. Notifying per image (not per tag) keeps one notification
+    // per distinct image; the inspect payload already carries every tag pointing at it.
+    for (const auto& image : m_runtime.Docker().ListImages(false, false, {{"reference", {Repository.Name}}}))
```
OneBlue: Unfortunately doing this here creates a race condition where a user could download all tags in a repo, and then re-tag one of the images and delete the original one, which would allow a user to "hide" a specific image from this callback. 

Instead of doing this, could we use the callback to get the list of the images that were pulled ?

OneBlue: The easiest solution I can think of would be to return a list of processed images out of `StreamImageOperation`

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -796,8 +796,15 @@ catch (...)
     EMIT_USER_WARNING(Localization::MessageWslcInstallCertsFailed(wslutil::GetErrorString(wil::ResultFromCaughtException())));
 }
 
-void WSLCSession::StreamImageOperation(DockerHTTPClient::HTTPRequestContext& requestContext, LPCSTR Image, LPCSTR OperationName, IProgressCallback* ProgressCallback)
+void WSLCSession::StreamImageOperation(
+    DockerHTTPClient::HTTPRequestContext& requestContext,
+    LPCSTR Image,
+    LPCSTR OperationName,
+    IProgressCallback* ProgressCallback,
+    std::vector<std::string>* PulledDigests)
```
OneBlue: nit: I would recommend returning `PulledDigests` to make this a bit simpler for the caller

OneBlue (dismissed):
LGTM. One minor comment (feel free to address in a followup)

OneBlue (approved):

PR #41496, opened 2026-09-01
Fix WSLC parser unit test argument overrides

benhillis:
## Summary
- update the remaining `Argument::Create` parser test call to use `ArgumentOverrides`
- restore compilation after the argument factory API change

## Validation
- `./FormatSource.ps1 -ModifiedOnly $true`
- full build attempted; CMake configuration was blocked because the restored `Microsoft.WSL.Kernel.6.18.40.1-1` package did not contain `build/native/bin/x64/modules.vhd`


beena352 (approved):

dkbennett (approved):

dkbennett:
looks like two PRs fixing the same issue submitted about a minute apart: #41497

PR #41500, opened 2026-09-01
wslc: add --all-tags to push for docker parity

ggarzia-MSFT:

Adds the `--all-tags` / `-a` flag to `wslc push` and `wslc image push` for Docker CLI parity, and aligns the default (no-flag) push behavior with `docker push`.

## PR Checklist

### Behavior change worth calling out

At the Engine API level, "push all tags" is expressed by omitting the `tag` query parameter on `POST /images/{name}/push` — the same thing the moby client does (`if !options.All { query.Set("tag", tag) }`).

`WSLCSession::PushImage` previously forwarded the parsed tag verbatim, so an untagged reference such as `wslc push myrepo/img` already omitted `tag` and pushed **every** tag. That means simply adding `--all-tags` would have made the flag a no-op relative to the default.

This PR therefore also makes the untagged case default to `latest`, matching `docker push` and matching what `wslc pull` already does. After this change:

| Command | Pushes |
|---|---|
| `wslc push myrepo/img` | `myrepo/img:latest` (and prints `Using default tag: latest`) |
| `wslc push myrepo/img:v1` | `myrepo/img:v1` |
| `wslc push myrepo/img --all-tags` | every tag in `myrepo/img` |
| `wslc push myrepo/img:v1 --all-tags` | error: `tag can't be used with --all-tags/-a` |

### Tests

- 11 command-line parser cases covering `--all-tags`, `-a`, both `push` and `image push` spellings, `--all-tags=true`, flag-before-positional ordering, and the negative cases `--alltags` and the case-sensitive `-A`.
- `ImagePushCommand_HasAllTagsArgumentWithDockerAlias` unit test.
- New `WSLCE2EImagePushTests.cpp` with `WSLCE2E_Image_Push_AllTagsListedInHelp`, `WSLCE2E_Image_Push_AllTagsRejectsTaggedReference`, and `WSLCE2E_Image_Push_AllTagsRejectsDigestReference`.

### Note for reviewers

The companion `pull --all-tags` PR introduces the same `ArgType::AllTags` entry and an equivalent `tag can't be used with --all-tags/-a` string under a `WSLCCLI_PullAllTagsWithTagError` key. Whichever lands second should reuse the neutral `WSLCCLI_AllTagsWithTagError` key added here rather than keeping both.

## Validation Steps Performed

Full `cmake --build .` succeeds. `wslc image push --help` was run against the local build and lists `-a  --all-tags  Push all tags of an image to the repository`. Automated coverage above; CI runs the test suite.


test/windows/wslc/e2e/WSLCE2EImagePushTests.cpp
```diff
@@ -0,0 +1,76 @@
...
+        VERIFY_IS_TRUE(result.StderrContainsSubstring(Localization::WSLCCLI_AllTagsWithTagError()));
+    }
+
+    WSLC_TEST_METHOD(WSLCE2E_Image_Push_AllTagsRejectsDigestReference)
+    {
+        const auto result =
+            RunWslc(L"image push debian@sha256:0000000000000000000000000000000000000000000000000000000000000000 -a");
+        result.Verify({.Stdout = L"", .ExitCode = 1});
+        VERIFY_IS_TRUE(result.StderrContainsSubstring(Localization::WSLCCLI_AllTagsWithTagError()));
+    }
```
ptrivedi, line 49: nit: 
```suggestion
    }
    
```

localization/strings/en-US/Resources.resw
```diff
@@ -2980,6 +2980,13 @@ On first run, creates the file with all settings commented out at their defaults
   <data name="WSLCCLI_AllArgDescription" xml:space="preserve">
     <value>Show all regardless of state.</value>
   </data>
+  <data name="WSLCCLI_AllTagsArgDescription" xml:space="preserve">
+    <value>Download all tagged images in the repository</value>
```
ptrivedi:
```suggestion
    <value>Download all tagged images in the repository.</value>
```

test/windows/wslc/e2e/WSLCE2EImagePushTests.cpp
```diff
@@ -0,0 +1,76 @@
...
+        VERIFY_IS_FALSE(taggedResult.StdoutContainsSubstring(L"Using default tag"));
+    }
+
+    WSLC_TEST_METHOD(WSLCE2E_Image_Push_AllTagsSkipsDefaultTag)
+    {
+        // --all-tags pushes every tag in the repository, so no tag is defaulted or reported.
+        const auto result = RunWslc(L"image push does-not-exist --all-tags");
+        result.Verify({.ExitCode = 1});
+        VERIFY_IS_FALSE(result.StdoutContainsSubstring(L"Using default tag"));
+    }
```
ptrivedi, line 76: It would be good to test multiple tags scenario -- creates multiple local tags, runs push --all-tags, and verifies that every tag gets to the registry. There are existing helpers in WSLCE2EImageTagTests.cpp and WSLCE2EPushPullTests.cpp that would help implement this flow

ptrivedi (commented):
few minor comments, lgtm otherwise

ggarzia-MSFT:
> few minor comments, lgtm otherwise

Addressed your comments

ptrivedi:
> ### 🔵 Needs a closer look
> The newly added user-facing localization strings for `--all-tags` include ambiguous/misleading wording that should be clarified before merge.
> 
> Review details
> ### Suppressed comments (2)
> **Previously missed (2)** — in code that hasn't changed since the last review.
> 
> **localization/strings/en-US/Resources.resw:2984**
> 
> * WSLCCLI_AllTagsArgDescription is the default help text for the ArgType::AllTags flag (ArgumentDefinitions.h). The current text is pull-specific ("Download"), which is easy to accidentally surface in push/help output if a command forgets to override the description; consider making this string neutral so it’s correct for both pull and push contexts.
>   **localization/strings/en-US/Resources.resw:2987**
> * The error message "tag can't be used with --all-tags/-a" is ambiguous (there is no "tag" option; this refers to a tagged/digested image reference) and also inconsistent with other CLI errors that use clearer sentence-case phrasing (e.g., "cannot be used"). Rewording to explicitly mention tagged/digested references would make it clearer for users.
> 
> * **Files reviewed:** 14/14 changed files
> * **Comments generated:** 0 new
> * **Review effort level:** Lite

Would be good to address these or provide an explanation of why later/not

ggarzia-MSFT:
> > ### 🔵 Needs a closer look
> > The newly added user-facing localization strings for `--all-tags` include ambiguous/misleading wording that should be clarified before merge.
> > Review details
> > ### Suppressed comments (2)
> > **Previously missed (2)** — in code that hasn't changed since the last review.
> > **localization/strings/en-US/Resources.resw:2984**
> > 
> > * WSLCCLI_AllTagsArgDescription is the default help text for the ArgType::AllTags flag (ArgumentDefinitions.h). The current text is pull-specific ("Download"), which is easy to accidentally surface in push/help output if a command forgets to override the description; consider making this string neutral so it’s correct for both pull and push contexts.
> >   **localization/strings/en-US/Resources.resw:2987**
> > * The error message "tag can't be used with --all-tags/-a" is ambiguous (there is no "tag" option; this refers to a tagged/digested image reference) and also inconsistent with other CLI errors that use clearer sentence-case phrasing (e.g., "cannot be used"). Rewording to explicitly mention tagged/digested references would make it clearer for users.
> > * **Files reviewed:** 14/14 changed files
> > * **Comments generated:** 0 new
> > * **Review effort level:** Lite
> 
> Would be good to address these or provide an explanation of why later/not

The first comment is factually wrong. The description is overridden at `ImagePushCommand.cpp:31`

The error message in the second comment is consistent with docker.

localization/strings/en-US/Resources.resw
```diff
@@ -2980,6 +2980,13 @@ On first run, creates the file with all settings commented out at their defaults
   <data name="WSLCCLI_AllArgDescription" xml:space="preserve">
     <value>Show all regardless of state.</value>
   </data>
+  <data name="WSLCCLI_AllTagsArgDescription" xml:space="preserve">
+    <value>Download all tagged images in the repository.</value>
```
OneBlue, line 2984: `Download` is a bit misleading since this applies to push and pull. Maybe we could just edit to `Download or upload all tagged images in the repository.` ?

src/windows/service/inc/wslc.idl
```diff
@@ -769,7 +769,7 @@ interface IWSLCSession : IUnknown
     HRESULT InspectVolume([in] LPCSTR Name, [out] LPSTR* Output);
 
     HRESULT Authenticate([in] LPCSTR ServerAddress, [in] LPCSTR Username, [in] LPCSTR Password, [out] LPSTR* IdentityToken);
-    HRESULT PushImage([in] LPCSTR Image, [in] LPCSTR RegistryAuthenticationInformation, [in, unique] IProgressCallback* ProgressCallback, [in, unique] IWarningCallback* WarningCallback);
+    HRESULT PushImage([in] LPCSTR Image, [in] LPCSTR RegistryAuthenticationInformation, [in] BOOL AllTags, [in, unique] IProgressCallback* ProgressCallback, [in, unique] IWarningCallback* WarningCallback);
```
OneBlue, line 772: Same feedback as the pull version: I recommend using flags for boolean switches in the COM API, but not a blocker for this change

OneBlue (approved):
LGTM, minor comments

localization/strings/en-US/Resources.resw
ggarzia-MSFT: This description is overridden at `ImagePushCommand.cpp:31`

src/windows/service/inc/wslc.idl
ggarzia-MSFT: Will make a follow up PR for this and pull

PR #41501, opened 2026-09-01
wslc: add --follow-link to cp

ggarzia-MSFT:

Adds `--follow-link` / `-L` to `wslc container cp`

Without `-L` the behavior is unchanged: the symlink itself is copied.

## PR Checklist

### CLI

* `ArgumentDefinitions.h` gains `ArgType::FollowLink` (`--follow-link`, alias `L`, `Kind::Flag`).
* `ContainerCpCommand` registers it alongside the existing `--archive`.

### Container &rarr; local: statting a path over the Docker Engine API

`StatArchivePath` issues `HEAD /containers/{id}/archive?path=...` and reads the `X-Docker-Container-Path-Stat` response header. This is the same call `ContainerStatPath` makes in the engine's own client — `cli.head(ctx, "/containers/"+containerID+"/archive", query, nil)` in `moby/moby:client/container_copy.go` — which is what `docker cp` uses to resolve a symlink.

`HEAD` is a good fit for `DockerHTTPClient`: `SendRequest` returns as soon as the response *header* has been consumed and parsed, handing the still-open socket back to the caller, so it never waits on a body. `StatArchivePath` drops the socket immediately. Each request gets its own freshly-connected socket via `SendRequestImpl`, so releasing it has no effect on subsequent calls.

On the engine side, the `HEAD` route maps to `headContainersArchive`, which sets the header via `setContainerPathStatHeader` and writes no body (`moby/moby:daemon/server/router/container/copy.go`); `GET` streams the tar in addition to that header. A missing container or path yields `404`, which maps to `std::nullopt`.

One consequence of `HEAD`: error responses carry no body, so `DockerHTTPException` is constructed with an empty body and message and callers have only the status code to work with.

The header is base64-encoded JSON, decoded with `wslutil::Base64Decode` into a new `docker_schema::ContainerPathStat`. The symlink test uses Go's `os.ModeSymlink` bit (`1 << 27`), since the `mode` field is a marshalled Go `os.FileMode`.

When the target is relative, it is joined against the parent directory of the source path — the same rebasing docker performs via `archive.SplitPathDirEntry`.

### Keeping the requested name

A followed link produces an archive named after the link's *target*, while the copy has to keep the name that was asked for. Container &rarr; local extraction therefore goes to a staging directory and the entries are moved up under the requested name afterwards; local &rarr; container stages the target's tree under the link's name (with its own nested links intact) because `tar.exe` cannot rename entries.

The requested name comes from `wsl::windows::common::filesystem::PosixBaseName`. A POSIX name may hold characters that no Windows file name can (`:` and `\` are legal in POSIX), so the copy is rejected with `WSLCCLI_CpSourceNameNotRepresentableError` rather than silently landing under the target's name instead.

### Interface changes

`IWSLCContainer::DownloadArchive` (`wslc.idl`) gains a `FollowLink` parameter:

```idl
HRESULT DownloadArchive([in, string] LPCSTR SrcPath, [in] BOOL FollowLink, [in] WSLCHandle OutHandle);
```

`IWSLCContainer` is internal and non-stable — it is rebuilt and ships in lockstep with its only clients — so changing an existing method's signature is safe here. The SDK-facing `IWSLCCompatContainer` in `WSLCCompat.idl` is untouched.

## Validation Steps Performed

* Full `cmake --build .` &mdash; clean.
* `wslc container cp --help` lists the new option:

  ```
  Options:
    -a  --archive      Archive mode (accepted for Docker CLI compatibility)
    -L  --follow-link  Always follow symlinks in SRC_PATH
    -?  --help         Shows help about the selected command
  ```

* 11 new parser cases in `CommandLineTestCases.h` covering `-L`, `--follow-link`, both copy directions, `=true`/`=false` on both spellings, combination with `-a`, and the negatives `-L=invalid`, `--followlink` and `-l` (aliases are case-sensitive).
* `ContainerCpCommand_HasFollowLinkArgumentWithDockerAlias` pins the name, alias, kind and optionality.
* E2E coverage in `WSLCE2EContainerCpTests.cpp`:
  * `WSLCE2E_Container_Cp_HelpListsFollowLink` asserts the flag is surfaced in help output.
  * `WSLCE2E_Container_Cp_LocalToContainer_FollowLinkCopiesTargetContents` — file link, dereferenced into the container under the link's name.
  * `WSLCE2E_Container_Cp_LocalToContainer_FollowLinkCopiesDirectoryTree` — directory link, staged under the link's name, with a nested link preserved as a link.
  * `WSLCE2E_Container_Cp_ContainerToLocal_FollowLinkCopiesTargetContents` — absolute link target.
  * `WSLCE2E_Container_Cp_ContainerToLocal_FollowLinkResolvesRelativeTarget` — relative link target, which only resolves against the directory holding the link.


src/windows/wslcsession/DockerHTTPClient.cpp
```diff
@@ -436,6 +436,35 @@ std::tuple<uint32_t, wil::unique_socket, bool> DockerHTTPClient::GetArchive(cons
...
 }
 
+std::optional<docker_schema::ContainerPathStat> DockerHTTPClient::StatArchivePath(const std::string& ContainerID, const std::string& Path)
+{
+    auto url = URL::Create("/containers/{}/archive", ContainerID);
+    url.SetParameter("path", Path);
+
+    // The engine reports the stat in a response header, so a HEAD request is enough.
+    auto [response, socket] = SendRequest(verb::head, url, {}, {});
+    socket.reset();
```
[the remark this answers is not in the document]
ggarzia-MSFT: PR desc was out of date, updated

src/windows/wslc/services/ContainerService.cpp
```diff
@@ -838,6 +838,24 @@ void ContainerService::CopyFromContainer(Session& session, const std::string& id
     THROW_IF_FAILED(container->DownloadArchive(srcPath.c_str(), ToCOMInputHandle(outputHandle)));
 }
 
+std::optional<std::string> ContainerService::ResolveContainerSymlink(Session& session, const std::string& id, const std::string& srcPath)
+{
+    [[maybe_unused]] auto operation = session.BeginContainerOperation();
```
OneBlue, line 843: Was there a specific reason to perform the symlink resolution in the wslc.exe side instead of doing it service side ? 

If not, I would recommend doing this via a in `DownloadArchive`. That would have the benefit of avoiding having to plumb the symlink information back to the client, and the operation would happen under the container lock, so there's no risk of it being deleted between the symlink resolution and the start of the copy

ggarzia-MSFT: No, there was not a specific reason.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -56,6 +56,72 @@ std::string FormatStatsIo(uint64_t Bytes)
     return WideToMultiByte(FormatHumanReadableSize(Bytes, c_statsIoPrecision));
 }
 
+// Container paths are POSIX. Returns the last path component, ignoring trailing separators.
+// Dot components and the root carry no name of their own and yield an empty string.
+std::string PosixBaseName(std::string_view Path)
```
dkbennett: This method should probably be in container services as a lower level thing than tasks.

This may be better off in common as well. Non-blocking, can be refactored in a followup.

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -56,6 +56,72 @@ std::string FormatStatsIo(uint64_t Bytes)
...
+
+    if (name == "/" || name == "." || name == "..")
+    {
+        return {};
+    }
+
+    return name.find_first_of("\\/:") == std::string::npos ? name : std::string{};
+}
+
+std::filesystem::path MakeStagingDirectory(const std::filesystem::path& Parent)
```
dkbennett: Same here, this is a lower level thing than tasks, should be in a more specialized helper or the container services level.

dkbennett (dismissed):
This is fine for now, but the entire container cp methods are way too low level to be at the task layer and should be refactored. Several of the helpers should probably go into common as well along with unit tests for them.

ggarzia-MSFT:
> This is fine for now, but the entire container cp methods are way too low level to be at the task layer and should be refactored. Several of the helpers should probably go into common as well along with unit tests for them.

Happy to make a follow up PR to address this if you want me to

dkbennett (changes requested):
Please address the latest comments from Copilot about correctness.

Recommend moving some of these filesystem methods into common and adding unit tests for them to verify the behavior.

As a followup, I think the cp method in general needs to be refactored into the service, way too much stuff going on in the tasks. If you want to do that feel free, otherwise I'll probably do it in a few days.

dkbennett (approved):
This looks good enough to me to checkin. The correctness issues from copilot stem from possible failure cases and robustness in the validation, which can be addressed in a followup. The test coverage looks good as does the refactor.

ggarzia-MSFT:
> This looks good enough to me to checkin. The correctness issues from copilot stem from possible failure cases and robustness in the validation, which can be addressed in a followup. The test coverage looks good as does the refactor.

I will double check the latest copilot comment, but a few of those are incorrect and ours behaves the same as docker

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -495,11 +593,60 @@ void ContainerCp(CLIExecutionContext& context)
...
+
+                // Only '/' and NUL are barred from a POSIX name, so the basename can hold characters that no
+                // Windows file name can. Copying under the resolved target's name instead would silently
+                // produce something other than what was asked for.
+                const auto unrepresentable = [&]() {
+                    constexpr std::wstring_view reserved = L"<>:\"/\\|?*";
+                    for (const auto character : requestedName)
+                    {
+                        if (character < L' ' || reserved.find(character) != std::wstring_view::npos)
+                        {
```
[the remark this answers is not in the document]
ggarzia-MSFT: None of the basenames you presented fail

src/windows/wslc/tasks/ContainerTasks.cpp
```diff
@@ -479,8 +558,27 @@ void ContainerCp(CLIExecutionContext& context)
...
+                    std::error_code cleanupError;
+                    std::filesystem::remove_all(stagingDir, cleanupError);
+                }
+            });
+
+            auto extractRoot = absTarget;
+            if (followLink)
+            {
+                stagingDir = wsl::windows::common::filesystem::MakeStagingDirectory(absTarget);
+                extractRoot = stagingDir;
```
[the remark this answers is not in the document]
ggarzia-MSFT: we match docker's behavior for `a:b` and CON is a valid filename

ggarzia-MSFT:
> > This looks good enough to me to checkin. The correctness issues from copilot stem from possible failure cases and robustness in the validation, which can be addressed in a followup. The test coverage looks good as does the refactor.
> 
> I will double check the latest copilot comment, but a few of those are incorrect and ours behaves the same as docker

We behave the same as docker for all of the cases presented by copilot

PR #41504, opened 2026-09-02
Fix dangling-else bugs in FAIL_FAST_IF and EMIT_USER_WARNING macros

Eamon2009:

`FAIL_FAST_IF` and `EMIT_USER_WARNING` were missing the standard `do { } while(0)` macro guard. When used as the single statement branch of an enclosing `if/else`, the `else` would incorrectly bind to the macro's internal `if` instead of the outer `if`, silently inverting control flow.

Windows counterpart: [#41513](https://github.com/microsoft/WSL/pull/41513)

## PR Checklist

- [x] **Closes:** Fixes dangling-else ambiguity in `FAIL_FAST_IF` and `EMIT_USER_WARNING`

### Changes
- Wrapped `FAIL_FAST_IF` in `do { } while ((void)0, 0)`
- Wrapped `EMIT_USER_WARNING` in `do { } while ((void)0, 0)`

Both macros were previously defined as bare `if` statements, which left callers vulnerable to dangling-else misassociation when used without braces:

```cpp
if (condition)
    EMIT_USER_WARNING("msg");
else
```
The do { ... } while ((void)0, 0) wrapper ensures each macro always expands to a single statement, preventing the outer else from binding to the internal if.

## Validation Steps Performed

- Verified that FAIL_FAST_IF and EMIT_USER_WARNING expand to single statements when used in if/else contexts without braces.
-  Windows part PR #41513 for consistency.

another PR #41342

### DIFF
--- a/src/linux/inc/lxwil.h
+++ b/src/linux/inc/lxwil.h
@@ -16,213 +16,213 @@ namespace wil {
 #define FAIL_FAST() raise(SIGABRT);
 #define FAIL_FAST_CAUGHT_EXCEPTION() FAIL_FAST()
 #define FAIL_FAST_IF(condition) \
-   do { \
-       if ((condition)) \
-       { \
-           FAIL_FAST() \
-       } \
-   } while ((void)0, 0)
+    do { \
+        if ((condition)) \
+        { \
+            FAIL_FAST() \
+        } \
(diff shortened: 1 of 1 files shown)


benhillis (approved):
Looks good, thank you for the contribution.

PR #41513, opened 2026-09-03
Harden Windows macros against dangling-else ambiguity

Eamon2009:

Hardens `EMIT_USER_WARNING` and `THROW_HR_WITH_USER_ERROR_IF` in `src/windows/common/ExecutionContext.h` by wrapping them in `do { ... } while (false)` statement wrappers. This is the Windows counterpart to the Linux fix in [#41504](https://github.com/microsoft/WSL/pull/41504).

## PR Checklist

- [x] **Closes:** Fixes dangling-else vulnerability on Windows (follow-up to #41504)

&lt;!-- Provide a more detailed description of the PR, other things fixed or any additional comments/features here --&gt;

Both `EMIT_USER_WARNING` and `THROW_HR_WITH_USER_ERROR_IF` were defined as bare `if` statements, leaving Windows callers vulnerable to dangling-else misassociation when the macros are used without braces:

```cpp
if (condition)
    EMIT_USER_WARNING(L"msg");
else
    // This else incorrectly binds to the macro's inner if
```
This change wraps both macros in do { ... } while (false) so they always expand to a single statement. This brings the Windows header into parity with the already-hardened Linux definition from #41504 and prevents the outer else from binding to the inner if generated by the macro.

## Validation Steps Performed
- Verified that EMIT_USER_WARNING and THROW_HR_WITH_USER_ERROR_IF now expand to single statements when used in if/else contexts without braces.
- Reviewed the Linux counterpart (#41504) to ensure consistency between platforms.

### DIFF
--- a/src/windows/common/ExecutionContext.h
+++ b/src/windows/common/ExecutionContext.h
@@ -35,7 +35,7 @@ namespace wsl::windows::common {
         { \
             THROW_HR_WITH_USER_ERROR(Result, Message); \
         } \
-    } while (false)
+    } while (false);
 
 #define EMIT_USER_WARNING(Warning) \
     do \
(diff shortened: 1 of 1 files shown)


benhillis (approved):
Change looks good to me, thank you for the contribution!

Issue #41519, opened 2026-09-04
DrvFs without metadata returns  EPERM  from  fchmod()


Zamiell:
### Windows Version

Microsoft Windows [Version 10.0.26100.9168]

### WSL Version

2.7.11.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.18.33.2-2

### Distro Version

Ubuntu 26.04.1 LTS

### Repro Steps

Use a WSL-native source file and a destination directory on a Windows drive mounted through DrvFs without the  metadata  option.

Confirm the mount configuration:

$ mount | grep ' on /mnt/c '
C:\ on /mnt/c type 9p (...,aname=drvfs;...,rw,...)

The mount options should not contain  metadata .

Create a writable source file with mode  0644 :

$ printf 'test\n' > /tmp/copyfile-source
$ chmod 0644 /tmp/copyfile-source
$ stat -c '%a' /tmp/copyfile-source
644

Change to any Windows directory where the current user has write access, then reproduce the failure with Node.js:

$ cd /mnt/c/path/to/a/writable/directory
$ node -e "require('node:fs').copyFileSync('/tmp/copyfile-source', './copyfile-destination')"
node:internal/fs/utils:...
Error: EPERM: operation not permitted, copyfile '/tmp/copyfile-source' -> './copyfile-destination'

The file contents are created successfully despite the reported failure:

$ cat ./copyfile-destination
test

The underlying failure can also be reproduced directly:

$ touch ./chmod-probe
$ stat -c '%a' ./chmod-probe
777
$ chmod 0644 ./chmod-probe
chmod: changing permissions of './chmod-probe': Operation not permitted

This occurs even though the Windows ACL permits the current user to modify the file. Writing the same destination with  fs.writeFileSync()  succeeds, as does copying it with GNU  cp .

Related downstream reports:

• https://github.com/libuv/libuv/pull/5257
• https://github.com/nodejs/node/issues/44261

### Expected Behavior

On a DrvFs mount without metadata support, arbitrary Unix modes cannot be represented. However, an unsupported  fchmod()  request that retains at least one write bit should either succeed as a best-effort operation or otherwise avoid turning an already-successful file copy into a reported failure.

[...]

PR #41521, opened 2026-09-04
Fix: update drvfs gid uid after default uid is set in oobe

chemwolf6922:

For a newly installed distro, the drvfs are initially mount as 0:0. And is not updated after oobe sets the default uid. This could cause access issues before the distro is restarted.

This PR updates the auto mounted drvfs gid uid before completing the oobe. The non-auto mounted shares are not remounted to avoid breaking user set gid uid values.

## PR Checklist

- [x] **Closes:** #41519

## Validation Steps Performed

Update tests:
UnitTests::UnitTests::ModernOOBE


src/linux/init/config.cpp
```diff
@@ -2206,6 +2217,93 @@ Return Value:
...
+    {
+        LOG_ERROR("setns failed {}", errno);
+        return -1;
+    }
+
+    const auto MountedVolumes = ConfigGetMountedDrvFsVolumes();
+    auto VolumesToRemount = AutomountedVolumes;
+    AutomountedVolumes.reset();
+    int Result = 0;
+    for (size_t Index = 0; Index < VolumesToRemount.size(); Index += 1)
```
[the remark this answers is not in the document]
chemwolf6922: This is redundant. To lead to this. That bit must have already been set by an impossible drive letter.

src/linux/init/init.cpp
```diff
@@ -654,6 +654,17 @@ try
             }
         }
 
+        if ((OobeResult == 0) && (defaultUidPresent == ConfigKeyPresence::Present) && (defaultUid >= 0) && UtilIsUtilityVm())
+        {
+            for (const auto Admin : {false, true})
+            {
+                if (ConfigRefreshDrvFsOwner(defaultUid, Admin, Config) < 0)
```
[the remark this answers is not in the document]
chemwolf6922: This new change is using the same uid as ConfigureUid.

test/windows/UnitTests.cpp
```diff
@@ -4679,6 +4679,13 @@ VERSION_ID="Invalid|Format"
         DistroFileChange distributionconf(L"/etc/wsl-distribution.conf", false);
         distributionconf.SetContent(L"[oobe]\ncommand = /bin/bash -c 'echo OOBE'\n");
 
+        constexpr auto drvFsTestPath = L"C:\\oobe-drvfs-test";
+        std::filesystem::create_directory(drvFsTestPath);
+        auto cleanupDrvFsTestPath = wil::scope_exit([&]() {
+            DeleteFileW(std::format(L"{}\\probe", drvFsTestPath).c_str());
+            RemoveDirectoryW(drvFsTestPath);
+        });
```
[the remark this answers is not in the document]
chemwolf6922: The non-elevated part is invalid as there is none. But it's better to put the test folder under temp.

sylvesterkaczmarek (commented):
One failure-path concern: the existing automount is detached before the replacement mount is attempted, and a failed `ConfigMountDrvFsVolumes()` leaves that drive untracked and unmounted while OOBE only logs the refresh failure and continues. Could a remount failure restore the previous mount/options, or keep the volume tracked for retry, so a transient error does not make the drive disappear until restart?

chemwolf6922:
Hi @sylvesterkaczmarek . 
Thanks for the feedback. The concern is valid. However, the rollback is basically the same operation as the remount. So that likely won't help if the remount fails. I made changes to the warning logic, so warnings collected during oobe gets presented to the user. Including any drvfs mount failures.

sylvesterkaczmarek:
Thanks, surfacing the OOBE warning is useful, and I agree that a literal rollback can hit the same failing mount operation. The remaining reliability point is narrower: `ConfigRefreshDrvFsOwner()` detaches the existing mount, clears its automount bit, and a failed replacement is not restored, so the warning makes the disappearance visible but does not preserve availability or a retryable desired state. If that transient-drive-loss behavior is an accepted OOBE tradeoff, documenting it would resolve my concern; otherwise the refresh needs a path that does not permanently lose the previous mount on replacement failure.

src/linux/init/config.cpp
```diff
@@ -2206,6 +2217,93 @@ Return Value:
...
+    auto& AutomountedVolumes = ConfigGetAutomountedDrvFsVolumes(Admin);
+    if (!Config.AutoMount || TargetNamespace == -1 || AutomountedVolumes.none())
+    {
+        return 0;
+    }
+
+    wil::unique_fd OriginalNamespace{UtilOpenMountNamespace()};
+    if (!OriginalNamespace)
+    {
+        return -1;
```
OneBlue, line 2256: nit: We should log an error if we ever hit this

src/linux/init/init.cpp
```diff
@@ -654,6 +658,23 @@ try
             }
         }
 
+        if ((OobeResult == 0) && (defaultUidPresent == ConfigKeyPresence::Present) && (defaultUid >= 0) && UtilIsUtilityVm())
```
OneBlue, line 661: If recommend skipping this is the new default UID is the same as what the drvfs shares were originally mounted with

sylvesterkaczmarek: Agreed. Skipping the refresh when the new default UID already matches the current DrvFs owner avoids an unnecessary detach/remount cycle. My remaining concern is only the failure path after a real ownership change begins.

src/linux/init/config.cpp
chemwolf6922: The error is already logged in the UtilOpenMountNamespace.

src/linux/init/init.cpp
chemwolf6922: Added per mount skipping logic in the refresh function. It may not be reliable to assume the original uid:gid here so the skipping logic is not put here.

src/linux/init/config.cpp
```diff
@@ -2206,6 +2219,103 @@ Return Value:
...
+            Result = -1;
+            ++Iterator;
+            continue;
+        }
+
+        VolumesToRemount.set(Index);
+        Iterator = AutomountedVolumes.erase(Iterator);
+    }
+
+    ConfigMountDrvFsVolumes(VolumesToRemount.to_ulong(), OwnerUid, Admin, Config);
```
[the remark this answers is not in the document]
chemwolf6922: Already discussed this above. The rollback is still a remount. Which likely won't help. The mount warning is flushed to the user. OOBE should not fail because of this.

Issue #41529, opened 2026-09-07
WSL 2.7.12 → 2.7.13 automatic MSI upgrade leaves runtime VHDs pending deletion; ERROR_FILE_NOT_FOUND after reboot


kungfudaibi:
### Windows Version

Windows 11 25H2, 10.0.26200.8037 (Hyper-V guest)

### WSL Version

2.7.12.0 → 2.7.13.0

### Are you using WSL 1 or WSL 2?

- [x] WSL 2
- [ ] WSL 1

### Kernel Version

6.18.33.2-2 (6.18.33.2-microsoft-standard-WSL2)

### Distro Version

Ubuntu installed with wsl --install ubuntu; exact release not recorded

### Other Software

Hyper-V with nested virtualization; PowerShell test clients. No Docker Desktop required.

### Repro Steps

We reproduced a post-reboot failure after the packaged automatic updater upgraded WSL 2.7.12.0 to 2.7.13.0 while WSL clients were repeatedly starting. The detailed MSI log and registry/file snapshots show pending deletion of runtime VHDs surviving their replacement.

Environment: disposable generation-2 Hyper-V VM, nested virtualization enabled, 4 vCPUs, 4 GiB RAM. Start from a checkpoint with WSL 2.7.12.0 and a working WSL 2 distro named Ubuntu.

1. Download the official [2.7.13 MSIX bundle](https://github.com/microsoft/WSL/releases/download/2.7.13/Microsoft.WSL_2.7.13.0_x64_ARM64.msixbundle). SHA256: `3c481b5739a988b22cb81f21737d0f496473ee69bffb1885cc52a77a02ce9fa2`.
2. Verify baseline Ubuntu starts with `wsl.exe -d Ubuntu -u root --exec /bin/true`.
3. In elevated guest PowerShell, set `UpgradeLogFile` (REG_SZ) under `HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Lxss\MSI` to an explicit log path, to retain the verbose MSI log. Set `AutoUpgradeViaMsix` (REG_DWORD) to 1.
4. Start three PowerShell jobs. Each runs `wsl.exe -d Ubuntu -u root --exec /bin/sleep 1` repeatedly for 180 seconds, with 200 ms between calls. Each client has a 5-second timeout, after which only that wsl.exe client is killed.
[...]

PR #41532, opened 2026-09-08
Protect legacy automatic MSI upgrades from service reactivation (experimental)

kungfudaibi:

Related to #41529.

During a legacy WSL 2.7.12 → 2.7.13 automatic MSI upgrade with concurrent WSL launches, the old uninstall can leave `system.vhd` and `tools/modules.vhd` scheduled for deletion. The upgraded installation works before reboot, but reboot deletes these runtime files and subsequent startup fails with `Wsl/Service/CreateInstance/CreateVm/HCS/ERROR_FILE_NOT_FOUND`.

This draft adds an experimental guard against WSLService reactivation during the legacy upgrade. **`WSL_EXPERIMENTAL_UPGRADE_GUARD` defaults to OFF.** The default build retains detailed MSI logs for exit code 3010. This is a reviewable prototype, not a complete, merge-ready resolution of every upgrade/startup failure.

## PR Checklist

- [ ] **Closes:** Related to #41529; intentionally does not close it while the mitigation is experimental.
- [ ] **Communication:** The issue was filed; this implementation approach has not been agreed with maintainers.
- [ ] **Tests:** Focused checks below passed, but pre-reboot availability and remaining integration coverage are unresolved.
- [x] **Localization:** No new end-user-facing strings.
- [x] **Dev docs:** Test instructions, evidence excerpts, and limitations added.
- [ ] **Documentation updated:** No user documentation change proposed at this stage.

The opt-in guard serializes cooperating updater processes, saves and flushes the original WSLService start type, and disables activation before invoking the legacy MSI. It does not retain an SCM service handle across MSI removal/recreation. Restoration changes the service only if it is still disabled; a different configuration is preserved.

The recovery journal lives outside the service registry key. A real rollback fixture demonstrated why: storing the journal under that key lost the record during MSI rollback and left the service disabled. With the independent journal, the same expected MSI failure restored the original service start type and cleared the record.

Recovery is attempted at updater startup, and the installed version is rechecked after acquiring the guard to avoid a stale upgrade decision. Busy Windows Installer status defers recovery while retaining the journal. MSI exit code 3010 keeps its diagnostic log even when the silent updater normalizes the result to success.

### Separate startup failure and limitations

The guarded trials also exposed a pre-reboot `E_UNEXPECTED`: both VHDs existed and neither had a pending-delete entry. In trial H, restarting only WSLService restored Ubuntu without repairing or replacing files. ETW recorded `Expected message LxInitMessageCreateSessionResponse, but socket WslCorePort was closed`; client termination events also preceded `CoImpersonateClient` failures with `0x800706E5`.

This is a different immediate failure mechanism from reboot deleting the VHDs. Its relationship to upgrade timing, the guard, or test-client cancellation is **not established**. It is deferred rather than hidden by an unconditional service restart.

Remaining limitations:

- The Windows Installer service-state check is not atomic with arbitrary external MSI transactions; the mutex coordinates only cooperating updater processes.
- Recovery retries at a subsequent updater start; there is no watchdog.
- Standalone crash/recovery tests passed, but crash recovery through the complete packaged updater was not validated end to end.
- Forced power loss and other installation/update routes were not tested.
- The actual `wslinstaller` target was built, not the complete WSL distribution.

## Validation Steps Performed

Environment: disposable Hyper-V VM, Windows 11 25H2 build 26200.8037, four virtual CPUs, 4 GiB RAM, nested virtualization, Ubuntu on WSL 2. Baseline was WSL 2.7.12; target was 2.7.13.

For each controlled trial, the same clean checkpoint was restored. The actual x64 Release installer was built from implementation commit `945900be3e8231d23b60febd0ffac26f3ffb35d4` with the guard ON or OFF and substituted into the official MSIX layout, locally re-signed for the test VM. The embedded official MSI was unchanged. Three clients launched `wsl -d Ubuntu -u root --exec /bin/sleep 1` for 180 seconds with a 200 ms gap and a five-second per-launch timeout followed by termination; package deployment began three seconds after the clients started. These are individual controlled trials, not a claim of statistical reliability.

| Trial | Guard | Raw MSI exit | Before Windows reboot | After Windows reboot |
| --- | --- | --- | --- | --- |
| F, same-source control | OFF | 3010 | Ubuntu worked; both VHDs were pending deletion | Both VHDs missing; original HCS `ERROR_FILE_NOT_FOUND` |
| E | ON | 0 | VHDs present, no VHD pending deletes; `E_UNEXPECTED` | VHDs present; Ubuntu and version checks passed |
| G, traced | ON | 0 | VHDs present, no VHD pending deletes; `E_UNEXPECTED` | VHDs present; Ubuntu and version checks passed |
| H, full trace | ON | 0 | Same startup error; a diagnostic WSLService-only restart restored Ubuntu before reboot | VHDs present; Ubuntu and version checks passed |

The control MSI log explicitly records `Info 1903` scheduling deletion of both `C:\Program Files\WSL\system.vhd` and `C:\Program Files\WSL\tools\modules.vhd`. Matching pending-file-operation entries had empty destinations. Both files existed before reboot and were absent afterward. Guarded runs did not schedule these VHD deletions (temporary `Config.Msi` rollback-file entries remained).

Additional validation:

- MSVC 19.44 x64 Release builds of the actual `wslinstaller` target with the option ON and OFF.
- Standalone CTest suite: 1/1 passed; a mutation omitting the disabling operation failed its contract check.
- Fake SCM cases cover original start types, exceptions, already-disabled/replaced/absent services, access and journal-flush failures, invalid journals, and busy-installer recovery deferral.
- Real SCM probes: normal restoration, process crash followed by recovery, concurrent recovery waiting, and crash followed by normal Windows reboot and recovery passed.
- Real MSI rollback fixture: expected exit 1603; old product restored, new product absent, original manual start type restored, journal cleared with the corrected independent journal.
- `clang-format --dry-run --Werror` on the guard/test C++ files and `git diff --check` passed.

See [test instructions](https://github.com/kungfudaibi/WSL/blob/fix/legacy-upgrade-service-race/tools/test/legacy-upgrade-guard/README.md) and [selected evidence and limitations](https://github.com/kungfudaibi/WSL/blob/fix/legacy-upgrade-service-race/tools/test/legacy-upgrade-guard/evidence.md). The original issue contains the initial reproduction log archive and screenshot. New full raw logs and ETL traces remain local; selected excerpts are committed, not the full trace files.

## AI assistance disclosure

**GPT-6 Astra (OpenAI Codex)** assisted with investigation, implementation, test harnesses, analysis, and preparation of this PR. Reported results came from actual compiler, test, and VM runs. This disclosure does not imply maintainer or independent human review.

### Review follow-up validation (3a11230)

Startup recovery now returns an HRESULT through a noexcept exception boundary. Service startup logs failures, and the diagnostic probe checks the result before reporting success. The upgrade constructor still fails closed if recovery cannot complete. The probe also includes <string> directly.

A regression expectation that recovery must not throw failed against the previous implementation. The updated contract tests pass and verify invalid-journal, busy-installer, and access-denied HRESULTs; preservation of the recovery record and disabled state; successful retry; and rejection of a new upgrade while Installer remains busy. CTest: 1/1 passed. Both standalone test/probe targets and the actual x64 Release wslinstaller target with WSL_EXPERIMENTAL_UPGRADE_GUARD=ON built successfully. clang-format and git diff --check passed. The earlier VM upgrade trials were not rerun for this exception-boundary change; their results above belong to the previously identified implementation commit.


kungfudaibi:

benhillis:
@chemwolf6922 - this looks quite similar to your change that disables the service during install. I don't know if this is needed.

OneBlue:
Thank you for doing this ! 

We recently shipped fixes (#40524 and #41415) to the MSI install logic which we think will solve the issue that this change is trying to mitigate, so closing this for now, but feel please to reopen if you still hit this after the new 2.9.X update (the fix isn't in 2.7)

PR #41542, opened 2026-09-08
Enable RedirectionGuard process mitigation

benhillis:
Enable the RedirectionGuard process mitigation in enforce mode as part of the existing WSL process mitigation setup.

This applies the policy consistently to processes that call `ApplyProcessMitigationPolicies`.


OneBlue (approved):
I remember looking into this a while ago and that there was a blocking issue that prevented us from enabling this flag, but I can't remember what it was :). 

Approving since I can't find a good reason not to enable it

benhillis:
> I remember looking into this a while ago and that there was a blocking issue that prevented us from enabling this flag, but I can't remember what it was :).
> 
> Approving since I can't find a good reason not to enable it

That's ominous....

Was it maybe that this isn't supported on some Windows builds we run on? I guess we'll find out once our nightly pass runs.

PR #41548, opened 2026-09-09
Host plugin Plan9 shares as the session user

benhillis:
## Summary

Plugin folder mounts currently use the HCS-managed Plan9 server, which accesses host paths with the service identity. Using the session token with that path is not reliable.

Host plugin shares in a dedicated per-user Plan9 server on a separate port, and pass that port to mini_init when mounting. This preserves the session user's Windows file permissions without changing the existing GPU shares.

Restore the folder-access regression test with both allowed and denied writes.

## Testing

- Full x64 Debug build
- `PluginTests::MountFolderAccess`
- `PluginTests::ApiErrors`
- `PluginTests::Success`

### DIFF
--- a/src/windows/service/exe/WslCoreVm.cpp
+++ b/src/windows/service/exe/WslCoreVm.cpp
@@ -2111,7 +2111,7 @@ void WslCoreVm::MountRootNamespaceFolder(_In_ LPCWSTR HostPath, _In_ LPCWSTR Gue
 
     {
         auto runAsUser = wil::impersonate_token(m_userToken.get());
-        if (!m_pluginPlan9Server)
+        if (!m_pluginPlan9Server || m_pluginPlan9Server->IsRunning() != S_OK)
         {
             auto server =
                 wsl::windows::common::wslutil::CreateComServerAsUser<p9fs::Plan9FileSystem, IPlan9FileSystem>(m_userToken.get());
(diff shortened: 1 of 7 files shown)


ptrivedi (dismissed):

ptrivedi (approved):

craigloewen-msft (approved):

craigloewen-msft (approved):

src/windows/service/exe/WslCoreVm.cpp
```diff
@@ -2101,7 +2107,21 @@ void WslCoreVm::MountRootNamespaceFolder(_In_ LPCWSTR HostPath, _In_ LPCWSTR Gue
...
 
     const auto flags = (ReadOnly ? hcs::Plan9ShareFlags::ReadOnly : hcs::Plan9ShareFlags::None) | hcs::Plan9ShareFlags::AllowOptions;
-    wsl::windows::common::hcs::AddPlan9Share(m_system.get(), Name, Name, HostPath, LX_INIT_UTILITY_VM_PLAN9_PORT, flags);
+    wsl::windows::common::security::EnableTokenPrivilege(m_userToken.get(), SE_CREATE_SYMBOLIC_LINK_NAME);
+
+    {
+        auto runAsUser = wil::impersonate_token(m_userToken.get());
+        if (!m_pluginPlan9Server || m_pluginPlan9Server->IsRunning() != S_OK)
+        {
+            auto server =
```
OneBlue, line 2116: Should we just fail if `m_pluginPlan9Server->IsRunning()` is false ? 

If that was the case, creating a new object would drop pre-existing shares right ?

OneBlue: I'm also not sure if we'd even be able to create a new object without destroying the previous once, since the port would already be in use

OneBlue (approved):

PR #41549, opened 2026-09-10
Add a publishing pipeline

florelis:

Add a new publishing pipeline that consumes the build artifacts from the wsl-github-release pipeline and pushes them to the multiple distribution channels.

## PR Checklist

This pipeline runs automatically after the release build completes, and is gated by manual approvals for each stage of the publishing process. The only stage that runs automatically is the submission to DCAT for internal distribution, which still requires a manual approval on the portal.

There are two versions of the pipeline.
* `wsl-publish` targets the DCAT production environment and uses an official pipeline template. It can only be run from production branches.
* `wsl-publish-ppe` targets the DCAT PPE environment and uses a non-official template. This can be used for testing and development from a dev branch. It skips stages for which we don't have a non-production environment, such as NuGet.

## Validation Steps Performed

Tested publishing to DCAT with the PPE version.


benhillis:
This is a lot bigger than what I was expecting. Can this not be integrated into the existing release pipeline?

OneBlue:
Agreed with Ben's comment. This looks like a refactor of our release pipeline, which we might want to do at some point, but given our release I would prefer a surgical change that adds support for adding DCAT release support on top of the existing pipeline instead

florelis:
I would argue that a separate publish pipeline is the surgical approach, since it doesn't mess much with the existing pipeline so there's less of a chance of breaking it.

If there are issues with the publishing steps, we can fix them and run the publish pipeline against the existing build without tagging a new release. That also made it way easier for me to iterate on this for testing. If we were to add the publishing steps to the existing pipeline and something fails there, fixing it would require a full new build. 

Most of this change is net new. The only part that is a refactor is that I split the job that does flighting to the Store and releases to GitHub. My understanding is that when we publish to DCAT we won't want to publish to the Store, so I would think this is needed. Although I can drop that part if that's preferred.

There's also the fact that the build pipeline is a OneBranch pipeline, and I wanted the new jobs to be 1ESPT for multiple reasons. I believe we'll be asked to move everything from OB to 1ESPT soonish, so I'd rather not add to the pile of what we will need to migrate later.
But more importantly, last time I checked, OB made it way harder to have stages that need to be started manually, which is something we need for automating flighting to multiple rings at different times.

ranm-msft (commented):
Splitting submission prep from the actual pushing reads a lot better than the old combined flight stage, and the empty-gate trick for getting ordering out of manually triggered stages is neat.

Two questions.

`store-flight-stage.yml` still consumes `$(StoreBrokerFlightId)`, but that stage now runs under a different pipeline definition than it did before. I can't find it defined in YAML anywhere, so I assume it's a variable on the release pipeline's definition, and those don't follow a template across to a new definition. Has it been set up on the publish pipeline as well, or moved into a variable group?

Second, are `store_flight` and `nuget` meant to be independent of the DCAT rollout? Both are manual with `dependsOn: []`, so nothing in the pipeline makes them wait on `dcat_internal` or the later rings the way `dcat_preview` and `dcat_broad` do. If they're supposed to follow DCAT validation the gate pattern would cover it; if they're deliberately separate channels for the same validated artifact then ignore me.

.pipelines/wsl-publish.yml
```diff
@@ -0,0 +1,43 @@
+# Publishes the artifacts produced by the wsl-github-release pipeline to production: DCAT PROD, Store,
```
OneBlue, line 1: If I'm understanding correctly, this pipeline is designed to directly push to GA right ? 

I'm not sure that this is something that we want to do from the ADO pipeline itself. The current design is to use the pipeline to push to the first ring (internal), and then once we feel ready manually push to pre-release and then GA

.pipelines/wsl-build-release-onebranch.yml
```diff
@@ -1,14 +1,4 @@
 parameters:
-- name: bypassTests
```
OneBlue, line 2: I would like to keep this feature. It's useful when we have emergency fix and a test failure that we decide to override, or when we want to test the release logic itself and not wait for tests

.pipelines/wsl-publish.yml
```diff
@@ -0,0 +1,43 @@
...
+# and NuGet. This pipeline builds nothing; it only reads artifacts from a completed release build and
+# pushes them where they need to go. See wsl-publish-ppe.yml for the non-production counterpart.
+
+trigger: none
+pr: none
+
+resources:
+  pipelines:
+    - pipeline: releaseBuild
```
OneBlue, line 10: If I'm understanding correctly, this will pick whatever release pipeline ran last and publish it. 

I would prefer to keep the existing behavior of building the package and then publishing it in the same run because using the latest run carries the risk of catching a test run, or an incorrect build (especially if we're releasing backports in the same time frame)

OneBlue (changes requested):

.pipelines/wsl-publish.yml
florelis: It auto-selects the last run by default when you run it manually, but you can also select any run you want.

And the intent is to not run it manually. Instead, it has a trigger so that whenever a build completes successfully, it will run on top of that specific build. We can limit that to only the main branch to make sure it doesn't run for dev builds.

.pipelines/wsl-publish.yml
florelis: No, it does not go directly to GA.

It automatically goes to the first ring. Then, whenever we want, we can start a wider ring from the pipeline itself without going to DCAT (like I imagine you have to go to Partner Center today).

.pipelines/wsl-publish.yml
benhillis: Agreed with Pierre, I like how our release pipeline is currently structured and would like to keep things as a single pipeline rather than multiple queued off each other.

PR #41576, opened 2026-09-11
Add WSLC network lifecycle events

beena352:
Add network lifecycle events to WSLC, including create, connect, disconnect, destroy, and prune events. The implementation supports Docker-compatible network event filters by ID and name, forwards events directly from Docker's event stream preserving Docker's event order and behavior, handles network events across VM restarts, and safely cleans up when network creation fails.

## PR Checklist

- WSLC now records and publishes network events from the Docker event stream
- Network event filters support full and prefix matches for network IDs and names. Events are also kept across VM restart when the event store remains available
- Network operations (create/delete/prune) return as soon as Docker's HTTP response is processed and do not wait for the corresponding event to be recorded. Rollback events (e.g. from a failed create) are forwarded like any other Docker event, matching Docker's own behavior

### DIFF
--- a/src/windows/wslcsession/WSLCSession.cpp
+++ b/src/windows/wslcsession/WSLCSession.cpp
@@ -42,6 +42,7 @@ constexpr auto c_containerdSocket = "/run/containerd/containerd.sock";
 constexpr auto c_storageVhdFilename = wsl::windows::wslc::DefaultStorageVhdName;
 constexpr uint32_t c_progressPrecision = 4;
 constexpr auto c_containerCreateEventTimeout = std::chrono::seconds{60};
+constexpr auto c_networkEventTimeout = std::chrono::seconds{60};
 
 // Default grace period to keep an otherwise-idle VM running before tearing it down (used when the
 // session's IdleTimeoutSec setting is 0/unset). This avoids thrashing the VM (repeated
@@ -459,8 +460,13 @@ try
         std::lock_guard containersLock(m_containersLock);
         std::lock_guard networksLock(m_networksLock);
 
+        // The runtime lock prevents mutators from awaiting publication during teardown.
+        WI_ASSERT(!m_pendingNetworkOperation);
+
         // Network metadata is rebuilt from dockerd on every VM start, so it is always dropped.
         m_networks.clear();
+        m_suppressedNetworkEvents.clear();
+        m_abandonedPruneEventPending = false;
 
         // Container wrappers are kept alive across idle teardown (only cleared on permanent shutdown)
         // so client COM references stay valid; RecoverState reattaches them to the restarted VM.
@@ -502,6 +508,9 @@ try
 
     m_containerEventTracking = m_runtime.Events().RegisterContainerCreate(
         std::bind(&WSLCSession::OnContainerCreated, this, std::placeholders::_1, std::placeholders::_2));
+
+    m_networkEventTracking = m_runtime.Events().RegisterNetworkUpdates(std::bind(
+        &WSLCSession::OnNetworkEvent, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, std::placeholders::_4));
 
     return S_OK;
 }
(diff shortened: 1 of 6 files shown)


src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -2488,6 +2503,179 @@ try
...
+{
+    static const std::map<NetworkEvent, std::string> actions{
+        {NetworkEvent::Create, "create"},
+        {NetworkEvent::Connect, "connect"},
+        {NetworkEvent::Disconnect, "disconnect"},
+        {NetworkEvent::Destroy, "destroy"},
+        {NetworkEvent::Prune, "prune"}};
+
+    // Mutators hold m_networksLock until their state changes are committed.
+    std::lock_guard networksLock{m_networksLock};
```
OneBlue, line 2537: What's the thinking for synchronizing the dockerd network notifications with the local network state ? Could we just forward the network notifications without enforcing synchronization here ?

beena352: This was based on an earlier discussion about keeping event publication synchronized with WSLC’s local network state. However, since list networks reads directly from dockerd and these events are intended to match Docker, I agree that the additional synchronization is unnecessary. I’ll forward the notifications directly. This also means transient create/destroy events from rollback will remain visible, matching Docker behavior.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -3355,7 +3375,7 @@ try
     THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_runtime.HasDocker());
     THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_INVALID_STATE), !m_runtime.HasVm());
 
-    std::lock_guard networksLock(m_networksLock);
+    std::unique_lock networksLock(m_networksLock);
```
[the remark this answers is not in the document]
beena352: This is intentional based on PR feedback. Network events are now forwarded directly from Docker without making create, delete, or prune wait for the event to be recorded. Docker event ordering is still preserved, but completion of the network operation is no longer an event-publication boundary. I’ve updated the PR description to reflect this behavior.

src/windows/wslcsession/WSLCSession.cpp
```diff
@@ -3123,19 +3140,22 @@ try
         THROW_DOCKER_USER_ERROR_MSG(e, "Failed to create network '%hs'", name.c_str());
     }
 
+    // Docker published a create event for this network, and publishes a destroy for the removal below.
+    auto removeNetworkCleanup =
+        wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [this, &name]() { m_runtime.Docker().RemoveNetwork(name); });
```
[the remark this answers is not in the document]
OneBlue: That would be a very unexpected error case so I think this would be acceptable

test/windows/WSLCTests.cpp
```diff
@@ -7299,6 +7326,321 @@ class WSLCTests
...
+        const std::string networkName = "wslc-test-network-events";
+        const std::string containerName = "wslc-test-network-events-container";
+        const std::string imageName = "debian:latest";
+        const std::string networkDriver = "bridge";
+
+        LOG_IF_FAILED(m_defaultSession->DeleteNetwork(networkName.c_str()));
+
+        auto now = [] { return duration_cast<seconds>(system_clock::now().time_since_epoch()).count(); };
+
+        const LONGLONG since = WaitForNextEventSecond();
```
OneBlue: I'm a bit confused by this, could we open a new event stream without a "since", then run the operation that we want to observed, and then read the events from the stream we previously opened ? That should be more resilient than relaying on second-level synchronization across processes 

Or is there something else than I'm missing here ?

beena352: The timestamp boundary was intentional because the current event stream does not capture its position when opened. With  since=0 , the first  GetNext()  can still return older buffered events, especially aggregate prune events that have no network identity to filter on.
I’ve updated the tests to open the stream first while using unique names and marker events to exclude older events. The remaining timeout only prevents the test from waiting indefinitely if an expected event is missing.

test/windows/WSLCTests.cpp
```diff
@@ -7180,27 +7173,73 @@ class WSLCTests
...
         {
-            VERIFY_SUCCEEDED(stream->GetNext(&eventJson));
+            VERIFY_SUCCEEDED(Stream->GetNext(&eventJson));
             events.push_back(wsl::shared::FromJson<wsl::windows::common::wslc_schema::Event>(eventJson.get()));
         }
 
         return events;
     }
 
+    std::vector<wsl::windows::common::wslc_schema::Event> ReadNetworkEventsThroughMarker(IWSLCEventStream* Stream)
```
OneBlue, line 7206: Instead of doing this, I think should we implement David's suggestion in #41608  and add a mechanism to subscribe to events emitted after the command starts. Maybe a special value like "-1" would be the easiest way to accomplish this, and then we wouldn't need this special handling in the tests

OneBlue: That can be done in a followup though

OneBlue (dismissed):

src/windows/service/inc/wslc.idl
```diff
@@ -827,6 +827,10 @@ interface IWSLCSession : IUnknown
     // RPC_E_DISCONNECTED. The client holds the returned token for the whole operation; releasing
     // it (or the client exiting) lets the VM idle-terminate again.
     HRESULT BeginContainerOperation([out] IUnknown** Operation);
+
+    // Used only for testing. Fails the next CreateNetwork after dockerd created the network, so the
```
dkbennett, line 831: Probably more of a question for @OneBlue  - do we want methods that exist solely for testing to be in the IDL?

dkbennett (commented):
`src\windows\wslc\arguments\ArgumentValidation.cpp:52` still allows only `type`, `event`, `container`, and `image`. Existing tests explicitly expect `--filter network=test` to fail. Therefore, the COM tests pass while users cannot exercise the feature through `wslc events`. The correct solution here IMO is to remove the argument validation in the CLI on the value of the keys and let the service be authoritative in that (I know I recommended adding this validation earlier - I was wrong). I believe hte correct place to validate filter keys is in WSLCSession::GetEvents or EventStore::CreateStream.

Recommend removing the enum and passing the original action directly:

```cpp
using NetworkEventCallback = std::function<void(
    const std::string& networkId,
    const std::string& action,
    const std::map<std::string, std::string>& attributes,
    std::int64_t time)>;
```

OnNetworkEvent can then record the action directly and it would automatically support all docker actions and future docker actions without requiring another wslc change.

Also would like an E2E test for the network events using the wslc events command line to verify that network lifecycle events work E2E.

beena352:
> `src\windows\wslc\arguments\ArgumentValidation.cpp:52` still allows only `type`, `event`, `container`, and `image`. Existing tests explicitly expect `--filter network=test` to fail. Therefore, the COM tests pass while users cannot exercise the feature through `wslc events`. The correct solution here IMO is to remove the argument validation in the CLI on the value of the keys and let the service be authoritative in that (I know I recommended adding this validation earlier - I was wrong). I believe hte correct place to validate filter keys is in WSLCSession::GetEvents or EventStore::CreateStream.
> 
> Recommend removing the enum and passing the original action directly:
> 
> ```c++
> using NetworkEventCallback = std::function<void(
>     const std::string& networkId,
>     const std::string& action,
>     const std::map<std::string, std::string>& attributes,
>     std::int64_t time)>;
> ```
> 
> OnNetworkEvent can then record the action directly and it would automatically support all docker actions and future docker actions without requiring another wslc change.
> 
> Also would like an E2E test for the network events using the wslc events command line to verify that network lifecycle events work E2E.

That makes sense. I was originally planning to cover the network-events CLI path in a follow-up, but now wslc events command added, I’ll add that E2E coverage here. I’ll also switch the events command to the shared filter argument and move supported-key validation into the service so  --filter network=...  works through the CLI.

src/windows/service/inc/wslc.idl
OneBlue: Yes that's OK. We do this in other places, like TriggerIdleTermination()

src/windows/wslcsession/WSLCSession.h
```diff
@@ -30,6 +30,7 @@ Module Name:
 #include <list>
 #include <optional>
 #include <unordered_map>
+#include <unordered_set>
```
OneBlue, line 33: I don't think this is needed anmore

test/windows/wslc/e2e/WSLCE2EEventsTests.cpp
```diff
@@ -169,18 +169,86 @@ class WSLCE2EEventsTests
...
+
+        result = RunWslc(std::format(L"network rm {}", networkName));
+        result.Verify({.Stderr = L"", .ExitCode = 0});
+        const auto destroyEvent = expectedEvent(L"destroy");
+        WaitForPseudoConsoleOutput(events, wsl::shared::string::WideToMultiByte(destroyEvent));
+
+        events.SendCtrlBreak();
+        VERIFY_ARE_EQUAL(0, events.Wait());
+        events.VerifyNoErrors();
+        stopReader.release();
```
OneBlue, line 228: nit: Instead of doing this I would recommend replacing ` events.SendCtrlBreak();` by `stopReader.reset()`

OneBlue (approved):

PR #41584, opened 2026-09-12
Don't return sockets to wslc plugin API calls

OneBlue:

This changes updates the WSLC plugin logic to always return regular handles to plugins instead of potentially returning sockets. 

This allows plugins to always call CloseHandle() without having to worry about the underlying handle types

## PR Checklist


OneBlue:
Closed in favor of https://github.com/microsoft/WSL/pull/41626

PR #41597, opened 2026-09-13
wslc: add USB device passthrough (`--usb`, `--privileged`)

### LINKED ISSUE #41031: Add USB device passthrough support for WSLC
Yes. There is currently no apparent way to pass USB devices into WSLC containers.
This limits the usefulness of WSLC for hardware-related workloads.

For example, USB devices can be attached to WSL successfully using usbipd-win:
```usbipd attach --wsl --busid <busid>```
The device is visible inside a normal WSL distribution:
```lsusb```
However, inside a WSLC container, the USB device is not visible:
```lsusb```
returns no devices, and:
```ls /dev/bus/usb```
returns:
```No such file or directory```

Currently, `wslc run --help` provides options such as `--gpus` and `--volume`, but there does not appear to be an equivalent to Docker's `--device` option.
This prevents WSLC containers from being used for many hardware-accelerated Linux workloads, such as robotics, cameras, embedded development, and scientific instruments.

I would like WSLC to provide a way to pass USB devices (or more generally Linux device nodes) into containers.
A Docker-compatible approach would be:
`wslc run --device /dev/bus/usb ...`
or a WSLC-specific mechanism such as:
`wslc run --usb all ...`
The solution should allow devices already attached through `usbipd-win` to WSL to be selectively exposed to WSLC containers.

A more general device passthrough mechanism (similar to Docker's `--device` or `--privileged`) would also be useful for:
- USB cameras (`/dev/video*`)
- USB serial devices (`/dev/ttyUSB*`)
- development boards
- FPGA/JTAG adapters
- industrial sensors

Using Docker Desktop with WSL2 is currently the main alternative.
Docker already supports device passthrough through options such as:
```
--device
--privileged
```
However, this introduces another container runtime layer on top of WSL.

One of the main advantages of WSLC is that it is a native Windows-managed Linux container environment with direct integration into WSL features such as GPU support. Adding device passthrough would make WSLC a more complete alternative for hardware-enabled Linux containers.
Another workaround is running applications directly inside a WSL distribution instead of a container, but this loses the isolation and reproducibility benefits of containers.

**Additional context**
Environment:
- Windows 11
- WSLC version: `2.9.3.0`
- USB forwarding: `usbipd-win`

WSLC has great potential for workloads that combine:
- GPU acceleration
- Linux environments
- Windows hardware integration
Examples:
- ROS2 robotics
- NVIDIA Isaac / FoundationPose
- RealSense and depth cameras
- USB cameras
- embedded Linux development
- FPGA tools

Adding USB/device passthrough support would significantly expand the practical use cases of WSLC
: )

### PR DESCRIPTION
**Pull request by Daniel-Abdi**
Fixes #41031.

Full disclosure: I made heavy use of Claude Code to build this out. I've reviewed and tested myself to ensure that the feature works with the hardware I have on hand, but testing with other usb devices may be warranted. 

## The problem

Containers started by `wslc` can't see USB devices, and there's no way to work
around it from the outside.

`usbipd attach --wsl` doesn't help, because containers don't run in the distro
VM. They get their own. I checked by comparing btime in `/proc/stat` and uptime
across the two, and they're clearly separate machines. Anything attached to the
distro VM is invisible to a container.

Bind mounting doesn't help either. `wslc run -v /dev/ttyUSB0:/dev/ttyUSB0` just
creates an empty directory in the container, since the node doesn't exist in
the container VM for it to bind to.

## The approach

Most of what's needed for USB/IP is already in the container VM. `vhci_hcd` is
there, and Consomme's localhost relay means `127.0.0.1:3240` from inside the VM
lands on `usbipd-win` over on Windows. The missing piece was something to speak
the protocol.

This adds that, across four layers:

1. CLI (`src/windows/wslc/`): a repeatable `--usb <busid>` option and a
   `--privileged` flag, validated the same way `--gpus` already is.
2. Service boundary (`wslc.idl`, `WSLCShared.idl`): bus IDs cross to
   `wslcsession` as a `WSLCStringArray`, and `--privileged` as a new container
   flag bit.
3. Guest (`src/linux/init/WSLCUsb.cpp`, new): the USB/IP client. It loads
   `vhci_hcd`, picks a free virtual port, sends `OP_REQ_IMPORT` to
   `usbipd-win`, hands the socket off to the driver through
   `/sys/devices/platform/vhci_hcd.0/attach`, waits for enumeration, then
   reports back the bus ID along with any character device nodes that showed up
   (`/dev/ttyUSB0` and friends).
4. Container (`WSLCContainer.cpp`): those nodes go to Docker as
   `HostConfig.Devices` with a `c 189:* rwm` cgroup rule, and `/dev/bus/usb` is
   bind mounted so libusb-style raw access works as well.

A mutex guards the attachment table. `WSLC_FORK::Thread` runs `ProcessMessages`
on a detached thread, and that path dispatches the full message list, so attach
and detach can genuinely land on more than one thread. Keeping the lock for the
duration of an attach also stops two of them picking the same free vhci port.

Attachments are reference counted. Several containers can share a device, and
it goes back to Windows once the last one using it is gone. A watcher
re-imports the device if it's unplugged and plugged back in, so it returns on
the same virtual port without anyone having to restart the container.

## Usage

Bind the device once on Windows, from an elevated prompt:

```
usbipd bind --busid 5-1
```

Then name it:

```
wslc run --rm -it --usb 5-1 ubuntu
```

or take everything that's bound:

```
wslc run --rm -it --privileged ubuntu
```

`--privileged` sets Docker's privileged mode too, so it behaves the way people
expect coming from `docker run --privileged`.

## Testing

New unit tests cover argument validation and option parsing.
`WSLCCLIArgumentUnitTests` checks that `ValidateUsb` takes bus IDs and `all`
and rejects malformed values. `WSLCCLIExecutionUnitTests` checks `--usb` and
`--privileged` parsing on both `run` and `create`.

- `WSLCCLI`: 465 of 465 passing, including the five new ones.
- `WSLCTests`: 212 passed, 0 failed, 1 skipped (`PullImageFromDockerHub`, which
  needs network access). Measured before the last rebase.

I also ran it against a CH340 USB serial adapter (`1a86:7523`, bus ID `5-1`):

- `--usb 5-1` and `--privileged` both produce a working `/dev/ttyUSB0`
  (`crw------- root root 188, 0`).
- Two containers held the device at the same time. One detached container kept
  it while a second read the same node.
- Unplugging and replugging while a container held the device recovered on its
  own, with the node still present afterwards and nothing for the user to do.
- `usbipd list` showed bus ID `5-1` back as `Shared` once the last container
  was removed.

An earlier revision of the branch was also checked with
`stty -F /dev/ttyUSB0 115200`, and with two containers sharing a device across
an unplug and replug.

Testing on a real device turned up six bugs that compiled and passed unit tests
regardless. The worst one: the virtual port was originally chosen after the
device had already been requested from `usbipd`, so any later failure left the
device detached from Windows and attached to nothing. The port is now picked
first. The rest were a `modprobe` invocation relying on a `PATH` that init
doesn't have, an unconditional `modalias` read on devices that often have none,
and a device node search that never descended into interface directories.

The same held for the round of changes that came out of review. Three faults
only showed up on hardware: node discovery missed usb-serial entirely, because
its `dev` file sits deeper than the search reached; two containers run back to
back failed the second time, because a device released by the first is missing
for as long as Windows takes to enumerate it again; and the same gap left
`--privileged` with no devices. All three are fixed, and everything above was
re-run against the result.

## Limitations

- The guest reaches `usbipd-win` on `127.0.0.1`, which works because Consomme
  sets up a localhost relay. A session running the wslrelay-based port relay
  instead (`WslcFeatureFlagsPortRelayWslRelay`) does not get that relay, and
  this would not reach the host. The address is chosen on the host side so
  there is a single place to fix that, but I have not tried to carry a
  configured address through to the guest, and `HostLoopback` is not currently
  set for WSLC sessions, so there is no name to use yet. Pointing at a `usbip`
  server on another machine would need the same plumbing.
- `--usb` is selective about which devices it imports, but not about what the
  container can then reach. Raw access is a bind mount of `/dev/bus/usb` with a
  `c 189:* rwm` rule, so a container given one bus ID can open any USB device
  imported into the VM, including ones imported for other containers. Scoping
  that properly means per-device rules that survive a replug, since the device
  number changes, and I would rather agree the approach first than guess at it.
- A device that comes back from a replug under a different node name, `ttyUSB1`
  where it had been `ttyUSB0`, keeps working through `/dev/bus/usb` but its
  class node is not remapped: Docker holds the paths from when the container
  was created. Refreshing those on a running container is not something Docker
  exposes, so it would need a different mechanism.
- `WSLCCompat.idl` is untouched, so `wslcsdk` consumers can't request USB
  devices yet. It's versioned separately and I'd rather not change it without
  your input, but I can add it here if you want.
- A device that changes VID:PID while attached, a board dropping into DFU for
  instance, comes back as something else and won't be re-imported.
- `usbipd bind` is on the user, from an elevated prompt. Nothing here tries to
  elevate on its own.

## Notes for reviewers

The commits go one per layer and should read in order. A few are deliberately
left as their own commits rather than folded in, since they record what testing
and review actually found:

- `fix three faults found running USB passthrough on real hardware`,
  `mount /dev/bus/usb for privileged containers too` and `fix three faults
  found testing the review changes` came out of running this against a real
  device.
- `lock the USB attachment table`, `release devices when a multi-device attach
  fails part way` and `address review feedback on USB passthrough` came out of
  review rather than any failure I hit in practice.

`address review feedback on USB passthrough` is the larger of those. It enforces
the existing `AllowWSLContainerPrivileged` policy at the service boundary,
stops `--privileged` failing on a host with nothing bound, corrects which
imports a container releases when a start fails part way, serializes the USB
messages against each other, and routes the SDK's privileged flag to the new
runtime bit.

I can squash them into the layers they belong to if you'd prefer a tidier
series.


Daniel-Abdi:
> Daniel-Abdi please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

OneBlue:
Thank you for submitting this change. 

There are a couple blocking issues with this: 

- This pull request implements USB support via usbpid, which is an external dependency that the user might not know about. We have a long term plan to add native USB support to WSL so that's how we're planning to implement USB suppport long term (this will give this feature a better UX and better performance)

- This change also implements --privileged support, which we're not looking to implement for now since it could allow containers to escape into the VM's root namespace, which would break our current security model

For these reasons, we've decided to close this change for now. Moving forward, I recommend following [Add a feature or bugfix to WSL](https://github.com/microsoft/WSL/blob/master/CONTRIBUTING.md#add-a-feature-or-bugfix-to-wsl) to discuss the design considerations of changes like this one before publishing code changes and splitting separate features in separate changes so they're easier to review.

Daniel-Abdi:
> Thank you for submitting this change.
> 
> There are a couple blocking issues with this:
> 
> * This pull request implements USB support via usbpid, which is an external dependency that the user might not know about. We have a long term plan to add native USB support to WSL so that's how we're planning to implement USB suppport long term (this will give this feature a better UX and better performance)
> * This change also implements --privileged support, which we're not looking to implement for now since it could allow containers to escape into the VM's root namespace, which would break our current security model
> 
> For these reasons, we've decided to close this change for now. Moving forward, I recommend following [Add a feature or bugfix to WSL](https://github.com/microsoft/WSL/blob/master/CONTRIBUTING.md#add-a-feature-or-bugfix-to-wsl) to discuss the design considerations of changes like this one before publishing code changes and splitting separate features in separate changes so they're easier to review.

Fair enough. Is there any specific timeline on native USB support? USBIPD is already listed as the solution for passing USB devices into WSL, so while I get that it's not ideal, I'm not sure if it's so bad as a stopgap. Ideally USB support for WSLC could be added in with this form of fix before then. Frankly I need USB support for my own workloads and would love to use WSLC instead of Docker (and Docker Desktop). If native USB support is something that's planned for something like next year or even later, then it may be a better idea to have some form of support before then, even if not ideal. 

Regarding the privileged mode, that's fair. Probably shouldn't have packaged it in but it makes USB passthrough to my containers much simpler from my side as a dev. 

Shedding any light on the native USB support would be greatly appreciated.

Thank you for your time

PR #41608, opened 2026-09-14
 Add wslc events command

beena352:
Adds a wslc events command for viewing container events.
The command supports --since, --until, and repeatable --filter options. It displays events in a Docker-style format, supports Ctrl-C cancellation, and streams output immediately when redirected.

## PR Checklist

Adds support for filtering events by type, event action, container, and image. Unknown filter keys return an error instead of being ignored.
The command keeps the session active while events are streamed and supports Ctrl-C while waiting for new events.


src/windows/wslc/services/SessionService.cpp
```diff
@@ -188,6 +283,58 @@ int SessionService::Run(Terminal& terminal, const Session& session, const std::v
...
+    std::thread cancellationThread{CancelCallWhenSignaled, cancelEvent, completedEvent.get(), GetCurrentThreadId()};
+    const auto cancellationCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
+        completedEvent.SetEvent();
+        cancellationThread.join();
+    });
+
+    while (true)
+    {
+        wil::unique_cotaskmem_ansistring eventJson;
+        const auto result = stream->GetNext(&eventJson);
```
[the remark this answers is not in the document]
beena352: Good catch. Fixed the server-side cancellation handling and added a regression test

src/windows/wslc/services/SessionService.cpp
```diff
@@ -16,14 +16,109 @@ Module Name:
...
+                first = false;
+            }
+
+            output.push_back(')');
+        }
+
+        return output;
+    }
+
+    void WriteOutput(HANDLE outputHandle, std::string_view output)
```
dkbennett: use context.Terminal.Output(); do not do write calls directly, the Terminal class has the handles routing, color, and redirection behavior standardized for the CLI. This method should not exist.

src/windows/wslc/services/SessionService.cpp
```diff
@@ -188,6 +283,58 @@ int SessionService::Run(Terminal& terminal, const Session& session, const std::v
...
+    const Session& session, LONGLONG since, LONGLONG until, const std::vector<std::pair<std::string, std::string>>& filterValues, HANDLE cancelEvent)
+{
+    std::vector<WSLCFilter> filterEntries;
+    filterEntries.reserve(filterValues.size());
+    for (const auto& [key, value] : filterValues)
+    {
+        THROW_HR_WITH_USER_ERROR_IF(
+            E_INVALIDARG,
+            Localization::MessageWslcInvalidFilter(MultiByteToWide(key)),
+            std::ranges::find(c_eventFilterKeys, key) == c_eventFilterKeys.end());
```
dkbennett: May consider creating ArgType::EventFilter with a custom validation that it matches what c_eventFilterKeys are valid so we throw before even creating a session if the arguments are invalid.

src/windows/wslc/services/SessionService.h
```diff
@@ -39,6 +39,12 @@ struct SessionService
     static wsl::windows::wslc::models::Session OpenOrCreateDefaultSession(Terminal& terminal);
     // Runs the given command and arguments in a session without a TTY, resolving the executable from PATH.
     static int Run(Terminal& terminal, const wsl::windows::wslc::models::Session& session, const std::vector<std::string>& arguments);
+    static void StreamEvents(
```
dkbennett: Recommend:
```cpp
struct EventStreamOptions
{
    LONGLONG Since{};
    LONGLONG Until{};
    std::vector<std::pair<std::string, std::string>> Filters;
};
```
to keep the signature clean and make adding more options easy should we need to do that. This also fits the pattern elsewhere with run options and create options, etc.

src/windows/wslc/tasks/SessionTasks.cpp
```diff
@@ -97,6 +97,29 @@ void ListSessions(CLIExecutionContext& context)
...
 }
 
+void StreamEvents(CLIExecutionContext& context)
+{
+    using namespace std::chrono;
+
+    WI_ASSERT(context.Data.Contains(Data::Session));
+    auto& session = context.Data.Get<Data::Session>();
+
+    LONGLONG since = floor<seconds>(system_clock::now()).time_since_epoch().count();
```
dkbennett: GetValue has an optional default value you can pass in and is guaranteed to always provide the correct type.

```cpp
auto now = floor<seconds>(system_clock::now()).time_since_epoch().count();
auto since = context.Args.GetValue<ArgType::Since>(now);
auto until = context.Args.GetValue<ArgType::Until>(0);
```

(for until, 0 should be the default value anyway, but doesn't hurt to be explicit)

src/windows/wslc/services/SessionService.cpp
```diff
@@ -188,6 +283,58 @@ int SessionService::Run(Terminal& terminal, const Session& session, const std::v
...
+    const auto disableCallCancellation = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, []() { CoDisableCallCancellation(nullptr); });
+
+    wil::unique_event completedEvent(wil::EventOptions::ManualReset);
+    std::thread cancellationThread{CancelCallWhenSignaled, cancelEvent, completedEvent.get(), GetCurrentThreadId()};
+    const auto cancellationCleanup = wil::scope_exit_log(WI_DIAGNOSTICS_INFO, [&]() {
+        completedEvent.SetEvent();
+        cancellationThread.join();
+    });
+
+    while (true)
```
dkbennett: Please express the stream continuation condition in the loop declaration rather than using an unconditional  while (true)  with a return in the body. For example, loop while  GetNext()  succeeds, then accept  WSLC_E_EVENT_STREAM_FINISHED  as normal completion and propagate any other result. This makes the stream termination condition explicit and avoids an unbounded loop that depends on control flow inside the body.

Also recommend using the FromJson helper, and terminal Output understands narrow strings (at least until Ben's change that removes it comes in, but that isn't in yet AFAIK):

```cpp
HRESULT result = S_OK;
while (SUCCEEDED(result))
{
    wil::unique_cotaskmem_ansistring eventJson;
    result = stream->GetNext(&eventJson);
    if (SUCCEEDED(result))
    {
        const auto event = wsl::shared::FromJson<wslc_schema::Event>(eventJson.get());
        terminal.Output(L"{}\n", FormatEvent(event));
    }
}

THROW_HR_IF(result, result != WSLC_E_EVENT_STREAM_FINISHED);
```

And note that the json parsing can throw, not sure how you want to handle that case, but without a try/catch it's going to unwind and land as a terminating error in the CLI.

src/windows/wslc/services/SessionService.cpp
```diff
@@ -188,6 +283,58 @@ int SessionService::Run(Terminal& terminal, const Session& session, const std::v
...
+    wil::com_ptr<IWSLCEventStream> stream;
+    THROW_IF_FAILED(session.Get()->GetEvents(
+        since, until, filterEntries.empty() ? nullptr : filterEntries.data(), static_cast<ULONG>(filterEntries.size()), &stream));
+
+    // Event output is UTF-8.
+    wsl::windows::common::ConsoleState console;
+    console.SetOutputCodePageUtf8();
+    const auto outputHandle = GetStdHandle(STD_OUTPUT_HANDLE);
+
+    THROW_IF_FAILED(CoEnableCallCancellation(nullptr));
```
dkbennett: I think most of this cancellation machinery is unnecessary. 

Keep context.CreateCancelEvent() as the connection to Main's existing Ctrl-C handling. Pass that event to the blocking service operation directly and remove the separate COM cancellation mechanism. The only reason that's needed is because GetNext() doesn't have it. If it needs cancellation GetNext should support it. I.e. change GetNext():

```cpp
HRESULT GetNext(
    [in, unique, system_handle(sh_event)] HANDLE CancelEvent,
    [out, string] LPSTR* EventJson);
```
See the build commands for reference on cancellation wiring.

src/windows/wslcsession/EventStore.cpp
```diff
@@ -137,17 +144,30 @@ bool EventStore::WaitForEvent(std::unique_lock<std::mutex>& Lock, uint64_t Seque
     // Eviction while parked wakes us too, so the caller reports the gap on its next pass.
     const auto ready = [&] { return m_terminating || SequenceNumber < m_firstSequenceNumber + m_events.size(); };
 
-    if (Until.has_value())
+    while (!ready())
```
dkbennett: Please centralize the wait conditions in named predicates or a small wait-state helper. Cancellation, session termination, event availability, and the until deadline collectively determine whether the stream should continue waiting.

src/windows/wslc/commands/EventsCommand.h
```diff
@@ -0,0 +1,32 @@
...
+Abstract:
+
+    Declaration of the EventsCommand.
+
+--*/
+#pragma once
+#include "Command.h"
+
+namespace wsl::windows::wslc {
+struct EventsCommand final : public Command
```
dkbennett: Docker has "docker system events" which is aliased to "docker events"

So the proper structure for this command should be that it is SystemEventsCommand.cpp under SystemCommand.h, with an alias in the root as "events".

That would match docker exactly.

src/windows/wslc/services/SessionService.cpp
```diff
@@ -16,14 +16,109 @@ Module Name:
...
+        while (!output.empty())
+        {
+            DWORD bytesWritten{};
+            THROW_LAST_ERROR_IF(!WriteFile(outputHandle, output.data(), gsl::narrow_cast<DWORD>(output.size()), &bytesWritten, nullptr));
+            THROW_HR_IF(HRESULT_FROM_WIN32(ERROR_WRITE_FAULT), bytesWritten == 0);
+            output.remove_prefix(bytesWritten);
+        }
+    }
+
+    void CancelCallWhenSignaled(HANDLE cancelEvent, HANDLE completedEvent, DWORD threadId) noexcept
```
dkbennett: I don't think any of this cancel machinery is necessary (more details in other comments)

dkbennett (changes requested):
There's some differences with docker that are notable:
| Area | Docker behavior | PR behavior | Result |
|---|---|---|---|
| Command aliases | `docker system events` and `docker events` | Only `wslc events` | Difference |
| Formatting | Supports `--format`, including `--format json` | Does not accept `--format` | Difference |
| Default stream position | Subscribes from invocation onward without replaying buffered events | Uses the current whole-second timestamp as `since` | Difference; can replay same-second buffered events |
| Timestamp precision | Preserves `timeNano` | Discards `timeNano` and always prints nine zeroes | Difference |
| UTC rendering | Prints `Z` for UTC | Prints `+00:00` | Byte-level difference |
| Container filter | Matches container name, full ID, or ID prefix | Matches only the full actor ID | Difference |
| Image filter | Matches the exact image or repository without its tag | Matches only the exact image string | Difference |
| Label filter | Supports `label=<key>` and `label=<key>=<value>` | Not supported | Difference |
| Event actions | Preserves Docker actions such as `kill`, `die`, and `stop` | Converts Docker `die` into `stop` and omits the distinct `die` event | Difference |
| Kill attributes | Includes the signal when supplied by the engine | Does not include `signal` | Difference |
| Ctrl-C | Silent cancellation without an error message | Prints a cancellation warning and exits with code 1 | Difference |

Command aliases is easy to fix with a little refactoring and file renaming.
Formatting should be easy to add json format similar to the usuall format command.
Stream position behavior should align with docker.
Timestamp / utc rendering should be fixed.
Ctrl+C behavior should be clean and not throw to be caught as cancellation. Cancel is the normal way to terminate so we don't need to print any operation cancelled warning in this case (which will happen if it exits via a cancel throw).

src/windows/wslc/services/SessionService.cpp
```diff
@@ -15,15 +15,67 @@ Module Name:
...
+        using namespace std::chrono;
+
+        const sys_seconds time{seconds{timestamp}};
+        try
+        {
+            auto output = std::format("{:%FT%T.000000000%z}", zoned_time{current_zone(), time});
+            output.insert(output.size() - 2, ":");
+            return output;
+        }
+        catch (...)
```
OneBlue: I recommend scoping down the specific exception so we don't accidently hide an unrelated error here

src/windows/wslcsession/EventStore.cpp
```diff
@@ -167,6 +203,11 @@ std::optional<wsl::windows::common::wslc_schema::Event> EventStore::Get(
 
     while (true)
     {
+        if (IsCancelled(CancelEvent))
```
OneBlue: Do we need an explicit check here ? `WaitForEvent()` also handles cancellations

src/windows/wslcsession/EventStore.cpp
```diff
@@ -131,34 +148,53 @@ std::optional<wsl::windows::common::wslc_schema::Event> EventStore::GetLockHeld(
...
+    const auto shouldWait = [&] { return !m_terminating && !IsCancelled(CancelEvent) && !eventAvailable() && !deadlineReached(); };
 
-    if (Until.has_value())
+    while (shouldWait())
     {
-        if (!m_updated.wait_until(Lock, Until.value(), ready))
+        if (Until.has_value())
         {
-            return false;
+            m_updated.wait_until(Lock, Until.value());
```
OneBlue: Shouldn't we exit the loop if wait_until() returns false here ?

src/windows/wslcsession/EventStore.cpp
```diff
@@ -238,14 +279,19 @@ HRESULT EventStream::RuntimeClassInitialize(
...
 }
 
-HRESULT EventStream::GetNext(LPSTR* EventJson)
+HRESULT EventStream::GetNext(HANDLE CancelEvent, LPSTR* EventJson)
 try
 {
     RETURN_HR_IF_NULL(E_POINTER, EventJson);
     *EventJson = nullptr;
 
+    if (IsCancelled(CancelEvent))
```
OneBlue: Do we need an explicit check here ? This is already handled inside the event store

src/windows/wslcsession/EventStore.cpp
```diff
@@ -131,34 +148,53 @@ std::optional<wsl::windows::common::wslc_schema::Event> EventStore::GetLockHeld(
...
+        {
+            m_updated.wait(Lock);
         }
-    }
-    else
-    {
-        m_updated.wait(Lock, ready);
     }
 
     THROW_HR_IF(E_ABORT, m_terminating);
```
OneBlue: I think we should also throw that error if the cancel event is set

src/windows/wslcsession/EventStore.cpp
```diff
@@ -12,6 +12,18 @@ namespace wsl::windows::service::wslc {
 
 namespace {
 
+    bool IsCancelled(HANDLE CancelEvent)
+    {
+        if (CancelEvent == nullptr)
+        {
```
OneBlue, line 18: I also recommend exiting if the caller's process dies, like what we do at the WSLCSession level

test/windows/wslc/e2e/WSLCE2EEventsTests.cpp
```diff
@@ -0,0 +1,179 @@
...
+            since,
+            until,
+            containerId,
+            DebianImage.NameAndTag()));
+        result.Verify({.Stderr = L"", .ExitCode = 0});
+
+        const auto lines = result.GetStdoutLines();
+        VERIFY_ARE_EQUAL(2u, lines.size());
+        VerifyEventLine(lines[0], std::format(L" container create {} (image={}, name={})", containerId, DebianImage.NameAndTag(), c_eventContainerName));
+        VerifyEventLine(lines[1], std::format(L" container start {} (image={}, name={})", containerId, DebianImage.NameAndTag(), c_eventContainerName));
```
OneBlue, line 165: I also recommend killing the container and validating that we receive the container stop & delete events here

dkbennett (dismissed):
The important requested changes were made, this should be good for now, but I would like to see a followup PR addressing docker behavior alignment, specifically:

- --since should not replay same-second buffered events and start at subscription time
- container filter should accept name and ID prefix
- image=debian should match debian:latest
- support applicable label filters
- support --format json
- preserve nanosecond timestamps
- preserve docker event vocabulary, such as distinct 'die' and 'stop' events.
- omitted since is not equivalent to since=now in Docker, which is what it is here. It should be 0.

Also my feedback earlier was in error in that this can use the other ArgType::Filter as-is and do the validation of the filters provided in the service, not the CLI. The CLI should verify the syntax is correct as key-value pairs but the service should decide what filters it considers valid, not the CLI. This could be addressed at the same time as the above items.

In addition:
- The comments Blue had, such as some of the redundancy in possible cancellation calls.
- the E2E test uses until value that is only one second in the future, which might be timing dependent and create flakiness in a CI test run under heavy load or a slower system.

And anything else that is not in alignment with docker that is reasonable to implement.

src/windows/wslcsession/EventStore.cpp
dkbennett: Docker does not print an error on cancellation for events command, so I believe this should exit quietly and successfully if the user terminates the events stream via a cancel. Seems a little odd since it would be indistinguishable from the command running the buffer to completion but that is how it works so IMO we should follow the same pattern.

We could print a cancel warning like build if we want to be a little more user friendly, which would go to stderr. I dont feel strongly about that but my default disposition is to follow docker exactly unless we intentionally want to deviate. I believe better user experience is a good reason to deviate though.

beena352: The runtime returns E_ABORT on cancellation, but the CLI handles it as normal completion when its own cancellation event is signaled. User cancellation therefore exits successfully without an error or warning; aborts unrelated to user cancellation still propagate.

dkbennett (approved):

src/windows/wslcsession/EventStore.cpp
```diff
@@ -12,14 +12,14 @@ namespace wsl::windows::service::wslc {
...
+    bool IsSignaled(HANDLE Handle)
     {
-        if (CancelEvent == nullptr)
+        if (Handle == nullptr)
         {
             return false;
         }
 
-        const auto result = WaitForSingleObject(CancelEvent, 0);
+        const auto result = WaitForSingleObject(Handle, 0);
```
OneBlue, line 22: nit: I recommend passing in an array of events and using WaitForMultipleObjects() to wait check all of the events at once. 

In a future change, it would be nice to also use an event to signal more events being available without having to set threadpool waits, but I think that's fine for this change and we can implement that in a future iteration

OneBlue (approved):
LGTM, feel free to do followups in future changes

src/windows/wslc/tasks/SessionTasks.cpp
```diff
@@ -97,6 +97,22 @@ void ListSessions(CLIExecutionContext& context)
...
 }
 
+void StreamEvents(CLIExecutionContext& context)
+{
+    using namespace std::chrono;
+
+    WI_ASSERT(context.Data.Contains(Data::Session));
+    auto& session = context.Data.Get<Data::Session>();
+
+    const auto now = floor<seconds>(system_clock::now()).time_since_epoch().count();
```
OneBlue, line 107: Related to my comment in #41576 , I recommend having a special value to represent "now" instead of doing this, since this relies on clock synchronization across processes, which can lead to surprising behaviors

PR #41618, opened 2026-09-15
Add Rust client for the OpenVMM VM service

damanm24:
This change consumes the following collateral from the Wsl-deps repository:
- Microsoft.WSL.OpenVMM nuget package which contains:
- OpenVMM
- wslopenvmm.dll and wslopenvmm.h (a Rust based dll that provides an RPC client to communicate with OpenVMM and corresponding header file definitions)


CMakeLists.txt
```diff
@@ -572,6 +570,31 @@ set(TAEF_LINK_LIBRARIES
     Wex.Common.lib
     Wex.Logger.lib)
 
+# Only enable OpenVMM when its Rust target is available.
+find_program(CARGO_EXECUTABLE cargo)
+if (CARGO_EXECUTABLE)
```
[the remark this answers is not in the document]
damanm24: This is now being checked into a feature branch, so our assumptions/requirements can be relaxed here.

CMakeLists.txt
```diff
@@ -276,6 +277,16 @@ if (NOT WSL_DEVICEHOST_BIN)
     file(CREATE_LINK ${WSL_DEVICE_HOST_SOURCE_DIR}/bin/${TARGET_PLATFORM}/wsldevicehost.pdb ${BIN}/wsldevicehost.pdb)
 endif()
 
+foreach(binary openvmm.exe openvmm.pdb wslopenvmm.dll wslopenvmm.pdb)
+    file(CREATE_LINK "${WSL_OPENVMM_SOURCE_DIR}/bin/${TARGET_PLATFORM}/${binary}" "${BIN}/${binary}")
```
OneBlue, line 281: nit: I would recommend skipping the symlink and consuming those directly from the nuget folder in `msipackage/package.wix.in`

CMakeLists.txt
```diff
@@ -276,6 +277,16 @@ if (NOT WSL_DEVICEHOST_BIN)
...
+endforeach()
+
+add_library(wslopenvmm_client SHARED IMPORTED GLOBAL)
```
OneBlue, line 284: Do we need to create a target for this library ? I was originally thinking of just adding the include path & updating the linking options in wslservice / wslcsession

OneBlue (approved):

PR #41619, opened 2026-09-15
Move objects shared by wslc.exe and the tests to src/windows/common

ggarzia-MSFT:
Moves 8 units (16 files) that are shared by wslc.exe and wsltests.dll out of src/windows/wslc/ and into src/windows/common, so they build once into common.lib instead of being linked twice. This also makes it easier to use objects such as Terminal and TableOutput in WSL if needed in the future.


dkbennett (commented):
We should also update the namespaces for these to be under common as well.

For example, Terminal class is under namespace wsl::windows::wslc, that should be changed to wsl::windows::common and updated accordingly in consuming code.

It may also be worth considering whether we want the "common" namespace, or if "wsl::windows" is sufficient to keep the namespaces from getting too long, or if we want something more specific, like "wsl::windows::cli" for things like Terminal, InputChannel, OutputChannel, TableOutput since the are all related to common CLI-related tooling, but I don't feel strongly about it. "common" seems either redundant to the 'windows' parent namespace or too generic and would prefer something more specific or omitting the generic.

ggarzia-MSFT:
> We should also update the namespaces for these to be under common as well.
> 
> For example, Terminal class is under namespace wsl::windows::wslc, that should be changed to wsl::windows::common and updated accordingly in consuming code.
> 
> It may also be worth considering whether we want the "common" namespace, or if "wsl::windows" is sufficient to keep the namespaces from getting too long, or if we want something more specific, like "wsl::windows::cli" for things like Terminal, InputChannel, OutputChannel, TableOutput since the are all related to common CLI-related tooling, but I don't feel strongly about it. "common" seems either redundant to the 'windows' parent namespace or too generic and would prefer something more specific or omitting the generic.

I would prefer to go with wslc::windows::cli just so that we are a bit more specific with the namespace naming

benhillis (approved):

PR #41637, opened 2026-09-17
Fix various protocol parsing issues

OneBlue:

This change solves 3 parsing bugs: 

1) Reject 'clientId' strings whose length extend passed the end of the message in Plan9's TGetLock / RGetLock handler

2) Reject truncated Plan9 TRead requests

3) Switch string arrays to using the ArrayFromSpan format instead of null separated strings. The previous implementation could read passed the message length in multiple places

This change also add a bit of logic in the plan9 entrypoint to run it ad-hoc, which allows to run unit tests on the plan9 server 

## PR Checklist


OneBlue:
> * This test hard-codes 127.0.0.1:1234, so it fails whenever another process or concurrently running WSL test owns that port. Please allocate an ephemeral port (and report/use the port returned by bind) or otherwise choose a free port so the protocol tests do not depend on the host environment.

That's fine, the tests do this in many places

src/windows/service/exe/LxssCreateProcess.h
```diff
@@ -71,20 +66,19 @@ class LxssCreateProcess
...
-            ArgumentsData.insert(ArgumentsData.end(), *e, *e + strlen(*e) + 1);
+            ++argumentCount;
         }
 
-        ArgumentsData.emplace_back('\0');
-
         wsl::shared::MessageWriter<CREATE_PROCESS_MESSAGE> message(LxInitCreateProcess);
         message.WriteString(message->PathIndex, Path);
-        gsl::copy(as_bytes(gsl::span(ArgumentsData)), message.InsertBuffer(message->CommandLineIndex, ArgumentsData.size()));
+        message.WriteStringArray(message->CommandLineIndex, Arguments, argumentCount);
```
dkbennett, line 81: this looks a lot simpler than the former code!

dkbennett (approved):

benhillis (approved):

PR #41640, opened 2026-09-17
Use the latest ESRP signing key code

OneBlue:
## PR Checklist


ranm-msft (commented):
Verified the key rotation is complete on this branch: I fetched both pipeline files at head f22400c and counted occurrences - build-job.yml has 4x CP-501332 and 0x CP-230012, package-stage.yml has 2x CP-501332 and 0x CP-230012. So all six sign/verify operations moved together and nothing is left straddling the old key.

Not approving only because I can't independently confirm CP-501332 is the correct ESRP key for this product - that part needs someone with the signing-config source of truth. If that's already confirmed, this looks ready to go from a completeness standpoint.

benhillis (approved):

